PR #10, opened 2017-05-25
Implement basic web browser

fgei:
Reviewer: @subsymbolic 
Asana: https://app.asana.com/0/0/341403501734105

**Description**: Remove all the old code and implement a basic web browser with a minSdk 16, latest build tools and support libraries, MVP implementation of the presentation layer to be able to test the business logic and zero lint error and warnings

**Steps to test this PR**:
1. travis will build and run all the test and should fail if there are tests failing or any lint error or warning

###### Reviewer Checklist:
- [x] **Ensure the PR does no harm by testing the changes thoroughly**

###### PR DRI Checklist:


app/src/main/res/layout/appbar.xml
```diff
@@ -1,146 +1,53 @@
...
+                android:layout_marginRight="8dp"
+                android:paddingStart="8dp"
+                android:paddingLeft="8dp"
+                android:paddingEnd="8dp"
+                android:paddingRight="8dp"
+                android:selectAllOnFocus="true"
+                android:maxLines="1"
+                android:inputType="textNoSuggestions"
+                android:imeOptions="actionSearch"
+                android:textColor="#000"
```
subsymbolic: Move all color references into colors.xml

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/browser/BrowserFragment.java
```diff
@@ -0,0 +1,262 @@
...
+                return false;
+            }
+        });
+    }
+
+    private void initSearchEditText(EditText searchEditText) {
+        searchEditText.setOnEditorActionListener(new TextView.OnEditorActionListener() {
+            @Override
+            public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
+                if(actionId == EditorInfo.IME_ACTION_SEARCH) {
```
subsymbolic: It might be worth responding to the enter key too so that we play well with attached keyboards. You can capture this with`KeyEvent.ACTION_DOWN` and `KeyEvent.KEYCODE_ENTER` and you could wrapp that in a boolean function to keep things readable. Maybe something like this?
```
private boolean wasEnterPressed(KeyEvent event) {
     return event.getAction() == KeyEvent.ACTION_DOWN && event.getKeyCode() == KeyEvent.KEYCODE_ENTER;
 }
```

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/browser/BrowserPresenterImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        browserView.goBack();
+    }
+
+    @Override
+    public void refreshCurrentPage() {
+        browserView.reload();
+    }
+
+    @Override
+    public void onPageStarted(@Nullable String url) {
```
fgei: Since this is called directly and only from WebViewClient.onPageStarted(WebView webView, String url, Bitmap favicon), shouldn't this only be an url? (or this method can get called by something other then an url?)

fgei:
with the exception for the uri class and a complete format code this should all be addressed and fixed!

I can't get to crash the app pressing the back button, any info on how to reproduce that?

subsymbolic:
Great! Regarding the crash it's a libc error so there isn't much of a stack trace. I see it consistently when pressing back on my Nexus 5 API 23 emulator, give it a try on an emulator too and let me know if you see. If not, we can assume mine is broken.

I'm also noticing that webpages aren't loading on API 24 and 25 emulators (the progress bar activates but the webpage section stays blank). Do you see this too?

fgei:
ok I could reproduce the crash when pressing back on the emulator (nexus 5 API 23) but not on a real nexus 5 API 23 that I use as a base test device! fixed now.

fixed also the webpage not loading on api >= 24 (my bad I returned true on shouldOverrideUrlLoading)

removed all the .idea/ files form the repo and re-format all code

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/browser/BrowserPresenterImpl.java
subsymbolic: Ahh yes you're right, url is always correct by this point.

subsymbolic:
Nice, I can see that the browser now loads pages on API 24 & 25. As for the crash I'm still seeing it on API 23 emulators (Nexus and Nexus 5X) but if it can't be reproduced on a real device I'm happy to assume it is an emulator only problem for now.

subsymbolic:
Looks good @fgei, feel free to merge (or squash) when you're ready.

fgei:
@subsymbolic ok will keep an eye out for the crash on the emulator!

PR #17, opened 2017-06-13
Bookmarks

fgei:
Reviewer: @subsymbolic 
Asana: https://app.asana.com/0/0/353085934262754
CC:

**Description**: Implement basic bookmarks in the app

**Steps to test this PR**:
Test 1
1. Load web page
2. Select Save Bookmark from the menu
3. Open Bookmarks and see the just saved bookmark

Test 2
1. Save a few bookmarks
2. Try to open each bookmark and confirm the browser loads the right url for each bookmark

Test 3
1. Press Edit on the Bookmark menu
2. Delete some bookmark by pressing the delete button or swiping it.
3. Confirm that the bookmark are properly deleted

Test 4
1. Press Edit on the Bookmark menu
2. Select a single Bookmark and edit the proper info
3. Confirm that the bookmark is edited, saved properly and open the right url

Test 5
1. Force close the app
2. Confirm the bookmarks are preserved when reopening the app

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR DRI Checklist:


app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/bookmarks/itemtouchhelper/DragItemTouchHelperCallback.java
```diff
@@ -0,0 +1,70 @@
...
+import android.support.v7.widget.RecyclerView;
+import android.support.v7.widget.helper.ItemTouchHelper;
+
+/**
+ * Created by fgei on 6/9/17.
+ */
+
+public class DragItemTouchHelperCallback extends ItemTouchHelper.Callback {
+
+    public static final float ALPHA_FULL = 1.0f;
```
subsymbolic: Make private

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/domain/bookmark/Bookmark.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+public interface Bookmark {
+    String getId();
+
+    String getName();
+
+    String getUrl();
+
+    int getIndex();
```
subsymbolic: Ok, I see why now, I'm happy for you to keep the index if you need it.

fgei: I thought it was the best way to deal with ordering the list, I see on iOS you can drag and switch items and the order is preserved when relaunching the app, since in the SharedPreference we are storing just "key - value", I though that was the easiest way to preserve that. If you have a cleaner idea, I'm open to change it and remove the index

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/bookmarks/BookmarkModel.java
```diff
@@ -0,0 +1,99 @@
...
+
+import com.duckduckgo.mobile.android.duckduckgo.domain.bookmark.Bookmark;
+
+import java.util.UUID;
+
+/**
+ * Created by fgei on 6/12/17.
+ */
+
+public class BookmarkModel implements Bookmark, Parcelable {
```
fgei: I see other projects tend to add the -model suffix for the presentation layer and the -entity suffix for the data layer. I agree that the model suffix is not really readable and definitely annoying to use, we could call the interface just IBookmark and change BookmarkModel to simply Bookmark but I'm not a fan of altering names for interfaces. In this case also I don't think we want the same suffix, assume we are adding another data source for the bookmark like a database, then the BookmarkJsonEntity name is not really appropriate I think

fgei: we could also just remove the interface and rename the BookmarkModel -> Bookmark and keep BookmarkEntity

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/domain/bookmark/Bookmark.java
subsymbolic: No, keep it, I think it's valid in this case. As I read on, I better saw why you used the approach you did.

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/bookmarks/BookmarkModel.java
subsymbolic: I'm with you on the on avoiding the "I" prefix for interfaces :)

PR #18, opened 2017-06-15
Tabs

fgei:
Reviewer: @subsymbolic 
Asana: https://app.asana.com/0/0/350058423386223
CC:

**Description**: Implement basic tabs
Known bad behavior: when closing all tabs in the tab switcher the tab automatically creates a new tab, this will be fixed in the Navigation Task (https://app.asana.com/0/0/361925093999571)

**Steps to test this PR**:
Test 1
1. Use the tab switcher to create new tabs and switch
2. Confirm that tabs are preserved when switching back

Test 2
1. Create a few tabs with some log in.
2. Click the forget button in the tab switcher
3. Access the previous URL with log in and see that the credentials are erased

Test 3
1. Create a few tabs
2. Make the system kill the process
3. Open the app and confirm that the session is restored

Test 4
1. Create a few tabs
2. Swipe the app from the system app switcher
3. Open the app and confirm that the session is not restored

###### Reviewer Checklist:
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**

###### PR DRI Checklist:


app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/data/base/BaseSharedPreferencesDataStore.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public List<T> getAll() {
+        List<T> out = new ArrayList<>();
+        for (Object item : sharedPreferences.getAll().values()) {
+            try {
+                T t = clss.newInstance();
+                t.fromJson(item.toString());
+                out.add(t);
+            } catch (InstantiationException e) {
+                e.printStackTrace();
```
fgei: ok, I'll keep this and remove everything in a different branch as per the new task

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/data/base/BaseSharedPreferencesDataStore.java
```diff
@@ -0,0 +1,62 @@
...
+            } catch (IllegalAccessException e) {
+                e.printStackTrace();
+            }
+        }
+        return out;
+    }
+
+    public void insert(JsonEntity jsonEntity) {
```
fgei: ah yes, otherwise what's the point of using generics? :)

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/data/base/JsonEntity.java
```diff
@@ -0,0 +1,17 @@
+package com.duckduckgo.mobile.android.duckduckgo.data.base;
+
+/**
+ * Created by fgei on 6/12/17.
+ */
+
+public abstract class JsonEntity {
```
fgei: I prefer the interface too, but we need the empty constructor to ensure SharedPreferencesDataStore can instantiate the JsonEntity. With just an interface can we be sure each entity will have a empty constructor? (I'm asking because I was stack a few minutes with the store not working correctly)

fgei:
Everything should be addressed now, bugs were due to the refactor, and some missing tests, I'll try to think about it and add some that will test this scenario.

The progress bar was acting like this even before this PR, however it should be fixed now!

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/data/base/JsonEntity.java
subsymbolic: No an interface won't force an empty constructor. If you get an early failure (e.g an exception is thrown) and you write a test to demand an empty constructor for each entity, then I'm ok with that. If you really want to keep it abstract go for it, it's a compromise either way.

Another option you have is rather than passing the class type into shared preferences, you could pass a lightweight object responsible for instantiating the entity. This would internally knows the correct type so you would need to build one for each type. That  keeps the code flexible and removes your dependence on a empty constructor.

subsymbolic:
@fgei I've added my final thoughts to the abstract class versus interface discussion but I'm happy with whatever you decide. Great work, merge when you're ready.

app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/data/base/JsonEntity.java
fgei: @subsymbolic so something like this?

`public class SharedPreferencesDataStore<T extends JsonEntity> {

    private SharedPreferences sharedPreferences;
    private EntityCreator<T> entityCreator;

    public static abstract class EntityCreator<T> {
        public abstract T newInstance();
    }

    public SharedPreferencesDataStore(Context context, String fileName, EntityCreator<T> entityCreator) {
        sharedPreferences = context.getSharedPreferences(fileName, Context.MODE_PRIVATE);
        this.entityCreator = entityCreator;
    }

    public List<T> getAll() {
        List<T> out = new ArrayList<>();
        for (Object item : sharedPreferences.getAll().values()) {
            T t = entityCreator.newInstance();
            t.fromJson(item.toString());
            out.add(t);

        }
        return out;
    }
 ...
}

public class BookmarkSharedPreferences extends SharedPreferencesDataStore<BookmarkJsonEntity> {

    private static final String PREF_NAME = "bookmarks";

    public BookmarkSharedPreferences(Context context) {
        super(context, PREF_NAME, new EntityCreator<BookmarkJsonEntity>() {
            @Override
            public BookmarkJsonEntity newInstance() {
                return new BookmarkJsonEntity();
            }
        });
    }
}`

subsymbolic: Looks great, I'd call the method `create` to avoid any confusion with the class level `newInstance` method

subsymbolic: Optional: If it doesn't cause you any problems, make the EntityCreator an interface too

fgei: and yes to both of them! :)

PR #20, opened 2017-06-21
Omnibar

fgei:
Reviewer: @subsymbolic 
Asana: https://app.asana.com/0/0/364465928695402

**Description**: Update the Omnibar UI to latest iOS build and latest Zeplin iOS design file

**Steps to test this PR**:
1. Open the app and see that the Omnibus UI changed.
2. Confirm that the overflow button still open the main menu.
3. Confirm that when pressing on the Omnibar, the back arrow become visible and when pressed bring the Omnibar to the previous state.
4. Confirm that when pressing on the Omnibar the "cancel text" button become visible only after start typing and goes aways after removing all the letters, and when pressing the button correctly delete all the letters. 

###### Reviewer Checklist:
- [x] **Ensure the PR does no harm by testing the changes thoroughly**

###### PR DRI Checklist:

### DIFF
--- a/app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/bookmarks/BookmarksActivity.java
+++ b/app/src/main/java/com/duckduckgo/mobile/android/duckduckgo/ui/bookmarks/BookmarksActivity.java
@@ -165,7 +165,7 @@ public class BookmarksActivity extends AppCompatActivity implements BookmarksVie
 
     private void initToolbar() {
         toolbar.setTitle(R.string.bookmarks_toolbar);
-        toolbar.setNavigationIcon(R.drawable.ic_arrow_back_black_24dp);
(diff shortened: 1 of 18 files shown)


subsymbolic:
Great work @fgei. Only one request - can you update the light on light text and status bar icons to black so that they are legible? You don't need to fully theme the affected screens just do what needs to be done to ensure that the user is never given a lesser experience from a new PR.

fgei:
@subsymbolic should be done now!

subsymbolic:
Looks good, although the status bar is still light on light. Feel free to merge once you've fixed that.

fgei:
yep @subsymbolic sorry, I forgot that "windowLightStatusBar" is only available from API 23 but the status bar is already tinted from API 21. in those case we still get light text on light background.

I set to tint the status bar background only from API 23, and leave the black status bar before that

PR #42, opened 2017-07-26
Add Brazilian Portuguese Translate (PT-BR)

thalezirnho:
New translate without the error: "Unpacking replacement libc6:i386"


thalezirnho:
What's the problem? The same as the previous?

subsymbolic:
Thanks for the contribution @thalezirnho. Unfortunately, we're not accepting PRs at the moment. That will hopefully change in the future - watch for it on our readme page.

PR #51, opened 2017-11-20
Feature/webview improvements

CDRussell:
Asana:
https://app.asana.com/0/471012189846020/481692274328411

**Description**:
Improve the browsing experience by making the following changes
1. Remove `NestedScrollView` as a wrapper around the `WebView` as it completely messed with the browser experience.
1. Introducing a `WebView` subclass which allows the toolbar to still react to swipes. This is basically a straight Kotlin port of the java implementation provided here https://github.com/takahirom/webview-in-coordinatorlayout. It's not perfect, but might be a good base to build upon. The code in here could do with tidying; as I say, a straight port of that repo's version. 
1. Instructing Android not to destroy the `BrowserActivity` on common config changes. The Activity getting destroyed was the main reason why the `WebView` didn't handle rotation etc... well
1. Better support pinch to zoom by hiding the vile default zoom controls that appear
1. Removing swipe to refresh on the web view. Would be nice to do this, but it doesn't play nice with the nested web view approach, so will require a little more work to implement. Will be handled here https://app.asana.com/0/414730916066338/483384580258310

**Steps to test this PR**:
1. Load page
1. Rotate
    - Ensure the WebView doesn't completely reload the web page and resizes appropriately
1. Scroll page up and down
    - Ensure the toolbar hides and reappears as expected
1. Pinch to zoom
    - Ensure the web page zooms in/out accordingly


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -111,10 +103,15 @@ class BrowserActivity : DuckDuckGoActivity() {
     private fun configureWebView() {
         webView.webViewClient = webViewClient
         webView.webChromeClient = webChromeClient
-        webView.settings.javaScriptEnabled = true
-        webView.settings.builtInZoomControls = true
-        webView.settings.setSupportZoom(true)
-        webView.settings.useWideViewPort = true
+
+        webView.settings.apply {
```
subsymbolic: Nice!

app/src/main/java/com/duckduckgo/app/browser/NestedWebView.kt
```diff
@@ -0,0 +1,119 @@
...
+    constructor(context: Context) : this(context, null)
+    constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
+
+    private var lastY: Int = 0
+    private val scrollOffset = IntArray(2)
+    private val scrollConsumed = IntArray(2)
+    private var nestedOffsetY: Int = 0
+
+    private val mChildHelper: NestedScrollingChildHelper = NestedScrollingChildHelper(this)
```
subsymbolic: Are we planning to maintain this as our own over time?  If so can we use our own style guide e.g remove the `m` prefix from properties. If not, I'd suggest putting it in the a third party package to hide it away.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -88,7 +89,7 @@ class BrowserWebViewClient @Inject constructor(private val requestRewriter: Duck
     /**
      * Access the webview url from another thread, jumps onto the main thread to achieve this
```
subsymbolic: Optional: Update the comment to match? e.g change it to `from another thread` to `from any thread`

app/src/main/java/com/duckduckgo/app/browser/NestedWebView.kt
CDRussell: Updated to use our own style.

I'm not sure if this class will be needed longer term. I suspect it'll either disappear completely, or have more work done to it. Don't think it'll live forever as is, so have tidied up a little.

PR #57, opened 2017-11-24
Feature/add home screen

CDRussell:
Asana Issue URL: https://app.asana.com/0/471012189846020/364465928695404

## Description
First pass at a home screen / browser landing page.

There are a few unanswered questions around navigation and UX that will likely result in further changes. However, as discussed, let's get this implementation in now while we're working out the answers to those questions separately. 

We'll _likely_ want to move the two separate Activities (BrowserActivity and HomeActivity) into a single Activity in future, and treat the _landing page_ and _browser view_ as two separate states within.

## Steps to Test this PR:
1. Run the app; note the new home screen
1. Tap the EditText; ensure you transition to allow you to enter your query
1. Test back navigation works as expected (for now, meaning you **can** return to the landing page)
1. Test that if you enter a search query (not a full URL) then that search query is shown and remains editable in the omnibar.
1. Test that once you navigate to a search result (or enter a full URL) the omnibar updates to show the URL and not a search term


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -40,10 +43,16 @@ class BrowserActivity : DuckDuckGoActivity() {
...
     @Inject lateinit var viewModelFactory: ViewModelFactory
 
+    private var acceptingRenderUpdates = true
+
     private val viewModel: BrowserViewModel by lazy {
         ViewModelProviders.of(this, viewModelFactory).get(BrowserViewModel::class.java)
     }
 
+    companion object {
+        fun starter(context: Context): Intent = Intent(context, BrowserActivity::class.java)
```
CDRussell: Rename function to simply `intent()`

subsymbolic:
When dismissing the keyboard on the home screen can we also transition back to the "inactive" state (with the logo and search in the middle).

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -62,27 +71,46 @@ class BrowserActivity : DuckDuckGoActivity() {
     }
 
     private fun render(viewState: BrowserViewModel.ViewState) {
+
+        if (!acceptingRenderUpdates) return
+
+        supportActionBar?.setDisplayHomeAsUpEnabled(false)
```
CDRussell: This probably shouldn't be done in here

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
```diff
@@ -35,13 +35,24 @@ class BrowserViewModel(
         private val queryUrlConverter: OmnibarEntryConverter,
         private val trackerDataProvider: TrackerDataProvider,
         private val trackerDetector: TrackerDetector,
-        private val trackerListService: TrackerListService) :
+        private val trackerListService: TrackerListService,
+        private val duckDuckGoUrlDetector: DuckDuckGoUrlDetector) :
         WebViewClientListener, ViewModel() {
 
+    data class ViewState(
```
CDRussell: Make `ViewModel` responsible for deciding what appears in the URL bar; don't let the `Activity` decide

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
CDRussell: Good team convention, i like it!

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
subsymbolic: I like that you jotted down my random stream of consciousness earlier and turned it into real feedback :)

PR #60, opened 2017-11-28
Feature/detecting keyboard dismissal

CDRussell:
Asana Issue URL: https://app.asana.com/0/471012189846020/487962201286661

## Description
Added custom `EditText` subclass which is able to detect when the keyboard is being dismissed using the back button. Using this, we can react to when the user hits to dismiss the keyboard but hasn't searched for anything yet, and navigate them back to the landing page.

This PR also tidies up some of the UX around focussing and keyboard showing. You should no longer have to tap twice on the URL to get the keyboard to appear.

## Steps to Test this PR:
1. From landing page, tap the text box. Dismiss the keyboard, ensuring you return to the landing page
1. From landing page, tap the text box. Tap outside of the omnibar. Ensure you are returned to the landing page
1. From landing page, tap the text box. type a search and hit enter. Repeat the 2 steps above and ensure you are NOT returned to the landing page.


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -64,6 +65,14 @@ class BrowserActivity : DuckDuckGoActivity() {
             it?.let { webView.loadUrl(it) }
         })
 
+        viewModel.navigation.observe(this, Observer {
+            when (it) {
+                BrowserViewModel.NavigationCommand.LANDING_PAGE -> {
```
subsymbolic: Optional: It took me a moment to process this perhaps the when/switch syntax is too much for a single option? I think an simple if with a static import is more readable.
```
           if (it == LANDING_PAGE) {
                finishActivityAnimated()
            }
```

CDRussell:
>One observation, when dismissing the the keyboard on the landing page our animation and the keyboard animation happen at the same time and glitch

Yeah, agreed that visually, this could be nicer. It would be great to dismiss the keyboard with no animation, or react to the keyboard animation finishing, but neither seem to be trivial to achieve. 

Probably niche enough of a case to forget about, I reckon.

subsymbolic:
If improving the keyboard animation isn't trivial let’s not burn time on it now. Let's create an asana task and put it at the bottom of the android list to see if we can fit it in before release

subsymbolic:
This main keyboard interactions are working really well now. Great work, feel free to merge.

CDRussell:
Separate ticket for the keyboard/back animation: https://app.asana.com/0/414730916066338/490474714901190

PR #62, opened 2017-11-29
Remove touch listener from root view; use dedicated webview container

CDRussell:
Asana Issue URL: https://app.asana.com/0/471012189846020/488688451095619

## Description
Listening to touches on the root view resulted in an oddity when swiping down from the url input area. This should be fixed now.

## Steps to Test this PR:
1. Swipe down from the URL bar
1. Ensure you are NOT navigated away from the browser/search views
1. Ensure the toolbar does NOT overlap any of the web view content


subsymbolic:
This unfortunately causes the omnibar to overlap the webview. We went through this exact sequence of issues on iOS too! omnibar_overlaps

CDRussell:
Toolbar should no longer overlap the webview.

PR #64, opened 2017-12-01
Do not block first party sites

subsymbolic:
Asana Issue URL: 
https://app.asana.com/0/488551667048375/487607508203633

## Description
Stops content blocker from blocking first party sites

## Steps to Test this PR:
TEST ONE
1. Open the app, wait a moment for networking to complete and lists to be downloaded (sync service is a later feature)
1. Visit `twitter.com` and note from the logs that `twitter.com` (the site we are visiting) and `twimg.com` (an associated url in the disconnect list) are no longer blocked

TEST TWO
1. Now visit `subsymbolic.net`
1. Note from the logs that `platform.twitter.com` is blocked

NOTE: A user click on a javascript link to a tracker should technically be considered "first-party" as the user has requested that url. There is a possibility that this edge case is not covered here. I have captured that in https://app.asana.com/0/414730916066338/492564757947847


app/src/androidTest/java/com/duckduckgo/app/trackerdetection/TrackerDetectorInstrumentationTest.kt
```diff
@@ -16,62 +16,107 @@
...
 
-@RunWith(AndroidJUnit4::class)
 class TrackerDetectorInstrumentationTest {
 
+    private val networkTrackers = NetworkTrackers()
+    private val trackerDetector = TrackerDetector(networkTrackers)
+
     companion object {
-        private val documentUrl = "http://example.com"
         private val resourceType = ResourceType.UNKNOWN
```
brindy: What's the thinking with resourceType? It's passed in as a parameter, but never used (for instance, if I change the code to ignore the passed in value and use ResourceType.IMAGE the tests still pass.

Is it really needed?

app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDetector.kt
```diff
@@ -36,6 +39,25 @@ class TrackerDetector @Inject constructor() {
...
     fun shouldBlock(url: String, documentUrl: String, resourceType: ResourceType): Boolean {
-        return clients.any { it.matches(url, documentUrl, resourceType) }
+
+        if (firstParty(url, documentUrl)) {
+            Timber.v("$url is a first party url")
+            return false
+        }
+
+        val matches = clients.any { it.matches(url, documentUrl, resourceType) }
+        val matchText = if (matches) "WAS" else "was not"
```
brindy: This line won't get stripped by ProGuard (assuming we plan to do that) and it seems to be only for the purpose of logging.

subsymbolic: Can you explain why this is a problem? Are you worried about efficiency? Security? don't see an issue with a convenience string containing the words "was" or "was not" existing here.

brindy: Just that it will become dead code eventually and I personally feel a better habit would be to avoid that, but if you're not bothered it's not that big a deal.

app/src/androidTest/java/com/duckduckgo/app/trackerdetection/TrackerDetectorInstrumentationTest.kt
subsymbolic: Good question. It is needed however it is not to relevant at this level, if it were a more complex type I would mock it. It is passed to the tracker clients and adblock specifically needs it to match rules correctly (fyi the problems that iOS and the extension are having with the tacker detection are around this area). 

The ResourceType itself is tested in ResourceTypeInstrumentationTest, the current behaviour there is an improved version of what brave has on iOS although we may expand and improve on it over time. An integration test suite (like the one the extension team are writing) would be a good addition for both android and ios to validate that the third party components we're using are working correctly. That goes well beyond the scope of this PR so I'll add a separate task for that.

app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDetector.kt
subsymbolic: I would use a ternary if they were available in kotlin... but sadly not. Do you know if this gets optimised away at the bytecode level - it would in java just I can't find anything on android. Please share if you have a reference. More curious than anything, I need the variable so as you're not adamant that it needs to go I'll keep it either way.

subsymbolic:
@brindy I've answered your resource questions above! Let me know if you need anything else for this PR.

app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDetector.kt
brindy: Do you know what, ProGuard probably strips dead code if the compiler doesn't optimise it out.

subsymbolic:
Thanks for the preview @brindy!

PR #66, opened 2017-12-05
Major tracker networks summary

subsymbolic:
Asana Issue URL: https://app.asana.com/0/488551667048375/471718733961910

NOTE: There is an outstanding question around the maxcdn network covered in https://app.asana.com/0/488551667048375/494979307307237

## Description
Add major networks summary

## Steps to Test this PR:
1. Visit evanscycles.co.uk
1. Open the privacy dashboard (it will be a little slow as we haven't implemented live data yet and are just serializing to share data from one activity to the next)
1. Ensure that major tracker network count is displayed


app/src/main/java/com/duckduckgo/app/trackerdetection/model/NetworkTrackers.kt
```diff
@@ -37,4 +57,13 @@ class NetworkTrackers @Inject constructor() : Serializable {
                 .map { it.networkName }
                 .firstOrNull()
     }
+
+    fun majorNetwork(url: String): String? {
+        return majorNetworks
+                .filter { network(url).equals(it.name, true) }
+                .map { it.name }
+                .firstOrNull()
```
brindy: This could probably be simpler and more efficient using a Kotlin array for majorNetworks and then use `.find` , rather than filtering -> map -> find first or null, e.g.  line 34 could be:

    majorNetworks = arrayOf(

then line 62 could be:

    return majorNetworks.find { network(url).equals(it.name, true)  }?.name

see https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/find.html

subsymbolic:
Thanks for the review Chris!

PR #72, opened 2017-12-11
Allow privacy protection to be toggled on and off

subsymbolic:
Asana Issue URL: https://app.asana.com/0/414730916066338/487581092466119

## Description
Allows the user to toggle privacy protection (really just tracker blocking at this stage) on and off

## Steps to Test this PR:
TEST ONE:
1. Enter and submit a search in the omnibar
1. Open the privacy dash
1. Note that the toggle is on by default
1. Switch it off and note that the tracker icons are all still green as the detection count is zero
1. Leave the dashboard and reenter it, note the toggle is still off
1. Switch the toggle on again

TEST TWO:
1. Go to cnn.com
1. Open the privacy dash
1. Switch it off and note that the tracker icons are now red
1. Leave the dashboard and note that cnn.com is reloaded with blocking off


brindy:
Possible wee UI bug: 

* Go to a site
* Open the privacy dashboard
* Turn off privacy 
* Things turn red
* Hit back/home
* The UI appears to reset to zero/green before closing

Maybe a timing issue because as you go back the webview refreshes and sends new live data to the UI that is now disappearing? Just looks a bit weird.

subsymbolic:
Nice catch, I'm not seeing this on my emulators or real devices (Samsung s7 edge, Nexus 5). Which device are you seeing it on? I'll try and reproduce it.

subsymbolic:
@brindy I managed to reproduce that issue on my emulator after all. LiveData objects are meant to be lifecycle aware however it looks like updates are still sent while the activity is in the process of finishing. I've added a check for that.

app/src/main/java/com/duckduckgo/app/privacymonitor/ui/PrivacyDashboardViewModel.kt
```diff
@@ -0,0 +1,151 @@
...
+    private fun resetViewState() {
+        viewState.value = ViewState(
+                domain = "",
+                heading = headingText(),
+                httpsIcon = httpsIcon(HttpsStatus.SECURE),
+                httpsText = httpsText(HttpsStatus.SECURE),
+                networksIcon = R.drawable.dashboard_networks_good,
+                networksText = networksText(),
+                majorNetworksText = majorNetworksText(),
+                majorNetworksIcon = R.drawable.dashboard_networks_good,
```
brindy: wrong icon

app/src/main/java/com/duckduckgo/app/privacymonitor/ui/PrivacyDashboardViewModel.kt
```diff
@@ -0,0 +1,151 @@
...
+            val heading: String,
+            val httpsIcon: Int,
+            val httpsText: String,
+            val networksText: String,
+            val networksIcon: Int = R.drawable.dashboard_networks_good,
+            val majorNetworksText: String,
+            val majorNetworksIcon: Int = R.drawable.dashboard_major_networks_good,
+            val toggleEnabled: Boolean,
+            val toggleText: String,
+            val toggleBackgroundColor: Int = R.color.midGreen
```
brindy: I would suggest using the `@*Res` annotations here to give compile time checking of the resource type (e.g. `@ColorRes`  if they work (don't seem to be for me)

subsymbolic:
Ready for rereview @brindy

app/src/main/java/com/duckduckgo/app/privacymonitor/ui/PrivacyDashboardViewModel.kt
subsymbolic: Good thinking

brindy:
LGTM :)

PR #75, opened 2017-12-18
Adds privacy grade to the dashboard

subsymbolic:
Asana Issue URL: https://app.asana.com/0/488551667048375/471718733961913

## Description
Shows privacy grade and upgrade details on the privacy dashboard

## Steps to Test this PR:
TEST ONE
1. Visit netflix.com
1. Ensure the privacy toggle is on
1. Note that the grade is A and that the site was upgraded from a B to an A
1. Switch the toggle off, close privacy grade and reopen it
1. Note the grade is just B now

TEST TWO
1. Visit duckduckgo.com
1. Ensure the privacy toggle is on
1. Note the grade is A and there is no upgrade text

TEST THREE
1. Visit cnn.com
1. Ensure the privacy toggle is on
1. Note that the grade is C and that the site was upgraded from a D to a C
1. Switch the toggle off, close privacy grade and reopen it
1. Note the grade is just D now


brindy:
Possible bug: There's a slight discrepancy between iOS and Android here when comparing the grade for Netflix with protection off.  iOS only uses a tracker count of 11 when calculating the score, but Android seems to be using a tracker count of 20+ (thus giving it +1 compared to iOS).

app/src/main/java/com/duckduckgo/app/privacymonitor/ui/PrivacyMonitorGradeExtension.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.duckduckgo.app.privacymonitor.PrivacyMonitor
+import com.duckduckgo.app.privacymonitor.model.PrivacyGrade
+import com.duckduckgo.app.privacymonitor.model.TermsOfService
+
+
+val PrivacyMonitor.score: Int
+    get() {
+        var score = baseScore
+        score += Math.ceil(trackerCount / 10.0).toInt()
+        if (majorNetworkCount > 0) {
```
brindy: iOS doesn't do this, however iOS does check to see if this domain is part of a major network and adds a point for that.  Could be iOS is missing this and Android is missing the "in major network" test.  Will check iOS against extensions again. :)

brindy: Yep, extensions doesn't appear to do this, but it does do the "in a" major network check:

https://github.com/duckduckgo/chrome-zeroclickinfo/blob/beta/js/site.js#L111

subsymbolic: This is equivalent to `inMajorTrackingNetwork`, I just found this clearer as the way extensions have it worded it sounds like a reference to the site when it is really a reference to trackers being in a major network.

brindy: Yeah, iOS is actually wrong here. We've been checking the current domain is part of a major tracker. Will create a separate PR to fix that in iOS.

PR #78, opened 2017-12-20
Extract query from DDG url (rather than storing the last searched query)

CDRussell:
Also, made some parameters non-null to simplify the logic in the DuckDuckGoUrlDetector class. There's only a single place that could have been supplying nulls, so handling that at the original call-site instead.

Asana Issue URL: https://app.asana.com/0/488551667048375/494978153394812

## Description
Fixing a bug where the omnibar would show the previous query erroneously under certain conditions.

## Steps to Test this PR:
1. Enter search query "test search" into omnibar (ensure "test search" is shown in the omnibar bar with the DDG search results page)
1. Enter "duckduckgo.com" (ensure full url containing "https://" is shown)
1. Enter term into the DDG web field; ensure the omnibar shows search query and not full url

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserViewModelTest.kt
@@ -101,13 +101,13 @@ class BrowserViewModelTest {
 
     @Test
     fun whenViewModelNotifiedThatUrlGotFocusThenViewStateIsUpdated() {
-        testee.onUrlInputStateChanged("", true)
+        testee.onOmnibarInputStateChanged("", true)
         assertTrue(testee.viewState.value!!.isEditing)
     }
 
     @Test
     fun whenViewModelNotifiedThatUrlLostFocusThenViewStateIsUpdated() {
-        testee.onUrlInputStateChanged("", false)
+        testee.onOmnibarInputStateChanged("", false)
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/com/duckduckgo/app/browser/BrowserViewModelTest.kt
```diff
@@ -122,6 +122,24 @@ class BrowserViewModelTest {
         assertEquals("duckduckgo.com", testee.viewState.value!!.url)
     }
 
+    @Test
+    fun whenUrlChangedWithDuckDuckGoUrlContainingQueryThenUrlRewrittenToContainQuery() {
+        testee.urlChanged("http://duckduckgo.com?q=test")
+        assertEquals("test", testee.viewState.value!!.url)
```
brindy: Feels like "url" is the wrong name for this now.  Perhaps it should be "addressBarText" instead?  Likewise urlInput in the BrowserActivity is also for more than just urls (ie search queries) so maybe should be called "addressBarTextInput"? 

@subsymbolic what do you think?

CDRussell: Good point

subsymbolic: Sounds good to me too. We call it the omnibar so how about `omnibarText` and `omnibarTextInput`?

PR #87, opened 2017-12-27
Feature/periodic configuration downloading

CDRussell:
Asana Issue URL: https://app.asana.com/0/471012189846020/481414989525314

## Description
Separation of tracker downloading and data loading, along with the scheduling of fetching updates for tracker data as well as HTTPS upgrade lists.

Makes of use [JobScheduler](https://developer.android.com/reference/android/app/job/JobScheduler.html) to handle the recurring background fetching of data. Currently set to approximately once every 24 hours, though this is deliberately inexact - Android will run the job when it thinks it best.

## Steps to Test this PR:
1. Uninstall existing app (breaking change to DB schema)
1. Ensure that data is downloaded when app first launches 
    - This can be confirmed by viewing using a Proxy, or by watching for the logcat message "Running immediate attempt to download app configuration"
1. Ensure that a job is scheduled with JobScheduler
    - This can be confirmed by using `adb shell dumpsys jobscheduler`
        - This will print output of existing jobs. Look for `com.duckduckgo.app.job.AppConfigurationJobService` in this list.
    - You can also force a job to run 
        - `adb shell cmd jobscheduler run -f com.duckduckgo.app 1`
        - requires Android 7.1 or greater to run this
1. ⚠️  Really important step - check tracker and HTTPs upgrade functionality still works!


app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDataLoader.kt
```diff
@@ -16,62 +16,56 @@
...
+        Timber.i("Looking for adblock tracker ${name.name} to load")
 
         if (trackerDataStore.hasData(name)) {
+            Timber.i("Found adblock tracker ${name.name}")
+
             val client = AdBlockClient(name)
             client.loadProcessedData(trackerDataStore.loadData(name))
+
+            // remove old client, then add the client again
+            trackerDetector.removeClient(client)
```
subsymbolic: Because this touches a JNI C++ component which allocates and deallocates memory the add and remove calls need to happen on the same thread across multiple syncs.

brindy: Maybe you could annotate the methods to say which thread they are expected to run on, or handle that in the wrapper explicitly?

But am curious why the memory (de)allocation is not considered thread safe?  (I don't understand why it would matter which thread the memory is allocated and deallocated on as it's all in the same process space)

subsymbolic: I've looked into this and found some really old references to it being important but I can't find any definitive evidence. Happy to lean on your JNI experience and drop it for now and investigate later if we run into issues.

subsymbolic:
@brindy I'm finishing this PR off. The one essential item was to fix db upgrades so our beta users wouldn't get a crash / need to reinstall. That's now done. Can you PR check my change? You'll need to do a clean when you switch to this branch as dagger will otherwise fail.

app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDataLoader.kt
brindy: How does this remove the old client?  The call here is removing the new client which the tracker won't know anything about it.  I can't see anything in the code which would indicate that the new client equates to the old client (e.g. hashCode/equals)

Would this be better off being a `removeClient(withName:)` type method?

subsymbolic:
Nice catch @brindy, I've added a check to ensure the client can't be added twice

app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDetector.kt
```diff
@@ -32,12 +31,17 @@ class TrackerDetector @Inject constructor(private val networkTrackers: TrackerNe
...
     private val clients = CopyOnWriteArrayList<Client>()
 
+    /**
+     * Adds a new client. If the client's name matches an existing client, old client is replaced
+     */
     fun addClient(client: Client) {
+        val existingClient = clients.firstOrNull { client.name == it.name }
+        if (existingClient != null) {
+            removeClient(existingClient)
+        }
```
brindy: You can replace these 3 lines with:

`clients.removeAll { client.name == it.name }`

brindy: ... or do that in the `removeClient` func and just call removeClient every time

brindy: and I just noticed that `TrackerDataLoader.kt` still calls removeClient for the disconnect list. I would be tempted to remove the removeClient func all together, but we talked about doing a bit of a refactor later so happy to leave it.

subsymbolic: Good idea. I've killed the removeClient method, we really don't need it.

brindy:
This is good to go. :)

PR #93, opened 2018-01-06
Allow app to be default browser and receive shared urls from other browsers

subsymbolic:
Asana Issue URL: 
https://app.asana.com/0/414730916066338/492769985844014
https://app.asana.com/0/414730916066338/364465928695406

## Description
Allows our app to be set as the default system browser. Also allows our app to open shared urls and text.

## Steps to Test this PR:
TEST ONE
1. Clear any default browsers you have set
1. Open an app with a url in it e.g SMS, evernote etc
1. Tap the link and select to open it with DuckDuckGo, select always
1. Ensure our app launches and opens the url
1. Ensure Duckduckgo is now the default browser (this can be checked under applications in settings)

TEST TWO
1. Open Chrome
1. Navigate to a page
1. Open the menu and select share
1. Select DuckDuckGo
1. Ensure our app launches and opens the url

TEST THREE
1. Open an application which contains text e.g email
1. Highlight some text and select share
1. Select DuckDuckGo
1. Ensure our app launches and the selected text is queries

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
@@ -16,21 +16,39 @@
 
 package com.duckduckgo.app.home
 
+import android.content.Intent
 import android.os.Bundle
 import android.support.v4.app.ActivityOptionsCompat
 import android.support.v7.app.AppCompatActivity
 import com.duckduckgo.app.browser.BrowserActivity
 import com.duckduckgo.app.browser.R
+import com.duckduckgo.app.global.intentText
 import kotlinx.android.synthetic.main.content_home.*
 
 class HomeActivity : AppCompatActivity() {
-
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_home)
 
         searchInputBox.setOnClickListener { showSearchActivity() }
+
+        if (savedInstanceState == null) {
+            consumeSharedText(intent)
+        }
+    }
+
+    override fun onNewIntent(intent: Intent?) {
+        super.onNewIntent(intent)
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
```diff
@@ -16,21 +16,39 @@
...
+        }
+    }
+
+    override fun onNewIntent(intent: Intent?) {
+        super.onNewIntent(intent)
+        consumeSharedText(intent)
+    }
+
+    private fun consumeSharedText(intent: Intent?) {
+        val sharedText = intent?.intentText
```
brindy: Optional: I think we should start to embrace the Elvis operator, ahuh-huh. ;-) 

This line could be written as:

`val sharedText = intent?.intentText ?: return`

and then the if statement below can be removed.

app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
```diff
@@ -16,21 +16,39 @@
...
+        if (sharedText != null) {
+            val intent = BrowserActivity.intent(this, sharedText)
+            startActivity(intent)
```
brindy: Optional: I wouldn't have noticed if I hadn't looked at this in Android Studio tbh, but `intent` shadows the function's argument of the same name.

app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
subsymbolic: Love an excuse to use the elvis operator 👍

subsymbolic:
That PR feedback push failed. I've just added it to https://github.com/duckduckgo/Android/pull/95 and merged it there.

PR #98, opened 2018-01-08
Add fire button and cache clearing

subsymbolic:
Asana Issue URL: 
https://app.asana.com/0/414730916066338/471193197990045

## Description
Shows fire button and allows cache and cookies to be cleared

## Steps to Test this PR:
TEST ONE
1. Login to a webpage e.g facebook
2. Press the fire button and select `Clear all data`
3. Open the webpage again ensure you are logged out

TEST TWO
1. Login to a webpage e.g facebook
2. Press the fire button and select `Cancel`
3. Ensure you are still logged in


app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private fun clearCache() {
+        val webView = WebView(context)
+        webView.clearCache(true)
+        webView.clearHistory()
+    }
+
+    private fun clearCookies() {
+        CookieManager.getInstance().removeAllCookies(null)
```
CDRussell: This method is async, so there is the slight risk that it would take a while to complete, and the user might visit a site before the cookie is completely removed. 

This seems like a small risk though, as I suspect however async it is, it will always beat the user. 

We could look to revisit this later and add a callback instead of null, and only inform the user all is done when the callback fires.

Thoughts?

app/src/main/res/layout/sheet_fire_clear_data.xml
```diff
@@ -0,0 +1,62 @@
...
+        android:paddingBottom="17dp"
+        android:paddingEnd="24dp"
+        android:paddingStart="24dp"
+        android:paddingTop="24dp"
+        android:text="@string/fireClearAll"
+        android:textColor="@color/brickOrange"
+        android:textSize="16sp"
+        android:textStyle="normal" />
+
+    <android.support.v7.widget.AppCompatTextView
```
CDRussell: Any reason we're explicitly using `AppCompatTextView` here instead of `TextView`?

app/src/main/res/layout/sheet_fire_clear_data.xml
```diff
@@ -0,0 +1,62 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:background="@color/white"
+    app:layout_behavior="android.support.design.widget.BottomSheetBehavior">
+
+    <TextView
+        android:id="@+id/clearAllOption"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="?attr/selectableItemBackground"
```
CDRussell: Does this work for you? I'm not seeing the ripple effect here for some reason. 

I still see the ripple effect elsewhere in the app using the same device. 😕

app/src/main/res/layout/sheet_fire_clear_data.xml
subsymbolic: There _was_ a reason, can't remember why now so I'll switch back!

app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
subsymbolic: Makes sense, I'm making the change

app/src/main/res/layout/sheet_fire_clear_data.xml
subsymbolic: Hmm, this seems to be caused by setting the max "peek height" in styles. Without it only part of the dialog shows in landscape. I'll investigate!

subsymbolic: Thanks, as per your advice I've added a separate Asana ticket for it https://github.com/duckduckgo/Android/pull/98

PR #110, opened 2018-01-14
bookmarks

brindy:
Asana Issue URL: https://app.asana.com/0/488551667048375/513412888988779

## Description

Allow user to add the current page as a bookmark, view their bookmarks, delete bookmarks and open bookmarks.

There's a bug around up navigation, but this has been fixed by @subsymbolic in an existing PR.

## Steps to Test this PR:

1. Browse to a page
1. Open the menu and select "Add Bookmark"
1. Open the menu and select "Bookmarks"
1. The newly added bookmark should be there
1. Go back to the home page
1. Open bookmarks and select the bookmark
1. The browser activity should launch with the url for the selected bookmark
1. Open the bookmarks
1. Tap on the delete icon
1. Tap cancel on the confirmation dialog - the bookmark remains
1. Tap on the delete icon
1. Tap ok on the confirmation - the bookmark is deleted


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -302,13 +303,21 @@ class BrowserActivity : DuckDuckGoActivity() {
         return false
     }
 
+    private fun addBookmark() {
```
CDRussell: I think this operation needs some visual feedback to confirm it worked. I pressed it twice thinking it hadn't worked the first time.

from view model back to activity with a "show bookmark added" command, perhaps, which manifests itself as a `Toast` or `Snackbar`?

app/src/main/res/values/strings.xml
```diff
@@ -108,4 +109,11 @@
         After all, the internet shouldn\'t feel so creepy, and getting the privacy you deserve online should be as simple closing the blinds.</string>
     <string name="aboutMoreLink">More at duckduckgo.com/about</string>
 
+    <!-- Bookmarks Activity -->
+    <string name="boomarksMenuTitle">Bookmarks</string>
+    <string name="bookmarksActivityTitle">Bookmarks</string>
+    <string name="deleteBookmarkContentDescription">Delete bookmark for %s</string>
+    <string name="bookmarkDeleteConfirmMessage">Are you sure you want to delete this bookmark?\n\n\'%s\'</string>
```
CDRussell: Not sure if in design, but it would be nice to be able to see the URL somewhere. maybe in the deletion dialog. not urgent, though

app/build.gradle
```diff
@@ -95,6 +96,9 @@ dependencies {
     implementation "com.google.dagger:dagger-android-support:$dagger"
     releaseImplementation 'com.faendir:acra:4.10.0'
 
+    // Anko
+    compile "org.jetbrains.anko:anko-commons:$ankoVersion"
```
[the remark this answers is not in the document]
CDRussell: Discussed on slack; using for its `async` feature and some UI convenience methods currently (but we might end up using it more now that it's in)

app/src/androidTest/java/com/duckduckgo/app/bookmarks/ui/BookmarksViewModelTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    var instantTaskExecutorRule = InstantTaskExecutorRule()
+
+    private val liveData = MutableLiveData<List<BookmarkEntity>>()
+    private val viewStateObserver: Observer<BookmarksViewModel.ViewState> = mock()
+    private val commandObserver: Observer<BookmarksViewModel.Command> = mock()
+    private val bookmarksDao: BookmarksDao = mock()
+
+    private val bookmark = BookmarkEntity(title = "title", url = "www.example.com")
+
+    private val testee: BookmarksViewModel by lazy {
```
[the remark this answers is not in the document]
brindy: Had this discussion with @subsymbolic as that was my thinking too, but I now believe you get a new instance of the entire test class for each test that is run so should be fine.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
brindy: Good call

app/src/main/res/values/strings.xml
brindy: Yeah, I even had the URL in the list as well at one point but it's not in the design or on iOS either so I figured if title was good enough for the list, it's probably ok for this.  Plus this is temporary until we implement the more advance book marks UI where it will just be deleted instantly and then show a snackbar with an undo button.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -308,6 +309,9 @@ class BrowserActivity : DuckDuckGoActivity() {
         val url = webView.url
         doAsync {
             viewModel.addBookmark(title, url)
+            uiThread {
```
CDRussell: If you're already on the UI thread, can't you just toast outside of the `async`?

PR #116, opened 2018-01-18
Tracking improvements

subsymbolic:
Asana Issue URL: https://app.asana.com/0/488551667048375/509644293434231

## Description
• Reduces incidence of resource load mismatch from other sites (still not perfect will revist)
• Optimises page speed by removing content blocking on trusted sites
• Improves parent tracker network matching
• Remove code coverage as it breaks unit test inspection

## Steps to Test this PR:
TEST ONE
1. Visit duckduckgo.com 
2. Note faster load speed now that we don’t run full site monitoring on trusted sites

TEST TWO
1. Visit cnn.com
2. In the middle of a load visit duckduckgo.com
3. Ensure site rating is still A

TEST THREE
1. Visit youtube.com
2. Note that google (which is parent network) is no longer blocked

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -18,6 +18,7 @@ package com.duckduckgo.app.browser
 
 import android.graphics.Bitmap
 import android.net.Uri
+import android.support.annotation.AnyThread
 import android.support.annotation.WorkerThread
 import android.webkit.WebResourceRequest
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -137,4 +144,16 @@ class BrowserWebViewClient @Inject constructor(
         return true
     }
 
+    @AnyThread
+    private fun WebView.urlFromWorkerThread(): String? {
```
CDRussell: Could do with comments

PR #120, opened 2018-01-18
favourites and saved searches migration

brindy:
Asana Issue URL: https://app.asana.com/0/488551667048375/471444661339649

## Description

Migrates legacy favourites to bookmarks.

Legacy code has been used and I tried to change as little as possible (but have deleted a fair chunk).

Haven't tested migrating from older databases but I would like to assume that code has been well tested already and it's the same code used here.

## Steps to Test this PR:

1.  Build and run Search and Stories (sorry!  Remember to update to get the required submodule and you may need to install a bunch of older android build tools versions)
1. Favourite a couple of stories
1. Perform a search and favourite it
1. Browse to a page and favourite it
1. Install new app
1. Go to bookmarks and confirmed migrated favourites and searches are present and all bookmarks open correctly.


app/src/main/java/com/duckduckgo/app/migration/legacy/LegacyDbContracts.java
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright (c) 2018 DuckDuckGo
```
subsymbolic: It's been a while since I've had to look at Java :)

PR #123, opened 2018-01-19
Browsing menu

subsymbolic:
Asana Issue URL: https://app.asana.com/0/488551667048375/477327275839581

## Description
Add menu styling to the app

## Steps to Test this PR:
Check that the menu renders well across devices and has the correct buttons enabled/disabled on the home and browser screens

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserPopupMenu.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserPopupMenu.kt
@@ -44,12 +44,7 @@ class BrowserPopupMenu : PopupWindow {
         anchorView.getLocationOnScreen(anchorLocation)
         val x = margin
         val y = anchorLocation[1] + margin
-
-        if (SDK_INT > 21) {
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserPopupMenu.kt
```diff
@@ -0,0 +1,60 @@
...
+import android.view.View
+import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
+import android.widget.PopupWindow
+
+class BrowserPopupMenu : PopupWindow {
+
+    constructor(layoutInflater: LayoutInflater, view: View = BrowserPopupMenu.inflate(layoutInflater))
+            : super(view, WRAP_CONTENT, WRAP_CONTENT, true) {
+
+        if (SDK_INT > 21) {
```
CDRussell: Is this deliberately doing something different for API 21? I would have thought 21 and up could have elevation supplied (ie, `>= 21`).

If it is deliberately doing something different for Lollipop, then cool. But if it was supposed to be `>= 21` then we can ditch the whole if, given we're `minSDK=21`.

PR #128, opened 2018-01-19
Adds home activity to DI so cookie dependency is injected

subsymbolic:
## Description
Fixes crash when pressing fire button

## Steps to Test this PR:
1. Launch the app
1. Tap on the fire button. It will no longer crash.


subsymbolic:
@CDRussell I'm pushing this through for the release but would love your retrospective check over it.

PR #135, opened 2018-01-22
Add dom storage to fix ebay search box

subsymbolic:
Asana Issue URL: https://app.asana.com/0/488551667048375/530368950629129

**Description**
Fixes ebay search bug by adding dom storage

**Steps to Test this PR:**
1. Visit ebay.com
1. Start typing a search and note that autocomplete shows
1. Submit search and note that results are shown - previously nothing happened


subsymbolic:
Good question. I have added additional "clearing" behavior so the fire button zaps this too. I'm concerned that disabling it entirely may break more sites. I'll ponder.

brindy:
Usually you have to enable the database to get persistent local storage and we don't do that, see:

https://developer.android.com/reference/android/webkit/WebSettings.html#getDatabaseEnabled()

But clearing the data in this way would cover that as well, I believe, so it's good.

PR #144, opened 2018-01-23
manually manage the web view in order to make sure it does not leak

brindy:
Asana Issue URL: https://app.asana.com/0/413877547933099/532522547697239

## Description

There appears to be a leak in WebView, so managing it manually allows us to use the application context which then allows the view model to be cleaned up properly and in turn fixes the bug.

Also added leak canary to the project to aid in future leak detection.

## Steps to Test this PR:
1.  Go to cnn.com
1.  Before it finishes loading, "burn" it and return to the home screen
1. Go to cnn.com again - should load properly

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
@@ -24,6 +24,7 @@ import com.duckduckgo.app.di.DaggerAppComponent
 import com.duckduckgo.app.job.AppConfigurationSyncer
 import com.duckduckgo.app.migration.LegacyMigration
 import com.duckduckgo.app.trackerdetection.TrackerDataLoader
+import com.squareup.leakcanary.LeakCanary
 import dagger.android.AndroidInjector
 import dagger.android.DispatchingAndroidInjector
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
```diff
@@ -66,6 +69,14 @@ class DuckDuckGoApplication : HasActivityInjector, HasServiceInjector, Applicati
         migrateLegacyDb()
     }
 
+    private fun installLeakCanary(): Boolean {
```
subsymbolic: Does this install in release as well?

app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
```diff
@@ -56,6 +57,8 @@ class DuckDuckGoApplication : HasActivityInjector, HasServiceInjector, Applicati
     override fun onCreate() {
         super.onCreate()
 
+        if (!installLeakCanary()) return
```
subsymbolic: Did you mean to return here? Won't this cause a crash?

PR #153, opened 2018-01-24
Add special handling for 3rd party keyboards which submit \n for done

CDRussell:
Asana Issue URL: 
https://app.asana.com/0/488551667048375/533375498923960
https://app.asana.com/0/488551667048375/533891143524672

## Description
Done button doesn't submit query on some 3rd party software keyboards
This also tackles the crash reports for when `keyEvent` is `null`.

## Steps to Test this PR:
1. Install TouchPal (https://play.google.com/store/apps/details?id=com.cootek.smartinputv5&hl=en_GB)
1. Set TouchPal as the keyboard
1. Use the app, and ensure the query submits when you hit the software "Done" button

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -28,11 +28,11 @@ import android.os.Build
 import android.os.Bundle
 import android.support.v7.widget.LinearLayoutManager
 import android.text.Editable
-import android.view.KeyEvent
+import android.view.KeyEvent.KEYCODE_ENTER
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
-import android.view.inputmethod.EditorInfo
+import android.view.inputmethod.EditorInfo.IME_ACTION_DONE
 import android.webkit.CookieManager
 import android.webkit.WebSettings.MIXED_CONTENT_COMPATIBILITY_MODE
 import android.webkit.WebView
@@ -277,7 +277,7 @@ class BrowserActivity : DuckDuckGoActivity(), BookmarkDialogCreationListener {
         }
 
         omnibarTextInput.setOnEditorActionListener(TextView.OnEditorActionListener { _, actionId, keyEvent ->
-            if (actionId == EditorInfo.IME_ACTION_DONE || keyEvent?.keyCode == KeyEvent.KEYCODE_ENTER) {
+            if (actionId == IME_ACTION_DONE || keyEvent?.keyCode == KEYCODE_ENTER) {
                 userEnteredQuery(omnibarTextInput.text.toString())
                 return@OnEditorActionListener true
             }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -260,6 +260,20 @@ class BrowserActivity : DuckDuckGoActivity(), BookmarkDialogCreationListener {
...
         omnibarTextInput.addTextChangedListener(object : TextChangedWatcher() {
+
+            override fun onTextChanged(
+                charSequence: CharSequence,
+                start: Int,
+                before: Int,
+                count: Int
+            ) {
+                // some 3rd party keyboards submit \n instead of an IME action
+                if (before == 0 && count == 1 && charSequence[start] == '\n') {
```
subsymbolic: Do we still need this is we get rid of the "shortMessage" ime option in the xml

CDRussell: I guess we're still at the mercy of the 3rd party keyboards... even if changing this changes this one individual keyboard's behaviour, now that we've seen it once i'm tempted to leave it in as a safety.

subsymbolic: Keyboard behavior is complex on android and I'm not keen to introduce patches unless we know they fix a legitimate issue lest we introduce something knew. For example I often share text from other applications to our app to search for stuff. If there's a line break in it this patch (that we're not sure we need) may break that.

As we were causing this by using an incorrect IME option, I'd rather remove that alone.

subsymbolic: If you can find evidence that we need this, I'll happily approve it.

CDRussell: >As we were causing this by using an incorrect IME option
I disagree we are to blame, or that this is an incorrect IME option. It seems like an entirely appropriate IME option to use here (https://developer.android.com/reference/android/text/InputType.html#TYPE_TEXT_VARIATION_SHORT_MESSAGE)

That said, I'm happy to remove `shortMessage`, and the patch and wait and see if there are any other reports.

>Keyboard behavior is complex
Ain't it just!

subsymbolic: The reason I feel it's wrong is because it introduces a bunch of behavior that just doesn't work with urls. For example, I think it's the cause of other keyboard issues we've seen around extra spacing being added when typing. This isn't our fault - google should document the behavior of these options better and be clear about which combinations work well together and which ones fight each other.

subsymbolic: > Ain't it just!
Yesssss! Damn you android keyboards. I get the feeling we'll get to know them all well. The change we've deferred here may yet be revived 😆

CDRussell: Like I would! 😏

PR #154, opened 2018-01-24
Initialise the privacy monitor live data directly in singleton instance

CDRussell:
Asana Issue URL: https://app.asana.com/0/488551667048375/533891143524674

## Description
There's a crash report that privacy monitor was not initialised. the existing code initialised it only in one activity, and if the process died while the user was on another activity, this would not be initialised again.

## Steps to Test this PR:
1. Hard to test as it is hard to reproduce. Make sure the new implementation doesn't break anything around privacy grade.


subsymbolic:
Good change, we'll need to remember this issue again during tabs implementation!

PR #155, opened 2018-01-25
Use toolbar as the popup menu's anchor view instead of toolbar menu item

CDRussell:
Asana Issue URL: https://app.asana.com/0/488551667048375/532221766945306

## Description
Some users are reporting that the popup menu crashes for them. It appears the view used as an _anchor_ is `null` for them.

Instead of trying to use the menu item itself, this PR switches to using the `toolbar` itself as the anchor, which should be visually identical but guaranteed to be non-null instead.

## Steps to Test this PR:
1. Hard to test unless you have a device which is crashing. But make sure it doesn't break anything or change the existing look and feel of the menu on both:
    1. HomeActivity
    1. BrowserActivity

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
@@ -66,10 +66,12 @@ class HomeActivity : DuckDuckGoActivity() {
 
     private fun configurePopupMenu() {
         popupMenu = BrowserPopupMenu(layoutInflater)
-        popupMenu.contentView.backPopupMenuItem.isEnabled = false
-        popupMenu.contentView.forwardPopupMenuItem.isEnabled = false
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/home/HomeActivity.kt
```diff
@@ -66,10 +66,12 @@ class HomeActivity : DuckDuckGoActivity() {
 
     private fun configurePopupMenu() {
         popupMenu = BrowserPopupMenu(layoutInflater)
-        popupMenu.contentView.backPopupMenuItem.isEnabled = false
-        popupMenu.contentView.forwardPopupMenuItem.isEnabled = false
-        popupMenu.contentView.refreshPopupMenuItem.isEnabled = false
-        popupMenu.contentView.addBookmarksPopupMenuItem.isEnabled = false
+        popupMenu.contentView.apply {
```
subsymbolic: Nice update

PR #179, opened 2018-02-06
Add long press context menu for images with option to download image

CDRussell:
Asana Issue URL: https://app.asana.com/0/414730916066338/538612605030335

## Description
Add ability to long press on an image in a WebView and see a context menu allowing for the image to be downloaded

Note, images won't be able to be downloaded from DDG SERP yet due to an ongoing issue there.

## Steps to Test this PR:
1. Visit a site which contains images, such as http://www.bbc.co.uk/news/business-42957834
1. Check that you can long press on an image (such as the FTSE 100 chart 📉) and see an option to download the image. Do so.
1. Verify you are prompted for the runtime permission for writing to storage.
1. Deny the permission, and ensure the app handles it gracefully enough.
1. Try again, this time grant the permission, and ensure the image downloads successfully.
1. Check that you cannot long press on the video on the page.
1. Check that you can long press on the `BBC` icon (very top left of the page) and that you can download this image - this is a different type of image from the FTSE chart but both should download successfully.
1. Also, check an image from theguardian.com as these images were previously not being named sensibly - it should name them sensibly now.

From here, it would be wise to revoke the storage permission, and test that PDFs still download successfully (using espc.com for instance, and clicking on a property, and then "schedule" to trigger a PDF download)


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -380,6 +386,23 @@ class BrowserActivity : DuckDuckGoActivity(), BookmarkDialogCreationListener {
...
+    override fun onCreateContextMenu(menu: ContextMenu, view: View, menuInfo: ContextMenu.ContextMenuInfo?) {
+        webView.hitTestResult?.let {
+            viewModel.userLongPressedInWebView(it, menu)
+        }
+    }
+
+    override fun onContextItemSelected(item: MenuItem): Boolean {
+        webView.hitTestResult?.let {
+            var url = it.extra
+            if(viewModel.userSelectedItemFromLongPressMenu(url, item)) {
```
subsymbolic: Clean up formatting

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
```diff
@@ -322,14 +328,33 @@ class BrowserViewModel(
...
     private fun openUrl(url: String) {
         command.value = Navigate(url)
     }
 
     fun onUserSelectedToEditQuery(query: String) {
         viewState.value = currentViewState().copy(isEditing = false, showAutoCompleteSuggestions = false, omnibarText = query)
     }
+
+    fun userLongPressedInWebView(target: WebView.HitTestResult, menu: ContextMenu) {
+        Timber.i("Long pressed on %s, (extra=%s)", target.type, target.extra)
```
subsymbolic: Can we use kotlin style strings? It is the project standard and I'd rather not mix styles. I know it's slightly less efficient but on the upside they are more readable and if efficiency becomes an issue we can fix it with some proguard config.

CDRussell: I've removed other instances of the _Timber style_ in favour of the _Kotlin style_ too.

CDRussell:
>Can we update the pdf download to also be a long press menu operation for consistency?

I'm not sure if we can? it looks like the PDF download is a more integral action whereby the `WebView` just tells us it can't handle some content; i don't see a way to hook into that.

We could perhaps start whitelisting types we suspect the browser can't handle (and PDF would be one) and handle that in a special way, but seems like a bigger thing than should be done in this task.

subsymbolic:
> I'm not sure if we can? it looks like the PDF download is a more integral action
It was just a thought, I should have marked it as optional as it really goes beyond the scope of this task.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
CDRussell: What site/image are you seeing that's giving you `downloadfile.bin`? I'm seeing sensible filenames by default when I'm testing.

screen shot 2018-02-07 at 09 41 57

subsymbolic: I had tried a few images on espc.com and the guardian.co.uk and they are all `downloadfile.bin` in our app but fine in Chrome. I just tried bbc and that looks good in our app too

CDRussell: Thanks; will investigate

CDRussell:
@subsymbolic I've added code to save the images to "public external storage directory" so that it will outlive the system's download cache. Doing so requires the runtime permission for `WRITE_EXTERNAL_STORAGE` so that's also included.

Re-review please

PR #180, opened 2018-02-07
Make activities enable popup options they want

subsymbolic:
Asana Issue URL: 
https://app.asana.com/0/414730916066338/550678398757789

## Description
Low hanging fruit refactor. Make activities enable popup menu options they want rather than disabling options they don't want

## Steps to Test this PR:
TEST ONE:
1. Run the app
1. Go to the home page
1. Ensure the disabled options do nothing & the enabled ones continue to perform the correct options

TEST TWO:
1. Run the app
1. Go to the home page
1. Ensure the disabled options do nothing & the enabled ones continue to perform the correct options

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserPopupMenu.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserPopupMenu.kt
@@ -69,7 +69,7 @@ class BrowserPopupMenu : PopupWindow {
 
     companion object {
 
-        const val margin = 30
+        private const val margin = 30
 
         fun inflate(layoutInflater: LayoutInflater): View {
             return layoutInflater.inflate(R.layout.popup_window_browser_menu, null)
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserPopupMenu.kt
```diff
@@ -51,7 +69,7 @@ class BrowserPopupMenu : PopupWindow {
 
     companion object {
 
-        val margin = 30
+        const val margin = 30
```
CDRussell: Be as well adding `private` here too

CDRussell:
When performing a search from the `HomeActivity`, I see the initialised privacy grade button appear momentarily again though. Would you mind checking if that regression was reintroduced in this work?

subsymbolic:
> Would you mind checking if that regression was reintroduced in this work?

I see that in develop as well :(

PR #184, opened 2018-02-08
Feature/ci optimisation

CDRussell:
Asana Issue URL: https://app.asana.com/0/414730916066338/538244745809854

## Description
~This isn't meant to solve the CI stability problems, but will hopefully shed some light on what they are when they do surface as a failure.~

This has grown a bit into something which I do now hope improves the CI stability problems.

• Enabling the test orchestrator for CI builds (you can opt into it as a developer by supplying the `-Porchestrator` flag but probably not usually needed) and bumping gradle version.
• More logging on the tests. Will print out which tests are running and will print out the XML test report failures when the tests fail.
• Allowing for main thread queries for the test DBs
• Ensuring the oft-failing test has the correct test unit rule applied which provide test versions of executors used by Room and Android Architecture Components.
• Better Dagger configuration to allow separation of production and test implementations of modules. This allows for the substitution of the real database with an in-memory one for instance, or for the job scheduler to be replaced with a fake job scheduler than doesn't actually schedule jobs (both of which are done already). We should also go further in ensuring there are no network requests made but this task has grown enough arms and legs for now. At least this PR will make that 
task easier to achieve

## Steps to Test this PR:
1. n/a


app/build.gradle
```diff
@@ -7,6 +7,7 @@ apply from: '../versioning.gradle'
 
 ext {
     VERSION_NAME = "4.2.0"
+    USE_ORCHESTRATOR = project.hasProperty('orchestrator') ? project.property('orchestrator') : false
```
subsymbolic: Nice! Does this change the speed of execution? Can't tell from travis, it's still running.

CDRussell: I doubt it will make it any faster. 😄 

But will report back after a few runs with some indication on impact.

CDRussell: Speed seems roughly the same. Comparing with the last few successful runs from `develop`, they are all between 19.5 mins and 23.5 mins. Similar range for the last few runs from this branch.

CDRussell:
@subsymbolic Can I get a re-review? There have been significant changes made since your last review, as the build was still failing.

I'm hopeful this will make the build stable again. But even if it doesn't, there is some decent changes in here that we will want in our project regardless.

app/src/androidTest/java/com/duckduckgo/app/di/FakeJobSchedulerModule.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import android.app.job.JobInfo
+import android.app.job.JobScheduler
+import android.app.job.JobWorkItem
+import dagger.Module
+import dagger.Provides
+import javax.inject.Singleton
+
+@Module
+class FakeJobSchedulerModule {
```
subsymbolic: How would you feel about calling this `StubJobSchedulerModule`? I feel like there isn't a clear distinction between Stubs and Fakes and there is little benefit in separating them so  it makes sense to use `Stub` for both. It seems I'm not alone https://stackoverflow.com/questions/346372/whats-the-difference-between-faking-mocking-and-stubbing#346440.

app/src/androidTest/java/com/duckduckgo/app/di/TestDatabaseModule.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.arch.persistence.room.Room
+import android.content.Context
+import com.duckduckgo.app.global.db.AppDatabase
+import dagger.Module
+import dagger.Provides
+import javax.inject.Singleton
+
+
+@Module(includes = [DaoModule::class])
+class TestDatabaseModule {
```
subsymbolic: I'd happily call this `StubDatabaseMobule` too.

app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
```diff
@@ -73,11 +73,11 @@ class DuckDuckGoApplication : HasActivityInjector, HasServiceInjector, Applicati
         migrateLegacyDb()
     }
 
-    private fun installLeakCanary(): Boolean {
+    protected open fun installLeakCanary(): Boolean {
         if (LeakCanary.isInAnalyzerProcess(this)) {
-            return false;
+            return false
```
subsymbolic: Yes to deleting those semi-colons!

PR #197, opened 2018-02-22
Fix landscape layout's squished controls; add scroll view to menu items

CDRussell:
Asana Issue URL: https://app.asana.com/0/488551667048375/567110983215934

## Description
Add scroll view to popup menu, and fix the horizontal squishing that happens to the back/forward/refresh controls on landscape.

## Steps to Test this PR:
1. Launch the popup menu in both portrait and landscape; verify the controls look good 
1. To test the scroll view, you will either need a small screen (in landscape most likely) or artificially create the problem by adding new menu items into `popup_window_browser_menu.xml`

### DIFF
--- a/app/src/main/res/layout/popup_window_browser_menu.xml
+++ b/app/src/main/res/layout/popup_window_browser_menu.xml
@@ -24,11 +24,11 @@
         android:id="@+id/header"
         android:layout_width="0dp"
         android:layout_height="50dp"
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintEnd_toEndOf="parent"
         android:background="@color/white"
         android:gravity="center"
-        android:orientation="horizontal">
+        android:orientation="horizontal"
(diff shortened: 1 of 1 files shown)


subsymbolic:
@CDRussell on which device was this happening? I'm not seeing it on my Pixel 2 XL. I would like to reproduce the issue so I can confirm that it is fixed.

CDRussell:
I was seeing it on the Samsung Galaxy S6. I'll confirm I can still reproduce the problem on `develop`.

CDRussell:
@subsymbolic , the problem only manifests itself after a menu item containing a checkbox is added, which is what is happening as part of #198 

So i'd suggest getting #198 merged first, and the problem should be reproducible. This PR will need a minor merge conflict resolved, but just a small one; after that it should be testable that this will fix the now-visible problem.

CDRussell:
screenshot_20180223-162817

Screenshot of the problem

PR #198, opened 2018-02-22
Feature/request desktop site

CDRussell:
Asana Issue URL: https://app.asana.com/0/488551667048375/529888388192278

## Description
Gives the user an option to request that the desktop site be loaded instead of a mobile-specific site. It achieves this by altering the user agent string accordingly.

Note, if the url has an obviously mobile-specific site (using the `m.xxxxxxxx.com`) format, we strip the `m.` out of the URL when jumping to the desktop version. Discussions welcomed around this.

Also included in this is the slight improvement over the general UA strings we submit; they no longer contain the device make or model in them. We do still include the rest of the details, including the Android version and `AppleWebKit/xxx.yy` version information to ensure sites render well. https://app.asana.com/0/414730916066338/537987832233112

## Steps to Test this PR:
1. Visit a site that redirects to a mobile-specific site (e.g., youtube.com)
1. Verify that you end up on the mobile site (m.youtube.com)
1. Use the overflow menu, and check to receive the desktop site
1. Verify that you end up on the desktop site
1. Repeat again to ensure you end up back on the mobile site


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -401,10 +416,13 @@ class BrowserActivity : DuckDuckGoActivity(), BookmarkDialogCreationListener, We
     @SuppressLint("SetJavaScriptEnabled")
     private fun configureWebView() {
         webView = layoutInflater.inflate(R.layout.include_duckduckgo_browser_webview, webViewContainer, true).findViewById(R.id.browserWebView) as WebView
+        defaultUserAgentString = webView.settings.userAgentString
+
         webView.webViewClient = webViewClient
         webView.webChromeClient = webChromeClient
 
         webView.settings.apply {
+            userAgentString = userAgentProvider.getUserAgent(defaultUserAgentString)
```
subsymbolic: Optional: Would it make sense to instantiate the userAgentProvider with the defaultUserAgentString?

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -28,8 +28,8 @@ interface SettingsDataStore {
 class AppSettingsPreferencesStore @Inject constructor(private val context: Context) : SettingsDataStore {
 
     companion object {
-        val SHARED_PREFERENCES_FILENAME = "com.duckduckgo.app.settings_activity.settings"
-        val KEY_AUTOCOMPLETE_ENABLED = "AUTOCOMPLETE_ENABLED"
+        const val SHARED_PREFERENCES_FILENAME = "com.duckduckgo.app.settings_activity.settings"
```
subsymbolic: I did the same thing in my branch :)

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
```diff
@@ -388,6 +392,27 @@ class BrowserViewModel(
         viewState.value = currentViewState().copy(findInPage = findInPage)
     }
 
+
+    fun desktopSiteModeToggled(urlString: String?, desktopSiteRequested: Boolean) {
+        if (desktopSiteRequested) {
+            command.value = BrowserViewModel.Command.DesktopMode
+        } else {
+            command.value = BrowserViewModel.Command.MobileMode
```
subsymbolic: Are we sharing too much logic with the activity? Perhaps the command should be to update the user agent?

app/src/main/java/com/duckduckgo/app/browser/userAgent/UserAgentProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package com.duckduckgo.app.browser.userAgent
+
+import android.net.Uri
+import android.os.Build
+import javax.inject.Inject
+
+
+fun Uri.isMobileSite() : Boolean = this.authority.startsWith("m.")
```
subsymbolic: Do you need the explicit `this`?

app/src/main/java/com/duckduckgo/app/browser/userAgent/UserAgentProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+
+fun Uri.toDesktopUri(): Uri {
```
subsymbolic: These extensions feel generic rather than related to the user agent. Can we move them into UriExtensions, format and unit test them?

app/src/main/java/com/duckduckgo/app/browser/userAgent/UserAgentProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+    return if(this.isMobileSite()) {
```
subsymbolic: Do you need the explicit `this`?

app/src/main/java/com/duckduckgo/app/browser/userAgent/UserAgentProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+        Uri.parse(toString().replaceFirst("m.", ""))
```
subsymbolic: Can we protect future maintainers by adding a unit test for this to ensure that a non-mobile url with an ".m" e.g "adam.example.com" stays unchanged by this method?

app/src/main/java/com/duckduckgo/app/browser/userAgent/UserAgentProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Example Default User Agent (From Chrome):
+ * Mozilla/5.0 (Linux; Android 8.1.0; Nexus 6P Build/OPM3.171019.014) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.137 Mobile Safari/537.36
+ *
+ * Example Default Desktop User Agent (From Chrome):
+ * Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.137 Safari/537.36
+ */
+class UserAgentProvider @Inject constructor() {
+
+    companion object {
```
subsymbolic: Can we place companion objects at the bottom? I got the same feedback in my PR :) https://kotlinlang.org/docs/reference/coding-conventions.html#source-code-organization

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
```diff
@@ -89,6 +89,12 @@ class WebViewRequestInterceptor @Inject constructor(
         return null
     }
 
+    private fun shouldStripRequestedWithHeader(request: WebResourceRequest): Boolean {
+        if(!request.isForMainFrame) return false
```
subsymbolic: Spacing / formatting

app/src/main/res/layout/popup_window_browser_menu.xml
```diff
@@ -95,6 +95,17 @@
                 android:enabled="false"
                 android:text="@string/find_in_page" />
 
+            <CheckBox
```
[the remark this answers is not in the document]
CDRussell: ✅

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserViewModelTest.kt
```diff
@@ -473,4 +475,49 @@ class BrowserViewModelTest {
         assertEquals("", testee.viewState.value!!.findInPage.searchTerm)
     }
 
+    @Test
+    fun whenUserSelectsDesktopSiteThenDesktopModeCommandIssued() {
+        testee.desktopSiteModeToggled("http://example.com", desktopSiteRequested = true)
+        verify(mockCommandObserver, Mockito.atLeastOnce()).onChanged(commandCaptor.capture())
```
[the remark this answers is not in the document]
CDRussell: There are other commands that might already fire (like `ShowKeyboard`) so we don't want to verify only one command which is the default Mockito flow.

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
CDRussell: As per the video chat conversation earlier, this might fall under any work that comes out of considering the wider command/view state approach.

Are you ok with deferring any more changes on this until the above has happened, @subsymbolic ?

app/src/main/java/com/duckduckgo/app/browser/userAgent/UserAgentProvider.kt
CDRussell: Nope; removed

app/src/main/java/com/duckduckgo/app/browser/userAgent/UserAgentProvider.kt
CDRussell: Good thinking!

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
CDRussell: Removed function entirely; was unused

CDRussell:
RR @subsymbolic. all feedback should be addressed or, at least, given a response. let me know if i missed any.

CDRussell:
@subsymbolic I think you've unearthed something bigger. 

It looks like `BrowserWebViewClient.onPageFinished()` is being called for every resource, including javascript ones. Therefore, it's calling through to `BrowserViewModel.loadingFinished()` which hides the loading indicator.

I _think_ this is unrelated to the changes made for desktop mode but an important thing to fix. I will create a new ticket for this.

CDRussell:
Identified the problem with sites like theguardian.com is that because they have a responsive web page rather than a dedicated mobile site, the content of the web page is identical for both desktop and mobile UA strings.

Instead, we have to ensure we change some `WebView` settings locally based on which browsing mode we are in which will affect the scale on the `WebView`.

subsymbolic:
This is really close. Portrait is looking great, landscape looks a little too "zoomed out". I have added screenshots of our app and chrome for comparison.

ourapp_bbc
ourapp_theguardian
chrome_bbc
chrome_theguardian

subsymbolic:
Update: I see the same for portrait websites as well when in desktop mode. Facebook is a good one to test on as it happens in both portrait and landscape modes.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -401,14 +409,15 @@ class BrowserActivity : DuckDuckGoActivity(), BookmarkDialogCreationListener, We
...
+
         webView.webViewClient = webViewClient
         webView.webChromeClient = webChromeClient
 
         webView.settings.apply {
+            userAgentString = userAgentProvider.getUserAgent()
             javaScriptEnabled = true
             domStorageEnabled = true
-            loadWithOverviewMode = true
-            useWideViewPort = true
```
subsymbolic: Why remove these two options from mobile? Overview mode was introduced to fix an issue https://github.com/duckduckgo/Android/pull/70.

CDRussell:
The result of these changes is that some responsive sites do not present differently under desktop or mobile modes. theguardian.com/uk is one such example site where we render both modes identically (despite Chrome rendering them correctly for each)

Some other responsive sites do work just fine though - arstechnica.com for example renders differently for us under desktop and mobile modes.

Will revisit fixing sites like theguardian.com/uk in another task.

PR #200, opened 2018-02-23
Add retention statistics

subsymbolic:
Asana Issue URL: https://app.asana.com/0/488551667048375/471827040453678

## Description
Add atb retentions statistics https://duck.co/help/privacy/atb

## Steps to Test this PR:
**TEST 1: Installation**
1. On installation ensure requests to atb and exti are made

**TEST 2: Searching**
1. When searching ensure that the atb parameter is appended to the request
1. Ensure that each search triggers a retention (set_atb) update via the atb.js call

Refer to Asana ticket and unit tests for deeper detail.

**NOTE:** Once in beta we will get the backend team to validate that our traffic looks correct


CDRussell:
>When searching ensure that the atb and set_atb parameters are appended to the request
I'm only seeing `atb` value being included in the request; no sign of the `set_atb` value.

screen shot 2018-02-27 at 09 54 59

Can you clarify if both should be included when searching, @subsymbolic ?

subsymbolic:
Good catch @CDRussell. I have updated the PR description. I initially thought the set_atb parameter was needed in both requests however it is only needed by the atb.js request.

app/src/main/java/com/duckduckgo/app/statistics/api/StatisticsRequester.kt
```diff
@@ -0,0 +1,78 @@
...
+    StatisticsUpdater {
+
+    override fun initializeAtb() {
+
+        if (store.hasInstallationStatistics) {
+            Timber.v("Atb already initialized")
+            return
+        }
+
+        service.atb()
```
CDRussell: We should handle the lint warning here about not doing anything with the result of `.subscribe()`.

Also, it isn't clear why this is needed `.observeOn(AndroidSchedulers.mainThread())` as we don't appear to do anything UI thread related. Doesn't it feel more natural to call through to a `data store` on an `IO` thread anyways, rather than the UI thread?

app/src/main/java/com/duckduckgo/app/statistics/api/StatisticsRequester.kt
```diff
@@ -0,0 +1,78 @@
...
+
+        val atb = store.atb
+        val retentionAtb = store.retentionAtb
+
+        if (atb == null || retentionAtb == null) {
+            initializeAtb()
+            return
+        }
+
+        service.updateAtb(atb, retentionAtb)
```
CDRussell: Same questions as above re main thread and handling the return value (or ignoring the lint warning)

app/src/main/java/com/duckduckgo/app/statistics/api/StatisticsRequester.kt
subsymbolic: Good point, there's no need to jump back on the main thread so I'll remove that. As for the warning, I'll suppress it. I'm not seeing it when I run `analyze` where are you seeing it?

CDRussell: Android Studio is highlighting it; maybe only happening in the newest versions of it.

screen shot 2018-02-27 at 11 43 47

subsymbolic: Definitely not seeing it. I'll update my tools and see if that makes a difference. Are you on the beta or canary channel?

CDRussell: Beta.

I can confirm it isn't flagged on AS 3.0.1 stable.

If you just want to know the suppress command, it's `@SuppressLint("CheckResult")`. If it was just a beta-quirk, I'd happily not take any action, but it looks like a valid thing to flag given it's marked with `@CheckReturnValue`.

CDRussell: >I have no idea why it's complaining about that second one where I do use the value!

I don't see where it is used. There is stuff happening inside the `subscribe` block, but the return value from this isn't assigned to anything.

subsymbolic: In `refreshRetentionAtb` the vlaue is used in the subscribe block in `store.retentionAtb = it.version`

CDRussell: That's not the complaint, though. It isn't about what is going on in the subscribe block.

The complain is that `.subscribe()` itself returns a value (specifically, a `Disposable`) and we aren't doing anything with it. 

i.e., 

    val disposable = service.updateAtb(atb, retentionAtb)
            .subscribeOn(Schedulers.io())
            .subscribe({
                Timber.v("Atb refresh succeeded")
                store.retentionAtb = it.version
            }, {
                Timber.v("Atb refresh failed with error ${it.localizedMessage}")
            }) 

This makes that lint warning go away, as you are assigning the return value. (it introduces another lint warning, though, since we don't use `disposable`.

In any case, ignoring the lint warning is the way to go.

PR #215, opened 2018-03-07
Move browsing behavior into a fragment for tabs reuse

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/574254613727475
Tech Design URL: https://app.asana.com/0/481882893211075/543574221254520

**Description**:
Refactor our browser functionality into a Fragment to support upcoming tabs implementation. This task was a restructure - there should be no noticeable difference to the end user.

NOTE: This PR has a large footprint as a lot of code was moved around, git hasn't done a great job of recognizing this. In terms of scope, I think fixing any issues that were introduced by this PR makes sense, however, fixing code that is identical but has just moved does not.

**Steps to test this PR**:
**TEST 1**
1. Launch the app
2. Navigate to a webpage and the press back ensuring everything still holds together

**TEST 2**
Perform the following
• Share link/text from external app to our app
• Open the app via the widget
• Open the app via the assistant

For each of the following states:
• App closed
• App running and on home page
• App running and on browser page

---
###### Internal references:


CDRussell:
> I think fixing any issues that were introduced by this PR makes sense, however, fixing code that is identical but has just moved does not

makes sense 👍

CDRussell:
There is a regression around multi-window. 

• Perform a search, such that the WebView is showing some content
• Invoke multi-window

**Expected** 
No change to the `WebView` contents

**Actual**
`WebView` hidden; user returned to initial landing page view.

ℹ️ On `develop`, there is a related oddity but slightly different. On `develop`, the same actions result in the `WebView` being reloaded, which is slightly better but still wrong.

We need to fix this somewhere, but up to you if that's included in your work or picked up sometime before tabs go live.

subsymbolic:
I'm happy either way, I think it makes sense to pick up the regression here and just get it fixed across the board. If this is blocking you, however, I'm also happy to merge now and I can send you a separate PR for it.

CDRussell:
Attempting to add a new bookmark crashes the app after accepting the dialog.

    java.lang.ClassCastException: com.duckduckgo.app.browser.BrowserActivity cannot be cast to com.duckduckgo.app.bookmarks.ui.BookmarkAddEditDialogFragment$BookmarkDialogCreationListener
        at com.duckduckgo.app.bookmarks.ui.BookmarkAddEditDialogFragment.userAcceptedDialog(BookmarkAddEditDialogFragment.kt:74)
        at com.duckduckgo.app.bookmarks.ui.BookmarkAddEditDialogFragment.access$userAcceptedDialog(BookmarkAddEditDialogFragment.kt:32)
        at com.duckduckgo.app.bookmarks.ui.BookmarkAddEditDialogFragment$onCreateDialog$alertBuilder$1.onClick(BookmarkAddEditDialogFragment.kt:52)
        at android.support.v7.app.AlertController$ButtonHandler.handleMessage(AlertController.java:162)
        at android.os.Handler.dispatchMessage(Handler.java:105)
        at android.os.Looper.loop(Looper.java:172)
        at android.app.ActivityThread.main(ActivityThread.java:6637)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)

subsymbolic:
Yikes! Ok some wrong import there or something. Good catch will fix.

CDRussell:
>If this is blocking you, however, I'm also happy to merge now

This is the big blocker which is basically causing the current (effective) code freeze, right? Once this is merged, your other work will be more normal branching / merging strategy?

If so, then yeah, i'm happy to defer any minor bugs to enable us to work from the same browser code base again

CDRussell:
I'm noticing inconsistent behaviour with the keyboard showing / not showing when launched from as the assist app, and launched from the widget. This is mostly random it seems when the app is launched when not already running.

This may well not be caused by this branch though, so just flagging as something else we need to include in the final tabs tidy up (I'm more than happy to help tidy up all these little loose ends btw)

CDRussell:
Does `private var pendingFileDownload: PendingFileDownload? = null` belong in the `Activity` or the `Fragment`?

If it remains in the `Activity`, is there ever the chance that one tab's download attempt will collide with another's?

CDRussell:
For the bookmarks crash, it might be enough to:

## Show the dialog differently
Show the dialog using the `childFragmentManager`

**Before**
`addBookmarkDialog.show(it.supportFragmentManager, ADD_BOOKMARK_FRAGMENT_TAG)`

**After**
`addBookmarkDialog.show(childFragmentManager, ADD_BOOKMARK_FRAGMENT_TAG)`

## Invoke callback differently
In the `userAcceptedDialog` method

**Before**
`val listener = activity as BookmarkDialogCreationListener`

**After**
`val listener = parentFragment as BookmarkDialogCreationListener`

⚠️ Haven't tested thoroughly though; just wanted to stop the crash locally so i could test out the rest of bookmarks

CDRussell:
This is great! If we get the bookmarks crash sorted, i'm happy deferring the other observed problems as either being existing or new but will be fixed before tabs goes out.

subsymbolic:
@CDRussell digging into the bookmarks implementation I don't like the hard casting of listeners as it makes the fragment fragile. I have refactored to remove that, let me know what you think.

I am looking at the other feedback now.

subsymbolic:
Regarding the pending file download, it currently behaves indetnically to before. It will get refactored further in the mutli tabs branch.

subsymbolic:
@CDRussell just fixed the multi window state issue. I think that's everything other than your keyboard observations which I think we need to fine-tune at the end when both tour PRs are in.

PR #232, opened 2018-03-28
Increase leaderboard count

subsymbolic:
Task/Issue URL: https://app.asana.com/0/488551667048375/525633294284205
Tech Design URL: N/A

**Description**:
This task updates the preexisting logic to wait until at least 30 sites are visited before displaying the tracker networks leaderboard on the Privacy Dashboard.

Note this was not quite the 1-liner expected. It introduces a new db table to support counting the total number of unique sites visited. Some lint tidy ups were also performed.

**Steps to test this PR**:
1. Visit 30 sites, preferably sites with lots of trackers (you may want to edit `LEADERNOARD_MIN_DOMAINS` in `PrivacyDashboardViewModel.kt` to reduce this to 10)
2. Open the dashboard
3. Ensure the leaderboard is displayed

**Migration test:**
Install this PR over a preexisting version of the app (whatever is in develop) to ensure the db migration succeeds.

****

---
###### Internal references:


brindy:
On iOS I made the number of sites dependent on the type of build (debug vs release) though mainly so that I could capture screenshots in the UI tests without having to visit 30 sites. Worth doing this on Android as well?

subsymbolic:
I think the debug option is a good idea although I don't think we need it yet and I'm not keen to refactor the unit tests to add it. How do you feel about us deferring that until we do screenshots? We can have a different flavor for screenshots and have the option in there.

app/src/main/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModel.kt
```diff
@@ -62,32 +62,39 @@ class PrivacyDashboardViewModel(private val settingsStore: PrivacySettingsStore,
...
-                networkTrackerSummaryPercent1 = if (showSummary) networkPercents[0].percent else 0.0f,
-                networkTrackerSummaryPercent2 = if (showSummary) networkPercents[1].percent else 0.0f,
-                networkTrackerSummaryPercent3 = if (showSummary) networkPercents[2].percent else 0.0f
+            showTrackerNetworkLeaderboard = showTrackerNetworkLeaderboard(domainCount, networkTally.count()),
+            trackerNetworkTally = networkTally
         )
     }
 
+    private fun showTrackerNetworkLeaderboard (domainCount: Int, networkCount: Int): Boolean {
+        return domainCount >= LEADERNOARD_MIN_DOMAINS && networkCount >= LEADERBOARD_MIN_NETWORKS
```
brindy: on iOS this is:

`sitesVisited() > visitedSitesThreshold && networksDetected().count >= 3`

Not sure which one is right tbh.

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.0'
+        classpath 'com.android.tools.build:gradle:3.2.0-alpha08'
```
brindy: was this upgrade to an alpha dependency deliberate?

subsymbolic: Yes, Android Studio Preview 3.2 requested the change.

app/src/main/java/com/duckduckgo/app/global/db/AppDatabase.kt
```diff
@@ -60,5 +62,11 @@ abstract class AppDatabase : RoomDatabase() {
                 database.execSQL("CREATE INDEX `index_tab_selection_tabId` on `tab_selection` (tabId)")
             }
         }
+
+        val MIGRATION_2_TO_3: Migration = object : Migration(2, 3) {
```
brindy: Doubt it will affect many folk, but should the sites visited that are already recorded in the existing table be copied in to the new table?

Happy to merge either way.

subsymbolic: That's a good point. The old `network_leaderboard` data isn't valid without the total domain count so the percentages will be all over the place. Adding the domains to the new table doesn't fix that as there's still data missing for all the sites visited that didn't have trackers. I've gone ahead and cleared the old `network_leaderboard` data so we can start gathering fresh valid data.

app/src/main/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModel.kt
subsymbolic: Good point. It's not clear to me either. I have updated android for consistency.

subsymbolic:
@brindy I've made db migration changes so probably worth a re-review before I merge.

PR #234, opened 2018-04-01
Added italian translation

anthologist:


ghavaun:
ghavaunbrooklyn@gmail.com

subsymbolic:
Thank you for submitting this. We have a translation service in place and will be using that to add additional languages in the future. We are not currently in a position to accept community PRs but that will hopefully change soon, keep an eye on our README file.

PR #236, opened 2018-04-03
Feature/downloading data uri

CDRussell:
Task/Issue URL: https://app.asana.com/0/547792610048271/565680240942081
Tech Design URL: 
CC: 

**Description**:
Adds support for saving images encoded using a data URI.

**Steps to test this PR**:
1. Visit a site which renders an image as a data URI (e.g., https://dopiaza.org/tools/datauri/examples/index.php)
1. Long press on the image, and ensure it can be saved
1. During the decoding, a notification will show that the image is downloading - similar to downloading normal images.
1. Once decoded and written to disk, the notification will show that the image is downloaded.

⚠️  Ensure you test on an Oreo device and a pre-Oreo device; notifications changed then. If the notification isn't registered to an appropriate channel upfront, the notification will never show.

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/DataUriParserTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/DataUriParserTest.kt
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2018 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.downloader
+
+import com.duckduckgo.app.browser.downloader.DataUriParser.ParseResult.Invalid
+import com.duckduckgo.app.browser.downloader.DataUriParser.ParseResult.ParsedDataUri
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Test
+
+class DataUriParserTest {
+
+    private lateinit var testee: DataUriParser
+
+    @Before
+    fun setup() {
+        testee = DataUriParser()
+    }
+
+    @Test
+    fun whenMimeTypeProvidedAsImagePngThenPngSuffixGenerated() {
+        val parsed = testee.generate("data:image/png;base64,AAAA") as ParsedDataUri
+        assertEquals("png", parsed.filename.fileType)
+    }
+
+    @Test
+    fun whenMimeTypeProvidedAsImageJpegThenJpegSuffixGenerated() {
+        val parsed = testee.generate("data:image/jpeg;base64,AAAA") as ParsedDataUri
+        assertEquals("jpeg", parsed.filename.fileType)
+    }
+
+    @Test
+    fun whenMimeTypeProvidedAsWildcardImageTypeThenDefaultJpgSuffixGenerated() {
+        val parsed = testee.generate("data:image/*;base64,AAAA") as ParsedDataUri
+        assertEquals("jpg", parsed.filename.fileType)
+    }
+
+    @Test
+    fun whenMimeTypeProvidedAsArbitraryImageTypeThenSuffixGeneratedWithThatType() {
+        val parsed = testee.generate("data:image/foo;base64,AAAA") as ParsedDataUri
+        assertEquals("foo", parsed.filename.fileType)
+    }
+
+    @Test
+    fun whenMimeTypeNotProvidedThenNoSuffixAdded() {
+        val parsed = testee.generate("data:,AAAA") as ParsedDataUri
+        assertEquals("", parsed.filename.fileType)
+    }
+
+    @Test
+    fun whenInvalidDataUriProvidedInvalidTypeTurned() {
+        val parsed = testee.generate("AAAA")
+        assertTrue(parsed === Invalid)
+    }
+
+    @Test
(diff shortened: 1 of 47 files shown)


app/src/androidTest/java/com/duckduckgo/app/browser/downloader/DataUriParserTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    fun whenMimeTypeProvidedAsImageJpegThenJpegSuffixGenerated() {
+        val parsed = testee.generate("data:image/jpeg;base64,AAAA") as ParsedDataUri
+        assertEquals("jpeg", parsed.filename.fileType)
+    }
+
+    @Test
+    fun whenMimeTypeProvidedAsWildcardImageTypeThenDefaultJpgSuffixGenerated() {
+        val parsed = testee.generate("data:image/*;base64,AAAA") as ParsedDataUri
+        assertEquals("jpg", parsed.filename.fileType)
+    }
```
brindy: Given that the user must have initiated this from an actual image I guess we're safe to assume it really is an image, but if we don't know what type then this feels as though we could easy be getting the image type wrong as often as we get it right.  

For images of unknown type we could use `BitmapFactory` to decode them, then use `Bitmap.compress` to re-encode them in to a known type (e.g. JPG).  One downside I see with this is animated images would likely just end up being a single frame, however I think that is better than just assuming they are a JPG.

app/src/main/java/com/duckduckgo/app/browser/WebViewLongPressHandler.kt
```diff
@@ -40,11 +40,11 @@ interface LongPressHandler {
 
 class WebViewLongPressHandler @Inject constructor() : LongPressHandler {
 
-    override fun handleLongPress(longPressTargetType: Int, longPressTargetUrl: String, menu: ContextMenu) {
+    override fun handleLongPress(longPressTargetType: Int, longPressTargetUrl: String?, menu: ContextMenu) {
         when (longPressTargetType) {
             WebView.HitTestResult.IMAGE_TYPE,
             WebView.HitTestResult.SRC_IMAGE_ANCHOR_TYPE -> {
-                if(URLUtil.isNetworkUrl(longPressTargetUrl)) {
+                if(URLUtil.isNetworkUrl(longPressTargetUrl) || URLUtil.isDataUrl(longPressTargetUrl)) {
```
brindy: Needs space after `if` (didn't used to bother flagging these but I can see changes elsewhere suggest `if (` is the format we're going after.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/DataUriParserTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    @Test
+    fun whenInvalidDataUriProvidedInvalidTypeTurned() {
+        val parsed = testee.generate("AAAA")
+        assertTrue(parsed === Invalid)
+    }
+
+    @Test
+    fun whenMimeTypeProvidedAsNonImageTypeThenNoSuffixGenerated() {
+        val parsed = testee.generate("data:text/plain;base64,AAAA") as ParsedDataUri
+        assertEquals("", parsed.filename.fileType)
```
brindy: Not sure why an image would be of type "text/plain", presumably this is just to test the cases relevant to this task, but I would expect the file type to be parsed out as "txt".

app/src/main/java/com/duckduckgo/app/browser/downloader/DataUriParser.kt
```diff
@@ -0,0 +1,98 @@
...
+
+        val suffix = determineSuffix(fileTypeGeneral, fileTypeSpecific)
+        val filename = UUID.randomUUID().toString()
+        val generatedFilename = GeneratedFilename(name = filename, fileType = suffix)
+
+        return ParsedDataUri(fileTypeGeneral, fileTypeSpecific, data, mimeType, generatedFilename)
+    }
+
+    private fun determineSuffix(fileTypeGeneral: String, fileTypeSpecific: String): String {
+        return when (fileTypeGeneral) {
```
brindy: Could use `MimeTypeMap` here? 

See https://developer.android.com/reference/android/webkit/MimeTypeMap.html#getExtensionFromMimeType(java.lang.String)

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/DataUriParserTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    @Test
+    fun whenMimeTypeProvidedAsImagePngThenPngSuffixGenerated() {
+        val parsed = testee.generate("data:image/png;base64,AAAA") as ParsedDataUri
+        assertEquals("png", parsed.filename.fileType)
+    }
+
+    @Test
+    fun whenMimeTypeProvidedAsImageJpegThenJpegSuffixGenerated() {
+        val parsed = testee.generate("data:image/jpeg;base64,AAAA") as ParsedDataUri
+        assertEquals("jpeg", parsed.filename.fileType)
```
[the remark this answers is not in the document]
CDRussell: Good point, and that's what happens now because of `getExtensionFromMimeType`

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/DataUriParserTest.kt
CDRussell: I'm just going to defer all of this logic to `getExtensionFromMimeType` instead of us trying to be clever.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/DataUriParserTest.kt
CDRussell: Agreed, and now happens because of `getExtensionFromMimeType`

app/src/main/java/com/duckduckgo/app/browser/downloader/DataUriParser.kt
CDRussell: Inspired! I looked for this and failed to find it. And it fixes every other piece of feedback you've raised too! 🙏

PR #262, opened 2018-05-01
Hotfix/5.3.1

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/654220091942217

**Description**:
Explicitly disables Google SafeBrowsing which will be on by default from WebView versions 66 onwards.

**Steps to test this PR**:
1. Visit http://testsafebrowsing.appspot.com/ and click on the first link
1. The page should load
1. Compare the new `meta-data` tag with the documentation to ensure it matches the docs (https://developer.android.com/reference/android/webkit/WebView)

You can repeat the above test from `develop` and the page should fail to load (assuming you have v66 of Chrome on your device). Note, it won't fail in a nice way - it will just kinda stall as we haven't implemented the `SafeBrowsing` callbacks to show a nice UI.

---
###### Internal references:

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,7 +6,7 @@ apply plugin: 'kotlin-kapt'
 apply from: '../versioning.gradle'
 
 ext {
-    VERSION_NAME = "5.3.0"
+    VERSION_NAME = "5.3.1"
     USE_ORCHESTRATOR = project.hasProperty('orchestrator') ? project.property('orchestrator') : false
 }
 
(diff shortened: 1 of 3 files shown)


nilnilnil:
👍 nice, thanks.

brindy:
Is it worth a comment in the manifest to direct people browsing the source code to some information? The name of that meta data item makes it sounds like browsing with us won't be safe.

CDRussell:
@brindy ?

>

brindy:
Yep, LGTM

PR #266, opened 2018-05-03
Fix share failure

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/611385370877136
Tech Design URL: N/A

**Description**:
Removes the manual launching of share intents into tabs allowing live data updates to trigger them. This fixes an issue where sharing from an external app such as Chrome sometimes failed, resulting in a blank tab.

**Steps to test this PR**:
1. With the app open, share a url from chrome to our app and ensure the shared url displays
1. With the app closes, share a url from chrome to our app and ensure the shared url displays
1. Now repeat the above steps with developer settings set to "Don't keep activities" on and "Background process limit" set to one
1. Reset the settings changed in the previous step back to defaults

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -19,7 +19,6 @@ package com.duckduckgo.app.browser
 import android.arch.core.executor.testing.InstantTaskExecutorRule
 import android.arch.lifecycle.Observer
 import android.arch.persistence.room.Room
-import android.net.Uri
 import android.support.test.InstrumentationRegistry
 import android.view.MenuItem
 import android.view.View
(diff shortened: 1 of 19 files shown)


CDRussell:
On the whole, many of the scenarios which were failing before are working great! 

There is one case where I can still reproduce the behaviour.

1. set "background process limit" to 1
1. enter tabs screen
1. close all tabs
1. switch to another app (e.g., Chrome) from tabs screen (so that there are no tabs showing at all)
1. share back to DDG app

I see two tabs open then; but the new tab screens wins and is shown in the foreground whereas the shared link is on the other tab.

subsymbolic:
Thanks for the review @CDRussell. That last case **should** be covered by the PR in https://github.com/duckduckgo/Android/pull/267. Let me know if you still see it there too.

subsymbolic:
It may be easier if I combine them - I'll close this one off and we can just work of https://github.com/duckduckgo/Android/pull/267

PR #268, opened 2018-05-09
Feature/default browser

CDRussell:
Task/Issue URL: https://app.asana.com/0/72649045549333/608078252761679
Tech Design URL: 
CC: 

**Description**:
Add some UI to encourage users to set us as their default browser.

Specifically, we add an additional step to onboarding; add a place in settings which shows whether we're already the default or not and can be clicked to go set it; and a timed prompt which will appear after a few days of usage.

## A/B Testing
There is the introduction of some simple feature flags which can belong to a variant. The idea is that we can allocate a user to a variant, and that particular variant will have 0 or more features enabled for it.

These are defined in `VariantManager`. Currently, there are 3 variants defined

     val ACTIVE_VARIANTS = listOf(
            Variant(key = "mw", weight = 25.0, features = listOf(ShowInOnboarding)),
            Variant(key = "mx", weight = 25.0, features = listOf(ShowInOnboarding, ShowTimedReminder)),
            Variant(key = "my", weight = 50.0, features = emptyList())
        )

Unlike iOS version, the weights don't have to add up to any specific number and are just relative to each other. They will be normalised internally. 

So there's 50% chance you won't see any of these features, 25% you'll see only the extra screen in onboarding and 25% chance you'll see both the extra onboarding screen as well as the timed reminder.

This will complicate your testing somewhat. If you want to see which variant you were allocated, monitor logcat for `Current variant is null; allocating new one`. You might want to locally hack the percentages to try out different variants. 

**Steps to test this PR**:

### Onboarding test
1. Install the app clean (or visit privacy features from settings) to see onboarding screens
1. Ensure there is now a 3rd onboarding screen
1. Test that settings button takes you to settings screen
1. Test that continue button dismisses onboarding

### Settings screen test
1. Visit settings
1. Confirm radio button for default browser is set to correct value
1. Click on it to visit settings
1. Exit without making a change
1. Confirm radio button resets to reflect reality
1. Click on it again and set DDG as default browser
1. Return to app and confirm radio button set correctly

### Showing Notification banner 
1. Wait long enough for the default grace period to pass (3 days in production; 20 seconds in develop)
1. Type full URL into omnibar - confirm no default browser banner shown
1. Search for something instead - confirm default browser notification is shown
1. Tap the notification banner - confirm info dialog shown (it's actually a new activity but looks like a dialog)
1. Confirm settings button takes you to system settings

### Dismissing notification banner

1. Click `X` button on notification banner. 
1. Confirm it disappears
1. Perform another search - confirm it doesn't come back (it'll never come back now).

### Searching before time elapsed test

1. Clear app data
1. try searching again and perform a search within 20 seconds of the app launching - confirm no notification banner show

## Pre-Nougat tests
1. Test on device older than Nougat - you should **never** see any options about default browser.

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/statistics/api/StatisticsRequester.kt
```diff
@@ -39,19 +44,20 @@ class StatisticsRequester(private val store: StatisticsDataStore, private val se
...
 
+        val variant = variantManager.getVariant()
+
         service.atb(ParamValue.appVersion)
             .subscribeOn(Schedulers.io())
             .flatMap {
-                store.atb = it.versionWithVariant
-                store.retentionAtb = it.version
-                service.exti(it.versionWithVariant, ParamValue.appVersion)
+                val fullAtb = it.version + variant.key
```
subsymbolic: The variant often needs to be interrogated independently of the atb, why store them together? I would suggest storing each independently and using a convenience method on the store e.g `atbWithVariant` to retrieve them together when needed.

app/src/main/java/com/duckduckgo/app/statistics/model/Atb.kt
```diff
@@ -16,12 +16,4 @@
...
-data class Atb(val version: String) {
-
-    val versionWithVariant: String get() = version + VARIANT
-
-    companion object {
-        const val VARIANT = "ma"
-    }
-
-}
+data class Atb(val version: String)
```
subsymbolic: Nice!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -285,6 +295,11 @@ class BrowserTabViewModel(
         )
 
         if (duckDuckGoUrlDetector.isDuckDuckGoQueryUrl(url)) {
+
+            newViewState = newViewState.copy(
+                    omnibarText = duckDuckGoUrlDetector.extractQuery(url) ?: "",
```
subsymbolic: Why update the omnibar text again? It is already happening above `omnibarTextForUrl(url)`

app/src/main/res/values/colors.xml
```diff
@@ -29,12 +29,20 @@
     <color name="brickOrange">#d03a10</color>
 
     <color name="powderBlue">#50BFFF</color>
+    <color name="powderBlue2">#B2D6F3</color>
     <color name="windowsBlue">#3c90c0</color>
+    <color name="softBlue">#66afe9</color>
 
-    <color name="lighOliveGreen">#93c04d</color>
+    <color name="lightOliveGreen">#93c04d</color>
```
subsymbolic: Nice tidy up!

app/src/main/res/values-sw600dp/dimens.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
```
subsymbolic: Formatting

app/src/main/java/com/duckduckgo/app/statistics/store/StatisticsSharedPreferences.kt
```diff
@@ -18,21 +18,41 @@ package com.duckduckgo.app.statistics.store
...
         get() = preferences.getString(KEY_ATB, null)
-        set(atb) = preferences.edit().putString(KEY_ATB, atb).apply()
+        set(atb) = preferences.edit { putString(KEY_ATB, atb) }
 
     override var retentionAtb: String?
         get() = preferences.getString(KEY_RETENTION_ATB, null)
-        set(retentionAtb) = preferences.edit().putString(KEY_RETENTION_ATB, retentionAtb).apply()
+        set(value) = preferences.edit { putString(KEY_RETENTION_ATB, value) }
+
+    override fun saveAtb(fullAtb: String, retentionAtb: String) {
```
subsymbolic: I mentioned this earlier too - I think the variant and atb should be separated

app/src/androidTest/java/com/duckduckgo/app/statistics/ExperimentationVariantManagerTest.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test(expected = IllegalArgumentException::class)
+    fun whenNoVariantsAvailableThen() {
+        whenever(mockStore.variant).thenReturn("foo")
+
+        assertEquals("foo", testee.getVariant(activeVariants).key)
+    }
+
+    @Test
+    fun whenVariantPersistedIsNotFoundInActiveVariantListThenNewVariantAllocated() {
```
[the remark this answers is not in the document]
CDRussell: Doesn't this mean once a user has been part of a variant once, they will _never_ be given a new variant?

Even after we release a new version of the app with a new list of variants, they will have their now redundant variant code saved. If we didn't allocate them a new one, they'd be effectively trapped forever out of a valid variant, no?

app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        sealed class DefaultBrowserFeature : VariantFeature() {
+            object ShowInOnboarding : DefaultBrowserFeature()
+            object ShowTimedReminder : DefaultBrowserFeature()
+        }
+    }
+
+    companion object {
+
+        // there must always be at least one active variant defined here
```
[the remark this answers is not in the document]
CDRussell: Mimicking existing behaviour I guess - where there was always a variant code in use `ma`. I think there might be a convention server-side to use underscore to denote _no active variant_.

So if the list of active variants is empty, we could fallback to `m_` as the variant code. Or would you expect us to _not_ submit any variant information in that case?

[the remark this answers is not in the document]
subsymbolic: This PR supersedes existing behavior. In the past we always had a variant but now we only want to add them for active experiments. When we are not running an experiment there should be no active variants.

app/src/androidTest/java/com/duckduckgo/app/statistics/ExperimentationVariantManagerTest.kt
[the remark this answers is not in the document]
subsymbolic: That's a good question. My understanding is that we're only thinking in terms of installation experiments at the moment. Are you ok with existing users, who may have already seen onboarding being reallocated? I think that may mess up your data. Incidentally, the app uses the new variant in some places and the old variant in others e.g your searches are showing the old one.

app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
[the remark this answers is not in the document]
CDRussell: As per discussion on a call, we will ensure list can be empty.

Furthermore, we will treat the case whereby a user is in no variant as them having `""` an empty string for their variant code. This will preserve existing logic of appending it onto the ATB version but will obviously not add anything on when there is no variant in use.

app/src/androidTest/java/com/duckduckgo/app/global/db/AppDatabaseTest.kt
```diff
@@ -43,27 +39,6 @@ class AppDatabaseTest {
         testHelper.runMigrationsAndValidate(TEST_DB_NAME, 3, true, AppDatabase.MIGRATION_2_TO_3)
     }
 
-    @Test
```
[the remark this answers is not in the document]
CDRussell: Yes, it was deliberate. There is so much bad voodoo going on with the CI system right now, and this was one of the tests that would often fail.

I'm happy to revisit this and add it back in once we have a solid CI again. But marking it as `@Ignore` just seems like more baggage than it's worth, in the same way we wouldn't accept commented out code. 

Added as task to backlog to revisit. https://app.asana.com/0/414730916066338/681857984148014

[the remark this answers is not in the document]
subsymbolic: My two motivations for not wanting to remove it altogether are:
• This is a valuable migration test that runs fine locally (just not on travis) and I still want to run it before releasing the app
• The removal is diagnostic, we may find the test is not to blame for the failure we were seeing. I *want* the annoying reminder to sort it out by either unignoring it or removing it once the diagnosis is done

Should this change be part of this PR?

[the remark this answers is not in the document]
CDRussell: Restoring

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
CDRussell: `omnibarTextForUrl(url)` must have come in from a recent merge, done after i'd made that change above. Will remove the second query extraction.

app/src/main/res/values-sw600dp/dimens.xml
CDRussell: That's strange; my formatter wants to do this for all XML files. Can you confirm what your formatter does for this file, @subsymbolic ?

app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
```diff
@@ -83,11 +87,18 @@ open class DuckDuckGoApplication : HasActivityInjector, HasServiceInjector, HasS
...
         configureDataDownloader()
+        recordInstallationTimestamp()
 
         migrateLegacyDb()
         notificationRegistrar.registerApp()
     }
 
+    private fun recordInstallationTimestamp() {
+        if (!appInstallStore.hasInstallTimestampRecorded()) {
+            appInstallStore.installTimestamp = System.currentTimeMillis()
```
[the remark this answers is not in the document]
CDRussell: For existing users, `installation timestamp` will actually mean time they upgraded to this app version, which is admittedly not the best name for that.

The need for the timestamp is that we want to be able to ensure they have been using the app for a minimum amount of time before showing them the default browser prompt. I don't mind that we'll have a timestamp for both new installs and upgrades. Otherwise, it will tricky to differentiate when to generate it as we don't have another way of detecting a new install vs an upgrade. At least not a way springing to mind - you have a plan for detecting this scenario?

app/src/main/res/values-sw600dp/dimens.xml
subsymbolic: Interesting, mine doesn't change it.

CDRussell: That's annoying. I've reverted the change, and we can sync up our formatting settings later.

Issue #168, opened 2018-02-01
Address should start with https:// after loading page


satyammittal:
Currently whenever I loads some webpage by entering address ex: https://www.duckduckgo.com

address bar looks like this
![screenshot_20180201-102711](https://user-images.githubusercontent.com/15922592/35661899-20d66a9a-073b-11e8-9cd2-325a52c31768.png)
It should look like this (start with https:// so that we can easily see the webpage domain name)
![screenshot_20180201-102728](https://user-images.githubusercontent.com/15922592/35661917-400ca29e-073b-11e8-93c4-c8db7a90314a.png)

Version: 4.1.2

subsymbolic:
This is great feedback, thanks for taking the time to share it. I have passed it on to our design team. They have plans to display a different view altogether (with https status) when in non-editing mode, this will hopefully make things clearer.

subsymbolic:
This issue still persists, we don't have time to look at it right now but will leave it open in case someone in the community has time.

satyammittal:
I would like to work on this issue.

subsymbolic:
That's great @satyammittal. Feel free to send us a PR when you're ready. If you have any questions or are making any architectural changes ping us either here or in the PR to chat through your approach.

PR #281, opened 2018-06-02
Move address bar text to start after loading page

WilderPereira:
Task/Issue URL: #168 

**Description**:
After loading a webpage with a large address it wasn't easy to see the webpage domain name. This PR fixes this issue.

**Steps to test this PR**:
- Open the app and enter a large url 
- Click on a link that loads a large url

solved_url

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -341,13 +341,6 @@ class BrowserTabFragment : Fragment(), FindListener {
 
         if (shouldUpdateOmnibarTextInput(viewState, viewState.omnibarText)) {
             omnibarTextInput.setText(viewState.omnibarText)
-
-            // ensures caret sits at the end of the query
-            omnibarTextInput.post {
-                omnibarTextInput?.let {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -347,6 +347,8 @@ class BrowserTabFragment : Fragment(), FindListener {
                 omnibarTextInput?.let {
                     it.setSelection(it.text.length)
                 }
+                // ensures the begin of the url is always displayed
```
CDRussell: Thanks for making this PR!

Looking at the change, I see it's basically undoing the lines immediately above it. i.e., the lines above set the selection to end of the text, and your change sets the selection back to the start.

I think the previous functionality of setting the selection to the end can be removed entirely now (it was needed for a previous omnibar design).

Would you mind testing if the existing `setSelection` call is removed, do we still need your change to set it to 0 again? I suspect we can remove all `setSelection` calls now.

If you're happy to test that out and deleting the code if that looks good, then we can get your PR merged in.

CDRussell:
That's great @WilderPereira; merged. Many thanks for the contribution! 👍

PR #283, opened 2018-06-04
Feature/https db write performance

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/688772474867111
Tech Design URL: 
CC: 

**Description**:
Performance issue observed while large number of HTTPS upgrade rows were being written to DB. The number of rows is north of 200k and the time taken to write this data could span tens of seconds on some devices.

While the DB was being written to, other DB writes are blocked. If the HTTPS list is written at a bad time, the user might experience a delay in seeing the browser or launching a new tab (or potentially a whole host of other bad UX).

This PR stops writing the HTTPS list as a single transaction. Instead, it breaks the list down into many smaller transactions, which can support other app write taking place throughout. The result is that the total time to write all HTTPS rows will _increase_ but that because there are opportunities for other DB writes throughout, the overall UX is greatly increased and that initial delay in seeing the browser screen is eliminated.

**Benchmarking**
screen shot 2018-06-01 at 14 43 19

I performed some (unscientific) benchmarking comparing a few things; total write time and time granted between writing chunks. Compared this across two physical devices, using both `Room 1.0.0` and `Room 1.1.0`. Based on these numbers, I decided to chunk on inserting 1,000 rows at a time and decided to upgrade `Room` based on overall improved performance.

**Consideration**
A downside is that we can no longer rely on a single DB transaction to protect the integrity of our data. Instead, I've had to add an external guard which protects against the scenario whereby

1. Data downloaded
1. Data partially written to disk
1. Process crashes or exception thrown

In this scenario, it would be possible for us to detect that the data has been successfully stored already and not finish writing the missing data.

I've added a guard which is stored in `SharedPreferences` to counter this. See the full comment on `HttpsUpgradeListDownloader` class for more details.

**Steps to test this PR**:
1. Fresh install - watch the logs for a line like `I/HttpsUpgradeListDownloader$downloadList: Received 205621 HTTPS domains; chunking by 1000 into 206 separate DB transactions`
1. As soon as this is visible in the logs, hit the `Continue` button to end onboarding
1. Observe how long it takes to present browser. Browser should show before HTTPS data is fully written (which can be observed by watching logs for `Successfully wrote HTTPS data`)

Pay particular attention to this logic to make sure this makes sense:

    if (response.isCached && dbWriteStatusStore.hasWriteCompleted() && httpsUpgradeDao.count() != 0) {
                Timber.d("HTTPS data already cached and stored")
                return@fromAction
            }

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApi.kt
+++ b/app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApi.kt
@@ -16,16 +16,12 @@
 
 package com.duckduckgo.app.autocomplete.api
 
-import com.duckduckgo.app.browser.omnibar.QueryUrlConverter
 import com.duckduckgo.app.global.UriString
 import io.reactivex.Observable
 import javax.inject.Inject
 
 
-open class AutoCompleteApi @Inject constructor(
-        private val autoCompleteService: AutoCompleteService,
-        private val queryUrlConverter: QueryUrlConverter
-) {
+open class AutoCompleteApi @Inject constructor(private val autoCompleteService: AutoCompleteService) {
 
     fun autoComplete(query: String): Observable<AutoCompleteApi.AutoCompleteResult> {
 
@@ -34,16 +30,18 @@ open class AutoCompleteApi @Inject constructor(
         }
 
         return autoCompleteService.autoComplete(query)
-                .flatMapIterable { it -> it }
-                .map { AutoCompleteSuggestion(it.phrase, UriString.isWebUrl(it.phrase)) }
-                .toList()
-                .onErrorReturn { emptyList() }
-                .map { AutoCompleteResult(query = query, suggestions = it) }
-                .toObservable()
+            .flatMapIterable { it -> it }
+            .map { AutoCompleteSuggestion(it.phrase, UriString.isWebUrl(it.phrase)) }
+            .toList()
+            .onErrorReturn { emptyList() }
+            .map { AutoCompleteResult(query = query, suggestions = it) }
+            .toObservable()
     }
 
-    data class AutoCompleteResult(val query: String,
-                                  val suggestions: List<AutoCompleteSuggestion>)
+    data class AutoCompleteResult(
+        val query: String,
+        val suggestions: List<AutoCompleteSuggestion>
+    )
 
     data class AutoCompleteSuggestion(val phrase: String, val isUrl: Boolean)
 
--- a/app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
+++ b/app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
@@ -59,11 +59,7 @@ class HttpsUpgradeListDownloader @Inject constructor(
             if (response.isSuccessful) {
                 Timber.d("Updating HTTPS upgrade list from server")
 
-                val domains = response.body()
-                if (domains == null) {
-                    Timber.w("Failed to obtain HTTPS upgrade list")
-                    return@fromAction
-                }
+                val domains = response.body() ?: throw IllegalStateException("Failed to obtain HTTPS upgrade list")
 
                 val startTime = System.currentTimeMillis()
 
--- a/app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
+++ b/app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
@@ -32,7 +32,8 @@ import javax.inject.Inject
  * The network cache could therefore deliver the cached value and we could skip writing to the DB thinking we'd already successfully written it all.
  *
  * To counter this, we need an external mechanism to track whether the write completed successfully or not.
- * Using this, we can detect if the write failed when we next download the data. We'll only skip the DB inserts if cache returns the list and the write flag is set such that it finished writing successfully.
+ * Using this, we can detect if the write failed when we next download the data. We'll only skip the DB inserts if cache returns the list and the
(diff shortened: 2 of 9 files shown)


app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApi.kt
```diff
@@ -16,16 +16,12 @@
...
 import com.duckduckgo.app.global.UriString
 import io.reactivex.Observable
 import javax.inject.Inject
 
 
-open class AutoCompleteApi @Inject constructor(
-        private val autoCompleteService: AutoCompleteService,
-        private val queryUrlConverter: QueryUrlConverter
-) {
+open class AutoCompleteApi @Inject constructor(private val autoCompleteService: AutoCompleteService) {
```
nilnilnil: Is this an exogenous change?

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
```diff
@@ -38,21 +51,47 @@ class HttpsUpgradeListDownloader @Inject constructor(
...
                     httpsUpgradeDao.deleteAll()
-                    httpsUpgradeDao.insertAll(*domains)
+                    Timber.v("Took ${System.currentTimeMillis() - startTime}ms to delete existing records")
+
+                    val chunks = domains.chunked(chunkSize)
+                    Timber.i("Received ${domains.size} HTTPS domains; chunking by $chunkSize into ${chunks.size} separate DB transactions")
+
+                    chunks.forEach {
+                        database.runInTransaction {
+                            httpsUpgradeDao.insertAll(it)
```
nilnilnil: Is `it` really a special sygil here?

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
```diff
@@ -38,21 +51,47 @@ class HttpsUpgradeListDownloader @Inject constructor(
...
+
             } else {
                 throw IOException("Status: ${response.code()} - ${response.errorBody()?.string()}")
             }
         }
     }
+
+    private inline fun modifyHttpsDatabase(function: () -> Unit) {
+        dbWriteStatusStore.updateStatus(false)
+        function()
```
nilnilnil: This seems odd that it can run any function, right?

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApi.kt
CDRussell: Use smaller words! 😆 

The change wasn't needed or caused by anything; just noticed this class had an unused property and including here as a general tidy-up.

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
CDRussell: `it` here is implicitly available because `chunks.forEach` takes a lambda with a single parameter. 

so `it` == `an individual chunk` in that scope

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
CDRussell: I made the `modifyHttpsDatabase` function to hopefully draw enough attention to the fact that the action, whatever it may be, is immediately wrapped with calls to update the status store state.

It can run any function, you're right. But hopefully given it's existence in this class, it's clear enough that you'd want to wrap a bunch of code which is going to interact with the DB. This function could easily be removed and we could just leave the `dbWriteStatusStore.updateStatus()` calls top and bottom of the rest of the code, but I had hoped this would draw attention to them in the right way.

I can address if it's making it less clear?

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
```diff
@@ -38,21 +51,47 @@ class HttpsUpgradeListDownloader @Inject constructor(
             val call = service.https()
             val response = call.execute()
 
-            if (response.isCached && httpsUpgradeDao.count() != 0) {
+            if (response.isCached && dbWriteStatusStore.hasWriteCompleted() && httpsUpgradeDao.count() != 0) {
```
brindy: Couple of points:

* Probably don't need the count check?

* Also, might be more elegant to store the etag once all the records are inserted and compare the etag in the response vs the one stored?

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeListDownloader.kt
```diff
@@ -18,41 +18,75 @@ package com.duckduckgo.app.httpsupgrade.api
...
-    fun downloadList(): Completable {
+    fun downloadList(chunkSize: Int = INSERTION_CHUNK_SIZE): Completable {
 
         Timber.d("Downloading HTTPS Upgrade data")
 
         return Completable.fromAction {
 
             val call = service.https()
             val response = call.execute()
+            val eTag = response.headers().get("etag")
```
brindy: constant for this? (optional)

Headers are case-insensitive - does this class still return the etag if the header comes back as say "ETag"? (it's kind of moot as we use etag in our HTTP requests)

CDRussell: Added a constant.

It uses a `equalsIgnoreCare` under the hood

PR #290, opened 2018-06-13
Feature/ui performance

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/705937877094329
Tech Design URL: 
CC: 

**Description**:
Breaks apart the `BrowserTabViewModel`'s view state into smaller view states so that upon each change to an item in the view state, only related UI elements are updated. 

**Steps to test this PR**:
1. _Vague instructions:_ Generally use the browser - kick the tyres and make sure no regressions were introduced in the process
1. _[optionally]_ Verify the Play Store pre-release performance results are positive

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -851,7 +851,10 @@ class BrowserTabFragment : Fragment(), FindListener {
             }
         }
 
-        inline fun renderIfChanged(newViewState: Any, lastSeenViewState: Any?, block: () -> Unit) {
+        /**
+         * This method will execute the given lambda only if the given view states differ
+         */
+        private inline fun renderIfChanged(newViewState: Any, lastSeenViewState: Any?, block: () -> Unit) {
             if (newViewState == lastSeenViewState) {
                 Timber.v("view state identical to last seen state; skipping rendering for ${newViewState.javaClass.simpleName}")
             } else {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -829,5 +703,217 @@ class BrowserTabFragment : Fragment(), FindListener {
         }
     }
 
+    class BrowserTabFragmentRenderer(private val browserTabFragment: BrowserTabFragment) {
```
subsymbolic: Do we need this class? I prefer to avoid circular references unless there is a good reason e.g delegate pattern.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -829,5 +703,217 @@ class BrowserTabFragment : Fragment(), FindListener {
...
+
+        private var lastSeenViewState: ViewState? = null
+
+        fun render(viewState: ViewState) {
```
subsymbolic: Can we remove this global render method? Now that we have individual view models we could have listeners on those that trigger the relevant render methods directly.

subsymbolic:
Tested this and it works great @CDRussell, just a couple of code suggestions.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
CDRussell: Will convert to `inner` class and ditch the explicit parameter

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
CDRussell: This was a big change footprint to make it happen, but it's done. Should be a further improvement on performance; good suggestion!

Issue #33, opened 2017-07-16
Add swipe down 'Refresh' functionality.


abhaymaniyar:
I think the app should have a refresh functionality like Google Chrome Android app. It is very comfortable to refresh webpages that way or there should be a refresh button at the bottom bar.

Please let me know if I should start working on this.

gvklive:
Have you implemented it?

abhaymaniyar:
Yes Sir! Will submit a PR soon.

abhaymaniyar:
@subsymbolic I have done some work on this also. As you mentioned about not accepting PRs at this point of time, should I create a PR for this or not?

subsymbolic:
If you've already done the work then feel free but it will likely languish here. We'll be better placed to accept contributions once this app has progressed past the prototype phase. Stay with us!

abhaymaniyar:
Ok then I am not creating any PR for this. Waiting to pass the prototype phase.

CDRussell:
I've looked at this functionality now and again and have never been happy with the UX when I previously added it. There was always something that just didn't feel right about my previous approaches.

That said, if there was a PR that implemented this well and maintained the existing functionality (such as the toolbar scrolling away as the `WebView` is scrolled) then we'd be open to accepting that PR.

It isn't being treated as a high priority feature internally, so happy to accept community contributions for it if it's desired.

arunsathiya:
Voting for this as well! This would be a nice feature to have.

samiuelson:
Yeah, not having swipe to refresh is so frustrating while browsing.

Usiel:
After playing with it for a bit I think I'm able to do it based on the previous PR by @malmstein.
Would you like me to create a PR? I'm not quite sure since the 'help wanted' label was removed. Could do it today after work.

malmstein:
@Usiel have a go at it if you like! Now that the `BottomNavigation` has been removed it should be easier to execute. Ping me if you need any help!

PR #293, opened 2018-06-18
Swipe to refresh

In2deep73:
Task/Issue URL: #33

**Description**: Add swipe to refresh functionality via an extension of the SwipeRefreshLayout class.  Used a dummy ListView object that re-routes gestures to the WebView to ensure the SwipeRefresh behaves smoothly and consistently

### DIFF
--- a/.idea/copyright/Apache_2_0.xml
+++ b/.idea/copyright/Apache_2_0.xml
@@ -1,6 +1,0 @@
-<component name="CopyrightManager">
-  <copyright>
-    <option name="myName" value="Apache 2.0" />
-    <option name="notice" value="Copyright (c) &amp;#36;today.year DuckDuckGo&#10; &#10;Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);&#10;you may not use this file except in compliance with the License.&#10;You may obtain a copy of the License at&#10;&#10;    http://www.apache.org/licenses/LICENSE-2.0&#10;&#10;Unless required by applicable law or agreed to in writing, software&#10;distributed under the License is distributed on an &quot;AS IS&quot; BASIS,&#10;WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.&#10;See the License for the specific language governing permissions and&#10;limitations under the License." />
-  </copyright>
-</component>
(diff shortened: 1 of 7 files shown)


CDRussell:
Hi @In2deep73, thanks for submitting this. Apologies it's taken me so long to look at it!

I like that the pull to refresh gesture feels pretty natural; that's something that never felt right when I looked at this functionality before.

There are a few areas that need improving however; the way in which the spinner is dismissed seems a bit fragile (e.g., `Thread.sleep)` and lingers longer than I feel it should. We may want to better tie the spinner more to the `WebView` callbacks around when the page is loading.

Additionally there are some merge conflicts which have appeared in the time since this was created, and some of the code doesn't adhere to our style guide.

**Suggestions**
1. Continue to iterate on this PR, OR
1. We can take this PR as a base and finish the remainder of the task in-house

Personally, I'd lean towards 2; I think your contributions around the layout look really good and solve a real problem for us, and it might be quicker for us to finish tidying the other parts of the code. And it means asking less of your time to finish this.

The disadvantage of 1. is that I might not have time regular time to review changes as they are being made here, so iterating on the code might take a while.

Thoughts?

In2deep73:
Hello @CDRussell,

I appreciate you taking the time to review and give feedback on this PR.  I'm glad that this PR solves a problem for you all and I am happy to go with option 2 and have it finished in-house.

I would like to mention that I have found a bug while navigating to AMP enabled sites;  The AMP sites render themselves with the exact dimensions of the WebView and use nested scrolling, such that the "top of page" check always returns true no matter how far down you scroll (resulting in unwanted refreshes).  I have had success with requesting the AMP site as a desktop site and then changing back to a mobile site.  By doing this the site loses it's reference to it's AMP enabled link while still retaining the mobile interface.  One solution for this I have thought of is to listen in the navigation bar for a URL change, and if the URL contains "/amp/" load the desktop version and then change back to the mobile version.  This solution seems a little sloppy to me.  Perhaps something could be changed with the user agent?  I'm sure you guys will find some kind of elegant fix for this, but I just figured I'd mention it to save some time for you all.

If you happen to have any questions about any of my code or why I made certain design decisions please don't hesitate to contact me.

Thank you for letting me part of this process

PR #302, opened 2018-07-03
Fix back/forward browser navigation buttons

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/727172188146397
Tech Design URL: 
CC: 

**Description**:
A regression meant that the view state was checked to see if it was identical to the last seen view state. If it was, rendering would skip. But for these forward/back buttons, they also relied on `WebView `state; not just the view state. So we were erroneously skipping updates for them because the view state was identical.

The fix is to include these in the view state itself, meaning it won't skip the rendering if `canGoBack()` or `canGoForward()` has changed.

`onProgressChanged` is the earliest place where the `WebView` starts reporting the correct new states. But I've included the update in `loadingFinished()` too just in case `onProgressChanged()` could be bypassed somehow.

**Steps to test this PR**:
1. Browse around, and note the back/forward buttons accurately reflect the `WebView`'s ability to navigate forward or back

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -90,7 +90,9 @@ class BrowserTabViewModel(
         val showFireButton: Boolean = true,
         val showMenuButton: Boolean = true,
         val canSharePage: Boolean = false,
-        val canAddBookmarks: Boolean = false
+        val canAddBookmarks: Boolean = false,
+        val canGoBack: Boolean = false,
+        val canGoForward: Boolean = false
     )
 
     data class OmnibarViewState(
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -244,11 +246,12 @@ class BrowserTabViewModel(
         autoCompleteViewState.value = AutoCompleteViewState(false)
     }
 
-    override fun progressChanged(newProgress: Int) {
+    override fun progressChanged(newProgress: Int, canGoBack: Boolean, canGoForward: Boolean) {
         Timber.v("Loading in progress $newProgress")
 
         val progress = currentLoadingViewState()
         loadingViewState.value = progress.copy(progress = newProgress)
+        browserViewState.value = currentBrowserViewState().copy(canGoBack = canGoBack, canGoForward = canGoForward)
```
subsymbolic: Optional: It feels like the browserViewState and loadingViewState should be merged as they closely coupled and we are updating both anyway.

CDRussell: It's tricky to draw the lines between them, as they do all feel pretty related.

However, I'd leave those two separate because they don't always both get updated. While a change in progress here does result in changes to both, there are other places where a change to the browser state **does not** result in changes to progress. So they are, at times, completely independent.

PR #308, opened 2018-07-20
Fix multiple parameters substitution error

PramodJoshi:
**Description**:
Fresh build on Android Studio 3.0.1 throws **"Multiple substitution specified in non-positional format"** error. Correctly formatted the substitutions in strings.xml, compiles and runs fine now.


PramodJoshi:
Seems like this is not required in the latest version of Android Studio 3.1.3.

subsymbolic:
Thanks for submitting this @PramodJoshi. As it is no longer required, I am now closing it out.

PR #310, opened 2018-07-23
Feature/bug fix webview state too large

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/533891143524678
Tech Design URL: 
CC: 

**Description**:
Attempting to save the `WebView`'s state using the usual Android state saving mechanism can fail with a `TransactionTooLargeException`. This will happen when the underlying `Bundle` exceeds a certain size. The size isn't well documented but can be anecdotally anywhere between 500KB and 1MB. This type of exception is dominating our crash reports.

This PR stops using the normal Android state saving mechanism, and reverts to an in-memory LRU cache instead. We have chosen 10MiB max cache size (after which `WebView` history for least recently used tabs will start becoming inaccessible).

**Steps to test this PR**:

_Testing - happy path_
1. Browse to a page, 
1. Browse to another. 
1. Hit the `HOME` button once loaded
1. Return to the app. Ensure the previous page is reloaded and the rough scroll position is restored.
1. Hit the `BACK` button. Verify the previous page is loaded (and that you don't just land on the new tab screen)
1. Add in multiple tabs and repeat the above - ensuring things work as expected while there is enough room in the cache.

_Testing - less happy path_
1. You might want to manually lower the cache size to test this PR out. See `WebViewSessionInMemoryStorage.CACHE_SIZE` for this value. Recommend setting it to around 5KB for testing (empirically, each visited site might add around 1KB to the size)
1. Additionally, you might want to get Android to throw out all `Activities` as soon you leave them from developer settings to force the state saving to kick in.
1. Monitor `logcat` for output given by `WebViewSessionInMemoryStorage` class - this will tell you what the size of the cache is and let you see how close you are to exhausting it.
1. Repeat the steps from the happy path, above. This time however, deliberately exhaust the cache.
1. When you bring the app to the foreground, verify that the previous page is loaded
1. This time, you should notice that you **can't** return to your previous pages in history. They are gone; we ran out of room in the cache to store the history. So we'll reload their previous page but they'll lose their browsing history.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/session/WebViewSessionStorage.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/session/WebViewSessionStorage.kt
@@ -35,7 +35,7 @@ class WebViewSessionInMemoryStorage : WebViewSessionStorage {
     private val map = object: LruCache<String, Bundle>(CACHE_SIZE) {
 
         /**
-         * We can calculate this however we choose, but it should match up with the value we used for cache size.
+         * We can calculate this however we choose, but it should match up with the value we use for cache size.
          * i.e., if we specify max cache size in bytes, we should calculate an approximate size of the cache entry in bytes.
          */
         override fun sizeOf(key: String, bundle: Bundle) = bundle.sizeInBytes()
--- a/app/src/main/java/com/duckduckgo/app/browser/session/WebViewSessionStorage.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/session/WebViewSessionStorage.kt
@@ -35,8 +35,8 @@ class WebViewSessionInMemoryStorage : WebViewSessionStorage {
     private val cache = object : LruCache<String, Bundle>(CACHE_SIZE_BYTES) {
 
         /**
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/duckduckgo/app/browser/session/WebViewSessionStorage.kt
```diff
@@ -35,8 +35,8 @@ class WebViewSessionInMemoryStorage : WebViewSessionStorage {
     private val cache = object : LruCache<String, Bundle>(CACHE_SIZE_BYTES) {
 
         /**
-         * We can calculate this however we choose, but it should match up with the value we use for cache size.
-         * i.e., we specify the max cache size in bytes, so we need to calculate an approximate size of the cache entry in bytes.
+         * Size (in bytes) of a single entry in the cache for the given key.
+         * We specify the max cache size in bytes, so we need to calculate an approximate size of the cache entry in bytes.
```
subsymbolic: Nice!

CDRussell:
>Do you want to hold off merging until @brindy has given it a test too?
Yup, sounds good.

Issue #288, opened 2018-06-08
Option to open external links in the appropriate app


SuperDuperUser:
Right now, if I search "Uber Google Play" in the DDG app and I click the link to the Play Store, it opens in the internal browser instead of opening in the Google Play app, where I would ideally like to have it open so I can install. Moreover, if I click the "Open in Play Store App" button that the site displays, it causes a net::ERR_UNKNOWN_URL_SCHEME error. Links to YouTube and Amazon also don't make its way over to their respective apps. It would be really handy if there was a setting that allowed links to open in their default apps, either on a page-by-page (clicking something in the drop-down menu) or general basis. 

There were a couple of issues about this that were closed already, but those took place before the transition to the "Privacy Browser" version of the app in January 2018 and have not been adressed since. Also, the old version of the app used to have an option that kind of worked, but since the revamp it's no longer there https://imgur.com/gNw1tyD

Version 5.6.1
LineageOS 15.1, Android 8.1
Motorola Nexus 6

CDRussell:
This has been partly implemented in #322 

The app will now launch the external app if it is installed for urls with custom schemes, such as the "Open in Play Store App" case highlighted.

It doesn't yet offer a way to open an equivalent native app from an HTTP/HTTPS page when the app which could accept that deep link is installed. ie, this hasn't been implemented yet 👇.

>Right now, if I search "Uber Google Play" in the DDG app and I click the link to the Play Store, it opens in the internal browser instead of opening in the Google Play app

gostaDrivekitt:
This feature is needed for apps using Webviews and deep links/custom scheme links to go back to the app.
Would be awesome to have so that the DuckDuckGo browser work seamlessly as Chrome on Android.

Poopooracoocoo:
I love how Firefox Focus handles links. Links to external apps don't open automatically and prompt me. I can open a link in the current tab (or for a link using the menu) in another browser or app by going to the menu.

I love browsing the web in Focus and then opening the ones I want in my browser. With Reddit I can open them in the app or not. I usually open Reddit links in the app to save them though. If I plan on commenting, I'll do it through the browser.

uraza:
I am facing this issue with Youtube links on my side: they keep opening in the browser and I cannot figure out how to work around it.

aitorvs:
Thanks for reporting this @uraza, indeed this bit is still to be done. I have created an internal task so that we can triage it.

https://app.asana.com/0/1198194956794324/1199533149091655/f

[...]

PR #322, opened 2018-08-07
Feature/custom url schemes

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/656818431539163
GitHub Issue: https://github.com/duckduckgo/Android/issues/288

Tech Design URL: 
CC: 

**Description**:
Detect custom schemes and determine if an external app can open it or not. If an app is installed which can handle it, we'll prompt the user to ensure they are happy with jumping outside of our app before launching an external app.

**Steps to test this PR**:
1. Search for "uber google play" 
1. Visit the search result for the Google Play Store
1. Click the floating button for "OPEN IN PLAY STORE APP"
1. Test the positive and negative buttons for the dialog which is shown

device-2018-08-07-144630

---
###### Internal references:


In2deep73:
Hello @CDRussell ,

I was working on something similar to this for my own personal use.  I hope I am not overstepping my bounds by critiquing this PR, but intent URLs such as:
intent://user?screen_name=DuckDuckGo#Intent;scheme=twitter;S.browser_fallback_url=http%3A%2F%2Ftwitter.com%2FDuckDuckGo;end
specify a fallback URL to navigate to if the app is not installed (the intended action is to navigate to DDG's twitter via the app if installed, else navigate to fallback URL).  Perhaps it would be nice if the fallback URL [val fallbackURL = intent.getStringExtra("browser_fallback_url")] was used when an ActivityNotFoundException is thrown or when the user explicitly chooses to remain in the DDG app.  Other than that, the only other suggestions I could make based on my own experience with this issue is to do something like val package = intent.getPackage() if no app is installed to handle a certain URL and redirect to the market:// URL of said app if intent.getPackage() is not null instead of giving a toast.  Also, maybe a function to determine whether a URL is a 'deep link' to an activity would be nice (deep links to apps use the https or http scheme which make them a bit trickier).  Also, URL's such as the one listed at the bottom of this message redirect to the play store and double load even if the app is installed.  The solution that I've found for this is to pass the entire URL into an intent [and possibly do webView.stopLoading() before launching the intent) and let the OS handle it (this would require a reliable method of determining whether a link is a deep link or if a specialized handler is available).  I hope some of this is helpful to you.  Thank you!

https://reddit.app.link/?campaign=scaled_inference&utm_name=scaled_inference&channel=xpromo&utm_source=xpromo&feature=mweb&utm_medium=mweb&%24og_redirect=https%3A%2F%2Fwww.reddit.com%2F&%24deeplink_path=%2F&%24android_deeplink_path=reddit%2F&mweb_loid=00000000001wfkrzu5&mweb_loid_created=1533163548549&mweb_user_id36=&mweb_user_name=&domain=www.reddit.com&geoip_country=US&user_agent=Mozilla%2F5.0%20(Linux%3B%20Android%209%3B%20Android%20SDK%20built%20for%20x86%20Build%2FPPP4.180612.007)%20AppleWebKit%2F537.36%20(KHTML%2C%20like%20Gecko)%20Chrome%2F67.0.3396.87%20Mobile%20Safari%2F537.36&base_url=%2F&referrer_domain=www.google.com&referrer_url=https%3A%2F%2Fwww.google.com%2F&language=&dnt=false&compact_view=true&adblock=false&session_id=s0EI10481bFKvLfzR7&loid=00000000001wfkrzu5&loid_created=1533163548549&reddaid=4OU25RNO52GONJAB&keyword=control_1&utm_term=control_1&tags=topbutton&utm_content=topbutton

brindy:
@CDRussell 

When the URL is an intent the open in external app functionality leaves the page with the intent URL showing and blank.  I’ve not been able to test this on a device yet, so not sure if it’s because it fails due to the play store not being installed in the emulator, but it also happens when you click cancel. Maybe we can "go back" when these links are clicked?

CDRussell:
@In2deep73 thanks for the detailed suggestions

>I hope I am not overstepping my bounds by critiquing this PR

Not at all! That's why we do these PRs in the open; there's always going to be things we miss. Some of what you said was deliberated omitted as I had limited time to tackle this task; and adding more into its scope would likely have caused the whole task to get delayed. But others you've raised is definitely valid.

>fallback URL to navigate to if the app is not installed

This is definitely something we should add in. I'll get on that.

>if no app is installed to handle a certain URL and redirect to the market:// URL

That might be something to consider for later, but needs a little consideration. Not all of our users get the app from the Play Store, and indeed some definitely don't want to get anything from it.

>maybe a function to determine whether a URL is a 'deep link' to an activity would be nice (deep links to apps use the https or http scheme which make them a bit trickier)

Agreed; I almost considered doing this as part of this task but realised it was extend the scope beyond the time I had available for it. Decided it would be best to come back to this later so it doesn't hold up this custom URL functionality.

>URL's such as the one below [for yelp] result in an **infinite loop** of asking which external app to use

Thanks for flagging this; I'll investigate that.

> I just wanted to try and save some time for you all but at the same time I do not want to disrupt your workflow.

Your efforts are appreciated! Thank you for raising all of the issues you have.

CDRussell:
@brindy 

>When the URL is an intent the open in external app functionality leaves the page with the intent URL showing and blank. I’ve not been able to test this on a device yet, so not sure if it’s because it fails due to the play store not being installed in the emulator, but it also happens when you click cancel. Maybe we can "go back" when these links are clicked?

Indeed, the case is likely because you don't have the Play Store on the emulator (there are emulators you can use which _do_ have the Play Store btw).

As for the cancel case, I did look at stepping the `webView` back but it can get a weird depending on how the custom scheme arrived. Some sites tend to do redirects to different `https` pages before eventually ending on the `intent://` or whatever. I thought it would risk breaking more by trying to be smart around the back navigation.

CDRussell:
>However, this isn't showing the fallback URL for me on an emulator without Play Store installed.

I wouldn't worry about this case; in this situation, Google Play Services is the match but it basically swallows the request to open the Play Store when the Play Store isn't installed. 

>I wonder if we should just show the fallback URL anyway if it's available as well as prompt to launch externally?

My worry here is what happens if the fallback URL, upon loading, also redirects to launch the app or Play Store?

> when I try it with Play Store installed when I hit back from the Play Store I get the prompt again

That's not good. How did you see this? Do you have exact steps to reproduce as I'm not seeing this behaviour?

brindy:
>> when I try it with Play Store installed when I hit back from the Play Store I get the prompt again

> That's not good. How did you see this? Do you have exact steps to reproduce as I'm not seeing this behaviour?

* One Plus One device (Android 5.1.1) (haven't checked it on others)
* Search for "uber play store"
* Open Uber on the Play Store link
* Tap "open in Play Store"
* Play store opens
* Immediately go back
* DDG opens and immediately shows the prompt again

CDRussell:
You got activities being killed off by any chance?

brindy:
> You got activities being killed off by any chance?

Yep! that was it

CDRussell:
Merging this despite there being some rough edges that can be tidied up later.

Rough edges being: 

- we might be able to tidy up the web view when launching external app so it isn't left in a blank-looking state
- links like that reddit one will try and launch two apps; the play store and the reddit app

PR #331, opened 2018-08-22
open in background tab

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/770186107328907
Tech Design URL: 
CC: 

**Description**:

Open tab in new background.

Note, waiting on final animation, but that should just be a change to the animation JSON.

**Steps to test this PR**:

1. Long press a link and select "open in background tab"

  * tab switcher button animates
  * number of tabs increments
  * tab switcher button shows correct number of tabs and unread tab dot

2. Open tab switcher

  * current tab has blue border
  * other tabs are slightly transparent
  * unread tabs have a blue dot
  * tab title might be the domain name with `www.` prefix stripped if the tab hasn't had chance to load yet

3. Select tab

 * Tab opens
 * Number of tabs remains the same
 * If there are no unread tabs the dot is removed
 * Open tab switcher -> selected tab now has blue border and others are slightly transparent

4. Confirm that open in background tab pixel has been fired (`mlp_b`)

5. Remove tabs from the tab switcher

  * Selected tab should update correctly
  * Tab switcher button should be updated  

6. Test (database) migration from a previous app version 

---
###### Internal references:


CDRussell:
screen shot 2018-08-23 at 13 46 34

The size of the tab icon seems short compared to the fire icon. Is this known/deliberate?

CDRussell:
The blue dot feels like it hangs around too long. I feel like it should disappear when i visit the tab switcher view; i don't necessarily want it hanging around until i've visited every tab but more as a reminder there's new tabs been added recently.

That might just be me, though. maybe worth considering if the current behaviour is right, but not a blocker.

CDRussell:
When I upgrade the app from a previous version, my already-open tabs get marked as me having not read them, even if I have.

I'm sure users will figure it out soon enough, but if the first time they see the dot is also when we confuse them about it, they may not like it.

app/src/main/java/com/duckduckgo/app/tabs/db/TabsDao.kt
```diff
@@ -81,4 +84,16 @@ abstract class TabsDao {
...
+
+    @Transaction
+    open fun insertTabAtPosition(tab: TabEntity) {
+        incrementPositionStartingAt(tab.position)
+        insertTab(tab)
+    }
+
+    fun lastTab(): TabEntity? {
+        val tabs = tabs()
+        return if (tabs.isEmpty()) null else tabs.last()
```
CDRussell: Could replace with `tabs().lastOrNull()` ?

chrismorast:
@CDRussell and @brindy , The fire icon has been updated in the new set to be slightly smaller, better matching the rest of the icons. All of these are available in Zeplin

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -54,13 +58,26 @@ class TabDataRepository @Inject constructor(private val tabsDao: TabsDao, privat
...
+            tabsDao.addAndSelectTab(TabEntity(tabId, data.value?.url, data.value?.title, true, position))
+        }
+    }
+
+    override fun addNewTabAfterExistingTab(url: String?, tabId: String) {
+        Schedulers.io().scheduleDirect {
+            val position = tabsDao.tab(tabId)?.position ?: -1
+            val uri = Uri.parse(url)
+            val host = uri.host
+            val title = host.dropPrefix("www.")
```
CDRussell: looks like `host` could be null here

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -97,3 +114,7 @@ class TabDataRepository @Inject constructor(private val tabsDao: TabsDao, privat
         }
     }
 }
+
+fun String.dropPrefix(prefix: String): String {
```
CDRussell: this already exists as a string extension function: `removePrefix`

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
```diff
@@ -84,6 +94,18 @@ class TabSwitcherAdapter(private val context: Context, private val itemClickList
...
+        val close: ImageView,
+        val tabUnread: View
     ) : RecyclerView.ViewHolder(root)
+
+    companion object {
+
+        @DrawableRes val backgroundSelected = R.drawable.tab_background_selected
+        @DrawableRes val backgroundDefault = R.drawable.tab_background
+
+        val alphaSelected = 1.0f
```
CDRussell: should the 4 variables in the companion object be styled as static constants? 

eg: `private const val ALPHA_SELECTED = 1.0f`

also, looks like this file needs its code formatted.

app/src/main/res/layout/view_tab_switcher_button.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
```
CDRussell: format?

CDRussell:
>Is there any way to defer the blue dots on the already open tabs and only generate once the tab open event is triggered from the new version?

@chrismorast I believe @brindy has fixed this issue.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -54,13 +58,26 @@ class TabDataRepository @Inject constructor(private val tabsDao: TabsDao, privat
...
+            tabsDao.addAndSelectTab(TabEntity(tabId, data.value?.url, data.value?.title, true, position))
+        }
+    }
+
+    override fun addNewTabAfterExistingTab(url: String?, tabId: String) {
+        Schedulers.io().scheduleDirect {
+            val position = tabsDao.tab(tabId)?.position ?: -1
+            val uri = Uri.parse(url)
+            val host = uri.host
+            val title = if (host == null) url else host.removePrefix("www.")
```
CDRussell: could replace with `val title = host?.removePrefix("www.") ?: url`

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -19,10 +19,12 @@ package com.duckduckgo.app.tabs.model
 
 import android.arch.lifecycle.LiveData
 import android.arch.lifecycle.MutableLiveData
+import android.net.Uri
 import com.duckduckgo.app.global.model.Site
 import com.duckduckgo.app.global.model.SiteFactory
 import com.duckduckgo.app.tabs.db.TabsDao
 import io.reactivex.schedulers.Schedulers
+import timber.log.Timber
```
CDRussell: This import isn't needed

PR #355, opened 2018-09-21
Add light widget

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/825447419911129
Tech Design URL: N/A

**Description**:
Adds a light widget to the app

**Steps to test this PR**:

**TEST 1: Upgrade path**
1. Ensure that the widget is already installed
1. Install the new version of the app
1. Ensure the original widget works

**TEST 2: General**
1. Add the light widget to your home screen
1. Tap it and ensure it launches the app

---
###### Internal references:


app/src/main/java/com/duckduckgo/widget/SearchWidget.kt
```diff
@@ -24,10 +24,10 @@ import android.widget.RemoteViews
 import com.duckduckgo.app.browser.BrowserActivity
 import com.duckduckgo.app.browser.R
 
-/**
- * Implementation of App Widget functionality.
- */
-class SearchWidget : AppWidgetProvider() {
+
+class SearchWidgetLight : SearchWidget(R.layout.search_widget_light)
```
CDRussell: I like this approach; should we do it for both light and dark?

    class SearchWidgetLight : SearchWidget(R.layout.search_widget_light)
    class SearchWidgetDark : SearchWidget(R.layout.search_widget)

And maybe we should rename `R.layout.search_widget` to have `_dark` on the end for better clarity/symmetry?

subsymbolic: I would prefer that too (and initially had that exact setup) except if we change the class name we break all the existing widgets that name.

So I've gone with concept of "dark" as the "default" and kept the resources the same to keep things consistent.

Issue #347, opened 2018-09-06
Using "site" query parameter results in "Unable to open this type of link"


GuiltyDolphin:
### App version

5.9.1 (50901) (via F-Droid)

### Device make and model

OnePlus 5

### Android version

8.1.0

### Steps to reproduce the bug

1. Into the address bar type `foo site:duckduckgo.com` (or any query + any site)

2. Press enter (try to load page)

### Expected behavior

A DuckDuckGo search opens with the query

### Actual behavior

A toast pops up briefly with the phrase "Unable to open this type of link"

dineshkv:
I believe this is a feature request.

GuiltyDolphin:
@dineshkv I think this is a bug (i.e., something broke) - I noticed it because I used to do it and it worked (it definitely works in version 5.7.5), but now no longer works

dineshkv:
Oh okay. If I want to work on this are PRs going to be accepted?

CDRussell:
Thanks for reporting the `site:` problem; that was an unintended regression of another recent change.

Happy to accept a PR for this if someone wants to pick it up. And thanks in advance, if so!

CDRussell:
The problem is likely to be found in `SpecialUrlDetector`. We recently switched things around to better handle custom app schemes, and in the process broke queries like this that look like a URL scheme but should just be treated as a normal search query.

https://github.com/duckduckgo/Android/blob/develop/app/src/main/java/com/duckduckgo/app/browser/SpecialUrlDetector.kt#L42

In2deep73:
I have some fixes to propose if @dineshkv decides not to work on this.

dineshkv:
@In2deep73 please go ahead as I have not started working on it yet.

CDRussell:
Additional test case to cover under this fix: searches like `std::fabs`

subsymbolic:
Thanks to everyone who has submitted their PRs and ideas here, it has been really valuable. Taking all the ideas into consideration I would like to move forward with separating the logic for items that have an authority from those that do not when detecting the type in our `SpecialUrlDetector`. I imagine splitting the logic into two categories as follows:

**Category 1: Items with a Uri.authority component. This covers web links and intent URIs**
The behavior for these would be unchanged. Examples:
- http://duckduckgo.com - Should continue to be classed as `UrlType.WebUrl`
- twitter://user?14504859 - Should continue to be classed as  `UrlType.IntentType` or `UrlType.Unknown` if twitter is not installed
[...]

PR #356, opened 2018-09-23
Fix issue 347 by adding plausibility check to URI schemes

JeffAment:
Task/Issue URL: #347 

**Problem:** Anything before the first instance of a ‘:’ is interpreted as a scheme, even if it is not plausible for it to be a valid scheme.  For example, for “foo site:duckduckgo.com”, the scheme is thought to be “foo site”; for “:wq”, the scheme is thought to be “”.  Hence, both examples are erroneously treated as intent links.

**Solution:** I thought this problem was a good candidate for a regular expression.  Before trying to process the uri as an intent, the “scheme” is now checked against a regex pattern to see if its validity is plausible.  If it is not plausible, a search query is launched.  This brings the behavior in line with how I would expect the browser to handle various schemes.

**Thoughts:** It bothers me that there are two lines in the when statement that both launch a search query.  Is there possibly a better way to write this?  I am a newcomer to Kotlin and I have not investigated this much as I would like to hear some thoughts on this approach before I delve deeper into this.

**Steps to test this PR:** 
**1.** Test various invalid schemes that do not conform to the regex pattern and verify that they result in a search query.
**2.** Test intent links and other valid/invalid schemes that conform to the regex pattern and verify that they behave as expected.

---
###### Internal references:


CDRussell:
Hey, thanks for getting the ball rolling on this.

I'm not sure if a regex is the best way here; it's hard to predict what a _valid_ scheme is. For instance, running your branch, if i enter `site:example.com foo` into the URL bar, it still returns that it can't handle this type of link. Really, here, it should perform a search just like if you enter `foo example.com`.

I wonder if instead, what we need is to check if any other apps could handle it inside `SpecialUrlDetector.buildIntent` function. And if no other app could handle it, treat it like a query.

     private fun buildIntent(uriString: String): UrlType {
        try {
            val intent = Intent.parseUri(uriString, 0)
            val fallbackUrl = intent.getStringExtra(EXTRA_FALLBACK_URL)

            val pm = context.packageManager
            val activities = pm.queryIntentActivities(intent, 0)

            Timber.i("Found ${activities.size} that could consume $uriString")
            return if(activities.size == 0){
                UrlType.SearchQuery(uriString)
            } else {
                UrlType.IntentType(url = uriString, intent = intent, fallbackUrl = fallbackUrl)
            }
        } catch (e: URISyntaxException) {
            Timber.w(e, "Failed to parse uri $uriString")
            return UrlType.Unknown(uriString)
        }
    }

Something like the above? I don't have time to test out the various cases right now to see if that breaks other things.

If you have time, you could experiment with this approach maybe? Otherwise, happy to see revisions on the regex approach which accounts for the case i described above.

And again, thanks for getting started on this!

In2deep73:
Hello @CDRussell,

Thank you for your response and suggestions.
> Hey, thanks for getting the ball rolling on this.

Not a problem; I enjoy doing these kinds of things, and being able to work on a piece of software that I actually use alongside the developers is, for lack of a better term, “wicked awesome.” 
> Really, here, it should perform a search just like if you enter foo site:example.com.

Yes!  I did not consider this.

>I wonder if instead, what we need is to check if any other apps could handle it inside SpecialUrlDetector.buildIntent function. And if no other app could handle it, treat it like a query.

I had originally thought about doing something like this, but upon more careful consideration I thought the regex method might deliver a more stable user experience.  I’ll illustrate this with a thought-provoking magnification of a semi-real(ish) problem: I don’t like how if I am, for example, on www.SuperImportantWebsite.eu.nl.co.uk and click on a valid intent link for which no specialized handler is available, that the browser would navigate away from the super important page and redirect me to a search engine.  I personally think it’s best in this case to err on the side of staying on the current page, and I feel like the end user should have a somewhat reasonable expectation of this behavior as well [when clicking on certain links – especially in the case of a hypothetical http link that may redirect to an intent link; this would be especially frustrating for the user].  Staying on the page and giving a toast also gives the user a better idea of what’s going on; I feel most users wouldn’t understand that they need to install an app to open this kind of link and would be very confused to see a search query launched.  Perhaps some kind of dialog box asking the user if they would like to launch a search query would alleviate these concerns somewhat, but it would still not be a behavior that I, as a user, would expect.

I propose that we keep the regex (see new commit), but I am all about that hashtag respect and would like to hear your thoughts on this as I trust your judgment over my own and would only want to keep the regex if you agreed it was preferable in light of the concerns I’ve raised.  I do like the idea of trying to launch a search query if no handler is available, but I feel like we might need an entire IntentURIUtilities.kt class with some more complex logic to implement this well.  If you disagree with my thought process, I would be happy to implement the packageManager fix you suggested and test it for you.  The concerns I raised about it aren’t really as bad as my paradoxical illustrative rant made them out to be, and I do understand that to some extent the problems I referenced will always exist in some capacity.  I just feel like the regex would be a slightly safer option for the time being.

I may have over-complicated things in the new commit, but I thought it was more interesting to implement it the way I did as opposed to hard-coding the “site” string into the regex, and it’s nice to have a separate blacklist ArrayList of strings to make future additions/subtractions/testing simpler.  I’m really not sure if this is the best idea, however, and would like to hear your input on this.

Thank you for taking the time to read this, sir!

ngsilverman:
Weighing in as an outsider here. @In2deep73: The current approach seems to assume that `site` is the only non-scheme keyword that might have a `:` after it. I frequently perform searches for `define:<word>` out of habit, and I imagine many people might arbitrarily use colons in what they intend to be searches.

**Re: what about valid intent links with no intent handler available on device?** This seems like an exceedingly rare case. Why would a website use intent links if they're not confident that users will be able to open them? My impression is that the preferred way of linking into apps it to use an HTTP scheme with a URL that their app declares it can handle. Android will then prompt the user to choose which app to use to open the link.

It might also be helpful to handle links on websites and the Search/URL bar differently. For both UX and privacy reasons, I don't think links from websites should ever result in a DuckDuckGo search unless they are specifically linking to a DuckDuckGo search, ie. something like `https://duckduckgo.com/?q=hello`.

I like @CDRussell's suggestion to check whether the query can be handled as an intent. It might be helpful to test how other browsers handle things as well, user expectations are shaped by what they've experienced elsewhere and we may discover some helpful solutions.

In2deep73:
Hello @ngsilverman,

Thank you for taking the time to give your feedback; I really appreciate it!

> I frequently perform searches for define:<word> out of habit, and I imagine many people might arbitrarily use colons in what they intend to be searches.

I agree with this as I also use `define:<word>` to search for things.

> This seems like an exceedingly rare case. Why would a website use intent links if they're not confident that users will be able to open them?

It is true that these kinds of links are becoming rarer, but I have still managed to come across some of these in my day to day browsing.

> It might also be helpful to handle links on websites and the Search/URL bar differently.

It seems as if clicked links and those manually entered into the search bar _are_ actually treated differently.  This was a misunderstanding on my part.

> It might be helpful to test how other browsers handle things as well

The thing here is that every browser seems to handle this differently, and I suspect many have more complex logic to process these kinds of links.  My original approach drew lots of inspiration from Firefox.  Firefox feels very much like it has something similar going on under the hood.  For example, "site:", "define:" and other schemes that conform to the regex will always be treated as valid schemes and will result in a redirection to an error page.  I have tested @CDRussell's approach and will elaborate on this in my message to him below so I don't repeat myself.

Thank you again for raising these issues!

In2deep73:
Hello @CDRussell,

I have implemented the package manager approach on my fork (https://github.com/In2deep73/Android/tree/issue347_package_manager_approach).  The only thing I did differently than you suggested was to change `return if(activities.size == 0)` to `return if(activities.size == 0 && fallbackUrl == null)` so that the link is not treated as a search query if a `fallbackUrl` exists.

This seems to work well, but it behaves a little differently than I had anticipated.  When I, for example, go to reddit.com (amongst others) and click on the "open in app" button (without the reddit app installed), I am redirected to a "Webpage not available" page that says "...could not be loaded because: net::ERR_UNKNOWN_URL_SCHEME".  If I manually type the address into the navigation bar, only then is it treated as a search query.  I am happy to see it behave this way, as I thought it would be better for the user to see an error with hints on what they need to do to resolve the issue vs launching an unwanted search query.

I have no strong preferences on which approach we should use.  I feel we have outlined the pros and cons of each approach adequately and will wait for you to make the final judgment on this.  If you decide to go with the regex method, I definitely think we should add "define" to the blacklist as @ngsilverman suggested.

JeffAment:
Closing this PR in light of comments posted in issue 347.

subsymbolic:
Thanks for submitting this and for all the brainstorming and exploration. This is the top reference in our internal task.

PR #361, opened 2018-09-26
Feature/send fire pixel on app launch

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/838465040462722
Tech Design URL: 
CC: 

**Description**:
Previously, there were ways in which the pixels for performing a 🔥could have been lost. This might have happened if during the fire animation, the user killed the app. When they next relaunched the app, we lost information about the 🔥. Their data would still have been safely deleted, as expected, but the pixel would have been lost.

**Steps to test this PR**:
1. From a fresh install, ensure the app launch pixel occurs.
1. Perform a 🔥 and allow the process to restart without intervention. Verify that 
    1. the fire pixel is sent
    1. the app launch pixel is **not sent**
1. Perform a 🔥, this time hit the home button during the fire animation. Wait 15s before relaunching the app from recents
    1. Verify the app doesn't re-show itself automatically
    1. Verify the fire pixel is sent
    1. Verify the app launch pixel **is** sent
1. Perform a 🔥, this time hit the home button during the fire animation. Kill the app entirely. Wait 15s before launching from the normal launcher icon
    1. Verify the app doesn't re-show itself automatically
    1. Verify the fire pixel is sent
    1. Verify the app launch pixel **is** sent

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/fire/DataClearingStore.kt
```diff
@@ -0,0 +1,72 @@
...
+package com.duckduckgo.app.fire
+
+import android.content.Context
+import android.content.SharedPreferences
+import android.support.annotation.VisibleForTesting
+import androidx.core.content.edit
+import javax.inject.Inject
+
+
+interface DataClearingStore {
```
brindy: This seems pretty generic, but is only used by the 🔥 pixel.  If someone were to come along and think they can reuse this for their pixels without knowing the history they would introduce some bugs.

A couple of potential options:
1) This should probably be clearer about its intent `incrementUnsentFirePixelCount` (which doesn't seem very nice, but is probably the simplest)

2) Make it more generic in its interface by taking the name of the pixel to count (but this is about to get extra complicated since some pixels can now take parameters)

CDRussell: What about if we rename the interface to `DataClearingUnsentPixelStore`? Or `UnsentDataClearingPixelStore`?

brindy: I'm less worried about the interface name, though having `Pixel` in there would probably be helpful.  It really depends if you want to keep this thing specific to the "Forget All" functionality or keep it named generically, but make it function generically as well.

I would probably go with `UnsentForgetAllPixelStore` as the interface name and leave the function names as is, rather than implementing generic functionality which YAGNI :)

app/src/main/java/com/duckduckgo/app/fire/UnsentForgetAllPixelStore.kt
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    private val preferences: SharedPreferences
+        get() = context.getSharedPreferences(FILENAME, Context.MODE_PRIVATE)
+
+    companion object {
+
+        @VisibleForTesting
+        const val FILENAME = "com.duckduckgo.app.fire.settings"
```
brindy: (Optional) Sorry - should have spotted this before.  Could this filename be better as well?

CDRussell: Good idea - went with `com.duckduckgo.app.fire.unsentpixels.settings`

PR #364, opened 2018-09-27
Add https upgrades summary pixel

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/839627367764303
Tech Design URL: N/A however spec was discussed in https://app.asana.com/0/361428290920652/832602107339150

**Description**:
- Add a summary `ehs` pixel to determine https failures versus successes
- Remove status error from `ehd` pixel. Status errors are generally connectivity (we never see them in kibana as the `ehd` pixel fails too) but they could skew summary counts
- Restrict errors to main frame

**Steps to test this PR**:
1. Go to a broken https upgrade site ensure that the failure `ehd` pixel is still fired.
It may be easier to remove the `httpsUpgrader.shouldUpgrade` check in the `isHttpsUpgradeSite` and visit https://badssl.com and ensure the pixel fires from there.
1. Navigate to a few sites successfully
1. Stop and relaunch the app ensure that the `ehs` summary pixel is fired on sync showing the count of failed versus total https sites in our upgrade list
1. Ensure that other app pixels continue to fire as normal

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeDataDownloader.kt
+++ b/app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeDataDownloader.kt
@@ -29,7 +29,7 @@ import com.duckduckgo.app.statistics.pixels.Pixel.PixelParameter.FAILURE_COUNT
 import com.duckduckgo.app.statistics.pixels.Pixel.PixelParameter.TOTAL_COUNT
 import com.duckduckgo.app.statistics.store.StatisticsDataStore
 import io.reactivex.Completable
-import io.reactivex.Completable.fromAction
+import io.reactivex.Completable.*
 import timber.log.Timber
 import java.io.IOException
 import javax.inject.Inject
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeDataDownloader.kt
```diff
@@ -107,4 +113,22 @@ class HttpsUpgradeDataDownloader @Inject constructor(
...
+
+            if (statisticsDataStore.httpsUpgradesTotal == 0) {
+                return@fromAction
+            }
+            val params = mapOf(
+                TOTAL_COUNT to statisticsDataStore.httpsUpgradesTotal.toString(),
+                FAILURE_COUNT to statisticsDataStore.httpsUpgradesFailures.toString()
+            )
+
+            pixel.fireCompletable(Pixel.PixelName.HTTPS_UPGRADE_SITE_SUMMARY, params).blockingGet()
```
CDRussell: the `blockingGet()` is a bit of a smell here. We shouldn't need to subscribe inside a stream like this.

Perhaps instead, we should wrap this in `Completable.defer` instead of `Completable.fromAction` as we're currently doing. That would then become 👇 

        fun reportUpgradeStatistics(): Completable {
        return Completable.defer {
            if (statisticsDataStore.httpsUpgradesTotal == 0) {
                return@defer complete()
            }
            val params = mapOf(
                TOTAL_COUNT to statisticsDataStore.httpsUpgradesTotal.toString(),
                FAILURE_COUNT to statisticsDataStore.httpsUpgradesFailures.toString()
            )

            return@defer pixel.fireCompletable(Pixel.PixelName.HTTPS_UPGRADE_SITE_SUMMARY, params).andThen {
                Timber.v("Sent https statistics")
                statisticsDataStore.httpsUpgradesTotal = 0
                statisticsDataStore.httpsUpgradesFailures = 0
            }
        }
    }

Note, the use of `andThen` to hook in the action to be done after it was successfully upload. 

Test it out, of course, to make sure I haven't screwed anything up! 😄 But I think that's _cleaner_ in Rx world.

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeDataDownloader.kt
```diff
@@ -107,4 +113,22 @@ class HttpsUpgradeDataDownloader @Inject constructor(
 
     }
 
+    fun reportUpgradeStatistics(): Completable {
```
CDRussell: We should consider renaming this class at some point, since it is no longer exclusively a `downloader` as the class name suggests - or move the `reportUpgradeStatistics` action to its own class?

app/src/main/java/com/duckduckgo/app/httpsupgrade/api/HttpsUpgradeDataDownloader.kt
subsymbolic: Great idea!

PR #379, opened 2018-10-15
Reduce sync frequency and remove https path from EHD pixel

subsymbolic:
Task/Issue URL:
https://app.asana.com/0/361428290920652/863340012156759
https://app.asana.com/0/361428290920652/863340012155762

**Description**:
* Removes the url path from EHD pixels.
* Increases app sync to 12 hours and failure backoff to 1 hour to reduce data usage and load on the backend
* Removes the simpleUrl extension method based on previous team conversations that it is not needed

**Steps to test this PR**:
**TEST PIXEL**
1. Turn on charles and watch app traffic
1. Visit https://badssl.com/abcd
1. Note the EHD pixel fires as https://badssl.com (without a path)

**CHECK SYNC LOGIC**
Check the code to ensure sync time has been updated to 12 hours.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -179,9 +179,10 @@ class BrowserWebViewClient @Inject constructor(
     }
 
     private fun reportHttpsUpgradeSiteError(url: Uri, error: String?) {
+        val host = url.host ?: return
         val params = mapOf(
             APP_VERSION to BuildConfig.VERSION_NAME,
-            URL to "${url.scheme}://${url.host}",
+            URL to "https://$host",
             ERROR_CODE to error
         )
         pixel.fire(HTTPS_UPGRADE_SITE_ERROR, params)
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -27,7 +27,6 @@ import android.support.annotation.WorkerThread
 import android.webkit.*
 import androidx.core.net.toUri
 import com.duckduckgo.app.global.isHttps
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -182,7 +181,7 @@ class BrowserWebViewClient @Inject constructor(
     private fun reportHttpsUpgradeSiteError(url: Uri, error: String?) {
         val params = mapOf(
             APP_VERSION to BuildConfig.VERSION_NAME,
-            URL to url.simpleUrl,
+            URL to "${url.scheme}://${url.host}",
```
CDRussell: Previously, we had a null check around the `url.scheme` - are we cool with not having that null check any more?

app/src/main/java/com/duckduckgo/app/global/job/JobBuilder.kt
```diff
@@ -31,9 +31,9 @@ class JobBuilder @Inject constructor() {
     fun appConfigurationJob(context: Context): JobInfo {
 
         return JobInfo.Builder(APP_CONFIGURATION_JOB_ID, ComponentName(context, AppConfigurationJobService::class.java))
-            .setPeriodic(TimeUnit.HOURS.toMillis(3))
+            .setPeriodic(TimeUnit.HOURS.toMillis(12))
```
CDRussell: 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
subsymbolic: We are. We also know we are dealing with a https site at this point so I have updated it to use the https scheme directly.

CDRussell: Even better! 👍

PR #380, opened 2018-10-17
Removes incorrect strlen(input) call

ryanhaining:
**Description**:
The function takes input and end, which implies that input is not a
C-string, and isn't a valid argument to strlen().

I'm getting crashes when running duckduckgo on android virtual devices (cuttlefish)
with address sanitizer and this changes gets it running again.

**Steps to test this PR**:
1. I don't know of public instructions on reproducing the crash, I'd need to get back to you on this if its necessary to demonstrate that this is actually fixing something


ryanhaining:
@subsymbolic

subsymbolic:
Thanks for submitting a PR. This code comes from a third party and will soon be removed from the project entirely so we are not improving it right now. Thanks again for contributing.

PR #387, opened 2018-10-29
Hotfix/5.12.1

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/885011869511894
Tech Design URL: 
CC: 

**Description**:
We made a change in `5.12.0` to mitigate a potential phishing attack, but sadly the callback used was only ever invoked in API 23 and newer. As such, we created a bug for Lollipop users where the app got into bad state (not able to add bookmarks, share pages, open new tabs).

This solution effectively reverts the phishing protection, but only for Lollipop users; those users will have the same functionality as before the fix went in. But newer devices will continue to be protected from that phishing attack. This seems like a good all-round compromise.

**Steps to test this PR**:
1. Test this on a Lollipop device - open a page/perform a search and ensure bookmarks/tabs/sharing works as expected
1. Test this out on a newer and ensure the above is all working well
1. Additionally, on a Marshmallow or newer device, test out to ensure the phishing attack still mitigated, as detailed in https://app.asana.com/0/0/865896277649322

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -111,20 +111,22 @@ class BrowserWebViewClient @Inject constructor(
 
         webViewClientListener?.loadingStarted()
 
+        if (!willGetNotifiedOfPageCommits()) onUrlChanged(url, webView)
+
         val uri = if (url != null) Uri.parse(url) else null
         if (uri != null) {
             reportHttpsIfInUpgradeList(uri)
         }
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -218,6 +220,18 @@ class BrowserWebViewClient @Inject constructor(
         return BrowserNavigationOptions(canGoBack, canGoForward)
     }
 
+    private fun willGetNotifiedOfPageCommits(): Boolean {
```
subsymbolic: This could be a property

PR #399, opened 2018-11-15
Fix to allow for IPs with ports and schemes with no intent matches

jlandrum:
Task/Issue URL: #347 

**Description**:
This is a proposal to fix certain URLs failing to load after app integration was added.
This works in two ways:
- By adding a new internally handled scheme of "search" which always runs a search. When an intent fails, it will re-submit the query with "search:" appended.
- By checking if the "scheme" is an IP address. This is more of a fault of the simplistic nature of the Android Uri parser which considers anything and everything before : to be a scheme. As such, ports become the path of a scheme of the IP.

**Alternate Approach**:
It may be better to replace the "search:" append with quotes, and update to always perform searches on quoted text. To the user, it would function exactly the same but would free up the scheme for use by the OS or installed apps.

**Steps to test this PR**:
1. Enter intent URLs such as market://details?id=com.duckduckgo.mobile.android
2. Enter IP addresses such as 192.168.0.1:8080
3. Enter things that may appear to be valid URLs but are intended to be searches such as define:duck
4. Enter a standard URL with and without scheme such as http://duckduckgo.com and duckduckgo.com

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -393,7 +393,7 @@ class BrowserTabFragment : Fragment(), FindListener {
                     if (appLinkCommand.appLink.fallbackUrl != null) {
                         webView?.loadUrl(appLinkCommand.appLink.fallbackUrl)
                     } else {
-                        showToast(R.string.unableToOpenLink)
+                        submitQuery("search:${appLinkCommand.appLink.url}")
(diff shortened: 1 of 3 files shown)


subsymbolic:
Thanks for submitting this. As it is quite tricky to fix without breaking other links and it requires some wider refactoring, we have logged the problem internally and will address it soon (see comment in https://github.com/duckduckgo/Android/issues/347). Thanks again for the contribution and discussion, all the brainstorming has been invaluable.

PR #406, opened 2018-12-13
When detecting mobile URLs, also includes "mobile."

CDRussell:
Task/Issue URL: https://app.asana.com/0/inbox/608920331025313/945591783985791/945614451233295
Tech Design URL: 
CC: 

**Description**:
Adds "mobile." in as a recognised mobile-specific URL

**Steps to test this PR**:
1. Visit facebook.com - it'll redirect to the mobile site by default
1. Verify "request desktop site" takes you back to the main facebook.com page

Ideally, if you know a site which uses the "mobile." format for mobile pages, test that too. I don't know of one to hand though*

*Twitter does use mobile.twitter.com, but it also forcefully puts us back to that even if request desktop site is requested, so it's not a great test case.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/UriExtension.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/UriExtension.kt
@@ -46,27 +46,22 @@ val Uri.hasIpHost: Boolean
         return baseHost?.matches(ipRegex) ?: false
     }
 
-private const val MOBILE_URL_PREFIX_SHORT = "m."
-private const val MOBILE_URL_PREFIX_LONG = "mobile."
+private val MOBILE_URL_PREFIXES = listOf("m.", "mobile.")
 
 val Uri.isMobileSite: Boolean
     get() {
         val auth = authority ?: return false
-        return auth.startsWith(MOBILE_URL_PREFIX_SHORT) || auth.startsWith(MOBILE_URL_PREFIX_LONG)
+        return MOBILE_URL_PREFIXES.firstOrNull { auth.startsWith(it) } != null
     }
 
 fun Uri.toDesktopUri(): Uri {
-    return if (isMobileSite) {
-        val urlString = toString()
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/global/UriExtension.kt
```diff
@@ -46,12 +46,24 @@ val Uri.hasIpHost: Boolean
         return baseHost?.matches(ipRegex) ?: false
     }
 
+private const val MOBILE_URL_PREFIX_SHORT = "m."
```
subsymbolic: Does it make sense to use a list here? I think it may simplify the logic and as a bonus we could add more of these as we find them without needing to rewrite the code. This would give us something along the lines of:
```
private val MOBILE_URL_PREFIXES = listOf("m.", "mobile.")

val Uri.isMobileSite: Boolean
    get() {
        val auth = authority ?: return false
        return MOBILE_URL_PREFIXES.firstOrNull { auth.startsWith(it) } != null
    }

fun Uri.toDesktopUri(): Uri {
    if (!isMobileSite) return this

    val newUrl = MOBILE_URL_PREFIXES.fold(toString()) { url, prefix ->
        url.replaceFirst(prefix, "")
    }
    return Uri.parse(newUrl)
}
```

CDRussell: _fancy_! 😄 

This has _slightly_ different than before, but that might be desirable; this code will continually strip mobile prefixes, rather than just doing it once.

So `https://m.mobile.example.com` and `https://mobile.m.example.com` will both resolve to `https://example.com`.

I think that's a desirable behavioural change. 🤔

CDRussell: I think that change is ok (and maybe even good, but at least ok 😄) so i've pushed the suggested changes.

Back at you @subsymbolic to RR.

PR #411, opened 2018-12-18
Explicitly disables WebSQL.

CDRussell:
this should be disabled already, as the docs state it is disabled by default. but given the recent Magellan, i figure it's best to be explicit so it doesn't accidentally get enabled in future.

Task/Issue URL: https://app.asana.com/0/414730916066338/949984894756229
Tech Design URL: 
CC: 

**Description**:
Explicitly disables WebSQL. This should be the default, but making this explicit (along with the rationale) should help ensure we don't accidentally enable it in future (without careful consideration)

**Steps to test this PR**:
1. Visit https://andrew.hedges.name/html5/sql.html in the app
1. Verify that it reports that `WebSQL` **is not** enabled

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -576,6 +576,9 @@ class BrowserTabFragment : Fragment(), FindListener {
                 useWideViewPort = true
                 builtInZoomControls = true
                 displayZoomControls = false
+
+                // explicitly disable database to try protect against Magellan WebSQL/SQLite vulnerability
+                databaseEnabled = false
                 mixedContentMode = WebSettings.MIXED_CONTENT_COMPATIBILITY_MODE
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -576,6 +576,9 @@ class BrowserTabFragment : Fragment(), FindListener {
                 useWideViewPort = true
                 builtInZoomControls = true
                 displayZoomControls = false
+
+                // explicitly disable database to try protect against Magellan WebSQL/SQLite vulnerability
```
subsymbolic: Does it make sense to add this to it's own method (could be an extension)? Something explicit like protectAgainstMagellanVulnerability(). I always worry that comments like this become ambiguous over time. For example, imagine we add two more storage settings underneath this one and it is no longer clear if the comment applies to just this line of code or all three.

PR #413, opened 2019-01-09
Feature/automatic data clearing home screen shortcut fix

CDRussell:
Task/Issue URL: Home screen shortcuts not launched correctly due to auto clearing
Tech Design URL: 
CC: 

**Description**:
With the introduction of auto clearing data (#405), we likely introduced (or at least exacerbated) a bug whereby sometimes we show a blank tab instead of the user's intended web page. 

This could present itself when launching the app from a home screen shortcut, or perhaps from a bookmark or sharing from another app.

**Problem**
The problem is concurrency-related, whereby we had multiple places a new tab could be created (and selected) and they would run on the `IO` scheduler (therefore, could be run on multiple threads concurrently).

**Solution**
The quickest fix for this is to stop using the `IO` scheduler for these tasks, and make use of a single-thread executor instead.

**Steps to test this PR**:
⚠️ As this is concurrency related, you might find your test device never (or rarely) shows the problem. Or conversely, always (or often) does, thus making testing it is fixed pretty tricky.

1. Enable data clearing option
1. Launch web page, add it to home screen shortcut
1. Navigate to another page
1. Background the app; wait for the allocated amount of time before data is cleared automatically
1. launch the app using the home screen shortcut
1. verify the app shows the intended home screen shortcut web page (and not the blank tab page)

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/InstantSchedulersRule.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/InstantSchedulersRule.kt
@@ -40,6 +40,7 @@ class InstantSchedulersRule : TestRule {
                 RxJavaPlugins.setIoSchedulerHandler { Schedulers.trampoline() }
                 RxJavaPlugins.setComputationSchedulerHandler { Schedulers.trampoline() }
                 RxJavaPlugins.setNewThreadSchedulerHandler { Schedulers.trampoline() }
+                RxJavaPlugins.setSingleSchedulerHandler { Schedulers.trampoline() }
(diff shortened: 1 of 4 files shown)


CDRussell:
Thanks @subsymbolic, I was able to reproduce that. Can you test again with the latest commit?

I think the issue was the view state was initializing to showing the web view content, instead of initializing to hiding it. This difference caused the incoming intent (launching the home screen shortcut) to get consumed immediately before the tabs were cleaned.

Now, with the latest commit, it should delay the processing of the intent until it's ready to consume it, as was originally intended.

PR #447, opened 2019-02-28
Fix list recreation

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1111964631553193
Tech Design URL: N/A

**Description**:
Removes recreation of entity list on every addition

**Steps to test this PR**:
1. Run a fresh install of the app
1. Allow entities to be downloaded (might take some time to sync)
1. Visit cnn.com and ensure that tracker list entities appear

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
+++ b/app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
@@ -52,7 +52,7 @@ class EntityListDownloader @Inject constructor(
                 Timber.d("Updating entity list data from server")
                 val body = response.body()!!
 
-                var entities = ArrayList<EntityListEntity>()
+                val entities = mutableListOf<EntityListEntity>()
                 body.entries.forEach {
                     val entity = it.key
 
--- a/app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
+++ b/app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
@@ -52,12 +52,12 @@ class EntityListDownloader @Inject constructor(
                 Timber.d("Updating entity list data from server")
                 val body = response.body()!!
 
-                var entities = emptyList<EntityListEntity>()
+                var entities = ArrayList<EntityListEntity>()
                 body.entries.forEach {
                     val entity = it.key
 
                     (it.value.properties + it.value.resources).forEach {
-                        entities += EntityListEntity(entityName = entity, domainName = it)
+                        entities.add(EntityListEntity(entityName = entity, domainName = it))
                     }
                 }
 


app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
```diff
@@ -52,12 +52,12 @@ class EntityListDownloader @Inject constructor(
                 Timber.d("Updating entity list data from server")
                 val body = response.body()!!
 
-                var entities = emptyList<EntityListEntity>()
+                var entities = ArrayList<EntityListEntity>()
```
CDRussell: A few suggestions:
1. `var entities` can become `val`
1. although equivalent, i prefer `mutableListOf<EntityListEntity>()` instead of `ArrayList<EntityListEntity>()`. I think it better signals intent that it's deliberately mutable
1. `domainName = it` shows a yellow warning in the IDE for me (implicit shadowed param `it`). could explicitly name it to avoid using `it` here?

         val entities = mutableListOf<EntityListEntity>()
         body.entries.forEach {
               val entity = it.key

               (it.value.properties + it.value.resources).forEach { domain ->
                    entities.add(EntityListEntity(entityName = entity, domainName = domain))
               }
          }

app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
```diff
@@ -52,12 +52,12 @@ class EntityListDownloader @Inject constructor(
...
                 body.entries.forEach {
                     val entity = it.key
 
                     (it.value.properties + it.value.resources).forEach {
-                        entities += EntityListEntity(entityName = entity, domainName = it)
+                        entities.add(EntityListEntity(entityName = entity, domainName = it))
```
CDRussell: Good catch! I was looking at this difference why you'd made this change thinking surely they'd be equivalent. Then I checked the source for `+=` operator overload... that's not intuitive! 😬

app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
subsymbolic: Scope creep!!! 😜

But good suggestions. I've done the first two but left the last one out as it's not introduced here, all over the app and I don't mind it. Let's cover that as a separate conversation.

CDRussell: a one-line change to make the file warning-free when you were already in there seems like a nice touch - thought that'd be right up your street. (and would have been quicker than this back and forth 😉)

but fair enough. it is out with the original PR intent.

subsymbolic: I'm just not sure I _agree_ with the warning. I'm sitting on the fence about whether we should conform or switch it off. I'll add it to next week's meeting as a point of style to agree on.

subsymbolic:
Many thanks for the reviews today @CDRussell.

app/src/main/java/com/duckduckgo/app/entities/api/EntityListDownloader.kt
CDRussell: Sure. For when the time comes: 

> In lambdas which are short and not nested, it's recommended to use the `it` convention instead of declaring the parameter explicitly. In nested lambdas with parameters, parameters should be always declared explicitly.

https://kotlinlang.org/docs/reference/coding-conventions.html

PR #452, opened 2019-03-05
Add try/finally around bloom filter creation

CDRussell:
Task/Issue URL: 
Tech Design URL: 
CC: 

**Description**:

**Steps to test this PR**:
1. 
1. 

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/httpsupgrade/HttpsUpgrader.kt
+++ b/app/src/main/java/com/duckduckgo/app/httpsupgrade/HttpsUpgrader.kt
@@ -95,8 +95,11 @@ class HttpsUpgraderImpl(
     @WorkerThread
     override fun reloadData() {
         dataReloadLock.lock()
-        httpsBloomFilter = bloomFactory.create()
-        dataReloadLock.unlock()
+        try {
+            httpsBloomFilter = bloomFactory.create()
+        } finally {
+            dataReloadLock.unlock()
+        }
     }
 
     companion object {


subsymbolic:
@CDRussell I know we discussed this however I am no longer sure what problem this solves. Either there is no exception to catch (in which case it does nothing) or there is and we should know what that is and handle it with intention. That code also hits the C++ layer, if that's throwing an exception the app will still crash.

CDRussell:
> @CDRussell I know we discussed this however I am no longer sure what problem this solves. Either there is no exception to catch (in which case it does nothing) or there is and we should know what that is and handle it with intention. That code also hits the C++ layer, if that's throwing an exception the app will still crash.

The problem, as I see it, is that we are currently `locking` without a guarantee of a subsequent `unlock`. This happens because we `lock`, perform an operation which might throw a runtime exception, and then `unlock` if no exception was thrown.

In the case where a runtime exception was thrown, our app might well carry on as normal depending on where it was called. If it happens during the `HttpsUpgradeDataDownloader` flow, then the sync will fail, but the app will continue. However, now that we've locked on a resource and never unlocked it, we've potentially created a deadlock that cannot be recovered from.

Wrapping the lock/unlock in a try/finally doesn't alter the existing exception handling behaviour. But it does remove one possible place our app could lock up.

It's advised in [java docs]:(https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html)
>care must be taken to ensure that all code that is executed while the lock is held is protected by try-finally or try-catch to ensure that the lock is released when necessary.
>
>In most cases, the following idiom should be used:

     Lock l = ...;
     l.lock();
     try {
         // access the resource protected by this lock
     } finally {
         l.unlock();
     }

There's an argument to be made that we shouldn't let the exception happen and carry on regardless. If we decide we want to change that behaviour, then this conversation might be moot. But if we want to keep this existing behaviour, we should tidy up what happens when an exception occurs I think.

subsymbolic:
I'm torn. On one hand, I want to follow the recommendation in the documentation and on the other I don't want a general try/catch block that could swallow and hide genuine problems from us that we then won't know to fix.

subsymbolic:
The solution here may be to refactor the code to remove the Reentrant lock.

brindy:
What about catching the exception and firing a pixel?  Won't stop people using the app, but we can monitor for problems.

subsymbolic:
It's an interesting idea but I don't think pixels are good replacement for stacktraces.

subsymbolic:
Does anyone object to refactoring this class to remove the ReEntrant lock (can be in a separate PR)? That would address all the concerns here - we could follow best practices and remove the try/catch.

subsymbolic:
It's just dawned on me that there is no catch here... so this is all fine. Approved!

Issue #455, opened 2019-03-16
Navigation bar clipping into action bar


JeffAment:
**App Version:** 5.19.0
**Device:** Pixel XL, Android 9.0
**The problem:** When scrolling up and down such that the navigation bar is hidden and then revealed, part of the navigation bar bleeds into the action bar.  See gifs below (it’s a little hard to see, so Gabe will show you the first one and the second one is a two-frame animation showing the part that is bleeding into the action bar flashing red so you can more easily see it).
**Steps to reproduce:** Scroll up and down until the navigation bar bleeds into the action bar.  Once the navigation bar clips into the action bar, it will stay there until you interact with the search bar or a new tab is opened etc.  On my device this is easily reproducible, ~~but I have not been able to reproduce this on an emulator yet (granted I could have tried harder).~~  Update: I have been able to reproduce this a few times on an emulator (Pixel XL, Android Q), but from my experience it takes a solid minute or two of scrolling for it to happen.

**Possible fix:** This no longer happens on my device after removing `android:fitsSystemWindows="true"` from the `AppBarLayout` in `include_omnibar_toolbar.xml` and moving it to the `CoordinatorLayout` in `fragment_browser_tab.xml`.

If this is something you would like to fix, I would be happy to submit a PR with this fix or review/test a PR that fixes this issue.  If this isn’t on the list of importance and you want to put this in your back pocket and save it for a rainy day, that’s cool too!  As always, if you need any more information on this or are having trouble reproducing this, feel free to ping me and I will do whatever I can to help.

![premiere mock_11](https://user-images.githubusercontent.com/21976019/54471391-9c9d6680-478e-11e9-9e1f-134033065d69.gif)
![DDGgif](https://user-images.githubusercontent.com/21976019/54471393-9dce9380-478e-11e9-8401-4ff72eb67727.gif)

subsymbolic:
Thanks for the bug report and for entertaining us @JeffAment  😆😆😆
Regarding your question about submitting a PR, that would be great! I should have time to review it later this week.

JeffAment:
Sounds good @subsymbolic.  I'll submit a PR as soon as I have a spare moment, most likely sometime tomorrow (Tuesday).  Also, thanks for not getting mad about the [very unprofessional] gif!  :smile:

CDRussell:
For the record, I love the gif! 😍

JeffAment:
[...]

PR #457, opened 2019-03-20
Fix issue 455

JeffAment:
Task/Issue URL:  #455

**Steps to test this PR**:
1. Launch a search query.
2. Scroll up and down such that the navigation bar is hidden and then revealed; verify that the navigation bar does not clip into the action bar.
3. Test other UI elements to make sure nothing else is broken.

Given the intermittent nature of this problem, I cannot say unequivocally that these changes will solve the problem.  What I can say, however, is that on my Pixel XL 9.0 I was always able to reproduce this bug within fifteen seconds, and after the changes I have been thoroughly abusing the app for days and have not seen any trace of this bug.

---
###### Internal references:

### DIFF
--- a/app/src/main/res/layout/fragment_browser_tab.xml
+++ b/app/src/main/res/layout/fragment_browser_tab.xml
@@ -20,6 +20,7 @@
     android:id="@+id/rootView"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    android:fitsSystemWindows="true"
     app:layout_scrollFlags="scroll|enterAlways">
 
     <include layout="@layout/include_omnibar_toolbar" />
--- a/app/src/main/res/layout/include_omnibar_toolbar.xml
+++ b/app/src/main/res/layout/include_omnibar_toolbar.xml
@@ -27,7 +27,6 @@
         android:id="@+id/appBarLayout"
         android:layout_width="match_parent"
         android:layout_height="?attr/actionBarSize"
-        android:fitsSystemWindows="true"
         android:theme="@style/AppTheme.Dark.AppBarOverlay"
         app:layout_scrollFlags="scroll|enterAlways">
 


subsymbolic:
Thanks for submitting this @JeffAment. I never managed to reproduce the original bleed issue you reported but would have accepted this change, however it introduces a new issue. When switching from landscape to portrait the omnibar can no longer be scrolled away.

To reproduce:
- Launch the app in landscape
- Enter a search
- Rotate to portrait
- Scroll to make the omnibar disappear, it is now glued in place

JeffAment:
Thanks for being perceptive and finding this bug, @subsymbolic!  I have not been able to reproduce this on my device and two emulators, and was wondering if you could tell me what kind of device this happened on?  I would be happy to look into this a bit more to see if I can find a solution that will work for both, but at the same time I perfectly understand if this has become a bigger time investment for you than you had originally anticipated.  If this is the case, please feel free to close this PR as I'd hate to ask more of your time and take you away from solving more important issues.

subsymbolic:
It happens on a Pixel 3 XL running Android 9. In case it helps, I have attached a gif of the issue.
top_bar_not_scrolling

subsymbolic:
@JeffAment I'm closing this PR however you are welcome to open it again if the no-scroll issue on Pixel 3 is fixed.

JeffAment:
@subsymbolic,

Sorry for the late reply!  I normally try and be more prompt than this, but had some unexpected personal issues arise.  I was not able to reproduce the issue that you had reported, as I do not have access to a Pixel 3 XL or the ability to emulate it.  I did try emulating devices with the same screen resolution and with notched screens, but was still not able to reproduce.  As of now, I do have some ideas that may work (and one that I believe will most likely work, but is not an ideal fix as it would add unnecessary overhead), but it is not fair to ask anyone to review a possibly endless stream of PR's that are based off complete guess work.  With that being said, I will keep this in mind and will open another PR should I ever find a way to reproduce this.  Conversely, I am also willing to test anyone elses future PR's (or even work in collaboration with anyone who can reproduce the Pixel 3 XL bug) if someone else should attempt to fix this, so please feel free to ping me should you ever need someone to test a fix for this, since I can very easily reproduce the original bug that I documented on my device.

Issue #383, opened 2018-10-18
No Basic Auth


dmaes:
### App Version
5.11.1 (via PlayStore)
### Device make and model
Motorola Moto G3
### Android version
7.1.2 (LineageOS 14.1-20171109-NIGHTLY-osprey)
### Steps to reproduce the bug
Visit a site with Basic Auth (try [httpbin.org](http://httpbin.org/#/Auth/get_basic_auth__user___passwd_))
### Expected behavior
A prompt opens to ask for your username and password
### Actual behavior
No prompt is opened, and you get a 401 status code with the message "unauthorized"

CDRussell:
Internal issue: https://app.asana.com/0/414730916066338/555891326561611

CDRussell:
Hi, thanks for raising this. We have this on our internal backlog but it's never bubbled up to the top priority wise so we haven't got to implementing it yet. If users vote on this issue with 👍 we'll maybe get it prioritised sooner.

dmaes:
Nice to hear it's already in the backlog, looking forward to see it implemented.

jd666:
Please implement.

dapacheco:
@CDRussell is there any desire to have a community PR to resolve this issue?

CDRussell:
Good suggestion; this probably should be flagged for community help to get it done. @subsymbolic you agree?

subsymbolic:
Great shout. It'll be a while before we can look at this, help from the community would be awesome. I've added the "help wanted" label.

CDRussell:
Implemented in https://github.com/duckduckgo/Android/pull/460

PR #460, opened 2019-03-22
383 - Add basic auth support

dapacheco:
Task/Issue URL: #383 
Tech Design URL: 
CC: 

**Description**:
Implement the `onReceivedHttpAuthRequest` callback in the `BrowserWebViewClient`.
Add a new dialog that is shown when the client gets a HTTP auth challenge, with supporting string resources.
Add new command to handle authentication requests.
Add tests to cover new functionality.

I've followed the Command pattern that is in use to communicate between the `BrowserTabFragment` and `BrowserTabFragment` and used the `SaveBookmarkDialogFragment` as a template for the new `HttpAuthenticationDialogFragment`.

**Steps to test this PR**:
Successful login:
1. Navigate to http://httpbin.org/basic-auth/test/12345
1. Verify Sign in dialog shown
1. Enter `test` as username, `12345` as password
1. Verify page is shown, content is JSON.

Failed login:
1. Navigate to http://httpbin.org/basic-auth/test/4321
1. Verify Sign in dialog shown
1. Enter `random` as username, `12345` as password
1. Verify auth dialog is shown again

Cancel login:
1. Navigate to http://httpbin.org/basic-auth/test/4321
1. Verify Sign in dialog shown
1. Cancel the dialog
1. Verify page is blank and URL is still http://httpbin.org/basic-auth/test/4321

Success:
http-auth-success

Failure and cancel:
http-auth-fail

---
###### Internal references:


subsymbolic:
Thanks for submitting the @dapacheco, I'll review it next week.

CDRussell:
@dapacheco thanks for this PR, it's a great addition! 

I haven't done a deep-dive through the code since @subsymbolic is planning to review it, but I skimmed it and like that's in very in-keeping with our style!  👌 

I gave the branch a run to see how it feels.

The correct creds and invalid creds paths are working perfectly for me; functionality and UX all as I would expect.

However, for the cancelation path, i think something needs tidied up here UX wise. 

**Observed behaviour**
As it stands, when the user hits cancel, the page is left in a weird state with the progress bar still showing and the URL doesn't really reflect the current page; it still shows the httpbin.org URL. if you try and refresh, you aren't offered the dialog again like i would expect.

**Desired behaviour**
I wonder if it would be acceptable to drop the `Navigate(BLANK_PAGE)` altogether, call the `request.handler.cancel()` and not try to further alter the `browserViewState`.

ie, simplify the `cancelAuthentication` callback to just:

        override fun cancelAuthentication(request: BasicAuthenticationRequest) {
            request.handler.cancel()
        }

The advantage of that, I think, would be that the user could just hit refresh to be re-prompted (and they won't continue to see the progress bar)

**Thoughts?** 
Chrome shows an error page for this scenario. That would be better than just leaving the user with a blank screen, but given it's from the user cancelling, hopefully it'll be obvious to them why they're not seeing the page they originally requested.

CDRussell:
_Example of the progress bar staying around after cancelling_

Kapture 2019-04-04 at 14 46 26

dapacheco:
@CDRussell thanks for the feedback.

Regarding the cancellation flow:
When I tested reloading the page I was submitting the request again via the _omnibar_ and not the `retry` function in the menu so I missed this scenario. :see_no_evil:

Clearing the view was a compromise in order to make it clear to the user that the page had failed to load. When I initially implemented this change, not doing this results in displaying the previous page's content.

However, I've just tested this again with your suggested change of just cancelling the authentication request, and not loading the `about:blank` page, and this does provide the desired behaviour. 👍 

auth-cancel

I'm more than happy to simplify the PR and make the change you have suggested, or to refine the current approach to match Chrome's behaviour which would mean:
- [ ] Adding an error view
- [ ] Clearing the progress bar
- [ ] Ensuring the refresh button works as expected

CDRussell:
I think adding an error view is probably too much to include here. I'm happy with the UX being the user is left with a blank screen in that scenario, and chatted it through with the designer who's also happy with that. So if you're happy removing the `about:blank` logic, i think that UX would be good.

Forget what I said earlier, I'll do a code review too since i've got some slack for it; see if we can get this merged soon.

_review incoming_

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -693,5 +698,21 @@ class BrowserTabViewModel(
...
+    }
+
+    override fun handleAuthentication(request: BasicAuthenticationRequest, credentials: BasicAuthenticationCredentials) {
+        request.handler.proceed(credentials.username, credentials.password)
+        // Attempt to save the credentials
+        command.value = SaveCredentials(request, credentials)
+    }
+
+    override fun cancelAuthentication(request: BasicAuthenticationRequest) {
+        Timber.v("cancelAuthentication, loading about:blank")
```
CDRussell: i think we can kill this log statement; it's no longer accurate but i don't think it's needed generally

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -693,5 +698,21 @@ class BrowserTabViewModel(
...
     }
+
+    override fun requiresAuthentication(request: BasicAuthenticationRequest) {
+        Timber.v("requiresAuthentication for [${request.site}]")
+        command.value = RequiresAuthentication(request)
+    }
+
+    override fun handleAuthentication(request: BasicAuthenticationRequest, credentials: BasicAuthenticationCredentials) {
+        request.handler.proceed(credentials.username, credentials.password)
+        // Attempt to save the credentials
```
CDRussell: we can remove this comment - it's clear from the surrounding code what's going on

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -166,6 +168,57 @@ class BrowserWebViewClient @Inject constructor(
...
+    @UiThread
+    override fun onReceivedHttpAuthRequest(view: WebView?, handler: HttpAuthHandler?, host: String?, realm: String?) {
+        Timber.v("onReceivedHttpAuthRequest ${view?.url} $realm, $host,  $currentUrl")
+
+        if (handler != null) {
+            Timber.v("onReceivedHttpAuthRequest - useHttpAuthUsernamePassword [${handler.useHttpAuthUsernamePassword()}]")
+            if (handler.useHttpAuthUsernamePassword()) {
+                val webViewDatabase = WebViewDatabase.getInstance(view?.context)
+
+                val credentials = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
CDRussell: might be neater to refactor this little block out to its own function

`buildAuthCredentials()` maybe? that would help hide the `SDK_INT` logic away and keep this function's logic clearer

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -166,6 +168,57 @@ class BrowserWebViewClient @Inject constructor(
...
+                    webViewDatabase.getHttpAuthUsernamePassword(host.orEmpty(), realm.orEmpty())
+                } else {
+                    @Suppress("DEPRECATION")
+                    view?.getHttpAuthUsernamePassword(host.orEmpty(), realm.orEmpty())
+                }
+
+                credentials?.let {
```
CDRussell: we tend to only use `?.let` for when we don't care about the `null` case (though i'm sure there are plenty of cases where we don't stick to that convention)

however, here, i think it would be clearer to give with an `if/else`

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -166,6 +168,57 @@ class BrowserWebViewClient @Inject constructor(
...
+    private fun showAuthenticationDialog(
+        view: WebView?,
+        handler: HttpAuthHandler,
+        host: String?,
+        realm: String?
+    ) {
+        webViewClientListener?.let {
+            Timber.v("showAuthenticationDialog - $host, $realm")
+
+            val siteURL = view?.url?.let {
```
CDRussell: As above, maybe more in-keeping with our project style to use `if/else`

app/src/main/java/com/duckduckgo/app/browser/ui/HttpAuthenticationDialogFragment.kt
```diff
@@ -0,0 +1,100 @@
...
+    private fun validateBundleArguments() {
+        if (arguments == null) throw IllegalArgumentException("Missing arguments bundle")
+        val args = arguments!!
+        if (!args.containsKey(KEY_TARGET_URL)) {
+            throw IllegalArgumentException("Bundle arguments required [KEY_TARGET_URL]")
+        }
+    }
+
+    private fun showKeyboard(titleInput: EditText, alert: androidx.appcompat.app.AlertDialog) {
+        titleInput.setSelection(titleInput.text.length)
```
CDRussell: what's this line needed for?

app/src/main/res/layout/http_authentication.xml
```diff
@@ -0,0 +1,68 @@
...
+    <com.google.android.material.textfield.TextInputLayout
+        android:id="@+id/httpAuthUsernameContainer"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="20dp"
+        android:layout_marginEnd="20dp"
+        android:layout_marginTop="20dp"
+        app:layout_constraintTop_toBottomOf="@id/httpAuthInformationText">
+
+        <EditText
```
CDRussell: We might want to mark these fields with autocomplete hints, for example, for the `username` one:

     android:importantForAutofill="yes"
     android:autofillHints="username"
     tools:ignore="UnusedAttribute" 

And `android:autofillHints="password"` for the password field.

That said, I haven't tested this, and we haven't done it anywhere else in the app before. So leaving a note here as a future note - feel free to ignore this for now.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -693,5 +698,21 @@ class BrowserTabViewModel(
     override fun externalAppLinkClicked(appLink: IntentType) {
         command.value = HandleExternalAppLink(appLink)
     }
+
+    override fun requiresAuthentication(request: BasicAuthenticationRequest) {
+        Timber.v("requiresAuthentication for [${request.site}]")
```
CDRussell: Can we kill this comment? At first glance i thought this might log out credentials. I realise that's totally not the case, but even just to avoid anyone else having a similar thought, there's probably no need for the comment anyway.

app/src/main/java/com/duckduckgo/app/browser/ui/HttpAuthenticationDialogFragment.kt
dapacheco: This function was copied from `SaveBookmarkDialogFragment`. I've refactored the `titleInput` parameter to a more generic name and removed this line as the `username` edit text has focus without making this call.

app/src/main/res/layout/http_authentication.xml
dapacheco: When testing I was able to use 1Password to fill in the form so I wasn't sure if these were needed. Happy to add for further compatibly.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
dapacheco: 👍 That seems like a reasonable convention.

app/src/main/java/com/duckduckgo/app/browser/ui/HttpAuthenticationDialogFragment.kt
CDRussell: Ah, I see. I think it was only added there because those inputs could be pre-populated with the page title and/or URL, and that line would make the cursor jump to the end of the text rather than the start.

But agree, not needed in this case.

app/src/main/res/layout/http_authentication.xml
CDRussell: If it's working for you as is, feel free to leave it. It can be a small addition later if it looks like it would be needed.

As I understand it, autofill heuristics will automatically determine fields' eligibility for autofill if it can. So that's probably why we're getting it for free. The EditText is probably inspected and it likely knows it's a username and password.

We can add these explicitly later in a follow up if needed.

CDRussell:
This is a great addition @dapacheco! I was hoping to tackle this around a year ago and just never got the time as other things kept popping up, so very pleased to see it get added.

Thanks again! much appreciated

Issue #473, opened 2019-04-13
HTML5 audio controls


usabilitest:
There seems to be an issue with PLAY control of HTML5 audio element in DuckDuckGo on Android.

subsymbolic:
Thanks for reporting this, we will investigate.

subsymbolic:
I can confirm that tapping the play button does nothing on https://tekeye.uk/html/html5-video-test-page however hitting the sound button and THEN pressing play does work. I have opened this up to the community to fix.

Internal task reference: https://app.asana.com/0/414730916066338/1118831380832661.

subsymbolic:
It looks like this issue can be solved by turning on `hardwareAcceleration` in the manifest. I'll test and include that in the next release. Thanks again for reporting @usabilitest.

subsymbolic:
This issue seems to be os specific (see comment in https://github.com/duckduckgo/Android/pull/474). The html5 video play button works fine on
- Lollipop devices
- Q devices
- all emulators

And does not seem to work on:
- Pie devices

We have not yet checked Marshmallow, Nougat or Oreo so these may be affected too.

It may be possible to overcome the issue with some javascript touch magic. I'll continue to leave this task open to the community.

CristianMG:
Hi 

I´m trying to reproduce the error, I have discovered that in my Xiaomi MI8 if I install after deleting all data(uninstall the application) reproduced video is a success, it seems that reproduction video is broken after deleting tabs and data inside the app.

I don't see any log error suspect moreover I will attach logs in this comment and video 

XIAOMI MI8 
Android version: 9
MIUI Version: GLOBAL 10.2.1 Stable 10.2.3(PEAMIXM) 

I will try to get more information about the error

Greetings

https://we.tl/t-2TYm2TXerV

nicolashleon:
Hi! 

I'm just trying to provide some additional information on the issue as well as the previous investigation on #474

After trying the release and develop versions of the app both seem to be working as expected for the following environment:

- **DDG Android App Version** 5.23.0(release) and 5.24.0(develop)
- **Device** Pixel 2
- **OS Version** Android 9
- **Chrome Version** 74.0.3729.136

In my opinion, this might be related to the WebView version or the OS but we will need to have some extra data regarding the environments in which the error was successfully reproduced to narrow the cause down.
 
Hope this helps!

subsymbolic:
[...]

PR #474, opened 2019-04-17
Provide a fix to the issue of HTML 5 video not playing #473

dapacheco:
Task/Issue URL: #473
Tech Design URL: 
CC: 

**Description**:
By design, media content requires a user _gesture_ for it to be enabled in a WebView. It is unclear what constitutes a _gesture_ from the API docs.

This setting can be overridden by using the following API - [setMediaPlaybackRequiresUserGesture(boolean)](
https://developer.android.com/reference/android/webkit/WebSettings.html#setMediaPlaybackRequiresUserGesture(boolean))

The default for this setting is _true_, which explains the current behaviour in the app and why toggling the sound allows the content to be played.

[This SO answer](https://stackoverflow.com/a/34535582/3861647) highlights this issue -  and provides a link to a bug that has been raised against Google and their decision to maintain this behaviour: https://bugs.chromium.org/p/chromium/issues/detail?id=178297

**Steps to test this PR**:
1. Navigate to a page with HTML 5 video - https://tekeye.uk/html/html5-video-test-page 
1. Try to play the media
1. Launch settings and disable 'Require gesture for media playback'
1. Reload tab
1. Try to play the media - the video should now play

To verify autoplay behaviour for a video use this site - http://camendesign.com/code/video_for_everybody/test.html, when gesture is required the video won't autoplay. However, when the gesture is not required loading the page will autoplay the video.

The following was taken into consideration:
- Setting the `setMediaPlaybackRequiresUserGesture` value to `false` would enable `autoplay` for all media which may have an undesired side-effect. This is why I added a toggle under settings.
- Using a mechanism similar to a theme change to apply the setting to existing tabs would force a reload of the page without alerting the user beforehand. So I went with an implementation that requires the user to reload the page after making the setting change.

There is still an issue of discoverability of this setting and how changing it modifies the behaviour of the web view.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt
@@ -56,12 +56,16 @@ class SettingsActivity : DuckDuckGoActivity(), SettingsAutomaticallyClearWhatFra
 
     private val defaultBrowserChangeListener = OnCheckedChangeListener { _, _ -> launchDefaultAppScreen() }
 
+    private val playMediaRequiresGestureChangeListener = OnCheckedChangeListener { _, isChecked ->
+        viewModel.onPlayMediaRequireGestureChanged(isChecked)
+    }
+
     private val lightThemeToggleListener = OnCheckedChangeListener { _, isChecked ->
         viewModel.onLightThemeToggled(isChecked)
     }
 
     private val autocompleteToggleListener = OnCheckedChangeListener { _, isChecked ->
-        viewModel.onAutocompleteSettingChanged(isChecked)
+        viewModel.onPlayMediaRequireGestureChanged(isChecked)
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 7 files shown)


subsymbolic:
Thanks for your PR @dapacheco, your investigation into this issue is really helpful. In terms of desirable UX we would like autoplay off by default and pressing play to start a video rather than a setting. If that doesn't look feasible, we'll need to investigate other options in-house.

app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt
```diff
@@ -56,12 +56,16 @@ class SettingsActivity : DuckDuckGoActivity(), SettingsAutomaticallyClearWhatFra
...
+        viewModel.onPlayMediaRequireGestureChanged(isChecked)
+    }
+
     private val lightThemeToggleListener = OnCheckedChangeListener { _, isChecked ->
         viewModel.onLightThemeToggled(isChecked)
     }
 
     private val autocompleteToggleListener = OnCheckedChangeListener { _, isChecked ->
-        viewModel.onAutocompleteSettingChanged(isChecked)
+        viewModel.onPlayMediaRequireGestureChanged(isChecked)
```
subsymbolic: Revert unintentional change.

dapacheco:
I had only tested this issue on a device running Android 9 (a Pixel XL) and had verified the behaviour described so I didn't try any other devices before looking for a possible fix.

However, I've been thinking that this behaviour seems more like a bug with the WebView than anything else.

The autoplay requiring a gesture seems resonable but the inability to play video without first toggling the sound by default seems broken, even if changing `setMediaPlaybackRequiresUserGesture` _fixes_ it.

I found an older device, running 5.0.1, to test this on and it works as expected without any changes. This was also running an older version of Chrome, but after updating to the latest version of Chrome that too worked.

I've also tested a number of emulators and I've not been able to reproduce the broken behaviour. Even one that matches the OS and Chrome version of the physical device I've tested on. Although emulators don't always show the same issues as devices.

Finally, I've just updated the Pixel to Android Q and this works as expected. As such, I'm going to close this PR given that the root cause appears to be a system bug that affects WebViews (I've been testing on a simple WebView app to isolate any DDG app changes from this problem).

subsymbolic:
Thanks for investigating @dapacheco. Much appreciated.

PR #475, opened 2019-04-19
Improve android back navigation

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1117886846776279
Tech Design URL: https://app.asana.com/0/361428290920652/1118398381563523 and https://app.asana.com/0/361428290920652/1118398381563522

**Description**:
Improves three aspects of android back navigation:
- Pressing back after launching from another app takes user back to the previous application (rather than via home screen)
- When pressing back to get to the home screen, users can now press forward again to return to webpage they were viewing
- Pressing back on a https-upgraded site navigates two steps to avoid sending user back to the http site which just upgrades again

**Steps to test this PR**:
MIGRATION TEST
1. Install the develop version of the app.
1. Open some tabs
1. Install this version and make sure nothing goes "boom"

FORWARD TEST
1. Go to a webpage
1. Press the back button to go back to the home page
1. Press the forward button to go back to the webpage

SHARE TEST
1. Share a link from twitter to our app
1. Now from our app, press back
1. This should return you to twitter (without going via the home screen)

HTTPS Test
1. Navigate to http://duckduckgo.com
1. Now navigate to http://facebook.com
1. Press back, this should take you to https://duckduckgo.com
1. Press back again, this should take you to the native home screen
---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -318,10 +318,16 @@ class BrowserTabViewModel(
         }
     }
 
+    /**
+     * Handles back navigation. Returns false if navigation could not be
+     * handled at this level, giving system an opportunity to handle it
+     *
+     * @return true if navigation handled, otherwise false
+     */
     fun onUserPressedBack(): Boolean {
-        val navigation = navigationOptions
+        val navigation = navigationOptions ?: return false
 
-        if (!currentBrowserViewState().browserShowing || navigation == null) {
+        if (!currentBrowserViewState().browserShowing) {
             return false
         }
 
@@ -341,13 +347,13 @@ class BrowserTabViewModel(
         site = null
         onSiteChanged()
 
-        omnibarViewState.value = currentOmnibarViewState().copy(omnibarText = "")
-        loadingViewState.value = currentLoadingViewState().copy(isLoading = false)
         browserViewState.value = currentBrowserViewState().copy(
             browserShowing = false,
             canGoBack = false,
             canGoForward = true
         )
+        omnibarViewState.value = currentOmnibarViewState().copy(omnibarText = "")
+        loadingViewState.value = currentLoadingViewState().copy(isLoading = false)
     }
 
     override fun progressChanged(progressedUrl: String?, newProgress: Int) {
@@ -393,7 +399,7 @@ class BrowserTabViewModel(
         if (!currentBrowserViewState().browserShowing) return
 
         browserViewState.value = currentBrowserViewState().copy(
-            canGoBack = true,
+            canGoBack = navigation.canGoBack || !skipHome,
             canGoForward = navigation.canGoForward
         )
     }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -43,11 +43,11 @@ import com.duckduckgo.app.browser.LongPressHandler.RequiredAction
 import com.duckduckgo.app.browser.SpecialUrlDetector.UrlType.IntentType
 import com.duckduckgo.app.browser.addtohome.AddToHomeCapabilityDetector
 import com.duckduckgo.app.browser.favicon.FaviconDownloader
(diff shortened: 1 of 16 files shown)


subsymbolic:
FYI @CDRussell, I had meant to put this in draft PR (and now can't) as I have some more tests to submit. Will let you know when they're in.

subsymbolic:
@CDRussell this is ready for review, ping me if you would like to zoom.

CDRussell:
for the FORWARD TEST, it appears to immediately show the loaded web page, but then also reload the web page. is that reload necessary?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -318,23 +385,30 @@ class BrowserTabViewModel(
...
+        if (!currentBrowserViewState().browserShowing) return
         val progress = currentLoadingViewState()
         loadingViewState.value = progress.copy(isLoading = true)
         pendingUrl = url
         site = null
         onSiteChanged()
     }
 
-    override fun navigationOptionsChanged(navigationOptions: BrowserNavigationOptions) {
+    override fun navigationOptionsChanged(navigation: BrowserNavigationOptions) {
```
CDRussell: if changing the param name, should probably change it in the super class too (lint warning)

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -268,12 +269,26 @@ class BrowserWebViewClient @Inject constructor(
...
 
+        override val stepsToPreviousPage: Int = if (stack.isHttpsUpgrade) 2 else 1
+        override val canGoBack: Boolean = stack.currentIndex >= stepsToPreviousPage
+        override val canGoForward: Boolean = stack.currentIndex + 1 < stack.size
+        override val hasNavigationHistory = stack.size != 0
+
+        private val WebBackForwardList.isHttpsUpgrade: Boolean
+            get() {
+                if (currentIndex < 1) return false
+                val current = currentItem.originalUrl
```
CDRussell: `currentItem` is flagged as `nullable`

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -268,12 +269,26 @@ class BrowserWebViewClient @Inject constructor(
...
-        return BrowserNavigationOptions(canGoBack, canGoForward)
+    interface BrowserNavigationOptions {
+        val stepsToPreviousPage: Int
+        val canGoBack: Boolean
+        val canGoForward: Boolean
+        val hasNavigationHistory: Boolean
     }
 
-    data class BrowserNavigationOptions(val canGoBack: Boolean, val canGoForward: Boolean)
+    data class WebViewNavigationOptions(val stack: WebBackForwardList): BrowserNavigationOptions {
```
CDRussell: needs formatted

app/src/main/java/com/duckduckgo/app/tabs/model/TabEntitiy.kt
```diff
@@ -31,6 +31,7 @@ data class TabEntity(
     @PrimaryKey var tabId: String,
     var url: String? = null,
     var title: String? = null,
+    var skipHome: Boolean = false,
```
CDRussell: In the original design, we were just planning on keeping this flag in the `viewModel` and not in the DB. why did we need to store it in the DB?

(i'm absolutely cool with this approach btw, just curious why)

CDRussell:
This PR fixes the single biggest complaint I personally have about our app, so I can't put into words how happy and excited I am to see this work completed. kudos! 🎉

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
subsymbolic: Ahh, so it is. I've checked the list size so it definitely exists, I'll make that more clear.

app/src/main/java/com/duckduckgo/app/tabs/model/TabEntitiy.kt
subsymbolic: That's a good question. It's because it's the db change that triggers the live data update that then results in opening a new tab.

PR #495, opened 2019-05-29
Enable WebView debugging during BuildConfig.DEBUG builds

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1124956056611274
Tech Design URL: 
CC: 

**Description**:
Allows `WebView`s to be remotely debugged when the app is running in `BuildConfig.DEBUG` mode only.

**Steps to test this PR**:
1. From a `DEBUG` build, using Chrome on the desktop, visit `chrome://inspect` and verify the `WebView` shows and can be inspected
1. Repeat the above on a `RELEASE` build and ensure it cannot be inspected

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -660,7 +660,9 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
             it.setFindListener(this)
         }
 
-        WebView.setWebContentsDebuggingEnabled(BuildConfig.DEBUG)
+        if (BuildConfig.DEBUG) {
+            WebView.setWebContentsDebuggingEnabled(true)
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -659,6 +659,8 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
 
             it.setFindListener(this)
         }
+
+        WebView.setWebContentsDebuggingEnabled(BuildConfig.DEBUG)
```
subsymbolic: Optional: When we optimize then this code could disappear if we use this format
```
if (BuildConfig.DEBUG) {
   WebView.setWebContentsDebuggingEnabled(true)
}
```

CDRussell: i'm not sold on the (premature) optimization angle, but i think your suggestion is more readable. so we agree - but for different reasons 😄

PR #498, opened 2019-05-31
Release privacy notification on day one of inactivity to all users

subsymbolic:
Task/Issue URL: 
https://app.asana.com/0/361428290920652/1114172401555990
https://app.asana.com/0/361428290920652/1114174327194184

Tech Design URL: N/A

**Description**:
Removes various notification scenarios and releases **privacy stats** notification on day 1 of inactivity (also maintains existing **clear data**  notification on day 3).

**Steps to test this PR**:
Open the notification scheduler and change the privacy notification worker schedule from 1 day to 10 seconds:
```
scheduleNotification(OneTimeWorkRequestBuilder<PrivacyNotificationWorker>(), 10, TimeUnit.SECONDS)
```

And the clear data notification from 3 days to 30 seconds
```
scheduleNotification(OneTimeWorkRequestBuilder<ClearDataNotificationWorker>(), 30, TimeUnit.SECONDS)
```

1. Run a clean install of the app and background it
1. Ensure that the privacy notification appears after roughly 10 seconds (the scheduler isn't terribly precise)
1. Tap it to launch the app and then background it
1. Ensure that the clear data notification appears after roughly 30 seconds
1. Tap it to launch the app, then background the app again
1. No more notifications should appear
1. Kill and restart the app. Again no new notifications should appear

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/notification/NotificationSchedulerTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/notification/NotificationSchedulerTest.kt
@@ -21,6 +21,8 @@ package com.duckduckgo.app.notification
 
 import androidx.work.WorkInfo
 import androidx.work.WorkManager
+import com.duckduckgo.app.notification.NotificationScheduler.ClearDataNotificationWorker
+import com.duckduckgo.app.notification.NotificationScheduler.PrivacyNotificationWorker
 import com.duckduckgo.app.notification.model.SchedulableNotification
 import com.duckduckgo.app.statistics.VariantManager
 import com.duckduckgo.app.statistics.VariantManager.Companion.DEFAULT_VARIANT
@@ -32,6 +34,7 @@ import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
+import kotlin.reflect.jvm.jvmName
 
 class NotificationSchedulerTest {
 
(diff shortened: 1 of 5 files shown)


app/src/androidTest/java/com/duckduckgo/app/notification/NotificationSchedulerTest.kt
```diff
@@ -52,24 +55,47 @@ class NotificationSchedulerTest {
...
-        return WorkManager
+    private fun notificationScheduled(workerName: String? = null): Boolean {
+        val workers = WorkManager
             .getInstance()
             .getWorkInfosByTag(NotificationScheduler.WORK_REQUEST_TAG)
             .get()
-            .any { it.state == WorkInfo.State.ENQUEUED }
+            .filter { it.state == WorkInfo.State.ENQUEUED }
+
+        if (workerName == null) {
```
CDRussell: can you explain the changes made in here?

subsymbolic: We originally only had one type of notification, now that we have a second we need to check that the correct one has been queued, it's a bit clunky. I can rework it if you think it's unreadable.

app/src/androidTest/java/com/duckduckgo/app/notification/NotificationSchedulerTest.kt
```diff
@@ -52,24 +55,47 @@ class NotificationSchedulerTest {
...
+    fun whenNoNotificationCanShowThenNoNotificationScheduled() = runBlocking<Unit> {
+        whenever(privacyNotification.canShow()).thenReturn(false)
         whenever(clearNotification.canShow()).thenReturn(false)
         testee.scheduleNextNotification()
         assertFalse(notificationScheduled())
     }
 
-    private fun notificationScheduled(): Boolean {
-        return WorkManager
+    private fun notificationScheduled(workerName: String? = null): Boolean {
```
CDRussell: it feels odd that this parameter is `nullable`, and it seems like the only reason it can be is to allow `assertFalse(notificationScheduled())` to be executed.

I think it would make it more readable to make another function to assert that no notifications have been scheduled, and change the existing function to accept a non-nullable class name.

Could you replace `assertFalse(notificationScheduled())` with something like this...? `assertNoNotificationsScheduled()`

        private fun assertNoNotificationsScheduled() {
        val workers = WorkManager
            .getInstance()
            .getWorkInfosByTag(NotificationScheduler.WORK_REQUEST_TAG)
            .get()
            .filter{ it.state == WorkInfo.State.ENQUEUED }

        assertTrue(workers.isEmpty())
    }

app/src/androidTest/java/com/duckduckgo/app/notification/NotificationSchedulerTest.kt
CDRussell: Ah, got you. yeah, something making it tricky for me to scan and understand it quickly. i wrote a suggestion above that would help my understanding of the code... lmk if that works for you

app/src/main/java/com/duckduckgo/app/notification/NotificationScheduler.kt
```diff
@@ -47,13 +46,10 @@ class NotificationScheduler @Inject constructor(
         val variant = variantManager.getVariant()
```
CDRussell: `variant` is now unused

app/src/main/java/com/duckduckgo/app/notification/NotificationScheduler.kt
```diff
@@ -72,7 +68,7 @@ class NotificationScheduler @Inject constructor(
 
     // Legacy code. Unused class required for users who already have this notification scheduled from previous version. We can delete this in a
     // couple of weeks once old notifications have cleared. We should also remove "open" access from ClearDataNotificationWorker.
-    class ShowClearDataNotification(context: Context, params: WorkerParameters): ClearDataNotificationWorker(context, params)
+    class ShowClearDataNotification(context: Context, params: WorkerParameters) : ClearDataNotificationWorker(context, params)
```
CDRussell: when has "a couple of weeks" passed? do we have a task scheduled for this cleanup?

subsymbolic: That's well and truly passed. No need for a follow up, I'll delete it now.

CDRussell: Oh, actually, maybe we should do it at the same time as https://app.asana.com/0/414730916066338/1119619712088571?

So either leave it be until that's done, or do that task while you're removing it?

subsymbolic: I did think about this but then figured that this notification shouldn't exist at all anymore unless some user newly installs the version that contained it. Stranger things have happened though and users may reset their app data and mimic a fresh install. I've left it in but updated the comment and the asana task. I think that's safest. Good shout!

Issue #499, opened 2019-06-02
Site-specific search doesn't work; attempts to open as a link instead and fails


arunsathiya:
When searching for `site:producthunt.com change password` using the search/URL box at the top, the process fails with a `Unable to open this type of link`.

Visiting DuckDuckGo.com and typing the same works though. 

It's just that searching on the URL box fails.

GuiltyDolphin:
See also #347

arunsathiya:
Oh, I completely missed that. I see an open PR against this issue. @subsymbolic, could you advise on which issue could be closed at this stage? 🤔

subsymbolic:
Thanks @arunsathiya and @GuiltyDolphin. I'll close this one as a duplicate as it came after https://github.com/duckduckgo/Android/issues/347.

PR #503, opened 2019-06-08
Issue/499 - Fix: Site-specific search doesn't work; attempts to open as a link instead and fails

donfuxx:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/499 https://github.com/duckduckgo/Android/issues/347
Tech Design URL: N/A

**Description**:
1. Fixed bug of searches in address bar with `site:` scheme leading to a `Unable to open this type of link` toast message instead of showing the search results, by adding a new site scheme case in the `SpecialUrlDetector`. 
1. Added unit tests for new search cases

**Steps to test this PR**:
1. Go to top address bar and type a search like `site:duckduckgo.com about`
1. See search results displayed
1. Alternatively run new unit tests

dgg-issue476

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/SpecialUrlDetectorImplTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/SpecialUrlDetectorImplTest.kt
@@ -127,4 +127,18 @@ class SpecialUrlDetectorImplTest {
         val type = testee.determineType("smsto:123-555-12323") as Sms
         assertEquals("123-555-12323", type.telephoneNumber)
     }
+
+    @Test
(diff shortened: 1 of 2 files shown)


subsymbolic:
Thanks for your PR. This is a deceptive one as `site` is not really a scheme. All of these examples should be converted into searches:
- `about site:duckduckgo.com`
- `site:duckduckgo.com about`
- `site:duckduckgo.com`

This applies to other special search terms like `related` and `define` too. It is tricky to fix this without doing wider refactoring, so we have logged it internally and will fix it soon. Thanks again for contributing and getting us to think about this issue again. It's a pain-point for many people and it's good to have it back on our radar.

Issue #476, opened 2019-04-20
Allow encoded spaces in url path


GuiltyDolphin:
App version: 5.21.2 (via F-Droid)

When opening a link (in a new tab, or typing the URL in manually) which has `search/X%20Y` at the end, the page will open as a DDG search (with the URL Decode IA showing) rather than on the expected webpage.

For example, the search `http://foo.com/search/bar%20baz` displays as a DDG search, but `http://foo.com/search/bar` loads the expected webpage.

subsymbolic:
This is now fixed and will be available in the next release. Thanks @GuiltyDolphin and @donfuxx.

PR #504, opened 2019-06-08
Allow encoded spaces in url path

donfuxx:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/476
Tech Design URL: N/A

**Description**:
1. fix urls like http://www.example.com/pa%20th leading to search query. spaces shouldn't be in encoded uriString, but decoded path can have spaces
1. Add unit test to prove valid urls like http://www.example.com/pa%20th are treated like web urls

**Steps to test this PR**:
1. Type `http://www.example.com/pa%20th` in top address bar and submit search
1. see the webpage with title `Example Domain` displayed
1. Alternatively run new unit test

dgg-issue476

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/global/UriString.kt
```diff
@@ -42,7 +42,7 @@ class UriString {
             if (uri.scheme != UrlScheme.http && uri.scheme != UrlScheme.https) return false
             if (uri.userInfo != null) return false
             if (uri.host == null) return false
-            if (uri.path.contains(space)) return false
+            if (inputQuery.contains(space)) return false
```
subsymbolic: This is great. Can you move it to the beginning of the method (before we start building and interrogating the uri)?

donfuxx: Done!

PR #509, opened 2019-06-14
Use counts in leaderboard

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1127123170200348

**Description**:
Switches to counts in the tracker leaderboard

**Steps to test this PR**:

Test migration and new data:
1. Switch to develop branch
1. Update`PrivacyDashboardViewModel.kt` variable `LEADERBOARD_MIN_DOMAINS_EXCLUSIVE` to 3
1. Run the app, visit a few pages and make sure the tracker leaderboard is showing
1. Switch to this branch
1. Run the app and ensure the migration succeeds
1. Visit the privacy dash, the leaderboard should reset
1. Visit a few pages and make sure the tracker leaderboard shows again

Check updated database:
1. Download the app.db database from app storage
1. Check the network_leaderboard table and ensure network entries are now counts not strings
1. Check the sites_visited table and ensure the entries are now counts not strings

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/privacy/renderer/TrackersRendererTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/privacy/renderer/TrackersRendererTest.kt
@@ -18,7 +18,7 @@ package com.duckduckgo.app.privacy.renderer
 
 import androidx.test.platform.app.InstrumentationRegistry
 import com.duckduckgo.app.browser.R
-import com.duckduckgo.app.privacy.db.NetworkLeaderboardDao.NetworkTally
+import com.duckduckgo.app.privacy.db.NetworkLeaderboardEntry
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
 import org.junit.Test
@@ -65,26 +65,26 @@ class TrackersRendererTest {
     }
 
     @Test
-    fun whenTotalDomainsIsZeroThenPercentageIsBlank() {
-        val text = testee.networkPercentage(NetworkTally("", 10), 0)
+    fun whenSitestVisitedIsZeroThenPercentageIsBlank() {
+        val text = testee.networkPercentage(NetworkLeaderboardEntry("", 10), 0)
         assertEquals("", text)
     }
 
     @Test
-    fun whenDomainCountIsZeroThenPercentageIsBlank() {
-        val text = testee.networkPercentage(NetworkTally("", 0), 10)
+    fun whenNetworkCountIsZeroThenPercentageIsBlank() {
+        val text = testee.networkPercentage(NetworkLeaderboardEntry("", 0), 10)
         assertEquals("", text)
     }
 
     @Test
     fun whenPortionIsRecurringFractionThenPercentageIsRoundNumber() {
-        val text = testee.networkPercentage(NetworkTally("", 10), 30)
+        val text = testee.networkPercentage(NetworkLeaderboardEntry("", 10), 30)
         assertEquals("33%", text)
     }
 
     @Test
     fun whenPortionIsHalfThenPercentageIs50Percent() {
-        val text = testee.networkPercentage(NetworkTally("", 10), 20)
+        val text = testee.networkPercentage(NetworkLeaderboardEntry("", 10), 20)
         assertEquals("50%", text)
     }
 
--- a/app/src/androidTest/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModelTest.kt
@@ -49,7 +49,7 @@ class PrivacyDashboardViewModelTest {
     private var settingStore: PrivacySettingsStore = mock()
     private var networkLeaderboard: NetworkLeaderboardDao = mock()
     private var networkTallyLiveData: LiveData<List<NetworkTally>> = mock()
-    private var domainsVisitedLiveData: LiveData<Int> = mock()
+    private var sitesVisitedLiveData: LiveData<Int> = mock()
     private var mockPixel: Pixel = mock()
 
     private val testee: PrivacyDashboardViewModel by lazy {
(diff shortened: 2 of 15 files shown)


subsymbolic:
@CDRussell as per our conversation earlier, I'm merging this so we can include it in the beta. I would still like a retrospective review.

app/src/androidTest/java/com/duckduckgo/app/privacy/renderer/TrackersRendererTest.kt
```diff
@@ -65,26 +65,26 @@ class TrackersRendererTest {
     }
 
     @Test
-    fun whenTotalDomainsIsZeroThenPercentageIsBlank() {
-        val text = testee.networkPercentage(NetworkTally("", 10), 0)
+    fun whenSitestVisitedIsZeroThenPercentageIsBlank() {
```
nilnilnil: `Sitest`?

app/src/main/java/com/duckduckgo/app/privacy/ui/TrackerNetworkLeaderboardPillView.kt
```diff
@@ -45,11 +45,11 @@ class TrackerNetworkLeaderboardPillView : FrameLayout {
         View.inflate(context, R.layout.view_network_tracker_pill, this)
     }
 
-    fun render(tally: NetworkTally?, totalDomains: Int) {
-        tally ?: return
-        icon.setImageResource(renderer.networkPillIcon(context, tally.networkName) ?: R.drawable.network_pill_generic)
-        val percentText = renderer.networkPercentage(tally, totalDomains)
-        icon.contentDescription = "${tally.networkName} $percentText"
+    fun render(netowrkEntity: NetworkLeaderboardEntry?, totalSitesVisited: Int) {
```
CDRussell: typo: `netowrkEntity`

app/src/main/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModel.kt
```diff
@@ -69,37 +69,37 @@ class PrivacyDashboardViewModel(
...
-            trackerNetworkTally = networkTally
+            showTrackerNetworkLeaderboard = showTrackerNetworkLeaderboard(domainCount, networkEntries.count()),
+            trackerNetworkEntries = networkEntries
         )
     }
 
-    private fun showTrackerNetworkLeaderboard(domainCount: Int, networkCount: Int): Boolean {
-        return domainCount > LEADERNOARD_MIN_DOMAINS_EXCLUSIVE && networkCount >= LEADERBOARD_MIN_NETWORKS
+    private fun showTrackerNetworkLeaderboard(siteVisitedCount: Int, networkCount: Int): Boolean {
+        return siteVisitedCount > LEADERBOARD_MIN_DOMAINS_EXCLUSIVE && networkCount >= LEADERBOARD_MIN_NETWORKS
```
CDRussell: It adds complexity that one of these uses `>` and the other uses `>=`. Since they aren't used anywhere else, I'm guessing we can just use the same operator for both, and change the constant accordingly.

Again, I realise this code was already like that, though. Just highlighting now since my attention was drawn to it.

CDRussell:
Minor code comments only; tested and all working as expected.

app/src/main/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModel.kt
subsymbolic: It expresses the original spec which was that this would show:
-  after 30 sites
and
- when we have at least 3 networks

I think leaving it expresses that better.

subsymbolic:
Many thanks for the review @CDRussell  🙌
I'll send new PR for the typos!! 🤪

app/src/main/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModel.kt
CDRussell: Then the spec isn't written perfectly clear. 

> 30
> 2

>=31
>=3

Either of these is preferable to mixing and matching. Anyway, not a suggestion that it needs immediately fixed, but we could make the code easier to read.

subsymbolic: I think it is readable in that we want 30 sites and there are 3 leaderboard slots in the UI. While it is generally nicer to be consistent switching to either >=31 or >2 just doesn't capture the way we would (and did) express this in natural language.

Deferring this discussion sounds good though, especially as the change wasn't introduced in this PR. Might poll the team on Monday to see if other people find it confusing too. Will be a good discussion.

CDRussell: And while I'm at it... 

 `showTrackerNetworkLeaderboard` would be better named `shouldShowTrackerNetworkLeaderboard` 

Ok I'm done now

subsymbolic: That's not part of this PR but it's a good suggestion, I'll update.

app/src/androidTest/java/com/duckduckgo/app/privacy/ui/PrivacyDashboardViewModelTest.kt
```diff
@@ -151,50 +151,50 @@ class PrivacyDashboardViewModelTest {
     }
 
     @Test
-    fun whenNetworkCountIsAtLeastThreeAndTotalDomainsIsOverThirtyThenShowSummaryIsTrue() {
-        val first = NetworkTally("Network1", 5)
-        val second = NetworkTally("Network2", 3)
-        val third = NetworkTally("Network3", 3)
-        testee.onTrackerNetworkTallyChanged(listOf(first, second, third))
-        testee.onDomainsVisitedChanged(31)
+    fun whenNetworkCountIsAtLeastThreeAndTotalSitesIsOverThirtyThenShowSummaryIsTrue() {
```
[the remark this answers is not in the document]
subsymbolic: The tests could be named better. Regarding boundaries here they are 2 and 3 and are covered by existing tests so off by one failures shouldn't be possible
- 2 is covered by whenNetworkCountIsLessThanThreeThenShowSummaryIsFalse
- 3 is covered by whenNetworkCountIsAtLeastThreeAndTotalSitesIsOverThirtyThenShowSummaryIsTrue
- 4 does not need to be covered, it is not part of a boundary

That said there are two variables under test (network count and site count) so we should have 4 tests to cover the combination of boundaries. I've updated this as well as the names. Not strictly part of this PR but a good suggestion.

PR #513, opened 2019-06-17
Feature/webview file deletion on data clear

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1126987722969423
Tech Design URL: 
CC: 

**Description**:
Performs cleanup on remaining files after the data has been cleared using the 🔥 button

**Steps to test this PR**:
ℹ️ Can only replicate the non-empty `Cookie-journal` file on `Marshmallow` and `Lollipop`, so ensure you use one of these emulators for your testing.

on **develop**, run these steps to ensure you can replicate the problem (to confirm the problem is definitely fixed):

1. Perform a clean install
1. Browse to a site which uses cookies (e.g., cnn.com, and accept cookies)
1. Delete the data using the 🔥 button
1. Pull the webview data directory 
    - `adb pull data/data/com.duckduckgo.mobile.android/app_webview`
1. Inspect the `Cookies-journal` file; verify you can see cnn.com cookies

now, checkout this branch, and repeat the steps above. this time, on the final step above, verify that the `Cookies-journal` no longer exists (it will be re-created when the `WebView` next accepts a cookie)

Additionally, check the general cookie handling works as you'd expect:
1. DDG SERP settings are not cleared when using the 🔥 functionality
1. Third party sites still have their cookies cleaned as expected

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/WebDataManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/WebDataManager.kt
@@ -23,6 +23,7 @@ import android.webkit.WebView
 import android.webkit.WebViewDatabase
 import com.duckduckgo.app.browser.session.WebViewSessionStorage
 import com.duckduckgo.app.fire.DuckDuckGoCookieManager
+import com.duckduckgo.app.global.file.FileDeleter
 import java.io.File
 import javax.inject.Inject
 
@@ -36,7 +37,8 @@ interface WebDataManager {
 class WebViewDataManager @Inject constructor(
     private val context: Context,
     private val webViewSessionStorage: WebViewSessionStorage,
-    private val cookieManager: DuckDuckGoCookieManager
+    private val cookieManager: DuckDuckGoCookieManager,
+    private val fileDeleter: FileDeleter
 ) : WebDataManager {
 
     override fun clearData(webView: WebView, webStorage: WebStorage, webViewDatabase: WebViewDatabase) {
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/duckduckgo/app/browser/WebDataManager.kt
```diff
@@ -64,6 +70,11 @@ class WebViewDataManager @Inject constructor(
         webView.clearFormData()
     }
 
+    override suspend fun deleteWebViewDirectory() {
```
subsymbolic: I find it confusing that this method is called `clearWebViewDirectory` but only clears _some_ of the web view directory. Furthermore it has knowledge of other clearing behavior e.g it knows it doesn't need to clear the cookies and storage because other things clear those.

It could be renamed to better capture what it is doing or take the excluded items as a parameter. If the former, it should also be private as it would still contain knowledge of an "overall clearing strategy" and so shouldn't be part of the public API (that should be the top-level method that does the overall clearing).

CDRussell: Good points, and 👍 for the pragmatic suggestion to not block on it.

Can discuss it more here: https://app.asana.com/0/414730916066338/1127379946457993

CDRussell: Given that, is this PR approved to go?

PR #515, opened 2019-06-19
Feature/restore partial network caching

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1127752918556318
Tech Design URL: 
CC: 

**Description**:

1. Retains some network caching after data clearing; only non-sensitive data can be cached.
1. For sensitive data (like autocomplete results) we will disable caching entirely.
1. Moves the `okHttp` networking cache to a new location (still in `cache` dir, but now in a named directory)

As a result, when the data is cleared, we can delete the `cache` dir, but exclude the `okHttpCache` directory from this deletion. The result is that (non-sensitive) configuration data will be kept after clearing the app data.

**Steps to test this PR**:
1. From a clean install, verify the new `okHttpCache` directory is used
1. Start typing a search (to trigger a few autocomplete results); verify that this **does not** result in new cache files being created
1. Perform some browsing
1. Clear the app data 🔥 
1. Verify that the contents of the `okHttpCache` aren't modified / deleted etc...
1. Verify that the app didn't download full datasets again from the server after clearing the data (using a proxy)

In code, verify that 
1. the pixel service uses the non-caching `okHttp` client.
1. the autocomplete service uses the non-caching `okHttp` client

More generally, check the caching still works as expected, and that the data left on the device after clearing the data is expectedly minimal, and that it contains no sensitive data at all.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/di/NetworkModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/di/NetworkModule.kt
@@ -28,6 +28,7 @@ import com.duckduckgo.app.feedback.api.FireAndForgetFeedbackSubmitter
 import com.duckduckgo.app.feedback.api.SubReasonApiMapper
 import com.duckduckgo.app.global.AppUrl.Url
 import com.duckduckgo.app.global.api.ApiRequestInterceptor
+import com.duckduckgo.app.global.api.NetworkApiCache
 import com.duckduckgo.app.global.job.JobBuilder
 import com.duckduckgo.app.httpsupgrade.api.HttpsUpgradeService
 import com.duckduckgo.app.job.AppConfigurationSyncer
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/duckduckgo/app/di/NetworkModule.kt
```diff
@@ -89,12 +92,13 @@ class NetworkModule {
 
     @Provides
     @Singleton
-    @Named("pixel")
-    fun pixelRetrofit(@Named("pixel") okHttpClient: OkHttpClient): Retrofit {
+    @Named("nonCaching")
+    fun nonCachingRetrofit(@Named("nonCaching") okHttpClient: OkHttpClient, moshi: Moshi): Retrofit {
         return Retrofit.Builder()
-            .baseUrl(Url.PIXEL)
+            .baseUrl(Url.API)
```
CDRussell: ℹ️ `baseUrl` is a required parameter. I've used `Url.API` as hopefully a sensible default, but each service can (and does) override this by providing the full URL it needs.

PR #530, opened 2019-07-16
Lint layout tidy-up

ibrahimAlii:
- Replace `FrameLayout` with `merge` which is more faster.

- Remove Redundant `namespace`


subsymbolic:
Thanks!

PR #531, opened 2019-07-18
Improve URL handling so we display the correct url at the correct time

subsymbolic:
Task/Issue URL: 
- https://app.asana.com/0/361428290920652/1125745835404857
- https://app.asana.com/0/361428290920652/1129928100535218
- https://app.asana.com/0/361428290920652/1132109259549317

Tech Design URL: https://app.asana.com/0/361428290920652/1125745835404855

**Description**:
- Improve URL handling so we display the correct url at the correct time, this improves user experience and protect against phishing attacks.
- Also fix an issue where omnibar text from previous url could show on home screen
- Make extraction of website title hardier so that it is displayed correctly in tabs

**Steps to test this PR**:

**Check that url, privacy dash and back/forward buttons update as expected during navigation:**
1. Visit facebook.com and note the url updates in the omnibar and the privacy dash is refreshed
1. Visit reddit.com and note the url updates in the omnibar and the privacy dash is refreshed
1. Tap a subreddit link and note the url updates again (this previously did not happen)
1. Now press back all the way to the home screen to make sure navigation works as expected
1. Ensure the omnibar displays no url when on home
1. Now use the forward multiple times to return to the subreddit page again

**Check against historical phishing attack:**
1. Visit https://soda.fm/spoof1.html
1. Ensure that the url does not change until off the spoofed page

**Check against latest phishing attack:**
1. Visit https://geekboy.ninja/p0c/spoof-bar.html
1. Ensure that the url does not change

**Check refresh works as it did historically**
1. Visit cnn.com
1. Note the number of trackers in the privacy dashboard
1. Select refresh to refresh the page and open the privacy dashboard again
1. Note that the number of trackers resets and repopulates
NOTE: Refresh relies on onPageStarted as it used to. Fixing this was beyond the scope of this project however I have created https://app.asana.com/0/361428290920652/1131849703601622 to cover it at a later date

**Check we only make fire one atb request when navigating to our serp**
1. Open Charles to watch app traffic
1. Perform a search to visit our serp
1. Ensure that the atb call only fires once

**Check other urls**
- Tabs
- Bookmarks
- Sharing
- Pin to home

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/bookmarks/ui/SaveBookmarkDialogFragment.kt
```diff
@@ -88,8 +88,8 @@ class SaveBookmarkDialogFragment : DialogFragment() {
     }
 
     private fun getExistingId(): Int? = arguments!!.getInt(KEY_BOOKMARK_ID)
-    private fun getExistingTitle(): String = arguments!!.getString(KEY_PREEXISTING_TITLE)
-    private fun getExistingUrl(): String = arguments!!.getString(KEY_PREEXISTING_URL)
+    private fun getExistingTitle(): String? = arguments!!.getString(KEY_PREEXISTING_TITLE)
+    private fun getExistingUrl(): String? = arguments!!.getString(KEY_PREEXISTING_URL)
```
subsymbolic: Updated to optional to match the `createDialogCreationMode` method which accepts optionals. This needs to match or a crash can occur.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -259,23 +259,13 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
             onMenuItemClicked(view.refreshPopupMenuItem) { refresh() }
             onMenuItemClicked(view.newTabPopupMenuItem) { browserActivity?.launchNewTab() }
             onMenuItemClicked(view.bookmarksPopupMenuItem) { browserActivity?.launchBookmarks() }
-            onMenuItemClicked(view.addBookmarksPopupMenuItem) { addBookmark() }
-            onMenuItemClicked(view.findInPageMenuItem) { viewModel.userRequestingToFindInPage() }
+            onMenuItemClicked(view.addBookmarksPopupMenuItem) { viewModel.onAddBookmarkSelected() }
+            onMenuItemClicked(view.findInPageMenuItem) { viewModel.onFindInPageSelected() }
             onMenuItemClicked(view.brokenSitePopupMenuItem) { viewModel.onBrokenSiteSelected() }
             onMenuItemClicked(view.settingsPopupMenuItem) { browserActivity?.launchSettings() }
-            onMenuItemClicked(view.requestDesktopSiteCheckMenuItem) {
```
subsymbolic: These webview.url calls all go via the view model instead now.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -513,11 +505,10 @@ class BrowserTabViewModel(
     private fun updateNetworkLeaderboard(event: TrackingEvent) {
         val networkName = event.trackerNetwork?.name ?: return
         networkLeaderboardDao.incrementNetworkCount(networkName)
-        networkLeaderboardDao.incrementSitesVisited()
```
subsymbolic: Removed, this is legacy code from before tour use of counts and so no longer makes sense. We only want to increment this when a new page is detected, which we already do elsewhere.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -97,36 +96,28 @@ class BrowserWebViewClient(
...
-
-        Timber.i("Page Load Time: ${System.currentTimeMillis() - tempTimer}ms for $url")
+        webViewClientListener?.navigationStateChanged(WebViewNavigationState(webView.copyBackForwardList()))
     }
 
     @WorkerThread
     override fun shouldInterceptRequest(webView: WebView, request: WebResourceRequest): WebResourceResponse? {
-        Timber.v("Intercepting resource ${request.url} on page $currentUrl")
-        return runBlocking { requestInterceptor.shouldIntercept(request, webView, currentUrl, webViewClientListener) }
+        val documentUrl = runBlocking { withContext(Dispatchers.Main) { webView.url } }
```
subsymbolic: The blocking call here is a bit icky, but a necessary evil as we can't use the currentUrl from onPageStarted. Happy to discuss alternatives.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -259,23 +259,13 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
-                viewModel.desktopSiteModeToggled(
-                    urlString = webView?.url,
```
subsymbolic: The webview.url calls in this Activity have been removed, we now go via the view model

CDRussell:
⚠️PR conflicts exist now @subsymbolic

CDRussell:
There is a problem with the "forward navigation".

**Steps to reproduce**
1. Enter `twitter.com` in the search bar and enter
2. Click the `Sign up` button
3. Click back (either on screen back button or device back button)
4. Try to navigate `forwards`. 👎it's greyed out

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -37,9 +37,8 @@ class BrowserWebViewClient(
 ) : WebViewClient() {
 
     var webViewClientListener: WebViewClientListener? = null
+    private var lastPageStarted: String? = null
 
-    private var currentUrl: String? = null
-    private var tempTimer = 0L
```
CDRussell: 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
CDRussell: Why is there two `runBlocking`? couldn't it all be done within the same `runBlocking` block? e.g., 👇

    @WorkerThread
    override fun shouldInterceptRequest(webView: WebView, request: WebResourceRequest): WebResourceResponse? {
        return runBlocking {
            
            val documentUrl = withContext(Dispatchers.Main) { 
                webView.url 
            }
            
            Timber.v("Intercepting resource ${request.url} on page $documentUrl")
            requestInterceptor.shouldIntercept(request, webView, documentUrl, webViewClientListener)
        }
    }

app/src/main/java/com/duckduckgo/app/browser/WebNavigationState.kt
```diff
@@ -0,0 +1,122 @@
...
+    val originalUrl: String?
+    val currentUrl: String?
+    val title: String?
+    val stepsToPreviousPage: Int
+    val canGoBack: Boolean
+    val canGoForward: Boolean
+    val hasNavigationHistory: Boolean
+}
+
+sealed class WebNavigationStateChange
```
CDRussell: Elsewhere, we tend to _block_ the subclasses of a sealed class using `{ }`. i think it makes it more readable. e.g,

    sealed class WebNavigationStateChange {
        data class NewPage(val url: String, val title: String?) : WebNavigationStateChange()
        data class UrlUpdated(val url: String) : WebNavigationStateChange()
        object PageCleared : WebNavigationStateChange()
        object Unchanged : WebNavigationStateChange()
        object Other : WebNavigationStateChange()
    }

app/src/main/java/com/duckduckgo/app/browser/WebNavigationState.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    override val stepsToPreviousPage: Int = if (stack.isHttpsUpgrade) 2 else 1
+
+    override val canGoBack: Boolean = stack.currentIndex >= stepsToPreviousPage
+
+    override val canGoForward: Boolean = stack.currentIndex + 1 < stack.size
+
+    override val hasNavigationHistory = stack.size != 0
+
+    override fun equals(other: Any?): Boolean {
```
CDRussell: Why do we need the overrides for `equals` and `hashCode`? can you add a comment explaining the reason if it's needed?

app/src/main/java/com/duckduckgo/app/global/model/Site.kt
```diff
@@ -26,24 +28,38 @@ import com.duckduckgo.app.trackerdetection.model.TrackingEvent
...
 interface Site {
 
-    val url: String
-    val privacyPractices: PrivacyPractices.Practices
-    val memberNetwork: TrackerNetwork?
+    /*
+     * The current url for this site. This is sometimes different to the url originally
+     * loaded as the url may change while loading a site
+     */
+    var url: String
```
CDRussell: do we need both `Site.uri` and `Site.url`? Can these differ from each other, or will they always be the same underlying address?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -192,18 +193,19 @@ class BrowserTabViewModel(
         }
     }
 
-    override val url: String?
+    val url: String?
         get() = site?.url
 
-    private var pendingUrl: String? = null
+   val title: String?
```
CDRussell: formatting

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -387,45 +376,97 @@ class BrowserTabViewModel(
...
-    override fun loadingStarted(url: String?) {
-        Timber.v("Loading started")
-        if (!currentBrowserViewState().browserShowing) return
-        val progress = currentLoadingViewState()
-        loadingViewState.value = progress.copy(isLoading = true)
-        pendingUrl = url
-        site = null
-        onSiteChanged()
-    }
+    override fun navigationStateChanged(newWebNavigationState: WebNavigationState) {
```
CDRussell: lint warning that param name doesn't match the supertype and may cause problems

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -387,45 +376,97 @@ class BrowserTabViewModel(
...
         if (!currentBrowserViewState().browserShowing) return
 
         browserViewState.value = currentBrowserViewState().copy(
-            canGoBack = navigation.canGoBack || !skipHome,
-            canGoForward = navigation.canGoForward
+            canGoBack = newWebNavigationState.canGoBack || !skipHome,
+            canGoForward = newWebNavigationState.canGoForward
         )
+
+        when {
```
CDRussell: bump `stateChange` up into `when` clause? e.g., 

    when (stateChange) {
        is NewPage -> pageChanged(stateChange.url, stateChange.title)
        is PageCleared -> pageCleared()
        is UrlUpdated -> urlUpdated(stateChange.url)
    }

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -434,8 +475,8 @@ class BrowserTabViewModel(
         }
     }
 
-    override fun titleReceived(title: String) {
-        site?.title = (title)
+    override fun titleReceived(newTitle: String) {
```
CDRussell: lint warning that param name doesn't match the supertype and may cause problems

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
subsymbolic: Oh, I really need to turn this lint option on!

app/src/main/java/com/duckduckgo/app/browser/WebNavigationState.kt
subsymbolic: I think that looks nicer too 👍

app/src/main/java/com/duckduckgo/app/browser/WebNavigationState.kt
subsymbolic: Normally the system provided equals is perfect for data objects but the private stack property (which we don't care about) messes that up here. I have added comments

app/src/main/java/com/duckduckgo/app/global/model/Site.kt
subsymbolic: We use both so I have included both for convenience as it seemed messy to build the Uri each time it was needed (and a bit inefficient). They will always be in sync as the url setter updates the uri and the interface for uri doesn't allow setting. That said, I have added a private backing field to make it safer again.

subsymbolic:
Thanks for a great PR review @CDRussell!

PR #532, opened 2019-07-18
Update gemfile.lock as Github is advising about a vulnerability

CDRussell:
Task/Issue URL: https://app.asana.com/0/608920331025315/1131903915391988
Tech Design URL: 
CC: 

**Description**:
This is likely only used by `Fastlane`, and even so, it's likely a feature of `Fastlane` we don't use so i'm not too worried about this.

However, I think just updating the version of `mini_magick` in the lock file should be enough. (in grabbing the latest version of `Fastlane`, it also updated some other libs)

---
###### Internal references:


subsymbolic:
Fastlane was indeed the cause. Thanks for updating this 👍

PR #546, opened 2019-08-07
Change request interceptor to more safely upgrade https requests

CDRussell:
Task/Issue URL: https://app.asana.com/0/608920331025315/1134654612159126
Tech Design URL: 
CC: 

**Description**:
Change the way we load HTTPS versions of site during HTTPS-upgrades to avoid a `WebView` crash

**Steps to test this PR**:
The easiest way to test this out is to hack `BrowserTabFragment.launchPopupMenu` as follows

    private fun launchPopupMenu() {
        //popupMenu.show(rootView, toolbar)
        webView?.loadUrl("http://dealnews.com")
    }

Now, you can tap the overflow menu to attempt to load that site. Within a few taps of that (or certainly less than 10) on my Pixel 2, the app crashes.

1. Test the hack above against `develop`, and verify it crashes
1. Test the hack above against this branch, and verify it does not crash
1. Verify all the interceptor logic still works as before (still blocks trackers, still upgrades http sites etc...)

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -22,6 +22,7 @@ import android.os.Build
 import android.webkit.*
 import androidx.annotation.UiThread
 import androidx.annotation.WorkerThread
+import com.duckduckgo.app.browser.WebViewRequestInterceptor.InterceptAction
 import com.duckduckgo.app.browser.model.BasicAuthenticationRequest
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
```diff
@@ -63,21 +62,19 @@ class WebViewRequestInterceptor(
...
         documentUrl: String?,
         webViewClientListener: WebViewClientListener?
-    ): WebResourceResponse? {
+    ): InterceptAction {
 
         val url = request.url
 
         if (shouldUpgrade(request)) {
-            val newUri = httpsUpgrader.upgrade(url)
-            webView.post { webView.loadUrl(newUri.toString()) }
```
subsymbolic: This is a suspend function. Am I right in thinking that changing this to `withContext(Dispatchers.Main)` would have solved the issue too? What's your motivation for the bigger footprint? A cleaner API? Testing? I have some questions around this PR (see other comments) so if you're keep to get the fix out we could go with the one-line fix and review the rest of it in a more relaxed manner.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -111,9 +112,27 @@ class BrowserWebViewClient(
...
+                withContext(Dispatchers.Main) {
+                    webView.loadUrl(interceptionAction.url.toString())
+                }
+            }
+
+            return@runBlocking convertActionToWebResourceResponse(interceptionAction)
+        }
+    }
+
+    private fun convertActionToWebResourceResponse(action: InterceptAction): WebResourceResponse? {
```
subsymbolic: This api doesn't read well for me - why does the InterceptAction sometime return the WebResourceResponse and not others? I feel like it should always provide it or never,

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -111,9 +112,27 @@ class BrowserWebViewClient(
     @WorkerThread
     override fun shouldInterceptRequest(webView: WebView, request: WebResourceRequest): WebResourceResponse? {
         return runBlocking {
+
             val documentUrl = withContext(Dispatchers.Main) { webView.url }
             Timber.v("Intercepting resource ${request.url} on page $documentUrl")
-            requestInterceptor.shouldIntercept(request, webView, documentUrl, webViewClientListener)
+            val interceptionAction = requestInterceptor.shouldIntercept(request, webView, documentUrl, webViewClientListener)
+
+            if (interceptionAction is InterceptAction.BlockAndLoadAlternativeUrl) {
```
subsymbolic: We have a when statement to process the return statement already, feels like a cognitive disconnect to also have an if statement for this special actions - I have to think about how we are handling `InterceptAction.BlockAndLoadAlternativeUrl` in two places. Why not combine it all in a when? That was we have one block of logic to handle each type rather than splitting it out in multiple places.

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
CDRussell: Good shout on the simpler approach to fixing this.

CDRussell:
I'm going to close this PR out in favour of #549 which fixes the core problem in a simpler way

PR #555, opened 2019-08-09
Add webview renderer gone diagnositc pixel

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1135024870354624

**Description**:
Adds a diagnostic pixel when the webview renderer disappear so we can determine if this is caused by a crash or memory pressure

**Steps to test this PR**:
This is a little tricky to test however one path can be tested by causing a webview crash.

1. Open Charles to watch app traffic
1. Update `BrowserTabFragment.navigate(url: String)` method to load a crashing url when "crash" entered into the omnibar
```
if(url.contains("crash")) {
} else {
   webView?.loadUrl(url)
}
```
3. Now open the app and type crash in the omnibar. The app will crash.
1. Launch the app to start a sync
1. Ensure the pixel is fired

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
@@ -18,14 +18,13 @@ package com.duckduckgo.app.browser
 
 import android.content.Context
 import android.webkit.HttpAuthHandler
+import android.webkit.RenderProcessGoneDetail
 import android.webkit.WebView
 import androidx.test.annotation.UiThreadTest
 import androidx.test.platform.app.InstrumentationRegistry
 import com.duckduckgo.app.browser.model.BasicAuthenticationRequest
-import com.nhaarman.mockitokotlin2.any
-import com.nhaarman.mockitokotlin2.mock
-import com.nhaarman.mockitokotlin2.never
-import com.nhaarman.mockitokotlin2.verify
+import com.duckduckgo.app.statistics.store.OfflinePixelDataStore
+import com.nhaarman.mockitokotlin2.*
 import org.junit.Before
 import org.junit.Test
 
@@ -38,6 +37,7 @@ class BrowserWebViewClientTest {
     private val specialUrlDetector: SpecialUrlDetector = mock()
     private val requestInterceptor: RequestInterceptor = mock()
     private val listener: WebViewClientListener = mock()
+    private val offlinePixelDataStore: OfflinePixelDataStore = mock()
 
     @UiThreadTest
     @Before
@@ -46,7 +46,8 @@ class BrowserWebViewClientTest {
         testee = BrowserWebViewClient(
             requestRewriter,
             specialUrlDetector,
-            requestInterceptor
+            requestInterceptor,
+            offlinePixelDataStore
         )
         testee.webViewClientListener = listener
     }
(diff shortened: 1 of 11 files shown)


app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
```diff
@@ -90,6 +92,26 @@ class BrowserWebViewClientTest {
         verify(listener).requiresAuthentication(authenticationRequest)
     }
 
+    @Test
+    fun whenRenderProcessGoneDueToCrashThenCrashDataStoreEntryIsIncremented() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
subsymbolic: I *thought* there was an annotation to specify the min test API but I couldn't find one so I went with this less elegant solution. Happy to update if you can think of it.

app/src/main/java/com/duckduckgo/app/statistics/api/OfflinePixelSender.kt
```diff
@@ -0,0 +1,58 @@
...
+ */
+class OfflinePixelSender @Inject constructor(
+    private val dataStore: OfflinePixelDataStore,
+    private val pixel: Pixel
+) {
+
+    fun sendWebRendererGonePixel(): Completable {
+        return defer {
+
+            if (dataStore.webRendererGoneCrashCount == 0 && dataStore.webRendererGoneOtherCount == 0) {
```
CDRussell: we call through to `dataStore` quite a lot in this function. it's probably fast enough, but it would be faster to read the values once into local variables and use them throughout the function.

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
CDRussell: `@SdkSuppress(minSdkVersion = Build.VERSION_CODES.O)` I think would do it

subsymbolic: Thanks the one!! Thanks :)

app/src/main/java/com/duckduckgo/app/di/StatisticsModule.kt
```diff
@@ -52,11 +50,14 @@ class StatisticsModule {
...
     }
 
-    @Provides
-    fun deviceInfo(context: Context): DeviceInfo = ContextDeviceInfo(context)
-
     @Provides
     fun pixel(pixelService: PixelService, statisticsDataStore: StatisticsDataStore, variantManager: VariantManager, deviceInfo: DeviceInfo): Pixel =
         ApiBasedPixel(pixelService, statisticsDataStore, variantManager, deviceInfo)
 
+    @Provides
```
CDRussell: should this be `singleton`?

app/src/main/java/com/duckduckgo/app/statistics/api/OfflinePixelSender.kt
```diff
@@ -0,0 +1,60 @@
...
+ * Most pixels are "send and forget" however we sometimes need to guarantee that a pixel will be sent.
+ * In those cases we schedule them to happen as part of our app data sync.
+ */
+class OfflinePixelSender @Inject constructor(
+    private val dataStore: OfflinePixelDataStore,
+    private val pixel: Pixel
+) {
+
+    fun sendWebRendererGonePixel(): Completable {
+        return defer {
```
CDRussell: We use `Completable.fromAction` in most of the other services. is there a particular reason for using `defer` here? (i'm not sure it matters, but just wonder if there was special intent behind it)

PR #560, opened 2019-08-15
Update crash pixels and add work scheduler to send these offline

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1135707350642503

**Description**:
- Add a high level crash pixel to allow us to monitor overall crash numbers
- Send pixels via regular scheduled job, to protect against case where app is too unstable to run.
- Include app version globally in pixels.
- Update existing web render pixels to be separate entities to allow us to use standard Grafana dashboards (ClickHouse is too slow)

**Steps to test this PR**:

**Setup**
Open Charles to watch app traffic

**Test web pixel sent**
1. Update `BrowserTabFragment.navigate(url: String)` method to load a crashing url when "crash" entered into the omnibar
```
if(url.contains("crash")) {
} else {
   webView?.loadUrl(url)
}
```
2. Open the app and type "crash" in the omnibar. The app will crash.
1. Launch the app to start an immediate sync
1. Ensure the `m_d_wrg` pixel is fired, check that it includes the app version as a param

**Test app crash pixel sent**
1. Revert previous changes and update `BrowserTabFragment.navigate(url: String)` method to crash after a url is loaded
```
webView?.loadUrl(url)
throw Exception("")
```
1. Open the app and type in a search term e.g "test" in the omnibar. The app will crash.
1. Launch the app to start an immediate sync
1. Ensure the `m_d_ac` pixel is fired, check that it includes the app version as a param

**Test pixels still sent in background**
1. Set:
```
private const val SERVICE_INTERVAL = 30L
private val SERVICE_TIME_UNIT = TimeUnit.SECONDS
```
2. Run the app, and as before, type in a search term e.g "test" in the omnibar. The app will crash.
1. Wait a while for the service to kick in and ensure the `m_d_ac` pixel is fired, check that it includes the app version as a param

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
@@ -108,7 +108,7 @@ class BrowserWebViewClientTest {
         val detail: RenderProcessGoneDetail = mock()
         whenever(detail.didCrash()).thenReturn(false)
         testee.onRenderProcessGone(webView, detail)
-        verify(offlinePixelDataStore, times(1)).webRendererGoneOtherCount = 1
+        verify(offlinePixelDataStore, times(1)).webRendererGoneKilledCount = 1
     }
 
     private class TestWebView(context: Context) : WebView(context)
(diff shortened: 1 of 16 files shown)


subsymbolic:
Thanks so much for the late night review @CDRussell 🙏

PR #564, opened 2019-08-19
Update themes.xml

vargbeaumont:
Right now on certain ROMs like OxygenOS the navigation bar is solid white which doesn't match the apps theme, this should make the navigation bar on these devices match the apps dark theme while keeping it the same on Pixel ROMs.

### DIFF
--- a/app/src/main/res/values/themes.xml
+++ b/app/src/main/res/values/themes.xml
@@ -75,6 +75,7 @@
     <style name="AppTheme.Dark.Transparent">
         <item name="android:statusBarColor">@android:color/transparent</item>
         <item name="android:windowDrawsSystemBarBackgrounds">true</item>
+        <item name="android:navigationBarColor">@android:color/transparent</item>
     </style>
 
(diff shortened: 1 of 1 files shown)


subsymbolic:
@vargbeaumont can you upload a screenshot before and another after your changes so we can visualize the problem and the fix? Thanks!

app/src/main/res/values/themes.xml
```diff
@@ -75,6 +75,7 @@
     <style name="AppTheme.Dark.Transparent">
         <item name="android:statusBarColor">@android:color/transparent</item>
         <item name="android:windowDrawsSystemBarBackgrounds">true</item>
+        <item name="android:navigationBarColor">@android:color/transparent</item>
```
subsymbolic: This only updates the initial Onboarding screen, is that what you wanted?

subsymbolic:
Closing this as it doesn't work as intended. Thanks for submitting this @vargbeaumont, feel free to resubmit when you had addressed my earlier comments.

Issue #559, opened 2019-08-15
Help us translate the app


CDRussell:
## New languages
We will be adding additional languages to the app and we're needing help in improving our translations. 

For now, we're only looking for improvements to the list of languages identified and are not looking to add completely new languages yet. We may request help on doing this in future, though.

Do you know any of the following languages, and want to help us improve? If so, please see https://github.com/duckduckgo/Android/blob/develop/TRANSLATIONS.md for instructions on how you can help us improve the app.

### List of languages

- Bulgarian (bg)
- Croatian (hr)
- Czech (cs)
- Danish (da)
- Dutch (nl)
- Finnish (fi)
- French (fr)
- German (de)
- Greek (el)
- Hungarian (hu)
- Italian (it)
- Lithuanian (lt)
- Norwegian bokmål (nb)
- Polish (pl)
- Portuguese (pt)
- Romanian (ro)
- Russian (ru)
- Slovak (sk)
- Slovenian (sl)
- Spanish (es)
- Swedish (sv)

coffeeplanter:
Hi, I've seen the current Russian translation. It's pretty accurate and tidy, in my opinion. But I'll make a pull-request.
And... why to translate string resources marked as non translatable? )

CDRussell:
non translatable should indicate things like the company name, where DuckDuckGo doesn't need to be translated. however, it's possible some things have been incorrectly marked as non-translatable that should be translated, or things are currently translated that shouldn't be.

A tidy-up PR would be awesome! 🙏

renanferrari:
> not looking to add completely new languages yet

Would it be ok to add language variants, since it may not be considered a "completely new language"?

I'm asking this because I believe **Brazilian Portuguese** would be a great addition in the list, based on the fact that it is – by far – the most widely spoken Portuguese variant in the world, while **European Portuguese** is only the fourth:

Rank | Country | Number of Portuguese Speakers (Est.)
-- | -- | --
**1** | **Brazil** | **204,000,000**
2 | Angola | 20,455,100
3 | Mozambique | 14,400,000
**4** | **Portugal** | **10,320,000**
5 | Guinea-Bissau | 272,400
6 | Cape Verde | 215,000
7 | Sao Tome and Principe | 199,910
8 | Macao | 6,800

[Source](https://www.worldatlas.com/articles/portuguese-speaking-countries.html)

[...]

PR #575, opened 2019-08-26
Improved and cleaned up the string resources, especially Russian translation

coffeeplanter:
Task/Issue URL:  #559 

**Description**:
- Removed non-translatable strings from all translations except default string resources.
- Changed Play Store to the correct official name: Google Play.
- Slightly corrected Russian translation.
- Slightly cleaned up the code in the resources.

**Steps to test this PR**:
1. Check the changes in the source code.
3. Run the app.
4. Check the correct placement of the changed labels.

---
Regards,
Ilya. )

### DIFF
--- a/app/src/main/res/values-bg/strings.xml
+++ b/app/src/main/res/values-bg/strings.xml
@@ -343,7 +343,7 @@
     <string name="sharePositiveFeedbackWithTheTeam">Има ли някакви подробности, които искате да споделите с екипа?</string>
     <string name="whatHaveYouBeenEnjoying">Кое Ви беше най-приятно?</string>
     <string name="awesomeToHear">Страхотно е да чуем това!</string>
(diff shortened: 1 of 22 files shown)


CDRussell:
Hi @coffeeplanter, thank you for the PR! I like the clean up of removing non-translatable strings, however I have some feedback on some of the changes:

- Thanks for cleaning up the non-translatable strings from all non-English strings files
- Thank you for the Russian change suggestions
- There is a small merge conflict with the Russian strings file that is needing resolved
- I'm not sure on the change from "Play Store" to "Google Play". While you are right that Google have updated the official name of the app store (again), the strings using Play Store were user-tested and performed well and I don't want to introduce a change here without considering the impact of the change. I'd rather this change wasn't included in this PR and considered separately.

coffeeplanter:
Done.

CDRussell:
Thank you so much for the contributions @coffeeplanter! 🙏

PR #593, opened 2019-09-12
Filter experimental variants

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1138059002087732

Tech Design: https://app.asana.com/0/481882893211075/1139758288255945

**Description**:

Adds a new filterBy lambda parameter to Variant so each of them can specify conditions for them to be allocated or not.

**Steps to test this PR**:
**Doesn't filter when filter returning true added**
1. Add a variant using `filterBy = { true }`
1. Install the app notice how that variant does not get filtered out after being allocated.

**Doesn't filter when no filter added**
1. Add a variant without creating a filter
1. Install the app notice how that variant does not get filtered out after being allocated.

**Filters correctly when filter returning false added**
1. Add a variant using `filterBy = { false }`
1. Install the app notice how that variant gets filtered out after being allocated and is replaced by the default variant instead.
---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
@@ -42,6 +42,7 @@ interface VariantManager {
             Variant(key = "sc", weight = 0.0, features = emptyList()),
             Variant(key = "se", weight = 0.0, features = emptyList()),
 
+            // All groups in an experiment (control and variants) MUST use the same filters
             Variant(key = "mw", weight = 1.0, features = emptyList()),
             Variant(key = "mx", weight = 1.0, features = listOf(VariantFeature.TabSwitcherGrid))
         )
--- a/app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
@@ -16,10 +16,12 @@
 
 package com.duckduckgo.app.statistics
 
+import android.os.Build
 import androidx.annotation.WorkerThread
 import com.duckduckgo.app.statistics.VariantManager.Companion.DEFAULT_VARIANT
 import com.duckduckgo.app.statistics.store.StatisticsDataStore
 import timber.log.Timber
+import java.util.Locale
 
 @WorkerThread
 interface VariantManager {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
```diff
@@ -90,14 +92,24 @@ class ExperimentationVariantManager(
     }
 
     private fun generateVariant(activeVariants: List<Variant>): Variant {
-        val weightSum = activeVariants.sumByDouble { it.weight }
+        val weightSum = filterByEnglishLocale(filterByApi(activeVariants)).sumByDouble { it.weight }
```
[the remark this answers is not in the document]
marcosholgado: As spoken earlier this was the wrong approach plus it wasn't even doing what it was supposed to. This is now gone and changed the approach as per the tech design 😄

app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
```diff
@@ -41,9 +42,15 @@ interface VariantManager {
...
             Variant(key = "se", weight = 0.0, features = emptyList()),
 
+            // All groups in an experiment (control and variants) MUST use the same filters
             Variant(key = "mw", weight = 1.0, features = emptyList()),
             Variant(key = "mx", weight = 1.0, features = listOf(VariantFeature.TabSwitcherGrid))
         )
+
+        private fun isEnglishLocale(): Boolean {
+            val locale = Locale.getDefault()
+            return locale != null && locale.language == "en"
```
subsymbolic: Can we use the `DeviceInfo.getLanguage()` method here?

marcosholgado: Doing this will complicate things a bit since it needs injecting and is currently needed in an interface where we can't really inject.

app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
```diff
@@ -33,18 +33,18 @@ interface VariantManager {
...
-        val DEFAULT_VARIANT = Variant(key = "", features = emptyList())
+        val DEFAULT_VARIANT = Variant(key = "", features = emptyList(), filterBy = { true })
 
         val ACTIVE_VARIANTS = listOf(
 
             // SERP variants. "sc" may also be used as a shared control for mobile experiments in
             // the future if we can filter by app version
-            Variant(key = "sc", weight = 0.0, features = emptyList()),
-            Variant(key = "se", weight = 0.0, features = emptyList()),
+            Variant(key = "sc", weight = 0.0, features = emptyList(), filterBy = { true }),
```
subsymbolic: Let's use a common filter lambda for clarity and readability, as discussed in the tech design:
`Variant(key = "sc", weight = 0.0, features = emptyList(), filterBy = { noFilter() }),`

marcosholgado: Done!

PR #596, opened 2019-09-15
Improve tracker and entity lookup performance

derek-cheung:
**Description**:
Whilst looking through the tracker detection code, I spotted an optimization that can be made when searching through both the entity and tracker databases. The current implementation searches the database with the parent host domain then recursively strips the subdomains off one by one. Each time a subdomain is stripped away, the result goes through a database search again.

Instead of doing a recursive search, we can generate the list of all domains to search immediately and have each entry in the database check against that list. This reduces it down to a single pass of the database entries no matter how many subdomains there are.

**Results**
I ran some tests with the existing code and the new code in order to compare the performance differences. I took 3 different sites, 1 with a low number of trackers (2), one with a medium number of trackers (15) and one with a high number of trackers (25). For each of these I measured the time it took to execute the `trackerDetector.evaluate` call in the `WebViewRequestInterceptor.shouldBlock` method. I ran each of these sites 7 times in a row and calculated a running average of the execution times. For the low number of trackers, I took the 10th result and for the other two I took the 100th result.

Low Trackers
Average time saved per execution = 2.58ms
Average % increase in performance = 4%
Total time saved per site load = 5.16ms

Medium Trackers
Average time saved per execution = 13.48ms
Average % increase in performance = 18.79%
Total time saved per site load = 202.14ms

High Trackers
Average time saved per execution = 17.89ms
Average % increase in performance = 28.37%
Total time saved per site load = 447.2ms

The results will vary even on sites with the same number of trackers. Things like varying average subdomain lengths of tracker urls and how deep the search has to go to find particular matches will impact times. But just from these results, the performance boost from making this change is pretty significant for sites with many trackers.

**Tests**
I didn't need to add any as existing tests already cover subdomain searching. In fact, I removed some tests that were specifically testing recursion that are no longer applicable.

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/entities/EntityMappingTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/entities/EntityMappingTest.kt
@@ -18,10 +18,7 @@ package com.duckduckgo.app.entities
 
 import com.duckduckgo.app.entities.db.EntityListDao
 import com.duckduckgo.app.entities.db.EntityListEntity
-import com.nhaarman.mockitokotlin2.mock
-import com.nhaarman.mockitokotlin2.never
-import com.nhaarman.mockitokotlin2.verify
-import com.nhaarman.mockitokotlin2.whenever
+import com.nhaarman.mockitokotlin2.*
 import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
 import org.junit.Before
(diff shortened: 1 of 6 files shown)


subsymbolic:
Thanks for submitting this @derek-cheung. We are currently rewriting our tracker detection module including entity-lookups however feel free to open an issue and/or PR against our new tracker detector, once it is out.

PR #597, opened 2019-09-18
Generate webview image when leaving BrowserActivity

CDRussell:
Task/Issue URL: https://app.asana.com/0/1126768040926570/1140520481588209
Tech Design URL: 
CC: 

**Description**:
This fixes an issue whereby a page might be loaded in a tab (TAB A) but the callback for page loaded is fired before anything is visible, producing a blank WebView preview. This would normally be fine if the user hit the tab switcher button. But if they instead leave the app, then they might return to the app later launching another tab (TAB B). From TAB B, they could launch the tab switcher. TAB B preview would be re-generated, but TAB A would remain white.

This PR ensures the preview is generated when the user is leaving the screen altogether, and this removes the need to generate a bitmap every time the page loads.

**Steps to test this PR**:
1. Launch a tab, and browse to a site; **do not** hit the tab switcher
1. Hit the home button
1. Launch the app into a new tab (e.g, use the widget, long press the app for the new tab shortcut etc...)
1. Hit the tab switcher button
1. Verify the original tab preview is populated
1. Generally browse around, open new tabs, and verify the webview preview logic holds

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -53,8 +53,7 @@ import androidx.core.view.isNotEmpty
 import androidx.core.view.isVisible
 import androidx.core.view.postDelayed
 import androidx.fragment.app.Fragment
-import androidx.lifecycle.Observer
-import androidx.lifecycle.ViewModelProviders
+import androidx.lifecycle.*
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.duckduckgo.app.autocomplete.api.AutoCompleteApi.AutoCompleteSuggestion
 import com.duckduckgo.app.bookmarks.ui.EditBookmarkDialogFragment
@@ -82,6 +81,7 @@ import com.duckduckgo.app.global.device.DeviceInfo
 import com.duckduckgo.app.global.view.*
 import com.duckduckgo.app.privacy.model.PrivacyGrade
 import com.duckduckgo.app.privacy.renderer.icon
+import com.duckduckgo.app.statistics.VariantManager
 import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.app.survey.model.Survey
 import com.duckduckgo.app.survey.ui.SurveyActivity
@@ -158,6 +158,9 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
 
     @Inject
     lateinit var previewPersister: WebViewPreviewPersister
+
+    @Inject
+    lateinit var variantManager: VariantManager
 
     val tabId get() = arguments!![TAB_ID_ARG] as String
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -245,6 +248,31 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         if (savedInstanceState == null) {
             viewModel.onViewReady()
         }
+
+        lifecycle.addObserver(object : LifecycleObserver {
+            @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+            fun onStop() {
```
marcosholgado: There is an issue if we only generate the preview image `onStop`. If you add several tabs using the overflow menu and go to the tab switcher, all the tabs created will have a blank preview image because `onStop` never gets called and, when it does, `viewModel.url` is null. But even if `viewModel.url` wasn't null it would only generate the preview for the latest tab. 

Without giving it much thought (haven't thought about the implications of doing it) I think you can potentially move this logic to `onHiddenChanged` to generate the preview every time the fragment is hidden.

CDRussell: Nice catch! I like that suggestion; will investigate this

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -452,23 +480,19 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
             is Command.LaunchLegacyAddWidget -> launchLegacyAddWidget()
             is Command.RequiresAuthentication -> showAuthenticationDialog(it.request)
             is Command.SaveCredentials -> saveBasicAuthCredentials(it.request, it.credentials)
-            is Command.GenerateWebViewPreviewImage -> generateWebViewPreviewImage(it.forceImmediate)
+            is Command.GenerateWebViewPreviewImage -> generateWebViewPreviewImage()
```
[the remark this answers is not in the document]
CDRussell: It's used again now!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
CDRussell: I've added the `onHiddenChanged` hook. We still need the `onStop` hook too, so both now co-exist.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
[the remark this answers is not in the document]
marcosholgado: I can't see where it's used, it is created and then used in the `when` statement but never assigned to command.value. Can you double check?

PR #601, opened 2019-10-06
Parse IP:PORT query as a URL

shamilatesoglu:
**Description**:
When a query with `IP:PORT` format entered in omnibar, the app performs search, instead of going to that address.

**Steps to test this PR**:
1. Run the app and enter an ip with port in `IP:PORT` format.
2. Ensure that the browser did not interpret that IP and port as a search query.


subsymbolic:
Great work, thanks for submitting this!

PR #624, opened 2019-11-07
Add devMode parameter onto PixelService requests

CDRussell:
Task/Issue URL: https://app.asana.com/0/inbox/608920331025313/1148562571541097/1148555304057513
Tech Design URL: 
CC: 

**Description**:
Adds `test=1` as query parameter to pixels

**Steps to test this PR**:
1. Send a pixel in `DEBUG` mode; verify `test=1` flag is added to request

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/statistics/api/PixelService.kt
+++ b/app/src/main/java/com/duckduckgo/app/statistics/api/PixelService.kt
@@ -16,6 +16,7 @@
 
 package com.duckduckgo.app.statistics.api
 
+import com.duckduckgo.app.browser.BuildConfig
 import com.duckduckgo.app.global.AppUrl
 import io.reactivex.Completable
 import retrofit2.http.GET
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/statistics/api/PixelService.kt
```diff
@@ -30,6 +31,7 @@ interface PixelService {
         @Path("pixelName") pixelName: String,
         @Path("formFactor") formFactor: String,
         @Query(AppUrl.ParamKey.ATB) atb: String,
-        @QueryMap additionalQueryParams: Map<String, String?> = emptyMap()
+        @QueryMap additionalQueryParams: Map<String, String?> = emptyMap(),
+        @Query(AppUrl.ParamKey.DEV_MODE) devMode: Int? = if (BuildConfig.DEBUG) 1 else null
```
subsymbolic: Would this add the `test` param to every request e.g test=null on release builds?

subsymbolic: Can we move this to the `defaultParameters` section in Pixel.kt so all the default params are grouped together.

CDRussell: >Would this add the test param to every request e.g test=null on release builds?

No; null values are omitted

>Can we move this to the defaultParameters section in Pixel.kt so all the default params are grouped together.

That's a nice idea, but I want to defer it. Declaring it differently to how we do it in `StatisticsService` isn't best for consistency, so if changing the way we add the header, i'd rather do it in both places. And I don't want to adopt that overhead of changing and testing `StatisticsService` right now, given the other work i'm currently focussed on.

subsymbolic: Looking at it again, I'm happy to forget about that suggestion entirely as other requests e.g atb.js use the pattern you have used.

PR #625, opened 2019-11-08
Restricts what URLs originating from a subframe in the WebView can do

CDRussell:
Task/Issue URL: https://app.asana.com/0/inbox/608920331025313/1148690082200045/1148578967949304
Tech Design URL: 
CC: 

**Description**:
Adds restrictions on what an iFrame can do when it's loading a URL

With this change, we only support 3rd party app launching if it was triggered from the main frame. For Android API 21-23, the functionality hasn't changed as we don't have information about whether on the main frame or not.

**Steps to test this PR**:
1. Test out the links in the linked Asana task on API 24+; verify we don't launch 3rd party apps
1. Test that links clicked on from the main frame still work as expected

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
```diff
@@ -160,4 +163,18 @@ class BrowserModule {
...
+    @Singleton
+    @Named(SpecialUrlHandler.MAIN_FRAME_HANDLER)
+    fun mainFrameUrlHandler(requestRewriter: RequestRewriter): SpecialUrlHandler {
+        return MainFrameUrlHandler(requestRewriter)
+    }
+
+    @Provides
+    @Singleton
+    @Named(SpecialUrlHandler.SUB_FRAME_HANDLER)
+    fun nonMainFrameUrlHandler(requestRewriter: RequestRewriter): SpecialUrlHandler {
```
marcosholgado: `requestRewriter` is not used here, consider removing from the constructor.

app/src/androidTest/java/com/duckduckgo/app/browser/SubFrameUrlHandlerTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import androidx.test.annotation.UiThreadTest
+import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
+import com.duckduckgo.app.browser.SpecialUrlDetector.UrlType.*
+import com.nhaarman.mockitokotlin2.mock
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+
+class SubFrameUrlHandlerTest {
```
marcosholgado: I can't see a `MainFrameUrlHandlerTest` class so I'm assuming the logic there has been tested somewhere else? If not, it might be a good idea to add them.

CDRussell: Good point; these should be added. Created https://app.asana.com/0/414730916066338/1149174387975737 for adding them after this PR.

PR #629, opened 2019-11-13
Add safer null handling around WebView copyBackForwardList calls

CDRussell:
Task/Issue URL: 
 - https://app.asana.com/0/0/1148502601574098
 - https://app.asana.com/0/0/1148504523095289

Tech Design URL: 
CC: 

**Description**:
Adds safer handling around access to `WebView`'s `copyBackForwardList` calls

**Steps to test this PR**:
1. With no easy way to replicate this for real, you might want to hack a line into `WebViewBackForwardListSafeExtractor` which randomly throws a `NullPointerException`, and check the app handles that happening. This should be a fair test 

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/navigation/WebViewBackForwardListSafeExtractorTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/navigation/WebViewBackForwardListSafeExtractorTest.kt
@@ -21,30 +21,26 @@ import android.webkit.WebBackForwardList
 import android.webkit.WebView
 import androidx.test.annotation.UiThreadTest
 import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
+import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
 import org.junit.Test
 
 class WebViewBackForwardListSafeExtractorTest {
 
-    @UiThreadTest
-    @Test(expected = NullPointerException::class)
-    fun whenCopyBackForwardListCalledTestWebViewThrowsException() {
-        val testWebView = TestWebViewThrowsNullPointerException(getInstrumentation().targetContext)
-        testWebView.copyBackForwardList()
-    }
+    private val context = getInstrumentation().targetContext
 
     @UiThreadTest
     @Test
     fun whenCopyBackForwardListCalledAndExceptionThrownThenNavigationListIsNull() {
-        val testWebView = TestWebViewThrowsNullPointerException(getInstrumentation().targetContext)
+        val testWebView = TestWebViewThrowsNullPointerException(context)
         assertNull(testWebView.safeCopyBackForwardList())
     }
 
     @UiThreadTest
     @Test
     fun whenCopyBackForwardListCalledAndNoExceptionThrownThenNavigationListIsNotNull() {
-        val testWebView = TestWebViewThrowsNullPointerException(getInstrumentation().targetContext)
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/com/duckduckgo/app/browser/navigation/WebViewBackForwardListSafeExtractorTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.test.annotation.UiThreadTest
+import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
+import org.junit.Assert.assertNull
+import org.junit.Test
+
+class WebViewBackForwardListSafeExtractorTest {
+
+    @UiThreadTest
+    @Test(expected = NullPointerException::class)
+    fun whenCopyBackForwardListCalledTestWebViewThrowsException() {
```
subsymbolic: Unit tests are often thought of as documentation. The place to check assumptions, how a system should and does behave. This unit test is really confusing in that it suggests that copying the back forward list should throw a NPE. Do we need this? I suggest we delete it.

CDRussell: Good point. I wanted the test to protect against someone accidentally changing the safe extractor in a way which no longer internally catches the exception. I think its existence adds value.

Maybe we name it better to be clear it's not expected behaviour that copying the list should throw the NPE?

app/src/androidTest/java/com/duckduckgo/app/browser/navigation/WebViewBackForwardListSafeExtractorTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun whenCopyBackForwardListCalledAndExceptionThrownThenNavigationListIsNull() {
+        val testWebView = TestWebViewThrowsNullPointerException(getInstrumentation().targetContext)
+        assertNull(testWebView.safeCopyBackForwardList())
+    }
+
+    @UiThreadTest
+    @Test
+    fun whenCopyBackForwardListCalledAndNoExceptionThrownThenNavigationListIsNotNull() {
+        val testWebView = TestWebViewThrowsNullPointerException(getInstrumentation().targetContext)
+        assertNull(testWebView.safeCopyBackForwardList())
```
subsymbolic: This should assert not null right?

app/src/androidTest/java/com/duckduckgo/app/browser/navigation/WebViewBackForwardListSafeExtractorTest.kt
CDRussell: Scratch that last comment; deleting it as per your suggestion.

app/src/androidTest/java/com/duckduckgo/app/browser/navigation/WebViewBackForwardListSafeExtractorTest.kt
CDRussell: Resolved

app/src/androidTest/java/com/duckduckgo/app/browser/navigation/WebViewBackForwardListSafeExtractorTest.kt
subsymbolic: Update: We discussed this over zoom and agreed to delete the additional test

PR #630, opened 2019-11-14
Fix abusing URI scheme

marcosholgado:
Task/Issue URL: https://app.asana.com/0/4665784270797/1148690082200046
Tech Design URL: 
CC: 

**Description**:
Certain URL schemes can cause undesired consequences when loaded as part of an iFrame. We released a patch for API 24+, however this one aims to solve the issue in lower APIs as well.

As part of this PR I have rolled back the changes done for the 24+ patch since they are not needed anymore.

**Steps to test this PR**:

**Normal behaviour**
1. Go to yelp.com, select any restaurant and click on the call button.
1. You should see a dialog asking you to open in another app.
1. Clicking on "Yes" should open the phone app.
1. Clicking on "No" should cancel the dialog.
1. Selecting "Close tab" should close the tab and remove from the tab switcher screen.

**Normal behaviour intents**
1. Open the Google PlayStore in the browser
1. Click in the "Open in play store app" button at the bottom of the screen
1. You should see a dialog asking you to open in another app
1. Clicking on "Yes" should open the phone app.
1. Clicking on "No" should cancel the dialog.
1. Selecting "Close tab" should close the tab and remove from the tab switcher screen.

**Abusing URI schemes**
1. Open http://qfe8.000webhostapp.com/death.html
1. You should see a dialog asking you to open in another app.
1. Immediately closing the dialog will triggered it again due to the website making several requests.
1. Selecting "Close tab" should dismiss the dialog and close the tab taking you back to a safe place.

**Multi selection test**
By default we directly launch the email, sms and phone apps without letting the user select any other apps that can also handle those intents. We do let them choose for other cases. To test this:

1. Install Skype
1. Change the line 442 in `BrowserTabFragment` to `openExternalDialog(intent, requireContext(), null, false)` to allow multiple selection for the phone scheme.
1. Open yelp, find a restaurant and tap the call button.
1. A dialog should appear with the normal options but when clicking on "Yes" that will trigger a new dialog/bottom sheet (depending on the OS version) asking you to choose the app that you want to use.

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -511,42 +513,57 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         }
     }
 
-    private fun externalAppLinkClicked(appLinkCommand: Command.HandleExternalAppLink) {
-        context?.let {
+    private fun openExternalDialog(intent: Intent, context: Context, fallbackUrl: String? = null, showFirstOption: Boolean = true) {
```
subsymbolic: The purpose of `showFirstOption` wasn't clear to me until I dug into the code and I would like to rename it although it's a tricky one to name. Do you think `useFirstExternalHandler` adds clarity?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -511,42 +513,57 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
+        context.let {
```
subsymbolic: Why use let here? Could you kill the context param and use the class context with `context?.let` instead? This addresses my previous context related comment also.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -511,42 +513,57 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
-                        .setMessage(getString(R.string.confirmOpenExternalApp))
-                        .setPositiveButton(R.string.openExternalApp) { _, _ -> it.startActivity(intent) }
-                        .setNegativeButton(android.R.string.cancel) { dialog, _ -> dialog.dismiss() }
-                        .show()
-                }
                 else -> {
-                    val title = getString(R.string.openExternalApp)
-                    val intentChooser = Intent.createChooser(intent, title)
-                    it.startActivity(intentChooser)
+                    if (activities.size == 1 || showFirstOption) {
```
subsymbolic: The higher-level `when` is no longer useful here and just adds confusion. Can we factor it out?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -511,42 +513,57 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
+                    }
                 }
             }
+        }
+    }
+
+    private fun launchExternalAppDialog(context: Context, onClick: () -> Unit) {
+        val isShowing = alertDialog?.isShowing
 
+        if (isShowing == null || !isShowing) {
```
subsymbolic: Update to `if (isShowing != true)`

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -62,29 +58,54 @@ class BrowserWebViewClient(
...
+    private fun shouldOverride(webView: WebView, url: Uri): Boolean {
+        try {
+            Timber.v("shouldOverride $url")
 
             val urlType = specialUrlDetector.determineType(url)
-            val urlHandler = getUrlHandler(requestOrigin)
-            urlHandler.handleUrl(webView, urlType, webViewClientListener)
 
+            return when (urlType) {
+                is SpecialUrlDetector.UrlType.Email -> consume { webViewClientListener?.sendEmailRequested(urlType.emailAddress) }
```
subsymbolic: I could imagine the  consume function working but it doesn't add clarity here - it makes the code less readable as some sections return and others don't. I'd rather we follow a consistent pattern of returning directly or from somewhere else

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -62,29 +58,54 @@ class BrowserWebViewClient(
...
      * API-agnostic implementation of deciding whether to override url or not
      */
-    private fun shouldOverride(webView: WebView, url: Uri, requestOrigin: RequestOrigin): Boolean {
-        return try {
-            Timber.d("Request origin: ${requestOrigin.javaClass.simpleName}, shouldOverride $url")
+    private fun shouldOverride(webView: WebView, url: Uri): Boolean {
+        try {
+            Timber.v("shouldOverride $url")
 
             val urlType = specialUrlDetector.determineType(url)
```
subsymbolic: Lint suggestion: `urlType` declaration can be moved into when

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -1133,6 +1134,15 @@ class BrowserTabViewModelTest {
         assertTrue(command is Command.LaunchNewTab)
     }
 
+    @Test
+    fun whenCloseCurrentTabThenDeleteTabFromRepository() = runBlocking {
```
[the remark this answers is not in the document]
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -426,16 +428,16 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
             is Command.DialNumber -> {
                 val intent = Intent(Intent.ACTION_DIAL)
                 intent.data = Uri.parse("tel:${it.telephoneNumber}")
-                activity?.launchExternalActivity(intent)
+                openExternalDialog(intent, requireContext(), null, false)
```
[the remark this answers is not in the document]
marcosholgado: Yeah, I have removed the context as a parameter following another comment below so this is not needed anymore :)

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Agreed! I was struggling trying to find a good name so I'd go with your suggestion. Another option could be 'useFirstActivityFound` or 'useFirstHandlerFound`, any thoughts?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Yes! Not sure why I had that parameter but agreed is not necessary!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
subsymbolic: They all sound better, happy with all of them.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -513,31 +513,28 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
-                        val activity = activities.first()
-                        val appTitle = activity.loadLabel(pm)
-                        Timber.i("Exactly one app available for intent: $appTitle")
-                        launchExternalAppDialog(it) { it.startActivity(intent) }
-                    } else {
-                        val title = getString(R.string.openExternalApp)
-                        val intentChooser = Intent.createChooser(intent, title)
-                        launchExternalAppDialog(it) { it.startActivity(intentChooser) }
-                    }
+                return
```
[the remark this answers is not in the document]
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -513,31 +513,28 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
 
-            when (activities.size) {
-                0 -> {
-                    if (fallbackUrl != null) {
-                        webView?.loadUrl(fallbackUrl)
-                    } else {
-                        showToast(R.string.unableToOpenLink)
-                    }
-                    return
+            if (activities.size == 0) {
```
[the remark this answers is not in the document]
marcosholgado: Done!

marcosholgado:
> This is getting close. There are a couple of tiny clean up tasks and we need to protect against long SMS url crashes as I can still crash the sms app with http://qfe8.000webhostapp.com/killsms.html from the HackerOne report and even a reboot doesn't fix it - the only way to recover is to clear data and caches which most users wouldn't know to do.
> 
> Can we limit the SMS url size to something really large but that doesn't crash e.g 10,000 chars (or whatever is appropriate)?

Thanks for the review! I have added a maximum length for the email, sms and phone schemes to avoid that issue. Let me know if you are happy with the sizes :)

cd1063:
How is a brand new coder supposed to get answers if he does not follow and learn from the ones with much greater knowledge.  I sorry for just starting out at an older age just wanting to learn something new.  the other major sites have encouraged me to learn and get answers to all my questions and here I getting shamed and kicked off of post I really need to learn about.  I never thought it would be like that at GitHub.   Tim

PR #640, opened 2019-12-02
Feature/craig/detect app referrer data

CDRussell:
Task/Issue URL: 

- https://app.asana.com/0/72649045549333/1151809625968850/f
- https://app.asana.com/0/72649045549333/1147201843767899/f

Tech Design URL: https://app.asana.com/0/72649045549333/1147201843767917/f
CC: 

**Description**:
Adds logic to extract app installation referrer data, if it's available. This is part of a test as to what data is exposed to the app through the referrer API. 

When the code runs, it'll first check local storage if it's had referrer data before. 

- If so, it'll load that and cache in memory.
- If not, it'll connect to the Google referrer library which will asynchronously return referrer information. This will then be stored locally, and cached in memory.

**Steps to test this PR**:
You need to test various flows here: 
1. the dev flow of installing the app from Android Studio
1. the dev flow installing from command line
1. the user flow of installing from Google Play (can use Google Play Ad-hoc installations; see below)

ℹ️Hint: Set logcat to filter for `referr`

### Testing Dev Flow - Android Studio
In this scenario, although some referrer information will be available to the app, it won't contain the campaign name we're looking for. As such, the referrer callback should happen, but we won't parse out any campaign suffix.

**Fresh install**
1. From run Android Studio, run the app as fresh install
1. Verify you see `Referrer information does not contain inspected campaign names`
1. Verify you see `Retrieved referrer data...`

**Run app again**
1. Verify you see `Already saw referrer data, but no campaign suffix saved`

### Testing Dev Flow - Terminal
Perform the same tests as above, but this time from the command line, sanity checking nothing breaks when installed from a different mechanism.

### Testing User Flow - Installing From Google Play
In this scenario, we want to closely emulate what a typical user would do to install the app; install it from Google Play. We can use the ad-hoc distribution to avoid having to do a _proper release_ to test this out 😌. You can create your own, but to save you time, _here's one I made earlier_. This is a `release` build but has **logging enabled** for convenience.

There are two links in the Asana task below: one which contains referrer information which matches in the app and one which does not.

#### Matching Campaign
1. Install the app using the **first link** in https://app.asana.com/0/72649045549333/1152096993579097/f
1. Open the app; verify logcat shows a match in campaign name `Found suffix XY`
1. Verify pixel `APP_REFERRER_INFO_AVAILABLE("m_ria")` sent, and contains a `d=` key.
     1. The duration will be printed in the logs. In particular, filter for `Retrieved referrer data; took`
     1. Using `DurationBucketMapper`, convert the duration into a bucket name.
     1. Verify the value for key `d` maps to the correct duration bucket name.
1. Verify pixel `APP_REFERRER_FOUND_TYPE_A("m_rfa")` sent
1. Restart the app, verify from the logs that `XY` is restored

#### Non-matching Campaign
1. Install the app using the **second link** in https://app.asana.com/0/72649045549333/1152096993579097/f
1. Open the app; verify logcat doesn't find a match `Referrer information does not contain inspected campaign names`
1. Verify pixel `APP_REFERRER_INFO_AVAILABLE("m_ria")` sent, and contains a `d=` key. (Even though it wasn't a match, we still got the information, and still want to see the timings)
     1. The duration will be printed in the logs. In particular, filter for `Retrieved referrer data; took`
     1. Using `DurationBucketMapper`, convert the duration into a bucket name.
     1. Verify the value for key `d` maps to the correct duration bucket name.
1. Restart the app; verify from the logs that it remembers it didn't find a campaign suffix previously.

### Notes

- `APP_REFERRER_FOUND_TYPE_B` cannot be tested until after the app is released, but please verify the logic looks good were it to match.
- Please test on emulator and real devices. A non Google-Play device would be an amazing test if you have one.
---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/referral/DurationBucketMapper.kt
+++ b/app/src/main/java/com/duckduckgo/app/referral/DurationBucketMapper.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2019 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.referral
+
+class DurationBucketMapper {
+
+    fun mapDurationToBucket(durationMs: Long): String {
+        for ((durationThreshold, bucketName) in durationBuckets) {
+            if (durationMs < durationThreshold) {
+                return bucketName
+            }
+        }
+        return DURATION_EXCEEDED_BUCKET_NAME
+    }
+
+    companion object {
+
+        private val durationBuckets = listOf(
+            Pair(100L, "0"),
+            Pair(200L, "1"),
+            Pair(500L, "2"),
+            Pair(1_000L, "3"),
+            Pair(1_500L, "4"),
+            Pair(2_000L, "5"),
+            Pair(2_500L, "6")
+        )
+
+        private const val DURATION_EXCEEDED_BUCKET_NAME = "7"
+    }
+}
--- a/app/src/main/java/com/duckduckgo/app/referral/DurationMeasuringReferrerRetriever.kt
+++ b/app/src/main/java/com/duckduckgo/app/referral/DurationMeasuringReferrerRetriever.kt
@@ -22,7 +22,10 @@ import com.duckduckgo.app.statistics.pixels.Pixel.PixelName
 import kotlinx.coroutines.withTimeoutOrNull
 import timber.log.Timber
 
-
+/**
+ * This is likely to be a fairly temporary class since it's being used to find out typical timings
+ * of referrer data being made available.
+ */
 class DurationMeasuringReferrerRetriever(
     private val referrerStateListener: AppInstallationReferrerStateListener,
     private val durationBucketMapper: DurationBucketMapper,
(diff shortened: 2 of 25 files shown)


app/src/main/java/com/duckduckgo/app/referral/DurationBucketMapper.kt
```diff
@@ -0,0 +1,44 @@
...
+            if (durationMs < durationThreshold) {
+                return bucketName
+            }
+        }
+        return DURATION_EXCEEDED_BUCKET_NAME
+    }
+
+    companion object {
+
+        private val durationBuckets = listOf(
```
subsymbolic: What's the purpose of the duration buckets? Why not send the ms time as a param? That would let you better tabulate, graph and manipulate the data.

CDRussell: Sending raw durations raised a concern around fingerprinting optics. i.e., that it would _look_ like we were sending a unique ID. Even though that's not what's going on, we don't need exact numbers, and bucketing the durations removes the potential for anyone to misinterpret the data.

subsymbolic: We've not worried that timing a 3rd party api may _look bad_ (we know it's not actually bad) before. Seem like we're being overly cautious here but I may be missing context, I'll dig in on Asana.

app/src/main/java/com/duckduckgo/app/referral/DurationMeasuringReferrerRetriever.kt
```diff
@@ -0,0 +1,88 @@
...
+import com.duckduckgo.app.statistics.pixels.Pixel
+import com.duckduckgo.app.statistics.pixels.Pixel.PixelName
+import kotlinx.coroutines.withTimeoutOrNull
+import timber.log.Timber
+
+/**
+ * This is likely to be a fairly temporary class since it's being used to find out typical timings
+ * of referrer data being made available.
+ */
+class DurationMeasuringReferrerRetriever(
```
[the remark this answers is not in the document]
CDRussell: thanks, much better!

PR #642, opened 2019-12-03
Amend app description string

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1152127350816673 
Tech Design URL: 
CC: 

**Description**:
In https://github.com/duckduckgo/Android/pull/622 we added a new string as an app description which read "Privacy, simplified" in order to avoid alignment issues with Android 10. Since we released that, some changes seem to have happened and we don't show "DuckDuckGo" as the app name in the system dialog anymore in Android 9 and below.

**Steps to test this PR**:
**Android 10**:
1. Launch the app on Android 10
1. Tap on Let's do It
1. Notice how DuckDuckGo should show as the app name and description in the system dialog

**Android 9**:
1. Launch the app on Android 9 or below
1. Tap on Let's do It
1. Notice how DuckDuckGo should show as the app in the system dialog
---
###### Internal references:

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -2,7 +2,7 @@
 
     <string name="appName" translatable="false">DuckDuckGo</string>
     <string name="duckDuckGoLogoDescription">DuckDuckGo logo</string>
-    <string name="appDescription">Privacy, simplified</string>
+    <string name="appDescription" translatable="false">DuckDuckGo&#160;</string>
     <string name="yes">Yes</string>
     <string name="no">No</string>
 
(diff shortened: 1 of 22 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -2,7 +2,7 @@
 
     <string name="appName" translatable="false">DuckDuckGo</string>
     <string name="duckDuckGoLogoDescription">DuckDuckGo logo</string>
-    <string name="appDescription">Privacy, simplified</string>
+    <string name="appDescription" translatable="false">DuckDuckGo&#160;</string>
```
marcosholgado: We add a blank space at the end so the description shows in Android 10. If the app name is the same as the description then no description is shown. Bing used to show both but now it doesn't (I take due to this).

PR #644, opened 2019-12-06
Allow users to recover from WebView renderer failures

cmonfortep:
Task/Issue URL: https://app.asana.com/0/0/1152512237237203/f
Tech Design URL: https://app.asana.com/0/0/1151307893053517/f
CC: 

### **Description**:
Introducing gracefully handling WebView crash errors into our Browser.
A WebView render process can be gone due to a rendering crash or memory pressure situations. When we receive such event from WebViewClient we will avoid crashing, and instead, we will show a snackbar to inform about what happened and the possibility to reload the current URL. Also, application will still be usable so users will be able to perform other actions as closing the tab or navigate to a new URL.

### **Steps to test this PR**:
As we can't reproduce easily memory pressure situations or crashes, we will need to change our source code a bit in order to manually produce a crash.
Easiest way is to call `recoverFromRenderProcessGone()` when user clicks in a menu option, example will be inside `onBrokenSiteSelected()`.

### **Scenarios to test**
***Scenario to test***:
- Open app
- submit an URL
- crash the app (with hardcoded code)
- Verify user can't navigate back/forward/find a page/report a site

***Scenario to test***:
- Open app
- submit an URL
- crash the app (with hardcoded code)
- Verify snackbar appears
- Reload from snackbar
- Verify current tab is closed and a new one is opened with same URL

**Scenario to test**:
- Open app
- submit an URL
- crash the app (with hardcoded code)
- Verify snackbar appears
- Reload from option menu
- Verify current tab is closed and a new one is opened with same URL

**Scenario to test**:
- Open app
- submit an URL
- Navigate to produce some history navigation
- crash the app (with hardcoded code)
- Verify snackbar appears
- press back
- Verify home is shown
- press back
- Verify exits the app

**Scenario to test**:
- Open app
- submit an URL
- crash the app (with hardcoded code)
- Verify snackbar appears
- Submit current or a new URL
- Verify current tab is closed and a new one is opened with expected URL

**Scenario to test**:
- Open app
- submit an URL
- crash the app (with hardcoded code)
- Verify snackbar appears
- Open privacy settings and disable or enable toogle
- Go back to the tab
- Verify current tab is closed and a new one is opened with same URL

**Scenario to test**:
- Open app
- submit an URL
- crash the app (with hardcoded code)
- Open a new tab
- submit an URL
- crash the app (with hardcoded code)
- Change between tabs, snackbar should appear

**Scenario to test**:
- Open app
- submit an URL
- crash the app (with hardcoded code)
- Open a new tab
- submit an URL
- crash the app (with hardcoded code)
- Send app to background
- Bring back to foreground the app
- Snakbar should appear

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -1122,7 +1122,9 @@ class BrowserTabViewModelTest {
 
         showErrorWithAction.action()
 
-        assertCommandIssued<Command.OpenInNewTab>()
+        assertCommandIssued<Command.OpenInNewTab> {
+            assertEquals("https://example.com", query)
+        }
     }
 
     @Test
@@ -1278,10 +1280,11 @@ class BrowserTabViewModelTest {
         verify(mockCommandObserver, never()).onChanged(commandCaptor.capture())
     }
 
-    private inline fun <reified T : Command> assertCommandIssued() {
+    private inline fun <reified T : Command> assertCommandIssued(instanceAssertions: T.() -> Unit = {}) {
         verify(mockCommandObserver, atLeastOnce()).onChanged(commandCaptor.capture())
         val issuedCommand = commandCaptor.allValues.find { it is T }
         assertNotNull(issuedCommand)
+        (issuedCommand as T).apply { instanceAssertions() }
     }
 
     private fun pixelParams(showedBookmarks: Boolean, bookmarkCapable: Boolean) = mapOf(
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -75,6 +75,7 @@ import io.reactivex.Single
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.TestCoroutineScope
 import kotlinx.coroutines.test.runBlockingTest
 import kotlinx.coroutines.withContext
 import org.junit.After
@@ -180,6 +181,8 @@ class BrowserTabViewModelTest {
 
     private lateinit var testee: BrowserTabViewModel
 
+    private val selectedTabLiveData = MutableLiveData<TabEntity>()
+
     @Before
     fun before() {
         MockitoAnnotations.initMocks(this)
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt
```diff
@@ -165,5 +165,10 @@ class SurveyActivity : DuckDuckGoActivity() {
                 viewModel.onSurveyFailedToLoad()
             }
         }
+
+        override fun onRenderProcessGone(view: WebView?, detail: RenderProcessGoneDetail?): Boolean {
+            viewModel.onSurveyFailedToLoad()
```
subsymbolic: Is this valid? It eventually leads to a call to `webView.gone()` however according to the docs:

> The given WebView can't be used, and should be removed from the view hierarchy, all references to it should be cleaned up

In the case of an `onRenderProcessGone` failure the `WebView` should removed and set to null not hidden.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1124,13 +1141,23 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
-                    browserLayout.show()
+                when (viewState) {
+                    is GlobalLayoutViewState.Browser -> {
+                        if (viewState.isNewTabState) {
+                            browserLayout.hide()
+                        } else {
+                            browserLayout.show()
+                        }
+                    }
+                    is GlobalLayoutViewState.Invalidated -> destroyWebView()
```
subsymbolic: Is this valid? It eventually calls `webview.destroy()` however according to the docs:
> The given WebView can't be used, and should be removed from the view hierarchy, all references to it should be cleaned up

I recommend creating an invalidate method that removes and nulls out the WebView rather than reusing this method that interacts with the Webview instance when we don't know what state it is in.

cmonfortep: Good point, I had the same doubt because "WebView can't be used" is not clear enough.
I did some code investigations and tests, and I deduced that they refer to continue rendering in that `WebView` instance, or any action that will directly involve the renderer process.

[While reading again the docs](https://developer.android.com/guide/webapps/managing-webview#termination-handle), their sample is using exactly that method call, so I would say that it's safe and correct in order to clean up WebView internal references.
```
// Remove the inflated view from the container and cleanup
// the dead webview specifically (if it is not GC'ed it will cause
// problems later, the next time the renderer dies)
(...)
view.destroy();
```

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt
cmonfortep: Yes, it's valid to call `webView.gone()` but not enough. I will improve this part in order to do a better clean up for that `WebView`.

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -1042,14 +1099,80 @@ class BrowserTabViewModelTest {
...
+    @Test
+    fun whenUserClicksOnErrorActionThenOpenCurrentUrlInNewTab() {
+        givenOneActiveTabSelected()
+        testee.recoverFromRenderProcessGone()
+        verify(mockCommandObserver, atLeastOnce()).onChanged(commandCaptor.capture())
+        val showErrorWithAction = commandCaptor.value as Command.ShowErrorWithAction
+
+        showErrorWithAction.action()
+
+        assertCommandIssued<Command.OpenInNewTab>()
```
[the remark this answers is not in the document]
cmonfortep: Sure, I will add a new testcase to cover that part.

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt
cmonfortep: I went for adding extra logic inside `showError` in order to fully remove the `WebView`, instead of creating a new method in `SurveyViewModel` + a new `Command` to destroy the `WebView`.
Since currently when an error happens user can't interact with Survey, I just destroy the WebView too, there's no need to keep it alive.

PR #647, opened 2019-12-13
Update tracker detection to use new tds data set

subsymbolic:
**Task/Issue URL:** 

- New TDS Support: https://app.asana.com/0/361428290920652/1147045768949348
- Embed TDS in App: https://app.asana.com/0/361428290920652/1147872489592776
- Add global/temp whitelist: https://app.asana.com/0/361428290920652/1147872489592777 

**Tech Design URL:**

- https://app.asana.com/0/361428290920652/1132331070554722

**Description**:

- Implement new tracker data set and entities
- Embed TDS data in app to protect first run users
- Remove legacy tracker detection code and libraries
- Add Global whitelist support to allow us to publish temporary fixes and recover quickly from any issues in the TDS data set.

**Steps to test this PR**

**Test Setup:**
- in `PrivacyDashboardViewModel` set `private const val LEADERBOARD_MIN_DOMAINS_EXCLUSIVE = 3` 

**Test Database Migration:**
- Install the app from develop
- Run the app and load some sites (at least 3 including one with many trackers)
- Now install the latest version, run the app and ensure everything runs fine
- Check that the network leaderboard pills in the Privacy Dashboard are reset
- Run the app and load some sites (at least 3 including one with many trackers)
- Check that the network leaderboard pills are now populated

**Test that trackers are detected and displayed as expected:**
- Open a site with many trackers such as cnn.com
- Tap on the privacy grade to open the Privacy Dashboard
- Ensure that trackers are listed
- Ensure that the network list shows major trackers first and that icons (where they exist) are populated

**Test that embedded tds still works even if download fails:**
- Use charles to block calls to `https://staticcdn.duckduckgo.com/trackerblocking/v2/tds.json`
- Install the app and ensure that tracker blocking still works as normal 

**Test that tds is updated from backend:**
- Update the embedded tds.json file content to be empty json`{}`
- Run a clean install of the app and allow app config download to complete
- Visit a site with trackers and ensure that these are still detected

**Test temporary/global whitelist:**
- Visit a site in our temporary whitelist https://duckduckgo.com//contentblocking/trackers-whitelist-temporary.txt e.g fidelity.com and ensure that no trackers are blocked on it
- Now cleanly install the app and block the call to the temporary whitelist
- Wait for the rest of the app config to download
- Visit the site previously tested and check that trackers are now blocked on it

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/trackerdetection/model/TdsTracker.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    @TypeConverter
+    fun fromCategories(value: List<String>): String {
+        return Adapters.stringListAdapter.toJson(value)
+    }
+}
+
+class Adapters {
+    companion object {
+        private val moshi = JsonModule().moshi()
```
subsymbolic: This is a bit icky - let me know if you have any ideas for how we can use DI to get a moshi instance into the converters.

CDRussell:
>Test that embedded tds still works even if download fails

From a clean slate, if i perform this test then the privacy grade never shows

app/src/main/java/com/duckduckgo/app/trackerdetection/model/TdsEntity.kt
```diff
@@ -14,10 +14,15 @@
...
-package com.duckduckgo.app.privacy.store
+package com.duckduckgo.app.trackerdetection.model
 
-interface PrevalenceStore {
+import androidx.room.Entity
+import androidx.room.PrimaryKey
 
-    fun findPrevalenceOf(entity: String): Double?
+@Entity(tableName = "tds_entity")
+data class TdsEntity(
```
CDRussell: Why do we need this and the `Entity` interface?

app/src/main/java/com/duckduckgo/app/trackerdetection/model/TrackingEvent.kt
```diff
@@ -16,12 +16,10 @@
...
 
-import com.duckduckgo.app.global.UriString.Companion.host
-
 data class TrackingEvent(
     val documentUrl: String,
     val trackerUrl: String,
-    val trackerNetwork: TrackerNetwork?,
-    val blocked: Boolean,
-    val entity: String = trackerNetwork?.name ?: host(trackerUrl) ?: trackerUrl
+    val categories: List<String>?,
```
CDRussell: is an empty list of categories the same as a null list? if so, could we make this non-nullable?

app/src/main/java/com/duckduckgo/app/privacy/ui/TrackerNetworksViewModel.kt
```diff
@@ -16,18 +16,23 @@
...
+import java.util.*
 
 class TrackerNetworksViewModel : ViewModel() {
 
     data class ViewState(
         val domain: String,
         val allTrackersBlocked: Boolean,
         val trackerCount: Int,
-        val trackingEventsByNetwork: Map<String, List<TrackingEvent>>
+        val trackingEventsByNetwork: SortedMap<Entity, List<TrackingEvent>>
```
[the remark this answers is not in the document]
subsymbolic: This data structure only exists for display purposes and since we display these as a **sorted** list of entities (with major trackers coming first) then it felt like a better fit for storage. We used to keep a separate list of sorted keys to allow us to pull items out of the map in the right order which made the logic harder to follow.

app/src/main/java/com/duckduckgo/app/trackerdetection/model/TdsEntity.kt
subsymbolic: Having the `Entity` interface allows the classes that do entity lookups to avoid being tightly coupled with the TDS list and its concrete entity type.

app/src/main/java/com/duckduckgo/app/trackerdetection/api/TdsJson.kt
```diff
@@ -0,0 +1,76 @@
...
+)
+
+data class TdsJsonOwner(
+    val name: String
+)
+
+class ActionJsonAdapter {
+
+    @FromJson
+    fun fromJson(actionName: String): Action? {
```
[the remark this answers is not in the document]
subsymbolic: We have one place where the action can be null and another place where it can't so capturing the nullability at the variable level rather than embedding it in the enum cases makes the code more understandable. I love than kotlin allows us to express this so well. If we were to introduce a null enum case we would required handle to consider that in the place where the action can't be null which wouldn't make sense.

app/src/main/java/com/duckduckgo/app/trackerdetection/model/TrackingEvent.kt
subsymbolic: Good suggestion, I'll make that an empty list.

app/src/androidTest/java/com/duckduckgo/app/trackerdetection/api/TdsEntityJsonTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    fun whenFormatIsValidThenBasicElementsAreConvertedCorrectly() {
+        val json = loadText("json/tds_entities.json")
+        val entities = jsonAdapter.fromJson(json)!!.jsonToEntities()
+        val entity = entities.first()
+        assertEquals(TdsEntity("21 Productions Inc", "21 Productions", 0.348), entity)
+    }
+
+    @Test
+    fun whenEntityIsMissingPrevlaendeThenPrevalenceIsSetToZero() {
```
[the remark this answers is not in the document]
subsymbolic: lol, I managed to cram a few typos in that one! 🤦‍♂

app/src/main/java/com/duckduckgo/app/trackerdetection/model/TrackingEvent.kt
subsymbolic: Now that I'm making the change it feels a little clunky to have an empty list here as other variables can in the type can also be null. Do you feel strongly about this? Happy to zoom if you do.

app/src/main/java/com/duckduckgo/app/trackerdetection/api/TdsJson.kt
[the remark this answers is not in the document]
CDRussell: what does it mean for an action to be null, though? that's what's unclear to me about the current code. what action should be taken if it's null?

app/src/main/java/com/duckduckgo/app/trackerdetection/model/TrackingEvent.kt
CDRussell: It's not so much that i feel strongly; it's more than i'm less sure what to do with a null list than an empty one. I suspect the answer is "treat them both the same" (in which case, i'd prefer not having to push handling both cases onto all other classes that use it)

i think the list case is different than the others. with the others, it's either null or it isn't. with a list, it's either null, or empty, or populated. 3 states instead of 2.

Again, i don't feel strongly if you think this is the best representation. but it seems more confusing to me to allow a null list here.

app/src/main/java/com/duckduckgo/app/trackerdetection/TdsClient.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    override fun matches(url: String, documentUrl: String): Client.Result {
+        val tracker = trackers.firstOrNull { sameOrSubdomain(url, it.domain) } ?: return Client.Result(false)
+        val matches = matchesTrackerEntry(tracker, url, documentUrl)
+        return Client.Result(matches, tracker.ownerName, tracker.categories)
+    }
+
+    private fun matchesTrackerEntry(tracker: TdsTracker, url: String, documentUrl: String): Boolean {
+        tracker.rules.forEach { rule ->
+            val regex = ".*${rule.rule}.*".toRegex()
```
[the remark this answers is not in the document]
subsymbolic: Great question! As we are first matching on domain and only iterating through a very short (usually empty) rule list if there is a domain match, it's not too heavy. Performance check suggests that things are better than before this PR. Checking this loop on a few really heavy sites, it generally executes in nanoseconds and I didn't see it hitting a full ms once. CPU and memory also look fine. 

timings

subsymbolic:
Thanks for the excellent and thorough review @CDRussell! I'll go ahead and merge this now however you covered two optional suggestions for switching to non-null references where I have a preference for keeping them as null-references. Let's hash out the nuances of those choices in our call tomorrow and I can follow up with a small tidy up PR depending on the outcome of that conversation.

app/src/main/java/com/duckduckgo/app/trackerdetection/model/TdsTracker.kt
CDRussell: My suggestion here isn't much of an improvement, but here goes (based on https://stackoverflow.com/a/55600994/1654145)

1. Add a field to `AppDatabase`'s companion object, for Moshi
1. Pass the moshi instance through DI to `DatabaseModule` when creating the `AppDatabase` instance
1. Grab `moshi` instance from `AppDatabase.moshi`

I would say it's still not great, but it is a mild improvement over manually instantiation of `JsonModule`. what do you think?

CDRussell: @Module(includes = [DaoModule::class])
    class DatabaseModule {

    @Provides
    @Singleton
    fun provideDatabase(context: Context, moshi: Moshi): AppDatabase {
        val db = Room.databaseBuilder(context, AppDatabase::class.java, "app.db")
            .addMigrations(*AppDatabase.ALL_MIGRATIONS.toTypedArray())
            .build()

        AppDatabase.moshi = moshi

        return db
    }



    abstract class AppDatabase : RoomDatabase() {

        ...

        companion object {

            lateinit var moshi: Moshi
        }
}

subsymbolic: Nice, I've merged but will add that to a tidy up PR after our conversation tomorrow.

PR #650, opened 2019-12-16
Hotfix/5.38.1

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1154099029852441
Tech Design URL: 
CC: 

**Description**:
Hotfix for `SecurityException` thrown during interaction with referrer library

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/referral/AppInstallationReferrerStateListener.kt
```diff
@@ -88,22 +88,28 @@ class PlayStoreAppReferrerStateListener @Inject constructor(
...
-                val parsedResult = appInstallationReferrerParser.parse(referrer)
-                referralResultReceived(parsedResult)
             }
-            FEATURE_NOT_SUPPORTED -> referralResultFailed(FeatureNotSupported)
-            SERVICE_UNAVAILABLE -> referralResultFailed(ServiceUnavailable)
-            DEVELOPER_ERROR -> referralResultFailed(DeveloperError)
-            SERVICE_DISCONNECTED -> referralResultFailed(ServiceDisconnected)
-            else -> referralResultFailed(UnknownError)
+
+            referralClient.endConnection()
```
cmonfortep: Because we are dealing with such unstable API, I wonder if we should try to execute `endConnection()` inside a `finally` block + another `try-catch` in case it fails (old java typical code when dealing with Files), or do you think this method is the only possible source of exception within this block of code?.

PR #656, opened 2019-12-21
When activity is found to be null, don't attempt further UI calls

CDRussell:
Task/Issue URL: https://app.asana.com/0/1154851166658740/1154845361621030
Tech Design URL: 
CC: 

**Description**:
I _think_ attempting to cancel the animations after detecting the `activity` is null is bad, and can actually be a cause of new `IllegalStateExceptions` being triggered.

**Steps to test this PR**:
1. Increase `TRACKERS_SECONDARY_DELAY` to a large value (~ 7_000L)
1. Load a webpage (e.g., cnn.com)
1. Wait for "scanning" text to appear
1. Press `back` a few times, until you have exited back to home screen
1. Wait for remainder of `TRACKERS_SECONDARY_DELAY` time to finish
1. Verify no exception thrown (in logcat)

If you try the above on `develop`, you should be able to see an `IllegalStateException` being thrown

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -1189,18 +1189,15 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
                     val site = viewModel.siteLiveData.value
                     val events = site?.trackingEvents
 
-                    val act = activity
-                    if (act != null) {
+                    activity?.let { activity ->
                         animatorHelper.createLoadedAnimation(
                             lastSeenCtaViewState?.cta,
-                            act,
+                            activity,
                             networksContainer,
                             loadingText,
                             omnibarViews(),
                             events
                         )
-                    } else {
-                        cancelAllAnimations()
                     }
                 }
             }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -1095,7 +1095,7 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         private const val MIN_PROGRESS = 10
         private const val MAX_PROGRESS = 100
         private const val TRACKERS_INI_DELAY = 700L
-        private const val TRACKERS_SECONDARY_DELAY = 7000L
+        private const val TRACKERS_SECONDARY_DELAY = 300L
 
         fun newInstance(tabId: String, query: String? = null, skipHome: Boolean): BrowserTabFragment {
             val fragment = BrowserTabFragment()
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1189,18 +1189,15 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
                     val site = viewModel.siteLiveData.value
                     val events = site?.trackingEvents
 
-                    val act = activity
-                    if (act != null) {
+                    activity?.let { activity ->
```
cmonfortep: I did a quick look to the previous merged PR and to this one.

I would like to share some thoughts here:
1. We should not reach this point if the activity is null, rather than doing a null check. I think the fix will be just to cancel the coroutine. From what I see, delay is suspending this block, if user leaves the screen and we don't cancel the coroutine, the coroutine will be resumed and that's why the crash will happen for `requireActivity()`.
2. Not canceling also means that we are leaking.

Suggestion: as a quick fix, without checking all the different coroutines that we are starting from `BrowserTabFragment`, I would keep the reference to the `SupervisorJob`, and cancel `onDestroy`. I think there are some CoroutineContext more aware of the lifecycle, but we can investigate that part in future tasks.

CDRussell: I like this idea! In theory, if doing so, we should assume code after `delay` won't run at all in that scenario, since `delay` checks if the coroutine is active for us IIRC

cmonfortep: That's what I expect. Any method from `kotlinx.coroutines` is aware of cancelations, so code after `delay` should not be executed if coroutine is canceled.

marcosholgado: This makes sense to me, I have pushed a new commit which includes @cmonfortep's change.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1093,7 +1095,7 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         private const val MIN_PROGRESS = 10
         private const val MAX_PROGRESS = 100
         private const val TRACKERS_INI_DELAY = 700L
-        private const val TRACKERS_SECONDARY_DELAY = 300L
+        private const val TRACKERS_SECONDARY_DELAY = 7000L
```
cmonfortep: //TODO: remove before merging :D

marcosholgado: well spotted! Back to 300L now!

PR #657, opened 2019-12-22
Add missing year and copyright owner to LICENSE

nwtgck:
Task/Issue URL: 
Tech Design URL: 
CC: 

**Description**:
Thank you for the wonderful project! 

I just added missing year and the name to `LICENSE` in the same way as iOS.
(iOS commit: <https://github.com/duckduckgo/iOS/commit/db20e92ad22122c8d2ecebc3761f102ca1a8eef2#diff-9879d6db96fd29134fc802214163b95a>)

Note that "O" is an upper case like "DuckDuckGO" because I used the same way as iOS. Should I use lower case "o" like "DuckDuckGo"?

**Steps to test this PR**:
1. 
1. 

---
###### Internal references:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -191,7 +191,7 @@ versions can be downloaded from https://duckduckgo.com/press.
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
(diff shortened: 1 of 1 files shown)


subsymbolic:
Thanks for contributing! This isn't however a change we need, it's the appendix section of the license and informational only.

Issue #638, opened 2019-12-01
Webview positioned partially offscreen


qrkz:
The fixed bottom navigation bar on webpages stay hidden until the user scrolls down the page and the address-bar of the browser hides.

The issue is that the browser doesn't take into account the height of addressbar while calculating the height of viewport(vh). Almost all other mobile browsers subtract the addressbar height from the viewport height.

shahsurajk:
Any particular website to test this scenario?

cd1063:
I have tried on several websites and not getting the same results. I'll try a few more more social and encrypted sites tonight and see if I get any different results.

subsymbolic:
Thanks for yous input and for seeking examples. A solid breaking example will help us triage and prioritize this. If an important site that many of our users need is breaking, we would bump the priority.

qrkz:
> Any particular website to test this scenario?

You can try [qrkz.github.io](https://qrkz.github.io/vj/pages/botnav.html).
What to look for:
When the page initially loads, you can only see a part of the fixed bottom navbar.
When the user scrolls the page only then is the navbar visible...

you can even try out any website which has a fixed bottom navigation bar.

qrkz:
> Thanks for yous input and for seeking examples. A solid breaking example will help us triage and prioritize this. If an important site that many of our users need is breaking, we would bump the priority.

We can see the bug by opening the PWA version of Pinterest or m.youtube.com and see how the bottom navbar isn't easily accessible... Bottom navbar in pinterest is almost inaccessible...

CDRussell:
I had a quick look at this today, and the issue is that the web view can actually off be positioned partially offscreen depending on the scrolling. When you scroll the page down, the toolbar slides up, as does the web view. So only when the toolbar is fully up off screen is the web view fully positioned onscreen.

It seems to be a combo of the scroll behaviour + `CoordinatorLayout` that is causing the issue. I found some potential hacky solutions but none that looked particularly great and I don't have time to investigate this much further right now.

Poopooracoocoo:
Related: #689

RichyRich7-crypto:
Try implementing floating keyboard.

Slion:
You can also reproduce on http://bbc.co.uk/news the accept cookies dialog is cut if the address bar is showing.

raboof:
Another simple reproducer is https://arnout.engelen.eu/fullpage.html

cmonfortep:
[...]

PR #674, opened 2020-01-17
Fix webview positioned partially offscreen (#638)

mattrob33:
Task/Issue URL: #638 
Tech Design URL: 
CC: @CDRussell 

**Description**:
After this change, bottom-anchored content in web pages will display anchored to the bottom of the screen, rather than being partially hidden offscreen.

**Steps to test this PR**:
1. Navigate to reddit.com or m.youtube.com

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -975,12 +975,8 @@ class BrowserTabViewModel(
 
     override fun onSystemLocationPermissionNeverAllowed() {
         locationPermission?.let { locationPermission ->
-            val neverAllowedPermission = LocationPermissionType.DENY_ALWAYS
-            onSiteLocationPermissionSelected(locationPermission.origin, neverAllowedPermission)
+            onSiteLocationPermissionSelected(locationPermission.origin, LocationPermissionType.DENY_ALWAYS)
             pixel.fire(Pixel.PixelName.PRECISE_LOCATION_SYSTEM_DIALOG_NEVER)
-            viewModelScope.launch {
-                locationPermissionsRepository.savePermission(locationPermission.origin, neverAllowedPermission)
-            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


subsymbolic:
@CDRussell are you happy to grab this one? You're probably the best person to review it since you investigated this issue last time.

CDRussell:
Thanks for answering the #help-wanted tag and for this contribution @mattrob33 !

I tested this out and it's certainly fixed the bottom issue, but it seems to have introduced unintended side effects:

1. Now the toolbar can be scrolled away, even if the `WebView` isn't scrolling. This can be seen on reddit.com when the bottom sheet / "use our app" component shows. If you try to scroll the screen, the toolbar moves around even though the web content is fixed. This isn't a major issue, and wouldn't itself be a blocker to including.
2. A bigger issue seems to be that now the top of the `WebView` content is being covered by the toolbar in some cases. I've attached a screengrab to illustrate the problem. That white youtube toolbar is hard to reach now as our native toolbar sits on top of it if you scroll to the top of the page. So to access that toolbar, you'd have to scroll all the way to top of the page, then try slowly scrolling down to hide our toolbar. 

I feel like your solution is much closer to what we need but still needs some tweaks yet. Do you have any thoughts on how to solve those issues?

can't-access-top-of-web-content

mattrob33:
Ah yikes. I'll look into CoordinatorLayout a bit more for a better adjustment. Ideally the WebView would always top-align with the bottom of the omnibox and bottom-align with the bottom of the screen, right?

CDRussell:
> Ideally the WebView would always top-align with the bottom of the omnibox and bottom-align with the bottom of the screen

I think top-aligning with the bottom of the omnibox, yes. Basically want to replicate how Chrome handles this as it does it well.

sanogueralorenzo:
Hello 👋 

I was going to open an issue regarding this but I believe this PR is addressing it 🤔 

Tested on latest release + develop (that doesn't have the bottom bar anymore 😢 ) and have the same unexpected behavior
image

PR #699, opened 2020-02-03
Fix: cancel coroutine when activity gets destroyed

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1159930185369809/f

**Description**:
This fixes some crashes when trying to commit a Fragment transaction when the activity is destroyed.
Additionally, I'm removing also a reference to the `currentFragment` when activity is destroyed as this is causing a leak.

**Steps to test this PR**:

**How to reproduce the crash**
In order to reproduce the crash, we need to hardcode a `delay` in our code and enable activities being killed. Also, we need to use current `develop` branch.

- Developer settings -> enable don't keep activities
- In code ->
```
    fun openMessageInNewTab(message: Message) {
        launch {
            val tabId = viewModel.onNewTabRequested()
delay(5000)
            val fragment = openNewTab(tabId, null, false)
            fragment.messageFromPreviousTab = message
        }
    }
```

1. Visit https://www.w3schools.com/code/tryit.asp?filename=G8SKXWB6508R
1. Press "Run"
1. Tap the "click on me" button or link
1. Navigate to tab switcher screen

**How to test this fix**
Checkout this branch.
Follow the same steps as in the previous description.
Crash should be gone.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -103,8 +103,12 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope {
         viewModel.awaitClearDataFinishedNotification()
     }
 
-    override fun onDestroy() {
+    override fun onStop() {
         job.cancel()
+        super.onStop()
+    }
+
+    override fun onDestroy() {
         currentTab = null
         super.onDestroy()
     }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -56,7 +56,7 @@ import kotlin.coroutines.CoroutineContext
 class BrowserActivity : DuckDuckGoActivity(), CoroutineScope {
 
     override val coroutineContext: CoroutineContext
-        get() = SupervisorJob() + Dispatchers.Main
+        get() = job + Dispatchers.Main
 
     @Inject
     lateinit var clearPersonalDataAction: ClearPersonalDataAction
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -56,7 +56,7 @@ import kotlin.coroutines.CoroutineContext
 class BrowserActivity : DuckDuckGoActivity(), CoroutineScope {
 
     override val coroutineContext: CoroutineContext
-        get() = SupervisorJob() + Dispatchers.Main
+        get() = job + Dispatchers.Main
```
CDRussell: Instead of defining the `coroutineContext` manually, alongside a reference to a `job`, what's your thoughts on using `MainScope` instead?

## Class definition
`class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {`

## Cancelling job
`coroutineContext.cancel()`

## Functionally equivalent?
I think this should be functionally equivalent, but would allow us to delete the `CoroutineContext` and `Job` references?

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -103,8 +103,12 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope {
         viewModel.awaitClearDataFinishedNotification()
     }
 
-    override fun onDestroy() {
+    override fun onStop() {
```
cmonfortep: Double-checking this, I saw that we are executing fragment transactions inside our coroutine, and so, we need to cancel any work at this point.
When activity goes to background, `onDestroy` won't be called (if activity is not killed), only `onStop`.

PR #711, opened 2020-02-13
Smoother progress bar

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1161499456202321
Tech Design URL: 
CC: 

**Description**:
The way the progress bar is updated it stalls most of the time giving our users a not great experience. This PR adds a smoother progress bar animation starting to 50% and overriding it if a newer update comes through. 

**Steps to test this PR**:
1. Launch different websites: SERP, youtube, etc and notice how the progress bar animation is smoother than with the previous version.

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -543,10 +543,6 @@ class BrowserTabViewModelTest {
     fun whenBrowsingAndViewModelGetsProgressUpdateThenViewStateIsUpdated() {
         setBrowserShowing(true)
 
-        testee.progressChanged(0)
-        assertEquals(0, loadingViewState().progress)
-        assertEquals(true, loadingViewState().isLoading)
-
         testee.progressChanged(50)
         assertEquals(50, loadingViewState().progress)
         assertEquals(true, loadingViewState().isLoading)
@@ -554,6 +550,19 @@ class BrowserTabViewModelTest {
         testee.progressChanged(100)
         assertEquals(100, loadingViewState().progress)
         assertEquals(false, loadingViewState().isLoading)
+    }
+
+    @Test
+    fun whenBrowsingAndViewModelGetsProgressUpdateLowerThan50ThenViewStateIsUpdatedTo50() {
+        setBrowserShowing(true)
+
+        testee.progressChanged(0)
+        assertEquals(50, loadingViewState().progress)
+        assertEquals(true, loadingViewState().isLoading)
+
+        testee.progressChanged(15)
+        assertEquals(50, loadingViewState().progress)
+        assertEquals(true, loadingViewState().isLoading)
     }
 
     @Test
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -17,8 +17,10 @@
 package com.duckduckgo.app.browser
 
 import android.Manifest
+import android.animation.Animator
 import android.animation.LayoutTransition.CHANGING
 import android.animation.LayoutTransition.DISAPPEARING
+import android.animation.ObjectAnimator
 import android.annotation.SuppressLint
 import android.app.Activity.RESULT_OK
 import android.app.ActivityOptions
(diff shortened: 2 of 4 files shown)


app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -556,6 +552,19 @@ class BrowserTabViewModelTest {
         assertEquals(false, loadingViewState().isLoading)
     }
 
+    @Test
+    fun whenBrowsingAndViewModelGetsProgressUpdateLowerThan50ThenViewStateIsUpdatedTo50() {
+        setBrowserShowing(true)
+
+        testee.progressChanged(0)
+        assertEquals(50, loadingViewState().progress)
```
malmstein: maybe reference the constant instead of using a magic number?

marcosholgado: Yeah I thought about it but then the constant would have to be public and it kinda put me off, but happy to change it though.

malmstein: fair enough, let's keep it like this! 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1186,14 +1193,39 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
         }
 
+        private fun animateProgressBar(newProgress: Int, isLoading: Boolean) {
+            progressBarAnimation?.cancel()
+
+            if (pageLoadingIndicator.progress > newProgress) {
+                pageLoadingIndicator.progress = 0
+            }
+
+            progressBarAnimation = ObjectAnimator.ofInt(pageLoadingIndicator, "progress", newProgress).apply {
```
cmonfortep: could this `ObjectAnimator` be moved into an individual class?

My motivation is just to avoid this class to get bigger, and avoid adding distraction when adding new features, if moving it away keeps the same size or makes the read more complex we can keep it here.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1186,14 +1193,39 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
             }
         }
 
+        private fun animateProgressBar(newProgress: Int, isLoading: Boolean) {
+            progressBarAnimation?.cancel()
+
+            if (pageLoadingIndicator.progress > newProgress) {
```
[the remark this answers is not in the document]
marcosholgado: Checks if the current progress is higher than the new one which means the animation would go from right to left rather than left to right making it a "deprogress toolbar" 
I could move it to a val but it reads fine to me, the alternative is something like `val isCurrentProgressHigherThanNewProgress` but we already know is doing that.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: I can't see how to do that without making it more difficult to understand. `ObjectAnimator` is a class in itself and here we are creating one with a listener. What I can do is using the crossinline listeners to remove noise though.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1195,24 +1196,16 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
-                })
-                start()
+            pageLoadingIndicator.apply {
+                if (progress > newProgress) {
+                    progress = 0
+                }
+                progressBarAnimation = ObjectAnimator.ofInt(this, "progress", newProgress).apply {
+                    duration = if (newProgress < MIN_PROGRESS_BAR) ANIM_DURATION_PROGRESS_LONG else ANIM_DURATION_PROGRESS_SHORT
+                    interpolator = AccelerateDecelerateInterpolator()
+                    addListener(onEnd = { if (!isLoading) hide() })
```
cmonfortep: neat! I liked this change to get rid off empty listeners

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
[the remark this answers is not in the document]
cmonfortep: ok, then based on you answer I would call it `shouldProgressDecrease`. As you said, we already see that we are checking which value is bigger but naming it using "why" we do that will help us.

marcosholgado:
@cmonfortep as discussed I have moved the code to an external class :)

PR #713, opened 2020-02-15
Dax dialog with two buttons

cmonfortep:
Task/Issue URL: https://app.asana.com/0/0/1162132608576977/f
Tech Design URL: 
CC: https://app.asana.com/0/0/1158941144031452/f

**Description**:
This PR is some preparation work that will be related (and be used) in a future task, where we will implement new Ctas with two buttons. 
For context: To show our CTAs as Dax dialogs we will make use of a Dax dialog with two buttons. Usually, the secondary button will be used to dismiss the dialog.
Design can be found here: https://app.asana.com/0/0/1158941144031452/f

Within the scope of this PR, you will find the following changes:
* Refactor current DaxDialog to handle a new secondary button
* Decoupling view-highlighting related logic from simple Dax dialogs into a more concrete dialog.
* Changes in UI according to new design
* Enable sending pixels when the secondary button is clicked
* Small improvement to avoid showing a dark status bar when the dialog appears

**Steps to test this PR**:
Currently, the new Dax dialog with two buttons is not used in any part of the project (but it will be). Testing the new dialog with 2 buttons will be included on a future PR once we have the experiment ready to be reviewed.

**Then, do we need to test something?** yes, ensure new changes don't affect how normal dialogs are shown. To testing that part:
1. Hardcode variant `mv`
2. Navigate to BrowserActivity
3. Visit any site that triggers a Dax dialog
4. Ensure that dialog appears and it's rendered as in previous versions (except for commented changes in the description).
Example:

> Possible triggers: a site with trackers, a site without trackers (wikipedia.org), a major entity (google).
> To test different Ctas, you may need to perform a fresh install for each type, as showing some of them will disable showing others.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
@@ -111,6 +111,12 @@ class CtaViewModel @Inject constructor(
         }
     }
 
+    fun onUserClickCtaSecondaryButton(cta: SecondaryButtonCta) {
+        cta.secondaryButtonPixel?.let {
+            pixel.fire(it, cta.pixelSecondaryButtonParameters())
+        }
+    }
+
     suspend fun refreshCta(dispatcher: CoroutineContext, isBrowserShowing: Boolean, site: Site? = null): Cta? {
         surveyCta()?.let {
             return it
--- a/app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
@@ -22,30 +22,44 @@ import android.graphics.Color
 import android.graphics.Rect
 import android.graphics.drawable.ColorDrawable
 import android.os.Bundle
+import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import androidx.fragment.app.DialogFragment
-import com.duckduckgo.app.browser.R
-import android.view.Gravity
-import kotlinx.android.synthetic.main.content_dax_dialog.*
 import android.view.animation.Animation
 import android.view.animation.OvershootInterpolator
 import android.view.animation.ScaleAnimation
 import android.widget.ImageView
 import android.widget.RelativeLayout
 import androidx.core.content.ContextCompat.getColor
+import androidx.fragment.app.DialogFragment
+import com.duckduckgo.app.browser.R
+import kotlinx.android.synthetic.main.content_dax_dialog.*
 
-class DaxDialog(
-    var daxText: String,
-    var buttonText: String,
+interface DaxDialog {
+    fun setDaxText(daxText: String)
+    fun setButtonText(buttonText: String)
+    fun setDialogAndStartAnimation()
+    fun onAnimationFinishedListener(onAnimationFinished: () -> Unit)
+    fun setPrimaryCtaClickListener(clickListener: () -> Unit)
+    fun setSecondaryCtaClickListener(clickListener: () -> Unit)
+    fun setHideClickListener(clickListener: () -> Unit)
+    fun setDismissListener(clickListener: () -> Unit)
+    fun getDaxDialog(): DialogFragment
+}
+
+class TypewriterDaxDialog(
+    private var daxText: String,
+    private var primaryButtonText: String,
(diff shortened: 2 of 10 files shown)


app/src/main/res/layout/content_dax_dialog.xml
```diff
@@ -84,61 +84,63 @@
...
-                    app:layout_constraintBottom_toBottomOf="parent"
-                    app:layout_constraintEnd_toEndOf="parent"
-                    app:layout_constraintStart_toStartOf="parent"
-                    app:layout_constraintTop_toBottomOf="@id/hiddenText" />
+                    app:cornerRadius="20dp" />
 
-            </androidx.constraintlayout.widget.ConstraintLayout>
+                <com.google.android.material.button.MaterialButton
+                    android:id="@+id/secondaryCta"
+                    style="@style/Widget.AppCompat.Button.Borderless.Colored"
```
malmstein: does it make sense to extract this to our own Button style?

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -111,6 +111,12 @@ class CtaViewModel @Inject constructor(
         }
     }
 
+    fun onUserClickCtaSecondaryButton(cta: SecondaryButtonCta) {
+        cta.secondaryButtonPixel?.let {
+            pixel.fire(it, cta.pixelSecondaryButtonParameters())
```
malmstein: is this ok being launched in the main thread?

malmstein:
If this is not going to be used for now, does it make sense to merge into develop or to keep a feature branch where you add multiple PRs? Just wondering if something were to change before it's finished, we'd have to change code that is potentially released to Play Store but is not being used yet.

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
cmonfortep: `ApiBasedPixel`:`Pixel` will excecute it in a background thread. So we are good here.

cmonfortep:
Thanks @malmstein for the review. About target branch on this PR, I'm happy with both approaches, I think both are valid but I will stick to the preferred one for the team. If it feels more natural creating a feature branch and merge everything there, I will move to that one and we can discuss this topic on our next mobile functional meetings.

marcosholgado:
Assigning to myself to test all the cta scenarios since I know them very well but @malmstein and others, feel free to do your reviews as well, I just want to make sure all the ctas still work as expected :)

subsymbolic:
@cmonfortep yes, please don't merge items not ready for release into develop.

app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
```diff
@@ -22,30 +22,44 @@ import android.graphics.Color
...
+import kotlinx.android.synthetic.main.content_dax_dialog.*
 
-class DaxDialog(
-    var daxText: String,
-    var buttonText: String,
+interface DaxDialog {
+    fun setDaxText(daxText: String)
+    fun setButtonText(buttonText: String)
+    fun setDialogAndStartAnimation()
+    fun onAnimationFinishedListener(onAnimationFinished: () -> Unit)
```
marcosholgado: The `onAnimationFinishedListener` is specific to the typing animation and is triggered when it finishes. I would suggest to move it to another interface or make it specific to the `TypewriteDaxDialog` class but this would make `DaxDialogHighlightView` a bit messier.
The other option is to assume the DaxDialog always is going to have the typing animation (which it does for now)  and maybe rename `TypewriteDaxDialog` to something more generic? Having `TypewriteDaxDialog` makes me think there is another version without the typing animation when in reality there isn't.
I don't think is massively important to stop the PR but it feels a very specific name for something that's always used like that.

cmonfortep: Actually, `onAnimationFinishedListener` can be moved into an interface and `DaxDialogHighlightView`will not be affected. The only part that will be affected is where we create those dialogs. But I liked that you commented on that. I think we can delay that change till we introduce a different Dax dialog, that is also the reason why I didn't segregate that interface into different ones.
As per the name, I'm fine changing the name or using this concrete one (because the interface takes the generic name), so I'm open to suggestions here. Tbh, I could come up with something else.

Issue #693, opened 2020-01-31
Should not be able to show or hide toolbar when typing in a website (when keyboard is open)


Poopooracoocoo:
It's pretty annoying.

I have no idea how Chrome for Android works with a keyboard and mouse but it is also worth thinking about those too.

tvalle:
Can I work on this one?

AndrewR8:
It seems that this issue needs to be fixed first #755

Cyberdroid1:
Well, since #755 has been fixed, would it be fixed now?

cmonfortep:
@Poopooracoocoo I think I understand what you mean hear, but just to be sure, could you post a gif or a video as sample? Also it would be nice to understand why it's annoying for you that behavior. Thanks.

Poopooracoocoo:
@cmonfortep Jesus, it's been a year and a half. I haven't used DDG Browser since then. 

It's worth just looking at Chrome's behaviour. Same applies for the other issues I created here.

cmonfortep:
It's ok, and sorry for the late response, just wanted to understand if there was something else I was missing.

We haven't received any further feedback related to the issue (neither from users using our embedded feedback option in settings). Closing this as there's no plan in the foreseeable future to support this request.

Poopooracoocoo:
ok but know that most users are just going to stop using the browser altogether and won't be bothered submitting feedback.

PR #721, opened 2020-02-22
 Fix Action bar hiding When IME is visible #693

tanoDxyz:
**App bar** will now remain visible and won't hide
      -> upon scrolling 
if the keyboard is open  and will resume it's default behavior on keyboard **close.**

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.duckduckgo.app.browser
+package com.duckduckgo.app.browser 
 
 import android.annotation.SuppressLint
 import android.app.ActivityOptions
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -38,6 +38,7 @@ import com.duckduckgo.app.feedback.ui.common.FeedbackActivity
 import com.duckduckgo.app.fire.DataClearer
 import com.duckduckgo.app.global.ApplicationClearDataState
 import com.duckduckgo.app.global.DuckDuckGoActivity
+import com.duckduckgo.app.global.KeyBoardVisibilityNotifier
 import com.duckduckgo.app.global.intentText
 import com.duckduckgo.app.global.view.*
 import com.duckduckgo.app.onboarding.ui.page.DefaultBrowserPage
@@ -54,6 +55,7 @@ import javax.inject.Inject
 
 class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
 
+    private lateinit var keyboardVisibilityNotifier: KeyBoardVisibilityNotifier
     @Inject
     lateinit var clearPersonalDataAction: ClearPersonalDataAction
 
@@ -97,17 +99,37 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
             renderer.renderBrowserViewState(it)
         })
         viewModel.awaitClearDataFinishedNotification()
+
+        keyboardVisibilityNotifier = KeyBoardVisibilityNotifier(this)
+
     }
 
+
+    override fun onResume() {
+        keyboardVisibilityNotifier.register(rootViewGroup,::onKeyboardStateChanged)
+        super.onResume()
+    }
+
+    private fun onKeyboardStateChanged(isOpen: Boolean) {
+        currentTab?.keyboardStateChanged(isOpen)
+    }
+
+    override fun onPause() {
+        keyboardVisibilityNotifier.unRegister(rootViewGroup)
+        super.onPause()
+    }
     override fun onStop() {
+        keyboardVisibilityNotifier.unRegister(rootViewGroup)
         openMessageInNewTabJob?.cancel()
         super.onStop()
     }
 
     override fun onDestroy() {
+        keyboardVisibilityNotifier.unRegister(rootViewGroup)
         currentTab = null
         super.onDestroy()
     }
+
 
     override fun onNewIntent(intent: Intent?) {
         super.onNewIntent(intent)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -54,6 +55,7 @@ import javax.inject.Inject
 
 class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
 
+    private lateinit var keyboardVisibilityNotifier: KeyBoardVisibilityNotifier
```
cmonfortep: This should be moved together with the rest of the private fields.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -97,18 +99,38 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
             renderer.renderBrowserViewState(it)
         })
         viewModel.awaitClearDataFinishedNotification()
+
+        keyboardVisibilityNotifier = KeyBoardVisibilityNotifier(this)
+
+    }
+
```
cmonfortep: Formatting should be fixed here.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -97,18 +99,38 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
...
+    private fun onKeyboardStateChanged(isOpen: Boolean) {
+        currentTab?.keyboardStateChanged(isOpen)
+    }
+
+    override fun onPause() {
+        keyboardVisibilityNotifier.unRegister(rootViewGroup)
+        super.onPause()
+    }
     override fun onStop() {
+        keyboardVisibilityNotifier.unRegister(rootViewGroup)
```
cmonfortep: I don't think we should put the same `unRegister` method call in all these 3 lifecycle methods. Usually, you want to unregister from the opposite lifecycle method where you registered the listener. That will mean that if you register `onResume` you unregister at `onPause`.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -97,18 +99,38 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
...
         super.onStop()
     }
 
     override fun onDestroy() {
+        keyboardVisibilityNotifier.unRegister(rootViewGroup)
         currentTab = null
         super.onDestroy()
     }
 
+
```
cmonfortep: Wrong file formating.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -97,18 +99,38 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
...
         })
         viewModel.awaitClearDataFinishedNotification()
+
+        keyboardVisibilityNotifier = KeyBoardVisibilityNotifier(this)
+
+    }
+
+
+    override fun onResume() {
+        keyboardVisibilityNotifier.register(rootViewGroup,::onKeyboardStateChanged)
```
cmonfortep: There should be an space before the second function param.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1124,9 +1136,10 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         private const val AUTHENTICATION_DIALOG_TAG = "AUTH_DIALOG_TAG"
         private const val DAX_DIALOG_DIALOG_TAG = "DAX_DIALOG_TAG"
 
+        private const val MIN_PROGRESS = 10
         private const val MAX_PROGRESS = 100
-        private const val TRACKERS_INI_DELAY = 500L
-        private const val TRACKERS_SECONDARY_DELAY = 200L
+        private const val TRACKERS_INI_DELAY = 700L
```
cmonfortep: Something weird happened here if all the following code appear as changed in this PR.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarScrolling.kt
```diff
@@ -25,17 +25,24 @@ import javax.inject.Inject
 
 class OmnibarScrolling @Inject constructor() {
```
cmonfortep: This file is wrongly formatted.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarScrolling.kt
```diff
@@ -25,17 +25,24 @@ import javax.inject.Inject
...
 
+
+    private var isScrollingEnabled = true
+
+    val isOmnibarScrollingEnabled:Boolean get() = isScrollingEnabled
```
cmonfortep: I would expect that I could safely call any of the methods in this class (`enableOmnibarScrolling`, `disableOmnibarScrolling`) without needing to check  `isOmnibarScrollingEnabled`. But not sure about this.

If so, I don't think we need this property here and we will avoid needing to update the value all the time.

app/src/main/java/com/duckduckgo/app/global/KeyBoardVisibilityNotifier
```diff
@@ -0,0 +1,53 @@
...
+
+package com.duckduckgo.app.global
+
+import android.content.Context
+import android.util.TypedValue
+import android.view.View
+import android.view.ViewTreeObserver
+
+
+class KeyBoardVisibilityNotifier(context: Context) {
```
cmonfortep: I am aware that currently there is no official method to detect that IME is shown in Android. According to some google devs, we should not checking IME visibility in their opinion, but I agree that there are some possible use cases where it could be interesting to do so.

I wonder if we should rely on this workaround (knowing that it's kind of hacky) or if we should just think about another approach to do the same that doesn't implies relying on the IME visibility.

For example, I can think about some possible scenarios where this may not work:
* Users using an external bluetooth keyboard
* How this reacts if user uses our app in split-screen mode

PR #734, opened 2020-03-04
Change app icon

malmstein:
Task/Issue URL: https://app.asana.com/0/72649045549333/1161651092205713

**Description**:
The iOS team saw a positive response from users after adding the feature that enabled users to change the app icon. We think that the same outcome can be achieved on the Android side.

**Disclaimer**:
It's known that this feature won't work on all devices. Depending on the launcher and OEM, it's possible that after selecting an icon it closes the app. If it doesn't, be aware that the icon change may take a few seconds before it's applied.

**Steps to test this PR**:
1. Open app 
2. Open Settings and Change App Icon
3. Select a new icon
4. Navigate to Apps Menu
5. Check new icon has been applied

**Screenshots**:
| Settings screen | Change Icon screen |
| ------ | ----- |
device-2020-02-28-164625|device-2020-03-04-103920 (1)


app/src/main/AndroidManifest.xml
```diff
@@ -11,7 +11,7 @@
     <application
         android:name="com.duckduckgo.app.global.DuckDuckGoApplication"
         android:allowBackup="false"
-        android:icon="@mipmap/ic_launcher"
+        android:icon="${appIcon}"
```
malmstein: We can use manifest placeholders to change the app icon based on the build version. This is done to replace the need of maintaining a separate folder in the source.

app/build.gradle
```diff
@@ -45,11 +45,19 @@ android {
         debug {
             applicationIdSuffix ".debug"
             pseudoLocalesEnabled false
+            manifestPlaceholders = [
```
malmstein: We can use manifest placeholders to change the app icon based on the build version. This is done to replace the need of maintaining a separate folder in the source.

CDRussell: I'm no longer seeing the blue icon for DEBUG builds. Any thoughts why?

Pixel 2; Android R

app/src/androidTest/java/com/duckduckgo/app/icon/ui/ChangeIconViewModelTest.kt
```diff
@@ -0,0 +1,97 @@
...
+
+        Mockito.verify(mockPixel).fire(Pixel.PixelName.CHANGE_APP_ICON_OPENED)
+
+        val viewState = testee.viewState.value!!
+        assertTrue(viewState.appIcons.isNotEmpty())
+        assertTrue(viewState.appIcons.find { it.selected }!!.appIcon.componentName == selectedIcon.componentName)
+    }
+
+    @Test
+    fun storesIconWhenUserSelectsIt() {
```
CDRussell: Preferred test naming format is `whenXThenY()`

app/src/main/java/com/duckduckgo/app/di/VariantModule.kt
```diff
@@ -35,4 +39,8 @@ class VariantModule {
 
     @Provides
     fun weightedRandomizer() = WeightedRandomizer()
+
+    @Provides
+    fun appIconModifier(context: Context, appShortcutCreator: AppShortcutCreator): IconModifier =
```
CDRussell: minor: this seems misplaced being in the `VariantModule`

app/src/main/java/com/duckduckgo/app/icon/api/AppIconModifier.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val context: Context,
+    private val appShortcutCreator: AppShortcutCreator
+) : IconModifier {
+
+    override fun changeIcon(previousIcon: AppIcon, newIcon: AppIcon) {
+        enable(context, newIcon)
+        disable(context, previousIcon)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+            appShortcutCreator.configureAppShortcuts(context)
```
CDRussell: I'd been thinking it was good the app shortcuts changed icon for free, but see now you made it happen. 👍

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -75,6 +77,13 @@ class SettingsSharedPreferences @Inject constructor(private val context: Context
         get() = preferences.getBoolean(KEY_AUTOCOMPLETE_ENABLED, true)
         set(enabled) = preferences.edit { putBoolean(KEY_AUTOCOMPLETE_ENABLED, enabled) }
 
+    override var appIcon: AppIcon
+        get() {
+            val componentName = preferences.getString(KEY_APP_ICON, AppIcon.DEFAULT.componentName) ?: return AppIcon.DEFAULT
+            return AppIcon.from(componentName)
+        }
+        set(appIcon) = preferences.edit { putString(KEY_APP_ICON, appIcon.componentName) }
```
CDRussell: Worth trying `preferences.edit(commit = true)` here since we know the app might crash soon after? Maybe try ensure it has time to flush the changes to disk.

app/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
```diff
@@ -155,7 +155,9 @@ interface Pixel {
         FEEDBACK_NEGATIVE_SUBMISSION("mfbs_%s_%s_%s"),
 
         AUTOCOMPLETE_BOOKMARK_SELECTION("m_aut_s_b"),
-        AUTOCOMPLETE_SEARCH_SELECTION("m_aut_s_s")
+        AUTOCOMPLETE_SEARCH_SELECTION("m_aut_s_s"),
+
+        CHANGE_APP_ICON_OPENED("mic"),
```
CDRussell: We should be using `m_` prefix for all new pixels unless there's a good reason not to

CDRussell:
Screen Shot 2020-03-05 at 9 14 08 AM

The amount of launcher icon assets we have now is huge. This might be the only way to do it, but thought I'd ask the question anyway to question assumptions: do we still need to provide all these `PNG` or could any of them be replaced with `XML` instead?

app/src/main/java/com/duckduckgo/app/di/VariantModule.kt
malmstein: I didn't want to create a Module just for this, probably `SystemComponentsModule` would be better

malmstein:
Unfortunately not @CDRussell. Adaptive icons are [only available API level 26 and above](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive). If and when we have minSDK 26 we can get rid of those pngs.

app/src/main/java/com/duckduckgo/app/fire/AutomaticDataClearer.kt
```diff
@@ -171,13 +182,19 @@ class AutomaticDataClearer(
         }
     }
 
-    private fun shouldClearData(cleanWhenOption: ClearWhenOption, appUsedSinceLastClear: Boolean): Boolean {
+    private fun shouldClearData(cleanWhenOption: ClearWhenOption, appUsedSinceLastClear: Boolean, appIconChanged: Boolean): Boolean {
         Timber.d("Determining if data should be cleared for option $cleanWhenOption")
 
+        if (appIconChanged) {
```
CDRussell: I like the way this is integrated into this class; feels like a natural place for it 👍

app/src/main/java/com/duckduckgo/app/launch/LaunchBridgeActivity.kt
```diff
@@ -68,11 +68,11 @@ class LaunchBridgeActivity : DuckDuckGoActivity() {
         } else {
             startActivity(OnboardingActivity.intent(this))
         }
-        finish()
+        // finish()
```
CDRussell: Can we kill these commented out lines? Also, if the intention is to delete it, why don't we call `finish()`?

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -85,6 +86,10 @@ class SettingsSharedPreferences @Inject constructor(private val context: Context
         }
         set(appIcon) = preferences.edit(commit = true) { putString(KEY_APP_ICON, appIcon.componentName) }
 
+    override var appIconChanged: Boolean
```
CDRussell: Could we add a comment here explaining this flag? i just know otherwise i'm going to forget it's a temporary flag just across the app restart

app/build.gradle
malmstein: This is now fixed, thanks for noticing!

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
malmstein: 👍

app/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
malmstein: Changed it to `m_ic` 👍

app/src/main/java/com/duckduckgo/app/launch/LaunchBridgeActivity.kt
malmstein: This was just a test, good catch!

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
malmstein: Yeah that make sense, I've added a comment explaining it

malmstein:
@CDRussell I've fixed the issue with the flag after app close. The tests were missing the flag `appUsedSinceLastClear` so they were green without having to be. I've added a few extra tests to make sure we are on the right side of things

PR #737, opened 2020-03-06
Fix space deletion from user query

subsymbolic:
Task/Issue URL: https://app.asana.com/0/72649045549333/1163849208167805/f

**Description**:
Fix issue where view state would update and delete space from user query

**Steps to test this PR**:
1. Install the app and add the widget to your home screen
1. Tap the widget to launch the interstitial search screen
1. Type a few words really fast and check that the spaces are not being deleted.

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/systemsearch/SystemSearchViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/systemsearch/SystemSearchViewModelTest.kt
@@ -86,6 +86,18 @@ class SystemSearchViewModelTest {
     }
 
     @Test
+    fun whenUserAddsSpaceToQueryThenViewStateMatchesAndSpaceTrimmedFromAutocomplete() = ruleRunBlockingTest {
+        testee.userUpdatedQuery(QUERY)
+        testee.userUpdatedQuery("$QUERY ")
+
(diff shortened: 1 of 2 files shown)


subsymbolic:
Thanks for the review @CDRussell!

JeffAment:
@subsymbolic,
Just wanted to give you a heads up that a similar issue still exists, but only when autocorrect does its magic or when using autosuggestions. :cold_sweat:

JeffAment:
I also just noticed some odd behavior when holding backspace to delete a word; sometimes only the first space after the word will be deleted, and other times the word will be deleted but reappear a few ms later.  I could open an issue on these problems if desired.

subsymbolic:
Oh, great feedback, thanks @JeffAment. We're releasing 5.46.0 right now so I'll include tweaks for this in the subsequent release - the main change will be stopping the view model updating the text view other when the clear button is pressed as it's causing side-effects.

subsymbolic:
@JeffAment I couldn't reproduce this but had some suspicions and hopefully fixed it in PR https://github.com/duckduckgo/Android/pull/773 🤞🤞🤞

If you have time to test it and let us know if it's fixed, that would be amazing. If not, that's ok too, we'll monitor for wider community feedback once it's out.

PR #747, opened 2020-03-20
Hotfix/5.47.1.fixes

malmstein:
Task/Issue URL:  https://app.asana.com/0/inbox/1157893581871899/1119619712088571/1167491198050484

**Description**:
Renaming `NotificationScheduler` introduced a bug that made it impossible for `DaggerWorkerFactory`  to create Workers. The classpath had changed and therefore no longer possible to find.

This PR fixes this issue, also adding a layer of security in `DaggerWorkerFactory` preventing it from crashing if the worker can't be found.

As part of this PR we also fixed a new issue that was introduced with the Change App Icon feature. On some devices the `themeResourceId` was not found after changing the icon, so we added a default option for that too based on the `AppTheme`

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/di/DaggerWorkerFactory.kt
+++ b/app/src/main/java/com/duckduckgo/app/di/DaggerWorkerFactory.kt
@@ -23,7 +23,11 @@ import androidx.work.WorkerFactory
 import androidx.work.WorkerParameters
 import com.duckduckgo.app.fire.DataClearingWorker
 import com.duckduckgo.app.global.view.ClearDataAction
-import com.duckduckgo.app.notification.AndroidNotificationScheduler.*
+import com.duckduckgo.app.notification.NotificationScheduler.ClearDataNotificationWorker
+import com.duckduckgo.app.notification.NotificationScheduler.PrivacyNotificationWorker
+import com.duckduckgo.app.notification.NotificationScheduler.SearchPromptNotificationWorker
+import com.duckduckgo.app.notification.NotificationScheduler.StickySearchNotificationWorker
+import com.duckduckgo.app.notification.NotificationScheduler.DismissSearchNotificationWorker
 import com.duckduckgo.app.notification.NotificationFactory
 import com.duckduckgo.app.notification.db.NotificationDao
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
```diff
@@ -66,7 +63,7 @@ class AndroidNotificationScheduler(
     }
 
     private suspend fun scheduleInactiveUserNotifications() {
-        workManager.cancelAllWorkByTag(UNUSED_APP_WORK_REQUEST_TAG)
+        workManager.cancelUniqueWork(UNUSED_APP_WORK_REQUEST_TAG)
```
CDRussell: is this change deliberate?

app/src/main/java/com/duckduckgo/app/di/DaggerWorkerFactory.kt
```diff
@@ -52,22 +56,28 @@ class DaggerWorkerFactory(
...
+                is PrivacyNotificationWorker -> injectPrivacyNotificationWorker(instance)
+                is SearchPromptNotificationWorker -> injectSearchPromptNotificationWorker(instance)
+                is StickySearchNotificationWorker -> injectStickySearchNotificationWorker(instance)
+                is DismissSearchNotificationWorker -> injectDismissSearchNotificationWorker(instance)
+                else -> Timber.i("No injection required for worker $workerClassName")
+            }
+
+            return instance
+        } catch (exception: Exception){
+            Timber.i("Worker $workerClassName could not be created")
```
CDRussell: Could switch to a higher log level. maybe `e` log level. and include the exception too?

`Timber.e(exception, "....")`

cmonfortep:
Taking this one @CDRussell @malmstein.

app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
malmstein: nope, reverted

PR #750, opened 2020-03-27
Feature/david/pull to refresh

malmstein:
Task/Issue URL:  https://app.asana.com/0/488551667048375/1168653724774521

**Description**:
Pull to Refresh is a very common pattern on Android when it comes to refresh the content in screen. This PR enables this mechanism, without removing the current option for refreshing**Steps to test 

**Screencast**:

| Pull to Refresh | Refresh from Menu |
| ------ | ----- | 
device-2020-03-27-142605|device-2020-03-27-143149

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -23,20 +23,37 @@ import android.annotation.SuppressLint
 import android.app.Activity.RESULT_OK
 import android.app.ActivityOptions
 import android.appwidget.AppWidgetManager
-import android.content.*
+import android.content.ClipData
+import android.content.ClipboardManager
+import android.content.ComponentName
+import android.content.Context
+import android.content.Intent
 import android.content.pm.PackageManager
 import android.content.res.Configuration
 import android.media.MediaScannerConnection
 import android.net.Uri
-import android.os.*
+import android.os.Build
+import android.os.Bundle
+import android.os.Environment
+import android.os.Handler
+import android.os.Message
 import android.text.Editable
-import android.view.*
+import android.view.ContextMenu
+import android.view.KeyEvent
+import android.view.LayoutInflater
+import android.view.MenuItem
+import android.view.View
 import android.view.View.*
+import android.view.ViewGroup
 import android.view.inputmethod.EditorInfo
-import android.webkit.*
+import android.webkit.ValueCallback
+import android.webkit.WebChromeClient
+import android.webkit.WebSettings
+import android.webkit.WebView
 import android.webkit.WebView.FindListener
 import android.webkit.WebView.HitTestResult
 import android.webkit.WebView.HitTestResult.*
+import android.webkit.WebViewDatabase
 import android.widget.EditText
 import android.widget.TextView
 import android.widget.Toast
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/fragment_browser_tab.xml
```diff
@@ -19,8 +19,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:id="@+id/rootView"
     android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    app:layout_scrollFlags="scroll|enterAlways">
```
malmstein: these flags don't affect the behaviour, it's set manually by `OmnibarScrolling`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -23,20 +23,37 @@ import android.annotation.SuppressLint
 import android.app.Activity.RESULT_OK
 import android.app.ActivityOptions
 import android.appwidget.AppWidgetManager
-import android.content.*
+import android.content.ClipData
```
CDRussell: There shouldn't be all of these changes to the import statements; when i format this file it reverts them back again and this is going to create lots of noisy PRs.

Can you check your code formatting settings to ensure they match the team settings? I can work with you offline if they're out of sync.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -462,6 +558,7 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
     }
 
     fun refresh() {
+        swipeContainer.isRefreshing = true
```
CDRussell: >Do not call this when refresh is triggered by a swipe gesture.

The documentation on `SwipeRefreshLayout` suggests we shouldn't set this here?

app/src/main/java/com/duckduckgo/app/browser/ui/ScrollAwareSwipeToRefreshLayout.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    interface CanWebViewScrollUpCallback {
+        fun canSwipeRefreshChildScrollUp(): Boolean
+    }
+
+    fun setCanWebViewScrollUpCallback(canWebViewScrollUpCallback: CanWebViewScrollUpCallback) {
+        this.canWebViewScrollUpCallback = canWebViewScrollUpCallback
+    }
+
+    override fun canChildScrollUp(): Boolean {
```
CDRussell: Instead of the `!!`, what about if changed the function like this?

     override fun canChildScrollUp(): Boolean {
        return canWebViewScrollUpCallback?.canSwipeRefreshChildScrollUp() ?: super.canChildScrollUp()
    }

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
malmstein: This is triggered when the button in the Popup Menu is pressed, so it's necessary.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
malmstein: I'm afraid those guidelines don't mention anything about Imports. On the other hand, Unused imports is a well know PMD warning (https://pmd.github.io/latest/pmd_rules_java_bestpractices.html#unusedimports) Wildcard imports are unnecessary, why is that better?

malmstein:
Closing this until we agree on UX flow with Design. Thanks for the review @CDRussell!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
CDRussell: >Wildcard imports are unnecessary, why is that better?

It's not a case of which is better; it's about not generating a ton of changes every time the code is touched by different people. I've no strong opinions on whether to wildcard imports or not, per se, but i've a very strong opinion we all need to be doing the same thing.

PR #756, opened 2020-04-01
Android: Dax Dialogs - quick improvements on big screens

cmonfortep:
Task/Issue URL: https://app.asana.com/0/0/1169355936992923/f
Tech Design URL: https://app.asana.com/0/0/1169315285491563/f
CC: 

**Description**:
When making the new onboarding and Dax dialogs the new default experience, we noticed that the UI for some big screens (specially tablets) could be improved. To avoid delaying the release, we decided just to introduce a few quick improvements. Those include:
- introducing a max width of 600dp for all devices
- centering the dialog

**Onboarding dialog**
Screenshot_1585744086
Screenshot_1585744094

**Intro dialog:**
Screenshot_1585744107
Screenshot_1585744114

**Dax dialog:**
Screenshot_1585744125
Screenshot_1585744130

**Steps to test this PR**:
1. Fresh install on a big screen device -> use a tablet
1. Ensure this first dialog during the onboarding process is shown as the attached image
1. Navigate to the browser screen
1. Ensure intro dialog is shown as the attached image (test it in both modes portrait and landscape)
1. Visit a site to trigger any Dax dialog (cnn.com)
1. Ensure Dax dialog si shown as the attached image (test it in both modes portrait and landscape)

Additional test:
1. you can try running the same flow on a phone
1. Ensure everything looks as before, except for some devices in landscape mode where the same rules should apply (max 600dp and centered)

Side note: COVID Cta stays as it is right now.

---
###### Internal references:

### DIFF
--- a/app/src/main/res/layout/include_new_browser_tab.xml
+++ b/app/src/main/res/layout/include_new_browser_tab.xml
@@ -57,7 +57,6 @@
         android:layout_width="0dp"
         android:layout_height="wrap_content"
         android:layout_marginTop="40dp"
-        app:layout_constraintHorizontal_bias="0"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent"
--- a/app/src/main/res/layout/include_new_browser_tab.xml
+++ b/app/src/main/res/layout/include_new_browser_tab.xml
@@ -54,12 +54,14 @@
 
     <include
         layout="@layout/include_dax_dialog_cta"
-        android:layout_width="match_parent"
-        android:layout_height="0dp"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
         android:layout_marginTop="40dp"
+        app:layout_constraintHorizontal_bias="0"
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/include_new_browser_tab.xml
```diff
@@ -54,12 +54,13 @@
...
-        android:layout_width="match_parent"
-        android:layout_height="0dp"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
         android:layout_marginTop="40dp"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toTopOf="parent" />
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintWidth_max="600dp"/>
```
malmstein, line 63: you are using this value 4 times, maybe it makes sense to extract as a dimension?

cmonfortep: Good point, I know they are the same number, but it's just because the same size worked well to constraint each dialog for the different screens. But there's no indication that if at some point we need to tweak one of those dialogs, the rest should change too. So they are not officially related between them.

Issue #755, opened 2020-03-30
Address bar loses focus when scrolling web page.


AndrewR8:
**Steps to reproduce**:

1) Open any web page.
1) Click on address bar.
2) Start scrolling web page.

**Expected result**:
I'm not sure.
It seems to me one of the following should happen:
1) Address bar loses focus and the keyboard disappears.
2) Address bar keeps focus.

**Actual result**:
Address bar loses focus but the keyboard does not hide.

cmonfortep:
Thanks for reporting this @AndrewR8 I think the **expected result proposal 1** could fit pretty well here.

cmonfortep:
Closing this issue, thanks for the contribution @AndrewR8

PR #758, opened 2020-04-02
Keyboards hides when address bar loses focus

AndrewR8:
issue #755


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -735,6 +735,7 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         omnibarTextInput.onFocusChangeListener =
             OnFocusChangeListener { _, hasFocus: Boolean ->
                 viewModel.onOmnibarInputStateChanged(omnibarTextInput.text.toString(), hasFocus, false)
+                if(!hasFocus)omnibarTextInput.hideKeyboard()
```
cmonfortep, line 738: NIT: better to use `hideKeyboardImmediately()` or `hideKeyboard()` that does the same as your code + request focus on a dummy view.

An additional note: please use curly brackets even if it's a single line of code, and ensure code follows our project style guide.

AndrewR8: @cmonfortep  just a question: why we request focus on a dummy view if this function is already requested in `hideKeyboard()`?

PR #764, opened 2020-04-05
Ensure navigation buttons are visible when BottomSheetDialog is displayed

### LINKED ISSUE #754: Navigation buttons hidden on some devices
**Background**
After clicking on fire button, a bottom dialog appears to request user confirmation to clear all tabs and data.

**To reproduce:**
- Install app
- Go to settings and switch to light theme (if you're not already on it)
- Go to the browser and press the fire button

Seen on:
Pixel 3 XL device, API 29
Pixel XL emulator, API 29

**Actual behavior:** Fire dialog appears and however system buttons disappear. See following screenshot:

**Expected behavior:** Fire dialog appears and system buttons are visible. Example, see how in the following screenshot navigation buttons are visible:

JeffAment:
**Description**:
When the fire button is clicked on the light theme, the **navigation buttons**, as well as the **icons in the action bar**, are hard, if not impossible, to see on Android 8+.  This PR brings the UI focus loss in line with what one may expect from a normal `Dialog` etc.  Attempts to fix #754

**Steps to test this PR**:
1. Ensure that the navigation buttons and actionbar icons are _clearly_ visible on light and dark theme with intended contrast.

Below are two screenshots from a **Pixel 3XL API 29**: the first is a picture of what the navbar and actionbar will look like after this PR, and the second is of an alternative UI candidate that does not dim the navbar since I thought it looked decent and could be another acceptable way of handling this since `BottomSheetDialogs` are not floating dialogs hence the navbar doesn't _necessarily_ need to be dimmed here in my opinion.  I would be happy to submit a PR implementing the second approach if desired.

Screenshot_1586118185

Screenshot_1586129038

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
@@ -17,6 +17,7 @@
 package com.duckduckgo.app.global.view
 
 import android.content.Context
+import android.os.Build
 import android.os.Bundle
 import android.view.View
 import com.duckduckgo.app.browser.R
@@ -33,6 +34,10 @@ class FireDialog(context: Context, private val clearPersonalDataAction: ClearPer
     init {
         val contentView = View.inflate(context, R.layout.sheet_fire_clear_data, null)
         setContentView(contentView)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            window?.decorView?.systemUiVisibility = View.NOT_FOCUSABLE
+        }
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {


app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
```diff
@@ -33,6 +34,10 @@ class FireDialog(context: Context, private val clearPersonalDataAction: ClearPer
     init {
         val contentView = View.inflate(context, R.layout.sheet_fire_clear_data, null)
         setContentView(contentView)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            window?.decorView?.systemUiVisibility = View.NOT_FOCUSABLE
```
cmonfortep, line 39: Not sure why this fixes the issue, but I would prefer to rely on themes / styles if possible. I did a quick look and found this article interesting https://medium.com/halcyon-mobile/implementing-googles-refreshed-modal-bottom-sheet-4e76cb5de65b 
I think modifying our `BottomSheetDialog` style and adding the following should work:
```
<item name="android:windowIsFloating">false</item>
<item name="android:statusBarColor">@android:color/transparent</item>
```

JeffAment:
Thank you for the review, and for brainstorming with me, @cmonfortep!

> I would prefer to rely on themes / styles if possible.

This was my initial thought as well.  I will elaborate on why I made changes in `FireDialog.kt` below.

> About making the navBar white...

The second approach would not always have a white navbar; the navbar would stay the color that it normally is for that particular theme.

> I think modifying our BottomSheetDialog style and adding the following should work:

I did try this, but this seems to make the navbar black.  I'm not sure if my logic is correct here, but I will try to outline the logic behind my decision to modify `FireDialog.kt`:

- I was thinking that `Theme.Design.BottomSheetDialog`, which `BottomSheetDialog` inherits from in `styles.xml`, possibly applies its own static theming here, similar to how a custom keyboard might.  When I add those two lines you mentioned, the navbar becomes black.  I also thought that it was possible that there could be a certain attribute in `themes.xml` that `BottomSheetDialog` pulls from that we were not expecting it to use, so I changed every color to `cornflowerBlue`, but the navbar still turned black when the fire button was clicked, again, much like a keyboard may ignore themes and apply its own.
- I tried changing what `BottomSheetDialog` inherits from to many different parents for testing, which include `ThemeOverlay.*.BottomSheetDialog`, `Theme.*.Dialog` etc.  These didn't change much as far as the proper focus loss of the navbar, but eliminated the black navbar that `Theme.Design.BottomSheetDialog` had when putting those two lines in `styles.xml`.
- As another test, I changed `FireDialog.kt` to inherit from `Dialog` instead of `BottomSheetDialog`.  After I did this, the focus loss looked correct.  Since making `BottomSheetDialog` inherit from `Dialog` in `styles.xml` didn't change the focus loss, but making `FireDialog.kt` inherit from `Dialog` did, I figured maybe the best place for these changes would be in `FireDialog.kt`.
- Again, I am not sure if this logic is sound, but this is my thought process here.  It honestly sounds to me like something that Google may need to fix in their libraries (I also tested this issue against the "Update app dependencies" PR that Craig recently submitted to test if anything changed, but the bug still persists).  I believe the only way to fix this may be to manually override some colors in `styles.xml` (or maybe adding missing attributes in `themes.xml`?), but I think that may be much messier than modifying `FireDialog.kt` and letting Android take care of it for us.  We can also keep the navbar in its normal, undimmed, state and preserve the color by changing the parent of `BottomSheetDialog` in `styles.xml` to one that does not apply its own theming if that is preferable.

Even in that article you linked, written pre Android 10, it seems as though the navigation buttons and actionbar icons did not dynamically change color when focus was lost, and the contrast looks similar, if not identical, to how the DDG app currently looks on Android 9 and below.  Happy to take a closer look at this issue if you think there is something I could be missing here, and that a more optimal solution exists.

Sorry for the long reply!

cmonfortep:
Thanks @JeffAment, it could be true that changing focus triggers a different theme to be applied, I didn't think about that. Still, I think it worths to explore setting this through themes, just for consistency with how we are used to work with other view components, it will be the first place any dev will try to look at when trying to understand this.

We did some work trying to change the nav bar color and decided to give support only to api27 and above, you can see it inside the description of this PR https://github.com/duckduckgo/Android/pull/726

About theme proposed solution, I want to remark that `android:statusBarColor` doesn't set the nav bar color but the one from the top, I don't know if you were trying to set the color to cornFlowerBlue using that attribute.

if navbar is black with that solution, I think is okish. Maybe we can investigate a bit more using `android:navigationBarColor` for API>27 like we did in the PR I was talking about, and see if it works.

cmonfortep:
@JeffAment We got a fix for this issue and we can close this PR. I wanted to thank you for the work you did here (and in many other contributions). Looking forward to future contributions from you, they are always very welcome here. 🖖

JeffAment:
@cmonfortep,

Happy to hear that a fix has been merged :tada:.  I am sorry I couldn't have been more prompt here; life circumstances and a hectic work schedule sort of got in the way.

All the best and stay well :rocket: :v:

PR #765, opened 2020-04-06
Animator helper crash

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1170001449666321
Tech Design URL: 
CC: 

**Description**:
In some cases when the trackers blocked CTA is shown it could lead to a crash when the tracker blocked animation finishes if some of the views have been killed and not longer exist.

**Steps to test this PR**:
**App crashes**
1. In the developer options set `Don't keep activities` to true.
1. Launch the existing app from develop.
1. Navigate to cnn.com and wait until the trackers blocked CTA is shown.
1. Send the app to the background and see a crash.

**App does not crash**
1. In the developer options set `Don't keep activities` to true.
1. Launch the app from this branch.
1. Navigate to cnn.com and wait until the trackers blocked CTA is shown.
1. Send the app to the background and the crash does not happen anymore.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
@@ -18,7 +18,6 @@ package com.duckduckgo.app.browser
 
 import android.animation.AnimatorSet
 import android.animation.ObjectAnimator
-import android.annotation.SuppressLint
 import android.app.Activity
 import android.graphics.drawable.AnimatedVectorDrawable
 import android.view.View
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
```diff
@@ -187,17 +186,23 @@ class BrowserTrackersAnimatorHelper {
         }
     }
 
-    @SuppressLint("ObjectAnimatorBinding")
-    fun animateFadeOut(view: Any): ObjectAnimator {
-        return ObjectAnimator.ofFloat(view, "alpha", 1f, 0f).apply {
-            duration = DEFAULT_ANIMATION_DURATION
+    private fun animateFadeOut(view: View?): ObjectAnimator? {
```
cmonfortep, line 189: As a quick fix, that guard is ok.
But, what's the root cause of that null? are the animations being canceled when the view is destroyed? is there something we should be checking before that null gets there?

marcosholgado: So... in some cases what happens is the activity gets destroyed, if the dialog is being shown it then triggers `onDismissDialog` which is executed and causes the crash. 

I have further changed the code and I think it makes more sense now.

PR #769, opened 2020-04-08
Optimize onboarding CTAs

marcosholgado:
Task/Issue URL: https://app.asana.com/0/608920331025315/1170386777246553
Tech Design URL: https://app.asana.com/0/608920331025315/1170386262731497
CC: 

**Description**:
This PR introduces new copy changes and dialog amends to optimize their content, that means
- Removing the highlight dax dialog since is not used anymore.
- Simplified logic around network dialogs since they now have a single step.
- New copy

Apart from that this PR also fixes a bug where after destroying the fragment while showing a fragment would cause a crash after relaunching the app.
- Listeners moved to the fragment.
- 0 param constructor implemented to avoid the crash.
- dialog removed during onCreate to avoid secondary effects.

**Steps to test this PR**:
**No trackers new copy**
1. Re-install the app
1. Go to wikipedia.org.
1. New copy should read `As you tap and scroll, I'll block pesky trackers. Go ahead — keep browsing!`

**Network new copy**
1. Re-install the app
1. Go to facebook.com or google.com
1. New copy should read `Heads up! %s is a major tracking network. Their trackers lurk on about %s of top sites 😱 but don't worry! I'll block %s from seeing your activity on those sites.`
1. Clicking in `Got It` closes the dialog and does not bring up a second one.

**Owned by network new copy**
1. Re-install the app
1. Go to instagram.com or youtube.com
1. New copy should read `Heads up! %s is owned by %s. %s\'s trackers lurk on about %s of top websites 😱 but don't worry! I'll block %s from seeing your activity on those sites.`
1. Clicking in `Got It` closes the dialog and does not bring up a second one.

**Rotation with dialog**
1. Re-install the app
1. Go to wikipedia.org.
1. Rotate the phone.
1. Dialog should still show without being created again.

**Send the app to background with dialog**
1. Re-install the app
1. Go to wikipedia.org.
1. Press the home button.
1. Relaunch the app again, the dialog should still show without being created again.

**Listeners still work**
1. Re-install the app
1. Go to cnn.com.
1. Dismissing the dialog still triggers finishing the trackers animation.
1. Refresh the site, dialog does not show again.
1. Re-install the app
1. Go to wikipedia.org.
1. Press hide and hide tips forever.
1. Go to facebook.com, dialogs should not be shown anymore.

**Null pointer crash fixed**
1. In the developer options set Don't keep activities to true
1. Re-install and launch the app.
1. Navigate to cnn.com and wait until the trackers blocked CTA is shown.
1. Send the app to the background and there is no crash.
1. Relaunch the app.
1. The dialog is not immediately shown.
1. After the site loads the dialog is shown again.

**Double dialogs**
1. In the developer options set Don't keep activities back to false
1. Re-install and launch the app.
1. Navigate to SERP and wait until the CTA is shown.
1. Send the app to the background.
1. Open a URL externally with DDG (i.e. go to your email, long click on a URL and select open in browser)
1. Notice how the previous dialog disappears and a new one is shown for the new site.

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1115,6 +1120,44 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         startActivity(AddWidgetInstructionsActivity.intent(context), options)
     }
 
+    override fun onDismissDialog() {
+        val configuration = viewModel.ctaViewState.value?.cta ?: return
```
malmstein, line 1124: Should this logic be moved to the `ViewModel` in a method called `onDialogDismissed()`?. That way it could also be tested

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1115,6 +1120,44 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
+        if (configuration is DaxDialogCta.DaxTrackersBlockedCta) {
+            animatorHelper.finishTrackerAnimation(omnibarViews(), networksContainer)
+        }
+        viewModel.onUserDismissedCta(configuration)
+    }
+
+    override fun onHideClick() {
+        context?.let {
+            val configuration = viewModel.ctaViewState.value?.cta ?: return
```
malmstein, line 1133: same here, this could be moved to the `ViewModel`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1115,6 +1120,44 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
+            launchHideTipsDialog(it, configuration)
+        }
+    }
+
+    override fun onPrimaryCtaClick() {
+        val configuration = viewModel.ctaViewState.value?.cta ?: return
```
malmstein, line 1139: ditto

app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
```diff
@@ -40,28 +35,33 @@ interface DaxDialog {
...
+    private var dismissible: Boolean = true
+    private var typingDelayInMs: Long = 20
+
+    private var dialogListeners: DaxDialogListeners? = null
+
+    override fun onAttach(context: Context) {
+        super.onAttach(context)
+        dialogListeners = parentFragment as? DaxDialogListeners
+        if (dialogListeners == null) {
+            throw ClassCastException("$parentFragment must implement DaxDialogListeners")
```
malmstein, line 62: 👍

CDRussell:
Following the steps to test, i didn't see the dax dialog for **Owned by network new copy**

CDRussell:
In the "Listeners still work" scenario, when i follow the steps listed, I don't see any dialog on wikipedia

app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
```diff
@@ -40,28 +35,33 @@ interface DaxDialog {
...
+class TypewriterDaxDialog : DialogFragment(), DaxDialog {
+
+    private lateinit var daxText: String
+    private lateinit var primaryButtonText: String
+    private var secondaryButtonText: String = ""
+    private var toolbarDimmed: Boolean = true
+    private var dismissible: Boolean = true
+    private var typingDelayInMs: Long = 20
+
+    private var dialogListeners: DaxDialogListeners? = null
```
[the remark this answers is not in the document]
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
```diff
@@ -40,28 +35,33 @@ interface DaxDialog {
...
-
-    private var onAnimationFinished: () -> Unit = {}
-    private var primaryCtaClickListener: () -> Unit = { dismiss() }
-    private var secondaryCtaClickListener: (() -> Unit)? = null
-    private var hideClickListener: () -> Unit = { dismiss() }
-    private var dismissListener: () -> Unit = { }
+interface DaxDialogListeners {
+    fun onDismissDialog()
+    fun onPrimaryCtaClick()
+    fun onHideClick()
```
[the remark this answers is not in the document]
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1384,58 +1425,15 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
...
-        private fun launchHideTipsDialog(context: Context, cta: Cta) {
-            AlertDialog.Builder(context)
-                .setTitle(R.string.hideTipsTitle)
-                .setMessage(getString(R.string.hideTipsText))
-                .setPositiveButton(R.string.hideTipsButton) { dialog, _ ->
-                    dialog.dismiss()
-                    launch {
-                        ctaViewModel.hideTipsForever(cta)
-                    }
+                val savedDaxDialog: Fragment? = childFragmentManager.findFragmentByTag(DAX_DIALOG_DIALOG_TAG)
```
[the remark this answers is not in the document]
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -281,6 +280,13 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         })
     }
 
+    private fun removeDaxDialogFromManager() {
+        val fragment: Fragment? = childFragmentManager.findFragmentByTag(DAX_DIALOG_DIALOG_TAG)
+        fragment?.let {
+            childFragmentManager.beginTransaction().remove(it).commit()
```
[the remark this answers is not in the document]
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
```diff
@@ -40,28 +35,33 @@ interface DaxDialog {
...
-    private val dismissible: Boolean = true,
-    private val typingDelayInMs: Long = 20
-) : DialogFragment(), DaxDialog {
-
-    private var onAnimationFinished: () -> Unit = {}
-    private var primaryCtaClickListener: () -> Unit = { dismiss() }
-    private var secondaryCtaClickListener: (() -> Unit)? = null
-    private var hideClickListener: () -> Unit = { dismiss() }
-    private var dismissListener: () -> Unit = { }
+interface DaxDialogListeners {
```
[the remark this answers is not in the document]
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/global/view/DaxDialog.kt
```diff
@@ -77,6 +77,18 @@ class TypewriterDaxDialog(
...
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        arguments?.let {
+            daxText = it.getString(ARG_DAX_TEXT, "")
+            primaryButtonText = it.getString(ARG_PRIMARY_CTA_TEXT, "")
+            secondaryButtonText = it.getString(ARG_SECONDARY_CTA_TEXT, "")
+            toolbarDimmed = it.getBoolean(ARG_TOOLBAR_DIMMED, true)
+            dismissible = it.getBoolean(ARG_DISMISSIBLE, true)
+            typingDelayInMs = it.getLong(ARG_TYPING_DELAY, 20)
```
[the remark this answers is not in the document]
marcosholgado: 👍

app/src/androidTest/java/com/duckduckgo/app/cta/ui/CtaTest.kt
```diff
@@ -209,7 +209,7 @@ class CtaTest {
 
     @Test
     fun whenCanSendPixelAndCtaIsPartOfHistoryThenReturnFalse() {
-        whenever(mockOnboardingStore.onboardingDialogJourney).thenReturn("e:0")
+        whenever(mockOnboardingStore.onboardingDialogJourney).thenReturn("i:0-e:0-s:0")
```
marcosholgado, line 212: Changed so it doesn't matter if the where in the history the cta was (first, last, middle...)

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: This is now done. In order to make it work I had to create two extra commands specific for DaxDialog actions that don't trigger the `cancelAllAnimations` function.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: and done!

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -52,7 +52,7 @@ import org.jetbrains.anko.longToast
 import timber.log.Timber
 import javax.inject.Inject
 
-class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
+class BrowserActivity : DuckDuckGoActivity(), DaxDialogListener, CoroutineScope by MainScope() {
```
marcosholgado, line 55: I've moved the listener here so it's easier to know if the dax dialog is shown or not. Keeping this in the fragment means having to check every single fragment to know if the dialog is being shown to remove it. By doing that there is also the risk of an IllegalStateException (since we do this onCreate) because some fragments might not be attached yet at that point.

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -389,5 +370,5 @@ fun DaxCta.addCtaToHistory(newCta: String): String {
 
 fun DaxCta.canSendShownPixel(): Boolean {
     val param = onboardingStore.onboardingDialogJourney?.split("-").orEmpty().toMutableList()
-    return !(param.isNotEmpty() && param.last().contains(ctaPixelParam))
+    return !(param.isNotEmpty() && param.any { it.contains(ctaPixelParam) })
```
marcosholgado, line 373: Changed this from last to any because it shouldn't matter where the position of cta in the history, if it exists it should not be sent again. I also updated the test for this.

marcosholgado:
@CDRussell you can ignore all the **new copy** tests for the re-review.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -281,6 +281,15 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         })
     }
 
+    private fun recoverDaxDialogFromActivity(): Fragment? = activity?.supportFragmentManager?.findFragmentByTag(DAX_DIALOG_DIALOG_TAG)
+
+    private fun removeDaxDialogFromActivity() {
+        val fragment = recoverDaxDialogFromActivity()
+        fragment?.let {
+            activity?.supportFragmentManager?.transaction { remove(it) }
```
cmonfortep, line 289: What's the difference between `remove` or `dialog.dismiss`?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -281,6 +281,15 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
         })
     }
 
+    private fun recoverDaxDialogFromActivity(): Fragment? = activity?.supportFragmentManager?.findFragmentByTag(DAX_DIALOG_DIALOG_TAG)
```
cmonfortep, line 284: `recoverDaxDialogFromActivity` sounds like we are recovering the dialog after something happened based on something we are storing in the activity. But we are just finding the dialog by tag inside the fragmentManager. What about using a simple name as `get`/`find`/`obtain`DaxDialog? I don't think `FromActivity` is needed.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
cmonfortep: I'm worried about that `currentTab` reference since we are not accessing by tab, and we are keeping the reference, I don't know if we should rely so strong on it for this to work.
The problem that you are describing is because multiple fragments trying to remove the same dialog?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: dismiss would trigger the listeners and the dialog won't be shown again if the activity gets killed which is not what we are after.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Renamed to `getDaxDialogFromActivity`

marcosholgado:
> I'm worried about that currentTab reference since we are not accessing by tab, and we are keeping the reference, I don't know if we should rely so strong on it for this to work.
The problem that you are describing is because multiple fragments trying to remove the same dialog?

@cmonfortep, the problem having the listeners in the fragment is that we would have to check all the existing fragments to see if the dialog was in any of them. Since we are doing it during `onCreate` it could happen that the fragment won't be attached at that point crashing the app, or if we handle the exception the fragment will eventually be attached, the dialog won't be removed and we will show the wrong dialog.

marcosholgado:
@CDRussell and @cmonfortep here are the two options with some pros and cons.

The main scenario here is:
1. Go to a SERP and see the SERP dialog
1. Take the app to the background.
1. Open externally any URL (i.e. a link via your email app using long tap and selecting open in browser).
1. Both dialogs are shown.

**Listener in activity Pros**
- The dialog only exists in the activity so removing it from it's easier and cannot result in a crash
**Listener in activity Cons**
- Usage of `currentTab`? I don't think that's a problem though, it is used consistently across more functionality to communicate with the fragment. 

**Listener in fragment Pros**
- Does not use `currentTab` and communicates directly with its view model.
**Listener in fragment Cons**
- The dialog could exist in any tab/fragment so we need to loop through all of them to check if it's there. While looping any of the fragments could not be attached yet which will result in a crash. Catching the crash means the dialog could still be shown (since the looping stops at that point) and will be shown. We do still check if we show a dialog at any fragment so that will be found and will stop a second one showing but we will be showing the wrong dialog.

cmonfortep:
@marcosholgado let's jump into a zoom call on Monday and brainstorm together.

PR #773, opened 2020-04-10
Interstitial search screen tweaks

subsymbolic:
**Task/Issue URL:**
- https://app.asana.com/0/488551667048375/1161178882056507
- https://app.asana.com/0/488551667048375/1170451534984707

**Description**:
* Coordinate timing so interstitial results drop in smoothly together rather than shifting items on screen
* Update the list order to prefer items that start with search query, then alphabetically
* Remove circular updates between the omnibar and ViewModel to improve text input experience.

**Steps to test this PR**:

**Test Interstitial Results**
1. Install a fresh version of the app and add the widget to your home screen (it is recommended that other versions are uninstalled first as it is really easy to add the wrong widget)
1. Tap the widget to launch the interstitial screen
1. Enter a query for example "Pl" notice that the autocomplete and app results appear smoothly together
1. Notice that 

**Validate interstitial Screen Text input**
1. Enter text in the search bar,
1. Enter a space
1. Delete the space
1. Delete the last word
1. Make sure this respond as expected

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
@@ -40,7 +40,7 @@ import kotlinx.coroutines.launch
 import timber.log.Timber
 import java.util.concurrent.TimeUnit
 
-typealias SystemSearchResult = Pair<AutoCompleteResult, List<DeviceApp>>
+data class SystemSearchResult(val autocomplete: AutoCompleteResult, val deviceApps: List<DeviceApp>)
 
 class SystemSearchViewModel(
     private var userStageStore: UserStageStore,
@@ -163,8 +163,8 @@ class SystemSearchViewModel(
     private fun updateResults(results: SystemSearchResult) {
         this.results = results
 
-        val suggestions = results.first.suggestions
-        val appResults = results.second
+        val suggestions = results.autocomplete.suggestions
+        val appResults = results.deviceApps
         val hasMultiResults = suggestions.isNotEmpty() && appResults.isNotEmpty()
 
         val updatedSuggestions = if (hasMultiResults) suggestions.take(RESULTS_MAX_RESULTS_PER_GROUP) else suggestions
@@ -172,7 +172,7 @@ class SystemSearchViewModel(
 
         resultsViewState.postValue(
             currentResultsState().copy(
-                autocompleteResults = AutoCompleteResult(results.first.query, updatedSuggestions),
+                autocompleteResults = AutoCompleteResult(results.autocomplete.query, updatedSuggestions),
                 appResults = updatedApps
             )
         )
--- a/app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
@@ -197,9 +197,13 @@ class SystemSearchViewModel(
     }
 
     fun userSubmittedQuery(query: String) {
+        if (query.isBlank()) {
+            return
+        }
+
         viewModelScope.launch {
             userStageStore.stageCompleted(AppStage.NEW)
-            command.value = Command.LaunchBrowser(query)
+            command.value = Command.LaunchBrowser(query.trim())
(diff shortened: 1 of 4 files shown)


JeffAment:
@subsymbolic,

The UX feels really smooth and refreshing :+1:.  For documentation sake, I also previously noticed some keystrokes were getting ghosted when the thread felt busy; this also seems to be fixed by this PR.

As a side note, I think maybe the reason you could not reproduce the autosuggestion/deletion issues may have something to do with the keyboard app.  I was using SwiftKey, and when I just tested this PR on an emulator with the default G-Board, I noticed that these two keyboards apparently handle auto-suggestions (SwiftKey appends a whitespace here, while G-Board does not) and whole-word deletions very differently.

Thank you for the ping, and for putting blind trust in my bug report without being able to reproduce it yourself.  Stay well! :pray:

JeffAment:
```
    private fun launchDeviceApp(command: LaunchDeviceApplication) {
        try {
            startActivity(command.deviceApp.launchIntent)
            finish()
        } catch (error: ActivityNotFoundException) {
            viewModel.appNotFound(command.deviceApp)
        }
    }
```

Minor nitpick: As a user, I expect that if I launch an app via the interstitial search and then press the back button, that I will be taken back to the existing `SystemSearchActivity` if pressing back in the selected app would close said app, but will be taken to a new/reset instance of `SystemSearchActivity` when an interstitial search is explicitly launched via the widget.

Prior to this PR, typing in the interstitial search, then clicking the home button, then explicitly reopening it via the widget gave us an empty canvas.  Now the search term is preserved in the search bar, but the search results are empty until the text change listener triggers.

CDRussell:
Thanks for the comments @JeffAment .

>As a user, I expect that if I launch an app via the interstitial search and then press the back button, that I will be taken back to the existing SystemSearchActivity

I have the opposite expectation; the search screen should feel like a transient state that is only shown when the user tapped on the widget. I don't think you should be able to navigate back to it. Google Search works the same way (at least on my device) where if you back out of a launched app, you don't see the search screen.


>typing in the interstitial search, then clicking the home button, then explicitly reopening it via the widget gave us an empty canvas. Now the search term is preserved in the search bar, but the search results are empty until the text change listener triggers.

I agree with the expectations here; clicking the widget would ideally always show an empty canvas. @subsymbolic, what's your thoughts on this? I don't think it's a blocker but it would be a good tidy-up task IMO.

app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
```diff
@@ -30,14 +30,18 @@ import com.duckduckgo.app.onboarding.store.isNewUser
...
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.Disposable
+import io.reactivex.functions.BiFunction
 import io.reactivex.schedulers.Schedulers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import timber.log.Timber
 import java.util.concurrent.TimeUnit
 
+typealias SystemSearchResult = Pair<AutoCompleteResult, List<DeviceApp>>
```
CDRussell, line 43: Why this construct over a data class? I've no objections, just wondering if there's an advantage of doing it this way?

subsymbolic: No, but now that you mention it _I_ have objections 😆I've switched to a data class, neater than a Pair and typealias.

app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
```diff
@@ -89,6 +89,7 @@ class SystemSearchViewModel(
     private fun currentResultsState(): SystemSearchResultsViewState = resultsViewState.value!!
 
     fun resetViewState() {
+        command.value = Command.ClearInputText
```
JeffAment, line 92: :+1: :smile:

subsymbolic:
Thanks both for your reviews, very much appreciated! And thanks for catching the canvas not clearing, we definitely want it to clear so I've updated it.

@CDRussell I'll merge this later today, just waiting on some internal testing.

PR #775, opened 2020-04-15
Fix download confirmation fragment illegal state exception

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1171376595453734

**Description**:
Fixes java.lang.IllegalStateException. Only shows the dialog if the state hasn't been saved yet as per FragmentManager "operations that would change saved state should not be performed if (isStateSaved) returns true". We could show the dialog with state loss however the activity is most likely being destroyed and there is no need to now show the confirmation dialog. 

**Steps to test this PR**:
* visit https://github.com/duckduckgo/Android/releases
* download our apk
* make sure that the confirmation dialog shows (if the app doesn't have permission yet you will need to provide permission and repeat the process as we only show the dialog when other confirmation dialogs have not been displayed)

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1098,7 +1098,9 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope {
 
     private fun requestDownloadConfirmation(pendingDownload: PendingFileDownload, downloadListener: FileDownloadListener) {
         fragmentManager?.let {
-            DownloadConfirmationFragment.instance(pendingDownload, downloadListener).show(it, DOWNLOAD_CONFIRMATION_TAG)
+            if (!it.isStateSaved) {
```
CDRussell, line 1101: I was wondering if we'd be better using `commitAllowingStateLoss()` but I think it makes no difference in the end. 👍

subsymbolic: Me too! I oscillated between them and then thought that since the dialog won't show in that circumstance because the activity is being killed, this is slightly closer to desired behavior. But I'd be happy with either too.

cmonfortep: @subsymbolic was this related to any race condition? e.g: calling `requestDownloadConfirmation` when activity was destroyed or being destroyed?

PR #776, opened 2020-04-15
Optimize animations

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1170386777246554
Tech Design URL: 
CC: 

**Description**:
Note: This PR targets the `onboarding_optimizations` branch to avoid reviewing unnecessary changes that have already been reviewed there. It will be merged after the `onboarding_optimizations` and directly to develop.

This PR adds some optimizations to the trackers animation:
1. New svg assets for all the logos, old png assets have been deleted.
1. If we don't have a logo for a tracker we use the first letter of the tracker name.
1. We've reduced the number of logos shown from 4 to 3.
1. If there are more than 3 trackers, the third logo shown is a generic image for "more trackers".

This PR also changes the privacy dashboard designs to be inline with the new tracker logos.

**Steps to test this PR**:
**More than 3 trackers**
1. Open `cnn.com`
1. The more trackers logo should be shown as the last logo (3 dots).

**Less than 3 trackers**
1. Go to `speedtest.net` 
1. Only one logo should be shown (Google)

**3 trackers**
1. Go to `unsplash.com/@levyphoto`
1. Three logos should be shown without the "more trackers" logo.

**Unknown logo**
1. Go to `unsplash.com/@levyphoto`
1. The third logo shown should be a K letter for Kantar Operations.

**Privacy dashboard**
1. Go to `cnn.com`
1. Open the privacy dashboard.
1. Notice how we now have logos for all the trackers where we previously didn't have. All of them have the first letter of the tracker name.

**Privacy dashboard theming**
1. Selected the light theme.
1. Go to `cnn.com`
1. Open the privacy dashboard.
1. Notice how the logos have a dark grey color.
1. Select dark theme.
1. Go to `cnn.com`
1. Open the privacy dashboard.
1. Notice how the logos have the dark grey color as in the light theme.

**CTA changes**
1. Install the app
1. Go to `cnn.com`
1. Notice how the dialog now only specifies two tracker names instead of three.

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
```diff
@@ -201,6 +265,8 @@ class BrowserTrackersAnimatorHelper {
...
         private const val TRACKER_LOGOS_DELAY_ON_SCREEN = 2400L
         private const val DEFAULT_ANIMATION_DURATION = 150L
-        private const val MAX_LOGOS_SHOWN = 4
+        private const val MAX_LOGOS_SHOWN = 3
     }
-}
\ No newline at end of file
+}
+
+data class TrackerLogo(val resId: Int, val trackerLetter: String = "")
```
cmonfortep, line 272: I wonder if we should set empty as `trackerLetter` default value. Seems to me that we should enforce that value to the caller.

app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
```diff
@@ -74,33 +79,92 @@ class BrowserTrackersAnimatorHelper {
...
+        imageView.id = View.generateViewId()
+        return imageView
+    }
+
+    private fun createResourcesIdList(activity: Activity, events: List<TrackingEvent>): List<TrackerLogo>? {
         if (activity.packageName == null) return emptyList()
+        val resourcesList = events.asSequence()
+            .mapNotNull { it.entity }
+            .sortedByDescending { it.prevalence }
+            .map {
```
cmonfortep, line 136: NIT: knowing that some sites have more than 30 trackers and we will show just 3 (but we need to know if there are more). What about mapping only 4 events?

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -134,14 +134,15 @@ sealed class DaxDialogCta(
             )
 
         override fun getDaxText(context: Context): String {
-            val trackersFiltered = trackers.asSequence()
-                .filter { it.entity?.isMajor == true }
-                .map { it.entity?.displayName }
+            val trackers = trackers.asSequence()
+                .mapNotNull { it.entity }
+                .sortedByDescending { it.prevalence }
```
cmonfortep, line 139: Is this the same logic that should be applied to the logos? If we apply any change here, should it be applied to the other logic too? if so, maybe we should consider some common logic, or one to rely on a list already sorted by the a domain component. (e.g: if we already sorted the tracker events, do we need to sort it again when rendering?)

app/src/main/res/layout/item_tracker_network_header.xml
```diff
@@ -45,12 +45,27 @@
...
+        android:layout_height="24dp"
         android:importantForAccessibility="no"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintTop_toTopOf="parent"
         tools:src="@drawable/network_pill_facebook_inc" />
 
+    <TextView
+        android:id="@+id/unknownIcon"
+        android:layout_width="24dp"
```
cmonfortep, line 58: I'm most probably wrong, but should this be `22dp` too?

app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
```diff
@@ -74,33 +79,92 @@ class BrowserTrackersAnimatorHelper {
         return createLogosViewList(activity, container, logos)
     }
 
-    private fun createLogosViewList(activity: Activity, container: ConstraintLayout, resourcesId: List<Int>?): List<View> {
+    private fun createLogosViewList(
+        activity: Activity,
+        container: ConstraintLayout,
+        resourcesId: List<TrackerLogo>?
```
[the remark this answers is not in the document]
marcosholgado: `createResourcesIdList` can return an `emptyList` which could end up here in which case it will return another `emptyList` to `startTrackersAnimation` to skip doing any animation.

app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
marcosholgado: As we talked in the call, his is used for both trackers with logos and without logos hence the use of a default value.

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
marcosholgado: This logic now happens in the `Site` as an extension function that is then used to pass a list of entities rather than a list of tracking events.

app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
marcosholgado: Done!

app/src/main/res/layout/item_tracker_network_header.xml
marcosholgado: We talked about this, the design calls for 24dp here.

app/src/main/java/com/duckduckgo/app/browser/BrowserTrackersAnimatorHelper.kt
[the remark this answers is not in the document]
cmonfortep: Exactly, but my point was that an `emptyList` is not nullable. Changing the arg type to `resourcesId: List<TrackerLogo>` will work and be more concrete.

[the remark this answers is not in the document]
marcosholgado: Gotcha! I've removed the nullable since it's not needed

PR #791, opened 2020-04-22
InterruptedException and IOException now excluded exceptions to alert on

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1172356004360461
Tech Design URL: 
CC: 

**Description**:
Adds `InterruptedIOException` and `InterruptedException` to the list that we don't want to crash the app on. These exceptions were observed to be happening in the 5.22.3 release. These are essentially the ones signalling that there was an internet issue or that the sync job was instructed to stop mid way through; they aren't real exceptions that should cause the app to stop.

**Steps to test this PR**:
1. Similar to https://github.com/duckduckgo/Android/pull/787, it's hard to replicate. But I found I could replicate by making the `TrackerDataDownloader` do lots of extra work, and killing the wifi off while it's looping. e.g., 👇

```
for (i in 0..100) {
    Timber.i("Running TDS download #$i")
    val call = trackerListService.tds()
    val response = call.execute()
}
```

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/AlertingUncaughtExceptionHandler.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/AlertingUncaughtExceptionHandler.kt
@@ -19,10 +19,13 @@ package com.duckduckgo.app.global
 import com.duckduckgo.app.global.exception.UncaughtExceptionRepository
 import com.duckduckgo.app.global.exception.UncaughtExceptionSource
 import com.duckduckgo.app.statistics.store.OfflinePixelCountDataStore
+import io.reactivex.exceptions.UndeliverableException
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.launch
+import timber.log.Timber
+import java.io.IOException
 
 class AlertingUncaughtExceptionHandler(
     private val originalHandler: Thread.UncaughtExceptionHandler,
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/global/AlertingUncaughtExceptionHandler.kt
```diff
@@ -40,4 +48,17 @@ class AlertingUncaughtExceptionHandler(
...
+
+    /**
+     * Some exceptions happen due to lifecycle issues, such as our sync service being forced to stop.
+     * In such cases, we don't need to alert about the exception as they are exceptions essentially signalling that work was interrupted.
+     * Examples of this would be if the internet was lost during the sync,
+     * or when two or more sync operations are scheduled to run at the same time; one would run and the rest would be interrupted.
+     */
+    private fun shouldIgnore(exception: Throwable?): Boolean {
+        return when (exception) {
+            is UndeliverableException, is InterruptedException, is IOException -> true
```
cmonfortep, line 60: How sure are we about ignoring `IOException`? I can think about many parts of the app that could crash with that error due to bug in our code. For example, working with `File`.

What's our goal here? If it's to ignore `IOException` that we catch and rethrow, we could wrap them into another exception very quick and ignore that one instead. I count 7 places where we do that. But I'm not sure if that's what we want or if I'm missing something else.

Happy to help here and do that change if needed.

CDRussell: Good point about `IOException`. Ultimately, the goal is to ignore exceptions which happen due to lifecycle issues such as the bg sync service being instructed to close by the system.

Revisiting the exceptions list, perhaps I over-generalized and should stick with `InterruptedException` and `InterruptedIOException`

Screen Shot 2020-04-23 at 7 55 08 AM

CDRussell: Changed this to look for `InterruptedIOException` instead of `IOException`

Issue #753, opened 2020-03-30
Downloading a file link with target=_blank opens a new blank tab


cmonfortep:
**Background**
PR https://github.com/duckduckgo/Android/pull/666/ added support for handling target=_blank links, which allows for a link to be opened in a separate tab.

Some of these links can be files that the user wants to download. Downloading any of these files results in the app opening a new blank tab.

**Steps to reproduce**
Navigate to http://www.granditaliaperth.co.uk/menus.html and click on the download link for the menu.

The download will start and a new blank tab will be opened.

**Objective**
Ensure that this action does not open a new blank tab.

rajkumaar23:
I would like to work on this issue.

PR #763, opened 2020-04-05
Handling target=_blank if link points to a file

AndrewR8:
Issue URL:  #753

**Steps to test this PR**:
1. Navigate to http://www.granditaliaperth.co.uk/menus.html and click on the download link for the menu.
1. The download will start and a new blank tab closes immediately

PR #798, opened 2020-04-27
Handling target=_blank if link points to a file

AndrewR8:
Issue URL: #753

**Description**:

**Steps to test this PR**:
1. Navigate to http://www.granditaliaperth.co.uk/menus.html and click on the download link for the menu.
2. The download will start and a new blank tab closes immediately

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -353,13 +353,7 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope, DaxDialogLi
 
     override fun onPause() {
         logoHidingListener.onPause()
-        dismissDownloadFragment()
         super.onPause()
-    }
-
-    private fun dismissDownloadFragment() {
-        val fragment = fragmentManager?.findFragmentByTag(DOWNLOAD_CONFIRMATION_TAG) as? DownloadConfirmationFragment
-        fragment?.dismiss()
     }
 
     private fun createPopupMenu() {
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -841,6 +841,14 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope, DaxDialogLi
             }
 
             it.setDownloadListener { url, _, contentDisposition, mimeType, _ ->
+                //handling target=_blank if link points to a file
+                if (viewModel.url == null) {     //in an empty tab
+                    launch {
+                        viewModel.closeCurrentTab()
+                        destroyWebView()
+                    }
+                }
+
                 requestFileDownload(url, contentDisposition, mimeType, true)
             }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -841,6 +835,14 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope, DaxDialogLi
             }
 
             it.setDownloadListener { url, _, contentDisposition, mimeType, _ ->
+                //handling target=_blank if link points to a file
+                if (viewModel.url == null) {     //in an empty tab
```
malmstein, line 839: Could we move this logic to the `ViewModel` and add a test for it?

You'd need to:
- Add a method in the `ViewModel` that handles when `viewmodel.url` is null and calls `viewModel.closeCurrentTab()`
- Create a new `Command` that will be returned from the new method in the `ViewModel`, something like `Command.FileDownload`
- Process that new Command in the `BrowserTabFragment` that calls `destroyWebView()`. 
- And finally add a test for all this logic

AndrewR8: Thanks for the detailed review @malmstein . I will try to implement this

cmonfortep:
@AndrewR8 one question from my side, is this aligned with the proposed solution in https://github.com/duckduckgo/Android/pull/763? 

My advice is to put on hold any extra change till the assignee reviews this PR. Just in case the approach changes and current review comments don't apply anymore. 👍

AndrewR8:
> @AndrewR8 one question from my side, is this aligned with the proposed solution in #763?

@cmonfortep It seems to me yes.

subsymbolic:
Thanks for submitting this @AndrewR8, I'll review it tomorrow.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
subsymbolic: Great feedback @malmstein. As the approach here matches the code around it and potentially fixes a frustrating user bug, I'm happy to accept the PR once we have a solid approach even without this feedback in. The PR will add value, even if it doesn't quite match our ideal internal style.

subsymbolic:
@AndrewR8 thanks for submitting this, it's great to see progress on such a frustrating user bug. There is one issue with this approach however, if multiple tabs are open the dialog appears over the first tab rather than the tab that requested the dialog.

In order to fix this we'll need to coordinate the events happening in the db and make sure the correct tab is selected. Feel free to look into this but it is complex. If we don't hear from you by Friday, we can use the work you have done so far and complete the implementation in-house.

AndrewR8:
@subsymbolic 
Due to this #812 issue there is no way to make sure at this moment  that the correct tab is selected when blank tab closes. May I fix this issue first in this PR? or open a new one?

subsymbolic:
Many thanks for your time on this @AndrewR8 and a big apology that our codebase changed after you had submitted your PR. We have since received and accepted another PR that fixes this in line with our latest changes as it also makes some db updates that will help up with upcoming work. Thanks again for submitting this and for your other contributions in the past.

PR #800, opened 2020-04-28
Feature/david/remove sticky search deprecated jobs

malmstein:
Task/Issue URL: https://app.asana.com/0/1157893581871903/1171986548569356

**Description**:
As part of the remove of the Sticky Search Experiment we also want to make sure that if there was a job already scheduled it's properly removed from the WorkManager

This PR adds that functionality, making it easier in the future to remove old jobs.

**Steps to test this PR**:
1. Install current Production apk
2. Create a release version of this branch and update 
3. Open the app
4. Use a db viewer to check that the old TAG `CONTINUOUS_APP_USE_REQUEST_TAG` is no longer present in the WorkManager

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/notification/AndroidNotificationSchedulerTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/notification/AndroidNotificationSchedulerTest.kt
@@ -19,10 +19,12 @@
 package com.duckduckgo.app.notification
 
 import androidx.test.platform.app.InstrumentationRegistry
+import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.WorkInfo
 import androidx.work.WorkManager
 import com.duckduckgo.app.CoroutineTestRule
-import com.duckduckgo.app.notification.NotificationScheduler.*
+import com.duckduckgo.app.notification.NotificationScheduler.ClearDataNotificationWorker
+import com.duckduckgo.app.notification.NotificationScheduler.PrivacyNotificationWorker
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/com/duckduckgo/app/notification/AndroidNotificationSchedulerTest.kt
```diff
@@ -98,17 +97,29 @@ class AndroidNotificationSchedulerTest {
         assertNoUnusedAppNotificationScheduled()
     }
 
+    @Test
+    fun allDeprecatedWorkIsCancelledUponStart() = runBlocking<Unit> {
```
subsymbolic, line 101: Our tests names follow a `when<condition>Then<expectedResult>` naming convention. This test doesn't seem to validate what the name suggest e.g if I delete the new call to `cancelAllUnnecessaryWork` this test still passes.

CDRussell:
@malmstein, can you advise where in the DB to look for the tag name? I can't find it, even when inspecting the DB on `develop` so I can't verify it's gone on this branch. thanks

app/src/androidTest/java/com/duckduckgo/app/notification/AndroidNotificationSchedulerTest.kt
malmstein: good catch, I will modify the test to enqueue a few deprecated jobs!

malmstein: and change the name of course

PR #803, opened 2020-04-29
Improved Portuguese Translation

yakkomajuri:
Hey guys, 

Just some changes here and there to improve the Portuguese translation.

A brief overview:

- Removed extremely formal language
- Fixed typos
- Removed words that either do not exist or sound like a bad translator version

Halfway through I realized a lot of what I was correcting was potentially correct in the Portuguese spoken in Portugal (I speak Brazilian Portuguese). If the choice is made to keep Portuguese as one language only, then it's up to you which spelling you'd prefer.

Finally, I apologize but wanted to contribute to this translation as a DuckDuckGo user, however I do not develop for Android nor do I use Kotlin/Java, so don't have any of the tools setup to run the app and test. I did try to keep most strings the same length and style, however.

Essentially, if someone thinks the changes make sense and want to run it, awesome. Otherwise, feel free to ignore.

Cheers and thanks for DuckDuckGo!

### DIFF
--- a/app/src/main/res/values-pt/strings.xml
+++ b/app/src/main/res/values-pt/strings.xml
@@ -1,6 +1,6 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
 
-    <string name="duckDuckGoLogoDescription">Logótipo do DuckDuckGo</string>
+    <string name="duckDuckGoLogoDescription">Logotipo do DuckDuckGo</string>
     <string name="duckDuckGoPrivacySimplified">Privacidade, simplificada.</string>
     <string name="yes">Sim</string>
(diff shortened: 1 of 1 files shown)


cmonfortep:
Hi @yakkomajuri, as you correctly saw, we don't currently support Brazilian Portuguese.

You can check the full list of supported languages here: https://github.com/duckduckgo/Android/blob/develop/TRANSLATIONS.md

We would love to support as many languages as possible, but at the same time, we need to keep a manageable size of translations.

You did great even not being familiar with our dev tools. Following our translation guidelines, I'm very sorry to discard this contribution.

PR #804, opened 2020-04-29
Fix issue in activity_user_survey.xml

john2020star:
Hi,

I found there is a issue in activity_user_survey.xml, you did not set `android:indeterminateTintMode="src_in"`, which can cause inconsistent color of ProgressBar in API Level 21. This pull request help you fix it.

Thanks and wish the information is useful for you.


subsymbolic:
We don't have any surveys running right now but I tested this with some dummy data and can confirm that the progress bar wasn't displayed on API 21 and is fixed by this change as per the gif.

survey_loader_fixed_api_21

subsymbolic:
Thanks for catching this and submitting a PR @john2020star!

PR #805, opened 2020-04-29
Add mechanism to remove deprecated WorkManager jobs

malmstein:
Task/Issue URL: https://app.asana.com/0/414730916066338/1173445999500153/f

**Description**:
- We use WorkManager to enqueue work that will be done in the background at some point in the future. 
- Some experiments will use this feature to schedule future work. Android: Sticky Search Notification is an example.
- When the experiment is removed, we need to make sure that those scheduled jobs are also removed from the user's device.

Our current implementation has `AndroidNotificationScheduler` cleaning up old jobs AND scheduling new work. The idea in this PR is to separate the responsibilities making it more readable.

---
###### Internal references:


app/src/androidTest/java/com/duckduckgo/app/job/JobCleanerTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    private var workManager = WorkManager.getInstance(context)
+    private lateinit var testee: JobCleaner
+
+    @Before
+    fun before() {
+        testee = AndroidJobCleaner(workManager)
+    }
+
+    @Test
+    fun allDeprecatedWorkIsCancelledUponStart() {
```
CDRussell, line 39: can we switch this test name to keep the convention `whenXThenY`?

e.g., `whenStartedThenAllDeprecatedWorkIsCancelled()`

app/src/androidTest/java/com/duckduckgo/app/job/JobCleanerTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        testee.cleanDeprecatedJobs()
```
CDRussell, line 40: should we schedule some deprecated work first? otherwise we might not actually be testing that the cleanup works if there were no jobs scheduled in the first place

app/src/main/java/com/duckduckgo/app/job/JobCleaner.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import androidx.work.WorkManager
+import com.duckduckgo.app.job.JobCleaner.Companion.allDeprecatedNotificationWorkTags
+
+interface JobCleaner {
+    fun cleanDeprecatedJobs()
+
+    companion object {
+        // below there is a list of TAGs that were used at some point but that are no longer active
+        // we want to make sure that this TAGs are cancelled to avoid inconsistencies
```
CDRussell, line 27: nit: grammar
>we want to make sure that **these** TAGs are cancelled to avoid inconsistencies

app/src/main/java/com/duckduckgo/app/job/JobCleaner.kt
```diff
@@ -0,0 +1,41 @@
...
+        private const val CONTINUOUS_APP_USE_REQUEST_TAG = "com.duckduckgo.notification.schedule.continuous" // Sticky Search Experiment
```
CDRussell, line 28: Could we combine the comment into the name and ditch the comment?

E.g., `private const val STICKY_SEARCH_CONTINUOUS_APP_USE_REQUEST_TAG = "com.duckduckgo.notification.schedule.continuous"`

malmstein:
@CDRussell I've changed how the `WorkManager` is initialized for testing according to the [documentation](https://developer.android.com/topic/libraries/architecture/workmanager/how-to/integration-testing). This has shown that we were not properly testing it, for that I've created a `TestWorker` that has no dependencies and it's sole purpose is to be tested

app/src/androidTest/java/com/duckduckgo/app/job/JobCleanerTest.kt
```diff
@@ -0,0 +1,102 @@
...
+        assertWorkersAreNotEnqueued()
+    }
+
+    private fun enqueueDeprecatedWorkers(){
+        allDeprecatedNotificationWorkTags().forEach {
+            val requestBuilder = OneTimeWorkRequestBuilder<TestWorker>()
+            val request = requestBuilder
+                .addTag(it)
+                .setInitialDelay(10, TimeUnit.SECONDS)
```
malmstein, line 76: we set this delay because if we don't the `Worker` will be immediately completed.

app/src/main/java/com/duckduckgo/app/job/JobCleaner.kt
```diff
@@ -0,0 +1,39 @@
...
+package com.duckduckgo.app.job
+
+import androidx.work.WorkManager
+import com.duckduckgo.app.job.JobCleaner.Companion.allDeprecatedNotificationWorkTags
+
+interface JobCleaner {
+    fun cleanDeprecatedJobs()
+
+    companion object {
+        private const val STICKY_SEARCH_CONTINUOUS_APP_USE_REQUEST_TAG  = "com.duckduckgo.notification.schedule.continuous" // Sticky Search Experiment
```
CDRussell, line 26: Now that the constant is prefixed with `STICKY_SEARCH`, we can kill the `// Sticky Search Experiment` comment

CDRussell:
> Looks like there's now some duplicate code left over in `NotificationScheduler`
> 
> 1. it still also tries to cancel unnecessary work
> 2. it also defines deprecated work tag list

Heads up that this comment hasn't been addressed

app/src/androidTest/java/com/duckduckgo/app/job/JobCleanerTest.kt
```diff
@@ -0,0 +1,127 @@
...
+    @Test
+    fun whenStartedThenAllDeprecatedWorkIsCancelled() {
+        enqueueDeprecatedWorkers()
+        assertDeprecatedWorkersAreEnqueued()
+        testee.cleanDeprecatedJobs()
+        assertDeprecatedWorkersAreNotEnqueued()
+    }
+
+    @Test
+    fun whenStartedAndNoDeprecatedJobsAreScheduledThenNothingIsRemoved() {
```
malmstein, line 67: @CDRussell this is the new test, that ensures old non deprecated work is not removed

PR #808, opened 2020-05-01
Remove cookies preserving fireproof websites

cmonfortep:
Task/Issue URL: https://app.asana.com/0/0/1170688450957740/f
Tech Design URL: https://app.asana.com/0/481882893211075/1170688450957739/f
CC: 

## **Description**:

**What's included in this PR?**
Implement logic to remove cookies preserving the ones related to fireproof website.

When clearing cookies:

- Try to directly remove cookies from WebView database, preserving the ones with hosts related to a fireproof website
- If process fails, fallback to remove all the cookies to avoid any leak
- Send pixels in the following scenarios:
- database path not found
- database can't be opened
- delete query fails

**What's NOT included in this PR?**
- UI improvements from design feedback
- Pixels from user interactions with Fireproof options

#### Test 1: No cookies preserved
1. Fresh install
1. Visit some different websites
1. Press fire button
1. Access device file explorer
1. Ensure cookie database doesn't exist or database is empty

#### Test 2: DDG preserved
1. Fresh install
1. Visit duckduckgo website
1. Set a theme (it will create a cookies)
1. Press fire button
1. Visit duckduckgo again
1. Ensure theme has been preserved

#### Test 3: Fireproof website preserved
1. Fresh install
1. Visit any website (e.g: twitter.com)
1. Perform an action that will create cookies (e.g: login)
1. Fireproof website
1. Press fire button
1. Visit fireproof website
1. Ensure state has been preserved (e.g: you are logged in)

#### Test 4: Database not found: Database not found, all cookies removed
1. Replace following line from `WebViewDatabaseLocator`:
Line 28: `private val knownLocations = listOf("/fakePath")`
1. Fresh install
1. Visit any website (e.g: twitter.com)
1. Press fire button
1. Ensure pixel `m_cdb_nf` sent
1. Ensure captured and recorded
1. Access device file explorer
1. Ensure cookie database doesn't exist or database is empty

#### Test 5: Error opening database, all cookies removed
1. Replace the following line from `SQLCookieRemover`, line 70:
```
            SQLiteDatabase.openDatabase(
                "fakepath",
                null,
                SQLiteDatabase.OPEN_READWRITE,
                DatabaseErrorHandler { Timber.e("COOKIE: onCorruption") })
```
1. Fresh install
1. Visit any website (e.g: twitter.com)
1. Press fire button
1. Ensure pixel `m_cdb_oe` sent
1. Ensure captured and recorded
1. Access device file explorer
1. Ensure cookie database doesn't exist or database is empty

#### Test 6: Error deleting cookies, all cookies removed
1. Replace the following line from `SQLCookieRemover`, line 87:
```
val number = delete("faketable", whereClause, excludedSites.toTypedArray())
```
1. Fresh install
1. Visit any website (e.g: twitter.com)
1. Press fire button
1. Ensure pixel `m_cdb_de` sent
1. Ensure captured and recorded
1. Access device file explorer
1. Ensure cookie database doesn't exist or database is empty

Additional tests:
Instead of using firebutton, you can use auto cleaner data from settings.

---
###### Internal references:


CDRussell:
There is a strange transition when going from Settings to Fireproof Website settings. I'm only noticing it on my Android R (DP3) Pixel 2 so there's very good chance it's not related to any changes you've made.
device-2020-05-06-151027

app/src/androidTest/java/com/duckduckgo/app/fire/WebViewCookieManagerTest.kt
```diff
@@ -17,66 +17,104 @@
...
 
-@Suppress("RemoveExplicitTypeArguments")
-class WebViewCookieManagerTest {
+private data class Cookie(val url: String, val value: String)
 
-    private lateinit var testee: WebViewCookieManager
+class WebViewCookieManagerTest {
+    @get:Rule
+    @Suppress("unused")
+    val coroutineRule = CoroutineTestRule()
```
CDRussell, line 36: Can you accept the lint suggestion to add the experimentation annotation? otherwise, there's a lot of yellow warnings that make it hard to see real lint issues

app/src/androidTest/java/com/duckduckgo/app/fire/WebViewCookieManagerTest.kt
```diff
@@ -17,66 +17,104 @@
...
+        givenCookieManagerWithCookies(ddgCookie, externalHostCookie)
+
         withContext(Dispatchers.Main) {
-            suspendCoroutine<Unit> { continuation ->
-                cookieManager.removeAllCookies { continuation.resume(Unit) }
-            }
+            testee.removeExternalCookies()
         }
+
+        verify(cookieManager, times(1)).setCookie(eq(ddgCookie.url), eq(ddgCookie.value), any())
```
CDRussell, line 64: nit: can ditch `times(1)` since that's the default impl of `verify`

app/src/androidTest/java/com/duckduckgo/app/fire/WebViewDatabaseLocatorTest.kt
```diff
@@ -0,0 +1,66 @@
...
+class WebViewDatabaseLocatorTest {
+
+    @Test
+    fun whenGetDatabasePathOnDeviceThenPathNotEmpty() {
+        val context = InstrumentationRegistry.getInstrumentation().targetContext
+        val webViewDatabaseLocator = WebViewDatabaseLocator(context)
+
+        val databasePath = webViewDatabaseLocator.getDatabasePath()
+
+        //If this test fails means WebViewDatabase path has changed its location
```
CDRussell, line 36: grammar suggestion:

>If this test fails, it means WebViewDatabase path has changed its location

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
```diff
@@ -0,0 +1,117 @@
...
+) : CookieRemover {
+
+    override suspend fun removeCookies(): Boolean {
+        return withContext(dispatcherProvider.io()) {
+            val databasePath: String = webViewDatabaseLocator.getDatabasePath()
+            if (databasePath.isNotEmpty()) {
+                val excludedHosts = getCookieHostsToPreserve()
+                return@withContext removeCookies(databasePath, excludedHosts)
+            } else {
+                pixel.fire(Pixel.PixelName.COOKIE_DATABASE_NOT_FOUND)
```
CDRussell, line 62: There might be a race condition here between the pixel trying to fire and the app process trying to restart. I think you might see some cases where pixels are lost because of this.

In other places where we want a pixel near to the app restart, we store the pixel offline and send them when the app next syncs.

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    private suspend fun openReadableDatabase(databasePath: String): SQLiteDatabase? {
+        return try {
+            SQLiteDatabase.openDatabase(
+                databasePath,
+                null,
+                SQLiteDatabase.OPEN_READWRITE,
+                DatabaseErrorHandler { Timber.e("COOKIE: onCorruption") })
```
CDRussell, line 74: Should this `DatabaseErrorHandler` not do something? send a pixel? crash the app?

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
```diff
@@ -0,0 +1,117 @@
...
+        } catch (exception: Exception) {
+            pixel.fire(Pixel.PixelName.COOKIE_DATABASE_OPEN_ERROR)
```
CDRussell, line 76: If the app is about to crash, there's a good chance this pixel will not be sent

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
```diff
@@ -0,0 +1,117 @@
...
+        openReadableDatabase(databasePath)?.apply {
+            try {
+                val whereClause = buildSQLWhereClause(excludedSites)
+                val number = delete(COOKIES_TABLE_NAME, whereClause, excludedSites.toTypedArray())
+                deleteExecuted = true
+                Timber.v("$number cookies removed")
+            } catch (exception: Exception) {
+                Timber.e(exception)
+                pixel.fire(Pixel.PixelName.COOKIE_DATABASE_DELETE_ERROR)
+                uncaughtExceptionRepository.recordUncaughtException(exception, UncaughtExceptionSource.COOKIE_DATABASE)
```
CDRussell, line 93: As above

app/src/main/java/com/duckduckgo/app/fire/GetCookieHostsToPreserve.kt
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.fire
+
+import com.duckduckgo.app.fire.fireproofwebsite.data.FireproofWebsiteDao
+
+class GetCookieHostsToPreserve(private val fireproofWebsiteDao: FireproofWebsiteDao) {
```
CDRussell, line 21: this returns a list of hosts which might include `.com`. I don't fully understand the logic here, so just raising the query that this is expected and OK?

cmonfortep: yes, it's expected and I can add a comment there. The reason is that we can't differentiate between domains and top-level domains (e.g: .com, .co.uk, .io, ...), but no cookies can exist for top-level domain (RFC2109).
The cookie host needs to be a FQHN (fully-qualified host name: host + top-level domain)

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
cmonfortep: Oh, thanks for pointing out that issue. Interesting 😅

I have a question: I'm sending this pixel to detect something unexpected happening related to the database and to detect it somehow from our side (e.g: grafana + alert). I was also using `recordUncaughtException` to receive the stack trace and to have some more info about the crash, but...

Will it be enough if we only rely on using:
`recordUncaughtException(exception, UncaughtExceptionSource.COOKIE_DATABASE)`?
I think exception will be stored on the database and we will send the pixel when app next syncs, but I prefer to double-check with you.

CDRussell: In this context though, you aren't crashing the app if there's a problem. The app might be about to restart, but it's not because of an uncaught exception, so I don't think it makes sense to try use uncaught exception handler here to get the pixel data. Maybe that makes sense in some of the other cases though?

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
CDRussell: In this case, it's probably fine to use the uncaught exception handler here and not sending your own pixel. one nuance: you are using the same `UncaughtExceptionSource.COOKIE_DATABASE` for all database problems, so maybe you'd need to add different sources for each potential database problem (like you currently have the different cases covered by different pixels)

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
cmonfortep: For the record. Here I'm not crashing the app, but neither in other parts related to the WebView database (if an error occurs, I capture the exception and fallback to removing everything via cookie manager). The difference with this case vs the others, it's that here I'm not capturing any exception whereas in the others I'm capturing the exception triggered by an external component and sending it back to us to know what's the issue with more detail.

Let's talk about this offline so we can discuss the correct approach for this. I would like to find a simple workaround instead of building a component to store pixels and trigger them when the app restarts only for this case.

cmonfortep: We talked about this offline: We are going to remove any usage of `recordUncaughtException`(1) here. Instead we will use `fire.pixel` (maybe losing some of them due to app restarting) to send a sneak peek of the stack trace as a parameter and use `OfflinePixelSender` to keep a real counter of how many times an issue happens.

(1) Reason we are not using `recordUncaughtException` is that we handle those exceptions. Responsibility for that class is to register any exception that makes the app to crash. We will introduce a new way of registering handled exceptions or illegal states but not in the scope of this PR.

app/src/main/java/com/duckduckgo/app/fire/CookieRemover.kt
cmonfortep: I went for the strategy of keeping a default behavior but sending a pixel. I'm wrapping the default SQLite handler into a custom class that also sends a pixel. 

By default, if you don't pass any handler there, sqlite will use a default one:
`mErrorHandler = errorHandler != null ? errorHandler : new DefaultDatabaseErrorHandler();`

cmonfortep:
@CDRussell this is ready for review. thanks.

app/src/main/java/com/duckduckgo/app/statistics/api/OfflinePixelSender.kt
```diff
@@ -139,4 +131,18 @@ class OfflinePixelSender @Inject constructor(
             SHOW_FILE_CHOOSER -> APPLICATION_CRASH_WEBVIEW_SHOW_FILE_CHOOSER
         }.pixelName
     }
+
+    private fun sendPixelCount(counter: KMutableProperty0<Int>, pixelName: Pixel.PixelName): Completable {
```
CDRussell, line 135: Slick! 🎉

PR #814, opened 2020-05-06
Make icons resilient so they can be used without issue across devices

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1174457371356692

**Description**:
Currently when we use an icon we need to also add a tint to the layout or style for low end devices. This is often omitted which can cause color glitches.

Thus in this task we:
* update the toolbar svgs to use a color attribute (rather than selector with attributes)
* add a selector (without inner attributes) for each theme to the styles that require it
* remove tint from overflow menu icon as no longer needed
* clean up names and remove 24dp form toolbar icons for consistency (we only really need to explicitly add dp when there are multiple versions)
* merge duplicate toolbar icons
* delete unused toolbar icons

**Steps to test this PR**:
Repeat the following test across these combinations:
- a recent API and API 21
- light theme and dark theme
- bottom bar experiment versus control

1. Launch to the home screen
1. Check that the icons color looks correct (it should be obvious if they are broken as they go bright pink!)
1. Open the menu and check that the enabled/disabled colors look correct
1. Visit every screen in the app and again check that the colors are correct 

---
###### Internal references:


app/src/main/res/layout/item_autocomplete_bookmark_suggestion.xml
```diff
@@ -28,7 +28,7 @@
         android:layout_height="24dp"
         android:layout_marginStart="16dp"
         android:importantForAccessibility="no"
-        android:src="@drawable/ic_bookmark_24dp"
+        android:src="@drawable/ic_autocomplete_bookmark_24dp"
```
malmstein, line 31: Does it make sense to keep the 24dp suffix? This usually is created when using the Asset Creator from Android Studio and I think we could just get rid of it

app/src/main/res/layout/layout_browser_bottom_navigation_bar.xml
```diff
@@ -46,7 +46,7 @@
         android:layout_weight="1"
         android:background="?selectableItemBackgroundBorderless"
         android:contentDescription="@string/back"
-        android:src="@drawable/ic_fire_black_24dp" />
+        android:src="@drawable/ic_fire_24dp" />
```
malmstein, line 49: same here, let's get rid of the suffix

app/src/main/res/values/attrs.xml
```diff
@@ -20,6 +20,9 @@
     <attr name="normalTextColor" format="color" />
     <attr name="toolbarIconColor" format="color" />
     <attr name="toolbarTextColor" format="color" />
+    <attr name="browserIconColorSelector" format="color" />
```
malmstein, line 23: this is really nice! we remove the duplicity and embrace light / dark theming at the same time!

app/src/main/res/values/attrs.xml
```diff
@@ -60,8 +63,6 @@
     <attr name="topCallToActionDescriptionColor" format="color" />
     <attr name="topCallToActionBackgroundColor" format="color" />
     <attr name="privacyDashboardLogoLetterTextColor" format="color" />
-    <attr name="bottomBarIconColor" format="color" />
```
malmstein, line 63: 👍

malmstein:
All tests passed on light / dark theme and bottom bar variant and control group

Tested on:
- API 21 Emulator
- API 29 Emulator
- Nexus 3a on Android 10

app/src/main/res/layout/item_autocomplete_bookmark_suggestion.xml
subsymbolic: Good thinking. Done!

subsymbolic:
Thanks for the review @malmstein 🎉

Issue #840, opened 2020-05-25
Please add web page dark mode


rogerxed:
Please add web page dark mode like opera browser, dark mode will reduce eye stress.

![Screenshot_20200525-212242](https://user-images.githubusercontent.com/65120278/82828237-39e54600-9ece-11ea-86cc-fdf91cf97b84.png)

KelvnPere:
Hi @cmonfortep I would love to work on this enhancement request.

cmonfortep:
Hi @KelvnPere, thanks for offering yourself to work on this. First, let me ask our product team about introducing this enhancement, and if they agree is something we can introduce I will ping you here.

@rogerxed just to clarify, are we talking about forcing the website to use dark mode when dark mode is enabled in our app, right?

KelvnPere:
That would be great @cmonfortep. Are they some minor fixes that help is needed. If any kindly drop a link. Thanks

rogerxed:
> That would be great @cmonfortep. Are they some minor fixes that help is needed. If any kindly drop a link. Thanks

Yes, Of course.

rogerxed:
> Hi @KelvnPere, thanks for offering yourself to work on this. First, let me ask our product team about introducing this enhancement, and if they agree is something we can introduce I will ping you here.
> 
> @rogerxed just to clarify, we are talking about forcing the website to use dark mode when dark mode is enabled in our app, right?

Yes, Of course.

cmonfortep:
Hi there @KelvnPere and others. I've been sharing this proposal with other members of the team and we need to explore this a bit more.

Basically, there are some sites where users could choose their own theme from the website settings, and we don't want to override those settings forcing the webview to use dark theme in some scenarios. A real example is DDG, users that visit our website, they can configure their them from https://duckduckgo.com/settings#theme 

We are open for suggestions about how to solve that usecase, I haven't had time to explore if there's any way to avoid webview overriding website settings.

sanogueralorenzo:
😱 Completely missed this issue before trying something out I opened a PR with a GIF showcasing it in case anyone wants to play around https://github.com/duckduckgo/Android/pull/848

Regarding your comment @cmonfortep it is a good point (although I haven't' personally faced it due to not using Themes on my end). When I had some issues with DarkMode on Kiwi Browser I just tend to remove it for that page the same way I would remove an Adblocker or use the whitelist feature from duckduckgo.

[...]

PR #848, opened 2020-06-06
Initial working implementation of webview darkmode

sanogueralorenzo:
Task/Issue URL:  https://github.com/duckduckgo/Android/issues/840
Tech Design URL: N/A
CC: @cmonfortep 

**Description**:
Hello 👋 A few days ago I moved from Kiwi browser (chromium fork) to DuckDuckGo and the biggest missing feature was having a settings option to enable Dark Mode within the web views (not just the theme).

Since it is an open-source project I tried it out following the same code standards that I found in the codebase. Open to suggestions and alternatives.

Update: Discovered the issue after playing around on my end, see comment [here](https://github.com/duckduckgo/Android/issues/840#issuecomment-640115154).

**Steps to test this PR**:
1. Open App -> Settings -> Enable Dark Mode Switch
2. Leaving Settings should refresh dark mode on any existing Tabs
3. New tabs created should have dark mode enabled

GIF to display current feature implementation
DuckDuckGo-DarkMode-GitHub-PR

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -320,7 +320,7 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope, DaxDialogLi
 
             @OnLifecycleEvent(Lifecycle.Event.ON_START)
             fun onStart() {
-                if (isVisible && WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK)) {
+                if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK)) {
                     WebSettingsCompat.setForceDark(webView!!.settings,
                             if (settingsDataStore.darkMode) WebSettingsCompat.FORCE_DARK_ON
                             else WebSettingsCompat.FORCE_DARK_OFF)
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -56,6 +56,8 @@ import androidx.lifecycle.Observer
 import androidx.lifecycle.OnLifecycleEvent
 import androidx.lifecycle.ViewModelProvider
 import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.webkit.WebSettingsCompat
+import androidx.webkit.WebViewFeature
 import com.duckduckgo.app.autocomplete.api.AutoComplete.AutoCompleteSuggestion
 import com.duckduckgo.app.bookmarks.ui.EditBookmarkDialogFragment
 import com.duckduckgo.app.brokensite.BrokenSiteActivity
@@ -87,6 +89,7 @@ import com.duckduckgo.app.global.device.DeviceInfo
 import com.duckduckgo.app.global.model.orderedTrackingEntities
 import com.duckduckgo.app.global.view.*
 import com.duckduckgo.app.privacy.renderer.icon
+import com.duckduckgo.app.settings.db.SettingsDataStore
 import com.duckduckgo.app.statistics.VariantManager
 import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.app.survey.model.Survey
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
```diff
@@ -86,6 +87,7 @@ class SettingsViewModel @Inject constructor(
         viewState.value = currentViewState().copy(
             loading = false,
             lightThemeEnabled = isLightTheme,
+                darkModeEnabled = settingsDataStore.darkMode,
```
amaury901130, line 90: check the indentation

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
```diff
@@ -44,6 +44,7 @@ class SettingsViewModel @Inject constructor(
         val loading: Boolean = true,
         val version: String = "",
         val lightThemeEnabled: Boolean = false,
+        val darkModeEnabled: Boolean = false,
         val autoCompleteSuggestionsEnabled: Boolean = true,
         val showDefaultBrowserSetting: Boolean = false,
         val isAppDefaultBrowser: Boolean = false,
```
amaury901130, line 50: You don't need to specify the type.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -311,6 +317,15 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope, DaxDialogLi
...
                 }
             }
+
+            @OnLifecycleEvent(Lifecycle.Event.ON_START)
+            fun onStart() {
+                if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK)) {
+                    WebSettingsCompat.setForceDark(webView!!.settings,
+                            if (settingsDataStore.darkMode) WebSettingsCompat.FORCE_DARK_ON
+                            else WebSettingsCompat.FORCE_DARK_OFF)
+                }
```
amaury901130, line 327: Don't force the `webView!!`.
Also this 
`if (settingsDataStore.darkMode) WebSettingsCompat.FORCE_DARK_ON else WebSettingsCompat.FORCE_DARK_OFF`
should be delegate to the viewModel `viewModel.style`
This too `WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK)`
`viewModel.isDarkModeSupported`

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
sanogueralorenzo: Thanks, forgot to apply the styleguide before opening the PR

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
sanogueralorenzo: Disagree for multiple reasons.

1. Kotlin type inference puts stress on the IDE (a small amount though although noticeable on big files)
2. Consistency, the rest of the data class has the types specified
3. Most important one, improves readability and serves as documentation (specially on more complex functions)

cmonfortep:
Thanks @sanogueralorenzo 
I've been playing a bit with this PR, and it truly works as you defined. However, we still have the problems we described in https://github.com/duckduckgo/Android/issues/840

- `FORCE_DARK_ON` seems like it's forcing the use of Dark colors but not respecting the website's Dark theme Some websites have a Dark them implemented, my expectation will be to use that one and not just changing to dark colors. That gives a poor user experience in my opinion. Maybe I would expect that behavior only in the case where a website doesn't have a Dark theme.

This is an example of wrong behavior for example. If you visit DuckDuckGo theme with Force Dark:

|Force Dark Disabled|Force Dark Enabled|
| ------ | ----- | 
|image|image|

There we can see the problems described:
- Not respecting website's Dark theme
- Forcing dark breaks how some options should look like (see light theme option is changed)
- Dark theme is not selected

Additionally, one expected behavior will be that if the user has selected another theme, that theme will be honored.

I found Vivaldi and Chrome having the expected behavior we want. Don't know how or if it's related to have chromium embedded instead of using a WebView like us. As an example, this is what happens when I enable dark mode on my phone.



If you want to explore another approach feel free to do so. I have the feeling that using force dark is not going to take us where we want in terms of our expectations. No problem at all if you prefer to close this PR and continue the discussions inside the related issue.

sanogueralorenzo:
Thanks for the comments @cmonfortep 

I'll close this PR and do some further investigation. Hope to see dark mode implemented soon (even if the only solution breaks some website themes).

PR #863, opened 2020-06-18
Add FireDialog theme in order to fix navigation buttons hiden on some devices

ezanetta:
Resolves [#754 ](https://github.com/duckduckgo/Android/issues/754)

**Description**:
* Added style to `FireDialog`, and override some properties in order to fix this issue :)

1. `windowIsFloating` as `true`
2. `navigationBarColor` as `?toolbarBgColor` (light/dark theme)
3. `statusBarColor` as `@android:color/transparent`

**Tested on:**
- Pixel XL api 29 emulator
- Pixel 3XL api 29 emulator
- Pixel 3a XL api 29 device

Old  |  New
:-------------------------:|:-------------------------:
Screen Shot 2020-06-17 at 20 45 51  | Screen Shot 2020-06-17 at 20 55 48

**Live preview**

live_demo

---
###### Internal references:

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -310,4 +310,10 @@
         <item name="android:textColor">?attr/colorAccent</item>
     </style>
 
+    <style name="FireDialog" parent="@style/Theme.Design.BottomSheetDialog">
+        <item name="android:windowIsFloating">false</item>
+        <item name="android:navigationBarColor">?toolbarBgColor</item>
+        <item name="android:statusBarColor">@android:color/transparent</item>
+    </style>
+
 </resources>
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -312,7 +312,6 @@
 
     <style name="FireDialog" parent="@style/Theme.Design.BottomSheetDialog">
         <item name="android:windowIsFloating">false</item>
-        <item name="android:navigationBarColor">?toolbarBgColor</item>
         <item name="android:statusBarColor">@android:color/transparent</item>
     </style>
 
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -310,4 +310,10 @@
         <item name="android:textColor">?attr/colorAccent</item>
     </style>
 
+    <style name="FireDialog" parent="@style/Theme.Design.BottomSheetDialog">
+        <item name="android:windowIsFloating">false</item>
+        <item name="android:navigationBarColor">?toolbarBgColor</item>
```
cmonfortep, line 315: This produces a UI issue on low api devices. We had a similar issue in the past, where using white color makes all navigation buttons not visible. To avoid that bug I suggest  you to create a new style for `v27`.

Also, even if they are the same, I think is better to use `colorPrimary` instead.

ezanetta: Oh, I didn't realize about that issue, sure, sounds good to me, thanks!

cmonfortep: No problem at all, it also happened to us in the past. Happy you can work on this and make it happen 🚀

ezanetta: I removed `android:navigationBarColor` from the main style, and move it to `v27` file and now it's working great. I tried on `api 29 / 21 / 23`. I tried with `colorPrimary` but the result was `@color/almostBlackDark` instead of the actual primary color, I found another style on `v27` using `<item name="android:navigationBarColor">?toolbarBgColor</item>` so I think it's ok? Let me know if we can improve anything 🙌

ezanetta: Also I noticed we can avoid the property `<item name="android:navigationBarColor">?toolbarBgColor</item>` and the `navigationBar` works fine, the nav buttons are visible, but the nav background is always black (I think it takes the navBar system default color)

cmonfortep: About using `?toolbarBgColor`, if we are already using that color is ok 👍 

And about about low api devices always being black is also ok, it's our expected behavior to avoid problems the problems I described with navigation buttons.

cmonfortep:
Thanks for the fix @ezanetta. This is now merged.

ezanetta:
> Thanks for the fix @ezanetta. This is now merged.

Thanks Cristian, I'm glad I can help 😄

PR #875, opened 2020-06-30
Rq Pixel should only be fired in SERP Header removal experiment

malmstein:
Task/Issue URL: https://app.asana.com/0/488551667048375/1182605057307457

**Description**:
As part of Test improvements to the double search header we added a new pixel because. We should have only added this for the SERP Header Removal Experiment Variant, but it was enabled for all experiments.

Objective
Fix this, so the Pixel is only fired in this variant

**Steps to test this PR**:
*SERP Header Removal Variant*
1. Enable variant zi
2. Open the app
3. Type a query in the native toolbar
4. Check that pixel rq_1 has been fired
5. Type the same query again
6. Check that the pixel rq_0 has been fired  

*Control group*
1. Enable variant zg
2. Open the app
3. Type a query in the native toolbar
4. Check that pixel rq_1 has not been fired
5. Type the same query again
6. Check that the pixel rq_0 has not been fired  

---
###### Internal references:


cmonfortep:
❤ those two last commits

PR #883, opened 2020-07-04
Issue #753 - Downloads that open in a new tab leave user on an empty tab

JKane10:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/753

**Description**:
* **Actual**: When downloading via a link that has target="_blank", the user is left with a blank tab.
* **Expected**: Ensure that this action does not open a new blank tab.
* Referenced the work of @AndrewR8 and the comments in PR https://github.com/duckduckgo/Android/pull/798

**Approach**:
1. Modified TabEntity to maintain a sourceTabId to reference the source tab by its tab id.
2. When opening a new tab, store the sourceTabId in the TabEntity db entry.
3. In the event that a download is opened in a new, blank tab (url is blank), callbacks will use tabRepo to select the source tab and delete the blank tab.

**Concerns**
* This still opens a blank tab in which the download options are displayed, but upon the user completing their selections, the blank tab is cleaned up and the source tab is selected and displayed.
* TabEntities now maintain a source whether it is needed or not - Could simplify this a bit by just passing the sourceTabId in the Fragment bundle as well?
* LeakCanary is identifying a leak, but struggling to track it down. If this approach is valid, could use some assistance on this.

**Steps to test this PR**:
1. Navigate to : https://www.w3schools.com/code/tryit.asp?filename=GGETHXPDSEMN
2. Run and select "Download in new tab".
3. Make the appropriate selection
4. New, blank tab should close and return to source tab.

---


cmonfortep:
Thanks for the contribution, I will review it this week.

cmonfortep:
@JKane10 Sorry for the delay here. I told you I was going to review your PR a few weeks ago and I didn't have time for it + personal time off. I'm back to the party and I will review it ASAP.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -505,6 +505,17 @@ class BrowserTabViewModel(
         viewModelScope.launch { removeCurrentTabFromRepository() }
     }
 
+    override fun closeAndSelectSourceTab() {
+        viewModelScope.launch { removeAndSelectTabsFromRepository() }
```
cmonfortep, line 509: Probably a typo here: `SelectTab` rather than `SelectTabs`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -505,6 +505,17 @@ class BrowserTabViewModel(
...
+    }
+
+    private suspend fun removeAndSelectTabsFromRepository() {
+        val currentTab = tabRepository.liveSelectedTab.value
```
cmonfortep, line 513: Sharing some thoughts with you here:

I see that we obtain the `currentTab` directly from `tabRepository.liveSelectedTab`, we do that here and in `BrowserViewModel`. I wonder if it could be better to move that logic inside the Repository, and directly work with `TabsDao.selectedTab()` from there. That way we will reduce the code in the `ViewModels` and put that logic close to the source of data. What do you think? For example `deleteAndSelectSource` could be renamed to `deleteCurrentTabAndSelectSource()`.

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
```diff
@@ -129,11 +129,19 @@ class BrowserViewModel(
     }
 
     suspend fun onNewTabRequested(isDefaultTab: Boolean = false): String {
-        return tabRepository.add(isDefaultTab = isDefaultTab)
+        return tabRepository.addWithSource(
+                isDefaultTab = isDefaultTab,
+                sourceTabId = tabRepository.liveSelectedTab.value?.tabId
```
cmonfortep, line 134: Related to my comment about moving `tabRepository.liveSelectedTab.value?.tabId` inside repository.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1087,10 +1087,21 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope, DaxDialogLi
         }
     }
 
+    private fun closeAndReturnToSourceIfBlankTab() {
+        if (viewModel.url == null) {
+            launch {
+                dismissDownloadFragment()
+                destroyWebView()
```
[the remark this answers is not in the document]
JKane10: Makes complete sense. I think I threw this in trying to squash the memory leak. I'll take it and the `dismessDownloadFragment` out so that the fragment lifecycle hands this instead.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
JKane10: Makes sense and will definitely reduce some unnecessary code.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -137,6 +149,20 @@ class TabDataRepository @Inject constructor(
         siteData.remove(tab.tabId)
     }
 
+    override suspend fun deleteAndSelectSource(tabToDelete: TabEntity) {
+        databaseExecutor().scheduleDirect {
+            deleteOldPreviewImages(tabToDelete.tabId)
+
+            tabToDelete.sourceTabId?.let { sourceTabId ->
```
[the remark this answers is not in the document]
JKane10: I originally grouped these together as there are also independent delete and select methods in the repo and figured it'd be better to do nothing than an unexpected half measure, but this is a good point. I've broken the two into separate steps inside the method.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -51,6 +51,18 @@ class TabDataRepository @Inject constructor(
...
     }
 
+    override suspend fun addWithSource(url: String?, skipHome: Boolean, isDefaultTab: Boolean): String {
+        val tabId = generateTabId()
+        add(
+            tabId,
+            buildSiteData(url),
+            skipHome = skipHome,
+            isDefaultTab = isDefaultTab,
+            sourceTabId = liveSelectedTab.value?.tabId
```
cmonfortep, line 61: Getting the value from `liveSelectedTab` works because when we call `addWithSource` (or `deleteCurrentTabAndSelectSource`) a ViewModel is observing that LiveData externally. But if we end up in a scenario where we call this method without anyone observing the `LiveData`, this will fail. I think is better to use `tabsDao.selectedTab()`.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -137,6 +149,23 @@ class TabDataRepository @Inject constructor(
         siteData.remove(tab.tabId)
     }
 
+    override suspend fun deleteCurrentTabAndSelectSource() {
+        val tabToDelete = liveSelectedTab.value
```
cmonfortep, line 153: Same here with this `LiveData`.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
JKane10: Makes sense, let me know how my latest approach looks. A bit new to coroutines still, so not sure if the usage of withContext is ideal, but we'd need to await the value from the DB before proceeding and can't access the DB on the main thread.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -137,6 +158,23 @@ class TabDataRepository @Inject constructor(
...
+    override suspend fun deleteCurrentTabAndSelectSource() {
+        databaseExecutor().scheduleDirect {
+            val tabToDelete = tabsDao.selectedTab()
+            tabToDelete?.tabId?.let {
+                deleteOldPreviewImages(tabToDelete.tabId)
+                tabsDao.deleteTab(tabToDelete)
+                siteData.remove(tabToDelete.tabId)
+            }
+
+            tabToDelete?.sourceTabId?.let { sourceTabId ->
```
cmonfortep, line 170: Checking twice this method, I was asking myself what will happen if `tabToDelete?.sourceTabId` is null. It will not crash but we will not select a new tab. It may produce an inconsistent state for users.
Additionally, we are doing a few operations that modify the database (delete + insert), but they are not inside a transaction. If one fails, we could leave the database in a weird state too.

To solve both things I was thinking about modifying `tabsDao.deleteTabAndUpdateSelection`, where we can have a `@Transaction`, and also a fallback to the firstTab.

```
    @Transaction
    open fun deleteTabAndUpdateSelection(tab: TabEntity, newSelectedTab: TabEntity? = null) {
        deleteTab(tab)

        if (newSelectedTab != null) {
            insertTabSelection(TabSelectionEntity(tabId = newSelectedTab.tabId))
            return
        }

        if (selectedTab() != null) {
            return
        }

        firstTab()?.let {
            insertTabSelection(TabSelectionEntity(tabId = it.tabId))
        }
    }
```

and the repository method will look like:
```
    override suspend fun deleteCurrentTabAndSelectSource() {
        databaseExecutor().scheduleDirect {
            val tabToDelete = tabsDao.selectedTab() ?: return@scheduleDirect

            deleteOldPreviewImages(tabToDelete.tabId)
            val tabToSelect = tabToDelete.sourceTabId
                .takeUnless { it.isNullOrBlank() }
                ?.let {
                    tabsDao.tab(it)
                }
            tabsDao.deleteTabAndUpdateSelection(tabToDelete, tabToSelect)
            siteData.remove(tabToDelete.tabId)
        }
    }
```

I haven't checked if that actually works or if it compiles 😅, just wanted to show the idea. What do you think? please double check from your side too in case you like the approach.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -81,7 +102,7 @@ class TabDataRepository @Inject constructor(
             }
             Timber.i("About to add a new tab, isDefaultTab: $isDefaultTab. $tabId, position: $position")
 
-            tabsDao.addAndSelectTab(TabEntity(tabId, data.value?.url, data.value?.title, skipHome, true, position))
+            tabsDao.addAndSelectTab(TabEntity(tabId, data.value?.url, data.value?.title, skipHome, true, position, null, sourceTabId))
```
cmonfortep, line 105: Not important: I see you added `null` as `previewImage` because you wanted to pass `sourceTabId` which comes next. To avoid that, you could use named params instead, and only send the ones you know its value.

```
            tabsDao.addAndSelectTab(
                TabEntity(
                    tabId = tabId,
                    url = data.value?.url,
                    title = data.value?.title,
                    skipHome = skipHome,
                    viewed = true,
                    position = position,
                    sourceTabId = sourceTabId
                )
            )
```

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -51,6 +53,25 @@ class TabDataRepository @Inject constructor(
         return tabId
     }
 
+    override suspend fun addWithSource(url: String?, skipHome: Boolean, isDefaultTab: Boolean): String {
+        val tabId = generateTabId()
+        var sourceTabId: String? = null
+
+        withContext(Dispatchers.IO) {
```
cmonfortep, line 60: 👍 you can perfectly use `withContext` here to change threads.

Because you mentioned that you were a bit new to coroutines, just a small nice tip 🙌.

If you check the signature of `withContext`:
```
public suspend fun <T> withContext(
    context: CoroutineContext,
    block: suspend CoroutineScope.() -> T
): T
```
That means `withContext` can return the value you are querying in the internal `block`.

You can do this:
```
        val sourceTabId = withContext(Dispatchers.IO) {
           tabsDao.selectedTab()?.tabId
        }
```

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
JKane10: Good call, would definitely want to ensure the entire transaction completes rather than risk leaving the DB in an undesired state. The code you provided works as is and looks good to me.

This also accounts for the unlikely situation that a user opens a link in the new tab, declines the dialog, closes the sourceTab, and then refreshes the download tab to reinitialize the download dialog and proceeds. The user will now return to the firstTab where as previously the user would just remain on the blank new tab.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
JKane10: Awesome, this is clean too!

JKane10:
Made changes related to the latest feedback and added 2 tests for the 2 new TabDataRepo methods. Let me know if the tests look sufficient.

JKane10:
> 
> 
> @JKane10 This looks great! I've also tested the latest revision and works as expected. Good job!
> 
> Sadly, I've just noticed we merged today a PR that has updated the database version, so you have some conflicts to solve here. Could you solve them? I will merge the PR after that.
> 
> Solving database code conflicts is a bit tricky: you will need to accept the 23.json from develop, update your database versions to 24, and that will generate a new json file for the database changes we introduced here. Also, update the database test cases to cover migration from 23 to 24. If it gets hairy, let me know and I can solve them for you. Happy to help here.

The DB implementation is really well done. Updating the migration and test seems straight forward. Let me know if I missed anything.

JKane10:
> 
> 
> @JKane10 Thanks for the contribution!!! Really enjoyed working with you here. We have been really looking forward to fixing this bug, so thanks for the changes. 🙌

Thanks for working with me on this @cmonfortep. Enjoyed contributing. I'll keep my eye out for other opportunities.

PR #885, opened 2020-07-07
Render ER diagrams with FloorPlan

julioz:
[FloorPlan](https://github.com/julioz/FloorPlan) translates Room schemas into entity-relationship diagrams, to make it easier to inspect changes in the persistance-layer schema of the app.

Here, I am configuring FloorPlan to output `SVG` files by mimicking the Room directory structure, into a sibling directory under the `:app` module.
The first commit is the plugin configuration, the second is checking in the actual diagrams.

You can manually generate the diagrams by running

```
./gradlew generateFloorPlan
```

Optionally, we could also make `compileDebugJava`/`kaptDebugKotlin` depend on `generateFloorPlan`, and we would get automatic diagram generation on builds for that module, similarly to how [Room hooks into the build process to generate the JSON schemas](https://stackoverflow.com/a/57103344/12511149).

On top of that, there's also an (incubating) IntelliJ plugin to help with discoverability of the schemas:
floorplan-intellij-plugin

Let me know your thoughts on this.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,6 +5,7 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
 apply from: '../versioning.gradle'
 apply from: "$rootDir/ktlint.gradle"
+apply plugin: 'com.juliozynger.floorplan'
 
 ext {
     USE_ORCHESTRATOR = project.hasProperty('orchestrator') ? project.property('orchestrator') : false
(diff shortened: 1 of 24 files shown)


cmonfortep:
Hi @julioz, I've been taking a look to your library, and also reading the post on your website. Nice library.
However, our database is pretty simple (we don't even have relations between entities), so currently we don't have any need of visualizing our database schema. Thanks for sharing anyways.

PR #896, opened 2020-08-05
Improve algorithm to generate a filename from an image download

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1187737416677993/f
Tech Design URL: 
CC: 

**Description**:
When trying to download images, our app currently tries to guess what a good filename for the download will be, and often fails to produce something nice; it'll often end in a .*bin file type as its current implementation is somewhat naive in turning a URL into a filename.

This PR improves that situation by walking back the URL path, trying each to make a filename, and then returning the best filename produced. Worst case, this should produce the same result as the old algorithm, but best case it will pick out a much better filename.

**Steps to test this PR**:
1. Visit https://vignette3.wikia.nocookie.net/guns/images/b/b1/FHJ84.jpg/revision/latest?cb=20160824141549 and long press to download the image; verify the filename produced is `FHJ84.jpg` and not `latest.bin` as it was before
1. Try other image downloads, and verify they produce a sensible enough filename

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2020 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.downloader
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class UriUtilsFilenameExtractorTest {
+
+    private val testee: FilenameExtractor = FilenameExtractor()
+
+    @Test
+    fun whenUrlEndsWithFilenameAsJpgNoMimeOrContentDispositionThenFilenameShouldBeExtracted() {
+        val url = "https://example.com/realFilename.jpg"
+        val mimeType: String? = null
+        val contentDisposition: String? = null
+        val extracted = testee.extract(url, contentDisposition, mimeType)
+        assertEquals("realFilename.jpg", extracted)
+    }
+
+    @Test
+    fun whenUrlContainsFilenameButContainsAdditionalPathSegmentsThenFilenameShouldBeExtracted() {
+        val url = "https://foo.example.com/path/images/b/b1/realFilename.jpg/other/stuff"
+        val mimeType: String? = null
+        val contentDisposition: String? = null
+        val extracted = testee.extract(url, contentDisposition, mimeType)
+        assertEquals("realFilename.jpg", extracted)
+    }
+
+    @Test
+    fun whenUrlContainsFilenameButContainsAdditionalPathSegmentsAndQueryParamsThenFilenameShouldBeExtracted() {
+        val url = "https://foo.example.com/path/images/b/b1/realFilename.jpg/other/stuff?cb=123"
+        val mimeType: String? = null
+        val contentDisposition: String? = null
+        val extracted = testee.extract(url, contentDisposition, mimeType)
+        assertEquals("realFilename.jpg", extracted)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/browser/downloader/FilenameExtractor.kt
```diff
@@ -0,0 +1,87 @@
...
+        if (latestGuess.contains(".")) {
+            guesses.bestGuess = latestGuess
+        }
+
+        if (pathSegments.size < 2) return TriedAllOptions
+
+        return NotGoodEnough
+    }
+
+    private fun guessFromUrl(url: String, contentDisposition: String?, mimeType: String?): String {
```
cmonfortep, line 55: That made me think this method was going to use only `url: String` to guess the fileName, but I think we will also use also the other params. Wanted to check if my assumption is true

app/src/main/java/com/duckduckgo/app/browser/downloader/FileDownloader.kt
```diff
@@ -27,7 +27,8 @@ import javax.inject.Inject
 
 class FileDownloader @Inject constructor(
     private val dataUriDownloader: DataUriDownloader,
-    private val networkFileDownloader: NetworkFileDownloader
+    private val networkFileDownloader: NetworkFileDownloader,
+    private val filenameExtractor: FilenameExtractor
```
cmonfortep, line 31: unused param `filenameExtractor`

app/src/main/java/com/duckduckgo/app/browser/downloader/FileDownloader.kt
```diff
@@ -57,8 +58,8 @@ class FileDownloader @Inject constructor(
     }
 }
 
-fun FileDownloader.PendingFileDownload.guessFileName(): String {
-    val guessedFileName = URLUtil.guessFileName(url, contentDisposition, mimeType)
+fun FileDownloader.PendingFileDownload.guessFileName(filenameExtractor: FilenameExtractor): String {
```
cmonfortep, line 61: No strong opinion, and not really important, just talking out loud.

It feels a bit weird that `PendingFileDownload` has a method `guessFileName` that receives a `FileNameExtractor` to guess the file name. I would expect the other way around: `FileNameExtractor.guessFileNameFrom(pendingFileDownload)` as a extension function that receives a `PendingFileDownload`. Inside that method we map `PendingFileDownload` to `(url, contentDisposition, mimeType)`

app/src/main/java/com/duckduckgo/app/browser/downloader/FilenameExtractor.kt
CDRussell: Good point; changed to `guessFilename()` which is hopefully clearer

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        assertEquals("realFilename.bin", extracted)
+    }
+
+    @Test
+    fun whenUrlContainsNoFileNameButLotsOfPathsSegmentsThenFirstSegmentNameIsUsed() {
+        val url = "https://example.com/foo/bar/files"
+        val mimeType: String? = null
+        val contentDisposition: String? = null
+        val extracted = testee.extract(url, contentDisposition, mimeType)
+        assertEquals("foo", extracted)
```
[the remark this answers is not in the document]
CDRussell: Fixed to return `.bin` if all no filetype available on the best guess

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
```diff
@@ -0,0 +1,142 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.downloader
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class UriUtilsFilenameExtractorTest {
```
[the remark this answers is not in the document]
CDRussell: 1. `contentDisposition` is preferred. that is unchanged from current behavior; we still fall back on `URLUtil.guessFileName` which decides on the precedence between them. I added a unit test in here.
1. I see `mimetype` being preferred, but again, deferring that logic to `URLUtil.guessFileName`. so for this scenario, the behavior is unchanged from the current behavior.

app/src/main/java/com/duckduckgo/app/browser/downloader/FileDownloader.kt
CDRussell: Agree, this is weird. I've removed that extension function on `pendingFileDownload` and put the logic all in `FileNameExtractor`

PR #898, opened 2020-08-07
Fix crash that could happen when trying to download a large data URI image

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1187803392412385/f
Tech Design URL: 
CC: 

**Description**:
We saw crash instances when a user attempts to save a data URI. This PR fixes that, and has a few related improvements to that area

- Simplifies the `DownloadConfirmationFragment`
    - Previously it was responsible for asking what the user wanted to do, as well as handling the response and trying to delete files, trigger the download etc...
    - Now it is just responsible for passing the user's choice back to the parent fragment
    - Removed a leak with this dialog

- Limits the title length for the long press menu. An exceedingly long data URI could trigger an ANR as we were trying to populate a `TextView` with potentially megabytes of data (and the UI already capped what it could output)

**Steps to reproduce crash**
1. Use an emulator or a device with a large URI data file.
1. In the developer settings, enable "Don't keep activities".
1. Go https://dopiaza.org/tools/datauri/examples/index.php
1. Long click on the image and save image. As soon as you save the image, move the app to the background.
1. Crash should show in the logcat. (java.lang.NullPointerException)

**Steps to test this PR**:
1. Repeat the above steps on this branch; verify no crash happens and the data uri image is saved ok
1. Try other downloads; verify they work as expected (e.g., https://file-examples.com/)
1. Test blank tabs for new downloads are still closed; follow steps in https://github.com/duckduckgo/Android/pull/883

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/WebViewLongPressHandler.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/WebViewLongPressHandler.kt
@@ -46,7 +46,7 @@ interface LongPressHandler {
 class WebViewLongPressHandler @Inject constructor(private val context: Context, private val pixel: Pixel) : LongPressHandler {
 
     override fun handleLongPress(longPressTargetType: Int, longPressTargetUrl: String?, menu: ContextMenu) {
-        menu.setHeaderTitle(longPressTargetUrl ?: context.getString(R.string.options))
+        menu.setHeaderTitle(longPressTargetUrl?.take(MAX_TITLE_LENGTH) ?: context.getString(R.string.options))
 
         var menuShown = true
         when (longPressTargetType) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/duckduckgo/app/browser/WebViewLongPressHandler.kt
```diff
@@ -46,7 +46,7 @@ interface LongPressHandler {
 class WebViewLongPressHandler @Inject constructor(private val context: Context, private val pixel: Pixel) : LongPressHandler {
 
     override fun handleLongPress(longPressTargetType: Int, longPressTargetUrl: String?, menu: ContextMenu) {
-        menu.setHeaderTitle(longPressTargetUrl ?: context.getString(R.string.options))
+        menu.setHeaderTitle(longPressTargetUrl?.take(MAX_TITLE_LENGTH) ?: context.getString(R.string.options))
```
cmonfortep, line 49: 👍

PR #901, opened 2020-08-08
Fixes swipe to refresh triggers early when address bar is hidden

### LINKED ISSUE #900: Swipe to refresh triggers when address bar is hidden
Device: Google Pixel XL
OS: Android 10
DDG Version: 56000

On http://raosrestaurants.com/, the swipe to refresh triggers when it isn't supposed to (see gif below; very easy to reproduce on my device). Possibly related to #638. I've had similar issues on a handful of other sites also. I don't have a way to reproduce the bug on the other sites, but what I can say is that in every instance, the address bar was hidden.

Not sure if it matters, but I would also like to note that the other sites I mentioned, unlike raosrestaurants.com, did not render themselves to fill the `WebView` and had room to scroll up and down [in the `WebView`] when the unwanted refresh happened; the bug only happened one time on each site and could not be reproduced. Perhaps a check to see if the address bar is fully docked would fix, or at least mitigate, this issue.

I really don't think this is relevant, but if you wanted to know which other sites I had issues on, I will list them below. Please note that it is **_very hard to reproduce on these sites, and likely has nothing to do with the sites themselves_**:
https://portal.311.nyc.gov/
https://www.chess.com/
https://www.5dimes.eu
http://cooper.edu/

-------
**Edit:**

I found two more examples of sites where this bug is easy to reproduce:

**reddit.com**
- Go to reddit.com
- Wait for the site to lose focus, and for the overlay to appear with the bottom dialog asking whether to continue in the Reddit app or with the browser.

**linkedin.com**
- Visit LinkedIn profiles until you hit the authwall. (example: https://www.linkedin.com/in/chadchampion a.k.a. "the final boss of LinkedIn")

Both of these examples seem to distill themselves to the base case we saw on raosrestaurants.com (i.e. they fill the `WebView`)

**_End of edit_**

-------
Raos

Usiel:
- Do not abort nestedScroll
- Enable swipe to refresh only when no vertical nestedOffset is indicated

Fixes #900

**Steps to test this PR**:
See #900 linked websites

While this fixes the issue it's not pretty unfortunately. I observed the following behavior with Chrome, but not quite sure how to implement it:
1. Swipe refresh enabled on load
2. User swipes down -> Swipe refresh disabled
3. User swipes up and triggers the overscroll animation (swipe refresh still disabled)
a) If user swipes up again while the overscroll animation is active -> Swipe refresh stays disabled
or
b) Overscroll animation finishes -> Swipe refresh enabled

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/DuckDuckGoWebView.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/DuckDuckGoWebView.kt
@@ -92,15 +92,15 @@ class DuckDuckGoWebView : WebView, NestedScrollingChild {
 
                 returnValue = super.onTouchEvent(event)
 
-                if (scrollY == 0 && lastClampedTopY) {
-                    // we have reached the top and are clamped -> enable swipeRefresh (by default always disabled)
-                    enableSwipeRefresh(true)
-
-                    stopNestedScroll()
-                } else if (dispatchNestedScroll(0, scrollOffset[1], 0, deltaY, scrollOffset)) {
+                if (dispatchNestedScroll(0, scrollOffset[1], 0, deltaY, scrollOffset)) {
                     event.offsetLocation(0f, scrollOffset[1].toFloat())
                     nestedOffsetY += scrollOffset[1]
                     lastY -= scrollOffset[1]
+                }
+
+                if (scrollY == 0 && lastClampedTopY && nestedOffsetY == 0) {
+                    // we have reached the top, are clamped vertically and nestedScrollY is done too -> enable swipeRefresh (by default always disabled)
+                    enableSwipeRefresh(true)
                 }
 
                 lastDeltaY = deltaY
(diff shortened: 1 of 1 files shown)


JeffAment:
Thank you for fixing this :pray: I see what you mean about the overscroll animation, but overall everything works much better with these changes.

I've added two other websites where this is easy to reproduce as an edit to my original post on #900; both of these sites also behave much better after these changes, and the issue only seems to happen when the overscroll animation is still active, though I do wonder if checking whether the address bar is fully docked via an `offsetChangedListener` etc. before enabling swipe to refresh would help this further, as the hidden address bar still seems to be the common denominator. Great work :+1:

cmonfortep:
@Usiel going to merge this PR as I don't want to delay the fix for our users.
The idea of @JeffAment about checking if the toolbar is fully shown seems interesting and maybe we want to explore it too. So if anyone has the time for that, feel free to go for it, I'm happy to review it or give support there.

Usiel:
@cmonfortep Sure makes sense. I'll see if I can tinker with Jeff's idea in the next few days and achieve a cleaner behavior hopefully (both code and UI).

PR #911, opened 2020-08-17
Optimize and refactor swipe-to-refresh in WebView

Usiel:
**Description**:

Changes in this PR try to emulate Chrome's swipe-to-refresh behavior better:
Swipe-to-refresh ...
- ... is disabled when user swipes down
- ... is enabled again only once the user is clamped at the top, the address bar is not consuming any more scroll and the over scroll effect has receded (another swipe up before that only triggers the over scroll effect again)

We
- introduce a new handler for delayed enable of swipe-to-refresh after over scroll (emulating Chrome behavior)
- and adapt existing logic in `onTouchEvent` to follow the same rule

I feel the user experience is improved with these changes. I'm not so sure about the implementation as I don't have enough Android experience to tell whether the approach using the handler is sound or not.

**Steps to test this PR**:
1. Open any page
2. Try to use swipe to refresh (we expect an overscroll glow effect)
3. Wait 600ms
-> Swipe to refresh should be active (no glow expected)
or
3. Immediately swipe up again (repeat as often as you wish)
-> Overscroll glow effect should appear, no swipe to refresh

Websites good for testing (thanks @JeffAment):
http://raosrestaurants.com/ (no scroll)
http://cooper.edu/ (horizontal scrolling element which messed up the logic previously)
https://www.openstreetmap.org/ (map)
and Twitter (implements custom swipe-to-refresh)

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/DuckDuckGoWebView.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/DuckDuckGoWebView.kt
@@ -18,8 +18,10 @@ package com.duckduckgo.app.browser
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.os.Handler
 import android.util.AttributeSet
 import android.view.MotionEvent
+import android.view.View
 import android.view.inputmethod.EditorInfo
 import android.view.inputmethod.InputConnection
 import android.webkit.WebView
@@ -35,6 +37,11 @@ import androidx.core.view.ViewCompat
  * Originally based on https://github.com/takahirom/webview-in-coordinatorlayout for scrolling behaviour
  */
 class DuckDuckGoWebView : WebView, NestedScrollingChild {
+    private var swipeRefreshHandler: Handler = Handler()
+    private var enableSwipeRefreshRunnable: Runnable = Runnable {
+        enableSwipeRefresh(true)
+    }
+    private var lastClampedTopYTimestamp: Long? = null
     private var lastClampedTopY: Boolean = false
     private var contentAllowsSwipeToRefresh: Boolean = true
     private var enableSwipeRefreshCallback: ((Boolean) -> Unit)? = null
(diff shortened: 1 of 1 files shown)


cmonfortep:
Thanks @Usiel, I will take a look this week.

cmonfortep:
While testing this I see some weird behavior (attached gif).

- Once the pull to refresh appears (but not fully triggered, because an incomplete user action or whatever), we are in a state where any movement (even horizontal) triggers the pull to refresh.
- Now it's harder to trigger the pull to refresh in any site. You can see in cnn.com how I'm trying to trigger it multiple times, and only doing a very slow movement triggers it.

weird pull to refresh

I was thinking about the chrome delay you were pointing out. I think it's nice we try to mimic it. But I would like to be sure that any ux improvement vs code complexity worths it. I would like to keep that part of the code simple, and if we add complexity, the improvement needs to be notorious for the user, and I'm not sure about that delay.

RicoHeat:
Anyway to disable it altogether? User option?

cmonfortep:
@RicoHeat I will shate that idea internally, but it's not something we can decide without including UI/product team. Thanks.

Usiel:
> I was thinking about the chrome delay you were pointing out. I think it's nice we try to mimic it. But I would like to be sure that any ux improvement vs code complexity worths it. I would like to keep that part of the code simple, and if we add complexity, the improvement needs to be notorious for the user, and I'm not sure about that delay.

Yep I agree. I thought during testing the behavior was much nicer than before, that's why I assumed it might be worth the extra complexity. I'll see if the issues you mentioned are fixable without expanding the code further. If not I can just close the PR. Hope to get to work on this next week, currently a bit busy unfortunately.

cmonfortep:
Thanks, @Usiel, sounds reasonable. I personally noticed sometimes hard to make the toolbar appear again. 

Eg:
- toolbar is hidden
- I try to scroll down to make it appear
- pull to refresh triggered instead of the toolbar showing

Feel free to do whatever is best for you here: closing and opening a new one (preferred), or continue any work here. You are the owner of this contribution, so I'm fine with both. Thanks for all trying to make it better in any case 👍

PR #916, opened 2020-08-24
Fire button education

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1190516087603930/f
Tech Design URL: 
CC: https://app.asana.com/0/72649045549333/1186511372568850/f

**Description**:
We would like to encourage users to try the fire button during onboarding.

This PR adds an experiment to test a new onboarding flow where we will ask the user to try Fire button.

For new users, after they see the first in-browser CTA (except SERP CTA), we will start highlighting the Fire button with a pulse animation. That animation will last till the user interacts with the button.
The first time the user interacts with the Fire button, we will show a CTA to encourage them to try clearing all data.
Fire button education variant also includes a Fire animation.

**Steps to test this PR**:

---- Fire education variant ----

>Test 1 -Full flow-: 
1. Hardcode `zr` variant
1. Perform fresh install
1. Navigate to Browser screen
1. Visit SERP Page
1. Ensure after dismissing SERP CTA, Fire button is not highlighted
1. Visit any other site
1. Ensure after dismissing CTA, Fire button is highlighted
1. Click on the Fire button
1. Ensure Fire dialog displayed and CTA appears: `Now tap below to burn your browsing history...`
1. Click on the "Clear All Tabs and Data"
1. Ensure Fire animation runs and app restarts

>Test 2 -Try Fire button before animation started-: 
1. Hardcode `zr` variant
1. Perform fresh install
1. Navigate to Browser screen
1. Click on the Fire button
1. Ensure Fire dialog displayed and CTA appears: `Now tap below to burn your browsing history...`
1. Dismiss dialog (go back to the browser)
1. Visit any non-SERP site
1. Ensure after dismissing CTA, Fire button is **not** highlighted

>Test 3 -Pulse animation always active till user interacts with it-: 
1. Hardcode `zr` variant
1. Perform fresh install
1. Navigate to Browser screen
1. Visit any non-SERP site
1. Ensure after dismissing CTA, Fire button is highlighted
1. Kill the app / send the app to background
1. Go back to the app
1. Ensure Fire button is still highlighted

>Test 4 -Hide all tips before fire education-: 
1. Hardcode `zr` variant
1. Perform fresh install
1. Navigate to Browser screen
1. Visit SERP Page
1. When CTA appears, **click on Hide all tips**
1. Visit any other site
1. Ensure no CTA and the Fire button is **not** highlighted
1. Click on the Fire button
1. Ensure Fire dialog displayed but no CTA shown
1. Click on the "Clear All Tabs and Data"
1. Ensure Fire animation runs and app restarts

>Test 5 -Edge case: Avoid browser CTA appearing on top of Fire dialog-: 
1. Hardcode `zr` variant
1. Perform fresh install
1. Navigate to Browser screen
1. Visit any site and **quickly** click on the Fire button
1. Wait till website finishes loading in the background
1. Ensure no browser CTA has appeared on top of the Fire dialog

---- Control variant: No Fire button education ----
>Test A -Full flow-: 
1. Hardcode `zm` variant
1. Perform fresh install
1. Navigate to Browser screen
1. Visit SERP Page
1. Ensure after dismissing SERP CTA, Fire button is **not** highlighted
1. Visit any other site
1. Ensure after dismissing CTA, Fire button is **not** highlighted
1. Click on the Fire button
1. Ensure Fire dialog displayed without any CTA
1. Click on the "Clear All Tabs and Data"
1. Ensure Fire animation did not run

---
###### Internal references:


CDRussell (commented):
This is the only question needing answered before merging I think. Otherwise, this is all tested and working well. Just going to do one more test on an older device just now.

CDRussell:
Seeing this in the logs, which is new

>W/LOTTIE: Animation contains merge paths. Merge paths are only supported on KitKat+ and must be manually enabled by calling enableMergePathsForKitKatAndAbove().

app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
```diff
@@ -16,44 +16,143 @@
...
         clearAllOption.setOnClickListener {
-            dismiss()
-            clearStarted()
+            onClearOptionClicked()
+        }
+        cancelOption.setOnClickListener {
+            cancel()
+        }
+
+        fireAnimationView.setAndPropagateUpFitsSystemWindows(false)
```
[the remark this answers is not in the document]
cmonfortep: This is a tricky one. I will need to add comment to document what's going on.

`// BottomSheetDialog wraps provided Layout into a CoordinatorLayout. We need to set FitsSystemWindows false programmatically to all parents in order to render layout and animation full screen.`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1172,6 +1165,7 @@ class BrowserTabFragment : Fragment(), FindListener, CoroutineScope, DaxDialogLi
     }
 
     override fun onDestroy() {
+        pulseAnimation.stop()
```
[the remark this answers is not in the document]
cmonfortep: Yes, I think we can pause the animation if the app goes to background, and leave `onDestroy` to clean the rest of view refs.

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -68,15 +67,37 @@ class CtaViewModel @Inject constructor(
...
+            withContext(dispatchers.io()) {
+                if (!variantManager.getVariant().hasFeature(FireButtonEducation)) return@withContext false
+                if (!daxOnboardingActive() || daxDialogFireEducationShown() || settingsDataStore.hideTips) return@withContext false
+
+                return@withContext dismissedCtaDao.any {
+                    it.ctaId == CtaId.DAX_DIALOG_TRACKERS_FOUND ||
+                            it.ctaId == CtaId.DAX_DIALOG_OTHER ||
+                            it.ctaId == CtaId.DAX_DIALOG_NETWORK
+                }
+            }
```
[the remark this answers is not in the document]
cmonfortep: sounds good, I can extract the whole `map` so this logic can only be applied to the expected flow

cmonfortep:
@CDRussell this is ready to review again. Thanks!

app/build.gradle
```diff
@@ -120,7 +120,7 @@ ext {
     retrofit = "2.8.1"
     ankoVersion = "0.10.4"
     glide = "4.11.0"
-    lottieVersion = "2.6.0-beta19"
+    lottieVersion = "3.4.0"
```
CDRussell, line 123: 👀  wow, we were behind!

app/src/main/java/com/duckduckgo/app/fire/FireAnimationLoader.kt
```diff
@@ -0,0 +1,39 @@
...
+import javax.inject.Inject
+
+class FireAnimationLoader @Inject constructor(
+    private val variantManager: VariantManager,
+    private val context: Context
+) : LifecycleObserver {
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_START)
+    fun onAppStart() {
+        if (variantManager.getVariant().hasFeature(VariantManager.VariantFeature.FireButtonEducation)) {
```
CDRussell, line 35: This is going to cause problems for detecting installation referrer, as it's eagerly allocating a variant before we can detect and assign a variant based on referrer.

Could we defer initializing the animation until `BrowserActivity`, or at least until after `LaunchViewModel.waitForReferrerData()` has finished?

app/src/main/java/com/duckduckgo/app/browser/PulseAnimation.kt
```diff
@@ -26,19 +26,40 @@ import android.widget.FrameLayout
...
 import androidx.core.view.doOnLayout
 import androidx.core.view.isVisible
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.OnLifecycleEvent
 import com.duckduckgo.app.global.view.setAllParentsClip
 
-class PulseAnimation {
+class PulseAnimation(private val lifecycleOwner: LifecycleOwner) : LifecycleObserver {
```
CDRussell, line 35: This is maybe ok, but just want to be sure: we are passing in a fragment to `PulseAnimation` and storing a reference to it. Is that likely to cause a leak?

Alternatively, do we need to pass in `lifecycleOwner` at all. as it is, the animation is responsible for registering itself and unregistering itself as a lifecycle observer. could we switch it around, so that the fragment is responsible for adding and removing `PulseAnimation` as a lifecycle observer?

app/src/main/java/com/duckduckgo/app/fire/FireAnimationLoader.kt
cmonfortep: Good catch, I didn't think about the referrer. For simplicity, I'm going to move it to `BrowserActivity`, and leave a comment saying that it can be moved back to the Application lifecycle once we have experiment results, as at that point we will remove the variant manager dependency.

app/src/main/java/com/duckduckgo/app/browser/PulseAnimation.kt
cmonfortep: Leak-wise we are fine here. Even if we have a ref to `lifecycleOwner`, `PulseAnimation` instance is only referenced from the fragment itself and it can be garbage collected when the fragment is destroyed.

About the `lifecycleOwner` comment. First, I'm super happy you are asking because it was something I spent some time thinking about, tried both approaches, and in the end I took the approach I thought it was more correct. But now I can validate it with you.

So, going to share here what made me move this register/unregister inside the class:
- PulseAnimation registers/unregisters itself as implementation detail based on animation playing or not. Fragment will not know as it doesn't affect it. Also why it's needed (pausing when goes to background) keeps as a detail inside the PulseAnimation.
- From the fragment, side code was simpler with this approach. Only need to call play / stop animation. No need to remember that we also need to register/unregister the instance as an observer.
- When we register observers in the fragment, usually it's because the fragment itself has the code or will orchestrate how to react to an event, here was not the case.

We can talk about this offline too, as it may be something very interesting to discuss! ping me if you want to do it :D

CDRussell: That explanation makes sense to me, thanks! I was mostly concerned about leaks more than whose responsibility it is to do what here anyway, and now you've confirmed that's not a worry, i'm happy. thanks the detailed explanation 👍

CDRussell:
@cmonfortep should we be calling `LottieCompositionFactory.fromRawRes(context, R.raw.fire_hero_rising)` on a background thread instead of the main thread?

CDRussell:
@cmonfortep, i still see this in the logs, suggesting we should be calling that function?

https://github.com/duckduckgo/Android/pull/916#issuecomment-680103341

cmonfortep:
1. Lottie already handles the threading. They don't block mainthread when parsing the json.
2. It doesn't matter, meaning it doesn't fail. But will double check that we are not doing more work than needed. IIRC, they check by key before loading (using as key the res id).
3. I saw it appearing when early loading the image and I think it does not apply there, but I will check too.

CDRussell (approved):
No matter what you find out about that merge paths warning, I think this is good to go!

cmonfortep:
All good here @CDRussell if the animation is already cached, we will not parse it twice. And that warning is pure informative when parsing the json (https://github.com/airbnb/lottie-android/blob/2354bad616c417aa246ec72a95aa372274d31250/lottie/src/main/java/com/airbnb/lottie/parser/ContentModelParser.java#L87).

CDRussell (approved):
nav bar looks better!

PR #940, opened 2020-09-24
Improve favicon rate

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1183582398892645
Tech Design URL: https://app.asana.com/0/1125189844152671/1194520756224089, https://app.asana.com/0/1125189844152671/1193640289465885
CC: 

**Description**:
This PR improves the favicon hit rate by using native APIs and selecting the best favicon available. It also separately stores them depending on their usage.

**Steps to test this PR**:

**Tabs**
1. Go to `tesla.com`.
1. Go to the tab switcher.
1. A favicon should be shown.
1. In the cache folder, under `faviconsTemp` a new folder with the tabId should have been created and inside you should find the favicon file.
1. Press the fire button.
1. The cache folder `faviconsTemp` should have been deleted.

**Bookmarks**
1. Go to `tesla.com`.
1. Add a bookmark.
1. Go to bookmarks.
1. A favicon should be shown.
1. In the cache folder, under `faviconsTemp` a new folder with the tabId should have been created and inside you should find the favicon file.
1. That same file should also exist in a new folder called `favicons`
1. Press the fire button.
1. The cache folder `faviconsTemp` should have been deleted but the `favicons` folder should still be there with the favicon file.
1. Go to bookmarks, the favicon should show.
1. Delete the bookmark.
1. The file inside the 'favicons` folder should have been deleted.

**Fireproof**
1. Go to `tesla.com`.
1. Fireproof the site.
1. In settings, go to fireproof websites.
1. A favicon should be shown.
1. In the cache folder, under `faviconsTemp` a new folder with the tabId should have been created and inside you should find the favicon file.
1. That same file should also exist in a new folder called `favicons`
1. Press the fire button.
1. The cache folder `faviconsTemp` should have been deleted but the `favicons` folder should still be there with the favicon file.
1. Go to settings and fireproof websites, the favicon should show
1. Delete the site.
1. The file inside the 'favicons` folder should have been deleted.

**Location**
1. Go to `mapquest.com`
1. Enable the device location permissions.
1. Allow DuckDuckGo to access the device's location.
1. A favicon should show in the `Grant mapquest.com` dialog.
1. In the cache folder, under `faviconsTemp` a new folder with the tabId should have been created and inside you should find the favicon file.
1. Select `always` or `deny always`.
1. Go to settings and location.
1. A favicon should show.
1. Press the fire button.
1. The cache folder `faviconsTemp` should have been deleted.
1. The file inside the `favicons` folder should have also been deleted (we only keep locations if the site is fireproofed).

**Fireproof location**
1. Go to `www.mapquest.com`
1. Enable the device location permissions.
1. Allow DuckDuckGo to access the device's location. 
1. Select `always` or `deny always`.
1. Fireproof the site.
1. Go to settings and fireproof sites, the favicon should show.
1. Go to settings and location, the favicon should show.
1. Press the fire button.
1. The cache folder `faviconsTemp` should have been deleted.
1. The file inside the `favicons` folder should still be there (the site is now fireproof)

**Keeping persisted favicons**
1. Go to `mapquest.com` 
1. Accept the permissions and select `always` or `deny always`
1. Add a bookmark.
1. Fireproof `mapquest.com`.
1. There should be on single file in the `favicons` directory with the tesla favicon.
1. Delete the site from one of bookmarks, location or fireproof.
1. The favicon file should still exist in the `favicons` folder.
1. Delete the site from one of the two places remaining.
1. The favicon file should still exist in the `favicons` folder.
1. Delete the site from the last place.
1. The favicon should now have been deleted from `favicons`.

**Fallback and swapping high quality favicons**
1. Go to `cnn.com` and add a bookmark.
1. Go to `tesla.com` and add a bookmark.
1. Go to bookmarks, both favicons should show.
1. Close all the tabs and manually delete the cache, this should also delete the `favicons` folder.
1. Go to bookmarks.
1. The `tesla.com` bookmarks should not show, instead our placeholder image will be shown. This happens because there is no fallback favicon for `tesla.com`.
1. The `cnn.com` favicon should eventually show up but it should be a low quality one.
1. Go to `cnn.com` and let the site load.
1. Go to `tesla.com` and let the site load.
1. Go to bookmarks.
1. Both favicons should now show up and the `cnn.com` favicon should now have a better quality.

**Update app version**
1. Install the app from `develop`.
1. Go to `cnn.com` and add to bookmarks.
1. Go to `tesla.com` and add to bookmarks.
1. Go to `youtube.com` and add to bookmarks.
1. Go to bookmarks, low quality favicons should show for cnn and youtube. Nothing for tesla.
1. Update the app from this branch.
1. Go to bookmarks.
1. Same low quality favicons should show as in the previous version.
1. Going to `cnn.com`, `tesla.com` or `youtube.com` should eventually update the favicons to higher quality ones.

---
###### Internal references:


CDRussell:
@marcosholgado, when I visit a site, i see it appear in `faviconsTemp` as expected. However, if I close that tab, I'd expect that tab's favicon to be deleted, but it's left on the disk. It doesn't seem to get deleted until the data is cleared.

Shouldn't we be deleting the favicon when the tab is closed?

CDRussell:
I noticed a problem with what seems to be a corrupt, or partially written, favicon file. I was working my way through the `Fallback and swapping high quality favicons` test case. Race condition on writing the file to disk; two things writing to the same file at once?

cnn-corrupt-favicon-actual
cnn-corrupt-favicon

marcosholgado:
> @marcosholgado, when I visit a site, i see it appear in `faviconsTemp` as expected. However, if I close that tab, I'd expect that tab's favicon to be deleted, but it's left on the disk. It doesn't seem to get deleted until the data is cleared.
> 
> Shouldn't we be deleting the favicon when the tab is closed?

Yup, good catch! Should be fixed now.

marcosholgado:
> I noticed a problem with what seems to be a corrupt, or partially written, favicon file. I was working my way through the `Fallback and swapping high quality favicons` test case. Race condition on writing the file to disk; two things writing to the same file at once?
> 

That's weird, haven't seen it before but I've been doing some experiments and yes, we could be writing the same file at once. Also this could happen if you exit the app while the file is being written (that's the most decent way I found to test this).

I've changed that code to block the thread so no race conditions should happen. I've also changed it to be non cancellable to avoid the scenario when the app gets killed while writing the file.

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconPersister.kt
```diff
@@ -61,46 +63,41 @@ class FileBasedFaviconPersister(
         }
     }
 
-    override suspend fun store(directory: String, subFolder: String, bitmap: Bitmap, domain: String): File {
-        return withContext(dispatcherProvider.io()) {
-
+    override suspend fun store(directory: String, subFolder: String, bitmap: Bitmap, domain: String): File? {
+        return runBlocking(dispatcherProvider.io() + NonCancellable) {
```
CDRussell, line 67: I don't think this is quite right. we shouldn't need to use `runBlocking` inside of a `suspend` function; using `withContext` will already wait until the `store` function has finished before return execution. Also, if this was called from different threads, you could still have multiple threads writing at the same time.

Is part of the problem around race conditions that `dispatcherProvider.io()` returns a dispatcher with a thread pool? so that even though you are using the same dispatcher each call, it might be delegating work to a different thread each time and so you'd potentially have multiple threads writing at the same time?

If the problem is limited only to the `store` function, would marking that as `synchronized` help any?

marcosholgado: Yeah that seemed to be the problem, `withContext` didn't wait at all and adding `runBlocking` did the trick. The whole idea was to synchronously execute the store method. 

I've changed the code a bit and extracted the store logic after `withContext` into a separate method and annotated with `@Synchronized`. Seems to work fine.

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
```diff
@@ -0,0 +1,186 @@
...
+            faviconPersister.copyToDirectory(cachedFavicon, FAVICON_PERSISTED_DIR, NO_SUBFOLDER, domain)
+        }
+    }
+
+    override suspend fun deletePersistedFavicon(url: String) {
+        val domain = extractDomain(url)
+        val remainingFavicons = persistedFaviconsForDomain(domain) - 1 // -1 because we are about to delete one
+        if (remainingFavicons <= 0) {
+            faviconPersister.deletePersistedFavicon(domain)
+        }
```
CDRussell, line 108: I think this would be clearer if you didn't subtract that minus 1 from the first line; i think it's a little confusing. is the logic essentially to check if there is only a single instance of the favicon, and delete it if so? so the logic boils down to `numberFavicons == 1`, or am i missing something? do you need to call `delete` if you find 0 favicons are returned from that count?

```
val remainingFavicons = persistedFaviconsForDomain(domain)
if (remainingFavicons == 1) {
    faviconPersister.deletePersistedFavicon(domain)
}
```

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private fun getFaviconUrl(url: String): Uri {
+        return if (url.toUri().host.isNullOrBlank()) {
+            "https://$url".toUri().faviconLocation() ?: throw IllegalArgumentException("Invalid favicon currentPageUrl")
+        } else {
+            url.toUri().faviconLocation() ?: throw IllegalArgumentException("Invalid favicon currentPageUrl")
+        }
+    }
```
CDRussell, line 143: two questions on this function

1. what is `currentPageUrl` - should that be a variable instead of a string literal?
2. do we need to throw the exception? could we make the function return nullable type `Uri?` and return `null` if we can't get extract a favicon URL from the url string? 

I don't feel too strongly about it, but the exception feels a little heavy handed, and given the exceptions are runtime and not checked, i think the `nullable` return type is safer here too, as we're forced to handle that and can't forget to catch the exception

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconPersister.kt
```diff
@@ -0,0 +1,143 @@
...
+            file
+        } else {
+            null
+        }
+    }
+
+    override suspend fun copyToDirectory(file: File, directory: String, newSubfolder: String, newFilename: String) {
+        withContext(dispatcherProvider.io()) {
+            val persistedFile = fileForFavicon(directory, newSubfolder, newFilename)
+            file.copyTo(persistedFile, overwrite = true)
```
CDRussell, line 61: given that this is writing files, should we `synchronize` this function too?

app/src/main/java/com/duckduckgo/app/di/AndroidBindingModule.kt
```diff
@@ -177,6 +178,9 @@ abstract class AndroidBindingModule {
     @ContributesAndroidInjector
     abstract fun brokenSiteNegativeFeedbackFragment(): BrokenSiteNegativeFeedbackFragment
 
+    @ContributesAndroidInjector
+    abstract fun siteLocationPermissionDialog(): SiteLocationPermissionDialog
```
CDRussell, line 182: curious why location permission dialog change in this PR?

app/src/main/java/com/duckduckgo/app/fire/fireproofwebsite/ui/FireproofWebsitesViewModel.kt
```diff
@@ -69,6 +72,12 @@ class FireproofWebsitesViewModel(
         )
     }
 
+    fun loadFavicon(url: String, view: ImageView) {
+        viewModelScope.launch {
```
CDRussell, line 76: the `viewModelScope` outlives the `ImageView` passed in here which makes me nervous. if you rotated the screen while the favicon was loading, is there a risk that we'd be trying to load an image into an `ImageView` which should be dead (because it belonged to the previous, unrotated `Activity`) does that matter?

app/src/main/java/com/duckduckgo/app/location/ui/LocationPermissionsActivity.kt
```diff
@@ -98,7 +98,7 @@ class LocationPermissionsActivity : DuckDuckGoActivity(), SiteLocationPermission
     }
 
     private fun editSiteLocationPermission(entity: LocationPermissionEntity) {
-        val dialog = SiteLocationPermissionDialog.instance(entity.domain, true)
+        val dialog = SiteLocationPermissionDialog.instance(entity.domain, true, "")
```
CDRussell, line 101: can we use named params here? I have no idea what that extra string param means without drilling into the method

app/src/main/java/com/duckduckgo/app/location/ui/LocationPermissionsViewModel.kt
```diff
@@ -79,6 +82,12 @@ class LocationPermissionsViewModel(
         )
     }
 
+    fun loadFavicon(url: String, view: ImageView) {
+        viewModelScope.launch {
```
CDRussell, line 86: similar concern as before around `viewModel` and `View` lifecycle.

app/src/main/java/com/duckduckgo/app/location/ui/SiteLocationPermissionDialog.kt
```diff
@@ -89,12 +108,18 @@ class SiteLocationPermissionDialog : DialogFragment() {
...
-
-            GlideApp.with(requireContext())
-                .load(faviconUrl)
-                .error(R.drawable.ic_globe_gray_16dp)
-                .into(imageView)
+            val tabId = args.getString(KEY_TAB_ID, "")
+
+            originUrl?.let { url ->
+                faviconJob?.cancel()
+                faviconJob = GlobalScope.launch {
```
CDRussell, line 115: how about `viewLifecycleOwner.lifecycleScope.launch {  }` instead of `GlobalScope.launch{ }` ?

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -201,7 +203,10 @@ class TabDataRepository @Inject constructor(
     override fun deleteAll() {
         Timber.i("Deleting tabs right now")
         tabsDao.deleteAllTabs()
-        GlobalScope.launch { webViewPreviewPersister.deleteAll() }
+        GlobalScope.launch {
```
CDRussell, line 206: i know this was here already, but launching a new coroutine here will likely mean a race condition, where we might finish deleting things before we restart the process, or maybe not.

the only function that calls `deleteAll()` is also a `suspend` function. what about if we made `deleteAll` also a `suspend` function, meaning you wouldn't need to `launch` a new coroutine? can you see any downsides?

CDRussell (commented):
This is looking great! There's a few questions, mostly about coroutine best practices, where i think there's room for improvement, but this PR is pretty close

app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
```diff
@@ -155,15 +155,27 @@ class DuckDuckGoFaviconManagerTest {
     }
 
     @Test
-    fun whenSaveToTempThenReplacePersistedFavicons() = coroutineRule.runBlocking {
+    fun whenSaveToTempIfFaviconHasBetterQualityThenReplacePersistedFavicons() = coroutineRule.runBlocking {
         val bitmap: Bitmap = Bitmap.createBitmap(1, 1, Bitmap.Config.RGB_565)
```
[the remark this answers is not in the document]
marcosholgado: Yeah, that makes sense. I previously had a test that checked the quality but is not needed anymore.

app/src/androidTest/java/com/duckduckgo/app/global/UriExtensionTest.kt
```diff
@@ -224,4 +224,14 @@ class UriExtensionTest {
         val favicon = "https://192.168.1.0".toUri().faviconLocation()
         assertEquals("https://192.168.1.0/favicon.ico", favicon.toString())
     }
+
+    @Test
+    fun whenUrlDoesNotHaveHostReturnNull() {
```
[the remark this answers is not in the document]
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksViewModel.kt
```diff
@@ -89,8 +98,15 @@ class BookmarksViewModel(val dao: BookmarksDao) : EditBookmarkListener, ViewMode
...
 
+    fun loadFavicon(url: String, view: ImageView) {
+        viewModelScope.launch {
+            faviconManager.loadToViewFromPersisted(url, view)
+        }
+    }
+
     fun delete(bookmark: BookmarkEntity) {
-        Schedulers.io().scheduleDirect {
+        viewModelScope.launch(dispatcherProvider.io()) {
```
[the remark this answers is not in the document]
marcosholgado: Yup, makes sense. I've added `NonCancellable`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -540,6 +540,27 @@ class BrowserTabViewModel(
         }
     }
 
+    override fun prefetchFavicon(url: String) {
+        faviconJob?.cancel()
```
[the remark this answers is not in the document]
marcosholgado: Changed!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1173,6 +1199,7 @@ class BrowserTabViewModel(
                 fireproofWebsiteRepository.fireproofWebsite(domain)?.let {
                     pixel.fire(PixelName.FIREPROOF_WEBSITE_ADDED)
                     command.value = ShowFireproofWebSiteConfirmation(fireproofWebsiteEntity = it)
+                    faviconManager.persistFavicon(tabId, domain)
```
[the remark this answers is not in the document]
marcosholgado: I added the named param for clarity. That `domain` comes directly from the fireproof website which has a domain rather than a url hence the confusion.

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
marcosholgado: Makes sense to me, changed to `== 1`

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
marcosholgado: Both things are coming from the old favicon code where we threw that same exception. It is captured in `downloadFromUrl` which is capturing at the same time exceptions from Glide. I've changed it to return null and wrapped the glide call with `runCatching`

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconPersister.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/di/AndroidBindingModule.kt
marcosholgado: We need to inject `FaviconManager` to show the favicon in the dialog.

app/src/main/java/com/duckduckgo/app/fire/fireproofwebsite/ui/FireproofWebsitesViewModel.kt
marcosholgado: I've changed that so it uses the Activity rather than the ViewModel. This involves quite a few changes, i.e. passing the lifecycle owner to the adapter as well as the faviconManager but it will avoid this problem.

app/src/main/java/com/duckduckgo/app/location/ui/LocationPermissionsActivity.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/location/ui/LocationPermissionsViewModel.kt
marcosholgado: Fixed as explained above.

app/src/main/java/com/duckduckgo/app/location/ui/SiteLocationPermissionDialog.kt
marcosholgado: Can't really use `viewLifecycleOwner` because at that point it crashes the app. I've changed it to `this.lifecycleScope`.

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
marcosholgado: Sounds good to me, only called from a suspend function so that makes sense, changed!

CDRussell (approved):
Great stuff!

PR #950, opened 2020-10-01
Blank web content if user is redirected to another site and new content does not load

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1196139873349464/f
Tech Design URL: https://app.asana.com/0/481882893211075/1195526928832633/f
CC: 

**Description**:
This PR fixes an issue of updating the URL while showing previous content during more time than expected.

How to reproduce:
1. Visit https://duckduckgo.github.io/privacy-test-pages/security/spoof-js-page-rewrite.html
2. Press run
3. click on "start"
4. Ensure user is redirected to a new site, and loading freezes with wrong content while showing duck.co in the omnibar

It could happen that genuine website (duck.co) loads and fake content does not appear. Try again after clearing all data (with fire button) to reproduce the issue (maybe you will need to do it multiple times, not very deterministic).

**Steps to test this PR**:
#Test 1: 
1. Open DDG app (and press fire to be sure we clear the cache)
1. Visit https://www.w3schools.com/code/tryit.asp?filename=GJA5PVPWMMIJ
2. Press run
3. click on "load and redirect"
4. Ensure url is updated to duck.co and fake content is visible for 1 second
5. Ensure web content hides after 1 second
6. Wait 10 seconds
7. Ensure new content appears

#Test 2:
Test the app browsing as you usually do on your daily basis:
Try searching something at DDG and click on some links.
Try to log in on any website (just to be redirected to another site automatically)
Ensure user experience is good as a user.

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -108,6 +108,8 @@ class BrowserWebViewClient(
                         webView.loadUrl(newUri.toString())
                         return true
                     }
+                    val navigationList = webView.safeCopyBackForwardList() ?: return false
```
subsymbolic, line 111: Could this break the loading of non-mainframe requests?

app/src/main/java/com/duckduckgo/app/browser/WebNavigationState.kt
```diff
@@ -48,6 +49,9 @@ fun WebNavigationState.compare(previous: WebNavigationState?): WebNavigationStat
     if (this is EmptyNavigationState)
         return PageNavigationCleared
 
+    if (this is OverrideURL)
+        return Redirected
```
subsymbolic, line 53: This state is confusing, it sounds like a redirect happened but every normal load will go through this state.  I also worry that we haven't married the state change and the side-effects together e.g this causes the page to blank out but something different causes it to unblank. Let's talk about it on a call.

app/src/main/java/com/duckduckgo/app/browser/WebNavigationState.kt
```diff
@@ -146,6 +150,26 @@ data class EmptyNavigationState private constructor(
     override val hasNavigationHistory: Boolean = false
 }
 
+@Suppress("DataClassPrivateConstructor")
+data class OverrideURL private constructor(
+    val nextUrl: String,
```
subsymbolic, line 155: This isn't used.

app/src/main/java/com/duckduckgo/app/browser/WebNavigationState.kt
cmonfortep: Cool that you spotted this too. As we discussed on our call, I was not very happy either with this approach. I will remove this logic from here as I don't think is adding any value. thanks.

cmonfortep:
Thanks @subsymbolic for the previous review. Pushed changes with a different approach. Could you take a second look? thanks

subsymbolic:
@cmonfortep this now has conflicts, can you resolve those and I'll review on Monday?

cmonfortep:
@subsymbolic conflicts are solved now. Also, I've just pushed some changes with the approach we discussed in our call today. This can be reviewed on Monday, thanks!

subsymbolic (approved):
Great!

Issue #851, opened 2020-06-07
NFC tags don't open in duckduckgo


andreaskern:
**1. What I do:**
Scan an NFC tag that holds web url

**2. What happens:**
Google Chrome opens the url

**3. What should happen:**
Duckduckgo opens the url

After some investigation it seems that duckduckgo is missing the intent-filter.
a quick search for 'nfc' in the google chrome app manifest reveals two entries.
I don't know if nfc permissions are needed to.

![image](https://user-images.githubusercontent.com/739277/83968983-42c21880-a8cd-11ea-888d-9016adfdef89.png)

https://chromium.googlesource.com/chromium/src/+/b71e98cdf14f18cb967a73857826f6e8c568cea0/chrome/android/java/AndroidManifest.xml#245

ashwinpilgaonkar:
Hi @cmonfortep, can I work on this?

cmonfortep:
@ashwinpilgaonkar sure!

ashwinpilgaonkar:
Hi @cmonfortep,

Since I dont have any physical NFC tags to test out my changes with, I'm trying to emulate it using another test application that I'm building. 

I just have one question - when the NFC tag is detected, does it broadcast an intent in a similar manner as the code snippet given below?
```
                Intent intent = new Intent("android.nfc.action.NDEF_DISCOVERED");
                intent.setType("text/plain");
                intent.putExtra(NfcAdapter.EXTRA_NDEF_MESSAGES, "Test text");
                sendBroadcast(intent);
```

cmonfortep:
According to the docs, yes, it does -> https://developer.android.com/guide/topics/connectivity/nfc/nfc#filter-intents

There's more information about the nfc tags also there https://developer.android.com/guide/topics/connectivity/nfc/nfc#dispatching

ashwinpilgaonkar:
> According to the docs, yes, it does -> https://developer.android.com/guide/topics/connectivity/nfc/nfc#filter-intents
> 
> There's more information about the nfc tags also there https://developer.android.com/guide/topics/connectivity/nfc/nfc#dispatching

Yes, I had gone through these docs earlier but could only find info pertaining to filtering an NFC intent. Whereas, I was more interested in finding out how to mock a physical NFC tag.

Anyway, I've made a PR with the changes, which I was able to test successfully using the following code snippet -
```
                Intent intent = new Intent();
                intent.setAction("android.nfc.action.NDEF_DISCOVERED");
[...]

PR #960, opened 2020-10-07
Added support - NFC tags

ashwinpilgaonkar:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/851
Tech Design URL: N/A
CC: 

**Description**:
On scanning an NFC tag that holds web URL, the DuckDuckGo Android app was unable able to open the URL. This PR now fixes that issue. I needed to set up a new `<intent-filter>` in `Manifest.xml`, that would look for `android.nfc.action.NDEF_DISCOVERED`.

**Steps to test this PR**:
1. Ideally should be tested using an NFC tag that holds a web URL.

2. I tested my changes by building another small application that sends a `android.nfc.action.NDEF_DISCOVERED` intent, thus mocking a physical NFC tag. I used the following code snippet to achieve this - 
```
                Intent intent = new Intent();
                intent.setAction("android.nfc.action.NDEF_DISCOVERED");
                intent.setData(Uri.parse("https://www.example.com"));
                startActivity(intent);
```
---
###### Internal references:

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -10,7 +10,6 @@
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
-    <uses-permission android:name="android.permission.NFC"/>
 
     <application
         android:name="com.duckduckgo.app.global.DuckDuckGoApplication"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -10,6 +10,7 @@
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.NFC"/>
 
     <application
         android:name="com.duckduckgo.app.global.DuckDuckGoApplication"
@@ -213,6 +214,17 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <data android:mimeType="text/plain" />
             </intent-filter>
+
+            <!-- Allow app to be opened using an NFC tag -->
+            <intent-filter>
+                <action android:name="android.nfc.action.NDEF_DISCOVERED" />
+                <category android:name="android.intent.category.DEFAULT"/>
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="https" />
+                <data android:scheme="http" />
+            </intent-filter>
+
         </activity>
 
         <activity


JeffAment:
I happen to have a spare NFC tag, so I tested this for you; everything seems to work as intended. Good work! :smile:

cmonfortep:
Thanks @JeffAment !!!! You appeared at the right moment here, magical!

btw, how was the user experience? I guess the first time you scan a tag, a system dialog appears requesting nfc permission, right? or does it work automatically without any user prompt?

ashwinpilgaonkar:
> btw, how was the user experience? I guess the first time you scan a tag, a system dialog appears requesting nfc permission, right? or does it work automatically without any user prompt?
>
I think the NFC permission is granted directly at the time of installation. It isn't requested during runtime like other permissions such as location, storage, etc.

When I tested it, it automatically works without any user prompt. Once the intent is sent, the app opens the URL. I'm assuming it works the same way with an actual NFC tag. @JeffAment would probably be able to give us an accurate answer to this.

JeffAment:
> Thanks @JeffAment !!!! You appeared at the right moment here, magical!

No problem, @cmonfortep; happy to help! Hocus-pocus :tophat: :rabbit: :tada:

> When I tested it, it automatically works without any user prompt. Once the intent is sent, the app opens the URL. I'm assuming it works the same way with an actual NFC tag. @JeffAment would probably be able to give us an accurate answer to this.

I can confirm that there was no permission prompt for me. I also checked the "App permissions" screen after successfully using the NFC feature, and it doesn't mention NFC permissions anywhere, as do other browsers with NFC support, so I believe that @ashwinpilgaonkar is correct in saying that the permission is granted directly at install-time without user intervention.

I would also like to mention that my Google Pixel XL Android 10 successfully opens URLs stored on the NFC tag even when the `android.permission.NFC` permission is omitted from the manifest. The official documentation on this is vague, but I'm thinking that the permission might not be necessary for this instance since it's just a simple intent filter for which Android does the heavy lifting. I'm not necessarily advocating for removing the permission, but I just wanted to note this. FWIW if it were up to me, I'd probably leave it in to avoid potential headaches down the road.

ashwinpilgaonkar:
> I would also like to mention that my Google Pixel XL Android 10 successfully opens URLs stored on the NFC tag even when the `android.permission.NFC` permission is omitted from the manifest. The official documentation on this is vague, but I'm thinking that the permission might not be necessary for this instance since it's just a simple intent filter for which Android does the heavy lifting.
>
I'm not a 100% sure about this, but I think that the NFC permission is essential for other NFC related activities - such as switching on WiFi/Bluetooth when you tap an NFC device. Probably isn't required for opening URLs though.

> I'm not necessarily advocating for removing the permission, but I just wanted to note this. FWIW if it were up to me, I'd probably leave it in to avoid potential headaches down the road.
>
Yes, I completely agree with you on that one.

cmonfortep:
Thanks both, @JeffAment and @ashwinpilgaonkar, since our goal here is just to open URLs from NFC tags, I want to double-check if that permission is needed or not. We will merge this PR anyways (with or without permission), agree with Jeff that keeping the permission will remove problems in the future, but before that, I just want to double-check if we need it for our current usage.

Two main reasons:
1. AFAIK, new permission in the manifest that are accepted during installation will block auto-updates for users, till they go to GPlay and accept that permission.
2. Optics are important in our app. We don't want to request something that we don't really need, that could also confuse the user when they see that request from us.

I shared the question internally, will report back here once we know something else.

Thanks both again!

cmonfortep:
Hi there again! I managed to get some NFC tags and do some deep testing with a few devices to validate if we need that NFC permission. It seems we don't need it (🎉 ) in order to accept NFC intents, so I would remove that permission from the manifest and we are good to go.

sounds good @ashwinpilgaonkar ?

app/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,7 @@
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.NFC"/>
```
cmonfortep, line 13:
```suggestion
```

cmonfortep (changes requested):
We can safely remove this permission from the manifest.

ashwinpilgaonkar:
> Hi there again! I managed to get some NFC tags and do some deep testing with a few devices to validate if we need that NFC permission. It seems we don't need it (🎉 ) in order to accept NFC intents, so I would remove that permission from the manifest and we are good to go.
> 
> sounds good @ashwinpilgaonkar ?

Yep, I've made the required changes.

cmonfortep (approved):
LGTM Thanks!

Issue #953, opened 2020-10-01
Android: Protect app from crashing due to unsafe input from 3rd parties


cmonfortep:
**Describe the bug**
Our app can be launched from 3rd party apps, and we access an Intent which was created by the other app.

> If the other app has a badly created Intent, or the calling app's process dies before we can process the Intent, it can cause our app to crash when trying to process it.

Other browsers have had to add a safety net around processing 3rd party intents previously
- Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1147992
- Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=469219

**Potential fixes**
- Firefox fixes by wrapping all incoming Intents into a SafeIntent, which has try/catches embedded in it to prevent crashes.
- Chrome runs all Intents through a sanitize method [**preferred**]
  - attempts to pull out an extra from the Intent.
  - the extra isn't there, but trying to get it will cause the exception to happen if the   - Intent is malformed or invalid. 
  - Returns an Intent with a nulled out extras bundle if exception is caught

Play Store:
Link to crashes 
Number of users affected: low
Total occurrences: low

If applicable, please include information about any settings that may be relevant to this issue (e.g. selected theme, auto clear option, etc).

**Expected behavior**
- Protect against our app crashing due to problems with 3rd party apps

**Smartphone (please complete the following information):**
 - DDG App Version: all versions.

abdulbasit75:
I would like to work on this issue, please assign me.

cmonfortep:
Sure @abdulbasit75, feel free to contribute on this issue. 👍

code-bluebird:
Hi, 
I would like to work on this issue, could somebody help me with these details.
- Which activity processes the received intent from 3rd parties ?
- and please provide any more details of error log if possible.

cmonfortep:
This is the stacktrace of the error described above:

```
android.os.BadParcelableException: 
  at android.os.Parcel.readParcelableCreator (Parcel.java:3059)
  at android.os.Parcel.readParcelable (Parcel.java:2981)
  at android.os.Parcel.readValue (Parcel.java:2883)
  at android.os.Parcel.readArrayMapInternal (Parcel.java:3261)
  at android.os.BaseBundle.initializeFromParcelLocked (BaseBundle.java:292)
  at android.os.BaseBundle.unparcel (BaseBundle.java:236)
  at android.os.BaseBundle.getBoolean (BaseBundle.java:899)
[...]

PR #967, opened 2020-10-15
Intentfix

wseemann:
Task/Issue URL: Android: Protect app from crashing due to unsafe input from 3rd parties #953
Tech Design URL: N/A
CC: N/A

Android: Protect app from crashing due to unsafe input from 3rd parties.

**Steps to test this PR**:
1. Send invalid intent to app, verify the app does not crash.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -29,6 +29,7 @@ import com.duckduckgo.app.bookmarks.ui.BookmarksActivity
 import com.duckduckgo.app.browser.BrowserViewModel.Command
 import com.duckduckgo.app.browser.BrowserViewModel.Command.Query
 import com.duckduckgo.app.browser.BrowserViewModel.Command.Refresh
+import com.duckduckgo.app.browser.intent.ContextUtils
 import com.duckduckgo.app.browser.rating.ui.AppEnjoymentDialogFragment
 import com.duckduckgo.app.browser.rating.ui.GiveFeedbackDialogFragment
 import com.duckduckgo.app.browser.rating.ui.RateAppDialogFragment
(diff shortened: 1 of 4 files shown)


cmonfortep (changes requested):
Thanks for submitting this PR.

This fixes the crash report we shared in https://github.com/duckduckgo/Android/issues/953, but it still crashes when accessing the intent at:
```
at com.duckduckgo.app.global.IntentExtensionKt.getIntentText(IntentExtension.kt:23)
        at com.duckduckgo.app.fire.DataClearerForegroundAppRestartPixel.registerIntent(DataClearerForegroundAppRestartPixel.kt:62)
        at com.duckduckgo.app.browser.BrowserActivity.onNewIntent(BrowserActivity.kt:132)
```

About the proposed solution:

It's ok to use, on every intent access, the new proposed class `ContextUtils`, but I would like to suggest also creating a method that sanitizes the intent (like Chrome does). That method could be used to reduce verbosity when accessing multiple times the same `Intent`.

The method logic will be:
1. try to get extra (`intent.extras`) from an intent (it will crash here)
2. If exception captured, return same intent without extras (so it will not crash anymore)
3. return sanitized intent

cmonfortep:
@wseemann are you open to continue working on this PR? Let me know if you are still interested or I can open it for new contributors. Thanks!

wseemann:
@cmonfortep I'm working on the revisions. Sorry for the delay, it's Monday so I was busy catching up on work at my full time job. I will try and get the revisions in within the next day or so.

cmonfortep:
No problem, take as time as you need here. I will be on the other side for anything you need here 👍

cmonfortep:
@wseemann Thanks so much for contributing, however, we need to decline this PR (as it is now) since we noticed the code comes from another repository and we only accept unique contributions.

wseemann:
@cmonfortep No problem. I figured an open source project would want to leverage code that already works from another quality open source project rather then reinvent the wheel. After all, that's kind of the point of making things open source.

PR #980, opened 2020-10-27
Show Location Permission message only once

malmstein:
Task/Issue URL: https://app.asana.com/0/1157893581871903/1198881849371917

**Description**:
In a previous release, we allowed sites to request the user's location. In order to remind our users that a site has location permission, we show a Toast indicating it. 

However, this Toast is very repetitive, because every time the site is refreshed this message appears again. This is particularly intrusive in sites that are refreshed often, like our SERP.

This PR fixes this, making sure that the message is only shown once per tab life span. Closing the tab and open another will prompt this message once.

**Steps to test this PR**:

***In current version in Production:***
1. Open app
2. Search for "Restaurants near me" in our SERP
3. Enable Precise Location
4. Allow system permission
5. Allow permission "Always" (Choosing only once this session will not show the Toast)
6. Refresh page
7. Toast should be shown
8. Refresh page
9. Toast should be shown

***In this branch***
1. Open app
2. Search for "Restaurants near me" in our SERP
3. Enable Precise Location
4. Allow system permission
5. Allow permission "Always" (Choosing only once this session will not show the Toast)
6. Refresh page
7. Toast should be shown
8. Refresh page
9. Toast should not appear 

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -866,7 +866,7 @@ class BrowserTabViewModel(
         val permissionEntity = locationPermissionsRepository.getDomainPermission(domain)
         permissionEntity?.let {
             if (it.permission == LocationPermissionType.ALLOW_ALWAYS) {
-                if (!locationPermissionMessages.containsKey(domain)){
+                if (!locationPermissionMessages.containsKey(domain)) {
                     setDomainHasLocationPermissionShown(domain)
                     command.postValue(ShowDomainHasPermissionMessage(domain))
                 }
@@ -875,7 +875,7 @@ class BrowserTabViewModel(
     }
 
     @VisibleForTesting
-    fun setDomainHasLocationPermissionShown(domain: String){
+    fun setDomainHasLocationPermissionShown(domain: String) {
         locationPermissionMessages[domain] = true
     }
 
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -874,8 +874,7 @@ class BrowserTabViewModel(
         }
     }
 
-    @VisibleForTesting
-    fun setDomainHasLocationPermissionShown(domain: String) {
+    private fun setDomainHasLocationPermissionShown(domain: String) {
         locationPermissionMessages[domain] = true
     }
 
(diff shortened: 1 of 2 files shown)


cmonfortep:
When testing this, the flow is different to the one described above:
1. Open app
1. Search for "Restaurants near me" in our SERP
1. Enable Precise Location
1. Allow system permission
1. Allow permission "Always" (Choosing only once this session will not show the Toast)
1. Toast should be shown -> Toast was not shown
1. Refresh page
1. Toast should not appear -> Toast was shown

location_toast

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -865,11 +866,19 @@ class BrowserTabViewModel(
...
-                command.postValue(ShowDomainHasPermissionMessage(domain))
+                if (!locationPermissionMessages.containsKey(domain)) {
+                    setDomainHasLocationPermissionShown(domain)
+                    command.postValue(ShowDomainHasPermissionMessage(domain))
+                }
             }
         }
     }
 
+    @VisibleForTesting
```
cmonfortep, line 877: is there anything we can do to avoid this?

Inside the test, instead of using `setDomainHasLocationPermissionShown`, what will happen if we call twice `loadUrl("https://www.example.com", isBrowserShowing = true)`? I expect the first call to set `locationPermissionMessages[domain] = true`, and the second one not showing the Toast. Is that correct?

malmstein: Not really, `loadUrl` does a whole other bunch of stuff, triggering several `onPageChanged` and `onPageUpdated` that defies the purpose. That was my initial approach but it doesn't work.

malmstein:
> When testing this, the flow is different to the one described above:
> 
> 1. Open app
> 2. Search for "Restaurants near me" in our SERP
> 3. Enable Precise Location
> 4. Allow system permission
> 5. Allow permission "Always" (Choosing only once this session will not show the Toast)
> 6. Toast should be shown -> Toast was not shown
> 7. Refresh page
> 8. Toast should not appear -> Toast was shown
> 
> location_toast

This is actually how it should work, you only get notified AFTER the page has loaded AND the dialog was shown. I'm updating the steps to test for clarity. Thanks @cmonfortep !

cmonfortep:
Thanks for updating the description @malmstein. I've been checking the original task and I don't see any specific behavior we should follow, so I would assume we are pursuing here "just to show the toast once per tab".

However, I don't understand something. When the user clicks on "always", why are we showing the toast after refreshing?
My expectation here will be:

1. Open app
2. Search for "Restaurants near me" in our SERP
3. Enable Precise Location
4. Allow system permission
5. Allow permission "Always" (Choosing only once this session will not show the Toast)
6. **Any** Refresh page
9. Toast should not appear

On a new tab:
1. Open a new tab
2. Search for "Restaurants near me" in our SERP
3. if user selected "Always"
4. Toast should appear

My proposal is to add the domain to `locationPermissionMessages` when the user selects "Always", so we don't show any new prompt for that "session". Does it make sense?

malmstein:
We've discussed this online and decided to:
- If the permission has just been allowed, then we don't show the Toast after refresh
- Opening the same domain in another tab will show the Toast

cmonfortep (approved):
LGTM. Tested and works as described.

Thanks for the pairing and for addressing the comments. :shipit:

malmstein:
Thanks @cmonfortep !

PR #985, opened 2020-10-30
Improve ViewModel Commands assert methods

cmonfortep:
Task/Issue URL: https://app.asana.com/0/414730916066338/1198986542400933/f
Tech Design URL: 
CC: 

**Description**:
We spotted that some of the ViewModel-Command Assert methods that are using `ArgumentCaptor` internally were not capturing the emitted values by a `captor`. That means that when getting values from that `captor` the result will be always an emptyList. This could give us false positives when using `assertCommandNotIssued` or similar methods.

This PR fixes that issue. Ensure to capture emitted values before accessing them.

**Steps to test this PR**:
1. Run the test
1. 

---
###### Internal references:


app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -3012,13 +3009,24 @@ class BrowserTabViewModelTest {
...
     private inline fun <reified T : Command> assertCommandNotIssued() {
-        val issuedCommand = commandCaptor.allValues.find { it is T }
-        assertNull(issuedCommand)
+        val defaultMockingDetails = DefaultMockingDetails(mockCommandObserver)
+        if (defaultMockingDetails.invocations.isNotEmpty()) {
+            verify(mockCommandObserver, atLeastOnce()).onChanged(commandCaptor.capture())
+            val issuedCommand = commandCaptor.allValues.find { it is T }
+            assertNull(issuedCommand)
+        } else {
+            assertTrue(true)
```
CDRussell, line 3018: Not doing a full review as I haven't tested, but just skimming it

`assertTrue(true)` seems unnecessary. We only want to assert when there's a chance of the assert failing; `assertTrue(true)` will always pass, so you're as well doing nothing at all.

cmonfortep: Good point, I will remove it.

malmstein:
Good stuff @cmonfortep!

malmstein (approved):

CDRussell:
@cmonfortep still planning on merging this?

cmonfortep:
@CDRussell yes 🤞 `assertCommandNotIssued` is now fixed but on test is failing. 1) test is incorrect 2) there's a small bug in the app. I will try to find some time this week and see if I can take a look and fix what's failing.

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -3016,13 +3013,24 @@ class BrowserTabViewModelTest {
...
+        }
     }
 
     private inline fun <reified T : Command> assertCommandIssuedTimes(times: Int) {
-        val timesIssued = commandCaptor.allValues.count { it is T }
-        assertEquals(times, timesIssued)
+        if (times == 0) {
+            assertCommandNotIssued<T>()
+        } else {
+            verify(mockCommandObserver, atLeastOnce()).onChanged(commandCaptor.capture())
```
CDRussell, line 3030: doesn't `verify` already provide support for what you're doing here?

`verify(mockCommandObserver, times(times)).onChanged(commandCaptor.capture())` ?

cmonfortep: Not really. Here we are asserting that a specific command of type `T` has not been emitted. 

`verify(mockCommandObserver, times(times)).onChanged(commandCaptor.capture())` will check any command emitted.

CDRussell: `it is T`

ah yeah, missed that detail

malmstein:
Anything outstanding here @cmonfortep?

cmonfortep:
Yes @malmstein, check my comment above (https://github.com/duckduckgo/Android/pull/985#issuecomment-744534640)

We should validate if the test that is failing is because the feature has a bug or the test is wrong. Since you are familiar with this feature, could you check that? Thanks.

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -2460,7 +2462,6 @@ class BrowserTabViewModelTest {
         givenDeviceLocationSharingIsEnabled(true)
         givenLocationPermissionIsEnabled(true)
         givenCurrentSite(domain)
-
         testee.onSiteLocationPermissionSelected(domain, LocationPermissionType.ALLOW_ONCE)
```
cmonfortep, line 2465: Hey @malmstein, This test `whenDomainRequestsSitePermissionAndUserAllowedSessionPermissionThenPermissionIsAllowed` is failing for me.

I took a look but I'm not familiar with the logic here. What I've seen is:
- when calling `onSiteLocationPermissionSelected` I guess we expect to simulate that for that domain a permission is selected. However the method does nothing because `locationPermission?.let {` -> locationPermission is `null` at that point.
- When we call `givenLocationPermissionIsEnabled`, `CheckSystemLocationPermission` is emitted then.

could you take a look? Thanks!

malmstein: Discussed online with @cmonfortep and refactored the test to verify the functionality.

malmstein (approved):
🚢

PR #986, opened 2020-10-31
fix: don't add empty domains to whitelist

donfuxx:
Task/Issue URL: https://app.asana.com/0/414730916066338/1199127532699902/f
Tech Design URL: 
CC: 

**Description**:

It doesn't seem to have any impact, but I found it odd during debugging that there is an empty domain in the whitelist:
Screenshot from 2020-10-31 13-59-51
This seemed to happen due to empty line at the end of the whitelist, so I added a check to see if the line is not blank before mapping it to the white list domain model.

An alternative solution would be to not allow constructing white list domain models without valid domains in the first place.

A unit test for `downloadTemporaryWhitelist` would be nice :upside_down_face: 

**Steps to test this PR**:
1. open browser, but with empty cache
1. check domains list of whitelist in debugger, no more empty domain added at the end

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/trackerdetection/api/TrackerDataDownloader.kt
```diff
@@ -86,7 +86,8 @@ class TrackerDataDownloader @Inject constructor(
             val body = response.body()!!
 
             appDatabase.runInTransaction {
-                temporaryTrackingWhitelistDao.updateAll(body.lines().map { TemporaryTrackingWhitelistedDomain(it) })
+                temporaryTrackingWhitelistDao.updateAll(body.lines().takeWhile { it.isNotBlank() }
```
cmonfortep, line 89: Edge case and shouldn't happen, but if the received file contains an empty line inside the body, we will not save all the received domains.

e.g:
```
domain1.com
domain2.com

domain3.com
domain4.com

```

We will store only domain1 and 2, but not domain3 and 4. All 4 domains should be stored.

cmonfortep (commented):
Thanks for your contribution. I've found something to check.

app/src/main/java/com/duckduckgo/app/trackerdetection/api/TrackerDataDownloader.kt
cmonfortep: One recommendation (haven't tested how it works): Instead of `takeWhile` use `filter` or `filterNot`.

donfuxx: you are right! I adjusted that to use `filter` instead.
I also added a unit test for that, but I put that in a separate PR since it would introduce a new dependency to Junit5, see https://github.com/duckduckgo/Android/pull/989
Testing simple things like this is faster using plain jvm unit tests and Junit5 allows it to keep your test classes really compact :wink:

cmonfortep (approved):
LGTM thanks for the contribution

PR #989, opened 2020-11-01
Fix empty entry whitelist optional addon: Introduce Junit5 unit tests

donfuxx:
Task/Issue URL: 
Tech Design URL: 
CC: 

**Description**:
same like PR https://github.com/duckduckgo/Android/pull/986 just with additional unit tests added. 
Adds [Junit5](https://junit.org/junit5/docs/current/user-guide/#writing-tests-assertions-third-party) test dependency which allows the use of ParameterisedTest which is really handy to test lots of scenarios in one test method.
Also removed some test dependencies that seemed unused (only used in androidTest)

**Note**: I have noticed that there wasn't a `test` source set added yet to your project, there was only an `androidTest` yet. So you might need to add another step to your CI to run `./gradlew testDebugUnitTest` to not miss this tests.

**Steps to test this PR**:
1. run ./gradlew testDebugUnitTest in terminal or use Android studio to run jvm unit tests by clicking `run test` icon next to `ListExtensionKtTest`
1. See green test results:
Screenshot from 2020-11-01 11-44-07

---
###### Internal references:


cmonfortep:
Quick comment: You can run parameterized tests also in junit4. See `DataClearerTimeKeeperTest` and the usage of `@RunWith(Parameterized::class)`. Will that work for you scenario?

Side note: I recommend you to open an issue first before submitting any PR, just to validate the proposal with us and avoid any extra work.

donfuxx:
Right, I have added the Junit4 style test to the other PR then https://github.com/duckduckgo/Android/pull/986 
I was a bit confused though why some of the plain unit tests are in `andriodTest` and therefore need to run the emulator. Tests like this could be run in the JVM which would be faster since the emulator doesn't need to startup etc.
I'll try to create an issue first next time, this was basically something I stumbled over when I was looking at the issue with theguardian.com website ;-)

cmonfortep:
Hey @donfuxx , just wanted to check, do we need this PR if we have https://github.com/duckduckgo/Android/pull/986? since like this was just adding junit5 and you sorted out that right?

cmonfortep:
Closing this as it has been already solved in #986

PR #991, opened 2020-11-03
Fix favicon display view recycling

### LINKED ISSUE #975: Bookmark Icons
Issue:
Bookmark icons are not the sites original one's, my guess this would be a fetching issue or a local issue, ive noticed this as in the image below i dont think DuckDuckGo Traffic has a google icon. 😂

Image:
IMG_20201024_084509

aitorvs:
Task/Issue URL: https://app.asana.com/0/1198194956794324/1198918323600485/f
Tech Design URL: 
CC: 

**Description**:
Favicons are displayed in a number of places, eg. tab switcher, bookmarks, fireproof websites etc.
All those places use a listview adapter. The bug appears when the view holder is recycled and the new item
to be displayed in it does not have a favicon. In which case, whatever icon was set in the recycled holder is showed.

This PR fixes that issue in the `FaviconManager` as it is the common entry point for bookmarks, tab switcher, fireproof websites etc

Fixes #975

NOTE: I also considered doing this in the `FaviconDownloader` by allowing a nullable `File` in the `loadFaviconToView` method. However this is too late as results in a UX that shows the previous (holder) icon and then it changes to the proper one.

**Steps to test this PR**:
1. open multiple tabs to trigger view recycling (20-25)
2. open as (one of the) last tab(s) one website without favicon (eg. http://http://aosabook.org)
3. go to tab switcher and scroll down
4. websites that have no favicon should show the (default) globe icon - previously they would show the icon of a previously recycled view)

Similar steps would work for bookmarks and fireproof websites

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
@@ -38,7 +38,6 @@ import com.nhaarman.mockitokotlin2.never
 import com.nhaarman.mockitokotlin2.verify
 import com.nhaarman.mockitokotlin2.whenever
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
 import org.junit.Before
 import org.junit.Rule
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
@@ -129,7 +129,7 @@ class DuckDuckGoFaviconManagerTest {
 
     @Test
     fun whenLoadToViewFromTempIfCannotFindFaviconThenDownloadFromUrl() = coroutineRule.runBlocking {
-        val view = ImageView(InstrumentationRegistry.getInstrumentation().targetContext)
+        val view: ImageView = mock()
         val url = "https://example.com"
 
         testee.loadToViewFromTemp("subFolder", url, view)
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
@@ -19,6 +19,7 @@ package com.duckduckgo.app.browser.favicon
 import android.graphics.Bitmap
 import android.widget.ImageView
 import androidx.core.net.toUri
+import androidx.test.platform.app.InstrumentationRegistry
 import com.duckduckgo.app.CoroutineTestRule
 import com.duckduckgo.app.bookmarks.db.BookmarksDao
 import com.duckduckgo.app.browser.favicon.FileBasedFaviconPersister.Companion.FAVICON_PERSISTED_DIR
@@ -37,6 +38,7 @@ import com.nhaarman.mockitokotlin2.never
 import com.nhaarman.mockitokotlin2.verify
 import com.nhaarman.mockitokotlin2.whenever
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
 import org.junit.Before
 import org.junit.Rule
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
```diff
@@ -64,13 +66,23 @@ class DuckDuckGoFaviconManager constructor(
...
         loadToViewFromDirectory(url, FAVICON_PERSISTED_DIR, NO_SUBFOLDER, view)
     }
 
     override suspend fun loadToViewFromTemp(subFolder: String, url: String, view: ImageView) {
+        // avoid displaying the previous favicon when the holder is recycled
+        preloadDefaultImageIntoView(view)
         loadToViewFromDirectory(url, FAVICON_TEMP_DIR, subFolder, view)
     }
 
+    private suspend fun preloadDefaultImageIntoView(view: ImageView) {
```
cmonfortep, line 80: This looks great. Just one proposal from my side. Since the implementation detail of what placeholder we use for favicons (`ic_globe_gray_16dp`) is inside the `faviconDownloader`, just to keep any similar logic in one place, will it be possible to move this method inside the `FaviconDownloader`?

cmonfortep (commented):
Great job finding this root cause and introducing this quick fix. Just one minor comment related to the place where we have declared the new method. But looks great!

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
aitorvs: Makes sense, will do

aitorvs: @cmonfortep addressed the comment. Let me know any other comments you might have. Feel free to resolve this conversation if all is OK

app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
```diff
@@ -106,6 +107,16 @@ class DuckDuckGoFaviconManagerTest {
         verify(mockFaviconDownloader).getFaviconFromUrl(url.toUri().faviconLocation()!!)
     }
 
+    @Test
+    fun whenLoadToViewFromPersistedIfCannotFindFaviconThenLoadDefaultFaviconIntoView() = coroutineRule.runBlocking {
+        val view: ImageView = mock()
```
cmonfortep, line 112: Minor comment: we have some tests using `mock()` and other tests using `ImageView(InstrumentationRegistry.getInstrumentation().targetContext)`. No strong preference between any, maybe I would go fo `ImageView`. Are you ok using the same approach in each test?

cmonfortep (approved):
Thanks @aitorvs, this looks great. Thanks for also filling in the description with all the details.

I'm excited about your first PR here! and ooking forward to seeing more coming! 🚀

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
```diff
@@ -64,10 +64,14 @@ class DuckDuckGoFaviconManager constructor(
     }
 
     override suspend fun loadToViewFromPersisted(url: String, view: ImageView) {
+        // avoid displaying the previous favicon when the holder is recycled
+        faviconDownloader.loadDefaultFaviconToView(view)
```
marcosholgado, line 68: I think we can move these calls inside `loadFaviconToView` in the `GlideFaviconDownloader`

app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
aitorvs: Oh, my bad, it was necessary to have the actual view before, now I can revert to `mock` all of them as it was before. I prob missed reverting one of them, will do

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
aitorvs: Yeah, I thought about that. Left it like this to avoid having `loadToViewFromDirectory` doing more than just loading the icon from the directory, it could an undesired side effect to reset the icon in the view when calling `loadToViewFromDirectory` if we call it from other places in the future.
What do you think?
Anyway given that that's not currently the case and that it is a private method I am also ok with moving it into `loadToViewFromDirectory` if you feel strongly about it.

marcosholgado: Nah, not feeling strongly about it :)

app/src/androidTest/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManagerTest.kt
aitorvs: @cmonfortep done. Resolving the comment :D

PR #993, opened 2020-11-04
Android: New internal Pixel api

cmonfortep:
Task/Issue URL: https://app.asana.com/0/414730916066338/1198967073193796/f
Tech Design URL: 
CC: 

**Description**:
Currently, our Pixel api relies on a network call to send any pixel. This covers most of our usecases when we need to send a Pixel to our backend.
There are some scenarios where this is not enough:
- Sending pixels when we are about to restart the app (pressing Fire)
- If network fails

This PR introduces a new method in our existing API: enqueuePixel. This method is opt-in, so we can use it in cases where we are afraid about losing pixels and those are important.

`enqueuePixel` will:
- store the pixel in the database, and
- try to send the pixel to remote API
- if succeeds, remote pixel
- if fails, next time another pixel is added or the application comes again to the foreground, try to send it again.

In this iteration, I discarded using a WorkManager to keep it simple and to make it as synchronous as possible (as soon as the pixel is stored in our database, we will try to send it). Making it synchronous will make it easier to test PRs that include any pixel validation step, the reviewer will not need to wait any amount of time to validate if a pixel is sent.

**Steps to test this PR**:

> Fresh install

**Test 1:**
1. Go to any part of the code where we use `pixel.fire`, and replace it for `pixel.enqueueFire`
1. Perform a fresh install
1. Go to the screen where you changed the pixel method call
1. Peform the action required to enqueue the pixel
1. ensure Pixel is sent (see logs `Pixel sent`)
1. ensure in the database, the pixel has been removed

**Test 2:**
1. Turn off data on your emulator or device
1. Go to the screen where you changed the pixel method call
1. Peform the action required to enqueue the pixel
1. ensure Pixel is not sent (we should not see any log saying `Pixel sent`)
1. ensure in the database, all unsent triggered pixels are present
1. Turn on data on your emulator or device
1. Perform any of the following actions: another action that enqueues a new pixel OR send the app to background and bring it back to foreground OR press fire button
1. ensure Pixel is sent (see logs `Pixel sent`)
1. ensure in the database, the pixel has been removed

> Database Migration

**Test 1:**
1. Install develop and run the app
1. Checkout this branch and install it on your device
1. Ensure migration succeeds

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
```diff
@@ -262,30 +271,98 @@ interface Pixel {
...
+                            .subscribe({
+                                pixelDao.delete(pixelEntity)
+                                Timber.v("Pixel sent: ${pixelEntity.pixelName} with params: temp temp)")
+                            }, {
+                                Timber.w(it, "Pixel failed: ${pixelEntity.pixelName} with params:  temp temp)")
+                            })
+                    )
+                }
+
+            }
```
aitorvs: Given that the API is already in RxJava land, why making the `PixelDao#pixels` return a `Flow`? this bit of code would read better if it would return just an `Observable` of the entity. What do you think?

Disclaimer: I know this PR is still WIP and I don't have yet much context in general but thought I'd drop some comments anyway. Feel free to disregard if necessary.

app/src/main/java/com/duckduckgo/app/statistics/store/PixelDao.kt
```diff
@@ -16,14 +16,18 @@
...
-import com.duckduckgo.app.cta.model.DismissedCta
+import androidx.room.*
 import com.duckduckgo.app.statistics.model.PixelEntity
 import kotlinx.coroutines.flow.Flow
 
 @Dao
 interface PixelDao {
     @Query("select * from pixel_store")
-    fun pixels(): Flow<PixelEntity>
+    fun pixels(): Flow<List<PixelEntity>>
```
aitorvs, line 26: Along with previous comment, what do you think about making it return an Observable so that we avoid crossing Flow<>RxJava boundaries?

app/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
cmonfortep: Good point, yeah! I was tempted at the beginning to migrate every to coroutines and flow, but then discarded it to reduce some time. Yes, it has sense to keep this part using rx.

Keep those comments coming, actually, you saved me some time on the testing side throwing that comment before I start with them.

app/src/main/java/com/duckduckgo/app/statistics/api/PixelSender.kt
```diff
@@ -0,0 +1,115 @@
...
+    @OnLifecycleEvent(Lifecycle.Event.ON_START)
+    fun onAppForegrounded() {
+        parentDisposable = pixelDao.pixelsObs()
+            .doOnEach { sendPixelsDisposable.clear() }
+            .map {
+                Timber.i("Pixel sending: $it")
+                it.forEach { pixelEntity ->
+                    sendPixelToRemote(pixelEntity)
+                }
+            }.subscribe()
```
aitorvs, line 57: The code below works perfectly without having issues when deleting the elements

```kotlin

        parentDisposable = pixelDao.pixelsObs()
            .subscribeOn(Schedulers.io())
            .flatMapIterable { it }
            .switchMapCompletable { pixel ->
                Timber.i("Pixel sending $pixel")
                pixelDao.delete(pixel)
                sendPixel(pixel)
            }
            .subscribe()
```

cmonfortep: I've just updated this class. Let me know if you see something we can improve. Thanks!

app/src/main/java/com/duckduckgo/app/statistics/store/PixelDao.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import androidx.room.*
+import com.duckduckgo.app.statistics.model.PixelEntity
+import io.reactivex.Observable
+
+@Dao
+interface PixelDao {
+
+    @Query("select * from pixel_store")
+    fun unsentPixels(): Observable<List<PixelEntity>>
```
aitorvs, line 27: nit: this is actually getting all pixels in the DB so I would just call it `pixels`. The name `unsentPixels` seems to hint that there's another kind of pixels in the DB, while it is not.
What do you think?

app/src/main/java/com/duckduckgo/app/statistics/api/PixelSender.kt
aitorvs: It looks awesome!

app/src/main/java/com/duckduckgo/app/statistics/model/PixelEntity.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    @TypeConverter
+    fun fromQueryParams(value: Map<String, String>): String {
+        return Adapters.queryParamsAdapter.toJson(value)
+    }
+}
+
+class Adapters {
+    companion object {
+        private val moshi = JsonModule().moshi()
```
CDRussell, line 49: It's a bit strange having to instantiate a dagger module, to then instantiate a `Moshi` instance which is marked as `Singleton`.

I don't know how big of a deal that is, though, but feels _off_ to me. Any thoughts on if it's a problem, or if there's a nicer way to handle this?

app/src/main/java/com/duckduckgo/app/statistics/store/PixelDao.kt
CDRussell: You're technically right @aitorvs, although for some reason I do get some sort of reassurance from the name `unsentPixels` that I don't get from `pixels` (as it could read like we save all of them forever without it)

maybe `pendingPixels()` would be clear and still reassuring? (I don't feel strongly about this)

CDRussell (commented):
Reviewed the code; just about to test it now

app/src/main/java/com/duckduckgo/app/statistics/model/PixelEntity.kt
aitorvs: This is not really an issue, when I first saw it, it also felt odd, but in the end this is bypassing all DI and just creating an instance of moshi. Only that is creating one with the same configuration that we use in the instances provided by DI.
Imo it would suffice to just do `Moshi.Builder().build()`. Prob we don't even need to add any other adapter for this.

Note: this pattern is also done in `TdsTracker.ky`

cmonfortep: Yeah, it's a pattern we use in different parts. I don't think is an issue neither, just an implementation detail on how to go from Map to String in order to store it in the database. I agree that could be confusing because we use `Moshi` also to communicate with different API, but it's unrelated to that instance we use there. Also better to keep away other adapters that could add unexpected transformations.

CDRussell: We could consider pulling out the builder (of the builder) somewhere outside of the Dagger modules at some point, so that both dagger and anything else that needs it go there, instead of everything going to Dagger module. Not a big deal; just odd.

Happy to leave as is, though 👍

aitorvs (commented):
Looks good to me. Nice work @cmonfortep

CDRussell (approved):
All tested and working well. Great stuff @cmonfortep 🎉

cmonfortep:
I went for `pendingPixelsDao.pixels()`. I liked the idea of being explicit in the class name level, and be more simple on method names. Hope that's ok for you folks @aitorvs @CDRussell

PR #1007, opened 2020-11-10
Add 3 fire button animations and make it a setting

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1198977345184094/f
Tech Design URL: 
CC: 

**Description**:
This PR brings to users a way to customize their experience on the Fire button. We are including 3 different animations (fire by default) and 1 option to disable it (based on users' feedback).

What changes are included?
- Add new 3 Fire button animations: Inferno, whirlpool, airstream
- Add an option to disable the animation
- Add a new option inside settings based on ✓ Design: Setting to select Fire button animation
- Add pixels approved during kickoff:
- Pixel which animation was used on each clear data
- Pixel when the user accesses new option
- Pixel when the user selects a new animation

**Steps to test this PR**:
**Test1: Default animation**
1. Fresh install
1. Clear data
1. Ensure Fire animation appears
1. Ensure pixel `m_fd_a` sent with params `fa=fai`

**Test2: Disable fire animation**
1. Go to settings
1. Click on Fire Button Animation
1. Ensure pixel `m_fas_o` sent
1. Select "None"
1. Ensure pixel `m_fas_s` sent with params `fa=fann`
1. Go back to the browser
1. Clear data
1. Ensure no Fire animation appeared
1. Ensure pixel `m_fd_a` sent with params `fa=fann`

**Test3: Whirlpool and Airstream animation**
1. Repeat Test 2 but choosing another animation
1. Ensure pixels are sent with expected params, same as in Test2 (whirlpool `fa=fawp`, airstream `fa=faas`)
1. Ensure expected animation executed when clearing data

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/fire/LottieAnimationLoader.kt
```diff
@@ -21,21 +21,25 @@ import androidx.lifecycle.Lifecycle
...
-import com.duckduckgo.app.browser.R
-import com.duckduckgo.app.statistics.VariantManager
+import com.duckduckgo.app.settings.db.SettingsDataStore
 import javax.inject.Inject
 
-class FireAnimationLoader @Inject constructor(
-    private val variantManager: VariantManager,
-    private val context: Context
-) : LifecycleObserver {
+interface AnimationLoader : LifecycleObserver {
```
aitorvs, line 27: I think this file is now an implementation detail of settings, I would move it into the settings package.
Also the naming is now confusing to me, it very specifically preloads file animations, however the interface and its impl names are very generic, as if it could be used for any animation.
I think the prev FireAnimationLoader naming was better

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
```diff
@@ -182,6 +195,19 @@ class SettingsViewModel @Inject constructor(
         )
     }
 
+    fun onFireAnimationSelected(selectedFireAnimation: FireAnimation) {
+        if (settingsDataStore.isCurrentlySelected(selectedFireAnimation)) {
+            Timber.v("User selected same thing they already have set: $selectedFireAnimation; no need to do anything else")
+            return
+        }
+        settingsDataStore.selectedFireAnimation = selectedFireAnimation
+        animationLoader.preloadSelectedAnimation()
```
aitorvs, line 204: I would have loved to be able to just listen to changes in settings. In that case, animationLoader would just subscribe to whatever is important to thee class, and preload accordingly.
This view model should not be worried about this unrelated business logic (like preloading animations etc.)

This would not necessarily be needed for this PR, but having settings that other classes can subscribe to, simplifies a lot thee logic in general

Thoughts?

app/src/main/java/com/duckduckgo/app/fire/LottieAnimationLoader.kt
cmonfortep: Very good points. I would move it to the settings package.

And about naming, using `FireAnimationLoader` sounds good now that the concrete class is `LottieAnimationLoader`.  TBH, all this code is pretty tied to Lottie because we want to precache the json file to improve performance, and I would have avoided an interface here. But the interface is because we can't `mock` final classes in AndroidTests...so it was tricky to find some generic names for something very specific. But I think your proposal is good.

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
cmonfortep: Yes, I agree with your proposal. Actually, that will be something we will have for free with DataStore (currently in Alpha https://developer.android.com/topic/libraries/architecture/datastore), and most probably I would have done too if the stored value was in Room.

cmonfortep:
@marcosholgado I will introduce some small changes to this PR. I will ping you once is ready again.

cmonfortep:
hey @marcosholgado This is ready for review again. 💯

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
marcosholgado: Yup, agreed. I wouldn't change it now just to be consistent with the rest of the settings but an overall migration to DataStore at some point would be nice.

app/src/main/java/com/duckduckgo/app/di/SystemComponentsModule.kt
```diff
@@ -47,4 +50,9 @@ open class SystemComponentsModule {
     @Provides
     fun appIconModifier(context: Context, appShortcutCreator: AppShortcutCreator): IconModifier =
         AppIconModifier(context, appShortcutCreator)
+
+    @Provides
```
marcosholgado, line 54: This seems to be tied to the app lifecycle, just wondering if it makes sense for it to be a Singleton?

app/src/main/java/com/duckduckgo/app/fire/LottieAnimationLoader.kt
marcosholgado: Same as @aitorvs said, it feels like `FireAnimationLoader` was a better name. I feel that if I see `AnimationLoader` in the codebase I might be tempted to use it for other animation related stuff.

How about changing the interface to `FireAnimationLoader` and the implementation to `LottieFireAnimationLoader`?

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -146,6 +159,10 @@ class SettingsSharedPreferences @Inject constructor(private val context: Context
         return currentlySelected == clearWhenOption
     }
 
+    override fun isCurrentlySelected(fireAnimation: FireAnimation): Boolean {
```
marcosholgado, line 162: Consider renaming to `isFireAnimationCurrentlySelected` or something similar to give a bit more context since is kinda confusing  with the other `isCurrentlySelected` methods.

app/src/main/res/values/string-untranslated.xml
```diff
@@ -24,4 +24,13 @@
     <string name="useOurAppNotificationDescription">Here\'s a simple way to reduce its reach.</string>
     <string name="useOurAppShortcutAddedText">Success! %s has been added to your home screen.</string>
     <string name="useOurAppDeletionDialogText">Checking your feed in DuckDuckGo is a great alternative to using the Facebook app!&lt;br/&gt;&lt;br/&gt;But if the Facebook app is on your phone, it can make requests for data even when you\'re not using it.&lt;br/&gt;&lt;br/&gt;Prevent this by deleting it now!</string>
+
+    <!-- Fire Animation settings -->
+    <string name="settingsFireAnimation">Fire Button Animation</string>
```
marcosholgado, line 29: In theory, if this is not an experiment we should be getting translations for all of these strings before merging.

marcosholgado (changes requested):
Tested on API 21 and above, all working as expected. I just left a few comments around the code.

app/src/main/java/com/duckduckgo/app/di/SystemComponentsModule.kt
cmonfortep: Because it's stateless and part of the application component, I usually tend to leave out the `@singleton`. Just micro-optimization in dagger that can avoid some checks when I know it's not really important to be singleton, and knowing how I will use it in our code too. So, it's same result at the end.

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
cmonfortep: Not sure about this one.

So, you have your settings, and you want to know if something is selected using `isCurrentlySelected`. Even tho there are currently 3 methods, the argument type will tell you what you ar checking. 

`isCurrentlySelected(fireAnimation)`is telling you that you are checking if that fire animation is selected inside settings.

adding `isFireAnimationCurrentlySelected(fireAnimation)` has the same result to me but longer. WDYT?

app/src/main/res/values/string-untranslated.xml
cmonfortep: ha! that's good to know. So we will need translations before merging...good to know.

Going to move them to `strings.xml` but then I expect bitrise failing until translations are ready...

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
marcosholgado: Is like `selectedFireAnimationSavedValue` or `automaticallyClearWhatSavedValue` where we basically follow the opposite approach, the return types there tell us what it is but we are still adding a more meaningful function name.

marcosholgado (approved):
All looking good to me 👍 , just left a comment following up one of the conversations but I don't feel strongly about it.

app/src/main/res/values/string-untranslated.xml
marcosholgado: That's the theory anyway although I probably have not been clear about it when we talked about translations. Feel free to merge and create another PR for translations once you have them if you want to release asap.

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
cmonfortep: That's slightly different. You will get a compile error declaring two methods with the same name and the same arguments even if they return different types.

But wanted to make clear that this is just a personal preference, and I'm pretty sure I'm not even consistent with myself 😅, so I'm ok changing the name too. I think my main issue was the name being too long in that case.

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -53,12 +56,15 @@ interface SettingsDataStore {
     var appNotificationsEnabled: Boolean
     fun isCurrentlySelected(clearWhatOption: ClearWhatOption): Boolean
     fun isCurrentlySelected(clearWhenOption: ClearWhenOption): Boolean
+    fun isCurrentlySelected(fireAnimation: FireAnimation): Boolean
     fun hasBackgroundTimestampRecorded(): Boolean
     fun clearAppBackgroundTimestamp()
 }
 
 class SettingsSharedPreferences @Inject constructor(private val context: Context) : SettingsDataStore {
```
aitorvs, line 64: @cmonfortep now that you're touching this class maybe is a good moment to clean the DI. SettingsSharedPreferences is both annotated with `@Inject` for constructor injection and also has a provides method in `BrowserAutoCompleteModule`. I think we should remove one of them

cmonfortep: I can do it. I will take a look tomorrow.

marcosholgado: I think the `BrowserAutoCompleteModule` should be deleted and then add `@Binds abstract fun bindSettingsDataStore(settingsDataStore: SettingsSharedPreferences): SettingsDataStore` into the `StoreModule`

aitorvs: If `SettingsSharedPreferences` is an implementation detail of `SettingsDataStore`, I think it is best not having a `Binds`. To avoid getting the `SettingsSharedPreferences` type (which should be private).
I think in the `StoreModule` we could add a provider method that takes `Context` and returns `SettingsDataStore`

marcosholgado: Haven't checked but I assumed we were currently returning `SettingsSharedPreferences` which was an injected parameter in the `BrowserAutoCompleteModule` @Provides method hence why `@Binds` made sense to me. If not, then disregard :)

marcosholgado: By the way, isn't it a very common pattern having an implementation constructor using `@Inject` and then using `@Binds` to provide it? I'm sure I've seen that multiple times pretty much everywhere. Like you say, you could then inject the actual implementation which you should not be able to so... I think we should get rid of the whole `StoreModule` to sort that out. Maybe it makes more sense to create a LHF task to address all those issues rather than just fixing one here?

aitorvs: Not sure if it is a common pattern, I'd say it is an anti-pattern. That basically makes both types available through DI, which may not be desired (I think it should not be desired). 
The purpose of having an interface is to hide the type that implements it. Exposing both the implementing type and interface type partly defeats the purpose of having an interface in the first place, but also may be dangerous. It is very easy to miss that I should be asking for one not the other, not clear which one I should be asking for, if one is marked as `Singleton` (or differently scoped) it can even cause bugs. 

Imo a better pattern would be to even have the implementing type, in this case the `SettingsSharedPreferences` as private so that it is not visible to anybody. But if that can't be done, then use a provider that only exposes the final `interface` type.

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -61,7 +60,7 @@ interface SettingsDataStore {
     fun clearAppBackgroundTimestamp()
 }
 
-class SettingsSharedPreferences @Inject constructor(private val context: Context) : SettingsDataStore {
+class SettingsSharedPreferences constructor(private val context: Context) : SettingsDataStore {
```
cmonfortep, line 63: @aitorvs I've removed the `@Inject` from this class. Let me know if it's ok or you prefer to remove the provider method.

cmonfortep:
@marcosholgado I've updated the PR including translations. Could you take a quick look at the last 4 commits? Thanks

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
marcosholgado: Tiny thing, you can remove `constructor` as well.

marcosholgado (approved):
Nothing special, just a minor comment but feel free to merge. Good job!

Issue #648, opened 2019-12-16
Feature request: Do not track


sinus-x:
Would it be possible to add "Do not track" option to the settings? It's the single feature I'm missing in comparison with Firefox Focus.
Thanks a lot, for both the app and the project in general

Poopooracoocoo:
DNT is itself a fingerprinting issue but choices and defaults are what is important.

sinus-x:
Well, yes. The W3C discontinued the project calling it "a failed experiment" a year ago, as I just found out.

There isn't much data on support, however, Google isn't using it and Safari dropped the support completly. Maybe there really is no reason for adding it, if it is ignored as a privacy and used as a tracking feature.

Ryan451:
@sinus-x DuckDuckGo Privacy Browser uses "Global Privacy Control" (GPC) instead of "Do Not Track" (DNT)

https://spreadprivacy.com/announcing-global-privacy-control/

Ryan451:
@sinus-x DDG Privacy Browser uses "Global Privacy Control" (GPC) instead of "Do Not Track" (DNT)
https://spreadprivacy.com/announcing-global-privacy-control/

cmonfortep:
Our browser supports GPC as commented above (enabled by default in settings). We don't have plans to introduce DNT support. Thanks.

PR #1020, opened 2020-11-24
Enable GPC by Default

cmonfortep:
Task/Issue URL: https://app.asana.com/0/1197921851987065/1199132975397325/f
Tech Design URL: 
CC: https://app.asana.com/0/0/1199349637642694/f

**Description**:
This PR enables GPC by default to new users.

**Steps to test this PR**:

**Setting enabled**
1. Fresh install
1. Go to settings
1. Ensure GPC is enabled

**Signal present and back navigation**
1. Go to globalprivacycontrol.org
1. At the top of the page it should say `GPC signal detected`.
1. Click on `test against the reference server`
1. glitch.me site should load and both the header and DOM signals should show as present.
1. Inspect the tab by going to chrome://inspect in your desktop browser, select the tab opened in your phone and click network.
1. Back in the app, go to the bottom of the site and click to `View page source` 
1. In the inspected tab, in networks, select the first request and inspect the headers. The sec-gpc header should be set to `1`.
1. In the phone tab, click the back button.
1. The header and DOM should still show as present.
1. Click the back button and the sign should still be detected
1. Click back and you should be taken to the home screen.
1. Click forward until you reach github again navigating as expected and with the header set correctly.

**Blank link**
1. Go to https://www.w3schools.com/code/tryit.asp?filename=GL0N23JVASNE and execute the code.
1. App should not crash

**Blank link signal present**
1. Go to https://www.w3schools.com/code/tryit.asp?filename=GLNSDACQ1Y8L and execute the code.
1. You should be taken to the glitch site and the header and DOM signals should be present.

**Page rewrite**
1. Open DDG app (and press fire to be sure we clear the cache)
1. Visit https://duckduckgo.github.io/privacy-test-pages/security/spoof-js-page-rewrite.html
1. Press Start
1. Ensure url is updated to duck.co and fake content is visible for 1 second
1. Ensure web content hides after 1 second

**Fireproof**
1. Enable the fireproof toggle in Settings -> Fireproof Websites
1. Go to different sites, i.e. facebook, twitter, netflix, wsj.com
1. After login in those sites you should see the fireproof dialog

---
###### Internal references:


aitorvs:
LGTM!

aitorvs (approved):

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -100,18 +100,23 @@ class BrowserWebViewClient(
...
                 is SpecialUrlDetector.UrlType.Web -> {
                     if (requestRewriter.shouldRewriteRequest(url)) {
                         val newUri = requestRewriter.rewriteRequestWithCustomQueryParams(url)
-                        webView.loadUrl(newUri.toString())
+                        webView.loadUrl(newUri.toString(), globalPrivacyControl.getHeaders())
                         return true
                     }
                     if (isForMainFrame) {
-                        webViewClientListener?.willOverrideUrl(url.toString())
+                        if (globalPrivacyControl.isGpcActive()) {
```
cmonfortep, line 114: I've found some issues after introducing this change, so most probably this is affecting the callbacks we usually receive when loading a site.

What I tested?
- https://duckduckgo.github.io/privacy-test-pages/security/spoof-js-page-rewrite.html -> with GPC on this is not working now
- Login detection (fireproof) -> not working neither

I couldn't spot the problem. I will take a look later today.

cmonfortep: This is Solved now

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -980,7 +980,9 @@ class BrowserTabFragment :
                 builtInZoomControls = true
                 displayZoomControls = false
                 mixedContentMode = WebSettings.MIXED_CONTENT_COMPATIBILITY_MODE
-                setSupportMultipleWindows(true)
+                // There's an issue when overriding urls (GPC) and supporting blank links.
+                // Disabled until we fix the issue.
+                setSupportMultipleWindows(false)
```
subsymbolic, line 985: Comments explaining bugs/workarounds get confusing fast as the code around them changes.
Can we move this into a separate self-documenting method e.g "disableMultipleWindowsToSupportGPC()".

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -100,20 +105,24 @@ class BrowserWebViewClient(
...
                     if (requestRewriter.shouldRewriteRequest(url)) {
                         val newUri = requestRewriter.rewriteRequestWithCustomQueryParams(url)
-                        webView.loadUrl(newUri.toString())
+                        webView.loadUrl(newUri.toString(), globalPrivacyControl.getHeaders())
                         return true
                     }
                     if (isForMainFrame) {
                         webViewClientListener?.willOverrideUrl(url.toString())
+                        if (globalPrivacyControl.isGpcActive()) {
+                            webView.loadUrl(url.toString(), globalPrivacyControl.getHeaders())
```
subsymbolic, line 121: I see two potential gotchas here:
* this could get called for subframes on lower APIs and potentially allow an ad banner to be loaded fullscreen or break a payment page. Perhaps we should only support this for API 24 and above where we can be sure the request is from the main frame.
* we could end up in an endless loop. Perhaps we could check that the header isn't already there. We get the full request from API 24

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
cmonfortep: > nit: Can you add the Asana task reference here? it will be easier to get the context back when tackling the fix

Thanks for the comments. We are enabling this again. This was just a temp proposal to release before the deadline. We have more time now and we will investigate further.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
cmonfortep: Hey @subsymbolic I've been exploring what you said about "subframes", but I've not been able to find any scenario where that method is called from an iframe. 🤷‍♀️  I'm still trying to find something but without any luck. For `shouldInterceptRequest` we receive requests that are coming from a subframes, but not in the other method.

For example, I've playing around using this: https://jsfiddle.net/cmonfortep/d1poxc75/12/ where I refresh the iframe. I see the request via `shouldInterceptRequest`, but not on the `shouldOverrideUrlLoading`.

app/src/main/java/com/duckduckgo/app/brokensite/api/BrokenSiteSender.kt
```diff
@@ -84,5 +87,6 @@ class BrokenSiteSubmitter(
         private const val MODEL_KEY = "model"
         private const val WEBVIEW_VERSION_KEY = "wvVersion"
         private const val SITE_TYPE_KEY = "siteType"
+        private const val GPC = "gpc"
```
marcosholgado, line 90: FYI doing this doesn't really do anything, we need to change the perl script that aggregates the data, otherwise that information will get lost.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
marcosholgado: I've been playing with user agents, requests, etc this week and gut feeling says to do all this in `shouldInterceptRequest` same way that we do the https update instead of `shouldOverrideUrlLoading` as I initially thought.

marcosholgado: I've tried to move the code around like I said to `shouldInterceptRequest` and made some changes on a separate branch `feature/marcos/gpc_default_on` based on what I've been doing to change the user agent on a separate project. I think it works or at least is pretty close, it definitely doesn't crash on a _blank link.

@cmonfortep we'll catch up on Monday because I'm not 100% convinced if my version is breaking https://duckduckgo.github.io/privacy-test-pages/security/spoof-js-page-rewrite.html (I kinda think is not)

app/src/main/java/com/duckduckgo/app/brokensite/api/BrokenSiteSender.kt
marcosholgado: I've tried to move the code around like I said to `shouldInterceptRequest` and made some changes on a separate branch `feature/marcos/gpc_default_on` based on what I've been doing to change the user agent on a separate project. I think it works or at least is pretty close

@cmonfortep we'll catch up on Monday because I'm not 100% convinced if my version is breaking https://duckduckgo.github.io/privacy-test-pages/security/spoof-js-page-rewrite.html (I kinda think is not)

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
marcosholgado: Closing this conversation, this code is now in `shouldInterceptRequest` plus it checks if the header already exists to avoid an infinite loop like @subsymbolic said which I was able to reproduce.

subsymbolic: Just saw the earlier message to me, so sorry I missed that. Glad this is resolved now 👍

app/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
```diff
@@ -52,6 +52,7 @@ interface Pixel {
         WEB_RENDERER_GONE_KILLED("m_d_wrg_k"),
         BROKEN_SITE_REPORTED("m_bsr"),
         BROKEN_SITE_REPORT("epbf"),
+        WEB_VIEW_REDIRECT_HEADERS("m_wb_rh"),
```
marcosholgado, line 55: @cmonfortep do we still want this pixel or should I remove it?

marcosholgado:
@cmonfortep I've updated the PR with my changes and added new test cases in the description. I left a few questions in the existing code related to the new pixel introduce and the new broken site variable. 

I've assigned the PR to you, is a bit weird because you created it but a lot has changed 😅

app/src/main/java/com/duckduckgo/app/brokensite/api/BrokenSiteSender.kt
cmonfortep: Yes, we know that, we added it just to advance some work. See comments in https://app.asana.com/0/1198504124118156/1198503585911423/f

app/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
cmonfortep: If we are moving away from `shouldOverrideUrlLoading`, this can be removed. 👍

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
```diff
@@ -106,6 +121,16 @@ class WebViewRequestInterceptor(
         return null
     }
 
+    private fun shouldAddGcpHeaders(request: WebResourceRequest): Boolean {
+        val headers = request.requestHeaders
+        return (
+            globalPrivacyControl.isGpcActive() &&
+                !headers.containsKey(GlobalPrivacyControlManager.GPC_HEADER.toLowerCase(Locale.ROOT)) &&
```
cmonfortep, line 128: I have a question: why are we using `toLowerCase` here? is it because when we add the headers, those are changed by something external to us? do you know why? all headers are lowercase? if it's a documented standard, why don't we change `GlobalPrivacyControlManager.GPC_HEADER` to be lowercase instead?

app/src/main/java/com/duckduckgo/app/brokensite/api/BrokenSiteSender.kt
marcosholgado: Cool, thanks!

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
marcosholgado: Good catch, yup the headers get changed to lowercase for some reason so I did during my initial testing and never removed that. I've changed the header to always be lowercase.

app/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
marcosholgado: Nice, removed!

cmonfortep (commented):
Good work @marcosholgado.

Sadly https://duckduckgo.github.io/privacy-test-pages/security/spoof-js-page-rewrite.html fails. 

I quickly checked what's the issue, and I think we should somehow merge or communicate `willOverrideUrl` with `redirectTriggeredByGpc`.
`willOverrideUrl` decides to show a blank screen until the page finishes loading. When we redirect the url to inject headers, progress completes and blank screen is discarded.

cmonfortep:
BTW, I've installed a apk on my phone to keep it using for a day. Everything else works like a charm, so that's a amazing!

Let me know if you want to zoom to know more about the spoofing fix.

marcosholgado:
We'll get there... We had a problem with back navigation which should now be fixed (description updated to test that). Apart from that we still have the spoofing problem. I've created https://github.com/duckduckgo/Android/pull/1038 with a possible solution that we can discuss, I think it works better than what we have and it makes a bit more sense but until we decide I didn't want to put all that code in this PR.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -319,7 +319,6 @@ class BrowserTabViewModel(
     private lateinit var tabId: String
     private var webNavigationState: WebNavigationState? = null
     private var httpsUpgraded = false
-    private var shouldShowNewProgress = true
```
cmonfortep: @marcosholgado I've removed this logic since I was detecting some issues when testing GPC. Sometimes the loading progress was not appearing at all. I've noticed the logic was affected when using low speed connection and clicking on a link before site finished loading.

Disabling worked great when testing this. And since we are just enabling GPC headers for sites that are Header consumers, the problem has low impact when it happens.

app/src/main/java/com/duckduckgo/app/globalprivacycontrol/GlobalPrivacyControlManager.kt
```diff
@@ -17,18 +17,43 @@
...
-class GlobalPrivacyControlInjectorJs @Inject constructor(private val appSettingsPreferencesStore: SettingsDataStore) : GlobalPrivacyControlInjector {
+class GlobalPrivacyControlManager(private val appSettingsPreferencesStore: SettingsDataStore) : GlobalPrivacyControl {
     private val javaScriptInjector = JavaScriptInjector()
+    private val headerConsumers = listOf(
+        "nytimes.com",
+        "globalprivacycontrol.org",
+        "global-privacy-control.glitch.me"
+    )
+
+    override fun shouldAddHeaders(url: Uri): Boolean {
```
marcosholgado, line 43: Consider adding tests for this method. Also test for subdomains, i.e. `static.nytimes.com`.

marcosholgado (changes requested):
Nice job @cmonfortep, just left a comment to add some tests to the new logic. Apart from that everything works like expect including subdomains. 👍

app/src/main/java/com/duckduckgo/app/globalprivacycontrol/GlobalPrivacyControlManager.kt
marcosholgado: Also on this note. This check only happens in the `WebViewRequestInterceptor` but we are also adding the headers in the `BrowserTabViewModel`. Should this check be added in the ViewModel `getUrlHeaders()` method? (line 531)

marcosholgado: Just gonna leave this as it is for now :D

marcosholgado (approved):

Ryan451:
@cmonfortep since DDG supports GPC, I think issue #648 should be closed now

PR #1026, opened 2020-11-27
Implement role manager default browser dialog in welcome page

aitorvs:
Task/Issue URL: https://app.asana.com/0/72649045549333/1199338275364983/f
Tech Design URL: 
CC: 

**Description**:
Use the role manager dialog during onboarding welcome page to prompt the user to select us as default browser

* Added the `DefaultRoleBrowserDialogExperiment` that create the intent for the role manager and use it in welcome page

* Skip the "default browser page" regardless of user selection

**Steps to test this PR**:
Note: Use variant `zu`

==Test case 1==
1. Device API >=29
2. Fresh install the app and launch it
3. In Welcome page press "let's do it"
4. Select ddg as default browser in the dialog
5. Verify app proceeds to main browser page
6. Verify pixel `m_db_s` with `fo` param set as true
7. Verify browser is set as default

==Test case 2==
1. Device API >=29
2. Fresh install the app and launch it
3. In Welcome page press "let's do it"
4. Press "Cancel" in the dialog
5. Verify app proceeds to main browser page
6. Verify pixel `m_db_ns` with `fo` param set as true
7. Verify browser is NOT set as default

==Test case 3==
1. Device API <29
2. Fresh install the app and launch it
3. In Welcome page press "let's do it"
4. Verify onboarding proceeds unmodified

==Test case 4==
1. Device with any API
2. Upgrade the app and launch it (onboarding should have been done already)
3. Verify onboarding is skipped and user taken to main browser page

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/onboarding/ui/page/WelcomePageViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/onboarding/ui/page/WelcomePageViewModelTest.kt
@@ -1,0 +1,179 @@
+/*
+ * Copyright (c) 2020 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.onboarding.ui.page
+
+import android.content.Intent
+import androidx.arch.core.executor.testing.InstantTaskExecutorRule
+import androidx.test.platform.app.InstrumentationRegistry
+import com.duckduckgo.app.CoroutineTestRule
+import com.duckduckgo.app.global.DefaultRoleBrowserDialogExperiment
+import com.duckduckgo.app.global.install.AppInstallStore
+import com.duckduckgo.app.runBlocking
+import com.duckduckgo.app.statistics.pixels.Pixel
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.verify
+import com.nhaarman.mockitokotlin2.whenever
+import junit.framework.Assert.assertTrue
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.channels.ConflatedBroadcastChannel
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.launch
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.mockito.Mock
+import org.mockito.MockitoAnnotations
+
+@ExperimentalCoroutinesApi
+class WelcomePageViewModelTest {
+
+    @get:Rule
+    var instantTaskExecutorRule = InstantTaskExecutorRule()
+
+    @get:Rule
+    @Suppress("unused")
+    val coroutineRule = CoroutineTestRule()
+
+    @Mock
+    private lateinit var pixel: Pixel
+
+    @Mock
+    private lateinit var appInstallStore: AppInstallStore
+
+    @Mock
+    private lateinit var defaultRoleBrowserDialogExperiment: DefaultRoleBrowserDialogExperiment
+
+    private val events = ConflatedBroadcastChannel<WelcomePageView.Event>()
+
+    lateinit var viewModel: WelcomePageViewModel
+
+    private lateinit var viewEvents: Flow<WelcomePageView.State>
+
+    @Before
+    fun setup() {
+        MockitoAnnotations.initMocks(this)
+        viewModel = WelcomePageViewModel(
+            appInstallStore,
+            InstrumentationRegistry.getInstrumentation().targetContext,
+            pixel,
+            defaultRoleBrowserDialogExperiment
+        )
+
+        viewEvents = events.asFlow().flatMapLatest { viewModel.reduce(it) }
+    }
+
+    @Test
+    fun whenOnPrimaryCtaClickedAndControlXpThenFinish() = coroutineRule.runBlocking {
+        val launch = launch {
+            viewEvents.collect { state ->
+                assertTrue(state == WelcomePageView.State.Finish)
+            }
+        }
+        events.send(WelcomePageView.Event.OnPrimaryCtaClicked)
+
+        launch.cancel()
+    }
+
+    @Test
+    fun whenOnPrimaryCtaClickedAndTreatedXpAndNoShowThenFinish() = coroutineRule.runBlocking {
+        whenever(defaultRoleBrowserDialogExperiment.shouldShowExperiment())
+            .thenReturn(false)
+
+        val launch = launch {
+            viewEvents.collect { state ->
+                assertTrue(state == WelcomePageView.State.Finish)
+            }
+        }
+        events.send(WelcomePageView.Event.OnPrimaryCtaClicked)
+
+        launch.cancel()
+    }
+
+    @Test
+    fun whenOnPrimaryCtaClickedAndTreatedXpAndShowThenEmitShowDialog() = coroutineRule.runBlocking {
+        whenever(defaultRoleBrowserDialogExperiment.shouldShowExperiment())
+            .thenReturn(true)
+        val intent = Intent()
+        whenever(defaultRoleBrowserDialogExperiment.createIntent(any()))
+            .thenReturn(intent)
+
+        val launch = launch {
+            viewEvents.collect { state ->
+                assertTrue(state == WelcomePageView.State.ShowDefaultBrowserDialog(intent))
+            }
+        }
+        events.send(WelcomePageView.Event.OnPrimaryCtaClicked)
+
+        launch.cancel()
+    }
+
+    @Test
+    fun whenOnPrimaryCtaClickedAndTreatedXpNullIntentThenFireAndFinish() = coroutineRule.runBlocking {
+        whenever(defaultRoleBrowserDialogExperiment.shouldShowExperiment())
+            .thenReturn(true)
+
+        val launch = launch {
+            viewEvents.collect { state ->
+                assertTrue(state == WelcomePageView.State.Finish)
+            }
+        }
+        events.send(WelcomePageView.Event.OnPrimaryCtaClicked)
+
+        verify(pixel).fire(Pixel.PixelName.DEFAULT_BROWSER_DIALOG_NOT_SHOWN)
+
+        launch.cancel()
+    }
+
+    @Test
+    fun whenOnDefaultBrowserSetThenCallExperimentShownFireAndFinish() = coroutineRule.runBlocking {
+        val launch = launch {
+            viewEvents.collect { state ->
+                assertTrue(state == WelcomePageView.State.Finish)
+            }
+        }
+        events.send(WelcomePageView.Event.OnDefaultBrowserSet)
+
+        verify(defaultRoleBrowserDialogExperiment).experimentShown()
+        verify(pixel).fire(
+            Pixel.PixelName.DEFAULT_BROWSER_SET,
+            mapOf(Pixel.PixelParameter.DEFAULT_BROWSER_SET_FROM_ONBOARDING to true.toString())
+        )
+
+        launch.cancel()
+    }
+
+    @Test
+    fun whenOnDefaultBrowserNotSetThenCallExperimentShownFireAndFinish() = coroutineRule.runBlocking {
+        val launch = launch {
+            viewEvents.collect { state ->
+                assertTrue(state == WelcomePageView.State.Finish)
+            }
+        }
+        events.send(WelcomePageView.Event.OnDefaultBrowserNotSet)
+
+        verify(defaultRoleBrowserDialogExperiment).experimentShown()
+        verify(pixel).fire(
+            Pixel.PixelName.DEFAULT_BROWSER_NOT_SET,
+            mapOf(Pixel.PixelParameter.DEFAULT_BROWSER_SET_FROM_ONBOARDING to true.toString())
+        )
+
+        launch.cancel()
+    }
+}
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
```diff
@@ -57,6 +81,23 @@ class WelcomePage : OnboardingPageFragment() {
         typingAnimation?.cancel()
     }
 
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
```
marcosholgado, line 84: I know this is an experiment but I feel this page could have a view model so this logic could leave there and be tested ala `DefaultBrowserPageViewModel`. 

How about creating a clean up task in Asana to create that view model, test it, etc? If we end up deciding to release to everyone we can do that so, otherwise we delete the task.

marcosholgado (changes requested):
This is looking good! I've added some comments and I also found a few problems.

If I select our browser as default I get the pixel `m_db_s with params: {fo=true}` as expected but I also get another one `m_db_s with params: {fo=true}`. This double pixel will mess up the current pixel stats and invalidate them when the users set the browser as default out of the onboarding.

I think the above is because you forgot to set the default browser to true `installStore.defaultBrowser = true` so `DefaultBrowserObserver` is kicking in and sending another pixel. Same applies when the result is canceled we should be setting that value to false. Have a look at `fireDefaultBrowserPixelAndResetTimesPressedJustOnce()` for an example. 
We do something different there as well and rather than check the intent result we check if the browser has indeed been set as default I guess just to be sure the browser has actually been set as default.

app/src/main/java/com/duckduckgo/app/global/install/AppInstallStore.kt
```diff
@@ -61,5 +67,6 @@ class AppInstallSharedPreferences @Inject constructor(private val context: Conte
         const val KEY_TIMESTAMP_UTC = "INSTALL_TIMESTAMP_UTC"
         const val KEY_WIDGET_INSTALLED = "KEY_WIDGET_INSTALLED"
         const val KEY_DEFAULT_BROWSER = "KEY_DEFAULT_BROWSER"
+        private const val ROLE_MANAGER_DIALOG_KEY = "ROLE_MANAGER_DIALOG_KEY"
```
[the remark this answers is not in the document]
aitorvs: Donw

app/src/main/java/com/duckduckgo/app/global/DefaultRoleBrowserDialogExperiment.kt
```diff
@@ -0,0 +1,62 @@
...
+                    null
+                }
+            }
+        }
+        return null
+    }
+
+    fun shouldShowExperiment(): Boolean {
+        return variantManager.getVariant().hasFeature(VariantManager.VariantFeature.SetDefaultBrowserDialog) &&
+            appInstallStore.newDefaultBrowserDialogCount < 2
```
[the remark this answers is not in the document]
aitorvs: done

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
```diff
@@ -40,12 +51,25 @@ class WelcomePage : OnboardingPageFragment() {
...
         beginWelcomeAnimation(ctaText)
     }
 
+    private fun launchDefaultBrowserDialogOrContinue() {
+        if (defaultRoleBrowserDialogExperiment.shouldShowExperiment()) {
+            val intent = defaultRoleBrowserDialogExperiment.createIntent(requireContext())
+            if (intent != null) {
+                startActivityForResult(intent, DEFAULT_BROWSER_ROLE_MANAGER_DIALOG)
+            } else {
+                onContinuePressed()
```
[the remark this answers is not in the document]
aitorvs: done

app/src/main/java/com/duckduckgo/app/statistics/VariantManager.kt
```diff
@@ -52,12 +55,21 @@ interface VariantManager {
...
             Variant(
                 key = "zs",
-                weight = 1.0,
+                weight = 0.0,
                 features = listOf(VariantFeature.FireButtonEducation),
                 filterBy = { isEnglishLocale() }
+            ),
+
+            // Role manager default browser dialog
+            Variant(key = "zt", weight = 1.0, features = emptyList(), filterBy = { isAtLeastApiVersion(Q) }),
```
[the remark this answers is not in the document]
aitorvs: Added test

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
aitorvs: Done

aitorvs:
@marcosholgado I think I addressed all comments

cmonfortep (changes requested):
When testing the behavior in <29 api devices, I've found there original behavior has changed.

Develop:
If the user goes to settings and does not select DDG, we don't navigate to the browser.

This PR:
However, when testing this PR, user navigates to the browser. See gif: 

default_browser

app/src/main/java/com/duckduckgo/app/onboarding/di/WelcomePageModule.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.duckduckgo.app.global.RealDefaultRoleBrowserDialogExperiment
+import com.duckduckgo.app.global.install.AppInstallStore
+import com.duckduckgo.app.onboarding.ui.page.WelcomePageViewModelFactory
+import com.duckduckgo.app.statistics.VariantManager
+import com.duckduckgo.app.statistics.pixels.Pixel
+import dagger.Module
+import dagger.Provides
+
+@Module
+class WelcomePageModule {
```
cmonfortep, line 30: One question: I see this new approach and I want to know more. would you suggest to create `@Module` per Activity/Fragment? why that instead of per feature? Asking that because we have `OnboardingModule` and I think this could be maybe there?

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingActivity.kt
```diff
@@ -38,7 +42,9 @@ class OnboardingActivity : DuckDuckGoActivity() {
 
     fun onContinueClicked() {
         val next = viewPager.currentItem + 1
-        if (next < viewPager.adapter!!.count) {
+        if (next < viewPager.adapter!!.count &&
+            !variantManager.getVariant().hasFeature(VariantManager.VariantFeature.SetDefaultBrowserDialog)
```
cmonfortep, line 46: Correct me if I'm wrong, but I think this condition here assumes that between the First Onboarding screen and the browser Activity there's only 1 screen, and that screen can be skipped when feature `SetDefaultBrowserDialog` is true. 

I think we should avoid having this condition here. What about just not adding pages to the adapter based on that variant feature, and then we can just keep the logic as it was?

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPageFragment.kt
```diff
@@ -32,6 +34,11 @@ abstract class OnboardingPageFragment : Fragment() {
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? =
         inflater.inflate(layoutResource(), container, false)
 
+    override fun onAttach(context: Context) {
```
cmonfortep, line 37: do we need this here?

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
```diff
@@ -24,28 +26,79 @@ import android.view.WindowManager
...
+        primaryCta.setOnClickListener { event(WelcomePageView.Event.OnPrimaryCtaClicked) }
 
         configureDaxCta()
         beginWelcomeAnimation(ctaText)
     }
 
+    override fun onStart() {
+        super.onStart()
+
+        lifecycleScope.launch {
```
cmonfortep, line 73: This new pattern looks great, and it remembers me of old good RxJava days :D

In any case, I would recommend that before introducing new patterns we could discuss them internally. Could you do a brief summary of this approach at our next mobile meeting? Thanks!

aitorvs:
> When testing the behavior in <29 api devices, I've found there original behavior has changed.
> 
> If the user goes to settings and does not select DDG, we don't navigate to the browser. However, when testing this PR, user navigates to the browser. See gif:
> 
> default_browser

What is the config of the app. Did you force any variant?

cmonfortep:
> > When testing the behavior in <29 api devices, I've found there original behavior has changed.
> > If the user goes to settings and does not select DDG, we don't navigate to the browser. However, when testing this PR, user navigates to the browser. See gif:
> 
> What is the config of the app. Did you force any variant?

Yes, sorry. I was using emulator api 28 with `zu` variant. Haven't test the other variant yet, so I can't tell you if it's also happening in there.

app/src/main/java/com/duckduckgo/app/global/DefaultRoleBrowserDialogExperiment.kt
```diff
@@ -0,0 +1,73 @@
...
+                } else {
+                    Timber.i("Browser role held")
+                    null
+                }
+            }
+        }
+        return null
+    }
+
+    override fun shouldShowExperiment(): Boolean {
```
cmonfortep, line 59: question: is there any way to reproduce this in our app?

cmonfortep (commented):
Looks great, just a few questions from my side.

app/src/main/java/com/duckduckgo/app/onboarding/di/WelcomePageModule.kt
aitorvs: The setup we currently have has every DI instance in one single component (scope) which is the app scope. However, we are creating one component (scope) per activity/fragment that are completely empty.
This means that everything in DI is visible to every component, even dependencies that are very specific to say a activity/fragment, as they are placed in the app scope. which is kinda anti-pattern for DI.

I thought to take baby steps in this PR to a better DI organization. In this, `WelcomePageModule` has only the dependencies for the welcome Page (eg. `welcomePageViewModelFactory`. Nobody else should see those dependencies in DI or be able to get them. And `WelcomePageModule` is added to the `WelcomePage` scope.

Hope this clarifies

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPageFragment.kt
aitorvs: I can move it to the `WelcomePage` fragment.

app/src/main/java/com/duckduckgo/app/global/DefaultRoleBrowserDialogExperiment.kt
aitorvs: Not right now, because we are adding the new (role manager) dialog just to onboarding. So the user will only see it once. It will never happen that count is above `1`
I placed this as a safeguard in case we want to use it also in settings in the near future--which was discussed in the near future.

The comment in that method gives the background as to why that safeguard is there.

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingActivity.kt
aitorvs: Agree. I will move the logic into the `DefaultBrowserDetector`

aitorvs:
> > > When testing the behavior in <29 api devices, I've found there original behavior has changed.
> > > If the user goes to settings and does not select DDG, we don't navigate to the browser. However, when testing this PR, user navigates to the browser. See gif:
> > 
> > 
> > What is the config of the app. Did you force any variant?
> 
> Yes, sorry. I was using emulator api 28 ~with `zu` variant~, default variant (I was forcing zu but since I'm api 28, I was assigned to default variant). Haven't tested the other variant yet, so I can't tell you if it's also happening in there.

I found and solved the issue. I've also addressed all comments in https://github.com/duckduckgo/Android/pull/1026/commits/832c8703ab237b3e2425ad3bceb803bc16c09f1a. Let me know if something is amiss.

cmonfortep:
When testing the default variant, found a similar bug to the one reported in my previous review.

Bug:
- User selects DDG as default browser via settings
- goes back to the app
- user stays in the default browser onboarding screen

Expected:
- User selects DDG as default browser via settings
- goes back to the app
- we move the user to the Browser activity

settings_bug


Tes rest of scenarios work as described 👍

cmonfortep (commented):
Code looks great to me, thanks for the changes. We should check the bug reported when testing scenario 3.

Apart from that, something I would like to see is some conversations with the rest of the team about the unidirectional pattern we are using in this PR. I think we should share it so others have the chance of asking questions, and be aligned here.
There's also a comment in this PR about changes in our dagger setup. I really appreciate the will of improving our codebase, but I think we are missing some knowledge sharing here. I think everyone will benefit from your expertise in the DI field, so I will really encourage you to share what you have in mind and how can we improve.

I'm out next week, so I'm passing this PR back to @marcosholgado. Thanks and great work!

aitorvs:
> When testing the default variant, found a similar bug to the one reported in my previous review.
> 
> Bug:
> 
> * User selects DDG as default browser via settings
> * goes back to the app
> * user stays in the default browser onboarding screen
> 
> Expected:
> 
> * User selects DDG as default browser via settings
> * goes back to the app
> * we move the user to the Browser activity
> 
> Tes rest of scenarios work as described 👍

I tried to repro it and could not. See attached

screen-20201204-203956_2

Do you have any more info that might help reproducing this issue?

marcosholgado (approved):
Good job! I've added two minor comments related to a couple of tests, feel free to change them or not.

app/src/androidTest/java/com/duckduckgo/app/onboarding/ui/page/WelcomePageViewModelTest.kt
```diff
@@ -0,0 +1,179 @@
...
+                assertTrue(state == WelcomePageView.State.Finish)
+            }
+        }
+        events.send(WelcomePageView.Event.OnPrimaryCtaClicked)
+
+        launch.cancel()
+    }
+
+    @Test
+    fun whenOnPrimaryCtaClickedAndTreatedXpAndNoShowThenFinish() = coroutineRule.runBlocking {
```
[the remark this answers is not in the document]
aitorvs: Makes total sense, done

app/src/androidTest/java/com/duckduckgo/app/onboarding/ui/page/WelcomePageViewModelTest.kt
```diff
@@ -0,0 +1,179 @@
...
+                assertTrue(state == WelcomePageView.State.ShowDefaultBrowserDialog(intent))
+            }
+        }
+        events.send(WelcomePageView.Event.OnPrimaryCtaClicked)
+
+        launch.cancel()
+    }
+
+    @Test
+    fun whenOnPrimaryCtaClickedAndTreatedXpNullIntentThenFireAndFinish() = coroutineRule.runBlocking {
```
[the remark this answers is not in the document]
aitorvs: Makes total sense, done

PR #1037, opened 2020-12-14
Add LE root certificates to Glide

aitorvs:
Task/Issue URL: https://app.asana.com/0/0/1199598987755231/f
Tech Design URL: 
CC: 

**Description**:
We need to add the LE root certificates to Glide so that fetchinf favicons don't fail in old devices (API <= 25) when LE starts issuing certificates with its own self-signed cert chain

* created custom okHttp and added LE root certs to its SSL socket factory

* registered custom okHttp as `GlideUrl` extension to handle fetches

**Steps to test this PR**:
API <= 25

1. Install app from develop
2. Navigate to https://valid-isrgrootx1.letsencrypt.org/
3. Verify in logcat that glide throws `CertPathValidatorException` when trying to fetch site favicon

API <= 25

1. Install app from this branch
2. Verify `Registering OkHttp-based ModelLoader for GlideUrl` message appears in logcat
3. Navigate to https://valid-isrgrootx1.letsencrypt.org/
4. Verify in logcat that glide does not throw `CertPathValidatorException`
5. Verify in logcat that `TabDataRepository` updates the favicon

API > 25

1. Install app from this branch
2. Verify `Registering OkHttp-based ModelLoader for GlideUrl` message does NOT appear in logcat
3. Navigate to https://valid-isrgrootx1.letsencrypt.org/
4. Verify in logcat that glide does not throw `CertPathValidatorException`
5. Verify in logcat that `TabDataRepository` updates the favicon

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
@@ -16,8 +16,54 @@
 
 package com.duckduckgo.app.global.image
 
+import android.content.Context
+import android.os.Build
+import com.bumptech.glide.Glide
+import com.bumptech.glide.Registry
 import com.bumptech.glide.annotation.GlideModule
+import com.bumptech.glide.integration.okhttp3.OkHttpUrlLoader
+import com.bumptech.glide.load.model.GlideUrl
 import com.bumptech.glide.module.AppGlideModule
+import com.duckduckgo.app.browser.certificates.rootstore.IsrgRootX1
+import com.duckduckgo.app.browser.certificates.rootstore.IsrgRootX2
+import okhttp3.OkHttpClient
+import okhttp3.tls.HandshakeCertificates
+import timber.log.Timber
+import java.io.InputStream
+import java.security.cert.X509Certificate
 
 @GlideModule
-class GlobalGlideModule : AppGlideModule()
+class GlobalGlideModule : AppGlideModule() {
+    override fun registerComponents(context: Context, glide: Glide, registry: Registry) {
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.N_MR1) {
+            try {
+                Timber.d("Registering OkHttp-based ModelLoader for GlideUrl")
+
+                val isrgRootX1 = IsrgRootX1(context)
+                val isrgRootX2 = IsrgRootX2(context)
+
+                val handshakeCertificates = HandshakeCertificates.Builder()
+                    .addTrustedCertificate(isrgRootX1.certificate() as X509Certificate)
+                    .addTrustedCertificate(isrgRootX2.certificate() as X509Certificate)
+                    .addPlatformTrustedCertificates()
+                    .build()
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
```diff
@@ -16,8 +16,54 @@
...
+                val isrgRootX2 = IsrgRootX2(context)
+
+                val handshakeCertificates = HandshakeCertificates.Builder()
+                    .addTrustedCertificate(isrgRootX1.certificate() as X509Certificate)
+                    .addTrustedCertificate(isrgRootX2.certificate() as X509Certificate)
+                    .addPlatformTrustedCertificates()
+                    .build()
+
+                val okHttpClient = OkHttpClient.Builder()
+                    .sslSocketFactory(handshakeCertificates.sslSocketFactory())
```
CDRussell, line 52: `.sslSocketFactory` is marked as deprecated with a suggestion to use another function. 

Should we be using the suggested alternative? Or if we can't for some reason, can you suppress this warning to indicate we are knowingly using it?

CDRussell (approved):
Approving, but we should resolve the deprecated function call by using the suggested alternative or suppressing warning before merging

app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
aitorvs: I was intentional about using that deprecated function. I prefer not to manually pass myself the system `TrusManager` to avoid having an issue there. The only difference is that the deprecated function uses reflection, which it does not impose any performance issues for this one instance.

CDRussell: If it's intentional, let's make it _obviously_ intentional; whereas now in code it looks accidental. Suppressing that warning (and perhaps a comment explaining that decision) would help indicate it was intentional.

aitorvs: Makes sense, I will suppress the warning

yschimke: It's deprecated for a reason, you should be probably be passing in `handshakeCertificates.trustManager()`.

It will work on Android via StandardAndroidSocketAdapter, but would fail if you happened to have the GMS Play provider or conscrypt installed.

aitorvs: oh, nice. Missed the API in `handshakeCertificates` where we can get the `trusManager`. I thought I was gonna have to  go through the `TrustManagerFactory`, `init` with `null` keystore and all.

I will use your suggestion then.

Now that I have you here, what was the reason to deprecate it, just curious.

yschimke:
@aitorvs pinging here for a marker https://letsencrypt.org/2020/12/21/extending-android-compatibility.html

PR #1043, opened 2020-12-21
Aligned bookmarks autocomplete matching algorithm with iOS

aitorvs:
Task/Issue URL: https://app.asana.com/0/0/1199525460093862/f
Tech Design URL: https://app.asana.com/0/0/1199656369239103/f
CC: 

**Description**:
Align the autocomplete matching algorithm with iOS. This diff:

* removes the max limit of 6 auto complete items

* impose a max limit of 2 bookmark autocomplete items

* aligns with the iOS algorithm to select bookmark autocomplete items

**Steps to test this PR**:

Autocomplete results limit

1. install app from develop
2. clean all bookmarks
3. search for "cnn"
4. verify the autocomplete result items are limited to 6 items

1. install app from this branch
2. clean all bookmarks
3. search for "cnn"
4. verify the autocomplete result items are not limited to 6 suggestions

Bookmark autocomplete items limit

1. install app from develop
2. search for "cnn" and add more than 2 bookmarks from the results
3. restart the app and search for "cnn"
4. verify the autocomplete items contains all bookmarks added in step 2 (more than 2)
5. verify the autocomplete items are limited to 6 suggestions

1. install app from this branch
2. (only if not done before) search for "cnn" and add more than 2 bookmarks from the results
3. (only if not done beforee) restart the app and search for "cnn"
4. verify the autocomplete items contains max of 2 bookmark items
5. verify the autocomplete items are not limited to 6 suggestions

Item deduplication

1. install the app from develop
2. add https://edition.cnn.com to bookmarks
3. in main browser, search for edition.cnn.com
4. verify edition.cnn.com appears twice (from local bookmarks and autocomplete results from backend)

1. install the app from this branch
2. (only if not done before) add https://edition.cnn.com to bookmarks
3. in main browser, search for edition.cnn.com
4. verify edition.cnn.com appears just once

Bonus test

1. install the app from this branch
2. install develop from iOS repo in iOS
3. have both Android and iOS apps with same bookmarks
4. verify searches match for both devices

---
###### Internal references:

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApiTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApiTest.kt
@@ -40,7 +40,7 @@ class AutoCompleteApiTest {
 
     @Before
     fun before() {
-        MockitoAnnotations.initMocks(this)
+        MockitoAnnotations.openMocks(this)
         testee = AutoCompleteApi(mockAutoCompleteService, mockBookmarksDao)
     }
 
--- a/app/src/androidTest/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApiTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApiTest.kt
@@ -19,15 +19,12 @@ package com.duckduckgo.app.autocomplete.api
 import com.duckduckgo.app.autocomplete.api.AutoComplete.AutoCompleteResult
 import com.duckduckgo.app.bookmarks.db.BookmarkEntity
 import com.duckduckgo.app.bookmarks.db.BookmarksDao
-import com.nhaarman.mockitokotlin2.verify
 import com.nhaarman.mockitokotlin2.whenever
 import io.reactivex.Observable
 import io.reactivex.Single
-import org.junit.Assert.assertSame
-import org.junit.Assert.assertTrue
+import org.junit.Assert.*
 import org.junit.Before
 import org.junit.Test
-import org.mockito.ArgumentMatchers.anyString
 import org.mockito.Mock
 import org.mockito.MockitoAnnotations
 
@@ -48,16 +45,6 @@ class AutoCompleteApiTest {
     }
 
     @Test
-    fun whenQueryIsNotBlankThenTryToGetBookmarksFromDAO() {
-        whenever(mockAutoCompleteService.autoComplete("foo")).thenReturn(Observable.just(emptyList()))
-        whenever(mockBookmarksDao.bookmarksByQuery(anyString())).thenReturn(Single.just(emptyList()))
-
-        testee.autoComplete("foo")
-
-        verify(mockBookmarksDao).bookmarksByQuery("%foo%")
-    }
-
-    @Test
     fun whenQueryIsBlankThenReturnAnEmptyList() {
         val result = testee.autoComplete("").test()
         val value = result.values()[0] as AutoCompleteResult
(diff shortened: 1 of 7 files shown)


subsymbolic:
Autocomplete is used by the interstitial SystemSearch that the widget and google search bar launch. Just an FYI to test that too to make sure everything still works as expected.

app/src/androidTest/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApiTest.kt
```diff
@@ -48,31 +45,94 @@ class AutoCompleteApiTest {
...
-        assertSame("https://example.com", value.suggestions[0].phrase)
+        assertEquals(
+            listOf(
+                AutoComplete.AutoCompleteSuggestion.AutoCompleteBookmarkSuggestion(phrase = "example.com", "title example", "https://example.com"),
+                AutoComplete.AutoCompleteSuggestion.AutoCompleteBookmarkSuggestion(phrase = "foo.com", "title foo", "https://foo.com"),
+                AutoComplete.AutoCompleteSuggestion.AutoCompleteSearchSuggestion(phrase = "bar.com", true),
+                AutoComplete.AutoCompleteSuggestion.AutoCompleteSearchSuggestion(phrase = "baz.com", true)
+            ),
+            value.suggestions
+        )
```
marcosholgado, line 136: Consider testing the ranking logic in `AutoCompleteApi`. It would be nice to have tests to know how the different scores affect the order in the bookmark results.

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
```diff
@@ -58,7 +61,7 @@ class AutoCompleteApi @Inject constructor(
             BiFunction { bookmarksResults, searchResults ->
                 AutoCompleteResult(
                     query = query,
-                    suggestions = (bookmarksResults + searchResults).distinct()
+                    suggestions = (bookmarksResults + searchResults).distinctBy { it.phrase }
```
marcosholgado, line 64: I'm not 100% sure about this but could be correct. Right now if I have two bookmarks stored with the `editorial.cnn.com` domain and I search by the domain using `editorial` or `cnn` I only get one result back because the domain is used as the phrase and we filter them out here.

I think that both results should be displayed here.

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
```diff
@@ -75,12 +78,69 @@ class AutoCompleteApi @Inject constructor(
             .toObservable()
 
     private fun getAutoCompleteBookmarkResults(query: String) =
-        bookmarksDao.bookmarksByQuery("%$query%")
+        bookmarksDao.bookmarksObservable()
+            .map { rankBookmarks(query, it) }
             .flattenAsObservable { it }
             .map {
-                AutoCompleteBookmarkSuggestion(phrase = it.url, title = it.title ?: "", url = it.url)
+                AutoCompleteBookmarkSuggestion(phrase = it.url.toUri().baseHost.orEmpty(), title = it.title.orEmpty(), url = it.url)
```
marcosholgado, line 85: Using the domain as the phrase causes a bug when you try to edit a bookmark result.

Once you see a bookmark result, if you click on the arrow icon the expected behaviour is for the URL to be added in the search bar so users can tweak it. The logic we have right now uses the phrase. We should change this behaviour so it adds the URL rather than the phrase when the result is a bookmark.

In the `BrowserTabFragment` line 881 when we set the `editableSearchClickListener` we should now differentiate between search results and bookmarks and pass the phrase or URL depending on that.

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
```diff
@@ -75,12 +78,69 @@ class AutoCompleteApi @Inject constructor(
...
+        } else if (domain?.startsWith(tokens.first(), ignoreCase = true) == true) {
+            rankedBookmark.score += 300
+        }
+
+        return rankedBookmark
+    }
+
+    private data class RankedBookmark(
+        val bookmarkEntity: BookmarkEntity,
+        var score: Int = -1
```
marcosholgado, line 144: nit: I know this is -1 because that's the iOS implementation and they have `0` for favourites. Maybe moving it to a const like `BOOKMARK_SCORE`? Your call, I'm not too fuzz about it.

app/src/androidTest/java/com/duckduckgo/app/autocomplete/api/AutoCompleteApiTest.kt
```diff
@@ -48,31 +45,94 @@ class AutoCompleteApiTest {
     }
```
marcosholgado, line 46: nit: Since we are changing this file, would you mind changing `initMocks` which is now deprecated to `openMocks' ?

marcosholgado (changes requested):
Great job @aitorvs! This looks great, I added some comments but the main thing is around editing a bookmark query.

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
aitorvs: So the main problem here is that sealed classes are not value types. And so the distinct does nothing because it is an equality on the instance reference.
I could try to use `data class`  instead

marcosholgado: Just thinking out loud to summarise what I think we are trying to do:

1. Remove from `searchResults` any items that have a `phrase` which exists in `bookmarkResults`
1. Remove from `bookmarkResults` any items with the same URL.
1. Return 1 + 2

Since everyone is on holiday I'm gonna ask on Asana if that's the intended behaviour or not.

aitorvs: Almost. What we want is to dedup the results (1+2).dedup.
I also guess that was the intention of the original `distinc`. However that was not de duplicating anything as it was only comparing instance references

aitorvs:
@marcosholgado I think I have addressed all comments, lmk

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
```diff
@@ -75,12 +79,73 @@ class AutoCompleteApi @Inject constructor(
             .toObservable()
 
     private fun getAutoCompleteBookmarkResults(query: String) =
-        bookmarksDao.bookmarksByQuery("%$query%")
+        bookmarksDao.bookmarksObservable()
+            .map { rankBookmarks(query, it) }
             .flattenAsObservable { it }
             .map {
-                AutoCompleteBookmarkSuggestion(phrase = it.url, title = it.title ?: "", url = it.url)
+                AutoCompleteBookmarkSuggestion(phrase = it.url.toUri().resourceNameAndPath(), title = it.title.orEmpty(), url = it.url)
```
marcosholgado, line 86: Just a small issue with this. If you try to save a bookmark for a URL with query parameters they won't be used when editing the query. For example if you go to duckduckgo.com where by default we add some query parameters and save that as bookmark, those query parameters are not re-added when trying to edit the query.

marcosholgado (changes requested):
Everything is looking good but there is still a small issue when trying to edit certain saved bookmarks if they have query parameters.

marcosholgado (approved):
Looks great! Nice job @aitorvs

marcosholgado (approved):
Nice job! 👍

PR #1065, opened 2021-01-15
Fix query change pixels

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1199568299173487/f
Tech Design URL: https://app.asana.com/0/481882893211075/1199538898514883/f
CC: 

**Description**:

This PR fixes the `rq_x` pixels that got broke when removing the double search header. The main issue was that pixels have a suffix `android_{formFactor}`, but also under the [iOS implementation](https://app.asana.com/0/481882893211075/1199403269515656/f) a few more fine tuning was done to improve them. This PR will:

* Remove the android_tablet/phone suffix from the rq_x pixels
* Override the user agent so that rq_x pixels are sent with the same UA as the webview
* Align the search bar behavior with iOS
* rq.0 will fire when
  * the user resubmits the search through the search box in our app.
  * the SERP is reloaded 
  * hitting refresh button
  * resubmitting query without editing
  * selecting autocomplete item that matches what was just searched
* rq.1 will fire when
  * the user has the SERP loaded in the current tab and submits a different search (by typing or selecting different autocomplete entry).

**Steps to test this PR**:
Note: to check `rq_x` are being sent, just filter for `Pixel interceptor` in logcat

The following should test all scenarios.

* open app and search reddit
* Verify `rq.x` pixels are not sent
* pull to refresh
* verify `rq.0` pixel is sent
* overflow menu -> refresh
* verify `rq.0` is sent
* tap on search bar 
* verify "reddit" word is not selected, cursor is at the end of the word and autocomplete appears
* select "🔎 reddit" from autocomplete
* verify `rq.0` is sent
* tap on search bar and select "🔎 reddit politics" (or any other reddit xxx autocomplete result)
* Verify `rq.1` pixel is sent
* click on a link in SERP, open it in a new tab and wait for it to load
* tap on the search bar
* verify autocomplete results don't appear
* verify URL in search bar is selected - ❌ this check will fail, I didn't manage to make it work
* close the app and re-open
* Verify `rq.x` pixels are not sent
* switch to the tab with the query search, ie "reddit..."
* verify `rq.1` pixel is sent
* (here use Charles to inspect UA)
* change the query in the search bar and ENTER
* verify the UA is the one used for webview when `rq.1` is sent
* pull to refresh
* verify the UA is the one used for webview when `rq.1` is sent

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
@@ -18,7 +18,9 @@ package com.duckduckgo.app.browser.omnibar
 
 import android.content.Context
 import android.graphics.Rect
+import android.text.Editable
 import android.util.AttributeSet
+import android.util.Patterns
 import android.view.KeyEvent
 import androidx.appcompat.widget.AppCompatEditText
 import com.duckduckgo.app.global.view.showKeyboard
@@ -33,10 +35,25 @@ class KeyboardAwareEditText : AppCompatEditText {
     constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
     constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
 
+    private var didFocusAlready = false
+
+    private fun Editable.isWebUrl(): Boolean {
+        return Patterns.WEB_URL.matcher(this.toString()).matches()
+    }
+
     override fun onFocusChanged(focused: Boolean, direction: Int, previouslyFocusedRect: Rect?) {
         super.onFocusChanged(focused, direction, previouslyFocusedRect)
+        // stop the auto-selection of text after the first focus
+        setSelectAllOnFocus(!didFocusAlready)
         if (focused) {
+            if (didFocusAlready && text != null && text?.isWebUrl() == false) {
+                // trigger the text change listener so that we can show autocomplete
+                text = text
+                // cursor at the end of the word
+                setSelection(text!!.length)
+            }
             showKeyboard()
+            didFocusAlready = true
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
@@ -43,7 +43,6 @@ class KeyboardAwareEditText : AppCompatEditText {
 
     override fun onFocusChanged(focused: Boolean, direction: Int, previouslyFocusedRect: Rect?) {
         super.onFocusChanged(focused, direction, previouslyFocusedRect)
-        // stop the auto-selection of text after the first focus
         setSelectAllOnFocus(!didFocusAlready)
         if (focused) {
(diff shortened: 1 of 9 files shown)


aitorvs:
@malmstein help wanted for the test step that I marked as ❌. I think it'd be no biggie to not have it, but if we did, better

malmstein:
@aitorvs:

verify URL in search bar is selected - ❌ this check will fail, I didn't manage to make it work
close the app and re-open
verify rq.1 pixel is sent

This step does not work for me either. In fact, the pixel is not sent after closing / opening the app. I'd argue that it should not be fired since the query hasn't changed.

app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
```diff
@@ -33,10 +35,25 @@ class KeyboardAwareEditText : AppCompatEditText {
...
 
+    private var didFocusAlready = false
+
+    private fun Editable.isWebUrl(): Boolean {
+        return Patterns.WEB_URL.matcher(this.toString()).matches()
+    }
+
     override fun onFocusChanged(focused: Boolean, direction: Int, previouslyFocusedRect: Rect?) {
         super.onFocusChanged(focused, direction, previouslyFocusedRect)
+        // stop the auto-selection of text after the first focus
```
malmstein, line 46: we don't usually leave comments unless they are really necessary. seems like the code is pretty self-explanatory, so I'd say it's safe to remove them

malmstein (commented):
Question about the steps to test, other than that looks good. Great job @aitorvs !

app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
aitorvs: Makes sense, will do

aitorvs:
> @aitorvs:
> 
> verify URL in search bar is selected - ❌ this check will fail, I didn't manage to make it work
> close the app and re-open
> verify rq.1 pixel is sent
> 
> This step does not work for me either. In fact, the pixel is not sent after closing / opening the app. I'd argue that it should not be fired since the query hasn't changed.

Did you close the app and re-open or just put it in background and re-open. The former would send the rq.1 pixel just because the app starts with with an empty `oldQuery` and the `newQuery` is whatever in the search bar of the tab fragment.
If you place the app in the background and then re-open, nothing will be sent.

malmstein:
@aitorvs it should not matter if the app was put in background or closed, for pixel purposes it makes no difference and the rq.1 should not be sent. I don't want to delay this much longer so there are two solutions:
1. Fix the behavior on app start
2. Explain this behavior in the task and get feedback from interesting parties. Based on that we can close this or fix it.

aitorvs:
> @aitorvs it should not matter if the app was put in background or closed, for pixel purposes it makes no difference and the rq.1 should not be sent. I don't want to delay this much longer so there are two solutions:
> 
> 1. Fix the behavior on app start
> 2. Explain this behavior in the task and get feedback from interesting parties. Based on that we can close this or fix it.

Done in https://github.com/duckduckgo/Android/pull/1065/commits/c935b095bc2b0e856e693e7adc2a1338e8ced46c

malmstein (approved):
Great stuff @aitorvs. Thanks for making the changes!

PR #1066, opened 2021-01-18
Keep HTTP auth credentials when site is fireproof(ed)

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1199538898514852/f
Tech Design URL: https://app.asana.com/0/0/1199538898514868/f
CC: 

**Description**:
This task fixes the issue that makes the HTTP auth credentials being removed upon firing even when the site is fireproofed.

As the `httpauth` table db where the credentials are stored is never read by Android, we can manually read/write/delete at will without interfering with Android.
We wrap the DB with Room just as a convenience.

**Steps to test this PR**:
Verify fireproofing works with migration
* install the app from develop branch
* visit https://httpbin.org/basic-auth/test/1111
* enter user = test, password = 1111
* fire
* visit https://httpbin.org/basic-auth/test/1111 
* verify the Auth dialog is shown again (credentials have been deleted)
* enter user = test, password = 1111
* upgrade the app from this branch
* launcher app and visit https://httpbin.org/basic-auth/test/1111 
* fire
* visit https://httpbin.org/basic-auth/test/1111 
* verify the Auth dialog is shown again (credentials have been deleted)
* enter user = test, password = 1111
* fireproof the site
* fire
* visit https://httpbin.org/basic-auth/test/1111 
* verify auth dialog is no longer shown and user is already authenticated

Verify fireproofing works with clean build
* Clean install the app from this branch
* visit https://httpbin.org/basic-auth/test/1111 
* enter user = test, password = 1111
* fire
* visit https://httpbin.org/basic-auth/test/1111 
* verify the Auth dialog is shown again (credentials have been deleted)
* enter user = test, password = 1111
* fireproof the site
* fire
* visit https://httpbin.org/basic-auth/test/1111 
* verify auth dialog is no longer shown and user is already authenticated

Verify credentials can be updated
* install the app from this branch
* visit https://httpbin.org/basic-auth/test/1111 
* enter user = test, password = 1111
* close and re-open the app
* visit https://httpbin.org/basic-auth/test/1111 
* verify user is logged in
* visit https://httpbin.org/basic-auth/test/2222
* verify auth dialog appears
* enter user = test, password = 2222
* verify user is logged in
* visit https://httpbin.org/basic-auth/test/1111 
* verify auth dialog appears

---
###### Internal references:

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -148,7 +148,6 @@ dependencies {
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayout"
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:$swipeRefreshLayout"
     implementation "androidx.webkit:webkit:$webkit"
-    implementation "androidx.security:security-crypto:1.1.0-alpha03"
     implementation "com.squareup.okhttp3:okhttp:$okHttp"
     implementation "com.squareup.okhttp3:okhttp-tls:$okHttp"
     implementation "com.squareup.retrofit2:retrofit:$retrofit"
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -148,6 +148,7 @@ dependencies {
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayout"
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:$swipeRefreshLayout"
     implementation "androidx.webkit:webkit:$webkit"
+    implementation "androidx.security:security-crypto:1.1.0-alpha03"
     implementation "com.squareup.okhttp3:okhttp:$okHttp"
     implementation "com.squareup.okhttp3:okhttp-tls:$okHttp"
     implementation "com.squareup.retrofit2:retrofit:$retrofit"
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStoreTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStoreTest.kt
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2021 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.httpauth
+
+import android.os.Build
+import android.webkit.WebView
+import androidx.test.filters.SdkSuppress
+import com.duckduckgo.app.CoroutineTestRule
+import com.duckduckgo.app.browser.httpauth.db.HttpAuthDao
+import com.duckduckgo.app.browser.httpauth.db.HttpAuthEntity
+import com.duckduckgo.app.fire.fireproofwebsite.data.FireproofWebsiteDao
+import com.duckduckgo.app.fire.fireproofwebsite.data.FireproofWebsiteEntity
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.verify
+import com.nhaarman.mockitokotlin2.whenever
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+
+@ExperimentalCoroutinesApi
+class WebViewHttpAuthStoreTest {
+
+    @get:Rule
+    @Suppress("unused")
+    val coroutineRule = CoroutineTestRule()
+
+    private val fireproofWebsiteDao: FireproofWebsiteDao = mock()
+    private val httpAuthDao: HttpAuthDao = mock()
+    private val webView: WebView = mock()
+
+    private val webViewHttpAuthStore = RealWebViewHttpAuthStore(coroutineRule.testDispatcherProvider, fireproofWebsiteDao, httpAuthDao)
+    private val webViewHttpAuthStoreWithNullHttpAuthDao = RealWebViewHttpAuthStore(coroutineRule.testDispatcherProvider, fireproofWebsiteDao, null)
+
+    @Test
+    fun whenSetHttpAuthUsernamePasswordThenInsertHttpAuthEntity() {
+        webViewHttpAuthStore.setHttpAuthUsernamePassword(
+            webView = webView,
+            host = "host",
+            realm = "realm",
+            username = "name",
+            password = "pass",
+        )
+
+        verify(httpAuthDao).insert(
+            HttpAuthEntity(
+                host = "host",
+                realm = "realm",
+                username = "name",
+                password = "pass"
+            )
+        )
+    }
+
+    @Test
+    @SdkSuppress(maxSdkVersion = Build.VERSION_CODES.N_MR1)
+    fun whenSetHttpAuthUsernamePasswordAndNullHttpAuthDaoThenCallWebViewSetHttpAuth() {
+        webViewHttpAuthStoreWithNullHttpAuthDao.setHttpAuthUsernamePassword(
+            webView = webView,
+            host = "host",
+            realm = "realm",
+            username = "name",
+            password = "pass",
+        )
+
+        verify(webView).setHttpAuthUsernamePassword("host", "realm", "name", "pass",)
+    }
+
+    @Test
+    fun whenGetHttpAuthUsernamePasswordThenReturnWebViewHttpAuthCredentials() {
+        whenever(httpAuthDao.getAuthCredentials("host", "realm"))
+            .thenReturn(HttpAuthEntity(1, "host", "realm", "name", "pass"))
+        val credentials = webViewHttpAuthStore.getHttpAuthUsernamePassword(webView, "host", "realm")
+
+        assertEquals(WebViewHttpAuthCredentials("name", "pass"), credentials)
+    }
+
+    @Test
+    @SdkSuppress(maxSdkVersion = Build.VERSION_CODES.N_MR1)
+    fun whenGetHttpAuthUsernamePasswordAndNullHttpAuthDaoReturnWebViewHttpAuthCredentials() {
+        whenever(webView.getHttpAuthUsernamePassword("host", "realm"))
+            .thenReturn(arrayOf("name", "pass"))
+        val credentials = webViewHttpAuthStoreWithNullHttpAuthDao.getHttpAuthUsernamePassword(webView, "host", "realm")
+
+        assertEquals(WebViewHttpAuthCredentials("name", "pass"), credentials)
+    }
+
+    @Test
+    fun whenClearHttpAuthUsernamePasswordThenDeleteAllCredentialsExceptExclusions() {
+        whenever(fireproofWebsiteDao.fireproofWebsitesSync())
+            .thenReturn(listOf(FireproofWebsiteEntity("http://fireproofed.me")))
+
+        webViewHttpAuthStore.clearHttpAuthUsernamePassword(webView)
+
+        verify(httpAuthDao).deleteAll(listOf("http://fireproofed.me"))
+    }
+
+    @Test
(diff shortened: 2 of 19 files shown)


app/src/main/java/com/duckduckgo/app/browser/httpauth/db/HttpAuthEntity.kt
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.httpauth.db
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+// CREATE TABLE httpauth (_id INTEGER PRIMARY KEY, host TEXT, realm TEXT, username TEXT, password TEXT, UNIQUE (host, realm) ON CONFLICT REPLACE);
```
cmonfortep, line 22: Question: how probable is it that WebView changes the schema of this database? is this schema defined in their codebase? if so, could we check if they have changed any time in the past?

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -215,15 +217,17 @@ class BrowserWebViewClient(
...
                 Timber.v("onReceivedHttpAuthRequest - useHttpAuthUsernamePassword [${handler.useHttpAuthUsernamePassword()}]")
                 if (handler.useHttpAuthUsernamePassword()) {
-                    val credentials = buildAuthenticationCredentials(host.orEmpty(), realm.orEmpty(), view)
+                    val credentials = view?.let {
+                        webViewHttpAuthStore.getHttpAuthUsernamePassword(it, host.orEmpty(), realm.orEmpty())
+                    }
 
                     if (credentials != null) {
-                        handler.proceed(credentials[0], credentials[1])
+                        handler.proceed(credentials.username, credentials.password)
```
cmonfortep, line 225: 👍 I like this change.

app/src/main/java/com/duckduckgo/app/browser/httpauth/db/WebViewStoreExtensions.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.os.Build
+import android.webkit.WebView
+import android.webkit.WebViewDatabase
+
+fun WebView.clearFormDataCompat() {
+    clearFormData()
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
+        val webViewDatabase = WebViewDatabase.getInstance(this.context)
+        webViewDatabase.clearFormData()
```
cmonfortep, line 27: NIT: could we suppress this warning since we already have the API version guard?

app/src/main/java/com/duckduckgo/app/browser/WebDataManager.kt
```diff
@@ -83,16 +83,10 @@ class WebViewDataManager @Inject constructor(
...
-     */
-    @Suppress("DEPRECATION")
-    private fun clearFormData(webViewDatabase: WebViewDatabase) {
-        webViewDatabase.clearFormData()
-    }
-
-    private fun clearAuthentication(webViewDatabase: WebViewDatabase) {
-        webViewDatabase.clearHttpAuthUsernamePassword()
+    private fun clearAuthentication(webView: WebView) {
+        GlobalScope.launch(Dispatchers.IO) {
```
cmonfortep, line 87: why is this triggered as a new global coroutine? That will return without ensuring it was completed.

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
```diff
@@ -0,0 +1,128 @@
...
+    @UiThread
+    fun setHttpAuthUsernamePassword(webView: WebView, host: String, realm: String, username: String, password: String)
+    @UiThread
+    fun getHttpAuthUsernamePassword(webView: WebView, host: String, realm: String): WebViewHttpAuthCredentials?
+    @UiThread
+    fun clearHttpAuthUsernamePassword(webView: WebView)
+}
+
+class RealWebViewHttpAuthStore(
+    private val dispatcherProvider: DispatcherProvider,
```
cmonfortep, line 81: This is currently not used, I guess we want to use it instead of `Dispatchers.IO`?

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
```diff
@@ -0,0 +1,128 @@
...
+    private val fireproofWebsiteDao: FireproofWebsiteDao,
+    private val httpAuthDao: HttpAuthDao? = null
```
cmonfortep, line 83: is this default param required? I like that we have the logic inside the `DatabaseModule` to decide why it's `null`, and I would expect the same from any place that creates an instance of this class.

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
```diff
@@ -0,0 +1,128 @@
...
+import com.duckduckgo.app.global.DispatcherProvider
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+
+data class WebViewHttpAuthCredentials(val username: String, val password: String)
+
+@VisibleForTesting
+fun WebView.setHttpAuthUsernamePasswordCompat(host: String, realm: String, username: String, password: String) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
cmonfortep, line 37: This is the same logic we have in `DatabaseModule`. I wonder if having it duplicated here without a clear connection between them can lead to possible bugs if something changes. is there any way to unify them or to connect them somehow? 

Checking this twice, it's like we have 2 approaches for the same. 
a) compat method that handles Android api version
b) we call this method after checking that `httpAuthDao` was null, and `httpAuthDao` is `null` based on the device Android Api version.

app/src/androidTest/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStoreTest.kt
```diff
@@ -0,0 +1,119 @@
...
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.verify
+import com.nhaarman.mockitokotlin2.whenever
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+
+@ExperimentalCoroutinesApi
+class WebViewHttpAuthStoreTest {
```
cmonfortep, line 36: There are a few warnings within this class that we may want to suppress.

app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
```diff
@@ -29,6 +36,23 @@ import javax.inject.Singleton
...
+    ): WebViewHttpAuthStore {
+        val httpAuthDb = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            // This database is a Room wrapper around the http_auth convenience database
+            Room.databaseBuilder(context, HttpAuthDatabase::class.java, "http_auth.db")
+                .setJournalMode(RoomDatabase.JournalMode.TRUNCATE)
+                .addMigrations(HttpAuthDatabase.MIGRATION_1_2)
+                .build()
+        } else null
+
+        return RealWebViewHttpAuthStore(DefaultDispatcherProvider(), fireproofWebsiteDao, httpAuthDb?.httpAuthDao())
```
cmonfortep, line 53: Should we create this instance using an injected Dispatcher provider?

app/build.gradle
```diff
@@ -148,6 +148,7 @@ dependencies {
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayout"
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:$swipeRefreshLayout"
     implementation "androidx.webkit:webkit:$webkit"
+    implementation "androidx.security:security-crypto:1.1.0-alpha03"
```
cmonfortep, line 151: question: where are we using this dependency?

cmonfortep (commented):
Great work! I really liked how you approached the problem and divided the different responsibilities. A few minor comments from my side. 

I've tested this on different devices and everything worked as expected.

app/build.gradle
aitorvs: we don't use it, my bad. It is a left over of something else I was doing. Removing it

app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
aitorvs: It's s good question, I saw that all usages of `DefaultDispatcherProvider` are not through DI but rather building the type directly. That's why I didn't want to provide it, to avoid confusion.
As it is right now, it is still an injected dependency, just it is not available in the DI graph.
I think it is worth refactoring all places where we call `DefaultDispatcherProvider()` and get that thru DI but not sure that place is this PR.

app/src/androidTest/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStoreTest.kt
aitorvs: Yep, will do

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
aitorvs: This was on purpose. 
The webview extensions are disconnected from the DAO and viceversa. So we should not remove API checking from one of them assuming it will only be called from the other one and that other one will perform the checking.
If we remove the API checking from one of them and rely on the other one, we'd be coupling this two classes, something that I think is generally undesirable.

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
aitorvs: I can remove the `=null` yep

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
aitorvs: Yes, good catch. That was the intention, will do

app/src/main/java/com/duckduckgo/app/browser/WebDataManager.kt
aitorvs: Yes, I will follow the same approach I did for other methods and remove the `GlobalScope` here

app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
cmonfortep: Agree, I don't think we need to do that refactor within this PR. 

However, we already have `CoroutinesModule` that will provide that `DispatcherProvider` for you. And we have some Modules using it already (`UncaughtExceptionModule`, `BrowserModule`,...) So I would say we can use it in the same way.

aitorvs: oh nice, I missed the fact that we had it already in the `CoroutinesModule`. Will do

app/src/main/java/com/duckduckgo/app/browser/httpauth/db/HttpAuthEntity.kt
aitorvs: The documentation says webview never reads from this db.
The [source code](https://github.com/chromium/chromium/blame/master/android_webview/java/src/org/chromium/android_webview/HttpAuthDatabase.java) for the creation of the DB has been untouched for 8 years.

I think it's unlikely it'll change however we could always just migrate the entire DB and use our own to ensure the system does not touch it.

aitorvs:
@cmonfortep thanks for the review. I have addressed all comments and and replied to [this one](https://github.com/duckduckgo/Android/pull/1066#discussion_r559648506)

aitorvs:
@cmonfortep `whenAddLoginDetectionThenJSInterfaceAdded` is failing for one of the bitrise devices. I was introduced in a previous PR. It is also failing locally, any clues? Maybe a flaky test?

cmonfortep (approved):
Thanks @aitorvs, great work! I see we managed to make this compatible for all api levels. Could we update the description? 

Apart from that, I've tested this on different devices and everything works great. :shipit:

Issue #1067, opened 2021-01-18
Duplicate bookmarks


ru2saig:
**Describe the bug**
By "Add Bookmark"ing a website multiple times, I'm able to get multiple bookmarks of the exact same webpage. 
See screenshot below:
![Screenshot_20210118-222631_DuckDuckGo](https://user-images.githubusercontent.com/42857541/104944170-d4097980-59dc-11eb-84b0-47aae4bf42bb.jpg)

**How to Reproduce**
  .0. Go to any website
  .1. Tap on Add Bookmark in the kebab menu a couple of times [ Optional ] Tap a couple more times for good measure
  .2. Going to Bookmarks, there are multiple bookmarks of the same page.

**Expected behavior**
Only one bookmark per a unique webpage.

**System and App Information**
 - DDG App settings: Settings -> 5.74.0 (57400)
 - Device: Samsung Galaxy S7 Edge SM-G935T
 - OS: Android 8.0.0 and Samsung Experience Version 9.0

cmonfortep:
Thanks for reporting the issue. I agree it's something we should try to avoid. Thanks.

PR #1068, opened 2021-01-18
fix for duplicate bookmarks #1067

bluetoothfx:
Task/Issue URL:  https://github.com/duckduckgo/Android/issues/1067

**Description**: In this scenario if user tap on Add bookmark menu multiple times, then it creates multiple entry to database. 

**Solution approach**: In this case we have taken action to either update if exist or insert bookmark item if this doesn't.

**Impact**: There is several impact for this solution. If user taps multiple times then it will always show **Bookmark added**.
Technically this is updating into database. In other browser they have handled this kind of different approach.  

**Steps to test this PR**:
Steps to test if the behavior persist or not:
1. Go to any website
2. Tap on Add Bookmark in the kebab menu a couple of times
3. Going to Bookmarks, there are multiple bookmarks of the same page.

### DIFF
--- a/app/schemas/com.duckduckgo.app.global.db.AppDatabase/30.json
+++ b/app/schemas/com.duckduckgo.app.global.db.AppDatabase/30.json
@@ -1,0 +1,846 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 30,
+    "identityHash": "65b30523736804a4973282ffbe3ef731",
+    "entities": [
+      {
+        "tableName": "tds_tracker",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`domain` TEXT NOT NULL, `defaultAction` TEXT NOT NULL, `ownerName` TEXT NOT NULL, `categories` TEXT NOT NULL, `rules` TEXT NOT NULL, PRIMARY KEY(`domain`))",
+        "fields": [
+          {
+            "fieldPath": "domain",
+            "columnName": "domain",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "defaultAction",
+            "columnName": "defaultAction",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "ownerName",
+            "columnName": "ownerName",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "categories",
+            "columnName": "categories",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "rules",
+            "columnName": "rules",
+            "affinity": "TEXT",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "columnNames": [
+            "domain"
+          ],
+          "autoGenerate": false
+        },
+        "indices": [],
+        "foreignKeys": []
+      },
+      {
+        "tableName": "tds_entity",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`name` TEXT NOT NULL, `displayName` TEXT NOT NULL, `prevalence` REAL NOT NULL, PRIMARY KEY(`name`))",
+        "fields": [
+          {
+            "fieldPath": "name",
+            "columnName": "name",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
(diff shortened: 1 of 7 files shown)


aitorvs:
@bluetoothfx thanks a lot for opening the PR. This is a great bug to solve for our users.

I have a comment about the approach. Although the PR solves the problem for the upcoming bookmark additions, it does not for the users that already have duplicates (all previous versions of the app). And ideally we'd also want to dedup those.
I think the core issue is that the bookmarks DB was probably not properly defined at creation and thus not handling duplicates properly. 
We could solve all cases doing a DB migration (they are all placed under `AppDatabase.kt`). I think something like this should work
```kotlin
val MIGRATION_29_TO_30: Migration = object : Migration(29, 30) {
    override fun migrate(database: SupportSQLiteDatabase) {
        // create temp table
        database.execSQL("CREATE TABLE IF NOT EXISTS `bookmarks_temp` (id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, title TEXT, url TEXT NOT NULL, UNIQUE (url) ON CONFLICT REPLACE)")
        // copy data across
        database.execSQL("INSERT INTO `bookmarks_temp` (id, title, url) SELECT * FROM `bookmarks`")
        // drop old table
        database.execSQL("DROP TABLE `bookmarks`")
        // rename new table
        database.execSQL("ALTER TABLE `bookmarks_temp` RENAME TO `bookmarks`")
    }
}
```
I guess we'd also need to add `@Insert(onConflict = REPLACE)` to the `BookmarksDao#insert` method.

I haven't tested the SQL above so might need some tuning.

Do you mind give this approach a go?

bluetoothfx:
@aitorvs Thanks, that is great input. I will do as you have instructed and will update it accordingly. I should consider the scenario that you have mentioned. Really nice solution.

aitorvs:
@bluetoothfx thanks for PR, neat! After testing I have couple comments:
* We also need to commit the Room schema file. When you compile your branch you should have a file created called `30.json`, just add it to this PR
* It'd be great if you'd also added test for the migration, you can take a look at `AppDatabaseTest.kt`. The are some test that pre-populate the DB with some data, then run the migration and verify all is good
* While testing I've noticed that while this approach solves the issue for app upgrades, it does not upon fresh app install. The main issue seems to be although Room respect the `unique` constraints in the DB, as the entity does not declare that constraint (nor we can declare it as Room does not yet support that feature in its public API), when the DB is newly created upon fresh install the bookmarks table will still allow duplicates. To repro the issue:
  * delete and re-install the app
  * add the same bookmark several times
  * bookmarks will be duplicated ❌

One simple solution for the last point above is to run the same migration upon DB creation to add the `unique` contraint. For that we could register a `RoomDatabase.Callback` that preforms that in `onCreate`.
```kotlin
abstract class AppDatabase : RoomDatabase() {
    ....
    val MIGRATION_29_TO_30: Migration = object : Migration(29, 30) {...}

    val BOOKMARKS_DB_ON_CREATE = object : RoomDatabase.Callback() {
        override fun onCreate(database: SupportSQLiteDatabase) {
            MIGRATION_29_TO_30.migrate(database)
        }
    }
    ....
}
```

then just registre that callback

```kotlin
class DatabaseModule {

    @Provides
    @Singleton
    fun provideDatabase(context: Context, migrationsProvider: MigrationsProvider): AppDatabase {
        return Room.databaseBuilder(context, AppDatabase::class.java, "app.db")
            .addMigrations(*migrationsProvider.ALL_MIGRATIONS.toTypedArray())
            .addCallback(migrationsProvider.BOOKMARKS_DB_ON_CREATE)
            .build()
    }
    ...
}
```

It'd be great if we also added a comment to the `BookmarkEntity` pointing out that every time we change the entity we also need to check (and probably change) `migrationsProvider.BOOKMARKS_DB_ON_CREATE`.

bluetoothfx:
@aitorvs I will do necessary changes. Thank you for your quick feedback.

aitorvs (approved):
This is awesome @bluetoothfx, thanks a lot for your contribution. I will approve and merge

Tested in devices with API 24 and API 30 and all works as expected.

_Test(ed) cases_
1. install app from develop
2. add same bookmark several times
3. upgrade app from this branch
4. go to bookmarks
5. verify duplicates are gone
6. add same bookmark several times
7. go to bookmarks
8. verify there are no duplicates

1. fresh install from this branch
2. add same bookmark several times
3. go to bookmarks
4. verify there are no duplicates

Issue #1063, opened 2021-01-13
Sites requesting Client Certificate not working on Android


theyoz:
Websites are not working if they require a client certificate. Client certificate are uploaded.

Step to reproduce:
 * Upload pkcs12 Client Certificate (I can provide you a certificate and test site)
 * Go to the site with Google Chrome it works (Android add it for app and/or vpn use)
 * Go to DuckDuckGo and it fails

cmonfortep:
@aitorvs I know you worked on something related to certificates in https://github.com/duckduckgo/Android/pull/1031 and maybe you are more familiar than me here. Could you check if this is due to a WebView limitation or it's a bug? Thanks

aitorvs:
@theyoz thanks a lot for reporting this. Can you provide us with the cert and test site to repro the issue?

theyoz:
**Issue 1. Install the certficate**
URL: [http://test.yoram.me/download/ddg.p12](http://test.yoram.me/download/ddg.p12)
Password: duckduckgo

Response:
HTTP/1.1 200 OK
Content-Type: application/x-pkcs12

Video explaining the problem: https://user-images.githubusercontent.com/26814402/104548307-32c79f80-5639-11eb-95be-72c882eca39a.mp4

**Issue 2. Accessing sites with Client Certificate**

When I go to a site requiring a client certificate, it fails under DDG but works under Chrome

Video explaining the problem: https://user-images.githubusercontent.com/26814402/104549175-fb59f280-563a-11eb-8ac9-370f132ea1c3.mp4

**_All videos are in H.265 - Under Windows you may need a codec._**

aitorvs:
@theyoz thanks a lot for this report and your thoroughness describing it. It is very useful.
I can confirm this is indeed bug on our side.

@cmonfortep I have created an internal task for this so that we can triage it https://app.asana.com/0/414730916066338/1199561121099458/f

theyoz:
Guys,

Not sure If you had time to work on it but this work for me: https://github.com/duckduckgo/Android/pull/1078

cmonfortep:
We have an internal task to work on this issue in our backlog. I've tagged the Issue as `Will Fix` by our team.

We recently updated our [contribution guidelines](Let's sync on this if you can @Robert, ) to improve how we deal with reported issue.

PR #1078, opened 2021-01-27
User Credential on Android system being presented when accessing Mutu…

theyoz:
Task/Issue URL:  https://github.com/duckduckgo/Android/issues/1063

**Description**:

Hi,

This code does the trick for our sites. It is important to note that RSA keys ONLY are requested by this pull request, I didn't test it with EC keys et al (see [choosePrivateKeyAlias](https://developer.android.com/reference/android/security/KeyChain#choosePrivateKeyAlias(android.app.Activity,%20android.security.KeyChainAliasCallback,%20java.lang.String[],%20java.security.Principal[],%20java.lang.String,%20int,%20java.lang.String)) field keyTypes

However RSA is the most common and if adding the rest to the String[] list still works, you may as well.

My apologies for proposing a change, I really want to you DuckDuckGo but without this I can't :)

This fix does not fix the certificate download but that is the easier part.

###### Internal references:
[From Issue](https://app.asana.com/0/414730916066338/1199561121099458/f)

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -16,11 +16,14 @@
 
 package com.duckduckgo.app.browser
 
+import android.app.Activity
 import android.graphics.Bitmap
 import android.net.Uri
 import android.net.http.SslError
 import android.net.http.SslError.*
 import android.os.Build
+import android.security.KeyChain
+import android.security.KeyChainException
 import android.webkit.*
 import androidx.annotation.RequiresApi
 import androidx.annotation.UiThread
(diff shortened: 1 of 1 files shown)


aitorvs:
Thanks for this PR! we'll be working internally in a generic solution that works for all sites. I think it's best to close this one for now and wait for the more generic one.

initpwn (changes requested):
Not generic.

theyoz:
Hi,

Just to let you know that the server on which I installed the test site will run out on 5 mars 2021. Let me know if you need any more help from me.
Thanks

cmonfortep:
Thanks @theyoz, and sorry for the unclarity here. We've decide to work on this internally, so I will close this PR. Thanks for the setup and your time here, but we are going to provide a generic solution that works for all sites.

PR #1091, opened 2021-02-05
Rename unprotected URL

jonathanKingston:
Task/Issue URL: https://app.asana.com/0/11472677167855/1185888722207947
Tech Design URL: 
CC: 

**Description**:

**Depends on SERP repo being merged**: https://dub.duckduckgo.com/duckduckgo/ddg/pull/21513

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/trackerdetection/api/TrackerListService.kt
+++ b/app/src/main/java/com/duckduckgo/app/trackerdetection/api/TrackerListService.kt
@@ -24,7 +24,7 @@ interface TrackerListService {
     @GET("https://staticcdn.duckduckgo.com/trackerblocking/v2.1/tds.json")
     fun tds(): Call<TdsJson>
 
-    @GET("/contentblocking/trackers-whitelist-temporary.txt")
+    @GET("/contentblocking/trackers-unprotected-temporary.txt")
     fun temporaryWhitelist(): Call<String>
 
 }


malmstein:
SERP PR is merged, @jonathanKingston is there anything we need to do in order to test this?

jonathanKingston:
@malmstein I was looking through the Android code and my understanding is `TrackerDataLoader` is surfaced to TrackerDetectorImpl which sole responsibility is to block tracker URLs in content pages.

The lists are now different: https://duckduckgo.com/contentblocking/trackers-whitelist-temporary.txt and https://duckduckgo.com/contentblocking/trackers-unprotected-temporary.txt so loading scripts from those two different domains into a page shouldn't be blocked with the change.

 Perhaps @subsymbolic would be able to say if there's a better way to test this? Perhaps checking the browser is downloading the right list with network tools?

Also I'd like to note this is just to fix the URL, there are still other places where the terminology should change etc.

subsymbolic:
@malmstein in case it helps, I would test this as follows:

* Install the app fresh to make sure none of the data already exists in our db from the previous end-point
* Launch the app and wait for all backend syncing to happen
* Visit a couple of urls in https://duckduckgo.com/contentblocking/trackers-unprotected-temporary.txt and make sure that NO trackers are identified
* You should also see a comment in the console that the site is whitelisted (as mentioned we still have some of that old terminology in the app)

malmstein:
Thanks @subsymbolic, I'll take care of it @jonathanKingston !

malmstein:
Thanks @jonathanKingston @subsymbolic, works as expected!

malmstein (approved):
:shipit:

PR #1099, opened 2021-02-12
Fixed search bar selection behavior

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1199931182139250/f
Tech Design URL: 
CC: 

**Description**:
When we fixed the requery pixels (https://github.com/duckduckgo/Android/pull/1065) we also had to tune the search bar behavior, causing a side effect where URLs in thee search bar would never be selected when the user taps on the search bar.

This PR fixes that issue

**Steps to test this PR**:

1. install/update from this branch
2. Ensure tests in https://app.asana.com/0/0/1199922227966104/f pass (to see when rq pixels are sent, filter by `rq_` in logcat)

1. install/update from this branch
2. open app and tap on search bar
3. search for "reddit"
4. select "🔎 reddit"
5. tap on the search bar
6. verify text is not selected and autocomplete results show up
7. open the first URL in the SERP
8. tap on the search bar
9. verify URL text is selected

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
@@ -19,6 +19,7 @@ package com.duckduckgo.app.browser.omnibar
 import android.content.Context
 import android.graphics.Rect
 import android.text.Editable
+import android.text.Selection
 import android.util.AttributeSet
 import android.util.Patterns
 import android.view.KeyEvent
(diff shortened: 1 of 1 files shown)


marcosholgado (changes requested):
All rq test passing but there is one thing that I've noticed. 

After selecting a URL (`didFocusAlready = false`) and going back to SERP the whole search team is selected and autocomplete is not triggered. I think if we remove `didFocusAlready` from the if in the line 49 it should be fixed. I think is enough with `text?.isWebUrl() == false` since the previous focus state should not matter anymore.

1. Search for lego.
1. Click on any search result.
1. Click the URL, the whole url gets selected as expected.
1. Using the back button go back to SERP.
1. Select the url. The full search term is selected rather than autocomplete being triggered.

aitorvs:
> All rq test passing but there is one thing that I've noticed.
> 
> After selecting a URL (`didFocusAlready = false`) and going back to SERP, the whole search team is selected and autocomplete is not triggered. I think if we remove `didFocusAlready` from the if in the line 49 it should be fixed. I think is enough with `text?.isWebUrl() == false` since the previous focus state should not matter anymore.
> 
> 1. Search for lego.
> 2. Click on any search result.
> 3. Click the URL, the whole url gets selected as expected.
> 4. Using the back button go back to SERP.
> 5. Select the url. The full search term is selected rather than autocomplete being triggered.

Thanks @marcosholgado  it makes sense, done!

marcosholgado (approved):
Nice job! 🚢  🇮🇹

Issue #1074, opened 2021-01-24
Text formatting isn't properly cleared on copy-paste


yakkomajuri:
**Describe the bug**

Pasting larger text (e.g. headers) into the address bar doesn't completely clear the formatting:

<img src="https://user-images.githubusercontent.com/38760734/105640534-85ddf580-5e76-11eb-9002-253f81b00784.jpeg"  height="400"/>

**How to Reproduce**
1. Copy some header text (like the title of this issue) into the address bar
2. Watch it stay as "large text"

**Expected behavior**

The text should be cleared of formatting and standardized/

**Smartphone (please complete the following information):**
 - DDG App Version: 5.74.0 zu
 - Device: OnePlus 8T
 - OS: Android 11
 - If applicable, add screenshots to help explain your problem.

cmonfortep:
Thanks for reporting that one. I will leave this task open in case anyone from our community wants to provide a fix.

yakkomajuri:
On [our repo](github.com/PostHog/posthog) we get a lot of community action from our `good first issue` tag - you could maybe add it?

I would give this a look if I wasn't too busy to setup an Android dev environment

hotchemi:
Opened https://github.com/duckduckgo/Android/pull/1105🙇

cmonfortep:
Closing this as it has been resolved. Thanks @hotchemi

PR #1105, opened 2021-02-17
fix: convert clip data into plain text before pasting to omnibar

hotchemi:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/1074

**Description**:

- This PR modifies `KeyboardAwareEditText` to be aware of pasting text from the clipboard and clear its style beforehand.
  - If API level >= M, we can just use [R.id.pasteAsPlainText](https://developer.android.com/reference/android/R.id#pasteAsPlainText)
  - If API level < M, we need to manually clear and set the data to primary clip again(see `ClipboardManager.convertClipToPlainText`)

#### Discussion

- Now `KeyboardAwareEditText` has kind of two responsibilities and I was thinking I should rename the class as `OmniToolbarEditText` or something, but since this is my 1st PR as an external contributor I'd like to keep the diff smaller as much as I can. Let me know your thoughts around here🙏

**Steps to test this PR**:

1. API level >= M
  - Launch the emulator with API Level M
  - Open https://www.theverge.com/2018/1/23/16922508/private-search-engine-duckduckgo-browser-extension-app
  - Copy bold "DuckDuckGo" from the title
  - Paste the clipboard clip to omnibar search input area
  - Confirm there's no rich text formatting
 
2. API level < M
  - Launch the emulator with API Level M
  - Open https://www.theverge.com/2018/1/23/16922508/private-search-engine-duckduckgo-browser-extension-app
  - Copy bold "DuckDuckGo" from the title
  - Paste the clipboard clip to omnibar search input area
  - Confirm there's no rich text formatting

##### API level >= M

| Before |  After |
:-------------------------:|:-------------------------:
|

##### API level < M

| Before |  After |
:-------------------------:|:-------------------------:
|
---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/KeyboardAwareEditText.kt
@@ -16,8 +16,11 @@
 
 package com.duckduckgo.app.browser.omnibar
 
+import android.content.ClipData
+import android.content.ClipboardManager
 import android.content.Context
 import android.graphics.Rect
+import android.os.Build
 import android.text.Editable
 import android.util.AttributeSet
 import android.util.Patterns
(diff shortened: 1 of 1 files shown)


cmonfortep:
Thanks, we will take a look as soon as possible.

malmstein (changes requested):
Thanks for the submission @hotchemi. This is indeed very helpful and fixes an issue, good job!

On the actual code:
1. How do you feel about not using extension functions? We don't really need this feature anywhere else and I'd prefer if the `convertClipToPlainText` function would only be known to the `KeyboardEditText`
2. On the naming of the `EditText` itself, we are working on a few things internally that will change its behavior, so it's better not to change anything else just yet.

hotchemi:
@malmstein Thank you for the quick review!

- RE: 1 I'm fine with your suggestion and let me make my aim clear just in case🙏 My intention here for using Kotlin's extension function is to provide a more readable way to define logic and actually those extensions are not available outside of `KeyboardAwareEditText`(because those are private, which I've occasionally seen in Kotlin [repo](https://github.com/JetBrains/kotlin/blob/cc51869a2abb590b7c830cece7a6bec94901fdfa/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/annotationProcessing.kt#L222-L234))🙇 

- RE: 2 Roger that!

malmstein:
@hotchemi thanks for clarifying, I completely missed the `private` on the extension functions declaration! It actually reads quite nice and I'm happy with it.

Thanks for the contribution!

malmstein (approved):
Happy with the changes because the extension functions are private to the `View`
:shipit:

hotchemi:
Thank you so much for your review😄🙏

PR #1118, opened 2021-03-03
Clean up DBs when clear personal data action executed

marcosholgado:
Task/Issue URL: https://app.asana.com/0/inbox/1125189844075764/1199685456065107/1200036478062018
Tech Design URL: 
CC: 

**Description**:
This PR adds an extra clean up process when the fire button is used to make sure the deleted data is indeed deleted. We do apply this new change to the cookies db and http_auth.db.

**Steps to test this PR**:
This PR should be tested against all APIs from 21 to 30

1. Launch the app.
1. Go to `https://samesitetest.com/cookies/set`which will store cookies.
1. Go to `https://httpbin.org/basic-auth/u/pw` and enter username=u password=pw.
1. Use the fire button
1. After the app has restarted open the device file explorer
1. Go to `data/data/com.duckduckgo.mobile.android.debug`
1. In the `app_webview` or `app_webview/Default` folder (depending on the API), right click on the `Cookies` DB and click synchronize. After that right click again and save the file in your Desktop (or wherever you want).
1. In the `databases` folder, right click on the `http_auth.db` file and click synchronize. After that right click again and save the file in your Desktop (or wherever you want).

Check the following:
1. In the `databases` folder, the file `http_auth.db-journal` should have been deleted.
1. Open the `Cookies DB` downloaded with a text editor and make sure you cannot find any traces from `samesite`
1. Open the `http_auth.db` downloaded with a text editor and search for `httpbin`. You should not find anything.

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
```diff
@@ -202,7 +203,16 @@ class BrowserModule {
...
-    fun webViewDatabaseLocator(context: Context): WebViewDatabaseLocator = WebViewDatabaseLocator(context)
+    @Named("webViewDbLocator")
+    fun webViewDatabaseLocator(context: Context): DatabaseLocator = WebViewDatabaseLocator(context)
+
+    @Provides
+    @Named("authDbLocator")
+    fun authDatabaseLocator(context: Context): DatabaseLocator = AuthDatabaseLocator(context)
+
+    @Provides
+    @Named("authDbCleaner")
```
aitorvs, line 214: I think we should remove this named qualifier. There is only one implementation of the `DatabaseCleaner`

app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
```diff
@@ -36,8 +38,9 @@ class DatabaseModule {
     @Singleton
     fun provideWebViewHttpAuthStore(
         context: Context,
+        @Named("authDbCleaner") databaseCleaner: DatabaseCleaner
```
aitorvs, line 41: ditto, we can remove the named qualifier

app/src/main/java/com/duckduckgo/app/fire/DatabaseCleaner.kt
```diff
@@ -0,0 +1,62 @@
...
+import android.database.sqlite.SQLiteDatabase
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import timber.log.Timber
+
+interface DatabaseCleaner {
+    suspend fun cleanDatabase(): Boolean
+}
+
+class AuthDatabaseCleaner(private val databaseLocator: DatabaseLocator) : DatabaseCleaner {
```
aitorvs, line 28: I would suggest to rather pass directly the db path as oppose to passing the locator.
It took me a while to track down what DB was being cleaned-up. Given that we're passing an interface (`DatabaseLocator`) that has two implementations.

In the `BrowserModule` (and its stub) we could do

```kotlin
    @Provides
    fun authDatabaseCleaner(@Named("authDbLocator") authDatabaseLocator: DatabaseLocator): DatabaseCleaner = AuthDatabaseCleaner(authDatabaseLocator.getDatabasePath())
```

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
```diff
@@ -58,4 +63,8 @@ class RealWebViewHttpAuthStore(
     override fun clearHttpAuthUsernamePassword(webView: WebView) {
         webViewDatabase.clearHttpAuthUsernamePassword()
     }
+
+    override suspend fun cleanHttpAuthDatabase() {
+        databaseCleaner.cleanDatabase()
+    }
```
aitorvs, line 69: I am struggling with the `databaseCleaner`. It right now is an interface that has only one implementation , the `AuthDatabaseCleaner`. Is this bc before we had also an impl for the app db?

My struggle is that `databaseCleaner.cleanDatabase()` gives me no information about what I am cleaning. So I need to rely on other things like, in this case, its wrapped in a method called `cleanHttpAuthDatabase`. But anyway I feel compelled to go and look how it is injected to see whether we have injected the proper implementation.

I'd suggest to modify the `DatabaseCleaner#cleanDatabase` signature so that it receives the db path. But that relies on the fact that we now only have one db cleaner.
Give that we don't know what will happen with the app db, I would still propose this.


Note: if we agree on this approach, you can forget my comments about passing the db path in the constructor as oppose to passing the `DatabaseLocator`. We'd now pass nothing

app/src/main/java/com/duckduckgo/app/browser/WebDataManager.kt
```diff
@@ -95,10 +102,15 @@ class WebViewDataManager @Inject constructor(
...
+        private const val DATABASES_DIRECTORY_NAME = "databases"
 
         private val WEBVIEW_FILES_EXCLUDED_FROM_DELETION = listOf(
             "Default",
             "Cookies"
         )
+
+        private val HTTP_AUTH_FILES_TO_DELETE = listOf(
+            "http_auth.db-journal"
+        )
```
aitorvs, line 114: This will always rely on the fact that http_auth.db will always use rollback journal. Which might not be always the case, as we don't currently control that db.
What do you think about on app start, setting the journal mode of the `http_auth.db` as `DELETE`? That will automatically delete the journal file after each transaction and save all this code

aitorvs, line 114: This assumes the `http_auth` db will always use a rollback journal, which might not be the case always.

We could also try to set the journal mode of the db to `DELETE`. That should delete the journal every time a transaction happens.
AFAIR from when I was doing PR #1066, this DB journal mode is set to `TRUNCATE`, and that might be why the journal file is not deleted after each transaction

I don't think we should worry about performance here, we're not doing intensive read/writes in the db

app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
marcosholgado: Done!

app/src/main/java/com/duckduckgo/app/fire/DatabaseCleaner.kt
```diff
@@ -0,0 +1,62 @@
...
+        var cleanUpExecuted = false
+        openReadableDatabase(databasePath)?.apply {
+            try {
+                execSQL("VACUUM")
+                cleanUpExecuted = true
+            } catch (exception: Exception) {
+                Timber.e(exception)
+            } finally {
+                close()
+            }
```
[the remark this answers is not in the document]
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
```diff
@@ -31,11 +35,21 @@ interface WebViewHttpAuthStore {
...
 
 class RealWebViewHttpAuthStore(
-    private val webViewDatabase: WebViewDatabase
+    private val webViewDatabase: WebViewDatabase,
+    private val databaseCleaner: DatabaseCleaner
 ) : WebViewHttpAuthStore {
+
+    init {
+        GlobalScope.launch {
+            databaseCleaner.changeJournalModeToDelete()
```
marcosholgado, line 49: Rather than manually deleting the journal, like @aitorvs suggested it works better to manually change the journal mode to delete. I am however not massively convinced about doing it here so happy to hear suggestion. I was thinking in a lifecyle observer oncreate of the app.

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
marcosholgado: I ended up passing the db path to the constructor instead of this but no strong opinions.

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
aitorvs: yeah, I think a better place is in an `onCreate` app observer

CDRussell:
@aitorvs , if you don't mind i'm gonna mark you as _assignee_ for this PR since you're thoroughly reviewing it anyway

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
marcosholgado: Changed!

app/src/main/java/com/duckduckgo/app/browser/WebDataManager.kt
marcosholgado: Agreed and done!

aitorvs (approved):

marcosholgado:
@aitorvs this should (finally) be ready, leaving it up to you if you want to double check the changes.

aitorvs (approved):
Nicely done! 🚢 🇮🇹

PR #1124, opened 2021-03-09
Android: streamline home screen load

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1200024092221607/f
Tech Design URL: 
CC: 

**Description**:
This PR includes changes to make our clear data restart process smoother and faster:
- Removed fade-in transitions
- DDG logo fixed position
- Removed Data clear message when the user clears data manually (automatic data clear still shows the confirmation message)
- layout optimizations
- dagger graph initialized only for the app process (fire process will not initialize dagger injector)
- Browser activity has a dummy omnibar to avoid showing a white full screen

**Steps to test this PR**:
Test 1:
1. Perform fresh install
1. Clear data in the 3 different possible scenarios (during onboarding, onboarding completed - widget cta, onboarding completed - no Ctas)

Test 2:
1. Ensure automatic data clear works as usual (includes optimization changes but data clear message shows up if app cleared in foreground)
 
> See videos posted in [product review](https://app.asana.com/0/1142021229838617/1200036139228532/f) to validate expected behavior

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1870,29 +1847,14 @@ class BrowserTabFragment :
             newTabLayout.setOnClickListener(null)
         }
 
-        private fun showHomeTopCta(configuration: HomeTopPanelCta) {
```
marcosholgado, line 1873: If we are removing the home top panel functionality, does it make sense to also remove all the references to it?
- `HomeTopPanelCta` sealed class
- References in the `BrowserTabViewModel`
- `include_top_cta.xml` layout

app/src/main/java/com/duckduckgo/app/browser/HomeBackgroundLogo.kt
```diff
@@ -0,0 +1,67 @@
...
+            interpolator = AccelerateInterpolator()
+            alpha(1f)
+            ddgLogoView.show()
+        }
+    }
+
+    private fun fadeLogoOut() {
+        Timber.v("hideLogo")
+        ddgLogoView.hide()
+
```
marcosholgado, line 61: nit: empty line

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -142,6 +145,14 @@ class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope() {
         }
     }
 
+    private fun configureEmptyClickListeners() {
```
marcosholgado, line 148: To avoid this since this is a mock view that is not supposed to work anyway, can we make the layout non-clickable? `android:clickable="false"`. I think that should do the trick.

marcosholgado (changes requested):
Nice job @cmonfortep! Looks amazing and much faster than before, specially on 21! I left a few comments but nothing major.

- [X] Tested in API 21 and 30
- [X] Light/dark mode
- [X] Clear data feature

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
cmonfortep: Good point, that was clever but I've just tested and it doesn't work.

why? because inside that custom view, in the `onFinishInflate` we are calling:

```
        setOnClickListener {
            super.callOnClick()
        }
```

I'm 99% sure we can remove that part, what do you think?

marcosholgado: Yeah I think so. I tested it and seems to be fine but if you are not sure, just leave it. Not the end of the world.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
cmonfortep: I was in doubt about this one. While I was ok removing parts that were slowing down the startup process + som warnings caused by that. I was ok moving the final cleanup into a LHF task. But since you asked for that, and you know about that HomeTopCta, I will remove it 100%

cmonfortep:
@marcosholgado could you re-review this again? Thanks!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Yeah, we don't use it anymore and I doubt we ever will 👍

marcosholgado (approved):
Thanks for the changes @cmonfortep! Ship it! 👍

PR #1136, opened 2021-03-18
Close correct tab when close window called

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1199674134590488
Tech Design URL: 
CC: 

**Description**:
This PR fixes a bug where we close the wrong tab after receiving a close window event.

**Steps to test this PR**:

**Google Sign in**
1. Install the app
1. Open a tab and go to SERP
1. Open another tab and go to a site with a google login like `etsy.com`
1. Sign in with your google account, once the flow has finished you should be returned to the tab which initiated the flow
1. Some sites require us to enable 3p cookies for the login with Google to work, in those cases you will not be signed in but you should still be returned to the correct tab.

**Facebook Sign in**
1. Install the app
1. Open a tab and go to SERP
1. Open another tab and go to a site with a Facebook login like `pinterest.com`
1. Sign in with your facebook account, once the flow has finished you should be returned to the tab which initiated the flow.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
@@ -216,9 +216,9 @@ class TabDataRepository @Inject constructor(
         }.join()
     }
 
-    override suspend fun deleteCurrentTabAndSelectSource() {
+    override suspend fun deleteCurrentTabAndSelectSource(tabId: String) {
         databaseExecutor().scheduleDirect {
-            val tabToDelete = tabsDao.selectedTab() ?: return@scheduleDirect
+            val tabToDelete = tabsDao.tab(tabId) ?: return@scheduleDirect
 
             deleteOldPreviewImages(tabToDelete.tabId)
             val tabToSelect = tabToDelete.sourceTabId
--- a/app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
@@ -216,7 +216,7 @@ class TabDataRepository @Inject constructor(
         }.join()
     }
 
-    override suspend fun deleteCurrentTabAndSelectSource(tabId: String) {
+    override suspend fun deleteTabAndSelectSource(tabId: String) {
         databaseExecutor().scheduleDirect {
             val tabToDelete = tabsDao.tab(tabId) ?: return@scheduleDirect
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
```diff
@@ -216,9 +216,9 @@ class TabDataRepository @Inject constructor(
         }.join()
     }
 
-    override suspend fun deleteCurrentTabAndSelectSource() {
+    override suspend fun deleteCurrentTabAndSelectSource(tabId: String) {
```
cmonfortep, line 219: @marcosholgado I think we talked about changing this method name to remove the part of `CurrentTab` now that we are passing `tabId` as parameter. Are you ok with that?

app/src/main/java/com/duckduckgo/app/tabs/model/TabRepository.kt
```diff
@@ -66,7 +66,7 @@ interface TabRepository {
      */
     suspend fun purgeDeletableTabs()
 
-    suspend fun deleteCurrentTabAndSelectSource()
+    suspend fun deleteCurrentTabAndSelectSource(tabId: String)
```
cmonfortep, line 69: Same here, I would remove the `currentTab` reference from the method name since we can pass any tab id as parameter.

cmonfortep (approved):
Just a comment related to method naming. 

Tested all the scenarios posted in the description and works as described.

Throwing a question here:
- Last time we tried to fix this issue we introduced a bug where the root cause was overriding the loadURL on facebook sites. I guess that code is not present anymore...so it's not possible to reproduce it again. What will happen if a similar code is introduced again? Should we test how we will react to a similar scenario?

app/src/main/java/com/duckduckgo/app/tabs/model/TabDataRepository.kt
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/tabs/model/TabRepository.kt
marcosholgado: 👍

marcosholgado:
> Just a comment related to method naming.
> 
> Tested all the scenarios posted in the description and works as described.
> 
> Throwing a question here:
> 
> * Last time we tried to fix this issue we introduced a bug where the root cause was overriding the loadURL on facebook sites. I guess that code is not present anymore...so it's not possible to reproduce it again. What will happen if a similar code is introduced again? Should we test how we will react to a similar scenario?

Thanks @cmonfortep. 

We are still reloading the URL when changing the user agent so we can reproduce. Currently if you sign in with Facebook the onCloseWindow method still gets called several times so that's why I added the test case specific to Facebook.

PR #1123, opened 2021-03-08
Refactor the ViewModelFactory to use multibindings

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1199916263225062/f
Tech Design URL: 
CC: 

**Description**:
The `ViewModelFactory` is in charge of creating all the view models for the whole app. Before it had to:
- depend on all the dependencies required by the view models
- have visibility of all the view model types
- classes that needed a view model had to have a reference to the view model factory

By using the plugin pattern in this class we:
- decoupled the view model `Factory` class from the actual factories that create the view models
- both factories and the view model `Factory` can now live in different modules if needed without creating circular dependencies

**Steps to test this PR**:
This PR affects pretty much the whole app as it touches all the view models we've got. To test the PR test the common app screens to see if something is amiss.

1. App compiles
2. install/upgrade the app from this branch
3. use the SERP to perform searches
4. form the SERP, open several tabs
5. add/remove bookmarks
6. add/remove fireproof sites
7. navigate to settings screen
8. use fire button
10. clear app data and go through onboarding UX
11. for all above, verify app works as expected

---
###### Internal references:
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #1145, opened 2021-03-24
Use providers in the VM factories to respect the dep di semantics

aitorvs:
Task/Issue URL: https://app.asana.com/0/30173902528854/1200098175084894/f
Tech Design URL: 
CC: 

**Description**:
https://github.com/duckduckgo/Android/pull/1123 introduced a bug that was detected by our pixels.

* That pull request delegated the creation of the view models (VMs) to its own factory
* The VM factories are `singleton` -- as they should be
* The VM factories create the VMs and inject their dependencies
* Because the VM factories are `singleton` that effectively makes all VM dependencies `singleton`. Regardless of whether they are actually declare as such or not
* This is not desired, VM factories thouls be singleton, but it should respect the di semantics of its deps

This PR makes all the factories to provide `providers` of the dependencies, rather than the actual instance. So that they respect how the instances declare themselves when injected, ie. `singleton` or multiple instances.

In particular, the effects can be seen with the `fw_` pixels. They are fired multiple times (one per fragment) when they should fire only once.

**Steps to test this PR**:
1. App compiles
2. install/upgrade the app from this branch
3. use the SERP to perform searches
4. form the SERP, open several tabs
5. add/remove bookmarks
6. add/remove fireproof sites
7. navigate to settings screen
8. use fire button
10. clear app data and go through onboarding UX
11. for all above, verify app works as expected

2. open several tabs and ensure they are active
3. ensure `Ask When Signing In` is enabled inside `settings -> fireproof sites`
4. open a new tab and log into facebook.com (or any other login you may want)
5. verify the dialog to fireproof the site appears
6. navigate to any other of the active tabs
7. verify no dialog appears
8. verify `m_fw_xx` are only shown once

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksViewModel.kt
@@ -37,6 +37,7 @@ import dagger.multibindings.IntoSet
 import io.reactivex.schedulers.Schedulers
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.launch
+import javax.inject.Provider
 import javax.inject.Singleton
 
 class BookmarksViewModel(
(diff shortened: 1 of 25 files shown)


cmonfortep (approved):
Thanks @aitorvs 

Just a minor comment:
I think providers are missing in `WelcomePageViewModelFactory`, could it be?

I've just tried to test a bit the app and everything seems to work as expected, GJ.
I've also put an eye on the logcat to ensure no duplicated pixels were sent.
I've checked that we are not sharing instances between fragments when they are not declared as singleton (`NextPageLoginDetection`), and we have a single instance when they are declared as singleton (`SiteFactory`). All good

cmonfortep:
BTW, please, remember to create a task in Asana for this task and add it to the release board 👍

aitorvs:
> I think providers are missing in `WelcomePageViewModelFactory`, could it be?
That factory is not part of the `ViewModelFactoryPlugin`s. `WelcomePage` fragment uses that factory instead of the generic app `ViewModelFactory`.
We should move it there, will do in a separate task/PR

aitorvs:
> BTW, please, remember to create a task in Asana for this task and add it to the release board 👍 (and add it to the description)

Will do

PR #1160, opened 2021-04-04
Czech - Fixed translation of 'Light Theme'

zdenekobornik:
**Description**:
Czech translation of "Light theme" is wrong. The word "Lehký" is being used when something is not heavy (it's related to weight and not to brightness). The proper translation is "Světlý".


cmonfortep:
Thanks for the contribution and for the good explanation. It makes sense to fix this copy.

PR #1186, opened 2021-04-12
Fixed Czech translation for two menu items

vojta-horanek:
**Description**:
This PR improves/fixes the translation of two menu items in the Czech language. Namely:

- `requestDesktopSiteMenuTitle` - "Stránka plochy" means pretty much nothing and the purpose of this menu items is not clear at all. I've changed it to "Verze pro počítač". It is the same string firefox uses in it's mobile web browser and it is actually understandable.
- `fireproofWebsiteMenuTitleAdd` - "Web s ochranou" is good, but I also feel like it is not clear and doesn't describe an action (the English version does). That is why I've changed it to "Ochránit web" which is an action.

Initially I also wanted to fix `settingsLightTheme`, but someone was a bit quicker than me 😃


cmonfortep:
Thanks for spotting those issues and providing a fix.

PR #1197, opened 2021-04-21
Add autofill

marcosholgado:
Task/Issue URL: https://app.asana.com/0/inbox/1125189844075764/1200187721780181/1200227836639831
Tech Design URL: 
CC: 

**Description**:
See task

**Steps to test this PR**:
See task

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) 2020 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.email
+
+import androidx.annotation.UiThread
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.Observer
+import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.asLiveData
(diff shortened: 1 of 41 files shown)


app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    override var nextAlias: String? = null
+
+    private val nextAliasLiveData: LiveData<String?> = emailDataStore.nextAliasFlow().asLiveData()
+
+    private val alias = Observer<String?> { alias ->
+        nextAlias = alias
+    }
+
+    private val isSignedInChannel = ConflatedBroadcastChannel(isSignedIn())
```
CDRussell, line 64: It's hard to read some of this, and knowing the difference between things like `nextAlias` and `getAlias()` etc..

Not a blocker for now, but would be good to revisit this and see if we:

1. could remove `LiveData` from this layer
2. need `ConflatedBroadcastChannel`
3. can clarify the difference between `nextAlias` and `getAlias` and make that API more obvious

app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
```diff
@@ -0,0 +1,150 @@
...
+    @UiThread
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun onAppBackgrounded() {
+        nextAliasLiveData.removeObserver(alias)
+    }
+
+    override fun getAlias(): String? = consumeAlias()
+
+    override fun isSignedIn(): Boolean {
+        return !emailDataStore.emailToken.isNullOrBlank() && !emailDataStore.emailUsername.isNullOrBlank() && !emailDataStore.nextAlias.isNullOrBlank()
```
CDRussell, line 83: Worried that if `emailDataStore.nextAlias` returns `null`, user will be left in a state that's hard to easily recover. And seems that scenario could easily happen if `fetchAliasFromService` network request fails for any reason

CDRussell (approved):

app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
marcosholgado: Yeah makes sense, I've removed the alias from the signedIn method so it is a bit better but I'll create a task to talk about error scenario.

app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
marcosholgado: 👍  makes sense, I've migrated to StateFlow and removed `nextAlias` because is not really needed.

Issue #1159, opened 2021-04-02
complete black after clear all tab am data


stevenleeS0ht:
**Describe the bug**
The app directly turn to black and not responsive after clear all tab and data.

**How to Reproduce**
1. Touch button clear all tab and data.
2. See error

If applicable, please include information about any settings that may be relevant to this issue (e.g. selected theme, auto clear option, etc).

**Expected behavior**

**Smartphone (please complete the following information):**
 - DDG App Version: 5.78.1 zw
 - Device: XIAOMI
 - OS: Android One 11
 - If applicable, add screenshots to help explain your problem.

arunsathiya:
I see the same issue, has been happening for a few months but never had a chance to submit the bug report. My workaround at the moment is to close the app from "recent apps" list of Android and re-open it.

stevenleeS0ht:
@CDRussell  and @subsymbolic , do you have any ideas on this issue?

stevenleeS0ht:
@CDRussell , yes, when black the only thing I can do is quit app. The screen is completely dark.

arunsathiya:
https://user-images.githubusercontent.com/18581859/113480953-a056cf80-94b4-11eb-8c29-ad3508056ecd.mp4

@CDRussell Here we go. We can see that I am making some searches on DuckDuckGo and then clear all data, but the app blacks out and I can scroll down to see the actual page in some sort of a layered background. I open the recent apps list to clear DuckDuckGo from it, and then re-launch the app.

cmonfortep:
Thanks for all the info here.

We still don't understand why it fails. At that point where the screen turns black there should be an animation, so I guess the root cause should be related to that part. We will try to look into this issue and try to find a way to mitigate the issue for those cases where the animation fails to appear.

@stevenleeS0ht do you have animations disabled on your device by any chance? I guess the answer is no, but just double-checking this.

arunsathiya:
@cmonfortep Good hint! I have `Window animation scale`, `Transition animation scale` and `Animator duration scale` turned off under `Developer options` on my Android settings. By just updating `Animator duration scale` to 1x, the fire animation appears okay on DuckDuckGo app. It would be great if this bug is resolved. Thanks!

stevenleeS0ht:
@cmonfortep , Yes, developer mode enable, all animation turn off. Is it possible not to show the animation when the animation in setting are disabled?

cmonfortep:
Yes, I guess we can try that strategy.

[...]

PR #1199, opened 2021-04-21
Fix fire animation

jamie23:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/1159
Tech Design URL:
CC:

**Description**:
When we check that the animation is enabled we can also check that the animator duration scale is set, if it is set to 0 then we skip the animation.

**Steps to test this PR**:
1. Set `Animator duration scale` to `Animation off`
2. Select `Clear all tabs and data`. 
3. Now we skip the animation unlike previously where we would get stuck on a black screen.
---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
@@ -20,6 +20,8 @@ import android.animation.Animator
 import android.animation.ValueAnimator
 import android.content.Context
 import android.os.Bundle
+import android.provider.Settings
+import android.provider.Settings.Global.ANIMATOR_DURATION_SCALE
 import androidx.core.content.ContextCompat
 import androidx.core.view.doOnDetach
 import androidx.core.view.isVisible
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
```diff
@@ -140,7 +142,12 @@ class FireDialog(
         }
     }
 
-    private fun animationEnabled() = settingsDataStore.fireAnimationEnabled
+    private fun animationEnabled() = settingsDataStore.fireAnimationEnabled && animatorDurationEnabled()
+
+    private fun animatorDurationEnabled(): Boolean {
+        val animatorScale = Settings.Global.getFloat(context.contentResolver, ANIMATOR_DURATION_SCALE, 1.0f)
```
aitorvs, line 148: I think default value should not be `1f` but rather `0.0f` To fall on the safe side when the value is not defined.

aitorvs:
Good job @jamie23 . I left one minor comment to error on the safe side.

I have tested this for:
1. window animation scale off
2. transition animation scale off
3. animator duration scale off
4. and any combination of the above

All working properly

app/src/main/java/com/duckduckgo/app/global/view/FireDialog.kt
jamie23: Good spot @aitorvs, I've made the appropriate change.

aitorvs (approved):
Awesome @jamie23! thanks a lot for the contribution.

Issue #1188, opened 2021-04-14
Favicons get swapped!


oldphones68:
**Describe the bug**
Favicons get swapped.
This seems close to the issue #975. 

**How to Reproduce**
1. Go to 'Bookmarks'
2. See error

Although in #975 seeming @aitorvs has provided a patch, but I'm seeing this bug is still existing.

**Expected behavior**
Favicons should show correctly. 

**Smartphone**
 - DDG App Version: 5.80.0
 - OS: Android 6

Ammako:
Yup. Mine shows github with a reddit favicon.

cmonfortep:
Could you provide more info like: how many bookmarks you have? does it happen right after opening the screen or when scrolling? Do you see the same error on the tabs screen or only in the bookmarks screen?

Ammako:
4 bookmarks

Right when opening the screen

Only bookmarks screen

cmonfortep:
Thanks. That's weird, we use the same class to retrieve bookmark icon in the bookmarks screen and the tab screen (the one where you see all your tabs). 

We will investigate what's happening there

oldphones68:
18 bookmarks so far.

It happens after opening the browser.

Happens on the bookmarks menu. Didn't noticed (yet) in the tab screen.

@cmonfortep

cmonfortep:
@oldphones68 @Ammako another question from my side.

The sites that have a wrong favicon, do they show a favicon of a site that is also a bookmark? For example: in the case of @Ammako "github with a reddit favicon", do you have both sites bookmarked?

cmonfortep:
We have created an internal task to further investigate this issue: https://app.asana.com/0/414730916066338/1200206978219280/f

Ammako:
I don't have reddit bookmarked, no. Only github. But it must have picked up the favicon from when I was browsing reddit, either way.

Fire button to clear data doesn't fix it. Github is fireproofed, reddit is not.

cmonfortep:
Thanks so much @Ammako, does GitHub show the wrong favicon in fireproof websites too? I guess the answer is yes, but just to double check with you as we can't reproduce it in our side.

Ammako:
Reddit data gets cleared due to not being fireproofed, github data stays because it's fireproofed. Reddit favicon stays.

(I'm not entirely sure what you meant, so don't hesitate to elaborate if this doesn't properly answer your question.)

[...]

PR #1200, opened 2021-04-22
Fix wrong favicon issue

malmstein:
Task/Issue URL: https://app.asana.com/0/1174433894299346/1200206978219280

**Description**:
Some users reported that they see the wrong favicon showing on their bookmark screen (e.g: GitHub with a Reddit favicon). 
The problem comes from us storing a favicon which does not belong to the currently opened tab. 

We also have changed the amount of times we are replacing the default icon, it being in the view is enough. We don't need Glide to load a placeholder, nor manually reset it. The recycler view does that for us

Fixes: https://github.com/duckduckgo/Android/issues/1188

**Steps to reproduce**:
First, we are going to simulate the issue:
1. Open `BrowserTabViewModel` and add a delay(10000) to the coroutine in line 627
2. Launch the app

Now we are ready to see the problem
1. Open app and navigate to a site like github.com
2. In the same tab, open another site like reddit.com
3. Quickly open the Tab Switcher
4. You'll see that favicon for reddit.com is the one from github.com

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
```diff
@@ -64,14 +64,10 @@ class DuckDuckGoFaviconManager constructor(
     }
 
     override suspend fun loadToViewFromPersisted(url: String, view: ImageView) {
-        // avoid displaying the previous favicon when the holder is recycled
```
cmonfortep, line 67: Let's take a look at this together.

When you have more than 3 bookmarks, you can filter them. In that case, bookmarks without favicon could end up using the favicon from a recycled view.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -627,9 +627,15 @@ class BrowserTabViewModel(
...
 
-    override fun iconReceived(icon: Bitmap) {
+    override fun iconReceived(url: String, icon: Bitmap) {
         val currentTab = tabRepository.liveSelectedTab.value ?: return
-        val url = currentTab.url ?: return
+        val currentUrl = currentTab.url ?: return
+        if (currentUrl != url) {
+            pixel.enqueueFire(AppPixelName.FAVICON_WRONG_URL_ERROR)
+            Timber.d("Favicon received for a url $url, different than the current one $currentUrl")
+            return
```
cmonfortep, line 636: why not store that favicon? I think it should be ok to store it if we don't use it for the current URL. That way we will increase our hit ratio for favicons. WDTY?

cmonfortep: We have created an internal task to persist possible best quality favicons if they belong to persisted site: https://app.asana.com/0/414730916066338/1200237930163733/f

app/src/main/java/com/duckduckgo/app/browser/favicon/FaviconManager.kt
cmonfortep: Talked about this via zoom and decided to rollback this change.

cmonfortep (approved):
LGTM. Tested and all works as described. GJ

PR #1205, opened 2021-04-27
UOA experiment 2nd iteration

marcosholgado:
Task/Issue URL: https://app.asana.com/0/inbox/1125189844075764/1200252649107473/1200229532781040
Tech Design URL: 
CC: 

**Description**:
This PR disabled the previous experiment and creates two new variants. The experimental variant now enables all the inactive notification as long as the don't clash with the UOA notification. If that's the case we delay the inactive notification by one day.

**Before testing**
In the `AndroidNotificationScheduler` change lines `85` and `89` so they pass `TimeUnit.SECONDS` instead of days.
 
In the `AndroidNotificationScheduler` change line 102 to `return duration * 10` to give the notification more than 1 second to be sent.

**Steps to test this PR**:
**ZY experimental variant**
Change the variant to always use `zy`

**Day 0**
1. Before installing the app change the line 39 in `AppInstallStore` to return 0
1. Install the app and launch it
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 0 day is 1`
1. `Scheduling notification for 10` -> scheduled for 1 day later
1. After 10 seconds the notification should show up
1. Dismiss the notification and bring the app to the foreground
1. Bring the app back to the foreground
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 0 day is 3`
1. `Scheduling notification for 40` -> scheduled for 4 days later
1. After 40 seconds the notification should show up

**Day 2**
1. Before installing the app change the line 39 in `AppInstallStore` to return 2
1. Install the app and launch it
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 2 day is 1`
1. `Scheduling notification for 20` -> scheduled for 2 days later
1. After 20 seconds the notification should show up
1. Dismiss the notification and bring the app to the foreground
1. Bring the app back to the foreground
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 2 day is 3`
1. `Scheduling notification for 30` -> scheduled for 3 days later
1. After 30 seconds the notification should show up

Testing any other days (step 1) should always return 10 for the privacy notification and 30 for the clear data notification since those notification cannot clash with the UOA notification out of day 0 and 2.

**ZX experimental variant**
Change the variant to always use `zx`

**Day 0**
1. Before installing the app change the line 39 in `AppInstallStore` to return 0
1. Install the app and launch it
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 0 day is 1`
1. `Scheduling notification for 10` -> scheduled for 1 day later
1. After 10 seconds the notification should show up
1. Dismiss the notification and bring the app to the foreground
1. Bring the app back to the foreground
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 0 day is 3`
1. `Scheduling notification for 30` -> scheduled for 3 days later 
1. After 40 seconds the notification should show up

**Day 2**
1. Before installing the app change the line 39 in `AppInstallStore` to return 2
1. Install the app and launch it
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 2 day is 1`
1. `Scheduling notification for 10` -> scheduled for 1 day later
1. After 20 seconds the notification should show up
1. Dismiss the notification and bring the app to the foreground
1. Bring the app back to the foreground
1. In the logcat, if you filter by `notification`you should see
1. `Inactive notification days installed is 2 day is 3`
1. `Scheduling notification for 30` -> scheduled for 3 days later
1. After 30 seconds the notification should show up

Testing any other days (step 1) should always return 10 for the privacy notification and 30 for the clear data notification since those notification cannot clash with the UOA notification out of day 0 and 2.

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
+++ b/app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
@@ -17,9 +17,12 @@
 package com.duckduckgo.app.notification
 
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
 import androidx.core.app.NotificationManagerCompat
 import androidx.work.*
+import com.duckduckgo.app.global.install.AppInstallStore
+import com.duckduckgo.app.global.install.daysInstalled
 import com.duckduckgo.app.global.plugins.worker.WorkerInjectorPlugin
 import com.duckduckgo.app.notification.db.NotificationDao
 import com.duckduckgo.app.notification.model.ClearDataNotification
@@ -48,7 +51,8 @@ class NotificationScheduler(
     private val clearDataNotification: SchedulableNotification,
     private val privacyNotification: SchedulableNotification,
     private val useOurAppNotification: SchedulableNotification,
-    private val variantManager: VariantManager
+    private val variantManager: VariantManager,
+    private val appInstallStore: AppInstallStore
 ) : AndroidNotificationScheduler {
 
     override suspend fun scheduleNextNotification() {
@@ -60,7 +64,7 @@ class NotificationScheduler(
         if (variant().hasFeature(VariantManager.VariantFeature.InAppUsage) && useOurAppNotification.canShow()) {
             val operation = scheduleUniqueNotification(
                 OneTimeWorkRequestBuilder<UseOurAppNotificationWorker>(),
-                3,
+                UOA_DURATION,
                 TimeUnit.DAYS,
                 USE_OUR_APP_WORK_REQUEST_TAG
             )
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
```diff
@@ -77,15 +81,29 @@ class NotificationScheduler(
...
             }
             else -> Timber.v("Notifications not enabled for this variant")
         }
     }
 
+    @VisibleForTesting
+    fun getDurationForInactiveNotification(day: Long): Long {
+        Timber.d("Inactive notification days installed is ${appInstallStore.daysInstalled()} day is $day")
+        var duration = day
+        if (variantHasInAppUsage() && (appInstallStore.daysInstalled() + day) == UOA_DURATION) {
```
marcosholgado, line 99: if the inactive notification clashes with the UOA notification on day 3 then add an extra day.

app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
```diff
@@ -77,15 +81,29 @@ class NotificationScheduler(
...
             (!variant().hasFeature(VariantManager.VariantFeature.RemoveDay1AndDay3Notifications) && clearDataNotification.canShow()) -> {
-                scheduleNotification(OneTimeWorkRequestBuilder<ClearDataNotificationWorker>(), 3, TimeUnit.DAYS, UNUSED_APP_WORK_REQUEST_TAG)
+                val duration = getDurationForInactiveNotification(CLEAR_DATA_DURATION)
+                scheduleNotification(OneTimeWorkRequestBuilder<ClearDataNotificationWorker>(), duration, TimeUnit.DAYS, UNUSED_APP_WORK_REQUEST_TAG)
             }
             else -> Timber.v("Notifications not enabled for this variant")
         }
     }
 
+    @VisibleForTesting
```
cmonfortep, line 95: I'm ok keeping this for the experiment, but it will be nice if we could refactor this so we don't need `@VisibleForTesting` -> a) test this class in a different way or b) if we want to test this method, we should extract it to a different class.

app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
```diff
@@ -146,6 +164,9 @@ class NotificationScheduler(
     companion object {
         const val UNUSED_APP_WORK_REQUEST_TAG = "com.duckduckgo.notification.schedule"
         const val USE_OUR_APP_WORK_REQUEST_TAG = "com.duckduckgo.notification.useOurApp"
+        const val UOA_DURATION = 3L
```
cmonfortep, line 167: NIT: `DURATION` was confusing to me, this made me think about something like "for how long is going to be active something".

It's clear when you read this line: `.setInitialDelay(duration, unit)`, but not in other scenarios where I don't have the context of what duration means. I propose to rename it to include `Delay`, so we have the context for that val.

cmonfortep (approved):
Minor comments. GJ

Tested and works as described.

app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
marcosholgado: Yeah I was trying to test that by getting the job, but it turns out we can only get the WorkInfo which doesn't include that information at all. 

I think if we go ahead with the experiment all this stuff could be moved into the `canShow()` method of each notification. I also wasn't too happy about the DURATION constants in this class and I believe they should be in each notification class. Will refactor all this if we end up releasing.

app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
marcosholgado: 👍  renamed to `UOA_DELAY_DURATION_IN_DAYS`

Issue #209, opened 2018-03-02
Import/Export bookmarks


IOI-655321:
You have bookmarks so please add a basic import/export to them.  Would make starting from scratch a lot smoother to be able to import from a file like a standard browser if we choose to.

RichyHBM:
Have done a first pass on this, got it exporting a basic CSV style bookmarks list

Is there any particular list type that would be preferred? If something like Json are there any limitations to using a 3rd party library?

subsymbolic:
Thanks for contributing @RichyHBM. You are welcome to submit a PR however we aren't in a position to accept many community PRs at the moment. This will hopefully change in the future, see https://github.com/duckduckgo/Android/blob/develop/CONTRIBUTING.md.

tulanevolk:
I replaced phone and would like to port my duckduckgo bookmarks from old to new. I would appreciate knowing how.

Thank you.

rlourette:
Same here

subsymbolic:
Thanks all for showing interest in import/export, we don't currently support it but do have it in our backlog. Keep 👍 it to show you want it. If the demand grows we may be able to bump the priority and get to it sooner.

jdcrutch:
Just switched from Chrome to DDG on my Adroid tablet, and was disappointed to find that I can't transfer my bookmarks.

dg-demers:
Please add the bookmark import/export function.
And until you do please tell me where the bookmark file is stored, what format it's in, and whether rooting my Android phone is necesssary to access it.

Jaason:
Actually, I have a similar question. I have opened more than >100 tabs. I want to check them, but it's easier to get through it on the desktop. Where I can find a file with opened tabs? Is there any?

kolmercm:
We need import/export features for DuckDuckGo! I have so many bookmarks I can't spend the time to transfer them one by one!

cooncreek40:
Yes, please implement a bookmark import feature. Also, set up a folder option for bookmark organization. These two issues are seriously the only items stopping my from setting DDG to my default browser.

Thank you.

kso9:
> Please add the bookmark import/export function.
> And until you do please tell me where the bookmark file is stored, what format it's in, and whether rooting my Android phone is necesssary to access it.

That would work for me too short term. Is the bookmark file readable or encrypted on Android?

alanStocco:
> > Please add the bookmark import/export function.
[...]

PR #1211, opened 2021-04-29
Support for Import / Export bookmarks

malmstein:
Task/Issue URL: https://app.asana.com/0/1174433894299346/1199665075469536

**Description**:
Some users have reported that they would like to import/export bookmarks from/to our browsers. Users have also reported that they lost all bookmarks when moving to a new device.

This PR adds support for that, focused on users that want to export their DDG bookmarks and import them in a new device

**Steps to test this PR**:
***Happy path***
1. Add a few bookmarks
2. In the Bookmarks screen tap on the overflow menu -> Export
3. Select a location for the file
4. Delete all bookmarks
5. In the Bookmarks screen tap on the overflow menu -> Import
6. Select the previously exported file
7. Check that all bookmarks have been imported

***Error***
1. Add a few bookmarks
2. In the Bookmarks screen tap on the overflow menu -> Export
3. Select a location for the file
4. Delete all bookmarks
5. In the Bookmarks screen tap on the overflow menu -> Import
6. Select an html file that does not use the Netscape Bookmark format
7. Check that nothing has been imported

**Screenshots**:
| Import  | Export |
| ------ | ----- | 
imported|exported|

Fixes: https://github.com/duckduckgo/Android/discussions/1170

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/bookmarks/service/BookmarksParser.kt
```diff
@@ -0,0 +1,84 @@
...
+        return buildString {
+            appendLine("<!DOCTYPE NETSCAPE-Bookmark-file-1>")
+            appendLine("<!--This is an automatically generated file.")
+            appendLine("It will be read and overwritten.")
+            appendLine("Do Not Edit! -->")
+            appendLine("<META HTTP-EQUIV=\"Content-Type\" CONTENT=\"text/html; charset=UTF-8\">")
+            appendLine("<Title>Bookmarks</Title>")
+            appendLine("<H1>Bookmarks</H1>")
+            appendLine("<DL><p>")
+            appendLine("    <DT><H3 ADD_DATE=\"1618844074\" LAST_MODIFIED=\"1618844074\" PERSONAL_TOOLBAR_FOLDER=\"true\">DuckDuckGo</H3>")
```
malmstein, line 46: because we don't store when a bookmark was added, we hardcode a value so we satisfy the [Netscape Bookmark File Format](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/aa753582(v=vs.85)?redirectedfrom=MSDN)

app/src/main/java/com/duckduckgo/app/bookmarks/service/BookmarksParser.kt
```diff
@@ -0,0 +1,84 @@
...
+            appendLine("    <DL><p>")
+            bookmarks.forEach { entity ->
+                appendLine("        <DT><A HREF=\"${entity.url}\" ADD_DATE=\"1618844074\" LAST_MODIFIED=\"1618844074\">${entity.title}</A>")
```
malmstein, line 49: ditto

app/src/main/res/layout/view_bookmark_entry.xml
```diff
@@ -22,7 +22,7 @@
     android:paddingTop="14dp"
     android:paddingBottom="14dp"
     android:paddingStart="16dp"
-    android:paddingEnd="16dp"
+    android:paddingEnd="10dp"
```
malmstein, line 25: change suggested in the Product Review to align the overflow menu of the toolbar with the overflow menu of each bookmark item

app/build.gradle
```diff
@@ -131,6 +131,8 @@ dependencies {
         exclude group: "org.jetbrains.kotlinx", module: "kotlinx-coroutines-debug"
     }
 
+    implementation 'org.jsoup:jsoup:1.13.1'
```
aitorvs, line 134: Is it possible to add this to refreshVersions?

app/src/main/java/com/duckduckgo/app/bookmarks/db/BookmarksDao.kt
```diff
@@ -29,6 +29,9 @@ interface BookmarksDao {
     @Query("select * from bookmarks")
     fun bookmarks(): LiveData<List<BookmarkEntity>>
 
+    @Query("select * from bookmarks")
+    fun bookmarksSync(): List<BookmarkEntity>
```
aitorvs, line 33: nit about naming: I get the `sync` means synchronous but it took me scanning thru the file to get that and only realised when I saw the other `bookmarks` method.
Could we call them `getBookmarks` and `getBookmarksSync` instead?

app/src/main/java/com/duckduckgo/app/bookmarks/di/BookmarksModule.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.duckduckgo.app.bookmarks.service.DuckDuckGoBookmarksExporter
+import com.duckduckgo.app.bookmarks.service.DuckDuckGoBookmarksImporter
+import com.duckduckgo.app.bookmarks.service.DuckDuckGoBookmarksParser
+import com.duckduckgo.app.global.DispatcherProvider
+import dagger.Module
+import dagger.Provides
+import javax.inject.Singleton
+
+@Module
+class BookmarksModule {
```
aitorvs, line 35: I think it is better to add a `ContributesTo(AppObjectGraph::class)` to avoid always touching the `AppComponent`

aitorvs, line 35: I think it is better to add a `ContributesTo(AppObjectGraph::class)` to avoid always touching the `AppComponent`

app/src/main/java/com/duckduckgo/app/bookmarks/service/BookmarkManager.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.net.Uri
+
+data class Bookmark(val title: String, val url: String)
+
+interface BookmarkManager {
+    suspend fun import(uri: Uri): ImportBookmarksResult
+    suspend fun export(uri: Uri): ExportBookmarksResult
+}
+
+class DuckDuckGoBookmarkManager constructor(
```
aitorvs, line 28: nit: can we call it `RealBookmarkManager` to follow the convention we've got in most places?

aitorvs, line 28: nit: can we call it `RealBookmarkManager` to follow the convention we've got in most places?
Also we don't need the `constructor`

app/src/main/java/com/duckduckgo/app/bookmarks/service/BookmarksImporter.kt
```diff
@@ -0,0 +1,59 @@
...
+}
+
+class DuckDuckGoBookmarksImporter(
+    private val contentResolver: ContentResolver,
+    private val dao: BookmarksDao,
+    private val bookmarksParser: BookmarksParser
+) : BookmarksImporter {
+
+    companion object {
+        private const val CHARSET = "UTF-8"
```
aitorvs, line 41: We can remove this `const`. We already have a `Charsets.UTF_8` that we can use in kotlin stdlib

app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksActivity.kt
```diff
@@ -76,6 +83,24 @@ class BookmarksActivity : DuckDuckGoActivity() {
...
+        super.onActivityResult(requestCode, resultCode, data)
+
+        if (requestCode == IMPORT_BOOKMARKS_REQUEST_CODE && resultCode == RESULT_OK) {
+            val selectedFile = data?.data
+            if (selectedFile != null) {
+                viewModel.importBookmarks(selectedFile)
+            }
+        }
+
+        if (requestCode == EXPORT_BOOKMARKS_REQUEST_CODE && resultCode == RESULT_OK) {
```
aitorvs, line 96: nit: I think it is better to use an `else` statement here. Else we force the reader of the code to understand if both `if`s can be executed.

app/src/main/java/com/duckduckgo/app/di/AppComponent.kt
```diff
@@ -79,7 +80,8 @@ import javax.inject.Singleton
         CertificateTrustedStoreModule::class,
         WelcomePageModule::class,
         HttpsPersisterModule::class,
-        EmailModule::class
+        EmailModule::class,
+        BookmarksModule::class,
```
aitorvs, line 84: Ditto, I'd suggest to remove this in favor of using `@ContributesTo(AppObjectGraph::class)` right in the definition of the dagger module

aitorvs:
@malmstein I left some suggestions

app/src/main/java/com/duckduckgo/app/di/AppComponent.kt
cmonfortep, line 84: I think we should start using `ContributesTo(AppObjectGraph::class)` in the module.

cmonfortep (approved):
So happy this is happening, GJ! everything well structured, very nice code!

I left some minor comments.

Side note about test files, it would be nice if we could use the concrete classname in the filename instead of the interface name, that way it will be easier to find the existing test from the IDE.

Testing:
 - Found an issue with the Search icon in the menu (bookmarks activity). Search appears to me even with less than 3 bookmarks (it should not appear in that case), but filtering only works when I have more then 3 (this is how it needs to be).
- import/export works as described

app/src/main/java/com/duckduckgo/app/bookmarks/db/BookmarksDao.kt
malmstein: I like that!

app/src/main/java/com/duckduckgo/app/bookmarks/service/BookmarkManager.kt
malmstein: makes sense, I'm changing all the other ones too.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksActivity.kt
```diff
@@ -100,21 +125,91 @@ class BookmarksActivity : DuckDuckGoActivity() {
...
+            }
+        }
+        return super.onOptionsItemSelected(item)
     }
 
     override fun onPrepareOptionsMenu(menu: Menu?): Boolean {
-        menu?.findItem(action_search)?.isVisible = viewModel.viewState.value?.enableSearch == true
+        val searchView = menu?.findItem(action_search)?.actionView as SearchView
+        searchView.setOnQueryTextListener(BookmarksEntityQueryListener(viewModel.viewState.value?.bookmarks, adapter))
+        searchView.isVisible = viewModel.viewState.value?.enableSearch == true
```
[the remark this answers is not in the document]
malmstein: this fixes the glitch you saw when the loupe was showing when it was not supposed to. good catch @cmonfortep !

malmstein:
Thanks @cmonfortep and @aitorvs !

oldphones68:
Although this issue is merged, but in the latest update (5.85.3), I'm not seeing any option to ExIm of bookmarks.
When it's supposed to go live?
@cmonfortep

RobertMStahl:
Please resolve this. Please, please, please.

koyae:
Since this was merged, would it be possible to modify #209 and mark it as **Resolved** rather than **Closed**? That issue has more watchers than this PR and the fact that the issue has been addressed is not visible on the main page; instead it's buried in [the discussion](https://github.com/duckduckgo/Android/discussions/1170) where things were moved. I'm just after spending a bunch of time looking for a workaround to this issue, since I didn't click through to the discussion until just a bit ago to find this.

Regardless, I'm so glad this feature is now available!

@oldphones68 I'm running 5.86.2 (58602) on Android 5.1 and I'm seeing the option to import/export.

oldphones68:
@koyae Yes! I know. And I also posted a ___Thank you post___ in the Discussion. Please see #1248.
However, thanks for replying. Have a good day!

PR #1228, opened 2021-05-06
Move the app lifecycle observers to be plugins

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1200284402862810/f
Tech Design URL: 
CC: 

**Description**:
The `DuckDuckGoAppliation` adds all application observers manually, like so:

```kotlin
ProcessLifecycleOwner.get().lifecycle.also {
  it.addObserver(this)
  it.addObserver(dataClearer)
  it.addObserver(appDaysUsedRecorder)
  it.addObserver(defaultBrowserObserver)
  it.addObserver(appEnjoymentLifecycleObserver)
  it.addObserver(dataClearerForegroundAppRestartPixel)
  it.addObserver(userStageStore)
  it.addObserver(pixelSender)
  it.addObserver(fireFireAnimationLoader)
  it.addObserver(tabsDbSanitizer)
  it.addObserver(webViewHttpAuthStore)
}
```

The main drawback is that the appliation has to depend on all the observer types, and every time we add/remove one observer we need to touch the applicaton.

We'd like to move that to use the plugin pattern, when the appliation will just get a set of `LifecycleObserver` plugins and add them. Liek so:

```kotlin
ProcessLifecycleOwner.get().lifecycle.also {
  it.addObserver(this)
  appLifecycleObserverPlugins.forEach { plugin ->
    it.addObserver(plugin)
  }
}
```

The `DuckDuckGoApplication` also manually initialised some other classes in lifecycle methods. I have also moved those to be a lifecycle observer. These are the classes:
* `appShortcutCreator.configureAppShortcuts()` was called in `onCreate` method. I moved that to the `AppShortcutCreatorLifecycleObserver` observer
* `appInstallStore` was initialised during `onCreate`. The `AppInstallStore` is also a lifecycle observer
* The `AppConfigurationSyncer` was initialised during `onCreate`. The `AppConfigurationSyncer` is also now a lifecycle observer

**Steps to test this PR**:
2. For every observer listed above, add timbers in their methods annotated with `OnLifecycleEvent`
3. open the app
4. verify the methods are executed in the correct app lifecycle event

The DuckDuckApplication logs all registered observers, this should be the list:
```
com.duckduckgo.app.fire.AutomaticDataClearer
com.duckduckgo.app.global.shortcut.AppShortcutCreatorLifecycleObserver
com.duckduckgo.app.browser.httpauth.RealWebViewHttpAuthStore
com.duckduckgo.app.fire.LottieFireAnimationLoader
com.duckduckgo.app.global.install.AppInstallSharedPreferences
com.duckduckgo.app.global.rating.AppEnjoymentAppCreationObserver 
com.duckduckgo.app.browser.defaultbrowsing.DefaultBrowserObserver
com.duckduckgo.app.job.AppConfigurationSyncer
com.duckduckgo.app.statistics.api.RxPixelSender
com.duckduckgo.app.onboarding.store.AppUserStageStore
com.duckduckgo.app.usage.app.AppDaysUsedRecorder
com.duckduckgo.app.tabs.db.TabsDbSanitizer
com.duckduckgo.app.fire.DataClearerForegroundAppRestartPixel
```

1. instal/upgrade from this branch
2. ensure the core features of the app work as usual, bookmarks, fireproof sites, fire button, browsing

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
@@ -21,17 +21,11 @@ import android.webkit.WebViewDatabase
 import androidx.room.Room
 import androidx.room.RoomDatabase
 import com.duckduckgo.app.browser.addtohome.AddToHomeCapabilityDetector
-import com.duckduckgo.app.browser.httpauth.RealWebViewHttpAuthStore
-import com.duckduckgo.app.browser.httpauth.WebViewHttpAuthStore
-import com.duckduckgo.app.fire.DatabaseCleaner
-import com.duckduckgo.app.fire.DatabaseLocator
-import com.duckduckgo.app.global.DispatcherProvider
 import com.duckduckgo.app.global.db.AppDatabase
 import com.duckduckgo.app.global.db.MigrationsProvider
 import com.duckduckgo.app.settings.db.SettingsDataStore
 import dagger.Module
 import dagger.Provides
-import javax.inject.Named
 import javax.inject.Singleton
 
 @Module(includes = [DaoModule::class])
@@ -39,13 +33,8 @@ class DatabaseModule {
 
     @Provides
     @Singleton
-    fun provideWebViewHttpAuthStore(
-        context: Context,
-        databaseCleaner: DatabaseCleaner,
-        @Named("authDbLocator") authDatabaseLocator: DatabaseLocator,
-        dispatcherProvider: DispatcherProvider
-    ): WebViewHttpAuthStore {
-        return RealWebViewHttpAuthStore(WebViewDatabase.getInstance(context), databaseCleaner, authDatabaseLocator, dispatcherProvider)
+    fun provideWebviewDatabase(context: Context): WebViewDatabase {
+        return WebViewDatabase.getInstance(context)
     }
 
     @Provides
--- a/app/src/main/java/com/duckduckgo/app/global/shortcut/AppShortcutCreator.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/shortcut/AppShortcutCreator.kt
@@ -23,17 +23,52 @@ import android.content.pm.ShortcutInfo
 import android.content.pm.ShortcutManager
 import android.os.Build
 import androidx.annotation.RequiresApi
+import androidx.annotation.UiThread
 import androidx.core.content.pm.ShortcutInfoCompat
 import androidx.core.graphics.drawable.IconCompat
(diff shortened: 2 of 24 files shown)


app/src/main/java/com/duckduckgo/app/di/DatabaseModule.kt
```diff
@@ -21,31 +21,20 @@ import android.webkit.WebViewDatabase
...
 
     @Provides
     @Singleton
-    fun provideWebViewHttpAuthStore(
-        context: Context,
-        databaseCleaner: DatabaseCleaner,
-        @Named("authDbLocator") authDatabaseLocator: DatabaseLocator,
-        dispatcherProvider: DispatcherProvider
-    ): WebViewHttpAuthStore {
-        return RealWebViewHttpAuthStore(WebViewDatabase.getInstance(context), databaseCleaner, authDatabaseLocator, dispatcherProvider)
```
aitorvs, line 48: I changed the approach to provide stubs for the `DispatcherProvider`, as it was way simpler to do it that way. You'll see I added a `StubCoroutineModule` that does that.
In that way, the `RealWebViewHttpAuthStore` is always taken from the real dagger module, and when in test, it takes the stub dispatcher provider

app/src/main/java/com/duckduckgo/app/global/shortcut/AppShortcutCreator.kt
```diff
@@ -23,17 +23,52 @@ import android.content.pm.ShortcutInfo
...
+) : LifecycleObserver {
+    @UiThread
+    @OnLifecycleEvent(Lifecycle.Event.ON_CREATE)
+    fun configureAppShortcuts() {
+        Timber.i("Configure app shortcuts")
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+            appShortcutCreator.configureAppShortcuts()
+        }
+    }
+}
```
aitorvs, line 65: This class is created because `AppShortcutCreator` accesses types that are not available in all APIs, ie. `ShortcutInfo`.
Android would throw with an error message saying `Lifecycles cannot access even other methods so you should make sure that your observer classes only access framework classes that are available in your min API level OR use lifecycle:compiler annotation processor.`
Rather than adding the suggested new annotation processor, we just create a new class, `AppShortcutCreatorLifecycleObserver`, that is the lifecycle observer, that does not access any non supported method.
This is basically what the suggested annotation processor does, creating a Proxy class.

marcosholgado (approved):
LGTM! 🚢  🇮🇹

PR #1234, opened 2021-05-12
Moved ViewModelFactory.kt into common gradle module

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1200319078848279/f
Tech Design URL: 
CC: 

**Description**:
Moved the VM factory into the `common` gradle module so that it can be used from other gradle modules.

See https://app.asana.com/0/414730916066338/1200319078848279/f for full context

**Steps to test this PR**:
This change is guarded by the compiler.

---
###### Internal references:

### DIFF
--- a/common/src/main/java/com/duckduckgo/app/global/ViewModelFactory.kt
+++ b/common/src/main/java/com/duckduckgo/app/global/ViewModelFactory.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 DuckDuckGo
+ * Copyright (c) 2021 DuckDuckGo
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/duckduckgo/app/global/ViewModelFactory.kt
```diff
@@ -20,7 +20,9 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import com.duckduckgo.app.global.plugins.view_model.ViewModelFactoryPluginPoint
 import javax.inject.Inject
+import javax.inject.Singleton
 
+@Singleton
```
aitorvs, line 25: This `Singleton` does not affect the behavior.
It should have been `singleton` at the time as factory should only be one but it was fine because the `ViewModelFactoryPluginPoint` is also `singleton`

cmonfortep (approved):
LGTM In the compiler I trust

PR #1247, opened 2021-05-20
Add nav parameter to autocomplete

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1200343978632902
Tech Design URL: 
CC: 

**Description**:
This PR introduces the new autocomplete parameter `is_nav`. The extra `isNav=true` field will then be used as a navigation query (meaning when the user clicks that autocomplete suggestion we load that URL). Results where this is `false` should be treated as searches (even if they look like a URL).

This PR also handles the case where `nav=null` in which case we default to the old way.

**Steps to test this PR**:
**Previous steps, setting up Charles**
1. Connect your device to charles and use the `autocomplete.json` response provided in the task.
1. Install and launch the app.
1. Add a bookmark to `http://m.youtube.com`
1. Do a search (i.e. `you`) and in Charles map the response to the `autocomplete.json` file.

**Test cases**
1. The bookmark result should take you to the URL.
1. The first autocomplete result has `isNav=false`, clicking on it should do a search.
1. The second autocomplete result has `isNav=false` and is a URL, clicking on it should do a search.
1. The third autocomplete result has `isNav=true`, clicking on it should load the URL.
1. The fourth autocomplete result doesn't have the nav attribute and is a URL, clicking on it should load the URL.
1. The fifth autocomplete result doesn't have the nav attribute and is a search term, clicking on it should do a search.
---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -971,8 +971,8 @@ class BrowserTabFragment :
             viewModel.fireAutocompletePixel(suggestion)
             withContext(Dispatchers.Main) {
                 val origin = when (suggestion) {
-                    is AutoCompleteBookmarkSuggestion -> FromAutocomplete(nav = null)
-                    is AutoCompleteSearchSuggestion -> FromAutocomplete(nav = suggestion.isUrl)
+                    is AutoCompleteBookmarkSuggestion -> FromAutocomplete(isNav = null)
+                    is AutoCompleteSearchSuggestion -> FromAutocomplete(isNav = suggestion.isUrl)
                 }
                 viewModel.onUserSubmittedQuery(suggestion.phrase, origin)
             }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -79,6 +79,7 @@ import com.duckduckgo.app.browser.model.BasicAuthenticationRequest
 import com.duckduckgo.app.browser.model.LongPressTarget
 import com.duckduckgo.app.browser.omnibar.KeyboardAwareEditText
 import com.duckduckgo.app.browser.omnibar.OmnibarScrolling
+import com.duckduckgo.app.browser.omnibar.QueryOrigin
 import com.duckduckgo.app.browser.session.WebViewSessionStorage
 import com.duckduckgo.app.browser.shortcut.ShortcutBuilder
 import com.duckduckgo.app.browser.tabpreview.WebViewPreviewGenerator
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -968,7 +969,11 @@ class BrowserTabFragment :
         GlobalScope.launch {
             viewModel.fireAutocompletePixel(suggestion)
             withContext(Dispatchers.Main) {
-                viewModel.onUserSubmittedQuery(suggestion.phrase)
+                val origin = when (suggestion) {
+                    is AutoCompleteSuggestion.AutoCompleteBookmarkSuggestion -> QueryOrigin.FromAutocomplete(nav = null)
+                    is AutoCompleteSuggestion.AutoCompleteSearchSuggestion -> QueryOrigin.FromAutocomplete(nav = suggestion.isUrl)
```
CDRussell, line 974: static imports would make this code a little more readable here

app/src/main/java/com/duckduckgo/app/browser/omnibar/QueryUrlConverter.kt
```diff
@@ -28,8 +28,13 @@ import javax.inject.Inject
 
 class QueryUrlConverter @Inject constructor(private val requestRewriter: RequestRewriter) : OmnibarEntryConverter {
 
-    override fun convertQueryToUrl(searchQuery: String, vertical: String?): String {
-        if (UriString.isWebUrl(searchQuery)) {
+    override fun convertQueryToUrl(searchQuery: String, vertical: String?, queryOrigin: QueryOrigin): String {
+        val isUrl = when (queryOrigin) {
+            is QueryOrigin.FromAutocomplete -> queryOrigin.nav ?: UriString.isWebUrl(searchQuery)
```
CDRussell, line 33: We have this logic twice: `queryOrigin.nav ?: UriString.isWebUrl(searchQuery)`; here and in `getAutoCompleteSearchResults`

I haven't thought about it too much here, but it feels like we'd able to have this logic happen only once and by the time we get to here we wouldn't have to do it again.

is not something to block on, and it might be necessary after all, so just posing the question if both are needed

CDRussell (approved):
Tested and working well!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: 👍

app/src/main/java/com/duckduckgo/app/browser/omnibar/QueryUrlConverter.kt
marcosholgado: That is correct! I've removed that code and tested it, everything works as expected.

CDRussell (approved):

PR #1261, opened 2021-05-31
Replace LiveData with StateFlow and SharedFlow

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1200346947958650/f
Tech Design URL: 
CC: 

**Description**:
Moved the `SettingsActivity` and its view model from using `LiveData` to `StateFlow` and `SharedFlow`.
* view state uses `StateFlow`
* commands use `SharedFlow`

I also update `kotlinx.coroutines` from 1.4.2 to 1.5.0, why?
* `1.4.3` fixes a bunch of bugs
* `1.5.0` deprecates `BroadcastChannel` which should be the replace for `command`s `LiveData` and recomments using `SharedFlow` (see [here](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/))
* `1.5.0` removes makes some of the experimental APIs stable, which means we can remove some of the `ExperimentalCoroutinesApi` annotations

I also added a test dependency for testing, turbine, why?
* testing flows is generally a pain and produces hard-to-read and verbose code
* turbine makes testing flows a lot easier with a functional API
* it is a test dependency, so it does not risk prod app
* for more context see also [this task](https://app.asana.com/0/1200233033441830/1199582721357326/f)

After this PR, we should also:
* Change any `ConflatedBroadcastChannel` to `StateFlow`
* Change any remaining `BroadcastChannel` to `SharedFlow`

**Steps to test this PR**:
2. go to settings
3. test settings features

---
###### Internal references:


app/src/androidTest/java/com/duckduckgo/app/settings/SettingsViewModelTest.kt
```diff
@@ -147,13 +162,15 @@ class SettingsViewModelTest {
...
+        testee.commands().test {
+            testee.onLightThemeToggled(false)
+            verify(mockAppSettingsDataStore).theme = DuckDuckGoTheme.DARK
+
+            assertEquals(Command.UpdateTheme, expectItem())
 
-        testee.command.blockingObserve()
-        verify(commandObserver, atLeastOnce()).onChanged(commandCaptor.capture())
-        assertEquals(Command.UpdateTheme, commandCaptor.firstValue)
+            cancelAndConsumeRemainingEvents()
```
malmstein, line 172: would it make sense to extract this somehow? seems easy to forget and it's used in every single test

versions.properties
```diff
@@ -5,14 +5,14 @@
 ## Generated by $ ./gradlew refreshVersions
 ## Please, don't put extra comments in that file yet, keeping them is not supported yet.
 
-version.androidx.appcompat=1.2.0
+version.androidx.appcompat=1.3.0
```
malmstein, line 8: did you mean to update appCompat here?

malmstein (commented):
I like how this looks. One thing I'm not sold is the use of backing properties or not. We use backing properties in some places and in other we don't. It would be good to be consistent and decide which way to go.


Maybe another topic for tomorrow?

Other than that, good job @aitorvs !

aitorvs:
> I like how this looks. One thing I'm not sold is the use of backing properties or not. We use backing properties in some places and in other we don't. It would be good to be consistent and decide which way to go.
> 
> Maybe another topic for tomorrow?
> 
> Other than that, good job @aitorvs !

Yeah, I don't have strong opinions tbh. I profoundly hate the `_` notation for backing properties.
Also, backing properties make more difficult to do mods in the flows that should be private.

versions.properties
aitorvs: I had updated it because of a reason that I can't remember now. So I will revert it.

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
```diff
@@ -82,17 +83,17 @@ class SettingsViewModel @Inject constructor(
...
+        data class ShowClearWhenDialog(val option: ClearWhenOption) : Command()
     }
 
-    val viewState: MutableLiveData<ViewState> = MutableLiveData<ViewState>().apply {
-        value = ViewState()
-    }
+    private val viewState = MutableStateFlow(ViewState())
 
-    val command: SingleLiveEvent<Command> = SingleLiveEvent()
+    private val command = MutableSharedFlow<Command>()
```
marcosholgado, line 96: Not sure `SharedFlow` is the correct thing to use for commands. `SharedFlow` is hot so you could lose commands when there isn't an observer like during a configuration change.

aitorvs: I think for command specifically we want the semantics to be those of share flow. Say a command is a toast or a dialog. We wouldn't want to show the toast every time we eg. rotate the screen.

malmstein: We are currently using these commands with `SingleLiveEvent` @marcosholgado. Otherwise we'd be using a normal `LiveData` for it. Personally, I'm ok with not showing the Toast again after a configuration change, specially because configuration changes in our app are very very rare.

malmstein:
> > I like how this looks. One thing I'm not sold is the use of backing properties or not. We use backing properties in some places and in other we don't. It would be good to be consistent and decide which way to go.
> > Maybe another topic for tomorrow?
> > Other than that, good job @aitorvs !
> 
> Yeah, I don't have strong opinions tbh. I profoundly hate the `_` notation for backing properties.
> Also, backing properties make more difficult to do mods in the flows that should be private.

Agree on the naming issue, I'm sure we can make it better.
I'm not sure I follow why is that an issue? The backing property reflects what the private val has, how is it different if we don't use the backing property?

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
marcosholgado: Yeah, it depends on the scenario really. For user commands it is ok but we also use commands for other things not directly related with a user action (and we probably shouldn't). I had a chat with @aitorvs earlier and we decided to go with a `Channel<Command>(capacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)` instead.

marcosholgado (approved):
LGTM! We can talk about backing properties in our weekly call and change this later if needed.

PR #1286, opened 2021-06-18
Fix/david/image download

malmstein:
Task/Issue URL: https://app.asana.com/0/715106103902962/1200473921845317

**Description**:
Currently we've got an average of 90k occurrences per day where we don't properly detect the file type and so a generic .bin extension is given, making not possible for the user to locate and/or open the file

This PR solves the issue of detecting the content-type of the file by making a HEAD request to the file URL and extracting the content-type and content-disposition if present

**Steps to test this PR**:
1. Make an image query to SERP
2. Download the image
3. You'll download a png / jpg image instead of a .bin file

---
###### Internal references:


app/src/main/java/com/duckduckgo/app/browser/downloader/DownloadFileService.kt
```diff
@@ -0,0 +1,31 @@
...
+import retrofit2.Call
+import retrofit2.Response
+import retrofit2.http.GET
+import retrofit2.http.HEAD
+import retrofit2.http.Url
+
+interface DownloadFileService {
+
+    @HEAD
+    fun downloadFile(@Url urlString: String): Call<Void>?
```
aitorvs, line 29: nit: we're not downloading the file, should we call it differently?

app/src/main/java/com/duckduckgo/app/browser/downloader/FilenameExtractor.kt
```diff
@@ -74,13 +75,13 @@ class FilenameExtractor @Inject constructor(
         return guessedFilename
     }
 
-    private fun bestGuess(guesses: Guesses): String {
+    private fun bestGuess(guesses: Guesses): FilenameExtractionResult {
         val guess = guesses.bestGuess ?: guesses.latestGuess
         if (!guess.contains(".")) {
             pixel.fire(AppPixelName.DOWNLOAD_FILE_DEFAULT_GUESSED_NAME)
-            return guess + DEFAULT_FILE_TYPE
+            return FilenameExtractionResult.Failure(guess)
```
aitorvs, line 82: I think this is a bit misleading. Normally, when modelling `Success` or `Failure`, the `Failure` contains the reason of failure. Here it contains a default value.
Why did we need to change the return type from `String` with the best guess, to a `Success` / `Failure` that anyway contains the best guess?

app/src/main/java/com/duckduckgo/app/browser/downloader/NetworkFileDownloader.kt
```diff
@@ -34,8 +39,58 @@ class NetworkFileDownloader @Inject constructor(private val context: Context, pr
...
+                    var updatedPendingDownload = pendingDownload.copy()
+                    if (response.headers().get("content-type") != null) {
+                        val mimeType = response.headers().get("content-type")
+                        updatedPendingDownload = updatedPendingDownload.copy(mimeType = mimeType)
+                    }
+
+                    if (response.headers().get("content-disposition") != null) {
+                        val contentDisposition = response.headers().get("content-disposition")
+                        updatedPendingDownload = updatedPendingDownload.copy(contentDisposition = contentDisposition)
+                    }
```
aitorvs, line 63:
```suggestion
                    var updatedPendingDownload = pendingDownload.copy(
                      mimeType = response.headers().get("content-type"),
                      contentDisposition = response.headers().get("content-disposition")
                     )
```

No need to make so many copies, `mimeType` and `contentDisposition` are already nullable.

app/src/main/java/com/duckduckgo/app/browser/downloader/NetworkFileDownloader.kt
```diff
@@ -34,8 +39,58 @@ class NetworkFileDownloader @Inject constructor(private val context: Context, pr
...
+
+                    when (val extractionResult = filenameExtractor.extract(updatedPendingDownload)) {
+                        is FilenameExtractor.FilenameExtractionResult.Success -> downloadFile(updatedPendingDownload, extractionResult.filename, callback)
+                        is FilenameExtractor.FilenameExtractionResult.Failure -> {
+                            downloadFile(pendingDownload, extractionResult.bestGuess, callback)
+                            callback.downloadFailed(context.getString(R.string.downloadManagerDisabled), DownloadFailReason.DownloadManagerDisabled)
+                        }
+                    }
+                } else {
+                    Timber.d("Connection failed ${response.errorBody()}")
```
aitorvs, line 73: Should we not call `callback.downloadFailed(...)` here as well?

initpwn (approved):

app/src/main/java/com/duckduckgo/app/browser/downloader/NetworkFileDownloader.kt
```diff
@@ -34,8 +38,50 @@ class NetworkFileDownloader @Inject constructor(private val context: Context, pr
             return
         }
 
-        val guessedFileName = filenameExtractor.extract(pendingDownload)
+        fileService.getFileDetails(pendingDownload.url)?.enqueue(object : Callback<Void> {
+            override fun onResponse(call: Call<Void>, response: Response<Void>) {
+                if (response.isSuccessful) {
+                    var updatedPendingDownload = pendingDownload.copy()
 
+                    if (response.headers().get("content-type") != null) {
```
marcosholgado, line 46: Following @aitorvs's comment, not sure if we can avoid doing multiple copies but it would be nice.

Is there a case where any of the headers could come back as null but `PendingFileDownload` already had a `mimeType` or `contentDisposition` value set? If the answer is no we can set those two parameters even though they come back as null. If we don't know the answer or we think is a possible edge case, we can keep this as it is.

marcosholgado (approved):
Nice job! I added a comment but feel free to skip.

app/src/main/java/com/duckduckgo/app/browser/downloader/NetworkFileDownloader.kt
malmstein: Not anymore, we use this call to populate the `mimeType` and `contentDisposition` of  `PendingFileDownload`. If the headers come back as null, DownloadManager won't be able to download the file, therefore we will eventually call `callback.downloadFailed()`

malmstein:
Thanks @aitorvs  @marcosholgado

PR #1308, opened 2021-07-08
Create internal flavor

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1200572408644661/f
Tech Design URL: 
CC: 

**Description**:
We want to add an `internal` production flavor that will be used to release the app for internal users. The main use-case for the this `internal` flavor is to host some features that serve no purpose in production, eg. debug/dev panel

While doing that, I have also cleaned-up a bit the `play` and `fdroid` flavors, more details below.

**Changes in this PR**
* add the `internal` flavor to the `app/build.gradle`
* `internal` flavor only has `release` version, similar to what we have for `fdroid`
 * If we see this causes friction while developing we will enable the `internalDebug` as well
* `internal` flavor does not contain anything yet
* Tidy up for `play` and `fdroid` flavors
 * both flavors were providing the same `AppInstallationReferrerParser` and `AppReferrerDataStore` implementations, so I moved those providers out of the `StoreReferralModule`
 * as it was, every flavor had to provide its own `StoreReferralModule` and `HttpsPersisterModule` modules, even if the flavor was not concern with those. This would have been the case for the `internal` module
 * what we do now is to provide default implementations in `main` and any flavor that wants to override the default behavior can do so using DI. This is why I ahve:
  * moved the F-Droid `HttpsPersisterModule` and the `StoreReferralModule` into `main` (this will serve as default impls)
  * removed the F-Droid `HttpsPersisterModule` and the `StoreReferralModule` to the main app
  * `play` flavor will override those to DI modules by its own `PlayHttpsPersisterModule` and `PlayStoreReferralModule`
  * the reason we used the F-Droid as default implementation is because F-Droid APK can't contain the `bloom` binary file nor the `com.android.installreferrer` library

**Steps to test this PR**:

Test Play flavor
1. in this branch, optionally place some timber statements `PlayStoreAppReferrerStateListener` and `PlayHttpsEmbeddedDataPersister`
1. install the playDebug from this branch
1. launch the app
1. verify the `PlayHttpsEmbeddedDataPersister` and `PlayStoreAppReferrerStateListener` are called respectively

Test Fdroid/Internal flabors
1. install the fdroid/internal flavors
1. launch the app
1. verify `EmptyReferrerStateListener` and `EmptyHttpsEmbeddedDataPersister` are called instead

---
###### Internal references:


CDRussell:
Is this still a draft @aitorvs ?

aitorvs:
> Is this still a draft @aitorvs ?

shoot I forgot, now it's ready

app/build.gradle
```diff
@@ -84,8 +87,8 @@ android {
     }
     variantFilter { variant ->
         def names = variant.flavors*.name
-        if (names.contains("fdroid") && variant.buildType.name == "debug") {
-            // We only need the F-Droid release version
+        if ((names.contains("fdroid") || names.contains("internal")) && variant.buildType.name == "debug") {
+            // We only need release version for F-Droid and internal builds
```
CDRussell, line 91: will add the same comment here as from the other PR; I think this phrasing could be clearer

CDRussell (approved):
Looking good to me! 👏

app/build.gradle
aitorvs: you mean comment in line 91? I tis the same comment. Just copied it across from the other PR. Maybe I missed something?

CDRussell: yeah that's clearer, thanks. Think my other suggestion was missed but it was basically what you've ended up with. 👍 

[lll
C9856D65-D05C-49A9-BE79-6E35EC1639BC

PR #1320, opened 2021-07-15
Feature/david/export dark theme

malmstein:
Task/Issue URL: https://app.asana.com/0/1174433894299346/1200458304261332

**Description**:
This PR moves the Theming preferences out of the app module. This is the first step towards the bigger refactor we have planned for themes. Besides that, it also allows Activity from other modules to use `AppTheme.Light` or `AppTheme.Dark` and respect the option that the user selected in the Settings screen.

---
###### Internal references:


common-ui/build.gradle
```diff
@@ -17,14 +17,22 @@
...
 
+    implementation project(path: ':di')
+
+    implementation Google.dagger
+
+    implementation Kotlin.stdlib.jdk7
     implementation AndroidX.appCompat
     implementation Google.android.material
     implementation AndroidX.constraintLayout
+
```
cmonfortep, line 35: Just double-checking, are all these changes related to this refactor? no need to explain why we use each library, just checking we are not including something not related.

cmonfortep (commented):
Everything looks good and works fine. I only have one question related to `SettingsDataStore` vs `ThemingSharedPreferences`, it seems we have the same method and logic inside both classes.

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -75,10 +75,14 @@ class SettingsSharedPreferences constructor(private val context: Context, privat
             }
         }
 
-    override var theme: DuckDuckGoTheme?
+    override var theme: DuckDuckGoTheme
         get() {
-            val themeName = preferences.getString(KEY_THEME, null) ?: return null
-            return DuckDuckGoTheme.valueOf(themeName)
+            val themeName = preferences.getString(KEY_THEME, null)
```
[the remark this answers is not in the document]
malmstein: good catch, this can be removed from here

common-ui/build.gradle
malmstein: yes, we need to add AppCompat to reference AppCompatActivity, etc...

cmonfortep (approved):
Ship it!

PR #1326, opened 2021-07-22
Update to autofill 3.1.1

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1200635726526715
Tech Design URL: 
CC: 

**Description**:
This PR updates autofill to 3.1.1. This new version contains a breaking change so we now also need to support email cohorts. Furthermore this PR moves the injection of the JS code to the `onPageStarted` method which simplifies the code a bit (not having to check if the JS was already injected) and prepares the app for future autofill versions.

You might need to run this command before doing anything to fetch the changes from the submodule.

`git submodule update submodules/autofill`

**Steps to test this PR**:
1. Checkout branch
1. Run `git submodule`
1. Confirm autofill commit hash is`e4a5cc4543eb350bd0da197947ccd3423c999f92 `

**Autofill still works**
1. Sign in with your duck address (this will also set your cohort)
1. After signing all the email functionalities should be there and work correctly:
1. Create Duck Address in the options menu should copy a private address in the clipboard.
1. Go to `https://duckduckgo.com/email/login`, dax should now show up in the email field.
1. Go to `github.com`, autofill should appear when clicking on Dax
1. Click on `Use yourname@duck.com`
1. A pixel should be sentm `m_e_uad` with a cohort parameter (could be `internal_beta`, `unknown`, depending on what autofill returned after signing in.
1. Click on Dax again and now on `Generate a private address`
1. A pixel should be sentm `m_e_ua` with a cohort parameter (could be `internal_beta`, `unknown`, depending on what autofill returned after signing in.
1. Click on Dax again and dismiss the tooltip
1. A pixel should be sentm `m_e_t_d` with a cohort parameter (could be `internal_beta`, `unknown`, depending on what autofill returned after signing in.
---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
+++ b/app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
@@ -44,8 +44,8 @@ class EmailJavascriptInterface(
     fun isSignedIn(): String = emailManager.isSignedIn().toString()
 
     @JavascriptInterface
-    fun storeCredentials(token: String, username: String) {
-        emailManager.storeCredentials(token, username)
+    fun storeCredentials(token: String, username: String, cohort: String) {
+        emailManager.storeCredentials(token, username, cohort)
     }
 
     @JavascriptInterface
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -3155,19 +3155,21 @@ class BrowserTabViewModelTest {
     @Test
     fun whenConsumeAliasThenPixelSent() {
         whenever(mockEmailManager.getAlias()).thenReturn("alias")
+        whenever(mockEmailManager.getCohort()).thenReturn("cohort")
 
         testee.consumeAlias()
 
-        verify(mockPixel).enqueueFire(AppPixelName.EMAIL_USE_ALIAS)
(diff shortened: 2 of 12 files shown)


app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
```diff
@@ -44,8 +44,8 @@ class EmailJavascriptInterface(
     fun isSignedIn(): String = emailManager.isSignedIn().toString()
 
     @JavascriptInterface
-    fun storeCredentials(token: String, username: String) {
-        emailManager.storeCredentials(token, username)
+    fun storeCredentials(token: String, username: String, cohort: String) {
```
cmonfortep, line 47: Just checking if we are sure that `cohort` will never be `null`.

cmonfortep (approved):
LGTM. Tested and it works as described. Just left a comment not related with this PR so feel free to skip it.

Side comment: I've noticed we have `EmailManager` which has access to `EmailDataStore` and has some logic like mapping cohort `null` to `UNKNOWN_COHORT`. I've noticed `WaitlistCodeNotification` directly accesses the datastore instead of going through the `EmailManager`. Should we change that? feel free to skip this one as it's not related to the PR.

app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
marcosholgado: Yup, already confirmed by the FE team.

PR #1349, opened 2021-08-04
Add common githooks folder to share hooks with the team

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1200701185339027/f
Tech Design URL: 
CC: 

**Description**:
This PR adds a `.githooks` folder where all commit hooks we think should be shared across the devs should be placed, and adds the first hook which runs spotless on all modules for every commit to ensure our code styling is always correct.

Unfortunatelly GIT does not allow to directly check-in the hooks in the `.git/hooks` directory. Also anything in those folders are not cloned. The rational seems to be that a clone repository is a local copy, so sharing the hooks would go against that de-centralised approach.

However, it is possible to share a folder where we can place those hooks, ie. `.githooks` and then tell git where it needs to look for them. The downside is that we are adding a config step but that should be ok.

The setup for a newcomer to the repo would be:
1. clone the repo
1. cd the repo directory
1. execute `git config core.hooksPath .githooks`

From that point forward, git will look into `.githooks` rather than `.git/hooks`

**Steps to test this PR**:
1. checkout this branch
1. run `git config core.hooksPath .githooks`
1. Go to `FeedActivity` line 141
1. Remove the space before the `{` in `supportFragmentManager.transaction {`
1. run `git commit -a -m test`
1. verify spotless runs and the commit is greated but it is empty ;)
1. Go to `AppUrl` line 19
1. Remove the space before the `{` in `class AppUrl {`
1. run `git commit -a -m test2`
1. verify spotless runs and the commit is greated but it is empty ;)

---
###### Internal references:


marcosholgado:
👋  I'm not sure this works for all the modules. Can we also test steps 3 and 4 in a different module? Like the di module to make sure it does also work there since that module doesn't apply spotless in its gradle file?

aitorvs:
> 👋 I'm not sure this works for all the modules. Can we also test steps 3 and 4 in a different module? Like the di module to make sure it does also work there since that module doesn't apply spotless in its gradle file?

arg, I had forgotten to modify the `android-library.gradle` to make sure all modules run spotless.
When commiting that change spotless ran in all modules fixing code styles in some files, as can be seen in this PR.

I will also another test step to the description

malmstein (approved):
Works as described, thanks @aitorvs and @marcosholgado!

PR #1350, opened 2021-08-04
[BUG] Fix crash when accidentally returning binding.root in onCreateView

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1200701185339109/f
Tech Design URL: 
CC: 

**Description**:
This PR fixes a crash that happened when submitting negative feedback.
In `MainReasonNegativeFeedbackFragment` and `SubReasonNegativeFeedbackFragment` we were doing some logic in `onCreateView` and then returning `binding.root`. But Android does not allow to use view binding inside `onCreateView` that should return an inflated view, which at that point it is not.

This PR moves the logic performed in the `onCreateView` method to the `onActivityCreated` method--which is the next lifecycle method to be called--so that we can continue using view binding

**Steps to test this PR**:
1. install from develop and launch the app
1. go to settings -> share feedback
1. click in sad face
1. BOOM! app crashes
1. install from this branch and launch the app
1. go to settings -> share feedback
1. click in sad face
1. verify app does not crash
1. select a sub-reason
1. verify app does not crash
1. complete the feedback and submit
1. verify app does not crash

---
###### Internal references:

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/feedback/ui/negative/mainreason/MainReasonNegativeFeedbackFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/feedback/ui/negative/mainreason/MainReasonNegativeFeedbackFragment.kt
@@ -17,9 +17,6 @@
 package com.duckduckgo.app.feedback.ui.negative.mainreason
 
 import android.os.Bundle
-import android.view.LayoutInflater
-import android.view.View
-import android.view.ViewGroup
 import androidx.core.content.ContextCompat
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.duckduckgo.app.browser.R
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/feedback/ui/negative/mainreason/MainReasonNegativeFeedbackFragment.kt
```diff
@@ -44,19 +41,15 @@ class MainReasonNegativeFeedbackFragment : FeedbackFragment(R.layout.content_fee
     private val listener: MainReasonNegativeFeedbackListener?
         get() = activity as MainReasonNegativeFeedbackListener
 
-    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+
         recyclerAdapter = MainReasonAdapter(object : (FeedbackTypeMainReasonDisplay) -> Unit {
```
cmonfortep, line 47: We could even create the adapter using `by lazy` directly in the `var` declaration if we like more that style. In any case 👍

cmonfortep (approved):
LGTM 👍  All good! Thanks for fixing it so quick.

PR #1387, opened 2021-08-18
Remote Config (II): add local file

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1200800571610917
Tech Design URL: 
CC: 

**Description**:
This PR adds a local version of the remote config to be used if we cannot use the remote endpoint.

**Steps to test this PR**:
Note: As of today the remote config and the local file are the same and share the same version.

**Local file persists**
1. In the local `privacy_config.json` file change the version to `1628840233004`
1. Install and launch the app
1. Check the `privacy_config` database and in the `privacy_config` table the version value should be `1628840233004`

**Higher version always win**
1. In the local `privacy_config.json` file change the version to `1628840233002`
1. Install and launch the app
1. Check the `privacy_config` database and in the `privacy_config` table the version value should be `1628840233003` (remote config prevails over local)
1. Do not uninstall the app and continue to the next test case

**eTag properly used**
1. Set up charles to your device and launch the app.
1. Launch the app
1. In Charles notice the call to the endpoint returns a 304 html code (not modified)

**More tests**
1. Rollback your previous changes to the local `remote_config.json` file and clear cache and data.
1. Using Charles, launch the app and intercept the call to the remote config and save it locally.
1. Check the `privacy_config` database and in the `privacy_config` table the version value should be `1628840233003` 
1. Change the saved file version to return a higher version `1628840233004` and map the response in Charles.
1. Relaunch the app.
1. Check the `privacy_config` database and in the `privacy_config` table the version value should be `1628840233004` 

---
###### Internal references:


privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/observers/LocalPrivacyConfigObserver.kt
```diff
@@ -0,0 +1,69 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@WorkerThread
+@Singleton
+@ContributesMultibinding(AppObjectGraph::class)
+class LocalPrivacyConfigObserver @Inject constructor(
+    private val context: Context,
+    private val privacyConfigPersister: PrivacyConfigPersister,
+    @PrivacyConfigCoroutineScope val coroutineScope: CoroutineScope,
```
aitorvs, line 44: I think we can safely remove the `PrivacyConfigCoroutineScope` here. I don't see any `couroutineScope` provider dep that is annotated with it. So presumably we're just getting the `applicationCoroutineScope`
Maybe what you really want is to annotate with `@AppCoroutineScope` here?

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/di/PrivacyConfigCoroutineScope.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import javax.inject.Scope
+
+/**
+ * Identifies a coroutine scope type that is scope to the Privacy Config lifecycle
+ */
+@Scope
+@MustBeDocumented
+@Retention(AnnotationRetention.RUNTIME)
+annotation class PrivacyConfigCoroutineScope
```
aitorvs, line 27: I think we can remove this one. See comment below

aitorvs (approved):
This is great!
Added a comment because I think we can remove `PrivacyConfigCoroutineScope` and use `AppCoroutineScope` instead

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/di/PrivacyConfigCoroutineScope.kt
marcosholgado: 👍

PR #1420, opened 2021-09-10
Add fire animation preview to settings

joshliebe:
Task/Issue URL: https://app.asana.com/0/42792087274227/1200315340580172/f

### Description
This PR adds fire animation previews to the selection dialog in settings.

### Steps to test this PR

1. Go settings and tap "Fire Button animation"
2. Verify that the different options show the corresponding animation when tapped

### UI changes

https://user-images.githubusercontent.com/3471025/132840907-b9dd5ac7-b075-440c-b970-192e2ba4f597.mp4

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/settings/FireAnimationActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/settings/FireAnimationActivity.kt
@@ -29,7 +29,6 @@ import com.duckduckgo.app.settings.clear.FireAnimation
 import com.duckduckgo.mobile.android.ui.view.setAndPropagateUpFitsSystemWindows
 import com.duckduckgo.mobile.android.ui.view.show
 import com.duckduckgo.mobile.android.ui.viewbinding.viewBinding
-import kotlinx.android.synthetic.main.sheet_fire_clear_data.*
 
 class FireAnimationActivity : DuckDuckGoActivity() {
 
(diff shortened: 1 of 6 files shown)


cmonfortep:
This is super cool @joshliebe. When we implemented the new animation, we found some issues loading the json lottie file on some devices (some delays to make the animation appear). I recommend you to test how this performs in different API levels/devices

marcosholgado:
FYI I had a chat with @joshliebe about low APIs. I tested this feature in my LG3 (21) and it was quite clunky but the same happens when you use the fire button anyways so our suggestion would be to either leave as it is because there's nothing we can do about it or remove the preview for lower APIs.

app/src/main/res/values/styles.xml
```diff
@@ -572,4 +572,12 @@
         <item name="android:backgroundTint">?attr/savedSiteEditTextUnderlineColor</item>
     </style>
 
+    <style name="Theme.AppCompat.Transparent.NoActionBar" parent="Theme.AppCompat.Light.NoActionBar">
```
cmonfortep, line 575: We already have an style in `themes.xml` for transparent activities `<style name="AppTheme.Transparent">`. Any reason to have it duplicated here?

cmonfortep (approved):
LGTM, nice job. 

I've tested this on my Samsung a41 and works great. I've noticed a small delay between choosing an option and the animation appearing but I'm ok with it. Since the user is just plain with settings, it's a minor thing. Not the same as when clearing data, so all good!

app/src/main/java/com/duckduckgo/app/settings/FireAnimationActivity.kt
```diff
@@ -0,0 +1,89 @@
...
+import android.os.Bundle
+import com.airbnb.lottie.LottieAnimationView
+import com.airbnb.lottie.RenderMode
+import com.duckduckgo.app.browser.databinding.ActivityFireAnimationBinding
+import com.duckduckgo.app.global.DuckDuckGoActivity
+import com.duckduckgo.app.settings.clear.FireAnimation
+import com.duckduckgo.mobile.android.ui.view.setAndPropagateUpFitsSystemWindows
+import com.duckduckgo.mobile.android.ui.view.show
+import com.duckduckgo.mobile.android.ui.viewbinding.viewBinding
+import kotlinx.android.synthetic.main.sheet_fire_clear_data.*
```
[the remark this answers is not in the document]
joshliebe: Good spot, will remove!

app/src/main/res/values/styles.xml
joshliebe: I tried that style initially but it has a transition when you start the activity that doesn't look that great:

https://user-images.githubusercontent.com/3471025/133599644-bb05d90e-90f1-42fb-b67c-0b89b0328c80.mp4

PR #1427, opened 2021-09-13
Android - change menu actions on pages that were added to Bookmarks or Favorites

anikiki:
Task/Issue URL: https://app.asana.com/0/414730916066338/1200559385909216/f

### Description

This PR adds the ability to remove favorites and edit bookmarks from the overflow menu.
Prior to this, the user could only add favorites and bookmarks from the overflow menu. The actions to "Add Bookmark" /  "add Favorite" were displayed in the overflow menu matter if the visited site was already favorite / bookmarked.

### Steps to test this PR

_Feature `Remove favorite`_

1. Visit a site that was never added as favorite.
2. Tap on the overflow menu and check you see in the list the text "Add Favorite".
3. Tap on "Add favorite". You should see the snackbar at the bottom mentioning the favorite was added.
4. Tap on the overflow menu and check you see in the list the text "Remove Favorite" where "Add Favorite" used to be.
5. Tap on "Remove favorite". You should see the snackbar at the bottom mentioning the site was deleted from favorites.
6. Tap on the overflow menu and check you see in the list the text "Add Favorite".

_Feature `Edit bookmark`_

1. Visit a site that was never bookmarked.
2. Tap on the overflow menu and check you see in the list the text "Add Bookmark".
3. Tap on "Add Bookmark". You should see the snackbar at the bottom mentioning the bookmark was added.
4. Tap on the overflow menu and check you see in the list the text "Edit Bookmark" where "Add Bookmark" used to be.
5. Tap on "Edit Bookmark". You should see a screen that allows you to modify the bookmark details.

_Other tests to make sure the behavior is as expected_

_Removing a favorite / bookmarked site from `Bookmarks` screen_

1. Visit a site that was never added as favorite and was never bookmarked.
2. Tap on the overflow menu and select "Add Favorite"
3. Tap on the overflow menu and select "Add Bookmark"
4. Tap on the overflow menu and check you see "Remove Favorite" and "Edit Bookmark" in the list.
5. From the overflow menu tap on "Bookmarks".
6. Remove the site from both sections: FAVORITES and BOOKMARKS and go back.
7. Tap on the overflow menu and check you see "Add Favorite" and "Add Bookmark" in the list.

_Undo a removed favorite_

1. Visit a site that was previously added as favorite.
2. Tap on the overflow menu and select "Remove Favorite". You should see the snackbar at the bottom mentioning the favorite was deleted.
3. Tap on the "UNDO" action on the snackbar.
4. Tap on the overflow menu and check you still have "Remove Favorite" in the list.

_Import a `bookmarks` file and visit one of the favorite /  bookmarked sites_ 

1. Tap on the overflow menu and select "Bookmarks"
2. On "Bookmarks" screen, tap on the overflow menu and select "Import". Import a bookmarks file and go back.
3. Visit a site that was never added as favorite and was never bookmarked.
4. Tap on the overflow menu and check that you see "Add Bookmark" and "Add Favorite" in the list.
5. Visit a site that was previously favorite / bookmarked (it was present in the file you imported).
6. Tap on the overflow menu and check that you see "Edit Bookmark" and "Remove Favorite" in the list.

### UI changes
Uploaded videos in the ticket as they're too big.


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -139,6 +140,7 @@ class BrowserTabViewModel(
     private val contentBlocking: ContentBlocking,
     private val networkLeaderboardDao: NetworkLeaderboardDao,
     private val bookmarksDao: BookmarksDao,
+    private val bookmarksRepository: BookmarksRepository,
```
malmstein, line 143: why do we need `BookmarksDao` and `BookmarksRepository`. If the Repo is already using Dao, can we just use one the two?

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -3510,6 +3523,89 @@ class BrowserTabViewModelTest {
         assertNull(privacyGradeState().privacyGrade)
     }
 
+    fun whenEditBookmarkRequestedThenDaoIsNotUpdated() {
+        coroutineRule.runBlocking {
```
malmstein, line 3527: can you update this to`whenEditBookmarkRequestedThenDaoIsNotUpdated() = coroutineRule.runBlocking` to follow the convention of the other tests?

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -3510,6 +3523,89 @@ class BrowserTabViewModelTest {
...
+            bookmarksListFlow.send(listOf(Bookmark(id = 1L, title = "", url = "www.example.com", parentId = 0L)))
+            loadUrl("www.example.com", isBrowserShowing = true)
+            testee.onBookmarkMenuClicked()
+            verify(mockBookmarksDao, never()).insert(any())
+        }
+    }
+
+    @Test
+    fun whenEditBookmarkRequestedThenEditBookmarkPressedPixelIsFired() {
+        coroutineRule.runBlocking {
```
malmstein, line 3537: ditto

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -3510,6 +3523,89 @@ class BrowserTabViewModelTest {
...
+            bookmarksListFlow.send(listOf(Bookmark(id = 1L, title = "", url = "www.example.com", parentId = 0L)))
+            loadUrl("www.example.com", isBrowserShowing = true)
+            testee.onBookmarkMenuClicked()
+            verify(mockPixel).fire(AppPixelName.MENU_ACTION_EDIT_BOOKMARK_PRESSED.pixelName)
+        }
+    }
+
+    @Test
+    fun whenEditBookmarkRequestedThenEditDialogIsShownWithCorrectUrlAndTitle() {
+        coroutineRule.runBlocking {
```
malmstein, line 3547: ditto

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -3510,6 +3523,89 @@ class BrowserTabViewModelTest {
...
+            assertTrue(commandCaptor.lastValue is Command.ShowEditSavedSiteDialog)
+            val command = commandCaptor.lastValue as Command.ShowEditSavedSiteDialog
+            assertEquals("www.example.com", command.savedSite.url)
+            assertEquals("title", command.savedSite.title)
+        }
+    }
+
+    @Test
+    fun whenRemoveFavoriteRequestedThenDaoInsertIsNotCalled() {
+        coroutineRule.runBlocking {
```
malmstein, line 3561: ditto

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1398,20 +1423,49 @@ class BrowserTabViewModel(
         }
     }
 
-    fun onAddFavoriteMenuClicked() {
+    private suspend fun editSiteBookmark(url: String) {
+        withContext(dispatchers.io()) {
+            bookmarksState.value?.firstOrNull { it.url == url }
```
malmstein, line 1428: why do we need to change the value of `LiveData ` here? Nothing has changed until the user interacts with the `Dialog`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1398,20 +1423,49 @@ class BrowserTabViewModel(
...
+            )
+            saveFavoriteSite(url, title ?: "")
+        }
+    }
 
-        saveSiteAsFavorite(url, title)
+    private fun removeFavoriteSite(url: String) {
+        viewModelScope.launch {
+            withContext(dispatchers.io()) {
+                currentCtaViewState().favorites.firstOrNull {
```
malmstein, line 1455: same here, the call to `Repository `already triggers a call through the flow, and the `ViewState `should be updated

malmstein (commented):
There are several tests that don't follow the convention of
`test() = coroutineRule.runBlocking {}`, could you update them?

Where did you get the translations from? ,Unless you got them using https://app.asana.com/0/0/1185914506984794, you don't need to provide them.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1980,6 +2034,18 @@ class BrowserTabViewModel(
...
 
+    private fun isBookmark(url: String?): Boolean {
+        if (url == null) return false
+        val bookmarks = bookmarksState.value
+        return bookmarks?.any { it.url == url } ?: false
+    }
+
+    private fun isFavorite(url: String?): Boolean {
+        if (url == null) return false
+        val favorites = currentCtaViewState().favorites
```
cmonfortep, line 2045: I'm a bit worried about using this ViewState for our goal. Right now is ok, but imagine that at some point we change the logic that updates the CtaViewState (e.g: we want to show only 10 favorites in the home tab), that will break our logic.

This scenario is different from how we are dealing with bookmarks. For bookmarks, we are using directly an observer. 

I have some proposals:
1. instead of using an attribute from this class, always go through the repository (similar to how we update `isWhitelisted` in the same  viewstate)
2. similar approach to bookmarks.

Let's sync on zoom about this :)

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -3510,6 +3523,89 @@ class BrowserTabViewModelTest {
         assertNull(privacyGradeState().privacyGrade)
     }
 
+    fun whenEditBookmarkRequestedThenDaoIsNotUpdated() {
```
cmonfortep: NIT: You can use `= coroutineRule.runBlocking {` instead.

cmonfortep (commented):
Good job @anikiki on your first contribution :)

I've just left a comment, we can discuss it over zoom, but otherwise this is looking good.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
cmonfortep: @malmstein curious about this one. I don't think this line is changing the value of `LiveData` right? I think is just using the latest value stored (and emitted) as a cache to perform a query.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
anikiki: Having both `BookmarksDao` and `BookmarksRepository` is not ideal. I see it as a first step towards moving away from using DAOs inside a viewmodel and in general for moving away from using entities in viewmodel (i.e BookmarkEntity).
I can remove the `BookmarksDao` as part of this PR or a separate one. The change is not big.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
anikiki: I can see how it may be confusing and will extract to another variable to make the code more clear.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1382,9 +1416,20 @@ class BrowserTabViewModel(
...
     }
 
-    suspend fun onBookmarkAddRequested() {
+    fun onBookmarkMenuClicked() {
         val url = url ?: return
-        val title = title ?: ""
+        viewModelScope.launch {
+            if (currentBrowserViewState().bookmark != null) {
+                pixel.fire(AppPixelName.MENU_ACTION_EDIT_BOOKMARK_PRESSED.pixelName)
+                editSiteBookmark(url)
```
cmonfortep, line 1424: See my other comment. If we can obtain the bookmark at this point, we could replace `editSiteBookmark` for `onEditSavedSiteRequested`.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1398,27 +1443,53 @@ class BrowserTabViewModel(
...
-    fun onAddFavoriteMenuClicked() {
-        val url = url ?: return
-        val title = title ?: ""
-
-        val buttonHighlighted = currentBrowserViewState().addFavorite.isHighlighted()
-        pixel.fire(
-            AppPixelName.MENU_ACTION_ADD_FAVORITE_PRESSED.pixelName,
-            mapOf(FAVORITE_MENU_ITEM_STATE to buttonHighlighted.toString())
-        )
+    private suspend fun editSiteBookmark(url: String) {
```
cmonfortep, line 1446: `url` is not used inside the method (I know we used it before).
 
a) We can replace the function parameter type on this method and receive a `SavedSite`. In that case, we could just use `onEditSavedSiteRequested(savedSite)` instead of this method.

b) We remove the function parameter and the code stays as it is.

I'm fine with both approaches.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2057,11 +2052,13 @@ class BrowserTabFragment :
...
                 newTabPopupMenuItem.isEnabled = browserShowing
                 addBookmarksPopupMenuItem?.isEnabled = viewState.canAddBookmarks
+                addBookmarksPopupMenuItem?.text =
+                    getString(if (viewState.bookmark != null) R.string.editBookmarkMenuTitle else R.string.addBookmarkMenuTitle)
                 addFavoritePopupMenuItem?.isEnabled = viewState.addFavorite.isEnabled()
-                if (viewState.addFavorite.isHighlighted()) {
-                    addFavoritePopupMenuItem.text = getString(R.string.addFavoriteMenuTitleHighlighted)
-                } else {
-                    addFavoritePopupMenuItem.text = getString(R.string.addFavoriteMenuTitle)
+                addFavoritePopupMenuItem.text = when {
```
cmonfortep, line 2058: 👍  I like how you are using `when` here.

cmonfortep (approved):
LGTM! You nailed your first contribution and I'm excited about it. 🎉 



I've just left a minor comment, but all good! 🚀

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
anikiki: Good catch! Updated.

Issue #1440, opened 2021-09-22
[Feature] Show Bookmark Folder name on Bookmark Saved Snackbar


4shutosh:
**Feature Description**
The user should be shown the folder name where the bookmark is saved, enabling it will give the user chooses to change the location at the moment of saving the bookmark.

cmonfortep:
@4shutosh similar to my comment in https://github.com/duckduckgo/Android/issues/1439. Please open a discussion to suggest any bookmark improvement. 

More info in our [contributions guidelines](https://github.com/duckduckgo/Android/blob/develop/CONTRIBUTING.md#contribute-code).

Issue #1439, opened 2021-09-21
[Feature] Ability to create bookmark folder from (edit screen) EditSavedSiteDialogFragment -> BookmarkFoldersActivity


4shutosh:
**Feature Description**
Users should be able to create a new folder from the EditSavedSiteDialogFragment/edit screen, which is much more convenient than going back to the bookmarks screen and creating a folder and then going back again to select that folder for the bookmark.

Even if the user goes to the bookmarks list screen and then wishes to check the folder there is no add folder directly from the edit screen. The user will need to come back to the parent list screen and then add the folder and then go back to update the bookmark location.

**Suggested Behaviour**
Giving add folder option on the BookmarkFoldersActivity solves the above problem.

cmonfortep:
Hey @4shutosh it's a nice idea.

However, please open a discussion for this one (you can add also https://github.com/duckduckgo/Android/issues/1440 in the same discussion since they are both related) to suggest any bookmark improvement. 

As per our [contributions guidelines](https://github.com/duckduckgo/Android/blob/develop/CONTRIBUTING.md#contribute-code), we only accept Bug reports as Issues.

Issue #1438, opened 2021-09-21
[Bug] Edit Bookmark does not display the correct bookmark location.


4shutosh:
### Describe the bug

Once the user saves a bookmark on a location. Edit Bookmark does not display the correct bookmark location, from the BrowserTabFragment.

### How to Reproduce

1. Open any webpage 
2. Save it as a bookmark in any folder other than the "Bookmarks" root folder
3. After saving edit the bookmark, from the same popup menu.
4. The name being defaulted "Bookmarks" is shown.

### Expected behavior

Precise Bookmark location name should be shown.

### Environment

```markdown
- DDG App Version: develop branch
- Device: Pixel 5
- OS: Android 11
```

malmstein:
@joshliebe can you take a look? I can reproduce this every time.

4shutosh:
This is going to be fixed in #1441

PR #1441, opened 2021-09-22
[Feature] Bookmarks Improvement

4shutosh:
Task/Issue URL: 
#1438
#1439 
#1440

PR related to the feature proposal mentioned in discussion #1442 

### Description
- Users can create a bookmark folder from the edit bookmark screen.
-  Edit Bookmark shows the correct Bookmark Location, From `BrowserFragment`

### Steps to test this PR
1. Add a bookmark
2. Edit the bookmark from the menu or from the snackbar
3. Select the Location
4. New bookmark folder can be created.
5. After creating a new bookmark folder, select the bookmark folder
6. Save : bookmark saved inside the new bookmark folder.


4shutosh:
Screenshot_20210922-124907_DuckDuckGo
Screenshot_20210922-124913_DuckDuckGo
Screenshot_20210922-124944_DuckDuckGo

cmonfortep:
@4shutosh thanks for your time here. As I posted in the different issues you created, let's first have a discussion before we introduce any improvement.

We accept contributions for bug fixes after we triage the issue and we add [the label "help wanted"](https://github.com/duckduckgo/Android/blob/develop/CONTRIBUTING.md#what-to-expect).

For feature requests/improvements, you need to open a discussion first. Inside the discussion, we will have a nice chat to know more about the proposal, understand what's the user value, etc. That's because we will need to share the proposed change with our Product team and/or Design team. If it's aligned with our roadmap and we are happy with the idea then we will be able to give you support with the improvements.

Hope that's ok with you, let's close this PR for now. We can open it again once I get the internal approval. Thanks again, and I think the proposal makes sense, but we need to check internally first 👍

4shutosh:
Thank you for the explanation @cmonfortep 🙌🏻 

That's unfortunate but will surely do the required. 

I've created a discussion here #1442

4shutosh:
Behavioral changes introduced in this PR:
- The edit bookmark shows the folder name now
- We can now create a new bookmark folder while editing the bookmark from the edit option either from snackbar or from the menu options.

@cmonfortep please review.
Do let me know with any changes are required along with new tests to be written or not.

cmonfortep:
Thanks @4shutosh. 

"The edit bookmark shows the folder name now". Sorry if it was not clear, but we discarded de idea of showing folder names in the snackbar (see my comment: https://github.com/duckduckgo/Android/discussions/1442#discussioncomment-1551699)

4shutosh:
Sorry for the confusion. Yes it is not on the snackbar but on the edit saved site screen.
image

4shutosh:
Here the "Edited" is the bookmark folder name.

cmonfortep:
Thanks. I will try to review the PR this week. Could you update the description? (just to reflect what this PR is finally addressing since we discarded some initial proposals)

cmonfortep:
@4shutosh just to give you some heads up. Testing this right now, looks great. Our designer is taking a look, should give me feedback on Monday.

4shutosh:
Thank you for the update 🙌🏻

cmonfortep:
@4shutosh thank you for your patience.

Feedback from the design team was really positive, that's really cool! They only asked if we can take the user back to the edit screen when the user created the new folder (same as in chrome).

Feedback is **optional** and I haven't checked how complex it is, so let me know if you are interested (and if you have time, it's ok if you can't) or not. In case you want to contribute on that, you can do it here or in another PR.

4shutosh:
That's great!

I'm interested in working on the feedback.

But I'll be able to do that (or any test related progress of this PR: as some tests are failing) after a week.

Please proceed with the tests fix if this feature is of priority for upcoming release within a week. In that case, I'll be continuing the feedback work on a new PR.

cmonfortep:
Sounds good, I will review this PR as it is now, and further improvements can be done on a separated PR.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/EditSavedSiteDialogFragment.kt
```diff
@@ -20,15 +20,18 @@ import android.os.Bundle
...
 import com.duckduckgo.app.bookmarks.ui.bookmarkfolders.AddBookmarkFolderDialogFragment
+import com.duckduckgo.app.bookmarks.ui.bookmarkfolders.BookmarkFoldersActivity
 import com.duckduckgo.app.browser.R
 import com.duckduckgo.app.global.view.TextChangedWatcher
 
 class EditSavedSiteDialogFragment : SavedSiteDialogFragment() {
 
     interface EditSavedSiteListener {
-        fun onSavedSiteEdited(savedSite: SavedSite)
+        fun onSavedSiteFavoriteEdited(favorite: SavedSite.Favorite)
```
cmonfortep, line 33: I would try to avoid creating 2 methods here if it's not really necessary.

Checking the implementations of `onSavedSiteBookmarkEdited` only 1 of the 3 is using bookmarkFolder -> `BrowserTabViewModel`. I will leave a comment there.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1547,33 +1567,37 @@ class BrowserTabViewModel(
...
-            }
+    override fun onSavedSiteFavoriteEdited(favorite: SavedSite.Favorite) {
+        viewModelScope.launch(dispatchers.io()) {
+            editFavorite(favorite)
         }
     }
 
-    fun onEditSavedSiteRequested(savedSite: SavedSite) {
-        command.value = ShowEditSavedSiteDialog(savedSite)
+    override fun onSavedSiteBookmarkEdited(bookmark: SavedSite.Bookmark, bookmarkFolder: BookmarkFolder) {
```
cmonfortep, line 1576: I don't think we need bookmarkFolder as parameter here. I would prefer to keep one single method and get benefit of `SavedSite` sealed class.

What about just receiving `bookmark`, and retrieve `bookmarkFolder` from db inside `editBookmark`?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -986,6 +993,15 @@ class BrowserTabViewModel(
         }
     }
 
+    private suspend fun getBookmarkFolder(bookmark: SavedSite.Bookmark?): BookmarkFolder? {
```
cmonfortep, line 996: any reason method param should be nullable? if it's not a valid argument, I would remove nullability.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1547,33 +1567,37 @@ class BrowserTabViewModel(
...
     fun onDeleteQuickAccessItemRequested(savedSite: SavedSite) {
         command.value = DeleteSavedSiteConfirmation(savedSite)
     }
 
-    private suspend fun editBookmark(bookmark: SavedSite.Bookmark) {
+    private suspend fun editBookmark(bookmark: SavedSite.Bookmark, bookmarkFolder: BookmarkFolder) {
         withContext(dispatchers.io()) {
             bookmarksRepository.update(bookmark)
         }
+        updateBookmarkAndFavoriteState(bookmark.url)
```
cmonfortep, line 1594: after updating the bookmark, I would retreive the bookmark folder using `getBookmarkFolder` here. That way we can remove the method param and go back to a single method + sealed class in other classes.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -277,8 +279,8 @@ class BrowserTabViewModel(
         object HideKeyboard : Command()
         class ShowFullScreen(val view: View) : Command()
         class DownloadImage(val url: String, val requestUserConfirmation: Boolean) : Command()
-        class ShowSavedSiteAddedConfirmation(val savedSite: SavedSite) : Command()
-        class ShowEditSavedSiteDialog(val savedSite: SavedSite) : Command()
+        class ShowSavedSiteAddedConfirmation(val savedSite: SavedSite, val bookmarkFolder: BookmarkFolder?) : Command()
```
cmonfortep, line 282: Not sure about this approach: we are using a sealed class (`savedSite`) to represent a Bookmark/Favorite, but now we have a second param that only applies to Bookmarks (`bookmarkFolder`). If that's the case, it should be part of the sealed class as well, because it only applies to one of the types.

`SavedSite` is a domain (layer) sealed class.

In here, I think you are passing `bookmarkFolder` to show bookmark folder name in the dialog. If that's the case, my approach will be to create a sealed class specific for those dialog. Then you can map your savedSite into your new saved site and use it there. But please, double check this approach, I could be missing something.

cmonfortep (changes requested):
Thanks, overall looks great. There are a few decisions we should double check, I left my comments bellow.

4shutosh:
Thank you for the detailed review @cmonfortep.
I'll understand and double check the issues within few days and make necessary ammendments.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
4shutosh: instead of checking the nullability of the bookmark (of the saved site got from the bookmarksRepository) every time I use this function, I thought to add the null check inside the function, hence the parameter is nullable.

Will think of a better way.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
4shutosh: Sure, this can be done. 

Trying this right away.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/EditSavedSiteDialogFragment.kt
4shutosh: Yes removing the bookmark folder from the parameter helped me remove this bad code.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
4shutosh: You are right.
I've now used a separate data class for the change in saved site to process interaction with the views.

4shutosh:
@cmonfortep I've improved the areas of code mentioned by you. Please take a look.

Also, please help me with the falling tests. Can't find a meaningful message in the [log](https://app.bitrise.io/build/032fb40f-ad0b-42c5-8850-abf30c17cf89) for the failure of `virtual-device-testing-for-android`. 

I haven't written any new tests related to this PR, please let me know if any new tests are required due to the introduction of new behavior.

cmonfortep:
Thanks, @4shutosh, will take a look next week. I'd suggest aligning with develop to ensure no conflicts/issues.

cmonfortep (changes requested):
Thanks @4shutosh, haven't reviewed all the changes. I've only focused on the dialogs and the browsertabviewmodel. Left some comments bellow.

While testing this, I've noticed an unexpected behavior: After editing a bookmark (changing folder), a snackbar saying "bookmark added" appears. Testing our latest app, the behavior is different. We only show the snackbar when the user adds a new bookmark, not after editing the bookmark.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -191,6 +197,7 @@ class BrowserTabViewModel(
         val canSharePage: Boolean = false,
         val canAddBookmarks: Boolean = false,
         val bookmark: SavedSite.Bookmark? = null,
+        val bookmarkFolder: BookmarkFolder? = null,
```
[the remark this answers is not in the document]
4shutosh: Yes you are right, the bookmarkFolder can be removed. Doing that right away.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/SavedSiteDialogFragment.kt
```diff
@@ -44,6 +41,7 @@ abstract class SavedSiteDialogFragment : DialogFragment() {
 
     abstract fun onConfirmation()
     abstract fun configureUI()
+    abstract fun configureClickListeners()
```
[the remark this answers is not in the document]
4shutosh: Done

4shutosh:
Sorry for the delay. @cmonfortep
I've done the improvements in the code as mentioned by you. Also removed the unnecessary snackbar behaviour after edit of bookmark.

This is getting stretched more than expected. I'll try my best to complete the changes related to this PR as early as possible now. Please help me with the review. 

Thank you again. 🙌🏻

malmstein:
@cmonfortep can you check on this?

app/src/main/java/com/duckduckgo/app/bookmarks/model/BookmarksRepository.kt
```diff
@@ -82,6 +87,10 @@ class BookmarksDataRepository(
         return bookmarksDao.getBookmarksByParentIdSync(parentId)
     }
 
+    override suspend fun getBookmarkFolderByParentId(parentId: Long): BookmarkFolder? {
+        return bookmarkFoldersDao.getBookmarkFolders().firstOrNull()?.find { it.id == parentId }
```
cmonfortep, line 91: Haven't checked, but I think filtering at SQL level is more efficient than getting the whole list of folders here, and then filtering. Can you create a new method in `BookmarkFoldersDao`, and just use it here?

app/src/main/java/com/duckduckgo/app/bookmarks/ui/EditSavedSiteDialogFragment.kt
```diff
@@ -42,18 +42,22 @@ class EditSavedSiteDialogFragment : SavedSiteDialogFragment() {
...
 
     private fun validateInput(newValue: String, existingValue: String) =
         if (newValue.isNotBlank()) newValue else existingValue
 
-    private fun populateFields(titleInput: EditText, urlInput: EditText) {
+    private fun populateFields(titleInput: EditText, urlInput: EditText, savedLocation: EditText) {
         titleInput.setText(getExistingTitle())
         urlInput.setText(getExistingUrl())
+        getExistingBookmarkFolderName()?.let {
+            if (it.isNotEmpty()) savedLocation.setText(getExistingBookmarkFolderName())
```
cmonfortep, line 59: `savedLocation.setText(it)` no need to call the same method twice.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1367,20 +1367,24 @@ class BrowserTabFragment :
...
             }
             .show()
     }
 
-    private fun editSavedSite(savedSite: SavedSite) {
-        val addBookmarkDialog = EditSavedSiteDialogFragment.instance(savedSite)
+    private fun editSavedSite(savedSiteChangedViewState: SavedSiteChangedViewState) {
+        val addBookmarkDialog = EditSavedSiteDialogFragment.instance(
+            savedSiteChangedViewState.savedSite,
+            savedSiteChangedViewState.bookmarkFolder?.id ?: 0,
```
cmonfortep, line 1385: I want to check with someone from the team, what's the difference between `savedSiteChangedViewState.bookmarkFolder?.id` vs `savedSiteChangedViewState.savedSite.parentId` when we have a bookmark. So maybe I will ask to double check this part.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/bookmarkfolders/BookmarkFoldersViewModel.kt
```diff
@@ -53,19 +60,42 @@ class BookmarkFoldersViewModel(
...
             val folderStructure = bookmarksRepository.getFlatFolderStructure(selectedFolderId, currentFolder, rootFolderName)
             onFolderStructureCreated(folderStructure)
         }
     }
 
+    private fun saveCurrentFolderAndSelectedFolderId(currentFolder: BookmarkFolder?, selectedFolderId: Long) {
+        viewState.value = currentViewState().copy(currentFolder = currentFolder, selectedFolderId = selectedFolderId)
+    }
+
+    fun newFolderAdded(rootFolderName: String) {
```
cmonfortep, line 74: Test missing for this.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/bookmarkfolders/BookmarkFoldersViewModel.kt
```diff
@@ -35,16 +37,21 @@ import javax.inject.Provider
 class BookmarkFoldersViewModel(
     val bookmarksRepository: BookmarksRepository,
     private val dispatcherProvider: DispatcherProvider
-) : ViewModel() {
+) : ViewModel(), AddBookmarkFolderListener {
 
     data class ViewState(
-        val folderStructure: List<BookmarkFolderItem> = emptyList()
+        val folderStructure: List<BookmarkFolderItem> = emptyList(),
+        var currentFolder: BookmarkFolder? = null,
```
cmonfortep, line 44: Sorry, can you explain a bit why we have added these 2 new properties to the viewstate? what are we trying to achieve?

cmonfortep (commented):
Thanks and sorry for the delay in the review. As you said, this is taking longer than we expected and it was difficult to remember what was missing here.

I'm leaving some comments. So far it's looking good. Just something important to mention:
- be sure new code is properly tested if the class has a test (otherwise is not necessary, e.g: activities/fragments)
- when aligning with develop, you may get lot of conflicts. We have introduced changes in our code style, so expect some conflicts there.

Thanks.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1618,6 +1639,11 @@ class BrowserTabViewModel(
         withContext(dispatchers.io()) {
             bookmarksRepository.update(bookmark)
         }
+        withContext(dispatchers.main()) {
```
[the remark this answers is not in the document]
4shutosh: Yes, there is no need to update that. Did not realize we were using the flow directly from the repository.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/bookmarkfolders/BookmarkFoldersViewModel.kt
4shutosh: In the process of understanding and making things work (was not able to understand the schema completely), primarily making use of the `getFlatFolderStructure` function these two properties were required.

But they are not required in the ViewState. I'll add them as local variables inside the viewModel.

4shutosh: I'm still not sure about the correctness of the implementation.

The major change in this class is the usage of `AddBookmarkFolderListener` to listen to the addition of a new bookmark folder and in turn update the view i.e the "flat folder structure".

As I was not able to understand what's happening in `getFlatFolderStructure` inside the `BookmarksDataRepository` I was only able to reuse the function.

To reuse the function there is a `rootFolderName` property required. Which was initially obtained from the string value (maybe to make translations compatible). So I did the same using a new Command object `NewFolderCreatedUpdateTheStructure`.

cmonfortep: @4shutosh I've been taking a look at this part. And I understand better what was your problem here.

This is my suggestion to simplify this:
- keep the `viewstate` as it was before. No need to add new properties.
- change the signature of `newFolderAdded` to receive those params from the activity.

Doing that you will simplify a lot this class.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/bookmarkfolders/BookmarkFoldersActivity.kt
```diff
@@ -85,15 +108,26 @@ class BookmarkFoldersActivity : DuckDuckGoActivity() {
         }
     }
 
+    private fun setNewlyCreatedSelectedFolderResult() {
```
cmonfortep, line 111: This is my proposal to avoid storing those constants in the ViewModel. Just get it from intent.extras since they never change. 

``` 
   private fun setNewlyCreatedSelectedFolderResult() {
        viewModel.newFolderAdded(
                rootFolderName = getString(R.string.bookmarksSectionTitle),
                selectedFolderId = intent.extras?.getLong(KEY_BOOKMARK_FOLDER_ID) ?: 0,
                currentFolder = intent.extras?.getSerializable(KEY_CURRENT_FOLDER) as BookmarkFolder?
        )
    }
```

4shutosh: Yes, this will make things simpler. Working on the conflict parts now will update the PR soon along with these implementations.

4shutosh:
> Feedback from the design team was really positive, that's really cool! They only asked if we can take the user back to the edit screen when the user created the new folder (same as in chrome).

I tried to work on this. This can be done in a couple of ways as per my understanding.

1 
```
    override fun onBookmarkFolderAdded(bookmarkFolder: BookmarkFolder, exitAndSetSelected: Boolean) {
            bookmarksRepository.insert(bookmarkFolder)
        }
        if (exitAndSetSelected) onItemSelected(bookmarkFolder)
        else command.value = NewFolderCreatedUpdateTheStructure
    }
```
But this will not give us the ideal behavior and the list will be visible for a split second.

2. Removing the list screen (`Select Location`) from back stack and maybe implementing the interface of folder addition directly on the `Edit Bookmark` screen. But this is not possible in our case as the `Select Location` list screen i.e `BookmarkFoldersActivity` is an activity. All the screens inside the chrome bookmark flow would be Fragments.

Please take a look and let me know any other changes required.

cmonfortep:
Feedback was optional. If you want to work on that, it's awesome! but my proposal is to do it on another PR once this one is merged. Just to avoid delaying more this feature.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/bookmarkfolders/BookmarkFoldersViewModel.kt
4shutosh: Added please check

app/src/main/java/com/duckduckgo/app/bookmarks/model/BookmarksRepository.kt
4shutosh: Created a new method please check.

4shutosh:
Sure I'll do that.
I've done all the changes required, resolved merge conflicts and the feature is working as expected. Please let me know any other changes required.

app/src/main/java/com/duckduckgo/app/bookmarks/model/BookmarksRepository.kt
```diff
@@ -96,6 +100,10 @@ class BookmarksDataRepository(
         return bookmarksDao.getBookmarksByParentIdSync(parentId)
     }
 
+    override suspend fun getBookmarkFolderByParentId(parentId: Long): BookmarkFolder {
```
cmonfortep, line 103: Can you add a test to cover this method?

What will happen when parentId is not found or doesn't return any folder? Please also test that scenario.

cmonfortep (commented):
Thanks @4shutosh. We are almost done here 🎉  

I've just left one last comment to cover with a unit test the method in the repository. I've tested the functionality and it works as expected.

app/src/main/java/com/duckduckgo/app/bookmarks/model/BookmarksRepository.kt
4shutosh: I've added the test.

In that case, the bookmark is treated as if it is stored in the default root folder under the name "Bookmarks".

cmonfortep (approved):
Thanks for the changes @4shutosh LGTM! 

Please, solve the conflicts with develop (probably the test has been moved into unit tests folder), and as soon as the check is green we will merge it. :)

cmonfortep:
Sharing here the stacktrace of the failing test.
```
com.duckduckgo.app.bookmarks.model.BookmarksDataRepositoryTest > whenGetBookmarkFolderByParentIdThenReturnBookmarkFolderForParentId[test(AVD) - 5.1.1] FAILED 
	java.lang.AssertionError: expected:<BookmarkFolder(id=1, name=name, parentId=11, numBookmarks=0, numFolders=0)> but was:<null>
	at org.junit.Assert.fail(Assert.java:89)
```

4shutosh:
Yes, got the problem in the test. I've pushed a commit with a change in the test. The tests should pass now.

4shutosh:
Hahaha nooooo, another incoming conflict 😭. Merging again and pushing the same.

4shutosh:
Yayyy 🎉 thank you for your guidance @cmonfortep 🙌🏻

PR #1464, opened 2021-10-04
Optimized imports

akm0012:
Simply optimized the imports within the project.


malmstein:
Thanks for this @akm0012, however we can't just merge a 101 files PR.  As you can see in our [Contributing Guidelines](https://github.com/duckduckgo/Android/blob/develop/CONTRIBUTING.md), we'd love your help in issues labelled as such, translations or bug fixes. 

In this specific case, organizing imports does not belong to any of those types.

akm0012:
Ahh, ok my bad! I'll take a look at those tags.

PR #1468, opened 2021-10-07
Enable non-transitive R files

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201118368747585/f

### Description
Currently our gradle is configured in such a way that all `R` resources are rechable from any module, without the need to specify them as a gradle dependency. This has the following cons:
* compilation times are higher, as aapt has to include references to all resources in all gradle modules
* incremental compilations are higher, as any resource change affects more modules than it should
* APK/AAB files are bigger than needed

See [asana task](https://app.asana.com/0/414730916066338/1201118368747585/f) for a more detailed description.

This PR enables the `android.nonTransitiveRClass=true` in the `gradle.properties` file so that modules can only reference resources in the current module, or in modules explicitly added as dependency

### Steps to test this PR

The app builds for all flavors

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,6 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 android.useAndroidX=true
+android.nonTransitiveRClass=true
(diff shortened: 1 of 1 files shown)


malmstein:
1 line change, 40 sec build improve. 💥

PR #1473, opened 2021-10-09
Add return type in single line function

nikit19:
### Description

Add return type in single line function to make it more readable and consistent with other single line functions in that file


cmonfortep:
Thanks for the contribution. However, the proposed change doesn't address any open issue or fixes any code style/warning in our code.

PR #1476, opened 2021-10-09
Use View Binding for click listeners in Privacy Practices Activity

nikit19:
### Description

Use View Binding for click listeners in Privacy Practices Activity

I have verified that the click listeners work just like before ✅


nikit19:
@cmonfortep Thanks for the review. I will make sure to create an issue before raising a pull request from next time

PR #1511, opened 2021-10-15
Issue fixed #1479 : [Bug] Invalid URLs in broken site Android reports

### LINKED ISSUE #1479: [Bug] Invalid URLs in broken site Android reports
### Describe the bug

We are receiving broken site reports where the URL is `null://null`.

That means we have an edge case where the user is able to report a broken site but the URL is null.

### How to Reproduce

1. (simulate low speed connection) On an emulator, cellular settings → configure network type as GSM
2. Open a new tab
3. Search or visit any website
4. Press back before site loads
5. Press forward in the overflow menu
6. Omnibar is empty, but the overflow menu has all options enabled
7. Report broken site → see URL is null://null

you can try also clicking on privacy grade while page is loading, you can report a broken site from there too.

### Expected behavior

When URL is null, we should not send the broken site report.

### Environment

```markdown
- DDG App Version: all versions
- Device: Any device
- OS: any OS version
```

Adw41t:
Task/Issue URL:
#1479 
### Description
Check if url has null://null and hide/disable "Report Broken Site" button
(Alternate Solution)

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -891,6 +891,11 @@ class BrowserTabViewModel(
         val domain = site?.domain
         val canWhitelist = domain != null
         val canFireproofSite = domain != null
+        val canReportSite = if (site == null) {
+            false
+        } else {
+            !(site?.url.equals("null://null", true))
+        }
         val addFavorite = if (!currentBrowserViewState.addFavorite.isEnabled()) {
             HighlightableButton.Visible(enabled = true)
         } else {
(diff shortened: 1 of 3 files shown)


cmonfortep:
@Adw41t if this is related to https://github.com/duckduckgo/Android/pull/1510 could we have one single PR for this?

Adw41t:
> @Adw41t if this is related to #1510 could we have one single PR for this?

@cmonfortep #1510 and #1511 are both related to issue #1479 
#1510 only adds a check for null://null in url
#1511 If url is null://null then hide the "Report Broken Site" button from menu and PrivacyDashboard

I believe #1511 will be proper way to handle the issue, but this change may require a review by DDG team. Otherwise #1510 is just a simple check.
Please let me know which will be appropriate, or if both are appropriate, then I will merge both in one PR and then close the other

PR #1530, opened 2021-11-03
Improve filename guesses when MIME type provided or dots are in the path

asilvr:
## Description

This PR addresses a few issues related to the filename guessing method used in the Download Manager. To guess the filename, this PR improves usage of the MIME type when a URL path segment matches the MIME type, is ambiguous, or contain dots.

This PR also adds test coverage for when MIME types are provided -- most of the existing logic was written assuming no MIME types, which left a gap in proper filename guesses. Saving the file with a guessed name that is closer to what the user expects is a user experience improvement, and also avoids replacing files with the same name unintentionally.

Below are a few issues and their expected and actual behavior.

### Issue: Not utilizing MIME types appropriately

For a URL such as `https://example.com/files/realFilename.pdf` and a valid MIME type (e.g. `application/pdf`):

- Expected filename: `realFilename.pdf`
- Actual filename: `files.pdf`

### Issue: Not appropriately handling dots in the URL path:

For a URL such as `https://example.com/files.folder/realFilename.pdf` and a valid MIME type (e.g. `aplication/pdf`):

- Expected filename: `realFilename.pdf`
- Actual filename: `files.folder`

For a URL such as `https://example.com/files.folder/realFilename.pdf` and no valid MIME type provided:

- Expected filename: `realFilename.pdf`
- Actual filename: `files.folder`

### Issue: Not appropriately handling ambiguous filenames:

For a URL such as `https://example.com/files/realFilename` and a valid MIME type (e.g. `application/pdf`):

- Expected filename: `realFilename.pdf`
- Actual filename: `files.pdf`

### Steps to test this PR

1. Navigate to a URL which contains a `.` in the path of an attachment

I've created these for testing, which I'll remove if/when this PR is merged:

 - https://d1x9zcshh2aa46.cloudfront.net/abc.def/example.pdf (saves as `abc.def`)
 - https://d1x9zcshh2aa46.cloudfront.net/abc.123/example.pdf (saves as `abc.123`)
 - https://d1x9zcshh2aa46.cloudfront.net/123.abc/example.pdf (saves as `123.abc`)
 - https://d1x9zcshh2aa46.cloudfront.net/123.456/example.pdf (saves as `123.456`)
 - https://d1x9zcshh2aa46.cloudfront.net/abc.pdf/example.pdf (saves as `abc.pdf`)

[...]

PR #1547, opened 2021-11-15
Fix guessing logic when a dot is present

asilvr:
### Description

This is a simplified version of https://github.com/duckduckgo/Android/pull/1530, focusing on improved handling when a `.` character is present in the guess. It also addresses an issue where URLs with multiple path segments and a MIME type was guessing the first segment of the path as the filename, as opposed to the last.

### Steps to test this PR

Navigate to any of the following URLs and observe the filename they're downloaded as

| URL      |  5.101.0 | This branch |
| ----------- | ----------- | ----------- | 
| https://d1x9zcshh2aa46.cloudfront.net/abc.def/example.pdf  | abc.def  | example.pdf |
| https://d1x9zcshh2aa46.cloudfront.net/abc.123/example.pdf | abc.123  | example.pdf |
| https://d1x9zcshh2aa46.cloudfront.net/123.abc/example.pdf | 123.abc  | example.pdf |
| https://d1x9zcshh2aa46.cloudfront.net/123.456/example.pdf | 123.456 | example.pdf |
| https://d1x9zcshh2aa46.cloudfront.net/abc.pdf/example.pdf  | abc.pdf | example.pdf |
| https://d1x9zcshh2aa46.cloudfront.net/pdfs/example.pdf | pdfs.pdf | example.pdf |
| https://d1x9zcshh2aa46.cloudfront.net/pdfs/example | pdfs.pdf | example.pdf |

_Note_: All of the above URLs have a MIME type of `application/pdf` and no content disposition.

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
@@ -86,7 +86,7 @@ class UriUtilsFilenameExtractorTest {
     @SdkSuppress(minSdkVersion = 22)
     fun whenUrlContainsAmbiguousFilenameButContainsPathSegmentsWhichLookLikeAFilenameAndMimeTypeProvidedThenFilenameShouldBeExtracted() {
         val url = "https://foo.example.com/path/dotted.path/b/b1/realFilename"
-        val mimeType: String? = "image/jpeg"
+        val mimeType = "image/jpeg"
         val contentDisposition: String? = null
 
         val extractionResult = testee.extract(buildPendingDownload(url, contentDisposition, mimeType))
@@ -97,10 +97,9 @@ class UriUtilsFilenameExtractorTest {
     }
 
     @Test
-    @SdkSuppress(minSdkVersion = 22)
     fun whenUrlContainsFilenameAndContainsMultiplePathSegmentsAndMimeTypeProvidedThenFilenameShouldBeExtracted() {
         val url = "https://foo.example.com/path/images/b/b1/realFilename.jpg"
-        val mimeType: String? = "image/jpeg"
+        val mimeType = "image/jpeg"
         val contentDisposition: String? = null
 
         val extractionResult = testee.extract(buildPendingDownload(url, contentDisposition, mimeType))
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
@@ -70,6 +70,47 @@ class UriUtilsFilenameExtractorTest {
     }
 
     @Test
+    fun whenUrlContainsFilenameButContainsPathSegmentsWhichLookLikeAFilenameThenFilenameShouldBeExtracted() {
+        val url = "https://foo.example.com/path/dotted.path/b/b1/realFilename.jpg"
+        val mimeType: String? = null
+        val contentDisposition: String? = null
+
+        val extractionResult = testee.extract(buildPendingDownload(url, contentDisposition, mimeType))
+        assertTrue(extractionResult is FilenameExtractor.FilenameExtractionResult.Extracted)
+
+        extractionResult as FilenameExtractor.FilenameExtractionResult.Extracted
+        assertEquals("realFilename.jpg", extractionResult.filename)
+    }
+
+    @Test
+    @SdkSuppress(minSdkVersion = 22)
+    fun whenUrlContainsAmbiguousFilenameButContainsPathSegmentsWhichLookLikeAFilenameAndMimeTypeProvidedThenFilenameShouldBeExtracted() {
+        val url = "https://foo.example.com/path/dotted.path/b/b1/realFilename"
+        val mimeType: String? = "image/jpeg"
+        val contentDisposition: String? = null
+
+        val extractionResult = testee.extract(buildPendingDownload(url, contentDisposition, mimeType))
+        assertTrue(extractionResult is FilenameExtractor.FilenameExtractionResult.Extracted)
+
+        extractionResult as FilenameExtractor.FilenameExtractionResult.Extracted
+        assertEquals("realFilename.jpg", extractionResult.filename)
+    }
+
+    @Test
+    @SdkSuppress(minSdkVersion = 22)
+    fun whenUrlContainsFilenameAndContainsMultiplePathSegmentsAndMimeTypeProvidedThenFilenameShouldBeExtracted() {
+        val url = "https://foo.example.com/path/images/b/b1/realFilename.jpg"
+        val mimeType: String? = "image/jpeg"
+        val contentDisposition: String? = null
+
+        val extractionResult = testee.extract(buildPendingDownload(url, contentDisposition, mimeType))
+        assertTrue(extractionResult is FilenameExtractor.FilenameExtractionResult.Extracted)
+
+        extractionResult as FilenameExtractor.FilenameExtractionResult.Extracted
+        assertEquals("realFilename.jpg", extractionResult.filename)
+    }
+
+    @Test
     fun whenUrlContainsFilenameButContainsAdditionalPathSegmentsAndQueryParamsWhichLookLikeAFilenameThenFilenameShouldBeExtracted() {
         val url = "https://foo.example.com/path/images/b/b1/realFilename.jpg/other/stuff?cb=123.com"
         val mimeType: String? = null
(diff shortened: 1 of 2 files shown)


asilvr:
@cmonfortep This PR is a reduced version of #1530 and is ready for a review (the tests I added pass here as well)

cmonfortep:
@asilvr could it be that PR description needs to be updated? I think all the MIME types references apply to the previous PR and not to this reduced version. Could you double check? Thanks.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
```diff
@@ -69,6 +69,47 @@ class UriUtilsFilenameExtractorTest {
...
+
+        val extractionResult = testee.extract(buildPendingDownload(url, contentDisposition, mimeType))
+        assertTrue(extractionResult is FilenameExtractor.FilenameExtractionResult.Extracted)
+
+        extractionResult as FilenameExtractor.FilenameExtractionResult.Extracted
+        assertEquals("realFilename.jpg", extractionResult.filename)
+    }
+
+    @Test
+    @SdkSuppress(minSdkVersion = 22)
```
cmonfortep, line 86: not sure if this is necessary, can you double check?

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
```diff
@@ -69,6 +69,47 @@ class UriUtilsFilenameExtractorTest {
...
+    fun whenUrlContainsAmbiguousFilenameButContainsPathSegmentsWhichLookLikeAFilenameAndMimeTypeProvidedThenFilenameShouldBeExtracted() {
+        val url = "https://foo.example.com/path/dotted.path/b/b1/realFilename"
+        val mimeType: String? = "image/jpeg"
```
cmonfortep, line 89: there's a warning because mimeTypes is always non-nullable.

You can use type inference here, and remove the type.
`val mimeType = "image/jpeg"`

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
```diff
@@ -69,6 +69,47 @@ class UriUtilsFilenameExtractorTest {
...
+
+        extractionResult as FilenameExtractor.FilenameExtractionResult.Extracted
+        assertEquals("realFilename.jpg", extractionResult.filename)
+    }
+
+    @Test
+    @SdkSuppress(minSdkVersion = 22)
+    fun whenUrlContainsFilenameAndContainsMultiplePathSegmentsAndMimeTypeProvidedThenFilenameShouldBeExtracted() {
+        val url = "https://foo.example.com/path/images/b/b1/realFilename.jpg"
+        val mimeType: String? = "image/jpeg"
```
cmonfortep, line 103: same here.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
```diff
@@ -69,6 +69,47 @@ class UriUtilsFilenameExtractorTest {
...
+
+        val extractionResult = testee.extract(buildPendingDownload(url, contentDisposition, mimeType))
+        assertTrue(extractionResult is FilenameExtractor.FilenameExtractionResult.Extracted)
+
+        extractionResult as FilenameExtractor.FilenameExtractionResult.Extracted
+        assertEquals("realFilename.jpg", extractionResult.filename)
+    }
+
+    @Test
+    @SdkSuppress(minSdkVersion = 22)
```
cmonfortep, line 100: ditto

cmonfortep (approved):
👍  thanks, all good. Just left some minor comments in the unit tests.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
asilvr: From what I gather, `@SdkSuppress(minSdkVersion = 22)` is used in the [existing MIME type tests](https://github.com/duckduckgo/Android/blob/784c87beef30a351436f8a5300950b6f7081c77e/app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt#L167-L195) because in SDK version 21 and below, `image/jpeg` defaults to a `.jpeg` file extension, and SDK 22 and above defaults to a `.jpg` file extension (dropped the `e`). So using it here just ensures that the test doesn't fail if/when we execute it against a lower SDK target.

In the case of this test case, I _think_ it's required to satisfy the assertion, but I will double-check.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
asilvr: Good catch. Will fix.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
asilvr: Good catch. Will fix.

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
asilvr: Since this test contains a guess with an extension which satisfies the MIME type, I believe that any SDK version will recognize this (versus the other test where the extension was ambiguous). I'll remove this and see if it still passes!

asilvr:
> @asilvr could it be that PR description needs to be updated? I think all the MIME types references apply to the previous PR and not to this reduced version. Could you double check? Thanks.

@cmonfortep In reference to this:

> It also addresses an issue where URLs with multiple path segments and a MIME type was guessing the first segment of the path as the filename, as opposed to the last.

IMO, this is relevant to the PR. While this PR itself contains no source code changes related to MIME types, URLs which use MIME types are beneficiaries of the contribution. Specifically, these two URLs from the description get incorrect filename guesses today due to the previous logic:

| URL      |  5.101.0 | This branch |
| ----------- | ----------- | ----------- | 
| https://d1x9zcshh2aa46.cloudfront.net/pdfs/example.pdf | pdfs.pdf | example.pdf |
| https://d1x9zcshh2aa46.cloudfront.net/pdfs/example | pdfs.pdf | example.pdf |

This is because they have multiple path segments, and there was no prior test coverage for responses containing multiple path segments and a MIME type (nearly all the tests don't provide a MIME type, which is totally OK but led to a gap in coverage). I believe the added test cases which specify a MIME type demonstrates the change, and lead to a more logical filename based on the response.

Hopefully that makes sense!

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
cmonfortep: Makes sense 👍

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
cmonfortep: Thanks for the explanation, I was not aware of that 👍

cmonfortep:
Thanks for the insights @asilvr. I will wait until we remove those warnings to merge the PR. This contribution is going to improve the user experience, and excited about it 🙌

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
asilvr: Removed this and it passed 🎉

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
asilvr: Fixed this and it passed 🎉

app/src/androidTest/java/com/duckduckgo/app/browser/downloader/UriUtilsFilenameExtractorTest.kt
asilvr: Fixed this and it passed 🎉

asilvr:
Should be good to go @cmonfortep! Sorry for the delay, I've been traveling for US holidays. 

Appreciate you helping facilitate this review+merge, and looking forward this user experience improvement. It was a pleasure to contribute!

PR #1550, opened 2021-11-18
Ensure AppTP restarts after an app update

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1201383124676579/f

### Description
During an app update, the app is killed while the APK is being swapped. The app is not automatically opened again afterwards, meaning the underlying VPNService in AppTP will not resume immediately.

By adding a BroadcastReceiver to listen for "app update" events, our process will be created and therefore the VPN Service will restart itself if it was previously running.

### Steps to test this PR

Add logcat filter: "Notified of app"

* Run the app and enable appTP
* Build the APK using `gradlew assembleID`
* Install it using `adb install -r`   (e.g., `adb install -r app/build/outputs/apk/internal/debug/duckduckgo-5.102.2-internal-debug.apk`)
* Verify that after a few seconds, appTP restarts (you'll see the VPN icon showing) and there is a logcat entry for being notified of an app update

### DIFF
--- a/vpn/src/main/AndroidManifest.xml
+++ b/vpn/src/main/AndroidManifest.xml
@@ -135,6 +135,12 @@
             </intent-filter>
         </receiver>
 
+        <receiver android:name="com.duckduckgo.mobile.android.vpn.appupdates.AppUpdateReceiver">
+            <intent-filter>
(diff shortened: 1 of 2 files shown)


aitorvs (approved):
I like noop methods :D
Awesome

PR #1551, opened 2021-11-18
Alternative method of determining originating app package

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1200811292912063/f

### Description
Changes the way we detect the originating app package to overcome some edge cases with the existing implementation.

### Steps to test this PR
 
- Enable appTP as normal
- Use some apps with known trackers to smoke test. Verify they show up correctly in the tracker dashboard
- Launch `Noom Health & Weight` app and `MX Player`. Verify both of these show up correctly in the tracker dashboard (these are two apps which weren't correctly identified using the existing method of obtaining app package)

### DIFF
--- a/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/requestingapp/DetectOriginatingAppPackageModern.kt
+++ b/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/requestingapp/DetectOriginatingAppPackageModern.kt
@@ -41,8 +41,20 @@ class DetectOriginatingAppPackageModern(
     }
 
     private fun getPackageIdForUid(uid: Int): String {
-        return packageManager.getNameForUid(uid) ?: OriginatingAppPackageIdentifierStrategy.UNKNOWN.also {
-            Timber.i("Failed to get package ID for UID: $uid")
+        val packages = packageManager.getPackagesForUid(uid)
+        if (packages.isNullOrEmpty()) {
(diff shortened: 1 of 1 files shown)


vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/requestingapp/DetectOriginatingAppPackageModern.kt
```diff
@@ -41,8 +41,20 @@ class DetectOriginatingAppPackageModern(
...
     private fun getPackageIdForUid(uid: Int): String {
-        return packageManager.getNameForUid(uid) ?: OriginatingAppPackageIdentifierStrategy.UNKNOWN.also {
-            Timber.i("Failed to get package ID for UID: $uid")
+        val packages = packageManager.getPackagesForUid(uid)
+        if (packages.isNullOrEmpty()) {
+            Timber.w("Failed to get package ID for UID: $uid")
+            return OriginatingAppPackageIdentifierStrategy.UNKNOWN
         }
+
+        if (packages.size > 1) {
```
CDRussell, line 50: if there are multiple packages, that's _interesting_ so we log it in the console (to maybe notice it in development) but not so interesting that we'd want anything more at this stage (e.g., a pixel)

in any case, we'll grab the `first` package from the list if multiple are returned, so there is a slight chance we'll mis-identify one app for another but I think it's still worth the change until we see examples where this happens

aitorvs (approved):
LGTM!

PR #1553, opened 2021-11-19
Only record pixel when different exception or over time threshold

joshliebe:
Task/Issue URL: https://app.asana.com/0/414730916066338/1182255336053236/f

### Description
This PR limits the frequency of duplicate exceptions from being stored into the `UncaughtException` DB. The time threshold between storing duplicates has been set to 1s.

### Steps to test this PR

This one is tricky to test but with some patience the issue can be simulated.

_To simulate several duplicate `UnsatisfiedLinkErrors` being sent, do the following:_
1. Checkout this branch.
2. Comment out the following in the app `build.gradle` (Line 94-99):

Screenshot 2021-11-19 at 08 09 25

3. Set the `TIME_THRESHOLD_MILLIS` in `UncaughtExceptionRepository` (Line 97) to `0L`.
4. Filter Logcat with "m_d_ac_g".
5. Run the app, it should crash on startup.
6. Keep opening the app until the system shows the "DuckDuckGo keeps crashing" dialog.
7. At this point several enqueued  `UnsatisfiedLinkError` pixels should be sent.
8. Verify that this is the case in Logcat.

_Now to test the fix, do the following:_
1. Set the `TIME_THRESHOLD_MILLIS` in `UncaughtExceptionRepository` (Line 97) to `10000L`.
2. Run the app, it should crash on startup.
3. Keep opening the app (With less than 10s between openings) until the system shows the "DuckDuckGo keeps crashing" dialog.
4. At this point only a single `UnsatisfiedLinkError` pixel should be sent.
5. Verify that this is the case in Logcat.

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/global/exception/UncaughtExceptionDaoTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/global/exception/UncaughtExceptionDaoTest.kt
@@ -115,6 +115,26 @@ class UncaughtExceptionDaoTest {
         }
     }
 
+    @Test
+    fun whenGetLatestExceptionThenReturnLatestExceptionFromDatabase() {
+        val exception1 = UncaughtExceptionEntity(id = 1, exceptionSource = GLOBAL, message = "foo", "version")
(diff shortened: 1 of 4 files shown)


joshliebe:
Thanks for the review @anikiki!

PR #1560, opened 2021-11-20
Add git hooks configuration to install script

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201403896220536/f

### Description
Add the git commit hooks configuration to the `install_tools.sh` script

### Steps to test this PR
Execute scripts and check hooks are setup correctly

### DIFF
--- a/install_tools.sh
+++ b/install_tools.sh
@@ -10,3 +10,5 @@ echo "Fetching ndk-bundle. Suppressing output to avoid travis 4MG size limit";
 sdkmanager "ndk-bundle" >/dev/null;
 echo "Fetching ndk-bundle complete";
 sdkmanager "system-images;android-22;default;armeabi-v7a";
+echo "Configuring pre-commit hooks"
+git config core.hooksPath .githooks


aitorvs:
* develop
  * **PR #1560**  (this PR)

CDRussell:
Not sure it makes sense adding the hook to this script. This script only really existed for setting up CI, and has a bunch of stuff in it that probably doesn't even make sense to have.

I get the desire for "one setup script" but that's not what this is in its current form. Maybe makes sense to create a different one specifically for your purpose?

aitorvs:
> Not sure it makes sense adding the hook to this script. This script only really existed for setting up CI, and has a bunch of stuff in it that probably doesn't even make sense to have any more.
> 
> I get the desire for "one setup script" but that's not what this is in its current form. Maybe makes sense to create a different one specifically for your purpose?

That interesting, because I used this script for setting up my environment twice, once when I first joined and this time when changed laptop.
Why is this script not suitable for setting up dev machines?

CDRussell:
I don't see a problem with it being added on here since it was actually useful for you in setting up your machine. It's probably less to do with it being suitable or not, and more that I just never considered it for that purpose. 

I had to configure some stuff manually before this script would run for me (related to [this](https://app.asana.com/0/0/1200861884946481/f)) hence why I never considered it before.

CDRussell (approved):

PR #1583, opened 2021-11-26
Catch CancelledKeyExceptions to prevent service crashing

CDRussell:
Task/Issue URL: https://app.asana.com/0/1199371158290272/1201433035781456/f

### Description
It is possible to experience a cancelled key exception when registering for events using the selector. These represent local issues with that connection and shouldn't propagate up to affect the `VPNService` running

### Steps to test this PR
This was initially observed when using [Lazada](https://play.google.com/store/apps/details?id=com.lazada.android&hl=en_US&gl=US), however it is intermittent there and unclear what the exact steps are to reproduce.

You can simulate what happens by manually cancelling the key.

**TcbDeviceToNetwork**, line 280
Add this: `tcb.selectionKey.cancel()`


aitorvs (approved):
This is an awesome catch :pray:

PR #1585, opened 2021-11-28
Allow configure local cache using local.properties file

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201433136298885/f

### Description
* Some time ago we enabled gradle local build cache
* Bugs in kotlin incremental compilation sometimes interfere with the local cache
* This generally happens when modifying Dagger components
* The changes are sometimes not picked up by the compiler
* The easy fix when that happens is to remove the local cache directory
* Given that this is a local cache, it makes sense to be able to enable/disable it locally through local.properties

This PR will:
* allow to enable/disable local build cache through a property in local.properties file
  * the name of the property will be local.build.cache and will take true or false values

### Steps to test this PR
1. remove `build-cache` directory from root project directory
1. run `./gradlew clean`
1. ensure the `local.properties` file DOES NOT contain the `local.build.cache` property
1. build the app, `./gradlew assembleID` for instance
1. verity the `build.cache` directory is NOT created
1. run `./gradlew clean`
1. set `local.build.cache = false` in `local.properties` file
1. build the app, `./gradlew assembleID` for instance
1. verity the `build.cache` directory is NOT created
1. set `local.build.cache = true` in `local.properties` file
1. build the app, `./gradlew assembleID` for instance
1. verity the `build.cache` directory IS now created

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -38,9 +38,31 @@ rootProject.children.each { subproject ->
 
 
 buildCache {
+
+    def getFile = { dir, filename ->
+        File file = new File("$dir$File.separator$filename")
+        file?.exists() ? file : null
+    }
+
+    def getLocalProperties = { dir ->
+        def file = getFile(dir, "local.properties")
+        if (!file) {
+            return null
+        }
+
+        Properties properties = new Properties()
+        properties.load(file.newInputStream())
+        return properties
+    }
+
     local {
+        def properties = getLocalProperties(rootDir)
+        if (properties != null) {
+            enabled = "true" == properties.getProperty("local.build.cache", "true")
+        } else {
+            enabled = true
+        }
         // configure local build cache directory so that it is local to the project dir
-        enabled = true
         directory = new File(rootDir, 'build-cache')
         removeUnusedEntriesAfterDays = 7
     }


aitorvs:
* develop
  * **PR #1585**  (this PR)

CDRussell:
@aitorvs it's unclear whether you intended to make local build cache opt in or opt out. the instructions suggest it's opt-in (disabled by default; "4. verity the build.cache directory is NOT created") but the code assumes it should use the build cache by default.

Can you clarify the intention and update either the test steps or the code to match?

settings.gradle
```diff
@@ -38,9 +38,31 @@ rootProject.children.each { subproject ->
...
+        properties.load(file.newInputStream())
+        return properties
+    }
+
     local {
+        def properties = getLocalProperties(rootDir)
+        if (properties != null) {
+            enabled = "true" == properties.getProperty("local.build.cache", "true")
+        } else {
+            enabled = true
```
CDRussell, line 63: As per top-level comment, this makes the build cache **enabled by default**. I think we're better making it disabled by default, but no strong opinions really

CDRussell (approved):
Noted that it's currently enabled by default (and we probably want to switch that), but confirmed that setting `local.build.cache=false` disables the build cache so that's working as expected

PR #1597, opened 2021-12-02
Avoid scoping errors in our Dagger setup

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201415414578549/f

### Description
This PR harmonises the way we will scope dependencies in Dagger so that we avoid past mistakes mixing scope annotations resulting in unpexpected behavior.
We have also ditched the `ObjectGraph` naming and use `Scope` all over.

Recommended to see the [asana task](https://app.asana.com/0/414730916066338/1201415414578549/f) for more information about this.

With this PR we will have the following dagger scopes
* `AppScope` -> bound to the life of the application
* `ActivityScope` -> should be used for Activities
* `FragmentScope` -> should be used for Fragments
* `QuickSettingsScope` -> only use the quick tile setting for AppTP
* `ReceiverScope` -> should be used for broadcast receivers
* `VpnScope` -> only used for Vpn service

_Exceptions_
We currently do NOT support nesting dagger scopes, this means that the `AppScope` is the parent of ALL remaining scopes and, ALL remaining scopes are siblings, ie.
* children scopes have access to the dependencies in the parent scope
* sibling scopes do not have access to the dependencies in other siblins

Because we don't yet support nesting, some vpn-related activities that SHOULD use the `ActivityScope` are actually using the `VpnScope`, all because they need some dependencies that are contributed to the `VpnScope`.
To track those for future fixes, we have added an annotation dubbed `@WrongScope` that provide information as to what should be the appropriate scope.
In the future, when nesting is supported, we just need to find usages of that annotation and fix the scope.

Incidentally, and because we are heavy users of multi-bindings and unfortunatelly those require the `JvmSuppressWildcards` annotations, I have added a sugar over `Set` and `Map` multibindings to avoid mistakes.

### Steps to test this PR
This PR should not change any behavior, just fixes the usage of dagger annotations. If anything breaks, should now break at build time (this is why the `WrongScope` annotation exists now)

* Just perform regular smoke tests for the app and AppTP

PR #1602, opened 2021-12-03
Implement Activity->Fragment dagger component nesting

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201432423603230/f

### Description
In PR #1597 we're harmonising the way we name dagger components and how we contribute dependencies to them. As a recap:
* Use @SingletonIn(AppScope::class) for app scope
* Use @SingletonIn(ActivityScope::class) for activity scope
* Use @SingletonIn(FragmentScope::class) for fragment scope

This a follow up PR to enable nesting those scopes:
* AppScope -> ActivityScope -> FragmentScope

Nesting means:
* child scopes can get access to dependencies in their parent scopes (but not the other way around), ie.
* `FragmentScope` would get access to deps in Fragment, Activity and App scopes
* `ActivityScope` would get access to deps in Activity and App scopes
* `AppScope` would get access to deps in App scope

The PR changes the fragments that were not in the `FragmentScope` to use that one instead. Functionality should not be affected at all

### Steps to test this PR
Smoke tests for app and AppTP

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/di/component/AppEnjoymentDialogFragmentComponent.kt
+++ b/app/src/main/java/com/duckduckgo/app/di/component/AppEnjoymentDialogFragmentComponent.kt
@@ -17,7 +17,7 @@
 package com.duckduckgo.app.di.component
 
 import com.duckduckgo.app.browser.rating.ui.AppEnjoymentDialogFragment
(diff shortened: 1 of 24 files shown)


aitorvs:
* develop
  * **PR #1597** 
    * **PR #1602**  (this PR)

PR #1603, opened 2021-12-03
Upgrade WorkManager to 2.7.1

marcosholgado:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201445526459254

### Description
This PR upgrades Work Manager to the latest version.

### Steps to test this PR

_Make sure Work Manager still works!_
Some areas to check
1. Remote privacy config is downloaded correctly
1. Notifications work as expected (change the notifications scheduler to a few seconds rather than day to test this)
1. Overall filter for "work" in the logcat to check everything is being scheduled.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -28,11 +28,16 @@
 
         <!-- Disable WorkManager's default initialization so we can use Dagger with it -->
         <provider
-            android:name="androidx.work.impl.WorkManagerInitializer"
-            android:authorities="${applicationId}.workmanager-init"
-            android:enabled="false"
+            android:name="androidx.startup.InitializationProvider"
+            android:authorities="${applicationId}.androidx-startup"
             android:exported="false"
-            tools:replace="android:authorities" />
+            tools:node="merge">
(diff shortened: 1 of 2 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -28,11 +28,16 @@
...
+            android:authorities="${applicationId}.androidx-startup"
             android:exported="false"
-            tools:replace="android:authorities" />
+            tools:node="merge">
+            <!-- If you are using androidx.startup to initialize other components -->
+            <meta-data
+                android:name="androidx.work.WorkManagerInitializer"
+                android:value="androidx.startup"
+                tools:node="remove" />
+        </provider>
```
aitorvs, line 40: Just curious, why do we need to add the `metada` now?

aitorvs (commented):
Also checked some of the workers in the VN and also seems to work as expected

app/src/main/AndroidManifest.xml
marcosholgado: It is in the documentation (https://developer.android.com/jetpack/androidx/releases/work), I actually tried without it and didn't work.

PR #1607, opened 2021-12-06
Use a Provider to lazily initialize the default user agent

marcosholgado:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201472407198007

### Steps to test this PR
We currently try to access the default user agent from `WebSettings` on startup. This is not necessary and having two side effects:
1. Delaying our startup time and hence being a factor in the cause of ANRs
2. Crash the app in some cases where the WebView process is not ready yet.

_User agent creation correctly referred_
1. Go to the task and apply the `Reproduce_ANR.patch` patch. This will create a new trace when the app launches.
2. Download https://github.com/Grigory-Rylov/android-methods-profiler to analyze the trace.
3. Launch the app, wait a few seconds and then close the app.
4. From the device file explorer, go to `data/data/com/duckduckgo.com.mobile.android.debug` and download the trace created.
5. Analyze the trace with Android Methods Profiler.
6. Search for `Chromium`
7. You shouldn't not see any instances created under the `com.duckduckgo.app.di.DaggerAppComponent.pixel`
8. You can also analyze the `prod.trace` trace included in the task to see the difference.

_User agent gets created correctly_
1. Launch the app
1. Filter the logcat by `user agent`.
1. Go to SERP, you should see a log message with the user agent.
1. Change to desktop mode, you should see the desktop user agent.

Note: The user agent log might be duplicated, this is expected.

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/WebViewRequestInterceptorTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/WebViewRequestInterceptorTest.kt
@@ -57,7 +57,7 @@ class WebViewRequestInterceptorTest {
     private val mockPrivacyProtectionCountDao: PrivacyProtectionCountDao = mock()
     private val mockGpc: Gpc = mock()
     private val mockWebBackForwardList: WebBackForwardList = mock()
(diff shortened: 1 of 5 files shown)


aitorvs (approved):
Awesome, work as expected :pray:

PR #1608, opened 2021-12-06
AppTP Health Monitor

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1201527834957243/f

### Description
Adds an automated health monitor for app TP, only for `internal` builds currently. The monitor will start and stop with the `VPNService` lifecycle. 

`Internal` builds will be able to view additional diagnostics in the diagnostics activity, as well as having access to a report button.

#### Bad health alert
When the health monitor detects a problem, and it doesn't seem to be clearing, it will throw up a notification alerting that the health might be degraded. Clicking on that notification will show the diagnostics screen, where a health report can be captured and sent for analysis.

The rules for what constitutes bad health will need to evolve as we learn more, as they are somewhat arbitrary right now. See  `HealthClassifier` for these rules.

#### Health metrics
When it's running, it will monitor metrics, such as:

- current memory usage
- number of socket exceptions
- number of exceptions writing to the TUN
- tracer packets (see below)

#### Tracer packets
When the monitor is running, it will frequently drop in a special _tracer_ `Packet` and measure the length of time it takes for that packet to move through the system. The number of successful/failed tracers will be used to trigger a bad health alert if they start to slow down, or stop reaching their final destination.

#### General 
- [ ] Test it doesn't automatically run when not an `internal` build

#### Testing `internal` builds

Suggested logcat filter: `AppTp Health|Cleaning up old health metrics|Sending health report|Pixel sent: m_atp_breakage_report|tracer`

- [x] Ensure app monitor does **not** start when the VPN service isn't running
- [x] Ensure monitor **does** start when the VPN service starts
- [x] Launch `DiagnosticsActivity` (available from main AppTP activity->overflow). Verify health stats are updating frequently.
- [x] Click the `Report` button, then hit back. Verify no report sent. (nothing will show with suggested logcat filter)
- [x] Click the `Report` button, then hit `Submit` leaving options untouched. Verify JSON contains system health report and user report (which shows empty notes and "undetermined" health since user didn't specify current health)
- [x] Click the `Report` button Choose "Everything seems fine" and submit. Verify JSON contains system health report and user report (which shows "good" health)
- [x] Click the `Report` button. Choose "Something's not working" and submit. Verify JSON contains system health report and user report (which shows "bad" health)
- [x] Click the `Report` button. Add some notes and submit. Verify JSON contains system health report and user report (which shows "notes" populated)

#### Triggering bad health

Test bad health by adding some pretend bad health metrics.

E.g., in `TunPacketWriter`, line 72, add `healthMetricCounter.onTunWriteIOException()` (this will record an exception every time there is a tracer packet, ie, every ~5 seconds)

- [x] Verify the diagnostics screen shows the red sad face to indicate it's detecting bad health
- [x] Verify the notification doesn't show yet. This won't show until there's been a prolonged bad health event, lasting across multiple samples. 
- [x] Wait until it's detected as a prolonged event, and verify the notification is triggered
  - One sample taken every ~30s, so it might take 1 or 2 mins to be detected as prolonged. 
  - If you're impatient, tweak the constants in `AppTPHealthMonitor` for `MONITORING_FREQUENCY_MS` and how many samples to wait when instantiating each of the `HealthRule`s (e.g., passing `samplesToWaitBeforeAlerting = 2` in constructor)
- [x] Verify clicking on the notification launches the diagnostics screen


vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
```diff
@@ -135,7 +154,7 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope(), N
                 returnCode = Service.START_REDELIVER_INTENT
             }
             ACTION_STOP_VPN -> {
-                stopVpn(VpnStopReason.SelfStop)
+                launch { stopVpn(VpnStopReason.SelfStop) }
```
CDRussell, line 157: This helps prevent an ANR

aitorvs (approved):
Works as expected. This is great!

- I have added one change because we're sending the reports under the app breakage pixel and not the new one added for health.
- also what do you think about taking the opportunity that we're touching both `TunPacketReader.kt` and `TunPacketWriter.kt` to add them to our DI infra? (currently they're both out and we need to create them manually from the tracker service.
I leave here a patch with that change in case you want to apply it, ie:
  - `$ pbcopy < file.txt`
  - In Intellij chose "Apply patch from clipboard
- I have perform the tests with and without the patch apply just in case


[file.txt](https://github.com/duckduckgo/Android/files/7734209/file.txt)

CDRussell:
> Works as expected. This is great!
> 
> * I have added one change because we're sending the reports under the app breakage pixel and not the new one added for health.
> * also what do you think about taking the opportunity that we're touching both `TunPacketReader.kt` and `TunPacketWriter.kt` to add them to our DI infra? (currently they're both out and we need to create them manually from the tracker service.
>   I leave here a patch with that change in case you want to apply it, ie:
>   
>   * `$ pbcopy < file.txt`
>   * In Intellij chose "Apply patch from clipboard
> * I have perform the tests with and without the patch apply just in case
> 
> [file.txt](https://github.com/duckduckgo/Android/files/7734209/file.txt)

This is a great idea. And thanks for doing the work for it and making it easy to integrate; much appreciated 🙏

PR #1609, opened 2021-12-07
Remove useless call to getWorkInfosByTag to reduce app onCreate execution time

aitorvs:
Task/Issue URL: https://app.asana.com/0/0/1201472407198058/f

### Description
We're seeing an increase of the app `onCreate` execution time. Some of the lifecycle observers may have a role in that increase. In particular the `TrackersDbCleanerScheduler` class was executing some useless code to try skip scheduling a worker when it was already scheduled.
That resulted in calling `getWorkInfosByTag` which is rather slow.

We can safely remove that call as we're already using `enqueueUniquePeriodicWork` in cobination with `KEEP` policy, which achieves the same

### Steps to test this PR
1. install from this branch
1. open the app
1. VERIFY `Registering application lifecycle observer: com.duckduckgo.app.privacy.cleanup.TrackersDbCleanerScheduler` appears in the logcat


marcosholgado (approved):
Nice one! 👍

PR #1616, opened 2021-12-10
Revert continuation indent to be 4

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201492317527942/f

### Description
In [this](https://app.asana.com/0/414730916066338/1201355216612743) task we make some improvements but changed some of the IDE configs we have before, like the continuation indent that went from 4 -> 8

This PR reverts that value to be 4

### Steps to test this PR
1. pull this branch
1. Open the IDE
1. verify that continuation indents are not massive anymore

PR #1617, opened 2021-12-10
Move the pixel dependency out of application class

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201464810536290/f

### Description
* The current app onCreate path takes longer than it should
* One of the main contributors is the pixel dependency in the app class
* The pixel dependency is used to:
  * send the mf pixel onCreate
  * send the ml pixel onStart
* the ml pixel doesn't affect the onCreate path
* the mf pixel doesn't need to be sent at onCreate, and can be scheduled to be sent later

This PR removes the pixel dependency from the app class by:
* creating a LifecycleEventObserver that will:
  * send the ml pixel onStart--except when the app is launched as a result of a fire button press
  * schedules the sending of the mf pixel

### Steps to test this PR
_test the ml pixel_
* install from this branch
* launch the app
* verify the ml pixel is sent
* press the fire button
* verify the ml pixel is NOT sent upon app launch

_test the mf pixel_
* in EnqueuedPixelWorker change KEEP for REPLACE in line 110
* build and install from this branch
* launch the app
* verify the mf pixel is NOT sent
* press the fire button
* verify the mf pixel is sent once

In addition, when taking a startup trace, the app onCreate path should not contain the pixel dependency anymore.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
@@ -17,9 +17,7 @@
 package com.duckduckgo.app.global
 
 import android.app.Application
-import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.OnLifecycleEvent
 import androidx.lifecycle.ProcessLifecycleOwner
 import com.duckduckgo.app.browser.BuildConfig
 import com.duckduckgo.app.di.AppComponent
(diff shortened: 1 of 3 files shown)


aitorvs:
* develop
  * **PR #1616** 
    * **PR #1617**  (this PR)

marcosholgado:
FYI `In addition, when taking a startup trace, the app onCreate path should not contain the pixel dependency anymore.` I can still see the pixel dependency during onCreate. I've attached a trace in the Asana task.

marcosholgado (approved):
Nice job!

PR #1626, opened 2021-12-15
Capture default UserAgent crashes

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1201519018374192/f

### Description
This PR intends to capture a crash when getting WebView default user agent on some devices.

### Steps to test this PR

_No changes_
1. run the app
2. visit any website
3. Filter logs by `User Agent is`
4. Ensure your UA matches the one on your device (e.g: check if chrome version is the one installed on your device)

_Capture crash_
1. Go to `provideDefaultUserAgent` method in `UserAgentProvider`
2. hardcode `throw RuntimeException("")`  before `WebSettings.getDefaultUserAgent(context)`
3. run the app
4. visit any website
5. Ensure app does not crash

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/useragent/UserAgentProvider.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/useragent/UserAgentProvider.kt
@@ -22,6 +22,8 @@ import android.webkit.WebSettings
 import androidx.core.net.toUri
 import com.duckduckgo.app.global.UriString
 import com.duckduckgo.app.global.device.DeviceInfo
+import com.duckduckgo.app.pixels.AppPixelName
+import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesTo
 import dagger.Module
@@ -37,7 +39,11 @@ import javax.inject.Provider
  * Example Default Desktop User Agent (From Chrome):
  * Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.137 Safari/537.36
  */
-class UserAgentProvider constructor(@Named("defaultUserAgent") private val defaultUserAgent: Provider<String>, private val device: DeviceInfo) {
+class UserAgentProvider constructor(
+    @Named("defaultUserAgent") private val defaultUserAgent: Provider<String>,
+    private val device: DeviceInfo,
+    private val pixel: Pixel
+) {
 
     private val baseAgent: String by lazy { concatWithSpaces(mobilePrefix, getWebKitVersionOnwards(false)) }
     private val baseDesktopAgent: String by lazy { concatWithSpaces(desktopPrefix, getWebKitVersionOnwards(true)) }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/useragent/UserAgentProvider.kt
```diff
@@ -111,6 +127,7 @@ class UserAgentProvider constructor(@Named("defaultUserAgent") private val defau
         const val SPACE = " "
         val mobilePrefix = "Mozilla/5.0 (Linux; Android ${Build.VERSION.RELEASE})"
         val desktopPrefix = "Mozilla/5.0 (X11; Linux ${System.getProperty("os.arch")})"
+        val fallbackDefaultUA = "$mobilePrefix AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/96.0.4664.104 Mobile Safari/537.36"
```
cmonfortep, line 130: I'm using the latest chrome version here as part of our UA. should not be an issue. If anyone knows a better way to handle this, please share.

app/src/main/java/com/duckduckgo/app/browser/useragent/UserAgentProvider.kt
```diff
@@ -96,10 +102,20 @@ class UserAgentProvider constructor(@Named("defaultUserAgent") private val defau
...
+    private fun defaultUserAgent(): String = runCatching {
+        defaultUserAgent.get()
+    }.let { result ->
+        if (result.isFailure) {
+            pixel.fire(AppPixelName.WEBVIEW_DEFAULT_USER_AGENT_CRASH)
+            return fallbackDefaultUA
+        }
+        return@let result.getOrDefault(fallbackDefaultUA)
+    }
+
```
aitorvs, line 118: I think this should rather be in the `BrowserModule`. There is no need to make changes to this calss.

```kotlin
    @Provides
    fun userAgentProvider(@Named("defaultUserAgent") defaultUserAgent: Provider<String>, deviceInfo: DeviceInfo): UserAgentProvider {
        try {
            return UserAgentProvider(defaultUserAgent, deviceInfo)
        } catch (e: Exception) {
            return ""
        }
    }
```

malmstein:
@cmonfortep besides the fact that it's broken,  can you make sure you fill in the PR template? All the relevant info is missing

cmonfortep:
Thanks, @malmstein, this is still in draft. I will update the description and fix the build before is ready for review.

app/src/main/java/com/duckduckgo/app/browser/useragent/UserAgentProvider.kt
marcosholgado: Did you mean change the `DefaultUserAgentModule`? :P

```
@ContributesTo(AppScope::class)
@Module
class DefaultUserAgentModule {
    @Provides
    @Named("defaultUserAgent")
    fun provideDefaultUserAgent(context: Context): String {
       try {
           return WebSettings.getDefaultUserAgent(context)
        } {catch (e: Exception) {
            return ""
        }
    }
}

marcosholgado: Yup 👍  to this

cmonfortep: I'm ok not changing the class, however, I don't think is a good idea to provide an empty string as a fallback.

WDTY about wrapping `WebSettings.getDefaultUserAgent(context)` into in our own class that will provide a default UA?

marcosholgado: I think the empty string was an example, we should return the actual UA there.

cmonfortep: I've discussed this with @aitorvs. If we don't need to send a pixel, we can follow this approach.

marcosholgado (approved):
Nice job! 🚢  🇮🇹

PR #1630, opened 2021-12-16
Fix/david/quick settings tile crash

malmstein:
Task/Issue URL: https://app.asana.com/0/1157893581871903/1201501533153168

### Description
We're seeing a crash from the Quick Settings tile because we are using a null value.


aitorvs (approved):
Awesome, works as expected :pray:

PR #1634, opened 2021-12-20
Upgrade to Kotlin 1.6.10

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1201539840759397/f

### Description
Upgrades the kotlin version. This creates some compilation errors, and an incompatibility with Anvil so that is updated too.

|Library|Before|After|Notes|
|---|---|---|---|
|Kotlin|1.5.10|1.6.10||
|Anvil|2.3.3|2.3.10-1-6-0|See [Anvil releases](https://github.com/square/anvil/releases) to explain prefix

### Steps to test this PR

- [ ] Ensure `./gradlew checks` passes (ignoring known flaky tests if they present)
- [ ] Smoke test the app to ensure no regressions have slipped in

### DIFF
--- a/vpn/src/main/java/dummy/ui/VpnControllerActivity.kt
+++ b/vpn/src/main/java/dummy/ui/VpnControllerActivity.kt
@@ -31,7 +31,11 @@ import com.duckduckgo.app.trackerdetection.api.WebTrackersBlockedRepository
 import com.duckduckgo.app.trackerdetection.db.WebTrackerBlocked
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.mobile.android.vpn.R
-import com.duckduckgo.mobile.android.vpn.model.*
+import com.duckduckgo.mobile.android.vpn.model.DataSizeFormatter
+import com.duckduckgo.mobile.android.vpn.model.TimePassed
+import com.duckduckgo.mobile.android.vpn.model.VpnTracker
+import com.duckduckgo.mobile.android.vpn.model.dateOfLastWeek
+import com.duckduckgo.mobile.android.vpn.model.dateOfPreviousMidnight
 import com.duckduckgo.mobile.android.vpn.stats.AppTrackerBlockingStatsRepository
 import com.duckduckgo.mobile.android.vpn.stats.AppTrackerBlockingStatsRepository.DataTransfer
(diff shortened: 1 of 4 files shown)


vpn/src/main/java/dummy/ui/VpnControllerActivity.kt
```diff
@@ -31,7 +31,11 @@ import com.duckduckgo.app.trackerdetection.api.WebTrackersBlockedRepository
 import com.duckduckgo.app.trackerdetection.db.WebTrackerBlocked
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.mobile.android.vpn.R
-import com.duckduckgo.mobile.android.vpn.model.*
+import com.duckduckgo.mobile.android.vpn.model.DataSizeFormatter
+import com.duckduckgo.mobile.android.vpn.model.TimePassed
+import com.duckduckgo.mobile.android.vpn.model.VpnTracker
+import com.duckduckgo.mobile.android.vpn.model.dateOfLastWeek
+import com.duckduckgo.mobile.android.vpn.model.dateOfPreviousMidnight
```
CDRussell, line 38: Interesting breakage when using `com.duckduckgo.mobile.android.vpn.model.*` which we had before: Anvil gets confused which `VpnPreferences` class to use... there is one we explicitly import and another that is brought in because of that wildcard.

anikiki: I think we have a rule to disallow star imports, not sure if it is enforced.

anikiki (approved):
Thanks for doing the upgrade! :tada: Didn't find any issues during smoke testing and `./.gradlew checks` worked as expected.

vpn/src/main/java/dummy/ui/VpnControllerActivity.kt
aitorvs: I think * imports should be avoided. Not sure if we've got a rule but I'd love to have it an enforce it 😬

CDRussell: I'll leave y'all to have that discussion in the new year and look forward to coming back to a lovely formatted codebase 😎

PR #1648, opened 2022-01-05
remove duplicate styles, ensure InfoPanel is used properly

malmstein:
Task/Issue URL: https://app.asana.com/0/488551667048375/1201027352531949/f

### Description
With AppTP, we introduced a new component called "InfoPanel". However, this component was not standarized across all screens that use it. This PR fixes that


joshliebe (approved):
LGTM ✅

PR #1650, opened 2022-01-06
ANR supervisor

aitorvs:
Task/Issue URL: https://app.asana.com/0/0/1201618744517735/f

### Description
Adding the ANR supervisor. More info [here](https://app.asana.com/0/1198194956794324/1201466259640556).

* the ANR code lives as an independent gradle library
  * anrs/ansr-api
  * anrs/anrs-impl
  * anrs/anrs-store
* the ANR supervisor hooks as a `BrowserLifecycleObserver` to monitor only when the app is opened
* supervisor starts in onOpen (app is opened)
* supervisor stops in onClose (app is closed)
* the ANR supervisor challenges the UI thread every ~5 seconds
* the ANR supervisor waits for ~2 seconds for the UI thread response
* if response doesn't come, ANR exception will be stored in a local DB
* every ANR exception has a unique MD5 hash so that we don't store duplicates
  * the unique hash is calculated using the stack trace
* using the `OfflinePixelWorker` the ANRs will be sent one by one
  * ie. for every execution of the `OfflinePixelWorker`, one ANR from the DB will be sent (and remove from the db)
* the ANR exceptions will be sent in a LIFO fashion
* an internal feature to force ANRs will be added to the "Dev Settings" so that we can test it

### Steps to test this PR

_Test ANR supervisor start stop_
- [x] open the app
- [x] verify the message `AnrSupervisor checking for ANRs...` appears periodicaly in logcat
- [x] close the app
- [x] vefiry `AnrSupervisor stopping...` log appears in logcat

_Test ANR pixel_
- [x] in this branch, change the policy to `ExistingPeriodicWorkPolicy.REPLACE` in `OfflinePixelScheduler:60`
- [x] build and install
- [x] open the app
- [x] go to "Settings" -> "Dev settings"
- [x] trigger ANR
- [x] verify the log error message `ANR Detected: ...` appears in the logcat with a very comprehensive stacktrace for all threads running in the app
- [x] close and reopen the app
- [x] verify the `m_anr_exception` pixel is sent and contains a parameter `stackTrace` with a base64 value
- [x] decode the base64 value - you can use [this website](https://onlinestringtools.com/convert-base64-to-string)
- [x] verify the stack trace for just the main thread is sent and it correctly identifies the point where the ANR originated


aitorvs:
* develop
  * **PR #1650**  (this PR)

statistics/src/main/java/com/duckduckgo/app/statistics/api/OfflinePixelSender.kt
```diff
@@ -55,7 +54,10 @@ constructor(
                 sendCookieDatabaseOpenErrorPixel(),
                 sendCookieDatabaseDeleteErrorPixel(),
                 sendCookieDatabaseCorruptedErrorPixel(),
-                sendUncaughtExceptionsPixel()))
+                sendUncaughtExceptionsPixel(),
+                *offlinePixels.map { it.send() }.toTypedArray()
```
marcosholgado, line 58: I take the idea here will be to remove all the other methods that are called here and instead use the new `OfflinePixel` interface for all the other offline pixels?

marcosholgado (approved):
Nice job! Working as expected!

statistics/src/main/java/com/duckduckgo/app/statistics/api/OfflinePixelSender.kt
aitorvs: Yep, that's the idea. Will create the task for it

arunsudharsan:
Awesome stuff @aitorvs !  

Given that we're using LIFO to retrieve and send the most recent ANR to the backend and subsequently remove it from the database, wouldn't the remaining ANRs stored in the database within the 3-hour window become outdated and persist indefinitely? This has the risk of potential increase in the database size in the long term right?

PR #1651, opened 2022-01-06
Bug: Choose default browser displayed twice

anikiki:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201620824138997/f

### Description
The "Choose default browser" dialog is displayed twice when you switch the app from foreground to background and back and then you tap on "I've been here before". See the video attached to the Asana task.

### Steps to test this PR

- checkout this branch
- tweak the variant manager by changing the weight of "zk" to 0.0 in VariantManager.kt
- install
- on Welcome Page screen you should see 2 buttons "I'm new!" and "I've been here before." Don't tap any of them yet.
- while on Welcome Page screen put the app in background and back to foreground a couple of times
- on Welcome Page screen tap on "I've been here before." and you should see a dialog to choose the default browser. Once you choose you'll not see the same dialog again, you'll see the next screen.


aitorvs (approved):
Nice catch @anikiki !!
Works as expected!

PR #1661, opened 2022-01-10
Move GPC headers to remote config

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1201624035922285

### Description
This PR deletes the hardcoded headers list for GPC in the app and instead uses the list from the remote config.

### Steps to test this PR
_Headers sent when navigating via links_
1. Install and launch the app
2. Go to https://globalprivacycontrol.org 
3. On the top of the site it should say `GPC signal detected`
4. Click below where it says `Test against the reference server`
5. Header should be present in the next site.


anikiki (approved):
Looks good and works as expected :tada:

PR #1663, opened 2022-01-11
Add webview version to ml pixel

karlenDimla:
Task/Issue URL: https://app.asana.com/0/414730916066338/1157223223343562/f

### Description
This PR adds the following:

1.) A provider for the webview's major version that uses the `WebViewCompat.getCurrentWebViewPackage` as the underlying implementation.
2.) Adds the "`webview_version`" as a parameter for the `ml` pixel.

### Steps to test this PR

1. Install and launch the debug version of the app from this branch
2. FIlter logcat by `V/RxBasedPixel: Pixel sent: ml`
3. The filtered message should include the webview_version=X in the params. Example: `V/RxBasedPixel: Pixel sent: ml with params: {webview_version=91} {}`
4. For API level 26 and below, the `webview_version` could be `unknown` in some cases.

### Testing
Needs to be tested in API Level 25 or below.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2022 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser
+
+import android.content.Context
+import androidx.webkit.WebViewCompat
+import com.duckduckgo.app.browser.WebViewVersionProvider.Companion.WEBVIEW_UNKNOWN_VERSION
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+
+interface WebViewVersionProvider {
+    companion object {
+        const val WEBVIEW_UNKNOWN_VERSION = "unknown"
+    }
+
+    fun get(): String
+}
+
+interface RawWebViewVersionProvider {
+    fun get(): String?
+}
+
+@ContributesBinding(AppScope::class)
+class WebViewCompatRawWebViewVersionProvider @Inject constructor(
+    private val context: Context
+) : RawWebViewVersionProvider {
+
+    override fun get(): String? = WebViewCompat.getCurrentWebViewPackage(context)?.versionName
+
+}
+
+@ContributesBinding(AppScope::class)
+class MajorWebViewVersionProvider @Inject constructor(
+    private val rawWebViewVersionProvider: RawWebViewVersionProvider
+) : WebViewVersionProvider {
+    companion object {
+        const val WEBVIEW_VERSION_DELIMITER = "."
+    }
+
+    override fun get(): String = rawWebViewVersionProvider.get().run {
+        if (this.isNullOrEmpty()) {
+            WEBVIEW_UNKNOWN_VERSION
+        } else this.captureMajorVersion()
+    }
+
+    private fun String.captureMajorVersion() = this.split(WEBVIEW_VERSION_DELIMITER)[0]
+}
--- a/app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
@@ -16,8 +16,6 @@
 
 package com.duckduckgo.app.browser
 
-import android.content.Context
-import androidx.webkit.WebViewCompat
 import com.duckduckgo.app.browser.WebViewVersionProvider.Companion.WEBVIEW_UNKNOWN_VERSION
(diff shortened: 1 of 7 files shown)


app/src/test/java/com/duckduckgo/app/pixels/EnqueuedPixelWorkerTest.kt
```diff
@@ -73,22 +80,24 @@ class EnqueuedPixelWorkerTest {
...
         whenever(unsentForgetAllPixelStore.pendingPixelCountClearData).thenReturn(1)
         whenever(unsentForgetAllPixelStore.lastClearTimestamp).thenReturn(System.currentTimeMillis())
+        whenever(webViewVersionProvider.get()).thenReturn("91")
 
         enqueuedPixelWorker.onStateChanged(lifecycleOwner, Lifecycle.Event.ON_CREATE)
         enqueuedPixelWorker.onStateChanged(lifecycleOwner, Lifecycle.Event.ON_START)
         enqueuedPixelWorker.onStateChanged(lifecycleOwner, Lifecycle.Event.ON_START)
 
-        verify(pixel).fire(AppPixelName.APP_LAUNCH)
+        verify(pixel).fire(AppPixelName.APP_LAUNCH, mapOf("webview_version" to "91"))
```
cmonfortep, line 101: any reason why you prefer using a string instead of `PixelParameter.WEBVIEW_VERSION` in the test?

app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+        const val WEBVIEW_VERSION_DELIMITER = "."
+    }
+
+    override fun get(): String = rawWebViewVersionProvider.get().run {
+        if (this.isNullOrEmpty()) {
+            WEBVIEW_UNKNOWN_VERSION
+        } else this.captureMajorVersion()
+    }
+
+    private fun String.captureMajorVersion() = this.split(WEBVIEW_VERSION_DELIMITER)[0]
```
cmonfortep, line 61: can this be empty? or null? or crash?

app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+package com.duckduckgo.app.browser
+
+import android.content.Context
+import androidx.webkit.WebViewCompat
+import com.duckduckgo.app.browser.WebViewVersionProvider.Companion.WEBVIEW_UNKNOWN_VERSION
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+
+interface WebViewVersionProvider {
```
cmonfortep, line 26: this is similar to the logic we have in `BrokenSiteActivity`, should we update that class to use this new provider?

app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+    companion object {
+        const val WEBVIEW_UNKNOWN_VERSION = "unknown"
+    }
+
+    fun get(): String
+}
+
+interface RawWebViewVersionProvider {
```
cmonfortep, line 34: I don't see why we have 2 interfaces with one implementation each. Should not be better to have 1 interface with 2 implementations? what do you think?

cmonfortep (commented):
Thanks @karlenDimla. This looks really nice. I didn't have time to test this yet, so I'm just posting a few comments I had in case you want to take a look at them.

I will test this later and let you know.

app/src/test/java/com/duckduckgo/app/pixels/EnqueuedPixelWorkerTest.kt
karlenDimla: It's mostly for readability and knowing exactly what to expect. If I had done this through TDD (disclaimer: i didn't), I would know that we wanted to have a paramter "webview_version" before I would know where and how to define it in the implementation.

app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
karlenDimla: Ah good question. This can definitely be empty if the version starts with a `.` for some reason. I don't think it will be null since: 1.) `this` is always non-null and not empty. 2.) if the delimiter doesn't exist - it will return the full value of `this`.

Is this case something we should handle?

app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
karlenDimla: Hmm, I thought about this but oddly enough the pixel param being sent for broken site reporting expects: `wvVersion=91.0.4472.114` which is passing the whole version not only the major one. Since the use case of this sounds pretty reasonable - as in this is only logged when reporting that a site is broken, is there anywhere we can check to know if we should pass the major version only here as well?

app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
cmonfortep: we talked about this, and clarified why we had 2 different interfaces. Decided to have an interface(provider) with 2 methods (full webview version and major version), and to create another interface to wrap the access to webviewCompat (datasource)

app/src/main/java/com/duckduckgo/app/browser/WebViewVersionProvider.kt
cmonfortep: Discusses about this. Decided that we should return `unknown` when `this.split(WEBVIEW_VERSION_DELIMITER)[0]` return empty or a value that can't be parsed as Int

cmonfortep (approved):
Thanks, @karlenDimla. Good job here. I'm excited about your first PR being merged. kudos!

PR #1670, opened 2022-01-13
AppTP: Ensure Enable VPN Switch uses proper API

malmstein:
Task/Issue URL: https://app.asana.com/0/1199371158290272/1201656298066485/f

### Description
Description
One of the issues described in Bug: AppTP can't be toggled ON is that the switch to enable the VPN doesn't work properly. 

Looking at the code, we are currently using the wrong API to listen to the widget updates.

Objective
Ensure we use the proper API for the Switch,  that is the CheckedChangeListener instead of the ClickListener

### Steps to test this PR

_Enable VPN_
- [x] With the VPN Disabled
- [x] Open Tracker screen and use the Switch
- [x] VPN should be enabled

_Disable VPN_
- [x] With the VPN Enabled
- [x] Open Tracker screen and use the Switch
- [x] VPN should be disabled


aitorvs (approved):
Nice, work as expected.

PR #1672, opened 2022-01-13
Add reported app default protection info to breakage report

aitorvs:
Task/Issue URL: https://app.asana.com/0/1199371158290272/1201539333793164/f

### Description
During the breakage it is benefitial to know whether the reported app was enabled by default of not.

This PR adds a field inside the `vpnTrackerLists` JSON object that contains whether the package ID was unprotected (ie. inside the exclusion list) or not. 
The new field is called `reportedAppUnprotectedByDefault` and can take a "true" or "false" value.

### Steps to test this PR

_Test_
- [x] enable AppTP
- [x] go to App tracking protection scree and click on "report an issue"
- [x] select any app and report
- [x] verify the breakage metadata contains the field `reportedAppUnprotectedByDefault` inside the `vpnTrackerLists` object
- [x] verify the field is `true` or `false` depending on whether the reported app is in the exclusion list or not

### DIFF
--- a/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/bugreport/VpnAppTrackerListInfoCollector.kt
+++ b/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/bugreport/VpnAppTrackerListInfoCollector.kt
@@ -17,8 +17,11 @@
 package com.duckduckgo.mobile.android.vpn.bugreport
 
 import com.duckduckgo.di.scopes.VpnScope
+import com.duckduckgo.mobile.android.vpn.apps.AppCategory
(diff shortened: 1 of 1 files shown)


aitorvs:
* develop
  * **PR #1672**  (this PR)

GuiltyDolphin:
@aitorvs will this catch games too? I'd like to know whether we would protect by default at all (whether because we detect a game, exclusion list, etc.)

aitorvs:
> @aitorvs will this catch games too? I'd like to know whether we would protect by default at all (whether because we detect a game, exclusion list, etc.)

oh yeah, forgot about that. It does now. Thanks for the nudge.

malmstein (approved):
:shipit:

PR #1689, opened 2022-01-19
Promote Add Widget in Settings

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1201677824560317/f

### Description
This PR adds the following:

- Add a new entry in Setting page for "Home Screen Widget" that allows the user to add widget through the settings page
- Send a new Pixel `ms_add_home_screen_widget_clicked` when the new setting item is clicked.
- Introduces a reusable `AddWidgetLauncher` that contains the logic on when to use the `AppWidgetManager` vs the `AddWidgetInstructions`.
- Use the `AddWidgetLauncher` in `BrowseTabFragment`.
- Introduce `WidgetAddedReceiver` that shows a toast whenever a widget is added through the app (both CTA and Settings)
- Changed the label for the widgets to "Favorites" and "Search" instead of "Favorites Widget" and "Search Widget"

### Steps to test this PR

Scenario 1: User has a device that supports automatically adding a widget.

1. Open Menu > Settings and click on "Home Screen Widget"
2. Launcher should be shown and allow the user to add the widget.
3. Logcat should show a message: `V/RxBasedPixel: Pixel sent: ms_add_home_screen_widget_clicked with params: {} {}`
4. If the user adds the widget, the system should show a toast similar or along the lines of: "Favorites widget added to home screen". Also confirm that the Add widget CTA has been dismissed from the browse tab page.
5. If the user doesn't add the widget, nothing changes.

Scenario 2: User doesn't have a device that supports automatically adding a widget.

1. Open Menu > Settings and click on "Home Screen Widget"
2. Logcat should show a message: `V/RxBasedPixel: Pixel sent: ms_add_home_screen_widget_clicked with params: {} {}`
3. The Add widget instructions page should be shown instead.

Scenario 3: Add widget CTA should still work as expected.
1. Click "Add widget" on the Add widget CTA in the browse tab.
2. If the device supports automatic adding of the widget, the launcher should be shown.
3. If the device doesn't support the automatic adding of the widget, the Add widget instructions page should be shown instead.

### UI changes
| Before  | After |
| ------ | ----- |
|before|after|
|Before: Pixel no toast|After: PIxel with toast|
|device-2022-01-21-123952|device-2022-01-21-110325|

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/widget/AddWidgetLauncher.kt
+++ b/app/src/main/java/com/duckduckgo/app/widget/AddWidgetLauncher.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2022 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.widget
+
+import android.annotation.SuppressLint
+import android.app.Activity
+import android.app.ActivityOptions
+import android.appwidget.AppWidgetManager
+import android.content.ComponentName
+import com.duckduckgo.app.widget.ui.AddWidgetInstructionsActivity
+import com.duckduckgo.app.widget.ui.WidgetCapabilities
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.widget.SearchAndFavoritesWidget
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import javax.inject.Inject
+
+interface AddWidgetLauncher {
+    fun launchAddWidget(activity: Activity?)
+}
+
+@ContributesBinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
+class AddWidgetCompatLauncher @Inject constructor(
+    private val defaultAddWidgetLauncher: AppWidgetManagerAddWidgetLauncher,
+    private val legacyAddWidgetLauncher: LegacyAddWidgetLauncher,
+    private val widgetCapabilities: WidgetCapabilities
+) : AddWidgetLauncher {
+
+    override fun launchAddWidget(activity: Activity?) {
+        if (widgetCapabilities.supportsAutomaticWidgetAdd)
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/duckduckgo/app/widget/AddWidgetLauncher.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.duckduckgo.app.widget.ui.AddWidgetInstructionsActivity
+import com.duckduckgo.app.widget.ui.WidgetCapabilities
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.widget.SearchAndFavoritesWidget
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import javax.inject.Inject
+
+interface AddWidgetLauncher {
+    fun launchAddWidget(activity: Activity?)
```
cmonfortep, line 36: Q: Do we need an`Activity` here? I've been checking the code inside the different implementations and I think we only need a context. 

EDIT: if context is what you need. I would encourage you to see if it's possible to automatically inject the context in the concrete class via constructor (but I'm not sure if our setup allows that).

app/src/main/java/com/duckduckgo/app/widget/AddWidgetLauncher.kt
```diff
@@ -0,0 +1,85 @@
...
+}
+
+@ContributesBinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
+class AddWidgetCompatLauncher @Inject constructor(
+    private val defaultAddWidgetLauncher: AppWidgetManagerAddWidgetLauncher,
```
cmonfortep, line 42: If both,`AppWidgetManagerAddWidgetLauncher` and `LegacyAddWidgetLauncher` are concrete classes of `AddWidgetLauncher`, we should inject them as `AddWidgetLauncher`. Otherwise, there's no benefit of them implementing the same interface.

app/src/main/java/com/duckduckgo/app/widget/AddWidgetLauncher.kt
karlenDimla: Good question. The `AppWidgetManager` doesn't depend at all on Activity object. However, the legacy implementation where `ActivityOptions.makeSceneTransitionAnimation(activity)` specifically requires an Activity. I actually would have wanted to inject the Activity instance to the LegacyAddWidgetLauncher class only, but that would mean that for the BrowserActivity and SettingsActivity component we would need to provide a binding for the Activity like `bindActivity(impl: BrowserActivity): Activity.` which I don't think we currently have.

cmonfortep: oh, I didn't spot that usage inside `LegacyAddWidgetLauncher`. Sounds reasonable then to keep it like this 👍

aitorvs (approved):
Awesome job @karlenDimla !!

* did go with Karl over the changes from product review, all good
* did go with Karl over the comments from Cristian, all lgtm

cc @cmonfortep

PR #1702, opened 2022-01-24
Fix TracerPacket memory leak

aitorvs:
Task/Issue URL: https://app.asana.com/0/0/1201701653646856/f

### Description
This PR fixes the memory leak caused by tracer buffer.
Tracer packets have an associated `ByteBuffer` needed to inser the packet in the network-to-device queue. One we take the packet out and process it, we were not releasing it. Causing a memory leak.

This PR:
* releases the bytebuffer associated to the tracer packet
* Incidentally
  * adds some log statements for RUN write errors
  * double ensures that `acquire` buffers are cleared so that they're ready to be used

### Steps to test this PR
Smoke tests for AppTP

PR #1703, opened 2022-01-24
Bye Bye tracer paket

aitorvs:
Task/Issue URL: https://app.asana.com/0/0/1201710091654544/f

### Description
We fixed the memory leak with the tracer packet in https://github.com/duckduckgo/Android/pull/1702. 
But because the tracer packet is highly invasive and we've seen how can affect the health of the VPN and because we haven't seen it results in true positives while testing internally, we're removing it for now not to spend more time debugging it and focus on the other metrics that relate to real system errors.

See [here](https://app.asana.com/0/0/1201710091654544/f) for more context.

**Why fixing and then removing?**
* In case we want to come back to it later, it is simpler to just revert this PR and be sure the memory leak is not there anymore

### Steps to test this PR

_Verify AppTP is gone_
- [x] enable AppTP
- [x] verify that the diagnostrics screen doesn't have any tracer packet metrics reported
- [x] verify in the logcat that nothing comes back when filtering by `tracer`
- [x] smoke tests on AppTP, open apps, see they load, verify AppTP blocks trackers


aitorvs:
* develop
  * **PR #1703**  (this PR)

malmstein (approved):
🔥 the packet

PR #1711, opened 2022-01-26
Return URL when we cannot parse it rather than throw an exception

marcosholgado:
Task/Issue URL:  https://app.asana.com/0/1125189844152671/1201646552983568

### Description
Rather than throwing an exception when we cannot parse a URL, return the URL.

### Steps to test this PR
Test pass


joshliebe (approved):
🚢 it!

PR #1720, opened 2022-02-01
Faster VPN bring-up

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1201711685020940/f

### Description
Currently it is the HeartBeat that detects when the VPN has been killed for reasons other than user intent or revokes. In such cases the HB will bring the VPN back up.
The HB is a worker, and so it is the scheduled periodically every 15min (min period supported for work manager).
The main issu with that approach is that many users may noticed the VPN switched off (as a result of VPN process killed) way before the HB had the time to bring it back up.

This PR uses a complementary new way to bring-up the VPN upon sudden kill.
* Added a new service `VpnStateMonitorService`
* The `TrackerBlockingVpnService` will bind to the new `VpnStateMonitorService` when started and stopped (or killed)
* Every time the binding is broken, the `VpnStateMonitorService` will get notified
* The `VpnStateMonitorService` will then use the same logic used in the HB to decide whether it needs to bring te VPN back up

To make sure it all works as expected, the `VpnStateMonitorService` runs in the browser process, similar to what happens with the HB.

### Steps to test this PR

_Test sudden VPN process kill_
- [x] enable AppTP
- [x] Verify `D/VpnStateMonitorService: Bound to VPN` message appears in logcat
- [x] use Android Studio or Intellij to kill just the `vpn` process. For that, open logcat, and select the `:vpn` process, then click on the 🟥 in the logcat window
- [x] verify the `D/VpnStateMonitorService: Sudden stop, restarting VPN` appears in logcat
- [x] verify AppTP is re-enabled

_Test user-initiated VPN switch-off_
- [x] enable AppTP
- [x] manually disable AppTP
- [x] verify the `D/VpnStateMonitorService: Nah, all good` log appears in logcat
- [x] verify AppTP stays disabled

_Test VPN revoke switch-off_
- [x] enable AppTP
- [x] enable another VPN to force a revoke
- [x] verify the `D/VpnStateMonitorService: Nah, all good` log appears in logcat
- [x] verify AppTP stays disabled

### DIFF
--- a/vpn/src/main/AndroidManifest.xml
+++ b/vpn/src/main/AndroidManifest.xml
@@ -100,6 +100,10 @@
             android:screenOrientation="portrait"/>
 
         <service
+            android:name=".service.state.VpnStateMonitorService" >
+        </service>
+
+        <service
             android:name=".service.TrackerBlockingVpnService"
             android:permission="android.permission.BIND_VPN_SERVICE"
             android:process=":vpn">
--- a/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/state/VpnStateMonitorService.kt
+++ b/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/state/VpnStateMonitorService.kt
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2022 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.mobile.android.vpn.service.state
+
+import android.app.Service
+import android.content.Intent
+import android.os.Binder
+import android.os.IBinder
+import androidx.annotation.WorkerThread
+import com.duckduckgo.app.di.AppCoroutineScope
+import com.duckduckgo.app.global.DispatcherProvider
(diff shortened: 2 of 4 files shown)


aitorvs:
* develop
  * **PR #1720** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/1720" target="_blank">Graphite</a>  (this PR)

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/state/VpnStateMonitorService.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @WorkerThread
+    private fun vpnBringUpIfSuddenKill() {
+        val lastHearBeat = vpnDatabase.vpnHeartBeatDao().hearBeats().maxByOrNull { it.timestamp }
+        if (true == lastHearBeat?.isAlive()) {
+            Timber.d("Sudden stop, restarting VPN")
+            TrackerBlockingVpnService.startService(applicationContext)
+        } else {
+            Timber.d("Nah, all good")
```
malmstein, line 77: How about a better message here? Something like, "We don't need to restart the VPN"?

malmstein (approved):
Couple of comments, but this looks really good. Big improvement

vpn/src/main/AndroidManifest.xml
```diff
@@ -99,6 +99,11 @@
             android:parentActivityName="com.duckduckgo.mobile.android.vpn.ui.tracker_activity.DeviceShieldTrackerActivity"
             android:screenOrientation="portrait"/>
 
+        <service
+            android:name=".service.state.VpnStateMonitorService"
+            android:permission="android.permission.BIND_VPN_SERVICE" >
```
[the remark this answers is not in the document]
aitorvs: Yup, good catch. We prob don't need it.

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/state/VpnStateMonitorService.kt
aitorvs: 👍

PR #1734, opened 2022-02-09
Fix native memory leaks

aitorvs:
Task/Issue URL: https://app.asana.com/0/0/1201748194657736/f

### Description
After auditing the AppTP runtime, [these](https://app.asana.com/0/0/1201778487362550/f) are the memory leaks we've found. Basically places where the buffers need to be released because they're no longer in used, but they're not.

### Steps to test this PR

_Memory leak test_
- [x] install from develop
- [x] enable AppTP and go to diagnostics screen
- [x] Open apps that send trackers, even run speed tests
- [x] verify (diagnostics screen) how the `Buffer allocations` never go down to `0`
- [x] repeat steps above
- [x] verify how the `Buffer allocations` go eventually to `0`

_AppTP smoke tests_
- [x] enable AppTP and open some of you regular apps to see all works as expected, content loads, apps continue working etc.


malmstein (approved):
Nice work @aitorvs, release them all!

PR #1774, opened 2022-02-27
Get data from device for email settings page

alistairjcbrown:
Task/Issue URL: https://app.asana.com/0/72649045549333/1201900580321807/f

### Description
Update to allow requesting user data from Android devices

### Steps to test this PR
1. Update `common/src/main/java/com/duckduckgo/app/global/AppUrl.kt` to have host value align with dev instance
2. Update autofill to https://github.com/duckduckgo/duckduckgo-autofill/pull/52
3. Build browser
4. Go to my dev instance
5. If needed, log in with production duck email username
6. See "Autofill" tab with duck email and private address generator

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
+++ b/app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
@@ -47,6 +47,15 @@ class EmailJavascriptInterface(
     }
 
     @JavascriptInterface
+    fun getUserData(): String? {
+        return if (isUrlFromDuckDuckGoEmail()) {
+            emailManager.getUserData()
+        } else {
+            ""
+        }
+    }
+
+    @JavascriptInterface
     fun storeCredentials(
         token: String,
         username: String,
--- a/app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
+++ b/app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
@@ -47,7 +47,7 @@ class EmailJavascriptInterface(
     }
 
     @JavascriptInterface
-    fun getUserData(): String? {
+    fun getUserData(): String {
         return if (isUrlFromDuckDuckGoEmail()) {
             emailManager.getUserData()
         } else {
--- a/app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
@@ -19,14 +19,13 @@ package com.duckduckgo.app.email
 import com.duckduckgo.app.email.api.EmailService
 import com.duckduckgo.app.email.db.EmailDataStore
 import com.duckduckgo.app.global.DispatcherProvider
-import com.squareup.moshi.JsonAdapter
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
```diff
@@ -100,6 +103,19 @@ class AppEmailManager(
...
+
+    override fun getUserData(): String? {
+        var userData = UserData(
+            emailDataStore.emailToken,
+            emailDataStore.emailUsername,
+            emailDataStore.nextAlias?.replace(DUCK_EMAIL_DOMAIN, ""),
+        )
+        return jsonAdapter.toJson(userData)
+    }
+
```
aitorvs, line 118: There is not need to create private mutable properties here that also don't go through our dependency injection.
This model is rather simple so I'd rather use `JSONObject`
Also the signature doesn't need to be nullable.

```suggestion
    override fun getUserData(): String {
        return JSONObject().apply { 
            put("token", emailDataStore.emailToken)
            put("userName", emailDataStore.emailUsername)
            put("nextAlias", emailDataStore.nextAlias?.replace(DUCK_EMAIL_DOMAIN, ""))
        }.toString()
    }

```

app/src/main/java/com/duckduckgo/app/email/EmailManager.kt
```diff
@@ -41,6 +43,7 @@ interface EmailManager {
 
     fun signOut()
     fun getEmailAddress(): String?
+    fun getUserData(): String?
```
aitorvs, line 46: nit: Signature doesn't need to be nullable
```suggestion
    fun getUserData(): String
```

app/src/main/java/com/duckduckgo/app/email/EmailJavascriptInterface.kt
```diff
@@ -46,6 +46,15 @@ class EmailJavascriptInterface(
         }
     }
 
+    @JavascriptInterface
+    fun getUserData(): String? {
```
aitorvs, line 50: Signature doesn't need to be nullable
```suggestion
    fun getUserData(): String {
```

aitorvs (changes requested):
I left some comments. Most important is about creating unnecessary public properties without using dependency injection and unnecessary data models like `UserData`

Also, can we remove the last commit that modifies the version of autofill? Bumping autofill should be in its own PR.

alistairjcbrown:
Thanks so much for reviewing @aitorvs 🙇
I've updated for your comments, and dropped the last commit -- how does this look now?

aitorvs (commented):
This LGTM code wise. Haven't tested it end to end. @marcosholgado can you help with that?

aitorvs (approved):

marcosholgado:
Adding a do not merge label, I will add unit tests before merging.

malmstein:
https://github.com/duckduckgo/Android/actions/runs/1909666883 has been running for 3 hours

marcosholgado (approved):
Already approved but LGTM. I added unit tests for the new code and a minor change to add the json constants to the companion object. Thanks @alistairjcbrown!

PR #1775, opened 2022-02-28
Make WebViewDatabase dependency lazy

joshliebe:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201472407198060/f

### Description
This makes the `WebViewDatabase` dependency lazy which should address some related ANRs and crashes.

### Steps to test this PR

- [ ] Verify that the app opens without error.


app/src/androidTest/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStoreTest.kt
```diff
@@ -41,13 +43,19 @@ class WebViewHttpAuthStoreTest {
     val coroutineRule = CoroutineTestRule()
 
     private val context = InstrumentationRegistry.getInstrumentation().targetContext
+    private val webViewDatabaseProvider: Provider<WebViewDatabase> = mock()
```
aitorvs, line 46: nit: you don't really need a mock here. `Provider` it's a SAM interface so you could just do a lambda.

app/src/main/java/com/duckduckgo/app/browser/httpauth/WebViewHttpAuthStore.kt
```diff
@@ -75,7 +76,7 @@ interface WebViewHttpAuthStore {
 )
 @SingleInstanceIn(AppScope::class)
 class RealWebViewHttpAuthStore @Inject constructor(
-    private val webViewDatabase: WebViewDatabase,
+    private val webViewDatabase: Provider<WebViewDatabase>,
```
aitorvs, line 79: If we want to ensure that everybody gets a provider of the database, what do you think about modifying the signature of the dagger provider method from `fun provideWebviewDatabase(context: Context): WebViewDatabase` to `fun provideWebviewDatabase(context: Context): Provider<WebViewDatabase>` ??

This way nobody can request for the concrete `WebViewDatabase` type.

app/src/androidTest/java/com/duckduckgo/app/fire/AuthDatabaseLocatorTest.kt
```diff
@@ -16,17 +16,31 @@
...
+@ExperimentalCoroutinesApi
 class AuthDatabaseLocatorTest {
 
+    private val context: Context = InstrumentationRegistry.getInstrumentation().targetContext
+
+    @Before
+    fun before() {
+        DefaultWebViewDatabaseProvider(context).get()
+
+        // Wait for the database path to be created
```
CDRussell, line 37: I don't have the full context here, but could we check for the existence of the "database path" ourselves and wait if it's not yet created (instead of the arbitrary sleep)?

app/src/androidTest/java/com/duckduckgo/app/fire/DatabaseCleanerHelperTest.kt
```diff
@@ -40,20 +41,24 @@ class DatabaseCleanerHelperTest {
...
     }
 
     @Test
     fun whenChangeJournalModeToDeleteThenReturnTrue() = runTest {
+        awaitPathCreation()
         assertTrue(testee.changeJournalModeToDelete(databaseLocator.getDatabasePath()))
     }
+
+    private suspend fun awaitPathCreation() {
+        while (databaseLocator.getDatabasePath().isEmpty()) {
```
CDRussell, line 60: Perhaps we should also guard against this looping forever?

Since we're in a coroutine already, we could use `withTimeout` to put a maximum amount of time that we're willing to wait (before an exception is thrown). This would be a nicer experience as a developer and for the CI system.

joshliebe: I've set the timeout to 5s, going to rebuild a few times to check for flakiness...

joshliebe: I've re-run this a few times without issue, so hopefully it should be okay now

app/src/androidTest/java/com/duckduckgo/app/fire/AuthDatabaseLocatorTest.kt
```diff
@@ -16,21 +16,43 @@
...
 
         // If this test fails, it means the Auth Database path has changed its location
         // If so, add a new database location to knownLocations list
         assertTrue(databasePath.isNotEmpty())
     }
+
+    private suspend fun awaitPathCreation() {
+        withTimeout(5000) {
+            while (authDatabaseLocator.getDatabasePath().isEmpty()) {
+                delay(100)
```
CDRussell, line 54: nit: this delay could probably be smaller, like 5ms or 10ms; don't see any problems with a tight retry loop given it's in a test

CDRussell (approved):

app/src/androidTest/java/com/duckduckgo/app/fire/AuthDatabaseLocatorTest.kt
joshliebe: Set it to 10 millis

joshliebe:
Thanks for the review @CDRussell!

PR #1777, opened 2022-02-28
Implementation task for Downloads UI [moved to modules]

anikiki:
Task/Issue URL: https://app.asana.com/0/0/1201892740921878/f

### Description
Refactored downloads and moved to modules.

### Steps to test this PR

- Install from this branch
- Download files (you can use this website or a similar one -> https://file-examples.com/)
- Check the below:
        - when download starts you see a snackbar mentioning this
        - when download finishes you see a snackbar mentioning this and an action to open the file
        - you can open the file or see an error if there is not app that can handle the specific file
        - if you download the same file multiple times the file name contains a "-x" before the extension, where x is an incremental number
- open the Downloads view from the menu
- Check the below:
        - you see the list of the downloaded files
        - for each file you see the name and the size. If the file is still downloading you see "In Progress" instead of the size
        - the list contains headers like "TODAY", "YESTERDAY", "PAST WEEK" etc. You can alter the download date (createdAt) in the database (app.db -> downloads) to simulate various scenarios.
        - tap on an item in the list and you should be able to open the file or see an error if there is not app that can handle the specific file
        - tap on the overflow menu for an item and share it
        - tap on the overflow menu for an item and delete it. You should see a snackbar with an action to UNDO. Both delete and undo should work properly.
        - notice there is a search in the toolbar. Try to search for an existing file / a file that doesn't exist and see how it works.
        - tap on the overflow menu to the whole screen and delete all files. You should see a snackbar with an action to UNDO. Both delete and undo should work properly.
        - when no files are downloaded (or they were deleted) you should see an empty view
        - try to download a large file (you can use this website or a similar one -> https://www.thinkbroadband.com/download).
        - while the file is downloading try to tap on its item and notice is not clickable.

### UI changes

Screenshot_20220405_180737


app/schemas/com.duckduckgo.app.global.db.AppDatabase/43.json
```diff
@@ -968,68 +968,12 @@
         },
         "indices": [],
         "foreignKeys": []
-      },
```
anikiki, line 971: The "downloads" table in "app.db" is now removed as it's moved to its own database in the `downloads-store` module.

app/src/main/java/com/duckduckgo/app/downloads/DownloadsActivity.kt
```diff
@@ -158,18 +161,30 @@ class DownloadsActivity : DuckDuckGoActivity() {
...
                     BaseTransientBottomBar.BaseCallback.DISMISS_EVENT_CONSECUTIVE -> { /* noop */ }
                 }
             }
         }).show()
     }
 
+    private fun handleDeleteAll(items: List<DownloadItem>) {
+        viewModel.deleteFilesFromDisk(items)
+        // Remove all unfinished downloads from DownloadManager.
+        (getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager?)?.let { manager ->
```
malmstein, line 174: Should this logic be in the `Activity`? Maybe the `ViewModel` can have `DownloadManager` as a collaborator?

app/src/main/java/com/duckduckgo/app/downloads/DownloadsActivity.kt
```diff
@@ -158,18 +161,30 @@ class DownloadsActivity : DuckDuckGoActivity() {
...
+            items.filter { it.downloadStatus != FINISHED }.forEach { manager.remove(it.downloadId) }
+        }
+    }
+
+    private fun cancelDownload(command: CancelDownload) {
+        (getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager?)?.remove(command.item.downloadId)
```
malmstein, line 180: ditto

app/src/main/res/drawable/ic_more_vertical.xml
```diff
@@ -0,0 +1,11 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:pathData="M10,18C10,16.9 10.9,16 12,16C13.1,16 14,16.9 14,18C14,19.1 13.1,20 12,20C10.9,20 10,19.1 10,18ZM12,10C10.9,10 10,10.9 10,12C10,13.1 10.9,14 12,14C13.1,14 14,13.1 14,12C14,10.9 13.1,10 12,10ZM12,8C13.1,8 14,7.1 14,6C14,4.9 13.1,4 12,4C10.9,4 10,4.9 10,6C10,7.1 10.9,8 12,8Z"
+      android:fillColor="#222222"
```
malmstein, line 8: ?attr/primaryIconColor

downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/FileDownloadNotificationChannelType.kt
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.downloads.impl
+
+import androidx.core.app.NotificationManagerCompat
+import com.duckduckgo.app.notification.model.Channel
+
+object FileDownloadNotificationChannelType {
+    val FILE_DOWNLOADING = Channel(
```
malmstein, line 23: why do we need two different channels for downloads instead of a single one?

malmstein (approved):
This looks great, nice stuff @anikiki ! I left a couple comments / suggestions but nothing blocking. If you can change the icon color that would be great.

downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/FileDownloadNotificationChannelType.kt
anikiki: It was like this and didn't want to change it as I didn't have the full context.

app/src/main/java/com/duckduckgo/app/downloads/DownloadsActivity.kt
anikiki: Updated, thanks!

app/src/main/java/com/duckduckgo/app/downloads/DownloadsActivity.kt
anikiki: Updated, thanks!

PR #1778, opened 2022-02-28
Send OS version with Downloads / Crash pixels

malmstein:
Task/Issue URL: https://app.asana.com/0/1174433894299346/1201879515350133

### Description
When triaging error reports in the Android app it would be helpful to be able to filter by OS version. This PR adds the OS pixel to the agreed pixels

### Steps to test this PR

_Download Pixels_
- [ ] Download a file using the Browser (thinkbroadband.com/download)
- [ ] Filter by Pixel Sent
- [ ] `m_download_request_started `with params: {os_version=30} {}
- [ ] `m_download_request_succeeded `with params: {os_version=30} {}

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -2552,10 +2552,7 @@ class BrowserTabViewModel(
         command.postValue(LoadExtractedUrl(extractedUrl = initialUrl))
     }
 
-    override fun onUrlExtracted(
-        initialUrl: String,
-        extractedUrl: String?
-    ) {
+    override fun onUrlExtracted(initialUrl: String, extractedUrl: String?) {
         val destinationUrl: String = if (extractedUrl != null) {
             trackingLinkDetector.lastTrackingLinkInfo = TrackingLinkInfo(trackingLink = initialUrl)
             Timber.d("Tracking link detection: Success! Loading extracted URL: $extractedUrl")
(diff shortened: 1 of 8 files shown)


anikiki (approved):
Looks good and works as expected! :tada: There are some checks failing, didn't look at them.

malmstein:
Thanks @anikiki !

aitorvs (changes requested):
I think it'd be better if we do it through an interceptor. No class need to be modified and it would be easier to know what pixels contain the OS version

app/src/main/java/com/duckduckgo/app/di/StatisticsModule.kt
```diff
@@ -97,7 +97,7 @@ class StatisticsModule {
         uncaughtExceptionRepository: UncaughtExceptionRepository,
         pixelSender: PixelSender,
         offlinePixels: DaggerSet<OfflinePixel>
-    ): OfflinePixelSender = OfflinePixelSender(offlinePixelCountDataStore, uncaughtExceptionRepository, pixelSender, offlinePixels)
+    ): OfflinePixelSender = OfflinePixelSender(offlinePixelCountDataStore, uncaughtExceptionRepository, pixelSender, appBuildConfig, offlinePixels)
```
anikiki, line 100: `appBuildConfig` is not needed anymore.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2552,7 +2552,10 @@ class BrowserTabViewModel(
         command.postValue(LoadExtractedUrl(extractedUrl = initialUrl))
     }
 
-    override fun onUrlExtracted(initialUrl: String, extractedUrl: String?) {
+    override fun onUrlExtracted(
+        initialUrl: String,
+        extractedUrl: String?
+    ) {
```
aitorvs, line 2558: nit: revert this change, this comfortably fits in the 150 char limit we have and there's no need to modify this file

app/src/main/java/com/duckduckgo/app/di/StatisticsModule.kt
aitorvs: +1

aitorvs (approved):
lgtm. Left one nit about unnecessary reformat of a file we don't need to touch.

Also a +1 to a comment from @anikiki

PR #1800, opened 2022-03-10
Load file scheme searches

nalcalag:
Task/Issue URL: [https://app.asana.com/0/1201807753394693/1201781057786340](https://app.asana.com/0/1201807753394693/1201781057786340)

### Description

Fix for a crash if user tries to load `file` scheme. For consistency `about` scheme is returning now SearchQuery UrlType instead Unknown.

### Steps to test this PR

_Feature 1_
- [ ] Make a search with a file scheme (e.g. 'file:' or 'file:///sdcard')
- [ ] Verify browser load the search instead showing "Open in another app" AlertDialog


CDRussell (approved):
Changes themselves look good! Tested `about://` and `file://` urls

We should add tests for these if we can. Probably in `SpecialUrlDetectorImplTest`. Approving PR on the basis some testing is included.

PR #1827, opened 2022-03-24
Improve the network connection monitor reporting

aitorvs:
Task/Issue URL: https://app.asana.com/0/72649045549333/1202005414608865/f

### Description
We expected Android to call the `onAvailable` and `onLost` callbacks in order. Meaning when there is a network switch, eg. from WIFI to CELL, Android would call `onLost(WIFI)` and then `onAvailable(CELL)`.

However that's not happening. Android can (in the example above) call `onAvailable(CELL)` when the WIFI signal is getting lost, and then `onLost(WIFI)` when it's definitely lost.

This was messing with our implementation because we never bother tracking what network was available or lost, we just assumed they'd be called in order.

The consequence of that was that we would report `currentNetwork` as `NO_NETWORK` when that was not the case.

### Steps to test this PR

_Test network connectivity_
- [x] install form this branch
- [x] enable AppTP
- [x] ensure you're on WIFI
- [x] filter logcat by `NetworkTypeCollector`
- [x] change to CELLULAR network (no need turn off WIFI)
- [x] verify `currentNetwork` changes to CELLULAR and it is `AVAILABLE`
- [x] verify `previousNetwork` changes to WIFI and it is `LOST`
- [x] change back to WIFI
- [x] verify `currentNetwork` changes to WIFI AVAILABLE and `previousNetwork` to CELLULAR AVAILABLE
- [x] verify `Lost of previously switched connection` appears after that

### DIFF
--- a/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/bugreport/NetworkTypeCollector.kt
+++ b/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/bugreport/NetworkTypeCollector.kt
@@ -82,27 +82,27 @@ class NetworkTypeCollector @Inject constructor(
 
     private val privateDnsRequest = NetworkRequest.Builder().build()
 
-    private val wifiNetworkCallback = object : ConnectivityManager.NetworkCallback() {
+    private val wifiNetworkCallback = object : NetworkCallback() {
         override fun onAvailable(network: Network) {
-            updateNetworkInfo(NetworkType.WIFI)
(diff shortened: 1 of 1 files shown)


aitorvs:
* develop
  * **PR #1827** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/1827" target="_blank">Graphite</a>  (this PR)
    * **PR #1828** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/1828" target="_blank">Graphite</a> 

malmstein (approved):
Works as expected, nice @aitorvs

PR #1839, opened 2022-04-04
Wrap shortcut creation in a runCatching

CDRussell:
Task/Issue URL: https://app.asana.com/0/276630244458377/1202055094426961/f

### Description
Wraps the call interacting with `ShortcutManager` in a `runCatching` block. There are times when this call will fail with an `IllegalStateException`. When this happens, we don't want a crash but can silently swallow the exception given the low-importance of app shortcuts being registered.

### Steps to test this PR
1. Test the app starts as normal
2. Tests the app shortcuts are created as normal (ie, that you can long press on the app icon and use the shortcuts)


joshliebe (approved):
LGTM! ✅

PR #1877, opened 2022-04-21
Crash in NetworkFileDownloader

anikiki:
Task/Issue URL: https://app.asana.com/0/414730916066338/1201550925053703/f

### Description
Wrapped code in try/catch for Android 10 specific crash in NetworkFileDownloader.

### Steps to test this PR
Can't be reproduced. It happens only on Android 10 and it's very infrequent on Pixel devices -> only 0.23% in the last 30 days.
Instead of crashing, now we display a snackbar and a notification to inform the user about the failed download. See the 2 attached screenshots.

### UI changes
| Show snackbar  | Show notification |
| ------ | ----- |
failed_to_download_snackbar|failed_to_download_notification|

### DIFF
--- a/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/NetworkFileDownloader.kt
+++ b/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/NetworkFileDownloader.kt
@@ -39,8 +39,7 @@ import javax.inject.Inject
 class NetworkFileDownloader @Inject constructor(
     private val context: Context,
     private val filenameExtractor: FilenameExtractor,
-    private val fileService: DownloadFileService,
-    private val appBuildConfig: AppBuildConfig
+    private val fileService: DownloadFileService
 ) {
 
     fun download(pendingDownload: PendingFileDownload, callback: DownloadCallback) {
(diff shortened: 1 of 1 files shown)


cmonfortep (approved):
LGTM 🎢

app/src/main/java/com/duckduckgo/app/browser/downloader/NetworkFileDownloader.kt
```diff
@@ -129,6 +132,26 @@ class NetworkFileDownloader @Inject constructor(
...
+        } else {
+            startDownload(manager, request, callback)
+        }
+    }
+
+    private fun startDownloadOnAndroidQ(manager: DownloadManager?, request: DownloadManager.Request, callback: FileDownloader.FileDownloadListener) {
+        try {
+            startDownload(manager, request, callback)
+        } catch (e: IllegalArgumentException) {
+            callback.downloadFailed("Failed when trying to enqueue the download on Android Q.", DownloadFailReason.Other)
```
CDRussell, line 150: Probably worth throwing in a ` Timber.e` here with the full stack trace?

CDRussell (approved):

app/src/main/java/com/duckduckgo/app/browser/downloader/NetworkFileDownloader.kt
```diff
@@ -129,6 +132,26 @@ class NetworkFileDownloader @Inject constructor(
             setNotificationVisibility(DownloadManager.Request.VISIBILITY_VISIBLE_NOTIFY_COMPLETED)
         }
         val manager = context.getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager?
+
+        // In some unknown cases enqueuing a download on Android Q throws an IllegalArgumentException. We want the user to still
+        // be able to use the app and inform them the download has failed. If this will happen in the future on other versions we want
+        // to see the crashes.
+        if (appBuildConfig.sdkInt == Build.VERSION_CODES.Q) {
+            startDownloadOnAndroidQ(manager, request, callback)
```
aitorvs, line 140: I think we should probably not make a distinction between `Q` and any other versions.
I have taken a look at AOSP and this can potentially still happen in all versions. The `ContentResolver.insert` method throws `IllegalArgumentException` if "the provider doesn't support any of the bundled arguments" which come from the `request`.
The exception is never `try-catch`ed inside AOSP but also the signature is never carried on.

anikiki: Thanks, updated to do this for all Android versions. I re-checked and we only got this crash on Android 10 and there are some people mentioning this being noticed on Android 10 only, however nobody managed to understand how / why is happening. 
There are a few different download-related crashes that indeed happen on all OS versions and they'll be investigated separately.

aitorvs (commented):
lgtm

PR #1882, opened 2022-04-22
Android: Crash missing migration from Downloads

anikiki:
Task/Issue URL: 

### Description
Changed database version to 42 as there is no reason to be 43. Version 43 was used in an intermediary PR and not totally removed.

### Steps to test this PR

**To reproduce:**
- install and launch the app from a commit prior to R #1776. For example take any of the latest tags, if easier ->  5.121.0.4-nightly or 5.121.0
- then **update app from develop**
- see crash because we're missing AppDatabase db migration from 42 → 43

**To test the fix:**
- install and launch the app from a commit prior to R #1776. For example take any of the latest tags, if easier ->  5.121.0.4-nightly or 5.121.0
- then **update app from this branch**
- the app is not crashing


aitorvs (approved):
Lgtm

PR #1898, opened 2022-04-29
Add minSupportedVersion to privacy features

joshliebe:
Task/Issue URL: https://app.asana.com/0/0/1202202373459017/f

### Description
This adds support for a min supported version in each privacy feature.

### Steps to test this PR
Comment out line: `privacyConfigPersister.persistPrivacyConfig(it)` (Line 42) in `PrivacyConfigDownloader`

- [x] Go to https://privacy-test-pages.glitch.me/privacy-protections/query-parameters
- [x] Verify that the test cases pass
- [x] Uninstall the app
- [x] Change the `minSupportedVersion` in `privacy_config.json` to 61230000
- [x] Install the app and go to https://privacy-test-pages.glitch.me/privacy-protections/query-parameters
- [x] Verify that the test cases no longer pass
- [x] Uninstall the app
- [x] Delete the `minSupportedVersion` in `privacy_config.json`
- [x] Install the app and go to https://privacy-test-pages.glitch.me/privacy-protections/query-parameters
- [x] Verify that the test cases pass


privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/version/VersionHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+import java.lang.NumberFormatException
+import javax.inject.Inject
+
+interface VersionHandler {
+    fun isSupportedVersion(minSupportedVersion: String?): Boolean
+}
+
+@ContributesBinding(AppScope::class)
+class RealVersionHandler @Inject constructor(val context: Context) : VersionHandler {
+    private fun getAppVersion(): String? {
```
marcosholgado, line 32: You can get this from AppBuildConfig

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/version/VersionHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+        } catch (e: NameNotFoundException) {
+            null
+        }
+    }
+
+    override fun isSupportedVersion(minSupportedVersion: String?): Boolean {
+
+        if (minSupportedVersion == null) return true
+
+        getAppVersion()?.let { appVersion ->
```
marcosholgado, line 45: Gut feeling says to use the code instead of the version, easier to use and compare.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/trackingparameters/TrackingParametersPlugin.kt
```diff
@@ -56,7 +58,8 @@ class TrackingParametersPlugin @Inject constructor(
             }
 
             trackingParametersRepository.updateAll(exceptions, parameters)
-            val isEnabled = trackingParametersFeature?.state == "enabled"
+            val isEnabled = trackingParametersFeature?.state == "enabled" &&
+                versionHandler.isSupportedVersion(trackingParametersFeature.minSupportedVersion)
```
marcosholgado, line 62: We should also add this to all the other plugins. It probably doesn't matter right now but the vpn module also contains its own remote config plugins so we probably need to find a way to be able to share it there as well... I can think of a different options but maybe moving to the public API of remote config is a nice way to do it? cc/ @aitorvs

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/trackingparameters/TrackingParametersPlugin.kt
```diff
@@ -19,19 +19,21 @@ package com.duckduckgo.privacy.config.impl.features.trackingparameters
...
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
+import javax.inject.Inject
 
 @ContributesMultibinding(AppScope::class)
 class TrackingParametersPlugin @Inject constructor(
     private val trackingParametersRepository: TrackingParametersRepository,
-    private val privacyFeatureTogglesRepository: PrivacyFeatureTogglesRepository
+    private val privacyFeatureTogglesRepository: PrivacyFeatureTogglesRepository,
+    private val versionHandler: VersionHandler
```
aitorvs, line 36: We don't really need this new `VersionHandler` class. We can just use `AppBuildConfig`

joshliebe: I've kept the `VersionHandler` so I don't have to repeat this everywhere:
```
if (minSupportedVersion == null) return true
return appBuildConfig.versionCode >= minSupportedVersion
```
I thought about adding that logic to `AppBuildConfig` but it doesn't feel like it makes sense to put it there.

joshliebe: `VersionHandler` has been removed.

marcosholgado (approved):
Nice job @joshliebe! Left a few nits.

@aitorvs I have already reviewed the VPN part and it made sense to me. There is now logic for the main toggle to take the min version into account although we never call `insert` but in my head it makes sense to have that there just in case we eventually want to be able to enable/disable the main vpn toggle.

aitorvs (commented):
LGTM!

I think we can revert the changes for all appTP classes. AppTP root feature `appTrackerProtection` ignores the `"state": "enabled` json property because AppTP is not a feature it can be enable/disabled as a whole.

The minSupported version in ApptP makes sense probably for its `settings`, which are AppTP but I think that can be done in a different task.

aitorvs (approved):

joshliebe:
Thanks for the reviews @aitorvs and @marcosholgado! I've updated the config and addressed all the comments ✅

PR #1905, opened 2022-05-03
Improve connectivity change reconfigurations

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202217410650615/f

### Description
This PR:
* Ensures that the `ConnectivityChangeReceiver` only reconfigures the VPN when the network type changes, not in every single trigger

Incidentally, it also
* Removes the `PrivateDnsSettingMonitor` class and performs that work directly in the `NetworkTypeCollector` as it is more readable
* Sets the `TUN_READ_IO_SAMPLES` trigger to 1min instead of 45sec, as 45sec was too low
* Makes the `VpnPreferences` an interface so that it can be faked. This requires some minor changes in `AndroidVpnReminderReceiverManager`
* Add to the `DeviceShieldPixels` the ability to enqueue pixels rather that directly send them. This was needed at the time in the `NetworkConnectivityHealthHandler` because we wanted to send pixels when the device/VPN doesn't have connection, but I was too lazy to do it, so I didn't add the daily pixel, which we need

### Steps to test this PR

_Private DNS_
- [x] Enable AppTP
- [x] Go to AppTP dev settings and make sure "Private DNS" toggle is ON
- [x] Go to Android settings -> Private DNS and manually configure to `one.one.one.one`, click save
- [x] Verify the VPN restarts
- [x] Verify in the diagnostics screen that `1.1.1.1` address is configured as DNS
- [x] Verify apps continue to work

_Network switching handling_
- [x] Enable ApptP
- [x]  Go to AppTP dev settings and make sure "Network Switch Handling" toggle is ON
- [x] ensure you're on WIFI network
- [x] Scroll down from top of the screen
- [x] Select mobile data
- [x] verify the VPN restarts only once, logcat `Connectivity change to...` prints once
- [x] Scroll down from top of the screen
- [x] Select WIFI
- [x] verify the VPN restarts only once, logcat `Connectivity change to...` prints once
- [x] Verify apps continue to work


aitorvs:
* develop
  * **PR #1905** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/1905" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/1905?utm_source=stack-comment).

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/bugreport/NetworkTypeCollector.kt
```diff
@@ -119,9 +121,20 @@ class NetworkTypeCollector @Inject constructor(
                 )
                 true
             } else {
-                Timber.v("Private DNS support is disabled or not set")
+                Timber.v("Private DNS disabled")
                 false
             }
+
+            // Check if VPN reconfiguration is needed
+            if (appTpFeatureConfig.isEnabled(AppTpSetting.PrivateDnsSupport) && vpnPreferences.isPrivateDnsEnabled != privateDns) {
```
malmstein, line 129: does this logic belong to the collector? it's not like the other network callbacks that actually change values used in collector.

malmstein: I just saw that [PrivateDnsSettingMonitor.kt](https://github.com/duckduckgo/Android/pull/1905/files#diff-0edd70d3143c981c20c1cc0af74c05f06dffd81c681502bbb7ab1f35023ccf34) is deleted. Why not leave it as it was with the added logic?

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/network/connection/ConnectivityChangeReceiver.kt
```diff
@@ -47,6 +49,16 @@ class ConnectivityChangeReceiver @Inject constructor(
             val networkType = intent.getIntExtra(ConnectivityManager.EXTRA_NETWORK_TYPE, ConnectivityManager.TYPE_DUMMY)
             if (networkType == ConnectivityManager.TYPE_VPN) return
 
+            val oldNetworkType = vpnPreferences.activeNetworkType
+
+            if (oldNetworkType == networkType.toString()) {
```
malmstein, line 54: should we use `networkType.lowercase()` instead? just in case some OEMs don't respect upppercase / lowercase?

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/network/connection/ConnectivityChangeReceiver.kt
```diff
@@ -47,6 +49,16 @@ class ConnectivityChangeReceiver @Inject constructor(
...
+                Timber.d("Network type didn't change, skip")
+                return
+            }
+
+            vpnPreferences.activeNetworkType = networkType.toString()
```
malmstein, line 59: ditto

malmstein (approved):
I had to make sure that "Network Switch Handling" was enabled in AppTP Dev Settings. Other than that works as expected. Just a couple nits.

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/bugreport/NetworkTypeCollector.kt
aitorvs: Yeah, the `NetworkTypeCollector` is probably not the ideal place, is prob mainly semantics around the name of the class. But I think it was better to move it here and delete the private dns setting monitor because it also simplifies the logic around the shared prefs.

PR #1927, opened 2022-05-10
Fix App Launch Pixel

malmstein:
Task/Issue URL: https://app.asana.com/0/1157893581871903/1202061396691566

### Description
The app launch pixel is sent on app start, and not on app create. 
This PR adds a new app launch pixel, that only fires when app con create. The old pixel name (not the actual pixel value) has been renamed and kept as per discussion on the Asana task.

### Steps to test this PR

_App Launch_
- [x] Start app
- [x] Verify "m_app_launch" pixel is sent
- [x] Verify "ml" pixel is sent
- [x] Press home
- [x] Go back to the app
- [x] Verify "m_app_launch" is not sent
- [x] Verify "ml" pixel is sent

_Fire button_
- [x] Start app
- [x] Verify "m_app_launch" pixel is sent
- [x] Verify "ml" pixel is sent
- [x] Press fire button
- [x] Verify "m_app_launch" is not sent
- [x] Verify "ml" pixel is not sent


app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt
```diff
@@ -19,7 +19,8 @@ package com.duckduckgo.app.pixels
 import com.duckduckgo.app.statistics.pixels.Pixel
 
 enum class AppPixelName(override val pixelName: String) : Pixel.PixelName {
-    APP_LAUNCH("ml"),
+    APP_LAUNCH_LEGACY("ml"),
+    APP_LAUNCH("m_app_launch_c"),
```
joshliebe, line 23: Should this have a more descriptive name? Not sure what the "c" is for.

app/src/test/java/com/duckduckgo/app/pixels/EnqueuedPixelWorkerTest.kt
```diff
@@ -101,6 +107,30 @@ class EnqueuedPixelWorkerTest {
...
+    @Test
+    fun whenOnStartAndLaunchByFireActionFollowedByAppLaunchAfterGracePeriodThenSendOneAppLaunchPixel() {
+        whenever(unsentForgetAllPixelStore.pendingPixelCountClearData).thenReturn(1)
+        whenever(unsentForgetAllPixelStore.lastClearTimestamp).thenReturn(System.currentTimeMillis())
+        whenever(webViewVersionProvider.getMajorVersion()).thenReturn("91")
+
+        enqueuedPixelWorker.onStateChanged(lifecycleOwner, Lifecycle.Event.ON_CREATE)
+        enqueuedPixelWorker.onStateChanged(lifecycleOwner, Lifecycle.Event.ON_START)
+
+        Thread.sleep(10500)
```
joshliebe, line 130: 10s is a long time, could we not add the grace period as a parameter in the constructor? Something like this:

```
class EnqueuedPixelWorker @Inject constructor(
    private val workManager: WorkManager,
    private val pixel: Provider<Pixel>,
    private val unsentForgetAllPixelStore: UnsentForgetAllPixelStore,
    private val webViewVersionProvider: WebViewVersionProvider,
    private val APP_RESTART_CAUSED_BY_FIRE_GRACE_PERIOD: Long = 10_000L
)
```

Then we can pass it a much shorter value for the test.

malmstein: I like that @joshliebe, nice suggestion,

app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt
malmstein: we've been using "c" for the count pixel, but I don't think we need it here.

app/src/main/java/com/duckduckgo/app/pixels/EnqueuedPixelWorker.kt
```diff
@@ -44,7 +44,8 @@ class EnqueuedPixelWorker @Inject constructor(
     private val workManager: WorkManager,
     private val pixel: Provider<Pixel>,
     private val unsentForgetAllPixelStore: UnsentForgetAllPixelStore,
-    private val webViewVersionProvider: WebViewVersionProvider
+    private val webViewVersionProvider: WebViewVersionProvider,
+    private val APP_RESTART_CAUSED_BY_FIRE_GRACE_PERIOD: Long = 10_000L
```
aitorvs, line 48: This constant should not be in the constructor. It is not needed there. I would revert this change and leave the constant where it was before

app/src/test/java/com/duckduckgo/app/pixels/EnqueuedPixelWorkerTest.kt
aitorvs: I don't like it 🤣 
Two issues with that constant in the constructor. It's a workaround in production code to a bad test. But more importantly, it will force us to provide a `Long` value into DI as well, which also makes kitties sad.

The class already has the `UnsentForgetAllPixelStore`. That class can be easily faked so that we don't need a `sleep` in the test

joshliebe (approved):
Tested and LGTM 👍

PR #1941, opened 2022-05-16
AppTP: Disabling tracking protection issue on API 23

anikiki:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202236755602993/f

### Description
Fixed isServiceRunning on API 23.
On API 23 services are still reported as running up to 1 min after they're stopped. Changed the code to check the started property for this API version only.

### Steps to test this PR

- install from this branch on a device running Android M
- go to Settings and enable AppTP
- enable and disable AppTP and check it's indeed enabled / disabled
- check the Before / After videos attached on the Asana task.

### DIFF
--- a/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
+++ b/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
@@ -505,6 +505,9 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope(), N
 
             for (service in manager.getRunningServices(Int.MAX_VALUE)) {
                 if (TrackerBlockingVpnService::class.java.name == service.service.className) {
+                    if (Build.VERSION.SDK_INT == Build.VERSION_CODES.M) {
+                        return service.started
+                    }
                     return true
                 }
             }


vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
```diff
@@ -505,6 +505,9 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope(), N
 
             for (service in manager.getRunningServices(Int.MAX_VALUE)) {
                 if (TrackerBlockingVpnService::class.java.name == service.service.className) {
+                    if (Build.VERSION.SDK_INT == Build.VERSION_CODES.M) {
```
aitorvs, line 508: We should always use `appBuildConfig`
```suggestion
                    if (appBuildConfig.sdkInt == Build.VERSION_CODES.M) {
```

aitorvs (approved):
Left a comment to use `appBuildConfig`. Else all good. Thanks a lot for doing this

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
anikiki: Can't use the `appBuildConfig` as `isServiceRunning` is declared in the companion object. I can pass it as param, however this is used in many places.

aitorvs: > Can't use the `appBuildConfig` as `isServiceRunning` is declared in the companion object. I can pass it as param, however this is used in many places.

Nah, nvm. Didn't realise it was static call. All good

PR #1942, opened 2022-05-16
Make selector calls that add keys to the FD_SET thread-safe using lock

aitorvs:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202275295244711/f

### Description
Steps to repro:
- enable AppTP
- enable protection for Google Maps (or any other high traffic app)
- go to Google Maps and scroll
- go back to AppTP and either disable it and enable it, or disable / enable protection for one app (other than Google Maps). This should trigger a new thread to be used along with a new selector and selected-key set instances.
- go back to Google Maps and scroll. You'll get a ConcurrentModificationException (this doesn't happen all the time)

### Steps to test this PR
_Test fix_
- [ ] try to repro the issue in develop
- [ ] once repro install from this branch
- [ ] verify it doesn't repro anymore


aitorvs:
* develop
  * **PR #1942** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/1942" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/1942?utm_source=stack-comment).

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/tcp/TCBCloser.kt
```diff
@@ -42,7 +42,7 @@ class TCBCloser @Inject constructor(private val socketWriter: TcpSocketWriter) {
         isFIN: Boolean,
     ) {
         val buffer = ByteBufferPool.acquire()
-        val tcb = requireNotNull(connectionParams.tcb())
+        val tcb = connectionParams.tcb() ?: return
```
aitorvs, line 45: @anikiki I incidentally relaxed this `requireNotNull` calls to avoid unnecessary VPN restarts.
Realised of this while testing. When we turn OFF and back ON the VPB sometimes we'd get `null` TCBs because unfortunately the structure that holds them is a `static` one. Crashing is fine because we'd detect that and bring the VPN back up, but continuing is also fine and prob more desirable for now to avoid "too many" unnecessary restarts.

Added a bit of explanation to the PR description

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/tcp/TCBCloser.kt
```diff
@@ -42,7 +42,7 @@ class TCBCloser @Inject constructor(private val socketWriter: TcpSocketWriter) {
         isFIN: Boolean,
     ) {
         val buffer = ByteBufferPool.acquire()
-        val tcb = requireNotNull(connectionParams.tcb())
+        val tcb = connectionParams.tcbOrClose() ?: return
```
aitorvs, line 45: Added a bit of info in the description. TCBs can be null at times because we use an LRUCache and when we restart the VPN.
Although `requireNotNull` is OK, it'd crash the VPN processor forcing a restart. 
It's better approach to just free resources and continue to avoid unnecessary VPN process restarts

anikiki (approved):

PR #1943, opened 2022-05-16
Migrate prefs to harmony when necessary

aitorvs:
Task/Issue URL: https://app.asana.com/0/1199371158290272/1202279501986167/f

### Description
The Harmony multi process shared preferences creates the shared prefs file under a different folder and does not migrate previously existing shared preferences.

This causes a bug in DDG, to repro the bug:
* install from `5.124.0.1-nightly`
* launch app and enable AppTP
* update from develop
* launch app and go to Settings -> App Tracking Protection
* expected: AppTP screen is open
* actual: onboarding is shown again

This PR adds a migration to the `VpnSharedPreferencesProviderImpl` that is executed when shared prefs are returned using the provider

### Steps to test this PR

_Check migration_
- [ ] install form `5.124.0.1-nightly`
- [ ] launch app and enable AppTP
- [ ] update app from this branch
- [ ] launch app and go to Settings -> App tracking protection
- [ ] verify onboarding is now shown

(cohort and AppTP pixels are also migrated, but not tested as migration works the same in all instalces)

### DIFF
--- a/vpn/src/androidTest/java/com/duckduckgo/mobile/android/vpn/prefs/VpnSharedPreferencesProviderImplTest.kt
+++ b/vpn/src/androidTest/java/com/duckduckgo/mobile/android/vpn/prefs/VpnSharedPreferencesProviderImplTest.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2022 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.mobile.android.vpn.prefs
+
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
+import android.content.SharedPreferences
+import androidx.core.content.edit
+import androidx.test.platform.app.InstrumentationRegistry
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Before
+import org.junit.Test
+import java.util.UUID
+
+class VpnSharedPreferencesProviderImplTest {
+
+    private val context: Context = InstrumentationRegistry.getInstrumentation().context.applicationContext
+    private lateinit var prefs: SharedPreferences
+    private lateinit var vpnPreferencesProvider: VpnSharedPreferencesProvider
+    private lateinit var NAME: String
+
+    @Before
+    fun setup() {
+        NAME = UUID.randomUUID().toString()
+        prefs = context.getSharedPreferences(NAME, MODE_PRIVATE)
+        vpnPreferencesProvider = VpnSharedPreferencesProviderImpl(context)
+    }
+
+    @Test
+    fun whenGetMultiprocessPreferencesThenMigrateToHarmony() {
+        prefs.edit(commit = true) { putBoolean("bool", true) }
+        prefs.edit(commit = true) { putString("string", "true") }
+        prefs.edit(commit = true) { putInt("int", 1) }
+        prefs.edit(commit = true) { putFloat("float", 1f) }
+        prefs.edit(commit = true) { putLong("long", 1L) }
+
+        val harmony = vpnPreferencesProvider.getSharedPreferences(NAME, multiprocess = true, migrate = true)
+
+        assertEquals(true, harmony.getBoolean("bool", false))
+        assertEquals("true", harmony.getString("string", "false"))
+        assertEquals(1, harmony.getInt("int", 0))
+        assertEquals(1f, harmony.getFloat("float", 0f))
+        assertEquals(1L, harmony.getLong("long", 0L))
+    }
+
+    @Test
+    fun whenGetMultiprocessPreferencesAndMigrateIsFalseThenDoNotMigrateToHarmony() {
+        prefs.edit(commit = true) { putBoolean("bool", true) }
+        prefs.edit(commit = true) { putString("string", "true") }
+        prefs.edit(commit = true) { putInt("int", 1) }
+        prefs.edit(commit = true) { putFloat("float", 1f) }
+        prefs.edit(commit = true) { putLong("long", 1L) }
+
+        val harmony = vpnPreferencesProvider.getSharedPreferences(NAME, multiprocess = true)
+
+        assertNotEquals(true, harmony.getBoolean("bool", false))
+        assertNotEquals("true", harmony.getString("string", "false"))
+        assertNotEquals(1, harmony.getInt("int", 0))
+        assertNotEquals(1f, harmony.getFloat("float", 0f))
+        assertNotEquals(1L, harmony.getLong("long", 0L))
+    }
+}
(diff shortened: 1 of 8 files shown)


aitorvs:
* develop
  * **PR #1943** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/1943" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/1943?utm_source=stack-comment).

vpn/src/androidTest/java/com/duckduckgo/mobile/android/vpn/prefs/VpnSharedPreferencesProviderImplTest.kt
```diff
@@ -0,0 +1,77 @@
...
+class VpnSharedPreferencesProviderImplTest {
+
+    private val context: Context = InstrumentationRegistry.getInstrumentation().context.applicationContext
+    private lateinit var prefs: SharedPreferences
+    private lateinit var vpnPreferencesProvider: VpnSharedPreferencesProvider
+    private lateinit var NAME: String
+
+    @Before
+    fun setup() {
+        NAME = UUID.randomUUID().toString()
```
aitorvs, line 39: Easy way to start with clean state

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/prefs/VpnSharedPreferencesProvider.kt
```diff
@@ -19,24 +19,75 @@ package com.duckduckgo.mobile.android.vpn.prefs
...
 class VpnSharedPreferencesProviderImpl @Inject constructor(
     private val context: Context
 ) : VpnSharedPreferencesProvider {
-    override fun getSharedPreferences(name: String, multiprocess: Boolean): SharedPreferences {
+    override fun getSharedPreferences(name: String, multiprocess: Boolean, migrate: Boolean): SharedPreferences {
         return if (multiprocess) {
-            context.getHarmonySharedPreferences(name)
+            if (migrate) {
+                Timber.v("Migrate and return preferences to Harmony")
+                return migrateToHarmonyIfNecessary(name)
```
aitorvs, line 49: `migrateToHarmonyIfNecessary` returns already the harmony prefs to save one call

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/prefs/VpnSharedPreferencesProvider.kt
```diff
@@ -19,24 +19,75 @@ package com.duckduckgo.mobile.android.vpn.prefs
...
+                is Int -> {
+                    destination.edit { putInt(key, originalValue) }
+                }
+                is Float -> {
+                    destination.edit { putFloat(key, originalValue) }
+                }
+                is String -> {
+                    destination.edit { putString(key, originalValue) }
+                }
+                else -> Timber.w("Could not migrate $key from $name preferences")
```
aitorvs, line 85: Only `putStringSet` would not be migrated, but we don't use it in any place

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/ui/onboarding/VpnStore.kt
```diff
@@ -42,11 +43,14 @@ interface VpnStore {
 }
 
 @ContributesBinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
```
aitorvs, line 46: We don't really need this but I thought it would save some migrations. No strong opinions, I can remove it

anikiki (approved):

pablobaxter (commented):
I noticed an uptick in the Harmony Github traffic page, and saw that you all began using Harmony.

First off, thank you! It makes me happy seeing that this library is helpful for others.

Secondly, I noticed some potential, yet minor, bugs in this PR, so leaving a few comments for you all to consider.

Feel free to reach out if there's any questions/suggestions for Harmony as well. I wrote it out of frustration with how Android gave up on multi-process shared preferences, without any decent alternative that is purely for Android.

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/prefs/VpnSharedPreferencesProvider.kt
```diff
@@ -19,24 +19,75 @@ package com.duckduckgo.mobile.android.vpn.prefs
...
+            }
+        }
+
+        destination.edit(commit = true) { putBoolean(MIGRATED_TO_HARMONY, true) }
```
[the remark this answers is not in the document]
aitorvs: Thanks for the feedback @pablobaxter. Harmony came in the right moment for us. We were feeling a bit of pain because how shared prefs deal with multiprocess. The library is great.

> I would recommend storing both preferences in a class variable, and returning the correct object via this getSharedPreferences()

Do you mean having like a `Map<String, SharedPreference>` to have those shared prefs instances in memory so that we don't need to go through `context.getSharedPreferences` or `context.getHarmonySharedPreferences` every time?

> all or nothing migration

huh, that's a good point. A tricky one also. The main issue for us is that we know shared preferences doesn't deal with multi-process properly, causing bugs because many times values are not synced across processes when we need it.
1. If we do all or nothing migration, when it's nothing we'll returned the android shared prefs, which we know it will fail sometimes.
2. If we don't do all or nothing and (right now) assumed migration will just be OK, it may not be. The result in this case is that the first value may not be correct, but the subsequent ones would be, given than Harmony is much more reliable than shared prefs for multi-process.

I think I prefer 2/ over 1/. Does that make sense?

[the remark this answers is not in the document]
pablobaxter: > Do you mean having like a Map<String, SharedPreference> to have those shared prefs instances in memory so that we don't need to go through context.getSharedPreferences or context.getHarmonySharedPreferences every time?

Ah, never mind. You are calling for the preference file via the function you provided. Both Harmony and Android already store the prefs implementation as a singleton internally. What I was alluding to was that if you call for the prefs first, then immediately call any `get` function, it could block until the underlying prefs file has loaded. This is true for both SharedPreferences and Harmony. By eagerly calling for preferences, you give both Android and Harmony ample time to load preferences, so you don't risk a thread blocking `get` or `edit` call.

Basically, something like this would be bad:

```
fun getSomeValue(): String {
    val prefs = context.getSharedPreferences(PREFS_NAME, MODE_PRIVATE) // First call for prefs, which loads the prefs file under the covers
    return prefs.getString("blah", null) // This would most likely block if the prefs object didn't finish loading the file
}
```

But this would improve the performance of `getSomeValue()` in the example above:

```
private val prefs = context.getSharedPreferences(PREFS_NAME, MODE_PRIVATE) // First call for prefs, which loads the prefs file under the covers
fun getSomeValue(): String {
    return prefs.getString("blah", null) // Probably won't block if ample time occurred between the prefs call above and this call
}
```

> huh, that's a good point. A tricky one also. The main issue for us is that we know shared preferences doesn't deal with multi-process properly, causing bugs because many times values are not synced across processes when we need it.
>
>    If we do all or nothing migration, when it's nothing we'll returned the android shared prefs, which we know it will fail sometimes.
>    If we don't do all or nothing and (right now) assumed migration will just be OK, it may not be. The result in this case is that the first value may not be correct, but the subsequent ones would be, given than Harmony is much more reliable than shared prefs for multi-process.
>
> I think I prefer 2/ over 1/. Does that make sense?

The problem with 2 is that you might set `MIGRATED_TO_HARMONY` to `true` even though one of the values above may have failed (apply doesn't return a boolean status, and fails silently), in which case, you won't retry to migrate again. I'm not saying failures will happen (they are rare for both Harmony and SharedPreferences), but they can happen. It's a risk, but just pointing out the options here.

PR #1954, opened 2022-05-24
Print feature

nalcalag:
Task/Issue URL: https://app.asana.com/0/72649045549333/1201387054895492/f

### Description
Allow users to print page and also react to buttons for printing from webView (`window.print()` JS)

### Steps to test this PR

_Use Case 1_
- Install from this branch
- Open any website you want to print
- Select Overflow menu > Print Page
- [x] Check print dialog show
- [x] Save as pdf

_Use Case 2_
- Install from this branch
- Go to open tabs screen
- Select Overflow menu
- [x] Check 'Print Page' menu item is not there

_Use Case 3_
- Install from this branch
- Open a new tab
- Select Overflow menu
- [x] Check 'Print Page' menu item is not there

_Use Case 4_
- Install from this branch
- Open any website with a print button on it
- Click on the button
- [x] Check print dialog show
- [x] Save as pdf

### UI changes
| Before  | After |
| ------ | ----- |
Screenshot_20220524-170246_DuckDuckGo|Screenshot_20220523-173941_DuckDuckGo|


app/src/main/java/com/duckduckgo/app/browser/print/PrintDocumentAdapterWrapper.kt
```diff
@@ -0,0 +1,51 @@
...
+package com.duckduckgo.app.browser.print
+
+import android.os.Bundle
+import android.os.CancellationSignal
+import android.os.ParcelFileDescriptor
+import android.print.PageRange
+import android.print.PrintAttributes
+import android.print.PrintDocumentAdapter
+
+class PrintDocumentAdapterWrapper(private val delegate: PrintDocumentAdapter) : PrintDocumentAdapter() {
```
anikiki, line 26: This class is not needed if we don't want to notify the activity when the printing finishes.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2766,6 +2778,19 @@ class BrowserTabFragment :
         }
     }
 
+    private fun launchPrint(url: String) {
+        (activity?.getSystemService(Context.PRINT_SERVICE) as? PrintManager)?.let { printManager ->
+            webView?.createPrintDocumentAdapter(url)?.let { printAdapter ->
+                val wrapper = PrintDocumentAdapterWrapper(printAdapter)
+                printManager.print(
+                    url,
+                    wrapper,
```
anikiki, line 2787: Use the `printAdapter` directly here as the wrapper is not needed anymore.

app/src/main/java/com/duckduckgo/app/browser/print/PrintDocumentAdapterWrapper.kt
nalcalag: True! I'll remove it 👍

app/src/main/java/com/duckduckgo/app/browser/print/PrintJavascriptInterface.kt
```diff
@@ -0,0 +1,31 @@
...
+class PrintJavascriptInterface(private val onPrintDetected: () -> Unit) {
+
+    @JavascriptInterface
+    fun print() {
+        onPrintDetected()
+    }
+
+    companion object {
+        const val JAVASCRIPT_INTERFACE_NAME = "Print"
+    }
```
aitorvs, line 30: nit: is there any reason for this constant to be in this class as it's never access from here?

app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
```diff
@@ -374,6 +378,11 @@ class BrowserModule {
         return AndroidFileDownloader(dataUriDownloader, networkFileDownloader)
     }
 
+    @Provides
+    fun printInjector(): PrintInjector {
+        return PrintInjectorJS()
+    }
+
```
aitorvs, line 385: nit: I know we have a mixture of patterns here, ie. sometimes we use provider methods in dagger modules, sometimes we directly annotated the provided class.
I personally prefer the later, eg.
```kotlin
class PrintInjectorJS @Inject constructor() : PrintInjector {
}
```

My reasons for that are:
* just looking at the `PrintInjectorJS` class I can see how it is provided in DI
* tomorrow we need to add dependencies to `PrintInjectorJS` and we only need to modify one file
* dagger modules couple the dagger graph with the build graph

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2422,6 +2426,19 @@ class BrowserTabViewModel @Inject constructor(
...
 
+    fun onPrintSelected() {
+        url?.let {
+            pixel.fire(AppPixelName.MENU_ACTION_PRINT_PRESSED)
+            command.value = PrintLink(removeAtbAndSourceParamsFromSearch(it))
+        }
+    }
+
+    fun printFromWebView() {
+        viewModelScope.launch(dispatchers.main()) {
```
anikiki, line 2437: The dispatcher can be omitted here as main is the default in viewModelScope. Also. you can call `onPrintSelected()` directly, without launching a coroutine.

app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
nalcalag: I also prefer constructor injection, I made it through provides method to be consistency with other JS injectors but I will be happy to change it 🙂

app/src/main/java/com/duckduckgo/app/browser/print/PrintJavascriptInterface.kt
nalcalag: This is the string name for the @JavascriptInterface. We create the instance outside this class giving it this `JAVASCRIPT_INTERFACE_NAME`, but then, if we want to call print function inside this interface, we'll need to use this interface name too. Therefore, I followed the same pattern you have with other JS interfaces (that I think it's good) and keep the interface name inside the companion object of the interface class although it is not called here. Does it make sense?

aitorvs: Makes sense 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
nalcalag: Yeah, it doesn't work unless you change to the main thread. Thanks a lot for the comment, I'm not very experienced with coroutines so I didn't know main was the default one and can be omitted 🙌

app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
nalcalag: I finally changed it to constructor injection. I thought that if someone creates another JSInterface in the future, they can have this one as an example to start doing it like this. Thanks for the suggestion @aitorvs  💯

anikiki (approved):

PR #1964, opened 2022-05-26
Add guard on long press handler to ensure DI initialization has happened

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202348097691007/f

### Description
We noted a small number of crashes around `viewModelFactory` not being initialized when long press of back button happened. 

Theory on root cause is that there exists a small window whereby `BrowserActivity` has created the new fragment and reassigned `currentTab` to point to it, but the fragment hasn't been fully committed yet. Any interaction with the fragment in that window will result in similar crashes because the lazy `viewModel` will touch the uninitialized `viewModelFactory`.

This PR doesn't change the way the fragments are created (which is a bigger undertaking). Instead it just adds a guard in the long press handler to check for initialization, otherwise, it drops the long press event.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2018,7 +2018,13 @@ class BrowserTabFragment :
     }
 
     fun onLongPressBackButton() {
-        viewModel.onUserLongPressedBack()
+        /*
+         It is possible that this can be invoked before Fragment is attached
+         If viewModelFactory isn't initialized, ignore long press
+         */
+        if (this::viewModelFactory.isInitialized) {
+            viewModel.onUserLongPressedBack()
+        }
     }
 
     private fun launchHideTipsDialog(


joshliebe (approved):
LGTM!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2018,7 +2018,13 @@ class BrowserTabFragment :
...
 
     fun onLongPressBackButton() {
-        viewModel.onUserLongPressedBack()
+        /*
+         It is possible that this can be invoked before Fragment is attached
+         If viewModelFactory isn't initialized, ignore long press
+         */
+        if (this::viewModelFactory.isInitialized) {
+            viewModel.onUserLongPressedBack()
+        }
```
aitorvs, line 2027: I think this is theoretically possible so worth a shot. I wonder why the initialisation would be that slow that the user would have time to click overflow menu and long press, that's prob seconds.

But, on the other side, this fragment has a ton of dependencies, so DoubleChecks on them can cause contingency

CDRussell:
Merged as hotfix; closing PR

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
CDRussell: It isn't from the overflow; it's long press on the system back button (both options are possible, but I know from the stack trace it's the system back button). So that is more easily accessible.

But also thinking this might be exacerbated when the main thread is getting hammered; perhaps the user has hit the back button and it's being interpreted as a long press (i've seen this happen sometimes under heavy main thread load).

I can replicate the crash by adding a synthetic delay before calling `commit` on the fragment transaction, and long pressing in this window.

It's not the most solid of theories, but might explain why it only affects small number of users since it needs a few things going wrong at the same time to happen. But, even if the theory is wrong, it's a safe enough change (albeit, ugly)

Issue #199, opened 2018-02-22
[Feature Request] See Desktop version of websites


leroivi:
Mobile version of websites are not always desirable, an option to always see Desktop version of website as long as it is enabled would be enjoyable

NB : Chrome already offers this feature

CDRussell:
Created a PR for this just half an hour ago! Will likely be in the next release (next week)

PR https://github.com/duckduckgo/Android/pull/198

PR #1987, opened 2022-06-07
Improved bookmarks screen performace.

anikiki:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202088634690441/f

### Description

Changes:

1/ Updated to insert all bookmarks in one go and all favorites in one go. 
2/ Removed the ListAdapter along with the DiffCallback from BookmarksAdapter / FavoritesAdapter. This was incredibly slow, taking more than 20 seconds to display a list, unlike a regular implementation of RecyclerView.Adapter.
3/ Removed the usage of findViewById() when updating the viewholder in both BookmarksAdapter and FavoritesAdapter. Updated the TwoLineListItem custom view to give access to the image.
4/ Updated the loadFavicon in both BookmarksAdapter and FavoritesAdapter to do all operations needed to find the favicon using the IO dispatcher. Updated the actual Glide call which loads the image to use the Main dispatcher and also forced cancelling image requests that are not needed (i.e the view is no longer on screen).
5/ Used StrictMode before and after the changes to prove the main thread is now free of unnecessary operations.

### Steps to test this PR

To see the issue:
- fresh install from develop
- import a large number of bookmarks / favorites. Example files are available in the asana task.
- while importing try to scroll up / down
- notice the app crashes

To test the fix:
- fresh install from this branch
- import a large number of bookmarks / favorites. Example files are available in the asana task.
- while importing try to scroll up / down
- the app should not crash
- additional tests: do a search in bookmarks

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
+++ b/app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
@@ -4,8 +4,9 @@ import android.content.ContentResolver
 import android.net.Uri
 import com.duckduckgo.app.bookmarks.db.BookmarkEntity
 import com.duckduckgo.app.bookmarks.db.BookmarksDao
+import com.duckduckgo.app.bookmarks.db.FavoriteEntity
+import com.duckduckgo.app.bookmarks.db.FavoritesDao
 import com.duckduckgo.app.bookmarks.model.BookmarksRepository
-import com.duckduckgo.app.bookmarks.model.FavoritesRepository
 import com.duckduckgo.app.bookmarks.model.SavedSite
 import org.jsoup.Jsoup
 
@@ -37,7 +38,7 @@ sealed class ImportSavedSitesResult {
 class RealSavedSitesImporter(
     private val contentResolver: ContentResolver,
     private val bookmarksDao: BookmarksDao,
-    private val favoritesRepository: FavoritesRepository,
+    private val favoritesDao: FavoritesDao,
     private val bookmarksRepository: BookmarksRepository,
     private val savedSitesParser: SavedSitesParser
 ) : SavedSitesImporter {
@@ -53,12 +54,18 @@ class RealSavedSitesImporter(
                 savedSitesParser.parseHtml(document, bookmarksRepository)
             }
 
-            savedSites.forEach {
-                when (it) {
-                    is SavedSite.Favorite -> favoritesRepository.insert(it)
-                    is SavedSite.Bookmark -> bookmarksDao.insert(BookmarkEntity(title = it.title, url = it.url, parentId = it.parentId))
-                }
+            savedSites.filterIsInstance<SavedSite.Bookmark>().map {
+                BookmarkEntity(title = it.title, url = it.url, parentId = (it as SavedSite.Bookmark).parentId)
+            }.apply {
+                bookmarksDao.insertList(this)
             }
+
+            savedSites.filterIsInstance<SavedSite.Favorite>().filter { it.url.isNotEmpty() }.map { site ->
+                FavoriteEntity(title = site.title.takeIf { it.isNotEmpty() } ?: site.url, url = site.url, position = site.position)
+            }.apply {
+                favoritesDao.insertList(this)
+            }
+
             ImportSavedSitesResult.Success(savedSites)
         } catch (exception: Exception) {
             ImportSavedSitesResult.Error(exception)
--- a/app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
+++ b/app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
@@ -56,14 +56,14 @@ class RealSavedSitesImporter(
 
             savedSites.filterIsInstance<SavedSite.Bookmark>().map {
                 BookmarkEntity(title = it.title, url = it.url, parentId = (it as SavedSite.Bookmark).parentId)
-            }.apply {
-                bookmarksDao.insertList(this)
+            }.also {
+                bookmarksDao.insertList(it)
             }
 
             savedSites.filterIsInstance<SavedSite.Favorite>().filter { it.url.isNotEmpty() }.map { site ->
                 FavoriteEntity(title = site.title.takeIf { it.isNotEmpty() } ?: site.url, url = site.url, position = site.position)
-            }.apply {
-                favoritesDao.insertList(this)
+            }.also {
+                favoritesDao.insertList(it)
             }
 
             ImportSavedSitesResult.Success(savedSites)
--- a/app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
+++ b/app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
@@ -56,7 +56,7 @@ class RealSavedSitesImporter(
             }
 
             savedSites.filterIsInstance<SavedSite.Bookmark>().map {
-                BookmarkEntity(title = it.title, url = it.url, parentId = (it as SavedSite.Bookmark).parentId)
+                BookmarkEntity(title = it.title, url = it.url, parentId = it.parentId)
             }.also {
                 it.asSequence().chunked(IMPORT_BATCH_SIZE).forEach { chunk ->
                     bookmarksDao.insertList(chunk)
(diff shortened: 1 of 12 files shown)


AndroidPoet:
if I may I would like to suggest you please remove "GlobalScope.launch(Dispatchers. Main)" and replace it with lifecycle aware coroutine scope, calling GlobalScope inside the adapter will launch a top-level coroutine and they are unbound and keeps running until finished or canceled. it might create a memory leak. GlobalScope is ok if we use it for the single task but using an inside adapter where the user will scroll a list and glide will load images it might create memory leak.

AndroidPoet:
Hi @anikiki I have created issues for memory leak https://github.com/duckduckgo/Android/issues/199
This is the report from "TabSwitcherActivity",
I have attached a screenshot for your reference,
 only "Favicon" is retaining 1.5 Mb this might be the reason for the slow performance of the bookmarks screen. I hope this will help. 
fevicon
0

anikiki:
> Hi @anikiki I have created issues for memory leak #199 This is the report from "TabSwitcherActivity", I have attached a screenshot for your reference, only "Favicon" is retaining 1.5 Mb this might be the reason for the slow performance of the bookmarks screen. I hope this will help. fevicon 0

Hey @AndroidPoet! Thanks for the input on these changes. At the moment this PR is still in progress (converted to draft) as other things need to be addressed, including what you mentioned in your first comment. I'm not actively working on it at the moment, however I'll come back to it as soon as I can.

AndroidPoet:
I guess there is no need to optimize this screen with coroutines, The main reason behind the slow performance is the Memory leak. I was testing the app the few main screens performed slowly due to deeply coupling of the objects. When those memory leaks will be fixed these screens will work perfectly. I am sharing these details so this might help in other areas where the actual reason might be related to a memory leak.

app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
```diff
@@ -53,12 +54,18 @@ class RealSavedSitesImporter(
...
             }
 
-            savedSites.forEach {
-                when (it) {
-                    is SavedSite.Favorite -> favoritesRepository.insert(it)
-                    is SavedSite.Bookmark -> bookmarksDao.insert(BookmarkEntity(title = it.title, url = it.url, parentId = it.parentId))
-                }
+            savedSites.filterIsInstance<SavedSite.Bookmark>().map {
+                BookmarkEntity(title = it.title, url = it.url, parentId = (it as SavedSite.Bookmark).parentId)
+            }.apply {
```
aitorvs, line 59: nit: I think `also {}` suits better here. Convention for `apply` is generally to mutate the receiver object and return it.

To be clear, not a blocking comment, just an FYI nit

aitorvs (dismissed):
Works as expected!

app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
```diff
@@ -53,12 +54,18 @@ class RealSavedSitesImporter(
...
+                bookmarksDao.insertList(this)
             }
+
+            savedSites.filterIsInstance<SavedSite.Favorite>().filter { it.url.isNotEmpty() }.map { site ->
+                FavoriteEntity(title = site.title.takeIf { it.isNotEmpty() } ?: site.url, url = site.url, position = site.position)
+            }.apply {
```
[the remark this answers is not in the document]
anikiki: Thanks, updated :+1:

app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
anikiki: Thanks, updated! :+1:

app/src/main/java/com/duckduckgo/app/global/view/FaviconImageView.kt
```diff
@@ -26,41 +26,50 @@ import androidx.core.net.toUri
...
-    val defaultDrawable = generateDefaultDrawable(this.context, domain)
-    Glide.with(context)
-        .load(file)
-        .diskCacheStrategy(DiskCacheStrategy.NONE)
-        .skipMemoryCache(true)
-        .transform(RoundedCorners(10))
-        .placeholder(defaultDrawable)
-        .error(defaultDrawable)
-        .into(this)
+    GlobalScope.launch(Dispatchers.Main) {
```
marcosholgado, line 43: Why are we using `GlobalScope` instead of `AppCoroutineScope`?

app/src/main/java/com/duckduckgo/app/bookmarks/ui/FavoritesAdapter.kt
```diff
@@ -171,9 +221,10 @@ sealed class FavoritesScreenViewHolders(itemView: View) : RecyclerView.ViewHolde
             }
         }
 
-        private fun loadFavicon(url: String) {
+        private fun loadFavicon(url: String, image: ImageView) {
             lifecycleOwner.lifecycleScope.launch {
-                faviconManager.loadToViewFromLocalOrFallback(url = url, view = itemView.findViewById(CommonR.id.image))
+                favIconRequestsChannel.send(FavIconRequest(url))
```
marcosholgado, line 226: Not 100% sure about this. Each time we send something to the channel it will access the disk even when is not needed. I think `faviconManager.loadToViewFromLocalOrFallback` could return true/false if a favicon already exists and in that case don't send to the channel.

app/src/main/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManager.kt
```diff
@@ -87,6 +87,20 @@ class DuckDuckGoFaviconManager constructor(
         }
     }
 
+    override suspend fun saveFaviconForUrl(url: String) {
```
marcosholgado, line 90: Should be tested?

app/src/main/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManager.kt
```diff
@@ -146,6 +160,11 @@ class DuckDuckGoFaviconManager constructor(
         }
     }
 
+    override suspend fun loadToViewFromLocalWithPlaceholder(tabId: String?, url: String, view: ImageView) {
```
marcosholgado, line 163: Should be tested?

marcosholgado (commented):
Left a few comments before approving.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksActivity.kt
```diff
@@ -147,35 +147,33 @@ class BookmarksActivity : DuckDuckGoActivity() {
 
     private fun observeViewModel(parentId: Long) {
         viewModel.viewState.observe(
-            this,
```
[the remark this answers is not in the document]
anikiki: Not "needed". It's the annoying `Lambda argument should be moved out of parentheses` and I did that so I won't see it all the time :)

app/src/main/java/com/duckduckgo/app/bookmarks/service/SavedSitesImporter.kt
```diff
@@ -53,12 +55,22 @@ class RealSavedSitesImporter(
                 savedSitesParser.parseHtml(document, bookmarksRepository)
             }
 
-            savedSites.forEach {
-                when (it) {
-                    is SavedSite.Favorite -> favoritesRepository.insert(it)
-                    is SavedSite.Bookmark -> bookmarksDao.insert(BookmarkEntity(title = it.title, url = it.url, parentId = it.parentId))
+            savedSites.filterIsInstance<SavedSite.Bookmark>().map {
+                BookmarkEntity(title = it.title, url = it.url, parentId = (it as SavedSite.Bookmark).parentId)
```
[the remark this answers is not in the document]
anikiki: Thanks, updated.

app/src/main/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManager.kt
anikiki: Missed that :sweat_smile: Added tests.

app/src/main/java/com/duckduckgo/app/browser/favicon/DuckDuckGoFaviconManager.kt
anikiki: Missed that :sweat_smile: Added tests.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/FavoritesAdapter.kt
anikiki: Thanks! Updated to not send the event when binding the view :+1:

marcosholgado (approved):
LGTM!

PR #2001, opened 2022-06-17
Update android library dependencies

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202464623944581/f

### Description
version.androidx.appcompat=1.4.1 -> 1.4.2
version.androidx.constraintlayout=2.1.3 -> 2.1.4
version.androidx.core=1.7.0 -> 1.8.0
version.com.airbnb.android..lottie=5.1.1 -> 5.2.0
version.kotlinx.coroutines=1.6.1 -> 1.6.2

### Steps to test this PR
General smoke test

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -9,17 +9,17 @@
 
 plugin.android=7.0.0
 
-version.androidx.appcompat=1.4.1
+version.androidx.appcompat=1.4.2
 
 version.androidx.arch.core=2.1.0
 
-version.androidx.constraintlayout=2.1.3
+version.androidx.constraintlayout=2.1.4
 
 version.androidx.localbroadcastmanager=1.1.0
 
 version.androidx.recyclerview=1.2.1
 
-version.androidx.core=1.7.0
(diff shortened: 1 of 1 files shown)


versions.properties
```diff
@@ -99,6 +99,7 @@ version.com.jakewharton.retrofit..retrofit2-kotlin-coroutines-adapter=0.9.2
 
 version.androidx.viewpager2=1.0.0
 
+## unused
```
nalcalag, line 102: Question: Why don't we remove this library if we are not using it?

nalcalag (approved):
Approved! I just left a question about the "unused" comment but not a blocker to merge 👍

versions.properties
CDRussell: Ah good question; I'd been assuming it was a `refreshVersions` bug but upon closer look, I think define `turbine` twice (slightly differently). One probably is genuinely unused.

I'll tidy that up before merging. Good spot!

PR #2003, opened 2022-06-17
Change the way modules are defined in gradle

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202115663541560/f

### Description
When we define gradle modules currently, there is a lot of boilerplate and repetition of the module name. This makes it harder to add new modules, harder to remove it later and is pretty error prone.

This PR introduces just a few subtle tweaks to the config which fixes the above. 

~Additionally, it moves all but the `app` module to a subfolder called `modules` which cleans up the top-level project, and makes it easier to find modules when navigating the project. It comes with a big footprint of change, but it's just from moving folders to the **modules** folder.~
(☝️ undone, after feedback against this)

#### Before (lots of bits to change after copying/pasting): 
```
include ":example-api"
project(":example-api").projectDir = new File("modules/example/example-api")
include ":example-impl"
project(":example-impl").projectDir = new File("modules/example/example-impl")
include ":example-store"; 
project(":example-store").projectDir = new File("modules/example/example-store")
```

#### After (change a single word):
````
"example".with {
     // you can mostly ignore this block as it's typically identical for each module, making it easy to copy/paste
}
````

### Steps to test this PR
Smoke test to make sure nothing is broken! But I would expect a compilation error if anything had been messed up.

### CODE REVIEW

File: app/build.gradle [hunk 1]
```diff
@@ -153,7 +153,7 @@ dependencies {
     anvil project(path: ':anvil-compiler')
     implementation project(path: ':anvil-annotations')
 
-    implementation project(path: ':appbuildconfig-api')
+    implementation project(path: ':app-build-config-api')
```
CDRussell, line 156: renamed for consistency with all other modules naming convention

settings.gradle
```diff
@@ -2,90 +2,96 @@ plugins {
...
+"voice-search".with {
+    include ":$it-api"; project(":$it-api").projectDir = new File("$it/$it-api")
+    include ":$it-impl"; project(":$it-impl").projectDir = new File("$it/$it-impl")
+    include ":$it-store"; project(":$it-store").projectDir = new File("$it/$it-store")
+}
 
-include ':bandwidth-impl'
-include ':bandwidth-store'
+"downloads".with {
+    include ":$it-api"; project(":$it-api").projectDir = new File("$it/$it-api")
```
karlenDimla, line 31: Could this be extracted to a function? I am not a gradle expert but seems like that would reduce the code repetition even more.

PR #2011, opened 2022-06-23
Support updating a login credential

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1202487666612311/f

### Description
Support updating a saved credential workflow

### Steps to test this PR

- [x] Visit https://trello.com/login (having not previously saved any credentials for it)
- [x] Attempt login. Verify you are prompted to save credentials.
- [x] Choose to save them.
- [x] Reload page, and verify you can autofill with the saved credentials from before
- [x] Reload again, and this time change the password before logging in
- [x] Verify you are prompted to **update** the login, not create a new one
- [x] Choose to update
- [x] Visit the credential management screen to verify the password was updated successfully
- [x] Revisit that login page. This time, enter a new username and attempt a login
- [x] Verify you are prompted to save a new login (not to update)


modules/autofill/autofill-api/src/main/java/com/duckduckgo/autofill/store/AutofillStore.kt
```diff
@@ -24,9 +24,19 @@ interface AutofillStore {
...
+    suspend fun updateCredentials(rawUrl: String, credentials: LoginCredentials)
 
     suspend fun getAllCredentials(): Flow<List<LoginCredentials>>
 
     suspend fun deleteCredentials(id: Int)
     suspend fun updateCredentials(credentials: LoginCredentials)
+
+    suspend fun containsCredentials(rawUrl: String, username: String, password: String): ContainsCredentialsResult
+
+    sealed interface ContainsCredentialsResult {
```
karlenDimla, line 36: NIT: Maybe this is better as an enum since all classes here are objects.

modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/ExistingCredentialMatchDetector.kt
```diff
@@ -0,0 +1,40 @@
...
+class ExistingCredentialStoreInterrogatingMatchDetector(private val autofillStore: AutofillStore) : ExistingCredentialMatchDetector {
+
+    override suspend fun determine(currentUrl: String, username: String, password: String): AutofillStore.ContainsCredentialsResult {
+        return autofillStore.containsCredentials(currentUrl, username, password)
+    }
+}
+
+@Module
+@ContributesTo(AppScope::class)
+object ExistingCredentialMatchDetectorStoreModule {
```
karlenDimla, line 34: NIT: Since these classes (a few other more I've noticed) are in the impl module, we should be able to use @ContributesBinding instead of creating a module

modules/autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/AutofillJavascriptInterfaceTest.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.mockito.kotlin.*
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class AutofillJavascriptInterfaceTest {
+
+    @get:Rule
+    var coroutineRule = CoroutineTestRule()
+
+    private val moshi = Moshi.Builder().build()
```
karlenDimla, line 50: NIT: I don't think this is used anywhere.

karlenDimla:
While testing the changes on this PR, I noticed that:
- When I have saved the credential and loaded the page, no key icon is visible BUT the area where they key icon used to be seems to work since when you click it, it launches the bottom sheet.
- When I enter a site with credentials saved, clicking on the login text area launches the bottom sheet. I think it used to only launch when clicking the key icon.

modules/autofill/autofill-api/src/main/java/com/duckduckgo/autofill/store/AutofillStore.kt
CDRussell: That's a good suggestion, but I'm tempted to keep it as `sealed` instead, though I don't have strong feelings either way.

1. If we do add further data to any of the result types, it's easier to do this with `sealed` approach
2. We tend to use `sealed` for results elsewhere too (I think more so than `enums`)

I'll leave it as is if there's no strong pushback

karlenDimla (approved):

PR #2014, opened 2022-06-23
Thread Safety for Selector Operations

nshuba:
Task/Issue URL: https://app.asana.com/0/1202279501986195/1202454731900853/f

### Description
**Background**: We were changing channel keys' interest sets in two threads (`TcpNetworkToDevice` and `TcpDeviceToNetwork`), which is unsafe [[1]](http://rox-xmlrpc.sourceforge.net/niotut/) [[2]](https://stackoverflow.com/a/12824411). To get around the problem, we were using `sleep(10)` in between `select` operations. Without the `sleep`, a deadlock occurs between the two threads (where one is stuck on `register` and the other on `select`), indicating a thread-safety issue. There is a related discussion [here](https://stackoverflow.com/a/3191377) on the dangers of using `sleep` to make Selectors behave.

**This PR**: Fixes how we use the Selector by making sure all Selector-related operations are done on a single thread - `TcpDeviceToNetwork`. This eliminates the need to `sleep` in between `select` operations, thereby improving performance.

### Steps to test this PR
- [x] smoke tests on AppTP
 - open apps that have trackers and verify it blocks
 - perform speed tests and verify the download/upload perf is similar or better than develop. Ping should be much faster (in my testing it went from ~24ms to ~9ms).
 - downloads and uploads work as in develop

### DIFF
--- a/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/tcp/TcpDeviceToNetwork.kt
+++ b/vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/tcp/TcpDeviceToNetwork.kt
@@ -60,7 +60,6 @@ import java.io.IOException
 import java.nio.ByteBuffer
 import java.nio.channels.CancelledKeyException
 import java.nio.channels.SelectionKey
-import java.nio.channels.SelectionKey.OP_READ
 import java.nio.channels.SelectionKey.OP_WRITE
 import java.nio.channels.Selector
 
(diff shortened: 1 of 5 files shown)


vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/tcp/TcpDeviceToNetwork.kt
```diff
@@ -312,13 +314,6 @@ class TcpDeviceToNetwork(
                 return
             }
 
-            if (!tcb.waitingForNetworkData) {
```
nshuba, line 315: This is not needed since we change the op to WRITE below

vpn/src/main/java/com/duckduckgo/mobile/android/vpn/processor/tcp/TcpPacketProcessor.kt
```diff
@@ -231,7 +231,6 @@ constructor(
             queues.networkToDevice.offerFirst(buffer)
 
             try {
-                selectionKey.cancel()
```
nshuba, line 234: Closing the channel cancels the keys, so we don't need to keep track of the `selectionKey` at all.

aitorvs (approved):
This is great @nshuba ! 
works as expected

PR #2017, opened 2022-06-28
Android: Allow the use of user certificates for nightly builds

anikiki:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202500711374963/f

### Description
Allowed user certificates for internal builds.

### Steps to test this PR

- checkout this branch
- build the release versions for play (`./gradlew assemblePR`), fdroid (`./gradlew assembleFR`), internal (`./gradlew assembleIR`) 
- install each of them and have Charles proxy configured
- verify you can see traffic in Charles only for the internal build


CDRussell:
Rather than having 3 network configs (fdroid, play, and internal) could we only add the `internal` override, and let it pick up the existing one otherwise? or are you forced to add it for the build types?

cmonfortep (approved):
LGTM

anikiki:
> Rather than having 3 network configs (fdroid, play, and internal) could we only add the `internal` override, and let it pick up the existing, default one otherwise? or are you forced to add it for all the build types?

This is the general approach when different configs are needed for different flavours. I'm not aware of a way to overcome the duplication.

PR #2019, opened 2022-06-29
Replace 'get' with indexing operator

TangHuaiZhe:
Task/Issue URL: 

### Description
Adopt idiomatic Kotlin , replace 'get' with the indexing operator.
### Steps to test this PR
No need, there is no bytecode difference between before and after this modification.


TangHuaiZhe:
Strange error:
image

PR #2021, opened 2022-06-29
Restore guard that was removed as a temporary fix

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1202525190970047/f

### Description
Remove a temporary fix now that the real fix is coming through a JS update.


CDRussell:
@karlenDimla, this will break the email sign in flow until the new JS is available (ETA: tomorrow). I will push the updated JS to this branch when it is available.

karlenDimla (approved):

PR #2025, opened 2022-06-30
Add flag to easily enable debug mode for autofill

CDRussell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202532114531456/f

### Description
Adds a boolean to enable autofill debug mode. This is will stay `false` for default so we always test against the production impl and only toggled locally to `true` for debugging autofill integration issues.

### Steps to test this PR
Set the flag to true and add a logcat filter for `console`; verify you see lots more console output
Set it back to false and verify everything working as expected

### DIFF
--- a/common/src/main/java/com/duckduckgo/app/autofill/JavascriptInjector.kt
+++ b/common/src/main/java/com/duckduckgo/app/autofill/JavascriptInjector.kt
@@ -43,7 +43,10 @@ class FileBasedJavaScriptInjector @Inject constructor() : JavascriptInjector {
 
     override fun getFunctionsJS(): String {
         if (!this::functions.isInitialized) {
-            functions = loadJs("autofill.js")
+            // this can be enabled to see more verbose output from the autofill JS; useful for debugging autofill-related issues
+            val debugMode = false
+
+            functions = loadJs(if (debugMode) "autofill-debug.js" else "autofill.js")
         }
         return functions
     }


common/src/main/java/com/duckduckgo/app/autofill/JavascriptInjector.kt
```diff
@@ -43,7 +43,10 @@ class FileBasedJavaScriptInjector @Inject constructor() : JavascriptInjector {
 
     override fun getFunctionsJS(): String {
         if (!this::functions.isInitialized) {
-            functions = loadJs("autofill.js")
+            // this can be enabled to see more verbose output from the autofill JS; useful for debugging autofill-related issues
+            val debugMode = false
```
karlenDimla, line 47: Suggestion: We could also align this with AppBuildConfig.isDebug so we allow debug every time it's a debug build.

CDRussell: Good idea, and that was my first thought too, but don't want the risk that we'll test everyday against different behaviour than we'd see in prod version. 

Want to test against the important one, and manually switch on extra debugging as a temporary local step.

CDRussell:
FYI @shakyShane in case you check out the native code in future

shakyShane:
Thanks @CDRussell  - also CC @GioSensation

PR #2027, opened 2022-06-30
Fix known issues for Autofill

karlenDimla:
Task/Issue URL: 
https://app.asana.com/0/0/1202518798495133/f
https://app.asana.com/0/0/1202518798495138/f

### Description
This PR includes:
- Add handling for Xiaomi devices when launching settings page
- Synchronizing functions accessing the Cipher and separating cipher for encryption and decryption

### Steps to test this PR

- [ ] Smoke test autofill

### DIFF
--- a/modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementDisabledMode.kt
+++ b/modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementDisabledMode.kt
@@ -64,7 +64,7 @@ class AutofillManagementDisabledMode : Fragment() {
         }
     }
 
-    @SuppressLint("InlinedApi")
+    @SuppressLint("InlinedApi", "DEPRECATION")
     private fun launchDeviceAuthEnrollment() {
         when {
             appBuildConfig.manufacturer == "Xiaomi" ->
--- a/modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementDisabledMode.kt
+++ b/modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementDisabledMode.kt
@@ -66,15 +66,18 @@ class AutofillManagementDisabledMode : Fragment() {
 
     @SuppressLint("InlinedApi")
     private fun launchDeviceAuthEnrollment() {
-        if (appBuildConfig.sdkInt >= Build.VERSION_CODES.R) {
-            requireActivity().startActivity(Intent(android.provider.Settings.ACTION_BIOMETRIC_ENROLL))
-        } else {
-            if (appBuildConfig.sdkInt >= Build.VERSION_CODES.P) {
+        when {
+            appBuildConfig.manufacturer == "Xiaomi" ->
+                // Issue on Xiaomi: https://stackoverflow.com/questions/68484485/intent-action-fingerprint-enroll-on-redmi-results-in-exception
+                requireActivity().startActivity(Intent(android.provider.Settings.ACTION_SETTINGS))
+            appBuildConfig.sdkInt >= Build.VERSION_CODES.R ->
+                requireActivity().startActivity(Intent(android.provider.Settings.ACTION_BIOMETRIC_ENROLL))
+            appBuildConfig.sdkInt >= Build.VERSION_CODES.P ->
                 requireActivity().startActivity(Intent(android.provider.Settings.ACTION_FINGERPRINT_ENROLL))
-            } else {
+            else ->
                 requireActivity().startActivity(Intent(android.provider.Settings.ACTION_SECURITY_SETTINGS))
-            }
         }
+
         requireActivity().finish()
     }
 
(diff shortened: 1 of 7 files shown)


secure-storage/secure-storage-impl/src/main/java/com/duckduckgo/securestorage/impl/encryption/EncryptionHelper.kt
```diff
@@ -69,66 +54,39 @@ interface EncryptionHelper {
 
 @ContributesBinding(AppScope::class)
 class RealEncryptionHelper @Inject constructor() : EncryptionHelper {
-    private val cipher = Cipher.getInstance(TRANSFORMATION)
+    private val encryptionCipher = Cipher.getInstance(TRANSFORMATION)
```
karlenDimla, line 57: The root cause here was Cipher is not thread safe. Prior to this change:
- There are 4 functions that use the same cipher instance: 2 encrypt methods and 2 decrypt methods.
- Cipher maintains an internal state. Calling .init() resets the states.
- On the event that there's a race condition that one of the 4 functions are called simultaneously before the functions complete, then some issues could be encountered like overwriting encryption and decryption state.

secure-storage/secure-storage-impl/src/main/java/com/duckduckgo/securestorage/impl/encryption/EncryptionHelper.kt
```diff
@@ -69,66 +54,39 @@ interface EncryptionHelper {
...
+    private val decryptionCipher = Cipher.getInstance(TRANSFORMATION)
```
karlenDimla, line 58: The fix I implemented has 3 parts:

1. I created a separate Cipher for encryption and decryption. This way we avoid the risk of encryption and decryption flow affecting each other since it is possible that we encrypt and decrypt at almost the same time.
2. I reduced the functions to just 1 encrypt and 1 decrypt functions. I delegated the string to byteArray transformation (vice versa) out of the EncryptionHelper.
3. To complete the fix, i synchronised the encrypt and decrypt functions so we know that only one encryption or one decryption process takes place at a time.

CDRussell: Sounds good to me, and thanks for the detailed explanation of the changes. 

My initial thought on an alternative approach would be to move the cipher creation inside the functions, so that they were local (and we wouldn't need to worry about synchronization), but that would be bad if `Cipher.getInstance(TRANSFORMATION)` was expensive to perform. I suspect it might be; do you happen to know?

I'm happy with the approach in the PR, but want to throw that in for consideration.

CDRussell:
I'm not sure the new intent is better; on the Xiaomi it takes you to a security page that doesn't have the device lock as an option. We're not [alone in finding this](https://stackoverflow.com/questions/68484485/intent-action-fingerprint-enroll-on-redmi-results-in-exception).

Given this is an edge case, can we consider taking them to the top-level settings for now? I'm weakly in favour of that but easily convinced we can do better.

CDRussell:
Couldn't smoke test directly, as autofill wasn't working. But I pulled latest from base branch into yours locally and confirmed it'll all work there.

secure-storage/secure-storage-impl/src/main/java/com/duckduckgo/securestorage/impl/encryption/EncryptionHelper.kt
karlenDimla: Will address this in a separate PR. For now, we keep it this way.

modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementDisabledMode.kt
```diff
@@ -66,15 +66,18 @@ class AutofillManagementDisabledMode : Fragment() {
...
     @SuppressLint("InlinedApi")
     private fun launchDeviceAuthEnrollment() {
-        if (appBuildConfig.sdkInt >= Build.VERSION_CODES.R) {
-            requireActivity().startActivity(Intent(android.provider.Settings.ACTION_BIOMETRIC_ENROLL))
-        } else {
-            if (appBuildConfig.sdkInt >= Build.VERSION_CODES.P) {
+        when {
+            appBuildConfig.sdkInt >= Build.VERSION_CODES.R ->
+                requireActivity().startActivity(Intent(android.provider.Settings.ACTION_BIOMETRIC_ENROLL))
+            appBuildConfig.manufacturer == "Xiaomi" ->
```
CDRussell, line 72: I don't know what will happen on a Xiaomi device newer than Android R. Is it worth doing the manufacturer check first so that all Xiaomi devices get the generic settings, regardless of OS version?

karlenDimla: Good point. Updated it for the sake of safety. I don't think it's a bad info anyway.

modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementDisabledMode.kt
```diff
@@ -66,15 +66,18 @@ class AutofillManagementDisabledMode : Fragment() {
 
     @SuppressLint("InlinedApi")
     private fun launchDeviceAuthEnrollment() {
```
CDRussell, line 68: We should probably add a suppress lint statement here since we handle the deprecation warning

```suggestion
    @Suppress("DEPRECATION")
    private fun launchDeviceAuthEnrollment() {
```

CDRussell:
Working as expected on Xiaomi 👍

PR #2060, opened 2022-07-13
Filter autofill credentials by specified subtype

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202582065335829/f

### Description
Filters credentials from being chosen as eligible for autofill unless the specified subtype matches (e.g., username, or password). This should improve autofill into multi-steps login forms.

### Steps to test this PR
Amazon.co.uk isn't yet correctly saving usernames, so that makes it a good test site.

- [ ] Log into amazon.co.uk
- [ ] Verify you are prompted to save
- [ ] Save credentials
- [ ] Log out and revisit login screen
- [ ] Tap on username field; verify autofill prompt **is not shown**
- [ ] Enter username and progress
- [ ] Tap on password field; verify **prompt is shown**
- [ ] Tap on (currently empty; will fix later) button to autofill; verify password is autofilled


modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementEditMode.kt
```diff
@@ -142,3 +140,8 @@ class AutofillManagementEditMode : Fragment() {
             }
     }
 }
+
+private fun String.convertBlankToNull(): String? {
+    if (this.isBlank()) return null
```
karlenDimla, line 145: NIt: could b e simplified to:
`
private fun String.convertBlankToNull(): String? = if (this.isBlank()) null else this
`

modules/autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/AutofillJavascriptInterfaceTest.kt
```diff
@@ -108,6 +109,139 @@ class AutofillJavascriptInterfaceTest {
...
+                loginCredential(username = null, password = "password3"),
+
+            )
+        )
+        initiateGetAutofillDataRequest()
+        assertCredentialsAvailable()
+        assertCredentialsContains({ it.password }, "password1", "password2", "password3")
+    }
+
+    private fun assertCredentialsContains(property: (LoginCredentials) -> String?, vararg expected: String?) {
```
karlenDimla, line 219: I think this can be simplified to:
```
private fun assertCredentialsContains(property: (LoginCredentials) -> String?, vararg expected: String?) {
        val numberExpected = expected.size
        val numberMatched = TestCallback.credentials?.filter { expected.contains(property(it)) }?.count()
        assertEquals("Wrong number of matched properties. Expected $numberExpected but found $numberMatched", numberExpected, numberMatched)
    }
```

modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/AutofillJavascriptInterface.kt
```diff
@@ -79,6 +83,18 @@ class AutofillJavascriptInterface(
         }
     }
 
+    private fun filterRequestedSubtypes(request: AutofillDataRequest, credentials: List<LoginCredentials>): List<LoginCredentials> {
```
karlenDimla, line 86: So subtype corresponds as well to what the user clicked, so for this scenario:
- User opens trello
- User saves password without username
- User clicks on username field

Actual result: no prompt is shown

Expected result: There's a login saved for trello but without username, so shouldn't it show still?

When I click on password field, the autofill prompt appears though.

karlenDimla: I tested all the steps above and it works as expected. Not sure if the scenario I mentioned above is an edge case but at least we have a bottom sheet autofill design on how to display when the login saved has no username but has password  for the domain.

modules/autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/AutofillJavascriptInterfaceTest.kt
CDRussell: Neat! good suggestion. done

modules/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/AutofillJavascriptInterface.kt
karlenDimla: Can we not these kinds of cases somewhere? Just to have the right expectation on unusual but not edge case-y scenarios.

PR #2065, opened 2022-07-14
Align how we define dependencies in app:build.gradle and other modules

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202597451015882/f

### Description
This PR removes the usage of migrates to the approach 1 mentioned in the linked asana task. This also fixes other usages of api project.

### Steps to test this PR

- [ ] Smoke test app

### DIFF
--- a/privacy-config/privacy-config-api/build.gradle
+++ b/privacy-config/privacy-config-api/build.gradle
@@ -28,7 +28,7 @@ java {
 
 dependencies {
 
-    api project(path: ':feature-toggles-api')
+    implementation project(path: ':feature-toggles-api')
 
     implementation Google.dagger
     implementation Kotlin.stdlib.jdk7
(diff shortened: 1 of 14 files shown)


privacy-config/privacy-config-api/build.gradle
```diff
@@ -28,7 +28,7 @@ java {
 
 dependencies {
 
-    api project(path: ':feature-toggles-api')
```
karlenDimla, line 31: I think this is actually a valid case depending on how you look at it. The api returns/uses/exposes a class from feature-toggles-api. So if we don't expose the :feature-toggles-api as an api project, then any module that would need to depend on privacy-config-api needs to automatically add the feature-toggles-api as a dependency as well.

Not 100% what is the better approach here.

marcosholgado (approved):
LGTM!

PR #2070, opened 2022-07-15
Increase test coverage for autofill related logic

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202611392391285/f

### Description
This PR includes the following:
- Add missing tests identified for autofill related classes/ logic
- Minor cleanup: Use AndroidJunit4 test runner instead of RobolectricTestRunner (http://robolectric.org/androidx_test/)
- Minor cleanup: Remove dagger/Anvil from autofill-store to align with our recommended standard

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -431,8 +431,8 @@ class BrowserTabViewModel @Inject constructor(
             object FinishTrackerAnimation : DaxCommand()
             class HideDaxDialog(val cta: Cta) : DaxCommand()
         }
-        data class InjectCredentials(val url: String, val credentials: LoginCredentials) : Command()
-        data class CancelIncomingAutofillRequest(val url: String) : Command()
+        class InjectCredentials(val url: String, val credentials: LoginCredentials) : Command()
+        class CancelIncomingAutofillRequest(val url: String) : Command()
         class EditWithSelectedQuery(val query: String) : Command()
         class ShowBackNavigationHistory(val history: List<NavigationHistoryEntry>) : Command()
         class NavigateToHistory(val historyStackIndex: Int) : Command()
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -431,8 +431,8 @@ class BrowserTabViewModel @Inject constructor(
             object FinishTrackerAnimation : DaxCommand()
             class HideDaxDialog(val cta: Cta) : DaxCommand()
         }
-        class InjectCredentials(val url: String, val credentials: LoginCredentials) : Command()
-        class CancelIncomingAutofillRequest(val url: String) : Command()
+        data class InjectCredentials(val url: String, val credentials: LoginCredentials) : Command()
```
CDRussell, line 434: why the introduction of `data` here? (noting that none of the other commands are defined that way)

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/jsbridge/response/AutofillJsonResponseWriterTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        val expectedJson = "userUnprotectedDomains = [];"
+        assertEquals(
+            expectedJson,
+            testee.generateUserUnprotectedDomains()
+        )
+    }
+
+    @Test
+    fun generateUserPreferences() {
+        val expectedJson = "userPreferences = {\n" +
```
CDRussell, line 107: It makes me a little anxious to do a literal string comparison for this, in that it might be unstable if the order of properties change, or the number of spaces in indentation changes etc...

But I think its addition as is is still an improvement of not having these tests.

CDRussell (approved):

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
karlenDimla: Mainly to simplify the comparison. But I can maybe just change this to check the attributes instead of exact instance, just for the sake of consistency. Is there a reason why these are not data classes?

CDRussell: Functionally, everything would work if they were all data classes but it would be a _slight_ unnecessary overhead to generate the extra things data classes generate when we mostly never need to use them again.

They are generally used as a one-time indication an event has happened (and with an associated transfer of data about that event), and not typically used for equality.

PR #2073, opened 2022-07-15
Add delay before showing autofill save/update prompts

CDRussell:
Without this delay, the form submission / navigation web events can be blocked, meaning we save the credentials but the user is left on the form, needing to hit the submission button again.

Task/Issue URL: https://app.asana.com/0/488551667048375/1202612639205162/f

### Description
As noted in the linked task, we can hit a problem with forms not submitting correctly if we immediately show a DialogFragment from the @Javainterface function. This is a strange problem, but also one hit on other platforms like iOS.

The workaround is to introduce a short delay before showing the DialogFragments for saving or updating a credential

Also included is a small tidy-up of coroutine usage around these areas, which weren't quite right before.

### Steps to test this PR

1. Visit https://trello.com/login
2. Enter credentials and hit login button
3. Verify save dialog is shown
4. Also verify the form was submitted (either you'll be logged in, or you'll at least see error message about incorrect login). On `develop`, the page wouldn't change and you'd have to tap login button again

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/ExistingCredentialStoreInterrogatingMatchDetector.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/ExistingCredentialStoreInterrogatingMatchDetector.kt
@@ -19,6 +19,8 @@ package com.duckduckgo.autofill.ui
 import com.duckduckgo.autofill.store.AutofillStore
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesBinding
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import javax.inject.Inject
 
 @ContributesBinding(AppScope::class)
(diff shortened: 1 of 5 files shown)


autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/ExistingCredentialStoreInterrogatingMatchDetector.kt
```diff
@@ -19,13 +19,17 @@ package com.duckduckgo.autofill.ui
...
+import kotlinx.coroutines.withContext
 import javax.inject.Inject
 
 @ContributesBinding(AppScope::class)
 class ExistingCredentialStoreInterrogatingMatchDetector @Inject constructor(private val autofillStore: AutofillStore) :
     ExistingCredentialMatchDetector {
 
     override suspend fun determine(currentUrl: String, username: String, password: String): AutofillStore.ContainsCredentialsResult {
-        return autofillStore.containsCredentials(currentUrl, username, password)
+        return withContext(Dispatchers.IO) {
```
karlenDimla, line 31: Why are we not using DispatcherProvider.io here?

karlenDimla (approved):
Just 1 question, but I tested the scenario and it works as expected. Out of 5 attempts, I haven't encountered any scenario where the autofill didn't proceed.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/ExistingCredentialStoreInterrogatingMatchDetector.kt
CDRussell: Laziness 😊

Will switch to that.

PR #2077, opened 2022-07-18
Add custom lint rule

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1202620566951065

### Description
This PR adds a custom lint rule to err when @Singleton is used.

### Steps to test this PR
- [x] Change a @SingleInstanceIn annotation for a @Singleton and run lint
- [x] Lint should fail

Note: The whole build will probably fail regardless so just build as usual, then in an impl module change the annotation and run lint.


cmonfortep (approved):
LGTM

PR #2078, opened 2022-07-19
Address remaining comments on MVP autofill

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202621847351497/f

### Description
This PR includes:
- Remove injected context in ActivityScoped AutofillManagement Screen ViewModel
- Move device auth check from browser fragment into AutofillJavascript to have a single place that defines availability of autofill. This also fixes the issue where when device auth is removed from device anduser attempts to autofill aftyer refreshing website. Previously, autofill is still enabled (bottom sheet is shown) but can't autofill since no device auth is available. Now that is fixed.

### Steps to test this PR

_Clipboard changes_
- [x] Add a new credential
- [x] Go to autofill management 
- [x] Copy username and password and verify that they are being copied correctly.

_Auth check changes_
- [x] Enrol device auth in device
- [x] Save a credential for a website
- [x] Attempt to autofill the website with previously saved credential. 
- [x] Confirm that this works
- [x] Remove device auth for device
- [x] Refresh page and attempt to autofill
- [x] Confirm that autofill bottom sheet doesn't show.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -193,7 +193,6 @@ import com.duckduckgo.autofill.CredentialUpdateExistingCredentialsDialog.Compani
 import com.duckduckgo.autofill.domain.app.LoginCredentials
 import com.duckduckgo.autofill.store.AutofillStore.ContainsCredentialsResult.*
 import com.duckduckgo.autofill.ui.ExistingCredentialMatchDetector
-import com.duckduckgo.deviceauth.api.DeviceAuthenticator
 import com.duckduckgo.di.scopes.FragmentScope
 import com.duckduckgo.voice.api.VoiceSearchLauncher
 import com.duckduckgo.voice.api.VoiceSearchLauncher.Source.BROWSER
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1580,20 +1576,18 @@ class BrowserTabFragment :
     }
 
     private fun configureWebViewForAutofill(it: DuckDuckGoWebView) {
-        if (deviceAuthenticator.hasValidDeviceAuthentication()) {
-            browserAutofill.addJsInterface(it, autofillCallback)
+        browserAutofill.addJsInterface(it, autofillCallback)
```
karlenDimla, line 1579: This now means that the JS interface and result listeners are added regardless if the feature is enabled. I think this is OK since they wouldn't be called anyway if the feature is disabled. 

This just means that we won't have to add these things when the user enables (adds device auth, becomes an internal user or enables the feature through settings) the autofill feature since it is already there.

CDRussell (approved):
I assume these testing steps are in the wrong order; otherwise, i've missed a test case so let me know.
  - Save a credential for a website
  - Enrol device auth in device


Working great, thanks

PR #2082, opened 2022-07-20
Migrate from @OnLifecycleEvent: common-ui module

nalcalag:
Task/Issue URL: https://app.asana.com/0/0/1202260594569384/f

### Description

@OnLifecycleEvent is now deprecated. Updated code in common-ui module to use `DefaultLifecycleObserver`

### Steps to test this PR

- Install from this branch
- Launch the app
- Open any fragment which uses viewBinding (e.g. Email Protection)
- Check in the Logcat that the following Logs appear:
  - [ ] Registering fragment lifecycle observer: com.duckduckgo.app.email.ui.EmailProtectionSignInFragment

### DIFF
--- a/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/viewbinding/FragmentViewBindingDelegate.kt
+++ b/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/viewbinding/FragmentViewBindingDelegate.kt
@@ -24,7 +24,6 @@ import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import androidx.viewbinding.ViewBinding
-import timber.log.Timber
 import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KProperty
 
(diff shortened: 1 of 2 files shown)


common-ui/src/main/java/com/duckduckgo/mobile/android/ui/viewbinding/FragmentViewBindingDelegate.kt
```diff
@@ -45,10 +46,14 @@ class FragmentViewBindingDelegate<T : ViewBinding>(
     init {
         fragment.viewLifecycleOwnerLiveData.observe(fragment) { lifecycleOwner ->
             lifecycleOwner.lifecycle.addObserver(
-                object : LifecycleObserver {
-                    @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
-                    fun onDestroy() {
+                object : DefaultLifecycleObserver {
+                    override fun onCreate(owner: LifecycleOwner) {
+                        Timber.d("Registering fragment lifecycle observer: ${fragment.javaClass.canonicalName}")
```
karlenDimla, line 51: I don't think we need to override onCreate just to log something right? I think we can just stick to overriding onDestroy. Wdyt?

nalcalag: Yeah, it is a bit confusing just override onCreate just for logging. I did it in the first place to test lifecycle was registered in the fragment and then, I decided to leave it for consistency when we register application lifecycle. But probably not worth it just to override for this.
Now that it' tested, I will remove the log 👍

karlenDimla (approved):

PR #2085, opened 2022-07-21
Create reusable OutlinedTextInput

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202628454434342/f

### Description
This PR creates a reusable TextInput initially aligned with the design to be used for Autofill. This can be further extended to be used on other parts of the app going forward (might require minor changes on demand in the future).

The new component is also added into the App Components Preview.

### Steps to test this PR

_View usages of the view on the preview_
- [ ] Open Settings
- [ ] Open "App components Design Preview"
- [ ] Open the "INTERACTIVE" tab
- [ ] Scroll all the way to the outlined text input
- [ ] Verify it looks OK in light and dark modes

### UI changes
| Light  | Dark |
| ------ | ----- |
|Light mode|Dark  mode|


common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/OutLinedTextInput.kt
```diff
@@ -0,0 +1,160 @@
...
+            endIconDrawable = drawable
+            isEndIconVisible = true
+            endIconContentDescription = contentDescription
+        }
+    }
+
+    private fun setupPasswordMode() {
+        binding.internalPasswordIcon.visibility = VISIBLE
+        binding.internalEditText.inputType = EditorInfo.TYPE_TEXT_VARIATION_PASSWORD or EditorInfo.TYPE_TEXT_FLAG_MULTI_LINE
+        // This has to be set before the transformationMethod. Else, it will force thas password to be visible.
```
CDRussell, line 119: typo: thas

common-ui/src/main/res/values/widgets.xml
```diff
@@ -173,7 +173,8 @@
         <item name="android:paddingBottom">14dp</item>
         <item name="android:paddingStart">@dimen/settingsItemHorizontalPadding</item>
         <item name="android:paddingEnd">@dimen/settingsItemHorizontalPadding</item>
-        <item name="android:textAppearance">@style/Widget.DuckDuckGo.SectionHeader.TextAppearance</item>
+        <item name="android:textAppearance">@style/Widget.DuckDuckGo.SectionHeader.TextAppearance
```
CDRussell, line 176: nit: undesireable formatting change

common-ui/src/main/res/values/widgets.xml
```diff
@@ -213,7 +214,8 @@
 
     <style name="Widget.DuckDuckGo.Button.Secondary">
         <item name="android:textColor">@color/cornflowerBlue</item>
-        <item name="android:background">@drawable/background_transparent_button_rounded_corners</item>
+        <item name="android:background">@drawable/background_transparent_button_rounded_corners
```
CDRussell, line 217: nit: undesireable formatting change

CDRussell (approved):

CDRussell:
Just noticed something: the password field with the copy and show password buttons, the buttons disappear when you focus on them and they don't come back. I don't think that's intentional?

CDRussell:
> @CDRussell they should come back once you change the focus - so by clicking on another textinput or when you make it uneditable again. Doesn't that happen for you? Can you give me steps to repro if ever.

That wasn't happening for me, and I'll get a screen recording soon. 

However, stepping back from that, even from your description that sounds a little off UX-wise to me. If I'm in the middle of typing a long password, there's a good chance I want to toggle seeing the password right there, without the need to defocus that field first. 

Are you sure we want to be hiding the show password button while editing?

CDRussell:
This is looking better to me! My only slight nit is that the vertical size of the password field changes when you reveal the text and shrinks when you hide it again, which is a little jarring. I guess that might always happen depending on the font used, but wonder if we could add some line spacing or such so that it doesn't happen in most cases, and you could toggle the password to reveal/hide without anything noticeably resizing?

karlenDimla:
@CDRussell I've updated this now to the logic agreed in https://app.asana.com/0/0/1202649606707235/f Let me know if you would want to look again. Else, I'll just merge it since it has been approved.

CDRussell:
Haven't re-tested, but feel free to merge. We'll have time to tweak if it's needing anything more.

Thanks for doing it, great to see components like this being created.

PR #2098, opened 2022-07-25
Support logins with missing usernames

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1202630343540911/f

### Description
Supports savings logins without usernames. This includes accepting the login, the "save" dialog and the "choose login" dialog when autofilling.

#### Saving a credential without a username (no previous credentials ever saved)
**Setup**
I don't have a suggestion for a site which only saves the password, so need to hack this unless you know one. 
In `BrowserTabFragment.onCredentialsAvailableToSave()`, add this as the first line of the function to fake the username coming through as `null` 
👉 `val credentials = credentials.copy(username = null)`

- [x] Visit a site and attempt a login with the hack above in place (e.g., https://trello.com/login)
- [x] Verify the onboarding dialog shows
- [x] Verify the title ends "save your Login"
- [x] Verify the subtitle starts "Logins are..."
- [x] Verify the main CTA is "Save Login"
- [x] Save the credential
- [x] Visit credential management, and verify it has saved the password but not the username

#### Saving a credential without a username (no onboarding)

**Setup**
- Ensure the hack from the previous test case is still in place.
- Ensure a previous login was saved (so onboarding dialog not shown)

**Instructions**
- [x] Delete the existing login from credential management for trello
- [x] Visit a site with a login and attempt to log in (e.g., https://trello.com/login)
- [x] Verify the save dialog shows
- [x] Verify the title is "Save **Password**?"
- [x] Verify there is no onboarding subtitle
- [x] Verify the main CTA is "Save **Password**"
- [x] Save the credential
- [x] Visit credential management, and verify it has saved the password but not the username

#### Testing Autofilling into a site with username missing
Easiest way to test this scenario is to manually edit an existing login; edit it by manually deleting the username and save. (or keep the incomplete entry created from the hack above if you have one saved already) 

- [x] Load a site which shows username and password (e.g., https://trello.com/login)
- [x] Tap on the username field; verify no prompt is shown yet
- [x] Tap on the password field; verify prompt is shown
- [x] Verify the text on the CTA is "Login for trello.com"
- [x] Accept autofill; verify only the password field is populated

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/saving/AutofillSavingCredentialsViewModel.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/saving/AutofillSavingCredentialsViewModel.kt
@@ -16,8 +16,11 @@
 
 package com.duckduckgo.autofill.ui.credential.saving
 
+import androidx.annotation.StringRes
 import androidx.lifecycle.ViewModel
 import com.duckduckgo.anvil.annotations.ContributesViewModel
+import com.duckduckgo.autofill.domain.app.LoginCredentials
+import com.duckduckgo.autofill.impl.R
 import com.duckduckgo.autofill.store.AutofillStore
 import com.duckduckgo.di.scopes.ActivityScope
 import javax.inject.Inject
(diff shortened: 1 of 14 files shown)


autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/saving/AutofillSavingCredentialsViewModel.kt
```diff
@@ -31,4 +34,45 @@ class AutofillSavingCredentialsViewModel @Inject constructor() : ViewModel() {
     fun showOnboarding(): Boolean {
         return autofillStore.showOnboardingWhenOfferingToSaveLogin
     }
+
+    fun determineTextResources(credentials: LoginCredentials): DisplayStringResourceIds {
```
karlenDimla, line 38: NIT: This could be part of the viewmodel's view state or maybe the LoginCredentials could be kept in the ViewModel rather than the fragment.

autofill/autofill-impl/src/main/res/values/autofill-impl-strings-untranslated.xml
```diff
@@ -24,13 +24,17 @@
     <string name="managementDisabledModeTitle">Secure your device to use Autofill</string>
 
     <string name="saveLoginDialogTitle">Save Login?</string>
+    <string name="saveLoginMissingUsernameDialogTitle">Save Password?</string>
     <string name="saveLoginDialogButtonSave">Save Login</string>
+    <string name="saveLoginMissingUsernameDialogButtonSave">Save Password</string>
     <string name="saveLoginDialogNotNow">Not Now</string>
     <string name="saveLoginDialogOnboardingExplanationSubtitle">Logins are stored securely on this device only, and can be managed from the Autofill menu in Settings.</string>
-    <string name="saveLoginDialogFirstTimeExplanationTitle">Do you want DuckDuckGo\nto save your Login?</string>
+    <string name="saveLoginDialogFirstTimeOnboardingExplanationTitle">Do you want DuckDuckGo\nto save your Login?</string>
```
karlenDimla, line 32: NIT: This wouldn't yield a necessarily good result on bigger fonts. Why do we want to force a line break here instead of letting the system handle it?

karlenDimla (approved):
I think the changes look good over all.

autofill/autofill-impl/src/main/res/values/autofill-impl-strings-untranslated.xml
CDRussell: Good point. I was using the line break as a shortcut to achieve a balanced layout of the words. What I actually wanted was to use `android:breakStrategy="balanced"`.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/saving/AutofillSavingCredentialsViewModel.kt
CDRussell: I agree it'd be better wiring it in through the view state, and that's likely what we'll end up doing anyway, but wanted to minimise changes on that right now.

I'm going to defer such further changes to another PR.

PR #2101, opened 2022-07-27
Autofill engine: support saving and updating last updated time and notes along with credentials

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202666734291055/f

### Description
This PR includes:
- Saving of last updated time upon saving of the credential
- Update of last updated time (UI and DB) when saving edits to a credential
- Update of notes and domain title(UI and DB) when saving edits to a credential
- Move reference to currently being edited credential from the CredentialModeFragment to the ViewModel removing any reference to the state from the fragment.

### Steps to test this PR

_Save credential saves last updated time_
- [x] Open a website
- [x] Save credential
- [x] Go to autofill management screen and open the credential you saved
- [x] Confirm that last updated time is displayed and is correct

_Updating credentials persists_
- [x] Go to autofill management screen and select any saved credential
- [x] Select edit from the context menu
- [x] Update all fields
- [x] Press the check icon to save the changes
- [x] Confirm that view mode contains all updated data and the last updated time is updated. If you changed the domain, it is possible that the icon in the toolbar will not show if we haven't loaded the icon yet. For icons that already exist, it should change automatically.
- [x] Exit autofill management screen and comeback
- [x] Confirm that edited credentials is persisted

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/di/AutofillModule.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/di/AutofillModule.kt
@@ -21,6 +21,7 @@ import com.duckduckgo.autofill.InternalTestUserChecker
 import com.duckduckgo.autofill.store.AutofillStore
 import com.duckduckgo.autofill.store.InternalTestUserStore
 import com.duckduckgo.autofill.store.RealInternalTestUserStore
+import com.duckduckgo.autofill.store.RealLastUpdatedTimeProvider
 import com.duckduckgo.autofill.store.SecureStoreBackedAutofillStore
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.securestorage.api.SecureStorage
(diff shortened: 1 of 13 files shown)


autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/di/AutofillModule.kt
```diff
@@ -43,6 +44,6 @@ class AutofillModule {
         context: Context,
         internalTestUserChecker: InternalTestUserChecker
     ): AutofillStore {
-        return SecureStoreBackedAutofillStore(secureStorage, context, internalTestUserChecker)
+        return SecureStoreBackedAutofillStore(secureStorage, context, internalTestUserChecker, RealLastUpdatedTimeProvider())
```
CDRussell, line 47: Should we declare a `@Provides` or `@ContributesBinding` for the time provider, instead of instantiating here? 

functionally, not going to matter too much, but more in line with how we do it elsewhere

CDRussell (approved):

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/di/AutofillModule.kt
karlenDimla: I added RealLastUpdatedTimeProvider / LastUpdatedTimeProvider in autofill-store so we don't have DI there that's why I chose to do it this way.

PR #2103, opened 2022-07-27
Remove api dependencies from privacy-config-api and vpn-api modules

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1202666690958570

### Description
This PR removes the dependency from feature-toggles-api in the privacy-config-api and vpn-api modules. It removes FeatureName and swaps it to a String.

### Steps to test this PR

_Remote config_
- [x] All tests pass
- [x] Try a few remote features and make sure they work as expected, some examples below.
- [x] Install and launch from this branch
- [x] Filter by user agent in the logcat
- [x] Go to thingiverse.com
- [x] User agent should not include the word `DuckDuckGo`
- [x] Go to `welt.de`
- [x] Privacy grade should be solid black
- [x] Go to the privacy dashboard (click on the privacy grade)
- [x] There should be a message saying protections are temporarily disabled due to breakage
- [x] Check the toggles, privacy config and vpn DB to make sure the data inserted matches the date from `https://staticcdn.duckduckgo.com/trackerblocking/config/v1/android-config.json`

### DIFF
--- a/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggle.kt
+++ b/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggle.kt
@@ -24,35 +24,7 @@ interface FeatureToggle {
      * @throws [IllegalArgumentException] if the feature is not implemented
      */
     fun isFeatureEnabled(
-        featureName: FeatureName,
+        featureName: String,
         defaultValue: Boolean = true
     ): Boolean
-}
-
-/**
- * Each feature toggle created needs a [FeatureName] which can be implemented using this interface
- */
-interface FeatureName {
-    val value: String
-
-    companion object {
-        /**
-         * Utility function to create a [FeatureName] from the passed in [block] lambda
-         * instead of using the anonymous `object : FeatureName` syntax.
-         *
-         * Usage:
-         *
-         * ```kotlin
-         * val feature = FeatureName {
-         *
-         * }
-         * ```
-         */
-        inline operator fun invoke(crossinline block: () -> String): FeatureName {
-            return object : FeatureName {
-                override val value: String
-                    get() = block()
-            }
-        }
-    }
 }
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -4188,7 +4188,7 @@ class BrowserTabViewModelTest {
 
     private fun givenUrlCannotUseGpc(url: String) {
         val exceptions = CopyOnWriteArrayList<GpcException>().apply { add(GpcException(url)) }
-        whenever(mockFeatureToggle.isFeatureEnabled(eq(PrivacyFeatureName.GpcFeatureName), any())).thenReturn(true)
+        whenever(mockFeatureToggle.isFeatureEnabled(eq(PrivacyFeatureName.GpcFeatureName.value), any())).thenReturn(true)
         whenever(mockGpcRepository.isGpcEnabled()).thenReturn(true)
         whenever(mockGpcRepository.exceptions).thenReturn(exceptions)
     }
--- a/app/src/androidTest/java/com/duckduckgo/app/email/EmailInjectorJsTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/email/EmailInjectorJsTest.kt
@@ -47,7 +47,7 @@ class EmailInjectorJsTest {
         testee =
             EmailInjectorJs(mockEmailManager, DuckDuckGoUrlDetector(), mockDispatcherProvider, mockFeatureToggle, javascriptInjector, mockAutofill)
 
-        whenever(mockFeatureToggle.isFeatureEnabled(AutofillFeatureName)).thenReturn(true)
+        whenever(mockFeatureToggle.isFeatureEnabled(AutofillFeatureName.value)).thenReturn(true)
         whenever(mockAutofill.isAnException(any())).thenReturn(false)
     }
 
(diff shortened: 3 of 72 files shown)


feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggle.kt
```diff
@@ -24,35 +24,7 @@ interface FeatureToggle {
      * @throws [IllegalArgumentException] if the feature is not implemented
      */
     fun isFeatureEnabled(
-        featureName: FeatureName,
+        featureName: String,
```
cmonfortep, line 27: I have some doubts about this change:
I liked using a specific type instead of a primitive here. So I wonder if we should see this from a different perspective instead of finding a way to break the `api` dependency.

How I see it: 
- FeatureName restricts the possible value we pass into that function
- FeatureName clearly indicates that we have a set of features that can be enabled/disabled
- We had some method interfaces that used FeatureName as arguments.
- Then I see, that different modules can enable/disable features. I expect those modules to contribute to the app values that implement `FeatureName`
- Maybe this is a different type of `api`. Saying this because I see a difference between: modules that own some logic and we are only consumers, vs modules that we can contribute to.
- In this case I think we are contributors in a way. 

I see this very similar to how we consume `statistics` module for `Pixels`. We have a bunch of modules that are extending `Pixel.PixelName` interface, and I think that's correct.

Happy to know more about the approach here.

aitorvs: The rational (trade-off) behind is is the following (assuming we don't want to break the `api` not depending in `api` rule):
* The reason to have a the `FeatureName` type is to have type-safe code as much as possible
* The `PrivacyFeaturePlugin` plugins are internally not type-safe anyway because they have to do `FeatureName.value`. So in that sense, a `String` instead of a `FeatureName` doesn't make a difference
* So now it comes to the caller of the `isFeatureEnabled`
* In this case, the idea is that even though that `isFeatureEnabled` is now taking a `String` as a parameter, every `feature-api` module will expose an `enum` listing all its supported features
* Callers using that `enum` are still as type-safe as before
* The only nuance is that, having a `FeatureName` type in the `isFeatureEnable` signature makes clear(er) in the that the caller should look for a `FeatureName` (extended) type
* And that's the only thing we're giving up with this trade-off
* Which I think it's easily solvable with just Javadocs

marcosholgado: What @aitorvs said plus... you could also do `FeatureName { "something" }` before so it wasn't safe at all :). In any of the cases the app will crash if you use a featureName that doesn't exist but that's explained in the Javadocs.

cmonfortep: I agree "it's solvable", and that `FeatureName` is replaceable by `String` with the current logic. The other arguments about type-safe are implementation details and I wouldn't focus too much there.

My point is that we are going to have similar scenarios in the future where we **contribute** to a module rather than just **consume** the module. As I said, `Statistics` module is a similar example.

So I'm not advocating we should  allow depending on other modules `api`, I'm advocating we should have another type of module (not sure if we consider `Statistics` as an `api` module or not). And `feature-toggles` should be that type of module where others can depend on.

marcosholgado: In my head the only way you contribute to a module is usually done in the impl module. In your statistics example, the macOS waitlist needs that module, but the dependency happens in the macos-impl module rather than the macos-api.

aitorvs: > As I said, Statistics module is a similar example
I don't think `statistics` falls into the same category here. The `Pixel.PixelName` interface will only be extended in `impl` modules, because what pixels a module sends should always be an implementation detail.

cmonfortep: Thanks! You are right on this: "only way you contribute to a module is usually done in the impl module" + "The Pixel.PixelName interface will only be extended in impl modules". I'm going to take another look at the code to understand then why we can't do the same from impl modules.

cmonfortep: Ok, I have more context now, thanks.

Now I see the logic inside FeatureToggle is just to find an existing `FeatureTogglesPlugin` based on a key. Makes sense to me we are moving to `String` there. That was my main concern because I thought it was necessary that `PrivacyFeatureName` extended `FeatureName` for more than that.

I guess my next question is:
- I assume the change to move from `val featureName: FeatureName` to `val featureName: String` in `PrivacyFeaturePlugin` was not required to break the `api` dependency. Is that correct? we could still use `PrivacyFeatureName` on that interface. Any reason why we moved to `String`?

marcosholgado: `PrivacyFeaturePlugin` is defined in the `privacy-config-api` module so it cannot have `FeatureName` because that was part of the `feature-toggles-api` module.

cmonfortep: Yes, I'm sorry I see my comment was confusing. I wanted to ask if we could have moved to `PrivacyFeatureName` instead of `String` when replacing `FeatureName`

marcosholgado: The problem with using `PrivacyFeatureName` in the `PrivacyFeaturePlugin` would be the same. We have modules (i.e. vpn and soon ad_click) that contribute to the plugin but by enforcing to use `PrivacyFeatureName` we would end up with the same api dependency.

aitorvs (approved):

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggle.kt
cmonfortep: I was assuming that we will only "contribute" from those modules (vpn and ad_click), and never expose anything from `api` module.

In any case, it was just a question to see if I was getting the big picture here, no worries. Thanks for your time explaining this in more detail.

marcosholgado: If that was the case then yes, you could use `PrivacyFeatureName` in the plugin and extend in the impl modules to create more specific feature names but we want to expose those features (like `AppFeatureName`) from the different `api` modules.

PR #2110, opened 2022-07-29
Disable save in edit mode when all fields are empty

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202673680180939/f

### Description
- Add logic for hiding the save icon in Edit mode when all fields are set to empty.

### Steps to test this PR

_Remove all fields on edit mode_
- [x] Open a saved credential in credential management screen
- [x] Click edit from the overflow menu
- [x] Verify that the check icon is visible
- [x] Remove one field and verify that the check icon is still visible
- [x] Remove all fields and verify that the check icon disappears.

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillSettingsViewModel.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillSettingsViewModel.kt
@@ -161,6 +161,14 @@ class AutofillSettingsViewModel @Inject constructor(
         addCommand(ShowListMode)
     }
 
+    fun allowSaveInEditMode(saveable: Boolean) {
+        _viewState.value.credentialMode.let { credentialMode ->
+            if (credentialMode is Editing) {
+                _viewState.value = _viewState.value.copy(credentialMode = credentialMode.copy(saveable = saveable))
+            }
+        }
+    }
+
     data class ViewState(
         val autofillEnabled: Boolean = true,
         val logins: List<LoginCredentials> = emptyList(),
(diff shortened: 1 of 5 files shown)


autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/SaveStateWatcher.kt
```diff
@@ -0,0 +1,140 @@
...
+            before: Int,
+            count: Int
+        ) {
+        }
+
+        override fun afterTextChanged(s: Editable?) {
+            _binding?.let {
+                _saveStateUpdater(
+                    !(
+                        it.domainTitleEditText.text.isEmpty() &&
```
CDRussell, line 103: Could we provide a list of EditTexts to watch instead of manually referring to each individually? Thinking if we could define the group once then there is less chance of error-prone changes in future

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillSettingsViewModel.kt
```diff
@@ -161,6 +161,14 @@ class AutofillSettingsViewModel @Inject constructor(
         addCommand(ShowListMode)
     }
 
+    fun allowSaveInEditMode(saveable: Boolean) {
+        if (_viewState.value.credentialMode is Editing) {
```
CDRussell, line 165: Nit: should we introduce a variable for `viewState.value` to remove the 3x repetitions

CDRussell (approved):

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/SaveStateWatcher.kt
karlenDimla: This is actually unused code so I am removing this.

PR #2108, opened 2022-07-28
Improve UI for credential entries in list view

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1202659292537113/f

### Description
Improves the look of the individual entries in the credential management list.

- Styles the title and subtitle to match designs
- Shows the title if it exists, otherwise falls back to showing the domain

### Steps to test this PR

- [x] Save a few autofill entries
- [x] Visit the credential management screen; verify the style of the title and subtitle look ok
- [x] Edit a credential and add a title; return to the listing page
- [x] Verify the title is shown instead of the domain
- [x] Re-edit the credential to remove the title and return to the listing page
- [x] Verify the domain is shown instead of a title

PR #2117, opened 2022-08-01
Improve credential grouping and sorting

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1202698381206771/f 

### Description
Improves the grouping and sorting of credentials in the management list view, taking into consideration accented characters, non standard-English letters, numbers etc... and applies a natural sorting to them.

### Steps to test this PR

- [x] Generate half a dozen different logins on any site
- [x] Edit one to have a title starting with a number; verify it gets its own numerical grouping
- [x] Edit one to have a title starting with a character (e.g, `$`); verify it gets put into the `#` placeholder
- [x] Edit one to have a title starting with an accented character (e.g, `ç`); verify it gets put alongside its base letter (e.g., `c`)
- [x] Edit one to have a title starting with a non-english letter (e.g, `あ`); verify it gets its own grouping

Repeat the above with adding the cases above into the domain instead of title


CDRussell:
* develop
  * **PR #2107** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2107" target="_blank">Graphite</a> 
    * **PR #2108** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2108" target="_blank">Graphite</a> 
      * **PR #2117** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2117" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2117?utm_source=stack-comment).

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialGrouper.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private fun sortGroups(groups: MutableMap<String, MutableList<LoginCredentials>>): SortedMap<String, MutableList<LoginCredentials>> {
+        return groups.toSortedMap(sorter.comparator())
+    }
+
+    private fun buildGroups(unsortedCredentials: List<LoginCredentials>): MutableMap<String, MutableList<LoginCredentials>> {
+        val groups = mutableMapOf<String, MutableList<LoginCredentials>>()
+        for (credential in unsortedCredentials) {
+            val initial = initialExtractor.extractInitial(credential).toString() ?: continue
```
karlenDimla, line 56: Should `extractInitial` return a `String?` instead? I see that in `CredentialInitialExtractor` `Normalizer.normalize(rawInitial, Normalizer.Form.NFKD).firstOrNull().toString()` which means it's possible to be `null`?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialGrouper.kt
```diff
@@ -0,0 +1,66 @@
...
+        return flattenedList
+    }
+
+    private fun sortGroups(groups: MutableMap<String, MutableList<LoginCredentials>>): SortedMap<String, MutableList<LoginCredentials>> {
+        return groups.toSortedMap(sorter.comparator())
+    }
+
+    private fun buildGroups(unsortedCredentials: List<LoginCredentials>): MutableMap<String, MutableList<LoginCredentials>> {
+        val groups = mutableMapOf<String, MutableList<LoginCredentials>>()
+        for (credential in unsortedCredentials) {
```
karlenDimla, line 55: Nit: can be `unsortedCredentials.forEach {}`

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialGrouper.kt
```diff
@@ -0,0 +1,66 @@
...
+            val initial = initialExtractor.extractInitial(credential).toString() ?: continue
+            val list = groups.getOrPut(initial) { mutableListOf() }
+            list.add(credential)
+        }
+        return groups
+    }
+
+    private fun shouldAddNewGroup(groups: List<ListItem>, initial: String): Boolean {
```
karlenDimla, line 63: This seem to be unused.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialInitialExtractor.kt
```diff
@@ -14,45 +14,61 @@
...
+            else -> {
+                Timber.v("Rejecting type %d for %s", type, rawInitial)
+                INITIAL_CHAR_FOR_NON_LETTERS
+            }
         }
+    }
 
-        return initial.uppercaseChar()
+    private fun getCharacterType(rawInitial: String): Byte {
+        val initial = rawInitial.firstOrNull() ?: return UNASSIGNED
```
karlenDimla, line 59: This will never be null right? why can't we just use first()?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialInitialExtractor.kt
```diff
@@ -14,45 +14,61 @@
...
 
-        if (initial == null || !characterValidator.isLetter(initial)) {
-            return INITIAL_CHAR_FOR_NON_LETTERS
+        return when (val type = getCharacterType(rawInitial)) {
+            LOWERCASE_LETTER,
+            UPPERCASE_LETTER,
+            OTHER_LETTER,
+            DECIMAL_DIGIT_NUMBER
+            -> {
+                Normalizer.normalize(rawInitial, Normalizer.Form.NFKD).firstOrNull().toString()
```
karlenDimla, line 49: Can this be null after you normalize the rawinitla?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialListSorter.kt
```diff
@@ -0,0 +1,91 @@
...
+class CredentialListSorterByTitleAndDomain @Inject constructor(
+    private val domainFormatter: AutofillDomainFormatter
+) : CredentialListSorter {
+
+    override fun sort(credentials: List<LoginCredentials>): List<LoginCredentials> {
+        return credentials.sortedWith(credentialComparator)
+    }
+
+    override fun comparator(): Collator {
+        return buildCollator()
```
karlenDimla, line 41: do we need a new instance of the collator here and the credentialComparator? I am not very familiar with this class but checking if they can't be reused.

karlenDimla (approved):
A few questions but the feature already works as expected.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialGrouper.kt
CDRussell: The function shouldn't return a `String?`, no, or the compiler would flag it. but you did highlight a related problem!

That function returns a non-nullable String (correctly) but the line you highlight would actually return the string literal "null" if the initial extraction returned null, because it would effectively be `null.toString()`. 

I'll add another guard in here which detects that null and returns the placeholder if it happens.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialInitialExtractor.kt
CDRussell: it definitely won't be null and it _shouldn't_ be an empty string but i worry about that being really easy to pass in empty strings; I'd rather add a bit of safety as opposed to a runtime crash if it does happen

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialInitialExtractor.kt
CDRussell: Unsure; docs don't specify and it's better to assume it can be (in some languages, for some characters, in some Android versions etc...)

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/sorting/CredentialListSorter.kt
CDRussell: Not 100% sure but I know it carries some state so happy to give each use case its own instance

PR #2119, opened 2022-08-01
Autofill autoprompt + pixel

GioSensation:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1202778915697069/f

### Description
This fires a pixel when autofill autoprompt is dismissed. **Note: This also adds to _.gitignore_ the source files for autofill and untracks them from the repo. This will improve the diffs for future autofill updates.**

### Steps to test this PR

_Pixel fired_
1. Visit a login page where you have credentials. For example, https://privacy-test-pages.glitch.me/autofill/autoprompt/1-standard-login-form.html (save credentials for this page if you haven't yet).
2. Provided the form is front and center and not covered on pageload you should be prompted to autofill your credentials automatically (this is the case on the page linked above).
3. Dismiss the prompt and look at the Xcode console. You should see a pixel called `m_autofill_logins_autoprompt_dismissed`.

_No pixel_
1. Visit a login page where you have credentials but the form does not trigger autoprompt. For example, https://privacy-test-pages.glitch.me/autofill/autoprompt/2-form-in-modal.html.
2. Dismiss the prompt and look at the Xcode console. You should not see a pixel called `m_autofill_logins_autoprompt_dismissed`.


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -415,8 +415,8 @@ class BrowserTabFragment :
     }
 
     private val autofillCallback = object : Callback {
-        override suspend fun onCredentialsAvailableToInject(credentials: List<LoginCredentials>) {
-            showAutofillDialogChooseCredentials(credentials)
+        override suspend fun onCredentialsAvailableToInject(credentials: List<LoginCredentials>, isAutoprompt: Boolean) {
+            showAutofillDialogChooseCredentials(credentials, isAutoprompt)
```
GioSensation, line 419: I feel like I'm passing this prop around a bit too much, but couldn't find a better way.

autofill/autofill-impl/build.gradle
```diff
@@ -33,6 +33,9 @@ dependencies {
     implementation project(path: ':device-auth-api')
     implementation project(path: ':browser-api')
     implementation project(path: ':autofill-store')
+    implementation project(path: ':statistics')
+    implementation project(path: ':app')
+    implementation project(path: ':app')
```
GioSensation, line 38: The IDE added these during my attempts at importing the `AppPixelName` stuff into my file. Not sure what's going on and if they are needed.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/selecting/AutofillSelectCredentialsDialogFragment.kt
```diff
@@ -28,6 +28,8 @@ import androidx.lifecycle.lifecycleScope
 import com.duckduckgo.anvil.annotations.InjectWith
 import com.duckduckgo.app.browser.favicon.FaviconManager
 import com.duckduckgo.app.global.extractDomain
+import com.duckduckgo.app.statistics.pixels.Pixel
+import com.duckduckgo.app.pixels.AppPixelName
```
GioSensation, line 32: This import is not working. Not sure how I should do this.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/selecting/AutofillSelectCredentialsDialogFragment.kt
```diff
@@ -108,6 +112,7 @@ class AutofillSelectCredentialsDialogFragment : BottomSheetDialogFragment(), Cre
             it.putBoolean(CredentialAutofillPickerDialog.KEY_CANCELLED, true)
             it.putString(CredentialAutofillPickerDialog.KEY_URL, getOriginalUrl())
         }
+        pixel.fire(AppPixelName.AUTOFILL_LOGINS_AUTOPROMPT_DISMISSED)
```
GioSensation, line 115: As mentioned, `AppPixelName` is not imported so the code doesn't compile.

app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt
```diff
@@ -244,6 +244,8 @@ enum class AppPixelName(override val pixelName: String) : Pixel.PixelName {
     EMAIL_DID_PRESS_WAITLIST_DIALOG_NOTIFY_ME("email_did_press_waitlist_dialog_notify_me"),
     EMAIL_DID_PRESS_WAITLIST_DIALOG_NO_THANKS("email_did_press_waitlist_dialog_dismiss"),
 
+    AUTOFILL_LOGINS_AUTOPROMPT_DISMISSED("m_autofill_logins_autoprompt_dismissed"),
```
CDRussell, line 247: Would recommend moving this to inside `autofill-impl` module. Define a new enum which implements `Pixel.PixelName` and add your new pixel there

autofill/autofill-impl/build.gradle
CDRussell: Keep `implementation project(path: ':statistics')` but kill the other two; they shouldn't be added

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/selecting/AutofillSelectCredentialsDialogFragment.kt
CDRussell: Replace `import com.duckduckgo.app.pixels.AppPixelName` with your own enum where the pixel is defined (see previous comment)

.gitignore
```diff
@@ -58,3 +58,13 @@ build-cache
...
+/node_modules/@duckduckgo/autofill/.eslintrc
+/node_modules/@duckduckgo/autofill/docs
+/node_modules/@duckduckgo/autofill/Gruntfile.js
+/node_modules/@duckduckgo/autofill/integration-test
+/node_modules/@duckduckgo/autofill/package.json
+/node_modules/@duckduckgo/autofill/packages
+/node_modules/@duckduckgo/autofill/playwright.config.js
+/node_modules/@duckduckgo/autofill/scripts
+/node_modules/@duckduckgo/autofill/src
+/node_modules/@duckduckgo/autofill/swift-package
```
marcosholgado, line 70: Can we also add seedrandom and sjcl here? :)

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -415,8 +416,8 @@ class BrowserTabFragment :
     }
 
     private val autofillCallback = object : Callback {
-        override suspend fun onCredentialsAvailableToInject(credentials: List<LoginCredentials>, isAutoprompt: Boolean) {
-            showAutofillDialogChooseCredentials(credentials, isAutoprompt)
+        override suspend fun onCredentialsAvailableToInject(credentials: List<LoginCredentials>, triggerType: LoginTriggerType) {
+            showAutofillDialogChooseCredentials(credentials, triggerType)
```
marcosholgado, line 420: I'm probably missing context but I don't get why this is needed?

From what I understand this callback is called in the `impl` module and we pass some information here (to the tab fragment) which is now going to call `credentialAutofillDialogFactory.autofillSelectCredentialsDialog` passing the information back `impl` so it can return a `DialogFragment`. 

This roundtrip seems rather unnecessary and is probably easily avoidable if the methods in `CredentialAutofillDialogFactory` returned a `DialogFragment` rather than a specific type that is never needed. In fact, those dialogs implement `DialogFragmentType` which has a method to return a `DialogFragment` and that implementation is a simple `return this`.

karlenDimla: Hmm, I do agree that this can be simplified. Since I think the logic within `viewModel.returnNoCredentialsWithPage `and  onCredentialsAvailableToSave can all be handled within `impl` (`AutofillStoredBackJavascriptInterface`).

Though I think even if we get the `DialogFragment` ,we would need the `BrowserTabFragment.showDialogHidingPrevious()` to show it since we don't have any reference to the `childFragmentManager` within the autofill feature. So maybe this call back could just be simplified to `DialogFragment -> {}`

.gitignore
```diff
@@ -58,3 +58,17 @@ build-cache
 report
 
 .DS_Store
+
+# Ignore everything in node_modules except autofill dist, license and readme
+# Info on the weird syntax here: https://stackoverflow.com/a/54870120/1948947
```
GioSensation, line 63: Weird gitignore syntax, but it does the job. Follow the link for the fun.

GioSensation:
I've removed a whole bunch of other files from the repo and use a smarter gitignore thingy so we're good in the future. The only files not ignored are in the autofill `dist/` folder plus the license and readme. We can circle back with a different autofill export at some point as discussed in Asana, but this achieves 90% of the benefits already.

I've also updated autofill to the right tag 5.0.0.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
marcosholgado: Yup, 100% the dialog will have to be shown by the tab fragment.

GioSensation:
@CDRussell I tried rebasing a couple of times, but I get all sorts of errors. I'm sure you will have a much easier time doing it, so I will defer to you.

CDRussell (approved):
I've fixed the git conflicts and tested; all looking good to me. 

_Waiting for CI checks to finish_

@GioSensation are you wanting me to merge immediately when checks pass, or is there anything else to line up first?

GioSensation:
Merged! Thanks for the support, Craig! 🎉

PR #2124, opened 2022-08-02
Autofill: Fix net.sqlcipher.database.SQLiteException in WebsiteLoginCredentialsDao

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202696696022583/f

### Description
This PR includes:
- Check SecureStorage.canAccessSecureStorage when determining if autofill is available. There's no point in enabling autofill if we can't decrypt/encrypt data correctly.
- Separate autofillEnabled and autofillAvailable since they are 2 separate things and have different side effects to the autofill feature.
- Handle scenarios where encryptedsharedprefs failed to load AND room database creation. We want to avoid encrypting / decrypting the database when the l1key is null. Moreover we don't want to automatically regenerate the L1key and use it to encrypt the database since it is possible that the encryptedsharedprefs failed and just can't access the already existing L1key.

### Steps to test this PR

_Autofill is not available for non-internal test user_
- [x] Use a non-internal build
- [x] Confirm that autofill isn't triggered when logging into any site
- [x] Confirm that autofill isn't available in settings

_Autofill is disabled via toggle_
- [x] Use an internal build and disabled autofill via credential management screen
- [x] Confirm that autofill isn't triggered when logging into any site
- [x] Confirm that autofill IS available in settings

_Autofill is works as expected_
- [x] Use an internal build and enable autofill via credential management screen
- [x] Confirm that autofill save dialog is show when logging in to a website
- [x] Confirm that autofill select dialog is shown logging in to a website
- [x] Confirm that credential management works as expected

_Simulated failure case_
- [x] Use an internal build
- [x] Confirm that autofill is triggered and is present in settings
- [x] Modify RealSecureStorageKeyStore and set encryptedPreferences to null
- [x] Re-run app and confirm that autofill is no longer available both when logging in and in settings
- [x] Verify no crash happens (previous crash point)
- [x] Remove changes in RealSecureStorageKeyStore
- [x] Re-run app and confirm that autofill is back and works as expected again

### DIFF
--- a/secure-storage/secure-storage-impl/src/main/java/com/duckduckgo/securestorage/impl/RealSecureStorage.kt
+++ b/secure-storage/secure-storage-impl/src/main/java/com/duckduckgo/securestorage/impl/RealSecureStorage.kt
@@ -25,75 +25,91 @@ import com.duckduckgo.securestorage.impl.encryption.EncryptionHelper.EncryptedSt
 import com.duckduckgo.securestorage.store.SecureStorageRepository
 import com.duckduckgo.securestorage.store.db.WebsiteLoginCredentialsEntity
 import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
 import javax.inject.Inject
 
 @ContributesBinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
 class RealSecureStorage @Inject constructor(
-    private val secureStorageRepository: SecureStorageRepository,
+    private val secureStorageRepositoryFactory: SecureStorageRepository.Factory,
     private val dispatchers: DispatcherProvider,
     private val l2DataTransformer: L2DataTransformer
 ) : SecureStorage {
 
-    override fun canAccessSecureStorage(): Boolean = l2DataTransformer.canProcessData()
+    private val secureStorageRepository by lazy {
+        secureStorageRepositoryFactory.get()
+    }
+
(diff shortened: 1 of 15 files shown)


karlenDimla:
Consider this scenario:
- User got the `5.130.0` and opened the app for the first time. For some reason, `encryptedsharedpref` **failed**
- This means that an encrypted `secure_storage_database_encrypted.db` whose key is lost forever (since it can't be saved) is saved on the device.
- User gets the hotifx so the crash will not happen for now.
- When we enable Autofill for all users, for this user Autofill will try to call SecureStorage and will get the **same crash again**. Why? Because the user already had a previously encrypted db that we don't have the key for.

This could be a very edge case affecting worse case 700 users (assuming all instances of the crash is unique). They can work around it by uninstalling the app / clearing data.

Because of that, I didn't add any special handling for this case.

secure-storage/secure-storage-impl/src/main/java/com/duckduckgo/securestorage/impl/RealSecureStorage.kt
```diff
@@ -25,75 +25,91 @@ import com.duckduckgo.securestorage.impl.encryption.EncryptionHelper.EncryptedSt
...
             }
-        }
+        } else emptyFlow()
 
     override suspend fun websiteLoginDetailsWithCredentials(): Flow<List<WebsiteLoginDetailsWithCredentials>> =
-        withContext(dispatchers.io()) {
-            secureStorageRepository.websiteLoginCredentials().map { list ->
-                list.map {
-                    it.toCredentials()
+        if (secureStorageRepository != null) {
```
CDRussell, line 95: Could use `secureStorageRepository?.let` instead of the null check then the `!!` usage

CDRussell (approved):

PR #2126, opened 2022-08-03
Support incomplete logins from JS

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202705344705784/f 

### Description
Changes the domain object for credentials coming from JS to support nullable fields.

~Additionally, adds a coroutine handler in the autofill processing to ensure exceptions don't crash the whole app.~

### Steps to test this PR

- [x] Hardcode data passed into `AutofillStoredBackJavascriptInterface.storeFormData(data)` to be missing username and password. Verify no crashes and credentials not saved
- [x] Harcode data to be missing username; verify this can be saved
- [x] Hardcode data to be missing password; verify this can be saved

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/AutofillJavascriptInterface.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/AutofillJavascriptInterface.kt
@@ -36,6 +36,7 @@ import com.duckduckgo.autofill.store.AutofillStore
 import com.duckduckgo.deviceauth.api.DeviceAuthenticator
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesBinding
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -47,10 +48,7 @@ interface AutofillJavascriptInterface {
     @JavascriptInterface
     fun getAutofillData(requestString: String)
 
-    suspend fun getRuntimeConfiguration(
-        rawJs: String,
-        url: String?
-    ): String
+    suspend fun getRuntimeConfiguration(rawJs: String, url: String?): String
 
     fun injectCredentials(credentials: LoginCredentials)
     fun injectNoCredentials()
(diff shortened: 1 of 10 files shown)


CDRussell:
* develop
  * **PR #2126** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2126" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2126?utm_source=stack-comment).

karlenDimla:
A few questions around the behavior:
- It seems like we are allowing to save incomplete data as long as it comes from the JS layer. However, if the user input incomplete data in a form (for example only password), we don't show the autofill save dialogg. Is this expected?
- Why are we just logging the crash? I mean it wouldn't be sent anywhere right? In that case, if something is wrong, we wouldn't know about it, unless we experience it ourselves. Wouldn't it be better to let the app crash instead?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/AutofillJavascriptInterface.kt
```diff
@@ -175,10 +173,17 @@ class AutofillStoredBackJavascriptInterface @Inject constructor(
...
         Timber.i("storeFormData called, credentials provided to be persisted")
 
-        getAutofillDataJob += coroutineScope.launch(dispatcherProvider.default()) {
+        getAutofillDataJob += coroutineScope.launch(dispatcherProvider.default() + handler) {
+
             val currentUrl = currentUrlProvider.currentUrl(webView) ?: return@launch
 
             val request = requestParser.parseStoreFormDataRequest(data).credentials
+
+            if (request.username == null && request.password == null) {
```
karlenDimla, line 182: NIT: Edge case - save dialog is shown when you put space only on both username and password. Looks a bit weird in edit mode.

CDRussell:
>It seems like we are allowing to save incomplete data as long as it comes from the JS layer. However, if the user input incomplete data in a form (for example only password), we don't show the autofill save dialogg. Is this expected?

If the user inputs incomplete data in a form, it'll be up the JS to decide whether that constitutes a login event that can be saved. So if the JS does tell us a credential can be saved, and even if it's partial, we should be able to handle that scenario. Does that fully answer the question?

CDRussell:
>Why are we just logging the crash? I mean it wouldn't be sent anywhere right? In that case, if something is wrong, we wouldn't know about it, unless we experience it ourselves. Wouldn't it be better to let the app crash instead?

Good question @karlenDimla, I've decided to pull this crash handling out of this PR altogether.

karlenDimla (approved):

PR #1844, opened 2022-04-06
Make CookieManager lazy

joshliebe:
Task/Issue URL: https://app.asana.com/0/414730916066338/1202085579672992/f

### Description
This makes the CookieManager dependency lazy (In a similar way to: https://github.com/duckduckgo/Android/pull/1775) which should address some related crashes.

### Steps to test this PR

_Feature 1_
- [x] Verify that the app opens without error.
- [x] Visit a site.
- [x] Use the fire button.
- [x] Open a first party cloaked link on: https://privacy-test-pages.glitch.me/privacy-protections/amp/

PR #2131, opened 2022-08-03
Add ServiceWorkerLifecycleObserver to BrowserActivity lifecycle instead of Application

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202712032214375/f

### Description
Another attempt to fix the crashes / anrs related to ServiceWorkerClient. Instead of setting the ServiceWorkerClient on application onStart() which might be too early in the process, we are delaying it to BrowserActivity's onCreate just before any webview is rendered.

### Steps to test this PR

- [x] Smoke test app
- [x] Load website with videos
- [x] Go to `http://privacy-test-pages.glitch.me/privacy-protections/request-blocking/` and make sure the service worker request is blocked (colour is not green)

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/serviceworker/ServiceWorkerLifecycleObserver.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/serviceworker/ServiceWorkerLifecycleObserver.kt
@@ -23,18 +23,15 @@ import androidx.webkit.WebViewFeature
 import androidx.webkit.WebViewFeature.SERVICE_WORKER_BASIC_USAGE
 import com.duckduckgo.app.browser.RequestInterceptor
 import com.duckduckgo.app.global.exception.UncaughtExceptionRepository
-import com.duckduckgo.di.scopes.AppScope
 import timber.log.Timber
 import javax.inject.Inject
-import dagger.SingleInstanceIn
 
-@SingleInstanceIn(AppScope::class)
 class ServiceWorkerLifecycleObserver @Inject constructor(
     private val requestInterceptor: RequestInterceptor,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/serviceworker/ServiceWorkerLifecycleObserver.kt
```diff
@@ -23,18 +23,15 @@ import androidx.webkit.WebViewFeature
...
 
-@SingleInstanceIn(AppScope::class)
 class ServiceWorkerLifecycleObserver @Inject constructor(
     private val requestInterceptor: RequestInterceptor,
     private val uncaughtExceptionRepository: UncaughtExceptionRepository,
 ) : DefaultLifecycleObserver {
-    override fun onStart(owner: LifecycleOwner) {
-        super.onStart(owner)
+    override fun onCreate(owner: LifecycleOwner) {
+        super.onCreate(owner)
```
joshliebe, line 34: Instead of changing the lifecycle, would it be possible to make the `serviceWorkerLifecycleObserver` "lazy" so that `getInstance()` is only called when that class is accessed for the first time?

Sort of like: https://github.com/duckduckgo/Android/pull/1844.

karlenDimla: Discussed in MM. It wouldn't really make sense to lazily instantiate `serviceWorkerLifecycleObserver` itself since it doesn't hold any internal state and the creation of `ServiceWorkerControllerCompat` (crash point) happens on lifecycle `onCreate() `anyway.

joshliebe (approved):
LGTM!

PR #2138, opened 2022-08-04
Expose autofill dialogs as DialogFragment to simplify API

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202722378440328/f 

### Description
Simplification of the way we create and show the DialogFragments in autofill.

### Steps to test this PR

(Make sure you build using `internal` variant)

- [ ] Visit a site with a login (e.g., trello.com/login)
- [ ] Attempt a login; verify the save dialog still shows and works correctly
- [ ] Save the login. Refresh the page and verify you are prompted to use the saved login when tapping on the email field
- [ ] Choose to autofill. Change the password and hit log in button; Verify the update password dialog still shows

### DIFF
--- a/autofill/autofill-api/src/main/java/com/duckduckgo/autofill/AutofillCredentialDialogs.kt
+++ b/autofill/autofill-api/src/main/java/com/duckduckgo/autofill/AutofillCredentialDialogs.kt
@@ -22,7 +22,7 @@ import com.duckduckgo.autofill.domain.app.LoginCredentials
 /**
  * Dialog which can be shown when user is required to select which saved credential to autofill
  */
-interface CredentialAutofillPickerDialog : DialogFragmentType {
+interface CredentialAutofillPickerDialog {
 
     companion object {
         const val TAG = "CredentialAutofillPickerDialog"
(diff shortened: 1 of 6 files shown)


autofill/autofill-api/src/main/java/com/duckduckgo/autofill/AutofillCredentialDialogs.kt
```diff
@@ -22,7 +22,7 @@ import com.duckduckgo.autofill.domain.app.LoginCredentials
 /**
  * Dialog which can be shown when user is required to select which saved credential to autofill
  */
-interface CredentialAutofillPickerDialog : DialogFragmentType {
+interface CredentialAutofillPickerDialog {
```
marcosholgado, line 25: Not for this PR but maybe for a future one. Looking at these interfaces I think they can probably be deleted altogether and move the companions inside the implementations. The exception is the usage in `AutofillCredentialsSelectionResultHandler` which lives in the `app` module and could also be moved to `impl` to avoid a few roundtrips between the module and the tab fragment.

marcosholgado (approved):
LGTM!

autofill/autofill-api/src/main/java/com/duckduckgo/autofill/AutofillCredentialDialogs.kt
CDRussell: Agree not for this PR, and agree this could be improved. Would be good to define a guideline around this scenario first, then come back and make the changes based on that guideline.

CDRussell: Will track and discuss more here: [Autofill: Move more of autofill dialogs into autofill-impl module](https://app.asana.com/0/1202727816093912/1202727816093912/f)

PR #2154, opened 2022-08-10
Update unprotected sites copy

joshliebe:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202605670624090/f

### Description
Update unprotected sites header copy from "Fireproof sites" to "Unprotected sites".

### Steps to test this PR

- [x] Go to Settings → Unprotected Sites
- [x] Verify that the title says "Unprotected sites" instead of "Fireproof sites"  

### UI changes
| Before  | After |
| ------ | ----- |
before|after


marcosholgado (changes requested):
Couple of problems here:

1. This change now breaks the fireproof section (settings -> fireproof sites). The actual fix should be changing the string in the `WebsiteAdapter` to point to `settingsPrivacyProtectionWhitelist`.
2. Doesn't matter because of the previous point but something to consider whenever changing a string that's already translated is that the translations will be broken so we have to be careful with that.

marcosholgado (approved):
LGTM!

PR #2162, opened 2022-08-11
Fix tracker detector crash

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202768642236786/f

### Description
This PR allows the tracker detector to return null which we then match to Client.Result(matches = false, isATracker = false)

### Steps to test this PR
- [x] All tests should pass 
- [x] Privacy tests should pass (done in https://github.com/duckduckgo/Android/actions/runs/2837958897)
- [x] Smoke test bat.js


anikiki (approved):
LGTM :tada: Works as expected!

PR #2172, opened 2022-08-15
Fix broken site reports

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1202794100930314

### Description
After changing the trackers algorithm we broke broken sites a bit. We send the full list of domains that we intercept instead of only the ones that we blocked.

### Steps to test this PR

- [x] Install the app
- [x] Go to `play.google.com`
- [x] Go to the privacy dashboard
- [x] You should see `No tracking requests blocked`
- [x] Report the site as broken using the button on that scree
- [x] In the logcat filter by `Pixel sent: epbf`
- [x] The blockedTrackers parameter should be empty
- [x] Do the same but this time from `cnn.com`
- [x] The privacy dashboard should say this time `Requests blocked from loading`
- [x] The blockedTrackers parameter should not be empty


aitorvs (approved):
Works as expected!

PR #2178, opened 2022-08-15
Add Dax dialog with lottie animation to autoconsent

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202800206441425/f

### Description
This PR adds a new Dax dialog including a lottie animation

### Steps to test this PR

- [x] Go to `cnn.com`
- [x] You should see the onboarding dialog saying that we have blocked trackers
- [x] Tap on `High Five!`
- [x] Refresh the page
- [x] You should now see the new autoconsent dialog with an animation inside of it.
- [x] Tap on `Manage cookie Pop-ups`
- [x] The cookie popup should be handled automatically and eventually disappear.
- [x] Go to `gitlab.com`
- [x] Wait for a few seconds and eventually you should see a toast saying that cookies where handled.
- [x] Go to SERP and you should see an onboarding dialog.
- [x] Clear data & cache
- [x] Go to `cnn.com` and dimiss the onboarding dialog
- [x] Refresh to trigger the autoconsent popup
- [x] This time tap on `No Thanks`
- [x] Cookie popup from `cnn.com` should still be visible.
- [x] Go to `gitlab.com` and the cookie popup should be visible.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202800206441425

PR #2155, opened 2022-08-10
Support other messages

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1202768897662116

### Description
This PRs adds support for OptOut, selfTest and autoconsentDone

### Steps to test this PR

- [x] Install the app, go to SERP and get rid of the Dax dialogs by clicking hide and hide tips forever.
- [x] Go to `https://www.lufthansa.com/de/en/homepage`
- [x] New Dax dialog should show up and you should see the cookie popup behind it.
- [x] Click Manage Cookie Pop-ups
- [x] Dax dialog and cookie pop-up should disappear
- [x] You should see a toast with a random message
- [x] Refresh the page and you should not see anything related with cookies (Dax dialog, pop-up or toast)
- [x] Go to `http://privacy-test-pages.glitch.me/features/autoconsent/`
- [x] You should see a toast with a random message and the button in the site should say `I was clicked`
- [x] Refreshing the site should trigger the same behaviour, toast should show and button should way `I was clicked`

PR #2129, opened 2022-08-03
Add store to CMP

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1202705137400821

### Description
This PR adds a store module to handle a couple of basic autoconsent settings and integrates them in the basic messages that we already had.
### Steps to test this PR
- [x] Install the app and go to SERP/
- [x] In the dax dialog click Hide an. Hide tips forever. 

_First popup handled_
- [x] Navigate to `http://privacy-test-pages.glitch.me/features/autoconsent/`.
- [x] You should see a new Dax Dialog and in the website a button that says `Reject all`
- [x] Click on `Manage Cookie Pop-ups`.
- [x] Dax dialog should disappear and the button should say `I was clicked`.
- [x] Refresh the page.
- [x] Button should say `I was clicked` but no Dax dialog should be shown again.
- [x] Fresh install the app
- [x] Navigate to `http://privacy-test-pages.glitch.me/features/autoconsent/`.
- [x] You should see a new Dax Dialog and in the website a button that says `Reject all`
- [x] Click on `No Thanks`.
- [x] Dax dialog should disappear and the button should say `Reject all`.
- [x] Refresh the page.
- [x] Button should say `Reject all` but no Dax dialog should be shown again.

PR #2181, opened 2022-08-16
Ignore top cmps

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1202800206441459

### Description
In this PR we ignore top cmps that we cannot handle.

### Steps to test this PR

- [x] Install the app from this branch
- [x] Go to `independent.co.uk`
- [x] You should see an onboarding dialog, close it
- [x] Refresh the page
- [x] You should not see the autoconsent dialog
- [x] Go to `cnn.com`
- [x] You should see the autoconsent dialog


marcosholgado:
* develop
  * **PR #2128** <a href="https://app.graphite.dev/github/pr/duckduckgo/android/2128" target="_blank">Graphite</a> 
    * **PR #2129** <a href="https://app.graphite.dev/github/pr/duckduckgo/android/2129" target="_blank">Graphite</a> 
    * **PR #2155** <a href="https://app.graphite.dev/github/pr/duckduckgo/android/2155" target="_blank">Graphite</a> 
    * **PR #2178** <a href="https://app.graphite.dev/github/pr/duckduckgo/android/2178" target="_blank">Graphite</a> 
    * **PR #2181** <a href="https://app.graphite.dev/github/pr/duckduckgo/android/2181" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/android/2181?utm_source=stack-comment).

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/PopUpFoundMessageHandlerPlugin.kt
```diff
@@ -17,24 +17,53 @@
...
-            if (repository.userSetting) {
-                return
+        try {
+            if (supportedTypes.contains(messageType)) {
+                val message: PopUpFoundMessage = parseMessage(jsonString) ?: return
+                if (repository.userSetting) return
+
+                val url: String = runBlocking(dispatcherProvider.main()) { webView.url.orEmpty() }
+                if (message.cmp.endsWith(IGNORE_CMP_SUFFIX)) return
+                if (message.url != url) return
```
muodov, line 49: You sure we need this check? Aren't we sure at this point that the message originated from `webView`?

marcosholgado: Probably not but I've seen in a few cases that moving quickly from one site with a cmp to another one with a cmp you might get the first popupFound message triggered while you are on the second site. The check is to be sure that both the url from the message and webview are the same before continuing.

For the record, I saw this when I moved the injecting to onPageFinished so I might be overdoing it here and I don't think I've seen that scenario before.

muodov: It should be fine as long as `webView.url` returns the same as `location.href` in JS. If it gets out of sync or differs (e.g. if the #anchor part is missing), we could miss things here

marcosholgado: Yeah I ran a few tests and seemed to work fine but let's remove it just in case, like I said I haven't seen this again since moving the injecting back to onPageStarted so we can discuss if I can recreate again at some point.

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/PopUpFoundMessageHandlerPlugin.kt
```diff
@@ -19,22 +19,44 @@ package com.duckduckgo.autoconsent.impl.handlers
...
+
     override fun process(messageType: String, jsonString: String, webView: WebView, autoconsentCallback: AutoconsentCallback) {
-        if (supportedTypes.contains(messageType)) {
-            if (repository.userSetting) {
-                return
+        try {
+            if (supportedTypes.contains(messageType)) {
+                val message: PopUpFoundMessage = parseMessage(jsonString) ?: return
+                if (repository.userSetting) return
+                if (message.cmp.endsWith(IGNORE_CMP_SUFFIX)) return
```
aitorvs, line 41: Just flagging to be absolutely sure that case is important here and we don't need to ignore it

aitorvs (approved):
Works as expected.

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/PopUpFoundMessageHandlerPlugin.kt
marcosholgado: Good catch! Fixed

PR #2208, opened 2022-08-19
Fix infinite authentication loop on older Android

karlenDimla:
Task/Issue URL: 
https://app.asana.com/0/488551667048375/1202821446406981/f
https://app.asana.com/0/488551667048375/1202828695381297/f

### Description
This PR includes:
- Ensuring that only one device auth flow is launched until it is completed. The issue is originally caused by the auth flow in older android device that launches a new activity. Once the auth is completed, the auth activity is closed and triggers the management screen's `onStart` where we call `launchDeviceAuth`. To fix this, we add a check before launching auth that there is no already ongoing auth. An auth flow is considered `ended` once we have received the result in the AutofillManagementActivity.
- Don't reauth when orientation changes. Used https://developer.android.com/guide/topics/manifest/activity-element#config android:configChanges to prevent the activity from being restarted. We needed to add ScreenSize since orientation alone for some reason doesn't work.

### Steps to test this PR

_Check Autofill management works_
- [x] Launch autofill on an older android device (API 23) with device auth enrolled
- [x] Open autofill management screen
- [x] Complete device auth
- [x] Verify that list view is shown
- [x] Exit Autofill management screen and open it again.
- [x] Cancel device auth
- [x] Confirm that you are returned to browser/settings page.

_Check Autofill management works on newer device_
- [ ] Launch autofill on a newer android device (Android 12) with device auth enrolled
- [ ] Repeat steps above and ensure that everything works as expected.

_Orientation change doesm't reauth_
- [x] Launch autofill management and complete device auth
- [x] Change orientation and confirm that device auth is not triggered
- [x] Confirm that backgrounding app and locking screen launches device auth still.

### DIFF
--- a/autofill/autofill-impl/src/main/AndroidManifest.xml
+++ b/autofill/autofill-impl/src/main/AndroidManifest.xml
@@ -20,6 +20,7 @@
     <application>
         <activity
             android:name="com.duckduckgo.autofill.ui.credential.management.AutofillManagementActivity"
+            android:configChanges="orientation|screenSize"
             android:exported="false" />
     </application>
 
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillManagementActivity.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillManagementActivity.kt
@@ -96,6 +96,7 @@ class AutofillManagementActivity : DuckDuckGoActivity() {
                 } else {
                     finish()
                 }
+                viewModel.onAuthenticationEnded()
             }
         } else {
             viewModel.disabled()
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillSettingsViewModel.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillSettingsViewModel.kt
@@ -127,7 +127,10 @@ class AutofillSettingsViewModel @Inject constructor(
     }
 
     fun launchDeviceAuth() {
-        addCommand(LaunchDeviceAuth)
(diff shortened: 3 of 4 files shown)


autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillManagementActivity.kt
```diff
@@ -96,6 +96,7 @@ class AutofillManagementActivity : DuckDuckGoActivity() {
                 } else {
                     finish()
                 }
+                viewModel.onAuthenticationEnded()
```
karlenDimla, line 99: At this point, the Auth activity has already been closed and onStart() should have been called.

autofill/autofill-impl/src/main/AndroidManifest.xml
```diff
@@ -20,6 +20,7 @@
     <application>
         <activity
             android:name="com.duckduckgo.autofill.ui.credential.management.AutofillManagementActivity"
+            android:configChanges="orientation|screenSize"
```
CDRussell, line 23: should probably copy what we do for `BrowserActivity`?

`android:configChanges="keyboardHidden|orientation|screenSize"`

CDRussell (approved):

autofill/autofill-impl/src/main/AndroidManifest.xml
karlenDimla: Hmm, Didn't seem necessary though. I don't think keyboardHidden is triggering any restart on the Activity atm. I will not add it for now. Maybe add it later if it will seem needed.

CDRussell: IIRC, `keyboardHidden` refers to a hardware keyboard's availability, not the software-based IME. what the best list is, i'm not sure (we're probably also missing `screenLayout` from both for instance).

PR #2209, opened 2022-08-19
Catch throwble if ServiceWorkerControllerCompat init

karlenDimla:
Task/Issue URL: https://app.asana.com/0/0/1202828695381301/f

### Description
- Change try-catch for ServiceWorkerControllerCompat initialization to catch Throwable instead of Exception only.
- So far all Throwable thrown by ServiceWorkerControllerCompat are webview/chromium related which we don't have controll of. Instead of letting the app crash and become unusable, we just hide this specific type of crashes.


marcosholgado (approved):
👍

PR #2251, opened 2022-09-01
Introduce max active tabs in BrowserActivity

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1201078435129729/f

### Description
PR introduces max active size on BrowserActivity.
This will test the hypothesis that allowing to have infinite tabs alive in the browserActivity causes memory and binder exceptions.

To validate the assumption, we are going to keep track of the latest 40 visited tabs in memory, and remove from the fragment manager the rest of them.

### Steps to test this PR

_Test 1_
- [ ] checkout the code and set `MAX_ACTIVE_TABS` in BrowserActivity to `4`
- [ ] install the app
- [ ] go to browser
- [ ] open a tab and visit a site (x4) 
- [ ] now open another tab, and visit a new site
- [ ] go to the first tab you created
- [ ] ensure it's recreated (meaning that was removed from the fragment manager)

_Test 2_
- [ ] checkout the code and set `MAX_ACTIVE_TABS` in BrowserActivity to `4`
- [ ] install the app
- [ ] go to browser
- [ ] open a tab and visit a site (x4) 
- [ ] visit the oldest one
- [ ] now open another tab, and visit a new site
- [ ] go to the first tab you created
- [ ] ensure it's NOT recreated (because you visited it recently)
- [ ] visit the second tab you created
- [ ] ensure it's recreated (now this was the oldest one)

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -78,6 +78,7 @@ import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
 import timber.log.Timber
 import javax.inject.Inject
+import kotlin.collections.ArrayList
 
 // open class so that we can test BrowserApplicationStateInfo
 @InjectWith(ActivityScope::class)
@@ -116,6 +117,8 @@ open class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope()
     @Inject
     @AppCoroutineScope
     lateinit var appCoroutineScope: CoroutineScope
+
+    private val lastActiveTabs = TabList()
 
     private var currentTab: BrowserTabFragment? = null
 
@@ -235,6 +238,8 @@ open class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope()
 
         if (tab.tabId == currentTab?.tabId) return
 
+        lastActiveTabs.add(tab.tabId)
+
         val fragment = supportFragmentManager.findFragmentByTag(tab.tabId) as? BrowserTabFragment
         if (fragment == null) {
             openNewTab(tab.tabId, tab.url, tab.skipHome)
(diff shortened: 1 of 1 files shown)


CDRussell:
Functionally all working as expected, however, I will raise concern around _when_ we are attempting to remove older tabs.

You have it firing from `viewModel.tabs.observe`, which is surprisingly noisy and called multiple times when a page is loading. (i added a log statement to confirm)
Screenshot 2022-09-02 at 11 48 03

That already seems non-ideal that it's firing so much, and i worry your code will make it worse. Is there another place you can do the cleanup, like when explicitly adding a new tab perhaps?

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -569,3 +594,13 @@ open class BrowserActivity : DuckDuckGoActivity(), CoroutineScope by MainScope()
         val newInstanceState: Bundle?
     )
 }
+
+// Temporary class to keep track of latest visited tabs, keeping unique ids.
```
CDRussell, line 598: "Temporary" makes me think this is temporary code that will be changed soon. Could we update the comment to indicate it's the data that is only held for a short while? (e.g., "used to keep track of the latest visited tabs in memory")

cmonfortep: Yes, the comment refers that is a class that will be changed soon, not related with the lifecycle of the data stored. The plan is: 1) validate assumption around transaction too large exception  2) put some more love into the code. If 1 fails, we will remove it. If it works, we can then extract this into a wrapper that handles fragmentManager + memory list.

CDRussell (approved):
Discussed the concern around the extraneous invocations in a call. Summary: 

- this change is considered somewhat temporary; we'll likely revisit based on the impact on out of memory exceptions etc...
- even if triggered frequently, the actual work it's doing should be small as it won't be operating on a big list of tabs (usually 0 or 1)

cmonfortep:
@CDRussell, I will address your comment in https://github.com/duckduckgo/Android/pull/2257

Will wait for that PR to be reviewed before merging this one.

PR #2259, opened 2022-09-06
Extract configuration related logic out from BrowserAutofill

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202922320754385/f 

### Description
Prior to this change, InlineBrowserAutofill does 2 things: 1.) configures the webview for the specific site being loaded and 2.) handles injection of credentials. Because of this 2 purposes, BrowserAutofill is injected in 2 places creating 2 instances for every tab. The problem with this is that AutofillStoredBackJavascriptInterface together with majority of its dependencies are duplicated when we only need them for purpose 2.

To solve it in this PR, we separate the 2 purposes into 2 separate group of classes. Thus we only inject exactly what we need in the BrowserWebViewClient.

### Steps to test this PR
- [ ] Smoke test Autofill
- [ ] Smoke test Email


karlenDimla:
* develop
  * **PR #2259** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2259" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2259?utm_source=stack-comment).

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/configuration/AutofillRuntimeConfigProvider.kt
```diff
@@ -0,0 +1,86 @@
...
+        val emailAvailable = determineIfEmailAvailable()
+
+        val json = runtimeConfigurationWriter.generateResponseGetAvailableInputTypes(credentialsAvailable, emailAvailable).also {
+            Timber.v("availableInputTypes for %s: \n%s", url, it)
+        }
+        return "availableInputTypes = $json"
+    }
+
+    // in the future, we'll also tie this into feature toggles and remote config
+    private fun determineIfAutofillEnabled(): Boolean =
```
marcosholgado, line 73: FYI we are already parsing autofill from the remote config so you should be able to use `FeatureToggle` and `Autofill` APIs here.

CDRussell: Thanks @marcosholgado. It's a good FYI. shouldn't be done in this PR to avoid moving lots around AND changing logic in one go. For a follow-up PR 👍

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/configuration/RuntimeConfigurationWriter.kt
```diff
@@ -0,0 +1,107 @@
...
+        emailAvailable: Boolean
+    ): String {
+        val availableInputTypes = AvailableInputSuccessResponse(credentialsAvailable, emailAvailable)
+        return availableInputTypesAdapter.toJson(availableInputTypes)
+    }
+
+    /*
+    * hardcoded for now, but eventually will be a dump of the most up-to-date privacy remote config, untouched by us
+    */
+    override fun generateContentScope(): String {
```
marcosholgado, line 56: Already synced on MM about this but for future reference. Let's keep an eye on this because I don't think we need to dump the whole remote config, only the specific autofill part. I doubt gpc or fingerprinting has any use for autofill. We also need to know if autofill uses exceptions and unprotected temporary.

CDRussell (approved):

PR #2261, opened 2022-09-06
Fix disk io on main thread, BrowserTabViewModel.initializeViewStates

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1202908047221081/f

### Description
The current implementation of `BrowserTabViewModel.initializeViewStates` can be very slow, as it involves reading persisted data (on the main thread) to populate the view states. This can be pretty taxing on the main thread (measured at blocking UI thread for over 300ms on one device). Most view states don't require providing dynamic init values, leaving only a few offenders:

- autofill
- voice search capabilities
- add to home capabilities
- accessibility preferences

This PR makes the logic to initialise the `BrowserTabViewModel` view states happen in two parts:
- a default init which is instant as it does no disk IO
- async init to update view states which require fetching data from disk

In practice both parts should initialise so quickly that the user doesn't notice any difference, but we apply sensible defaults to the view states in case fetching the data is noticeably delayed.

#### Accessibility 
- [ ] Visit settings->accessibility, and change the font size to largest and to allow force zooming
- [ ] visit twitter.com (text size should be large and force zooming supported)
- [ ] Relaunch the app (so that `BrowserTabViewModel` re initialised and twitter page reloaded)
- [ ] Verify font size still large, and force zooming still supported

#### Autofill
ℹ️ Autofill being available or not can be tested by using different types (enabled on `internal`; disabled on `play`) or by hacking `val showAutofill = autofillStore.autofillAvailable` to return a hardcoded value

- [ ] Test with autofill enabled; verify autofill is in the overflow menu
- [ ] Test with autofill disabled; verify autofill is not in the overflow menu

#### Add to home
ℹ️ Can adjust whether add to home supported or not by manually changing `AddToHomeSystemCapabilityDetector.isAddToHomeSupported()`

- [ ] Test with add to home returning true; verify add to home is an option in the overflow menu
- [ ] Test with add to home returning false; verify add to home is not an option in the overflow menu

#### Voice search 
ℹ️ Can adjust whether voice search should be shown by hardcoding response to `RealVoiceSearchAvailability.shouldShowVoiceSearch()`

- [ ] Test with should show voice search returning false; verify voice search button isn't shown in url bar
- [ ] Test with should show voice search returning true; verify voice search button is shown in url bar

#### Testing an unexpectedly large delay in the second init
- [ ] Hardcode a delay in the init function. e.g., inside `initializeViewStatesFromPersistedData` add `delay(5_000)`
- [ ] Launch the app and verify there is a sensible UI while waiting for the synthetic delay period to pass

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -36,6 +36,7 @@ import androidx.lifecycle.Observer
 import com.duckduckgo.adclick.api.AdClickManager
 import com.duckduckgo.anvil.annotations.ContributesViewModel
 import com.duckduckgo.app.accessibility.data.AccessibilitySettingsDataStore
+import com.duckduckgo.app.accessibility.data.AccessibilitySettingsSharedPreferences.Companion.FONT_SIZE_DEFAULT
 import com.duckduckgo.app.autocomplete.api.AutoComplete
 import com.duckduckgo.app.autocomplete.api.AutoComplete.AutoCompleteResult
 import com.duckduckgo.app.autocomplete.api.AutoComplete.AutoCompleteSuggestion
@@ -279,9 +280,9 @@ class BrowserTabViewModel @Inject constructor(
     )
 
     data class AccessibilityViewState(
-        val fontSize: Float,
-        val forceZoom: Boolean,
-        val refreshWebView: Boolean
+        val fontSize: Float = FONT_SIZE_DEFAULT,
+        val forceZoom: Boolean = false,
+        val refreshWebView: Boolean = false
     )
 
     data class FindInPageViewState(
@@ -2142,24 +2143,47 @@ class BrowserTabViewModel @Inject constructor(
     }
 
     private fun initializeViewStates() {
+        initializeDefaultViewStates()
+        viewModelScope.launch {
+            initializeViewStatesFromPersistedData()
+        }
+    }
+
+    private fun initializeDefaultViewStates() {
         globalLayoutState.value = Browser()
-        browserViewState.value = BrowserViewState().copy(
-            addToHomeVisible = addToHomeCapabilityDetector.isAddToHomeSupported(),
-            showAutofill = autofillStore.autofillAvailable
-        )
+        browserViewState.value = BrowserViewState()
         loadingViewState.value = LoadingViewState()
         autoCompleteViewState.value = AutoCompleteViewState()
-        omnibarViewState.value = OmnibarViewState(
-            showVoiceSearch = voiceSearchAvailability.shouldShowVoiceSearch()
-        )
+        omnibarViewState.value = OmnibarViewState()
         findInPageViewState.value = FindInPageViewState()
         ctaViewState.value = CtaViewState()
         privacyGradeViewState.value = PrivacyGradeViewState()
-        accessibilityViewState.value = AccessibilityViewState(
-            fontSize = accessibilitySettingsDataStore.fontSize,
-            forceZoom = accessibilitySettingsDataStore.forceZoom,
-            refreshWebView = false
-        )
+        accessibilityViewState.value = AccessibilityViewState()
+    }
+
+    private suspend fun initializeViewStatesFromPersistedData() {
+        withContext(dispatchers.io()) {
+            val addToHomeSupported = addToHomeCapabilityDetector.isAddToHomeSupported()
+            val showAutofill = autofillStore.autofillAvailable
+            val showVoiceSearch = voiceSearchAvailability.shouldShowVoiceSearch()
+            val fontSize = accessibilitySettingsDataStore.fontSize
+            val forceZoom = accessibilitySettingsDataStore.forceZoom
+
+            withContext(dispatchers.main()) {
+                browserViewState.value = currentBrowserViewState().copy(
+                    addToHomeVisible = addToHomeSupported,
+                    showAutofill = showAutofill
+                )
+                omnibarViewState.value = currentOmnibarViewState().copy(
+                    showVoiceSearch = showVoiceSearch
+                )
+                accessibilityViewState.value = currentAccessibilityViewState().copy(
+                    fontSize = fontSize,
+                    forceZoom = forceZoom,
+                    refreshWebView = false
+                )
+            }
+        }
     }
 
     fun onShareSelected() {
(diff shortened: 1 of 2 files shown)


CDRussell:
* develop
  * **PR #2261** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2261" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2261?utm_source=stack-comment).

CDRussell:
cc @karlenDimla for touching voice search view state initialisation

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2142,24 +2143,47 @@ class BrowserTabViewModel @Inject constructor(
...
         privacyGradeViewState.value = PrivacyGradeViewState()
-        accessibilityViewState.value = AccessibilityViewState(
-            fontSize = accessibilitySettingsDataStore.fontSize,
-            forceZoom = accessibilitySettingsDataStore.forceZoom,
-            refreshWebView = false
-        )
+        accessibilityViewState.value = AccessibilityViewState()
+    }
+
+    private suspend fun initializeViewStatesFromPersistedData() {
```
CDRussell, line 2164: `initializeViewStatesFromPersistedData` isn't quite technically correct, as it's more about looking up values that might come from anywhere and not just persisted.. open to better naming suggestions

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2142,24 +2143,47 @@ class BrowserTabViewModel @Inject constructor(
     }
 
     private fun initializeViewStates() {
+        initializeDefaultViewStates()
+        viewModelScope.launch {
```
malmstein, line 2147: why not launching the coroutine inside the `initializeViewStatesFromPersistedData()` function?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2142,24 +2143,47 @@ class BrowserTabViewModel @Inject constructor(
...
-        )
+        accessibilityViewState.value = AccessibilityViewState()
+    }
+
+    private suspend fun initializeViewStatesFromPersistedData() {
+        withContext(dispatchers.io()) {
+            val addToHomeSupported = addToHomeCapabilityDetector.isAddToHomeSupported()
+            val showAutofill = autofillStore.autofillAvailable
+            val showVoiceSearch = voiceSearchAvailability.shouldShowVoiceSearch()
+            val fontSize = accessibilitySettingsDataStore.fontSize
```
malmstein, line 2169: I'm confused with this. If the `AccessibilitySettingsDataStore ` already has a default value that we are observing a few [lines below](https://github.com/duckduckgo/Android/blob/57d3b50bb746c5d4ec04a9637655bd1a0ba31dc5/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt#L598) to, why do you need to fetch this values here?

malmstein (commented):
I'm not the main reviewer so I'd leave this @cmonfortep, but I'm confused about what are we trying to do here. 

Can you elaborate on the _"This can be pretty taxing on the main thread"_ statement? What's the difference now and then?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2142,24 +2143,47 @@ class BrowserTabViewModel @Inject constructor(
...
+
+            withContext(dispatchers.main()) {
+                browserViewState.value = currentBrowserViewState().copy(
+                    addToHomeVisible = addToHomeSupported,
+                    showAutofill = showAutofill
+                )
+                omnibarViewState.value = currentOmnibarViewState().copy(
+                    showVoiceSearch = showVoiceSearch
+                )
+                accessibilityViewState.value = currentAccessibilityViewState().copy(
```
cmonfortep, line 2180: I don't think we need to initialize the `ViewState`. If we have a default value for the `ViewState`, initialization can happen inside the observer we create in `observeAccessibilitySettings`.

cmonfortep (approved):
LGTM, left a comment about some code I think can be removed. 

I've tested the scenarios and they work as described.

One note about the changes. I'm open to testing this and validating if we see any positive impact, even though I'm a bit skeptical that we are going to see something.

I'm worried to rely too much launching async calls and create new edge cases. There are other strategies that could be safer than launching coroutines from every consumer. For example, we have repositories that initialize their properties in memory, and consumers can get the values from the main thread (e.g: privacy config repositories, `exceptions` can be consumed from main thread).

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
CDRussell: Agreed.

I was trying to best replicate what was there before, including that initial initialization before `observeAccessibilitySettings`, but agree that should be unnecessary. Most likely unnecessary before too. Will remove.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
CDRussell: Answered: https://github.com/duckduckgo/Android/pull/2261/files#r972995609

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
CDRussell: No strong reason either way. One weak reason being that it makes it more obvious when looking at `initializeViewStates()` that the second operation is async.

CDRussell:
> LGTM, left a comment about some code I think can be removed.
> 
> I've tested the scenarios and they work as described.
> 
> One note about the changes. I'm open to testing this and validating if we see any positive impact, even though I'm a bit skeptical that we are going to see something.
> 
> I'm worried to rely too much launching async calls and create new edge cases. There are other strategies that could be safer than launching coroutines from every consumer. For example, we have repositories that initialize their properties in memory, and consumers can get the values from the main thread (e.g: privacy config repositories, `exceptions` can be consumed from main thread).


>There are other strategies that could be safer than launching coroutines from every consumer.

This PR is a workaround to view states being initialised in a way which blocks the main thread while reading default values from disk; in an ideal world we wouldn't need this because each data source would better handle/document data being async or not. So I agree, however that's a larger conversation around fixing the issues at the source (e.g., pre-loading in memory). I'm happy to make incremental changes in this area before it's _perfect_.

PR #2276, opened 2022-09-14
bump dashboard

shakyShane:
just bumping the dashboard to the latest version on feature branch


cmonfortep (approved):
LGTM

CI checks are not passing but is not related to any change here. It will be fixed on my feature branch.

PR #2311, opened 2022-09-23
Android: Debounce filtering for bookmarks

anikiki:
Task/Issue URL: https://app.asana.com/0/1200905986587319/1203036016848820/f

### Description
Added debounce for bookmarks filtering.

### Steps to test this PR

- [x] Import bookmarks / favorites / folders (there is a file attached to the Asana task if you don't have one).
- [x] Perform searches on bookmarks / favorites / folders.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksActivity.kt
@@ -23,6 +23,7 @@ import android.view.Menu
 import android.view.MenuItem
 import androidx.appcompat.app.AlertDialog
 import androidx.core.view.isVisible
+import androidx.lifecycle.coroutineScope
 import androidx.recyclerview.widget.ConcatAdapter
 import com.duckduckgo.anvil.annotations.InjectWith
 import com.duckduckgo.app.bookmarks.model.BookmarkFolder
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksActivity.kt
```diff
@@ -274,7 +280,12 @@ class BookmarksActivity : DuckDuckGoActivity() {
         searchBar.onAction {
             when (it) {
                 is SearchBar.Action.PerformUpAction -> hideSearchBar()
-                is SearchBar.Action.PerformSearch -> listener.onQueryTextChange(it.searchText)
+                is SearchBar.Action.PerformSearch -> {
+                    searchJob += lifecycle.coroutineScope.launch {
```
malmstein, line 284: Have you considered debouncing in the listener itself? It has a reference to the `viewModel`, so you could use `viewModel.viewModelScope.launch` and avoid canceling the job in the `Activity`.

It would also remove the `DEBOUNCE_PERIOD ` from here

malmstein (approved):
Just a suggestion about the place of the debounce, not feeling too strong about it. Works as expected, good job @anikiki!

app/src/main/java/com/duckduckgo/app/bookmarks/ui/BookmarksActivity.kt
anikiki: Good suggestion, thanks! Updated :+1:

malmstein (approved):
nice!

PR #2346, opened 2022-10-02
Site Permissions new tests

nalcalag:
Task/Issue URL: https://app.asana.com/0/0/1203086312631189/f

### Description
Add tests for Site Permissions screens and site-permissions module

### Steps to test this PR

Run tests


app/src/test/java/com/duckduckgo/app/sitepermissions/PermissionsPerWebsiteViewModelTest.kt
```diff
@@ -0,0 +1,195 @@
...
+import org.mockito.kotlin.whenever
+
+@ExperimentalCoroutinesApi
+class PermissionsPerWebsiteViewModelTest {
+
+    @ExperimentalCoroutinesApi
+    @get:Rule
+    var coroutineRule = CoroutineTestRule()
+
+    private lateinit var viewModel: PermissionsPerWebsiteViewModel
```
emartynov, line 54: Sorry for the subjective comment out of nowhere.

Since you control the lifecycle of this class, you could use: 
1. The `init` block for view model construction
2. A lazy property to instantiate the view model on the first use
3. Extract function/builder of view model and call it in every test

The second is preferred for me since every test has quick navigation to this function if the reviewer wants to see details of this construction.

site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsManageTest.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    private lateinit var testee: SitePermissionsManager
+
+    private val mockSitePermissionsRepository: SitePermissionsRepository = mock()
+
+    private val url = "https://domain.com/whatever"
+
+    @Before
+    fun before() {
+        testee = SitePermissionsManagerImpl(mockSitePermissionsRepository)
```
cmonfortep, line 49: you don't need the `before()` method. You can directly have:

`private val testee = SitePermissionsManagerImpl(mockSitePermissionsRepository)` as class attribute

site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsManageTest.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test
+    fun givenListOfPermissionsThenFilterGranted() = runTest {
+        val tabId = "tabId"
+        val resources = arrayOf(PermissionRequest.RESOURCE_AUDIO_CAPTURE, PermissionRequest.RESOURCE_VIDEO_CAPTURE)
+        whenever(mockSitePermissionsRepository.isDomainGranted(url, tabId, PermissionRequest.RESOURCE_AUDIO_CAPTURE)).thenReturn(true)
+        whenever(mockSitePermissionsRepository.isDomainGranted(url, tabId, PermissionRequest.RESOURCE_VIDEO_CAPTURE)).thenReturn(false)
+
+        val permissionsGranted = testee.getSitePermissionsGranted(url, tabId, resources)
+        assertEquals(1, permissionsGranted.size)
```
cmonfortep, line 60: should we assert if we received the expected one?

site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsManageTest.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test
+    fun givenListOfPermissionsThenFilterNotSupportedAndReturnOnlyPermissionsAllowedToAsk() = runTest {
+        val resources =
+            arrayOf(PermissionRequest.RESOURCE_VIDEO_CAPTURE, PermissionRequest.RESOURCE_MIDI_SYSEX, PermissionRequest.RESOURCE_AUDIO_CAPTURE)
+        whenever(mockSitePermissionsRepository.isDomainAllowedToAsk(url, PermissionRequest.RESOURCE_VIDEO_CAPTURE)).thenReturn(true)
+        whenever(mockSitePermissionsRepository.isDomainAllowedToAsk(url, PermissionRequest.RESOURCE_AUDIO_CAPTURE)).thenReturn(false)
+
+        val permissionsAllowedToAsk = testee.getSitePermissionsAllowedToAsk(url, resources)
+        assertEquals(1, permissionsAllowedToAsk.size)
```
cmonfortep, line 71: Same. could we assert if we received the expected one? it's ok to check the size, but I guess we should assert we got the expected one.

site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsRepositoryTest.kt
```diff
@@ -0,0 +1,262 @@
...
+    private val mockSitePermissionsDao: SitePermissionsDao = mock()
+    private val mockSitePermissionsAllowedDao: SitePermissionsAllowedDao = mock()
+    private val mockSitePermissionsPreferences: SitePermissionsPreferences = mock()
+
+    private val url = "https://domain.com/whatever"
+    private val domain = "domain.com"
+
+    @Before
+    fun before() {
+        repository = SitePermissionsRepositoryImpl(
```
cmonfortep, line 62: Ditto. This can be extracted into a class attribute and we could remove the `@Before` method.

site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsRepositoryTest.kt
```diff
@@ -0,0 +1,262 @@
...
+        val permission = PermissionRequest.RESOURCE_VIDEO_CAPTURE
+        val tabId = "tabId"
+        val sitePermissionAllowedEntity = SitePermissionAllowedEntity(domain, tabId, permission, setAllowedAtTime(12))
+        whenever(mockSitePermissionsAllowedDao.getSitePermissionAllowed(domain, tabId, permission)).thenReturn(sitePermissionAllowedEntity)
+
+        assertTrue(repository.isDomainGranted(url, tabId, permission))
+    }
+
+    @Test
+    fun whenSitePermissionsWasGrantedWithin24hThenReturnPermissionNotGranted() {
```
cmonfortep, line 125: whenSitePermissions**WasGrantedWithin24h**? Probably more than 24h ago?

app/src/test/java/com/duckduckgo/app/sitepermissions/PermissionsPerWebsiteViewModelTest.kt
```diff
@@ -0,0 +1,195 @@
...
+    private val mockSitePermissionsRepository: SitePermissionsRepository = mock()
+    private val mockLocationPermissionsRepository: LocationPermissionsRepositoryAPI = mock()
+    private val mockSettingsDataStore: SettingsDataStore = mock()
+    private val mockPixel: Pixel = mock()
+
+    private val domain = "domain.com"
+
+    @Before
+    fun before() {
+        viewModel = PermissionsPerWebsiteViewModel(
```
cmonfortep, line 65: NIT: we could move `viewmodel` as a class attribute, and move the other methods to the tests that need that setup. That way we won't need the `before()` method. This is optional because we don't have any established guidelines.

I have a personal preference to avoid, if possible, `before()` methods on tests. (Maybe I will propose it)
Basically, I prefer helper methods (the ones you already have like `loadAskForPermissionsPrefs`) and use them on each test case than having them in the `before()` method.
why?
- Easier to read tests if all required code is in each test case. Logic inside `before()` is usually hidden
- Avoid mocking objects that are not required on your tests
- if a test needs a different setup, you will need to override things that were initialized in the `before()`
- etc.

app/src/test/java/com/duckduckgo/app/sitepermissions/SitePermissionsViewModelTest.kt
```diff
@@ -0,0 +1,197 @@
...
+
+    private val mockSitePermissionsRepository: SitePermissionsRepository = mock()
+    private val mockLocationPermissionsRepository: LocationPermissionsRepositoryAPI = mock()
+    private val mockGeoLocationPermissions: GeoLocationPermissions = mock()
+    private val mockSettingsDataStore: SettingsDataStore = mock()
+    private val mockPixel: Pixel = mock()
+
+    @Before
+    fun before() {
+        viewModel = SitePermissionsViewModel(
```
cmonfortep, line 62: ditto.

cmonfortep (approved):
LGTM Thanks for the good effort here. Tests are well detailed and we are covering many test scenarios for each class.

just left some optional comments about a preference I have. I explained some of the benefits, happy to talk about it if you want to know more.

site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsManageTest.kt
nalcalag: I can add that assertion too. Thanks for the suggestion!

site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsRepositoryTest.kt
nalcalag: Good catch! Thanks :)

app/src/test/java/com/duckduckgo/app/sitepermissions/PermissionsPerWebsiteViewModelTest.kt
nalcalag: I agree on this case because it's true that some of the mocks in `before()` are override in some functions. I will change it 👍

app/src/test/java/com/duckduckgo/app/sitepermissions/SitePermissionsViewModelTest.kt
nalcalag: In this case I will keep it because it will introduce much more lines of code and the logic is shared between all tests. I will change the constructor to the variable declaration though 👍

nalcalag:
* develop
  * **PR #2346** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2346" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2346?utm_source=stack-comment).

PR #2356, opened 2022-10-04
Migrate Site Permissions UI to available components

nalcalag:
Task/Issue URL: https://app.asana.com/0/0/1203073463571585/f

### Description
Use new design component `SectionHeaderTextView` and removed some legacy code I needed to keep in order to display existing layouts

### Steps to test this PR

_Site Permissions dialog_
- Install from this branch
- Go to www.mictest.cc
- Tap on "Test my cam" button
- [ ] Check site permissions dialog appears

_Site Permissions screen_
- Install from this branch
- Go to Settings > Site Permissions
- [ ] Check everything looks good


nalcalag:
* develop
  * **PR #2356** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2356" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2356?utm_source=stack-comment).

aitorvs (commented):
You can also take this opportunity to remove the `companion object` in `SitePermissionsRequestedType`. Doesn't seem to be used

site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
```diff
@@ -100,18 +102,24 @@ class SitePermissionsDialogActivityLauncher @Inject constructor(
...
-                pixel.fire(SitePermissionsPixelName.SITE_PERMISSION_DIALOG_ALLOWED, mapOf(PixelParameter.SITE_PERMISSION to pixelParamValue))
-                onPermissionAllowed()
-            }
-            setNegativeButton(R.string.sitePermissionsDialogDenyButton) { _, _ ->
-                pixel.fire(SitePermissionsPixelName.SITE_PERMISSION_DIALOG_DENIED, mapOf(PixelParameter.SITE_PERMISSION to pixelParamValue))
-                sitePermissionRequest.deny()
-            }
-            setCancelable(false)
-            show()
+        (activity as FragmentActivity?)?.supportFragmentManager?. let { fragmentManager ->
```
aitorvs, line 105: code style: I think we should remove the space between the `.` and the `.let`
```suggestion
        (activity as FragmentActivity?)?.supportFragmentManager?.let { fragmentManager ->
```

site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
```diff
@@ -254,18 +262,23 @@ class SitePermissionsDialogActivityLauncher @Inject constructor(
...
-            setPositiveButton(R.string.systemPermissionsDeniedDialogPositiveButton) { _, _ ->
-                val intent = Intent()
-                intent.action = Settings.ACTION_APPLICATION_DETAILS_SETTINGS
-                val uri = Uri.fromParts("package", activity.packageName, null)
-                intent.data = uri
-                context.startActivity(intent)
-            }
-            setNegativeButton(R.string.systemPermissionsDeniedDialogNegativeButton) { _, _ -> }
-            show()
+        (activity as FragmentActivity?)?.supportFragmentManager?. let { fragmentManager ->
```
aitorvs, line 265: Ditto
```suggestion
        (activity as FragmentActivity?)?.supportFragmentManager?.let { fragmentManager ->
```

site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
```diff
@@ -254,18 +262,23 @@ class SitePermissionsDialogActivityLauncher @Inject constructor(
...
+                .setMessage(contentRes)
+                .setPositiveButton(R.string.systemPermissionsDeniedDialogPositiveButton)
+                .setNegativeButton(R.string.systemPermissionsDeniedDialogNegativeButton)
+                .addEventListener(object : EventListener() {
+                    override fun onPositiveButtonClicked() {
+                        val intent = Intent()
+                        intent.action = Settings.ACTION_APPLICATION_DETAILS_SETTINGS
+                        val uri = Uri.fromParts("package", activity.packageName, null)
+                        intent.data = uri
+                        activity.startActivity(intent)
```
aitorvs, line 277:
```suggestion
                        context.startActivity(intent)
```

site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
aitorvs, line 105: This is a very confusing and prone to error public API now. 
For one I don't think `Activity` should be part of the public API. But even in that case, we're asking for an `Activity` but internally expect now a `FragmentActivity`.

I fail to understand why it's the responsibility of a library to actually show the dialog using `supportFragmentManager`. This is logic that belongs to the user of the library, not the library itself.

nalcalag: `supportFragmentManager` was not needed at the time of the API implementation. I agree cast activity to `FragmentActivity` is not safe enough for the API. Alternative solutions I can think of:

-  I can cast it to `DuckDuckGoActivity` which should always be true if you call it from another activity ✅ 
- Pass `fragmentManager` as parameter to the API function
- Change the new component builder to something easier and default several values -> Not scoped for this PR

In the beginning I was just injecting activity context for showing dialogs, however I also needed `activity` for checking `shouldShowRequestPermissionRationale` in system permissions helper. That's why the activity is needed.

This library has wrapped all the functionality related to site permissions and system permissions combined. So you send the request and this library will handle check which permissions needs to request to the user and in case user denies/allows site permissions + system permissions it will automatically grant or deny them to the website. Therefore, the user of the library will not need to think about which dialogs or system permissions should show to the user.

site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
nalcalag: You are right! I was using it before this change, not sure why I changed it. Thanks! 💯

malmstein (approved):
nice one, thanks @nalcalag!

PR #2366, opened 2022-10-06
ADS: Fix list items text selector

malmstein:
Task/Issue URL: https://app.asana.com/0/1202857801505092/1203118789302972

### Description
The newly created OneLineListItem doesn't respect the disabled state. It's text color remains the same as when the component is enabled.

### Steps to test this PR

_In internal build flavour_
- [ ] Open Settings
- [ ] App Components Preview
- [ ] Move to List Items tab
- [ ] Check last item under One list Item
- [ ] It should be disabled, and the text color different to the other items


common-ui/src/main/java/com/duckduckgo/mobile/android/themepreview/ui/component/ComponentViewHolder.kt
```diff
@@ -99,6 +99,16 @@ sealed class ComponentViewHolder(val view: View) : RecyclerView.ViewHolder(view)
...
             }
+
+            view.findViewById<OneLineListItem>(R.id.oneLineListItemDisabled).apply {
+                setClickListener { Snackbar.make(this.rootView, component.name, Snackbar.LENGTH_SHORT).show() }
+                isEnabled = false
+            }
+
+            view.findViewById<OneLineListItem>(R.id.oneLineListItemCustomTextColor).apply {
+                setClickListener { Snackbar.make(this.rootView, component.name, Snackbar.LENGTH_SHORT).show() }
+            }
```
aitorvs, line 110: Why do we need this now?

common-ui/src/main/res/values/attrs_lists.xml
```diff
@@ -16,17 +16,18 @@
 
 <resources>
 
-    <!--Common lists attributes-->
+    <!--Default lists attributes-->
     <attr name="listItemTitleColor" format="color"/>
     <attr name="listItemSubtitleColor" format="color"/>
```
nalcalag, line 21: NIT: Maybe we could update these attribute names to `listItemPrimaryTextColor` and `listItemSecondaryTextColor` for consistency.

common-ui/src/main/res/layout/component_two_line_item.xml
```diff
@@ -151,13 +151,37 @@
...
+            android:id="@+id/twoLineSwitchListItemWithSecondaryTextColorOverlay"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            app:primaryText="Two Line Item"
+            app:secondaryText="With custom Secondary Text color"
+            app:leadingIcon="@drawable/ic_globe_gray_16dp"
+            app:secondaryTextColorOverlay="@color/red_text_color_selector"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@id/twoLineSwitchListItemWithPrimaryTextColorOverlay"/>
```
nalcalag, line 185: NIT: You haven't added onClickListener for these new list items in the fragment. Maybe it was intentionally, so feel free to ignore it 🙂

nalcalag (approved):
Approved 🎉  I've just added a couple of comments which are optional and there is also a question from @aitorvs.
_Question_: Showing switch or leading icon are not compatible but you can set both attributes instead having an enum `attr`. I've checked switch overrides leading icon which I suppose was expected? Anyways, we can talk about this offline because this is not scoped for this PR 👍

common-ui/src/main/java/com/duckduckgo/mobile/android/themepreview/ui/component/ComponentViewHolder.kt
malmstein: These two are variants that weren't showcased earlier:
- DIsabled item (so text color is different to the enabled one)
- Item with custom text color

common-ui/src/main/res/values/attrs_lists.xml
malmstein: good idea!

PR #2376, opened 2022-10-10
ADS: Fix Divider Lint

malmstein:
Task/Issue URL:  https://app.asana.com/0/inbox/1157893581871899/1203130286824893/1203130286824895

### Description
Wrong assertion in DividerTest
I also removed the AlertDialogTest since it's not working. I'll get it fixed and back once we migrate all dialogs

### Steps to test this PR
Build should be green

---
  - https://app.asana.com/0/0/1203130286824893


lint-rules/src/main/java/com/duckduckgo/lint/ui/NoAlertDialogDetector.kt
```diff
@@ -73,7 +73,7 @@ class NoAlertDialogDetector : Detector(), SourceCodeScanner {
                 androidSpecific = true,
                 implementation = Implementation(
                     NoAlertDialogDetector::class.java,
-                    Scope.RESOURCE_FILE_SCOPE
+                    Scope.JAVA_FILE_SCOPE
```
CDRussell, line 76: I don't know what this change does, so just flagging to ensure it was deliberate.

CDRussell (approved):
`:lint-rules:test` passing for me now 👍

lint-rules/src/main/java/com/duckduckgo/lint/ui/NoAlertDialogDetector.kt
malmstein: In Lint terms, the scope defines when is this rule being inspected. I had this wrongly set to resource files, so Lint wasn't really working. `JAVA_FILE_SCOPE` applies to all source files, Java and Kotlin, which is what we want.

PR #2390, opened 2022-10-12
Pull exception lists from Privacy Config

nshuba:
Task/Issue URL: https://app.asana.com/0/0/1203265536113085/f

### Description
Update AppTP to pull exception lists from Privacy Config

### Steps to test this PR

_Clean Install and List Update_
- [ ] Pull this branch and update `PrivacyConfigService` to pull from https://jsonblob.com/api/jsonBlob/1032428887661101056
- [ ] Install fresh
- [ ] Smoke tests for AppTP
- [ ] Verify that Chrome does not appear in list of protected apps
- [ ] Verify that protection for Clock app is disabled
- [ ] Verify (through logs) that `sb.scorecardsearch.com` is allowed when running Speedtest
- [ ] Use https://jsonblob.com/1032428887661101056 to update the version of the config and the hash within the `appTrackerProtection` object
- [ ] Make changes to the exclusion lists within the `appTrackerProtection` object
- [ ] Force quit the app and start again (to force update the privacy config)
- [ ] Verify that your changes are applied. __Note__: there is a bug with `vpn_app_tracker_exception_rules` table where the table is wiped on VPN restart. So changes to `appTrackerAllowList` will not be reflected. This bug was pre-existing and will be fixed in a future PR.

_DB Update_
- [ ] Install from `develop` and then install from this branch
- [ ] Smoke tests for AppTP
- [ ] Verify in App Inspector that VPN DB contains all 3 exclusion lists


app-tracking-protection/vpn-store/src/main/java/com/duckduckgo/mobile/android/vpn/store/VpnDatabase.kt
```diff
@@ -137,6 +137,19 @@ abstract class VpnDatabase : RoomDatabase() {
...
+                database.execSQL("DROP TABLE IF EXISTS `vpn_app_tracker_exclusion_list`")
+
+                database.execSQL(
+                    "CREATE TABLE IF NOT EXISTS `vpn_app_tracker_exclusion_list` " +
+                        "(`packageName` TEXT PRIMARY KEY NOT NULL, " +
+                        "`reason` TEXT NOT NULL DEFAULT 'UNKNOWN')"
+                )
+            }
+        }
+
```
aitorvs, line 152: Why do we want to rename to `packageName`. I think `packageId` is not really a misnomer and I'd rather avoid the table drop and the noise of touching so many files just for renaming one field.

nshuba: I was on the fence about this one as well. But our other lists use `packageName` as their field so I though it best if all our lists used either `packageName` or  `packageId` for consistency. This way we can even require this field to be set in our lists in the upcoming privacy-config schemas.

aitorvs: Why do we need to rename the db table column to enforce any schema in the JSON?
Our db tables don't (purposely) share schemas with the protocols over the wire.

Also I rather not rename fields to avoid having to think about cases like what happens to users that update the app, but now we drop the table, they need to re-fetch from remote but the eTag hasn't changed.
For example, I haven't tested this PR but I doubt it'd work in all cases, because the `vpn_app_tracker_exclusion_list_metadata` db table–that holds the eTag– is not dropped, and probably is should be as well.

I get the reason to rename, but I don't think is worth the trouble, we `packageId` in plenty of places anyways.

nshuba: Those are great points - I didn't realize the details of the update/eTag mechanism. Let me see if I can re-work this to take in the new name from the schema but not change the DB.

app-tracking-protection/vpn-store/src/main/java/com/duckduckgo/mobile/android/vpn/dao/VpnAppTrackerBlockingDao.kt
```diff
@@ -87,9 +87,6 @@ interface VpnAppTrackerBlockingDao {
     @Query("SELECT * from vpn_app_tracker_exclusion_list")
     fun getAppExclusionListFlow(): Flow<List<AppTrackerExcludedPackage>>
 
-    @Insert
-    fun setExclusionListMetadata(appTrackerExclusionListMetadata: AppTrackerExclusionListMetadata)
```
nshuba, line 91: To keep the PR smaller and safer, I did not fully remove all metadata references. Do we think it's ok or should I remove the relevant tables and references? Or do we want to implement a similar mechanism in the PrivacyConfig?

app-tracking-protection/vpn-store/src/main/java/com/duckduckgo/mobile/android/vpn/store/VpnDatabase.kt
```diff
@@ -204,6 +204,12 @@ abstract class VpnDatabase : RoomDatabase() {
             }
         }
 
+        private val MIGRATION_31_TO_32: Migration = object : Migration(31, 32) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("ALTER TABLE `vpn_app_tracker_exclusion_list` ADD COLUMN `reason` TEXT NOT NULL DEFAULT 'UNKNOWN'")
```
aitorvs, line 209: Given that we see crash increase every time we do a migration, and the current theory is that some users have a pretty full disk and we never vacuumed previous db tables improvements. Maybe we should do a `dataase.execSQL("VACUUM")` as first statement?

@marcosholgado what do you think?

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpPrivacyFeaturePlugin.kt
```diff
@@ -16,21 +16,31 @@
...
 import com.duckduckgo.di.DaggerSet
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.privacy.config.api.PrivacyFeaturePlugin
 import com.squareup.anvil.annotations.ContributesMultibinding
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
 import javax.inject.Inject
+import logcat.logcat
+
+private const val APPTP_FEATURE_SIGNATURE_KEY = "apptp_feature_signature"
```
aitorvs, line 31: nit: we're mixing `companion object` with high level const. I don't have strong preferences, but in this file maybe we should pick one so that all const are grouped in the same place (?)

app-tracking-protection/vpn-api/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpSetting.kt
```diff
@@ -25,6 +25,7 @@ enum class AppTpSetting(override val value: String, override val defaultValue: B
     AlwaysSetDNS("alwaysSetDNS"),
     CPUMonitoring("cpuMonitoring"),
     ProtectGames("protectGames"),
+    ExceptionLists("exceptionLists", defaultValue = true),
```
aitorvs, line 28: Why the default value set to `true`? This is not a feature that will/can be disabled

aitorvs (commented):
Left couple comments, need to do another round of testing tomorrow

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpPrivacyFeaturePlugin.kt
nshuba: Good point - fixed!

app-tracking-protection/vpn-api/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpSetting.kt
nshuba: Right, there should be no way to disable it. But why not be safe and set the default to `true`?

aitorvs (approved):
LGTM!

PR #2397, opened 2022-10-13
Privacy Dashboard new UI updated

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203158752116062/f

### Description
This PR (follow up of https://github.com/duckduckgo/Android/pull/2081) integrates the new privacy dashboard (JS) which includes extended information about all type of trackers.

### Steps to test this PR

_Site with trackers_
- [x] navigate to a site with trackers (examples in https://app.asana.com/0/0/1202620291996749/f)
- [x] click on the privacy shield
- [x] ensure dashboard renders tracker information

_Site without trackers_
- [x] navigate to a site without trackers (examples in https://app.asana.com/0/0/1202620291996749/f)
- [x] click on the privacy shield
- [x] ensure dashboard says no trackers found

_Major tracking site_
- [x] navigate to a major tracking site (examples in https://app.asana.com/0/0/1202620291996749/f)
- [x] click on the privacy shield
- [x] ensure dashboard says is a major tracking site

_http site_
- [x] navigate to a HTTP site (examples in https://app.asana.com/0/0/1202620291996749/f)
- [x] click on the privacy shield
- [x] ensure dashboard says connection is not secure

For extra points: [you can checkout this branch and run maestro tests](https://github.com/duckduckgo/Android/tree/feature/cristian/privacy_dashboard/cookie_animation/ad-click/ad-click-impl/src/test/resources/maestro_tests/ad_click_detection_flows)


browser-api/src/main/java/com/duckduckgo/app/tabs/BrowserNav.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.tabs
+
+import android.content.Context
+import android.content.Intent
+
+interface BrowserNav {
+    fun openInNewTab(context: Context, url: String): Intent
```
marcosholgado, line 23: nit: should be documented

package.json
```diff
@@ -18,6 +18,6 @@
     "@duckduckgo/autoconsent": "github:duckduckgo/autoconsent#v3.0.0",
     "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#2.6.0",
     "@duckduckgo/privacy-reference-tests": "github:duckduckgo/privacy-reference-tests#1.0.0",
-    "@duckduckgo/privacy_dashboard": "github:more-duckduckgo-org/duckduckgo-privacy-dashboard#ca7b5f444455b587a6bafab9c2b8a44ec99dad53"
+    "@duckduckgo/privacy_dashboard": "github:duckduckgo/privacy-dashboard#main"
```
marcosholgado, line 21: nit: should be pointing to a version rather than main to avoid future errors. Suggest to push the frontend team to do proper releases :)

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/ProtectionStatusViewStateMapper.kt
```diff
@@ -0,0 +1,53 @@
...
+@ContributesBinding(AppScope::class)
+class AppProtectionStatusViewStateMapper @Inject constructor(
+    private val contentBlocking: ContentBlocking,
+    private val unprotectedTemporary: UnprotectedTemporary
+) : ProtectionStatusViewStateMapper {
+
+    override fun mapFromSite(site: Site): ProtectionStatusViewState {
+        val enabledFeatures = mutableListOf<String>().apply {
+            if (!contentBlocking.isAnException(site.url)) {
+                add(ContentBlockingFeatureName.value)
```
marcosholgado, line 42: I don't fully understand why this is needed? Perhaps a comment explaining would be useful.

The other part is that I don't think this scales. If you needed to add 40 features the constructor wouldn't be very nice :D. Probably a plugin to provide these values would be better. I suggest for now just leaving a comment with the plugin idea in case we come back here in the future to add more features.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/RequestDataViewStateMapper.kt
```diff
@@ -0,0 +1,128 @@
...
+        "Embedded Content"
+    )
+
+    override fun mapFromSite(site: Site): RequestDataViewState {
+        val installedSurrogates = mutableListOf<String>()
+
+        val uniqueEntityDomainState = mutableMapOf<String, List<TrackerStatus>>()
+        val requests: List<DetectedRequest> = site.trackingEvents.map {
+            val withSchemeTrackeUrl = kotlin.runCatching { it.trackerUrl.toUri().withScheme().toString() }.getOrNull() ?: return@map null
+            val trackerEvent = it.copy(trackerUrl = withSchemeTrackeUrl)
```
marcosholgado, line 66: nit: `withSchemeTrackeUrl` has a typo, should be `withSchemeTrackerUrl`

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/RequestDataViewStateMapper.kt
```diff
@@ -0,0 +1,128 @@
...
+
+interface RequestDataViewStateMapper {
+    fun mapFromSite(site: Site): RequestDataViewState
+}
+
+@ContributesBinding(AppScope::class)
+class AppSiteRequestDataViewStateMapper @Inject constructor() : RequestDataViewStateMapper {
+
+    private val publicSuffixDatabase = PublicSuffixDatabase()
+    private val allowedCategories = listOf(
```
marcosholgado, line 52: I have a generic comment. My gut feeling is that I've already seen a lot of this code in classes like `Site`, `SiteMonitor`, `TrackerDetector`, etc. I'm guessing all that code now lives here and we are not duplicating it? Just double checking.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -182,16 +193,16 @@ class PrivacyDashboardHybridViewModel @Inject constructor(
 
         viewModelScope.launch(dispatcher.io()) {
             if (enabled) {
-                userWhitelistDao.delete(currentViewState().siteProtectionsViewState.site.domain)
+                userWhitelistDao.delete(currentViewState().siteViewState.domain)
```
marcosholgado, line 196: nit: we should ideally try to avoid using daos directly and instead start adding methods to UserAllowListRepository when needed. Not really asking to do it here but FYI :)

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -48,37 +48,72 @@ import javax.inject.Inject
 @ContributesViewModel(ActivityScope::class)
 class PrivacyDashboardHybridViewModel @Inject constructor(
     private val userWhitelistDao: UserWhitelistDao,
-    private val contentBlocking: ContentBlocking,
     private val pixel: Pixel,
     private val dispatcher: DispatcherProvider,
     @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
```
marcosholgado, line 53: This scope is not used anymore.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -48,37 +48,72 @@ import javax.inject.Inject
...
+        val entityName: String,
+        val ownerName: String?,
+        val pageUrl: String,
+        val prevalence: Double,
+        val state: State,
+        val url: String
     )
 
-    data class SiteProtectionsViewState(
+    sealed class State {
```
marcosholgado, line 97: I would probably renamed this to RequestState to be more specific because just by reading it I think it refers to the actual state of the view.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/di/JsonModule.kt
```diff
@@ -0,0 +1,47 @@
...
+@ContributesTo(AppScope::class)
+object JsonModule {
+
+    @Provides
+    @SingleInstanceIn(AppScope::class)
+    @Named("privacyDashboard")
+    fun moshi(moshi: Moshi): Moshi {
+        return moshi.newBuilder()
+            .add(
+                PolymorphicJsonAdapterFactory.of(State::class.java, "state")
```
marcosholgado, line 41: The State, Blocked and Allowed classes live in the view model, I'm not sure if that's by design but if someones changes them there thinking they are specific to the view model and adds something that's not expected when parsing using moshi I guess it could create some issues.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -48,37 +48,72 @@ import javax.inject.Inject
...
+        data class Blocked(val blocked: Any = Any()) : State()
```
marcosholgado, line 98: Why Any? Can be more specific here?

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -48,37 +48,72 @@ import javax.inject.Inject
...
+        data class Allowed(val allowed: Reason) : State()
```
marcosholgado, line 99: The val should probably be named `reason` for consistency, `allow` doesn't really make a lot of sense.

marcosholgado (approved):
Overall LGTM, left a few comments.

package.json
cmonfortep: Yes, totally agree, forgot to mention that they are already aware of this and we will have proper release numbers.

I'm going to use hash commit until releases are ready. I was using `main` during development just for convenience, but once this is released, we would like to target something "static".

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/ProtectionStatusViewStateMapper.kt
cmonfortep: gotcha, I know what you mean. Yes, if at some point the number of privacy features we need to depend on here grows, we could change the way we provide this list. Using plugins is an approach I also like, so I will leave a comment. Right now, it's only one feature, so it would be premature to extract that part.

I will leave a comment linking the docs with a brief explanation. "List of enabled features that are supported by the privacy dashboard".  (right now only contentblocking) https://duckduckgo.github.io/privacy-dashboard/example/docs/interfaces/Generated_Schema_Definitions.ProtectionsStatus.html#enabledFeatures

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/RequestDataViewStateMapper.kt
cmonfortep: if we are talking about `allowedCategories`, this is specific to the privacy dashboard UI (that's why inside the ViewStateMapper), so not sure if similar logic is duplicated somewhere else. If so, I think is also ok, since the list of supported categories in the domain layer doesn't need to necessarily match the UI layer.

if you were referring to something else, please ask again :D

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
cmonfortep: Thanks, I've created a task to migrate all daos. https://app.asana.com/0/1202552961248957/1203198577387937/f

I will keep it under my radar.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
cmonfortep: Thanks for this one, actually I think you are right and I prefer to keep it readable. 👍

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/di/JsonModule.kt
cmonfortep: yes this part is tricky, I'm not a fan of how moshi deals with sealed classes tbh. But I haven't found a way to improve that part since they rely on a `String` to indicate the json `key`.

cmonfortep: I guess I'm going to leave a comment to at least avoid those unexpected issues.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
cmonfortep: 😅 no, we can't. We need a placeholder value there so generated json is the expected one:
```
"state": {
        "blocked": {}
      }
```
tried a few things, but if you put `null` or empty values, Moshi discards adding the object. That's why ended up with that patch. Not the best thing tho, I agree.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
cmonfortep: this is just to mimic the values on the UI ([so the privacy dashboard](https://duckduckgo.github.io/privacy-dashboard/example/docs/interfaces/Generated_Schema_Definitions.DetectedRequest.html#state)). I see it can lead to confusion and write a comment in the ViewModel.

PR #2406, opened 2022-10-14
[Part 4] Implementation for Exclusion List in AppTP

anikiki:
Task/Issue URL: https://app.asana.com/0/0/1203134519344743/f

### Description
Added filters for app exclusion list. This PR doesn't contain any update on copy.

### Steps to test this PR

- [x] Fresh install from this branch.
- [x] Enable AppTP from Settings.
- [x] Tap on the Protected Apps item (x of your apps are protected from ...) in the `Activity` section.
- [x] On `All Apps` screen notice the `Protection Enabled (x)` filter is applied and has the correct number of apps in parentheses and all apps displayed in the list are protected.
- [x] Select different filters (`All`, `Protection Enabled`, `Protection Disabled`) and notice the list is correctly updated.
- [x] Go back to the previous screen.
- [x] Tap on the Unprotected Apps item (x of your apps have protection automatically disabled) in the `Activity` section.
- [x] On `All Apps` screen notice the `Protection Disabled (x)` filter is applied and has the correct number of apps in parentheses and all apps displayed in the list are unprotected.
- [x] Try to enable protection on some apps in the list and notice the list is updated.

### UI changes
| Filters popup  | "All" filter selected |
| ------ | ----- |
|Filters_popup|All|

| "Protection Enabled" filter selected  | "Protection Disabled" filter selected |
| ------ | ----- |
|Protection_enabled|Protection_disabled|


PIYUSH-MISHRA-00 (commented):
Implementation for Exclusion List in AppTP

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ManageAppsProtectionViewModel.kt
```diff
@@ -74,7 +79,19 @@ class ManageAppsProtectionViewModel @Inject constructor(
             .launchIn(viewModelScope)
     }
 
-    internal suspend fun getProtectedApps() = excludedApps.getAppsAndProtectionInfo().map { ViewState(it) }
+    internal suspend fun getProtectedApps(): Flow<ViewState> {
+        return excludedApps.getAppsAndProtectionInfo()
```
malmstein, line 83: This is accessing disk every single time, would it make sense to keep an in memory list? we avoid accessing IO each time we change the filter and it would be much faster. In my device the change from Protection Disabled to Protection Enabled takes half a second

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ui/TrackingProtectionExclusionListActivity.kt
```diff
@@ -90,6 +92,8 @@ class TrackingProtectionExclusionListActivity :
         bindViews()
         observeViewModel()
 
+        viewModel.applyAppsFilter(getAppsFilterOrDefault())
```
malmstein, line 95: `observeViewModel` launches a coroutine that calls `viewModel.getProtectedApps()`. viewModel.applyAppsFilter() will also trigger the same pipeline. I think you can remove one of the two calls.

app-tracking-protection/vpn-impl/src/main/res/layout/activity_tracking_protection_exclusion_list.xml
```diff
@@ -56,6 +56,16 @@
                     app:panelDrawable="@drawable/ic_alert_color_24"
                     android:layout_margin="16dp"/>
 
+            <TextView
+                android:id="@+id/excludedAppsFilterText"
```
malmstein, line 60: this needs a  `android:background="?selectableItemBackground"` so there's touch feedback

malmstein: I think we can keep this invisible until the first `ViewState` is received. Otherwise we can get to the situation where the drawable is visible but there's no text in the `TextView`

Screenshot_20221018_141431

malmstein (approved):
Couple suggestions for the UI and calls to the `VM`. Nothing blocking, looking great @anikiki !

app-tracking-protection/vpn-impl/src/main/res/layout/activity_tracking_protection_exclusion_list.xml
anikiki: - Added `android:background="?selectableItemBackground"`.
- `I think we can keep this invisible until the first ViewState is received` - this is done in the latest code (will be available in Part 5) along with loading states and empty banner.

PR #2409, opened 2022-10-14
Automate update ref tests

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203158712142231/f 

### Description
This PR adds a new workflow to automatically update reference tests once a day. In case of a new version published it will create a new PR with those changes triggering all the necessary tests.

### Steps to test this PR

- [ ] This workflow triggered https://github.com/duckduckgo/Android/pull/2413 which automatically added a change in a reference tests and update the library to 1.0.0 although that change is not visible because on develop is already 1.0.0

### DIFF
--- a/.github/workflows/build-debug-apk.yaml
+++ b/.github/workflows/build-debug-apk.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches:
       - develop
+    paths-ignore:
+      - '**.md'
+      - '.github/**'
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,9 +7,11 @@ on:
       - develop
     paths-ignore:
       - '**.md'
+      - '.github/**'
   pull_request:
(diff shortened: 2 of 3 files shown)


marcosholgado:
* develop
  * **PR #2409** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2409" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2409?utm_source=stack-comment).

cmonfortep (approved):
LGTM (a bit confusing that failure means an update is available, but I guess is how `ncu` works 🤷)

marcosholgado:
> LGTM (a bit confusing that failure means an update is available, but I guess is how `ncu` works 🤷)

It's a feature for ncu, normally it will always succeed wether there is a new version available or not. For CI we want to be able to know when there is or isn't and we can make that happen by forcing an exit 1 when there is an update available. Probably makes sense the other way around but is how it works :D

PR #2398, opened 2022-10-13
Improve always on promotion

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203149173362393/f

### Description
Implement the new always On promotion.

This PR:
* implements the new bottom sheet for always on promotion
* removes the code from the previous dialog used for promotion
* handles the case for when user enables always on and lock down options in Android settings

### Steps to test this PR

_Test always on promotion dialog_
- [x] install from this branch, launch and enable AppTP
- [x] Disable appTP
- [x] Modify last entry (highest ID) in `vpn_service_state_stats` table db in the `vpn.db` so that `stopReason` is set to `UNKNOWN`
- [x] Enable AppTP
- [x] verify the new bottom sheet dialog is shown
- [x] Click on not now
- [x] verify the dialog is dismissed
- [x] Disable appTP
- [x] Modify last entry (highest ID) in `vpn_service_state_stats` table db in the `vpn.db` so that `stopReason` is set to `UNKNOWN`
- [x] Enable AppTP
- [x] click on `X`
- [x] verify the dialog is dismissed
- [x] Disable appTP
- [x] Modify last entry (highest ID) in `vpn_service_state_stats` table db in the `vpn.db` so that `stopReason` is set to `UNKNOWN`
- [x] Enable AppTP
- [x] click on open settings
- [x] verify settings are opened
- [x] go back to the app
- [x] verify AppTP is enabled
- [x] Disable appTP
- [x] Modify last entry (highest ID) in `vpn_service_state_stats` table db in the `vpn.db` so that `stopReason` is set to `UNKNOWN`
- [x] Enable AppTP
- [x] click on open settings and enable always on
- [x] Disable appTP
- [x] Modify last entry (highest ID) in `vpn_service_state_stats` table db in the `vpn.db` so that `stopReason` is set to `UNKNOWN`
- [x] Enable AppTP
- [x] verify the dialog DOES NOT appear
- [x] Go to Android VPN settings and disable always on
- [x] Disable appTP
- [x] Modify last entry (highest ID) in `vpn_service_state_stats` table db in the `vpn.db` so that `stopReason` is set to `UNKNOWN`
- [x] Enable AppTP
- [x] verify dialog appears, tap on settings, enable both always on and lock down mode, return to DDG ap
- [x] verify info panel appears, in yellow with message `Action required...`
- [x] tap on `change settings` link
- [x] verify VPN settings appears, disable lock down and go back to DDG app
[...]

PR #2420, opened 2022-10-18
Add lint custom rules for strings

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203158712142223/f

### Description
This PR creates lint rules so strings with placeholder always have instructions and to force placeholders to include their position.

### Steps to test this PR
CI should pass

### DIFF
--- a/lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
+++ b/lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2022 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.lint.strings
+
+import com.android.SdkConstants
+import com.android.ide.common.resources.usage.ResourceUsageModel
+import com.android.tools.lint.detector.api.Category.Companion.CUSTOM_LINT_CHECKS
+import com.android.tools.lint.detector.api.Context
+import com.android.tools.lint.detector.api.Implementation
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.ResourceXmlDetector
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.XmlContext
+import com.android.tools.lint.detector.api.XmlScanner
+import org.w3c.dom.Attr
+import org.w3c.dom.Document
+import org.w3c.dom.Node
+
+@Suppress("UnstableApiUsage")
+class MissingInstructionDetector : ResourceXmlDetector(), XmlScanner {
+
+    private fun ignoreFile(context: Context) =
+        context.file.name.startsWith("donottranslate") ||
+            ResourceUsageModel.isAnalyticsFile(context.file) ||
+            !context.project.reportIssues
+
+
+    override fun visitDocument(context: XmlContext, document: Document) {
+
+        if (ignoreFile(context)) {
+            return
+        }
+
+        val root = document.documentElement
+        val translatable: Attr? = root?.getAttributeNode(SdkConstants.ATTR_TRANSLATABLE)
+        if (translatable?.value != SdkConstants.VALUE_FALSE) {
+            if (context.file.parentFile?.name == SdkConstants.FD_RES_VALUES) {
+                visitNode(context, document)
+            }
+            return
+        }
+
+    }
+
+    private fun visitNode(context: XmlContext, node: Node) {
+        val nodeType = node.nodeType
+        if (nodeType == Node.TEXT_NODE) {
+            val text = node.nodeValue
+            val instruction = node.parentNode.attributes.getNamedItem("instruction")
+            val translatable = node.parentNode.attributes.getNamedItem("translatable")
+            val translatableValue = (translatable as Attr?)?.value
+
+            if (translatableValue != "false" && PLACEHOLDERS.firstOrNull { text.contains(it) } != null && instruction == null) {
+                context.report(
+                    MISSING_INSTRUCTION, node, location = context.getNameLocation(node),
+                    "Missing instructions attribute"
+                )
+            }
+        }
+
+        // Visit children
+        val childNodes = node.childNodes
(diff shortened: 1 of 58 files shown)


marcosholgado:
* develop
  * **PR #2420** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2420" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2420?utm_source=stack-comment).

lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
```diff
@@ -0,0 +1,107 @@
...
+            return
+        }
+
+    }
+
+    private fun visitNode(context: XmlContext, node: Node) {
+        val nodeType = node.nodeType
+        if (nodeType == Node.TEXT_NODE) {
+            val text = node.nodeValue
+            val instruction = node.parentNode.attributes.getNamedItem("instruction")
```
malmstein, line 64: as discussed in https://github.com/duckduckgo/Android/pull/2398/files#r996270777, can you add a check for the annotation?

marcosholgado: I don't really understand what you mean? Which annotation? The discussion there is about the fact that a string inside a strings.xml file (or strings-vpn.xml in that case) must be translated or have a `translatable=false` to pass lint. 

This PR is to trigger a lint error if a string with placeholders doesn't have instructions but it ignores strings in donottranslate.xml and strings with `translatable=false`

malmstein: I'm an idiot and I'll see myself out

lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
```diff
@@ -0,0 +1,116 @@
...
+            val child = childNodes.item(i)
+            visitNode(context, child, parentInstruction)
+            i++
+        }
+    }
+
+
+    companion object {
+        val PLACEHOLDERS = listOf(
+            "%s", "%d", "%1\$"
```
CDRussell, line 99: are `%s` and `%d` still required here given your other lint rule?

lint-rules/src/main/java/com/duckduckgo/lint/strings/PlaceholderDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+            val child = childNodes.item(i)
+            visitNode(context, child)
+            i++
+        }
+    }
+
+
+    companion object {
+        val PLACEHOLDERS = listOf(
+            "%s", "%d"
```
CDRussell, line 84: nit: a niche case that might get missed: floats (which i think are valid placeholders)

CDRussell (approved):
Looks great! 🎉

Left a suggestion and some questions I couldn't figure out from the code, but nothing blocking.

lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
```diff
@@ -0,0 +1,116 @@
...
+    private fun visitNode(context: XmlContext, node: Node, parentHasInstruction: Boolean = false) {
+        var parentInstruction = parentHasInstruction
+        val nodeType = node.nodeType
+        if (nodeType == Node.TEXT_NODE) {
+            val text = node.nodeValue
+            val instruction = node.parentNode.attributes.getNamedItem("instruction")
+            val translatable = node.parentNode.attributes.getNamedItem("translatable")
+            val translatableValue = (translatable as Attr?)?.value
+
+            if (translatableValue != "false" && PLACEHOLDERS.firstOrNull { text.contains(it) } != null && (instruction == null && !parentHasInstruction)) {
```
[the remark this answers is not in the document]
marcosholgado: Yup, done :)

lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.w3c.dom.Attr
+import org.w3c.dom.Document
+import org.w3c.dom.Node
+
+@Suppress("UnstableApiUsage")
+class MissingInstructionDetector : ResourceXmlDetector(), XmlScanner {
+
+    private fun ignoreFile(context: Context) =
+        context.file.name.startsWith("donottranslate") ||
+            ResourceUsageModel.isAnalyticsFile(context.file) ||
```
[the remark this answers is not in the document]
marcosholgado: Probably not, it ignores files called analytics.xml, it is basically a copy&paste from an existing lint rule. In any case, the idea is to ignore as much files as possible

lint-rules/src/test/java/com/duckduckgo/lint/strings/PlaceholderDetectorTest.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    @Test
+    fun whenStringHasDPlaceholderWithOrderThenSuccess() {
+        TestLintTask.lint().files(
+            xml(
+                "res/values/strings.xml",
+                """
+                <?xml version="1.0" encoding="utf-8"?>
+                <resources>
+                    <string name="macos_windows" instruction="">Windows coming soon! %1$\d</string>
```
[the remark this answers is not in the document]
marcosholgado: Ha... actually it is a bug. It should not allow empty instructions :), fixed!

lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    override fun visitDocument(context: XmlContext, document: Document) {
+
+        if (ignoreFile(context)) {
+            return
+        }
+
+        val root = document.documentElement
+        val translatable: Attr? = root?.getAttributeNode(SdkConstants.ATTR_TRANSLATABLE)
+        if (translatable?.value != SdkConstants.VALUE_FALSE) {
```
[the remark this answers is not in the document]
marcosholgado: Not deliberate, just forgot about it, fixed!

lint-rules/src/main/java/com/duckduckgo/lint/strings/MissingInstructionDetector.kt
marcosholgado: I'm going to say yes because the idea is to make sure that if a placeholder exists we require instructions and all of those are placeholders. If we were to delete the other rule then we would be missing a few cases.

lint-rules/src/main/java/com/duckduckgo/lint/strings/PlaceholderDetector.kt
marcosholgado: Oki so, this is interesting. For the record this is wrong, as you say there are many more placeholders and different scenarios but the logic is mind blowing to get it right so instead and to save time I added the typical ones that we use. I've also now added f and o. Should have probably added b to honest but follow up maybe.

PR #2425, opened 2022-10-20
Update reference tests to latest version

daxmobile:
- Automated reference tests dependency update

This PR updates the reference tests dependency to the latest available version and copies the necessary files.
- [x] All tests must pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,6 @@
     "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#5.1.0",
     "@duckduckgo/autoconsent": "github:duckduckgo/autoconsent#v3.0.0",
     "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#3.1.0",
-    "@duckduckgo/privacy-reference-tests": "github:duckduckgo/privacy-reference-tests#1.0.0"
+    "@duckduckgo/privacy-reference-tests": "github:duckduckgo/privacy-reference-tests#1666183539.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,6 +17,6 @@
     "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#5.1.0",
     "@duckduckgo/autoconsent": "github:duckduckgo/autoconsent#v3.0.0",
     "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#3.1.0",
-    "@duckduckgo/privacy-reference-tests": "github:duckduckgo/privacy-reference-tests#1.0.0"
+    "@duckduckgo/privacy-reference-tests": "github:duckduckgo/privacy-reference-tests#1666183539.0.0"
```
cmonfortep, line 20: FYI: weird version number, but not a bug in the workflow. It's the version published in https://github.com/duckduckgo/privacy-reference-tests/releases

marcosholgado: Yeah I know, they are using a timestamp for the version.

cmonfortep: I've noticed, `npm update` throws an error:

```
cristian@Cristians-MacBook-Pro Android % npm update 
npm ERR! code 1
npm ERR! The git reference could not be found
npm ERR! command git --no-replace-objects checkout 1666183539.0.0
npm ERR! error: pathspec '1666183539.0.0' did not match any file(s) known to git
```

Note: I see the workflow uses internally `npm install` which I found was pretty buggy when I was testing npm. Not sure why `ncu` returns 1666183539.0.0 as version, if it doesn't exist in their repo

PR #2477, opened 2022-11-01
Extend OutLinedTextInputView to support restricting to single lines

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203277996585144/f

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


CDRussell:
* develop
  * **PR #2477** <a href="https://app.graphite.dev/github/pr/duckduckgo/android/2477" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/android/2477?utm_source=stack-comment).

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/OutLinedTextInput.kt
```diff
@@ -93,6 +94,17 @@ class OutLinedTextInputView @JvmOverloads constructor(
                     showKeyboard()
                 }
             }
+            getInt(R.styleable.OutLinedTextInputView_android_maxLines, -1).also {
+                if (it >= 0) {
+                    binding.internalEditText.maxLines = it
+                }
+            }
+
+            getInt(R.styleable.OutLinedTextInputView_android_inputType, InputType.TYPE_NULL).also {
```
karlenDimla, line 103: We are setting this already in `setupPasswordMode()` and `setupTextMode()`.  I think this currently overrides what the functions set. Do we only need this to allow singleLine only? 

I think we can either:
1. Retain android:inputType and remove setting inputType in the functions mentioned BUT we would need to manually set the inputType for password fields that we have now to have EditorInfo.TYPE_TEXT_VARIATION_PASSWORD.
2. Don't use android:inputType and just check if maxLines == 1 and pass a boolean to setupPasswordMode and setupTextMode to know if we need to set EditorInfo.TYPE_TEXT_FLAG_MULTI_LINE.


I think option 2 is better for our needs today.

karlenDimla (changes requested):
The changes currently breaks the handling for `app:isPassword`.

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/OutLinedTextInput.kt
karlenDimla: We could also just take `android:inputType` when the field is not a password / isPassword not set. Might be a simpler approach actually.

malmstein: @CDRussell How about setting a custom type with three modes? 

Something like:
```
<attr name="type" format="enum" >
    <enum name="multi_line" value="0"/>
    <enum name="single_line" value="1"/>
    <enum name="password" value="2"/>
</attr>
```
Each mode has it's own configuration, caters for all usages, and has multi_line as the default type.

I really don't like exposing inputType and maxLines as a public API

karlenDimla: @malmstein good suggestion! That should work as well and we can even  remove `app:IsPassword` and just rely on this `type`.

Just out of curiosity, what is the down side of exposing `inputType` and `maxLines`?

CDRussell: I like that approach @malmstein , thanks. will give that a try.

>what is the down side of exposing inputType and maxLines?

Curious about this too. on the one hand, it's sort of leaking implementation details to expose the underlying attributes on the `EditText`. on the other hand, seems like a pain in having to redefine the wheel to allow all existing `EditText` properties to be configurable.

karlenDimla: Thanks @malmstein I agree! I think when creating custom components I am always torn between making components work for as much as possible configurations vs having a defined set of capabilities. I think your explanation above make sense as adding more possible customisations could result to breakage/inconsistencies.

CDRussell: Switched approach as per suggestion above. Thanks!

common-ui/src/main/res/layout/component_outline_text_input.xml
```diff
@@ -21,8 +21,7 @@
     android:orientation="vertical"
     android:paddingStart="@dimen/keyline_4"
     android:paddingTop="@dimen/keyline_5"
-    android:paddingEnd="@dimen/keyline_4"
-    android:paddingBottom="@dimen/keyline_5">
```
CDRussell, line 25: This is strange; having this padding makes the bottom input view get squished. Flagged in another task: https://app.asana.com/0/1174433894299346/1203157856956586

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/OutLinedTextInput.kt
```diff
@@ -220,6 +231,17 @@ class OutLinedTextInputView @JvmOverloads constructor(
         doOnNextLayout { isHintAnimationEnabled = hintAnimationEnabledSnapshot }
     }
 
+    private fun TypedArray.getInputType(): Type {
+        val inputTypeInt = getInt(styleable.OutLinedTextInputView_type, INPUT_TYPE_MULTI_LINE.value)
+        return Type.values().firstOrNull { it.value == inputTypeInt } ?: INPUT_TYPE_MULTI_LINE
```
karlenDimla, line 236: I was slightly confused by this but I think `Type.values()[inputTypeInt]` should be more readable and simpler.

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/OutLinedTextInput.kt
```diff
@@ -154,8 +162,6 @@ class OutLinedTextInputView @JvmOverloads constructor(
     private fun setupPasswordMode() {
         binding.internalPasswordIcon.visibility = VISIBLE
         binding.internalEditText.inputType = EditorInfo.TYPE_TEXT_VARIATION_PASSWORD or EditorInfo.TYPE_TEXT_FLAG_MULTI_LINE
-        // This has to be set before the transformationMethod. Else, it will force the password to be visible.
```
karlenDimla, line 157: removing this forces the password to be single line even if we set the inputType just above to EditorInfo.TYPE_TEXT_FLAG_MULTI_LINE. (You'll see that if you open the edit mode in password it won't expand anymore). If that is the desired behavior, then i think we can remove ditorInfo.TYPE_TEXT_FLAG_MULTI_LINE in line 156.

karlenDimla (approved):
I have some suggestions that will be good to apply. After that, this is good to go.

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/OutLinedTextInput.kt
CDRussell: `Type.values()[inputTypeInt]` changes the behavior in an undesirable way. 
- That relies on the _index_ to do the matching, but isn't matching on the actual specified _value_. 
- The way we have the code now, both the `value` and `index` match up, so your code would run fine. but if we had to add attributes with values not in order, or skipped some, then `index` and `value` wouldn't match up.
- If somehow something doesn't match up, your suggestion throws an exception rather than nulling out and letting us `?:` for a safe default

Thinking more, it might make sense to not even specify a value explicitly and rely on ordering alone, but I'm going to stick with the existing, explicit code for now.

PR #2479, opened 2022-11-01
Fix bug where exceptions were matching full hostname instead of SLD

nshuba:
Task/Issue URL: https://app.asana.com/0/0/1203278838524404/f

### Description
Our exceptions are based on SLDs, but we were matching on FQDNs. Fix to match exceptions the same way we match domains in the blocklist.

### Steps to test this PR
- [x] Smoke tests on AppTP
- [x] Use AppTP dev settings to add some App/Tracker exception rules and verify they work as intended


aitorvs (approved):
LGTM

PR #2483, opened 2022-11-02
Improve the performance of AppTP apps screen - Adding View Types

malmstein:
Task/Issue URL: https://app.asana.com/0/inbox/1157893581871899/1203215850863402/1203215917938975

### Description
Improved performance of the list by using view types in the adapter


app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ui/ExclusionListAdapter.kt
```diff
@@ -0,0 +1,318 @@
...
+
+        private val context: Context = binding.root.context
+
+        fun bind(
+            isListEnabled: Boolean,
+            excludedAppInfo: TrackingProtectionAppInfo,
+            position: Int,
+            listener: ExclusionListListener
+        ) {
+            val appIcon = excludedAppInfo.packageIcon
```
anikiki, line 244: We need to revert this change to:
`val appIcon = itemView.context.packageManager.safeGetApplicationIcon(excludedAppInfo.packageName)`
and remove the `packageIcon` from the `TrackingProtectionAppInfo` as it invalidates the diff as it creates a new drawable object all the time. 
To check this:
Select "All", on the first item change the state of the toggle and notice the whole list is updated instead of updating only that item.

anikiki (approved):

PR #2489, opened 2022-11-04
Re-enable CPU usage alerts tests

nshuba:
Task/Issue URL: None

### Description
Fix flaky CPU usage alerts tests by adding a time buffer and re-enable them.

### Steps to test this PR
Unit test changes only - no testing needed. Tested 5 times on GitHub CI and tests passed: https://github.com/duckduckgo/Android/actions/runs/3390388374/jobs/5635459177


app-tracking-protection/vpn-impl/src/test/java/com/duckduckgo/mobile/android/vpn/health/AppTPCPUMonitorTest.kt
```diff
@@ -180,7 +180,7 @@ class AppTPCPUMonitorTest {
 
         assertStartWorker()
 
-        verify(mockDeviceShieldPixels).sendCPUUsageAlert(eq(expectedAlert))
+        verify(mockDeviceShieldPixels, timeout(1000)).sendCPUUsageAlert(eq(expectedAlert))
```
aitorvs, line 183: Do we know why we need this

nshuba: Can't be 100% sure as I still could not repro locally. My guess is that it just takes a bit of delay to spin up the worker.

aitorvs: Not sure you really need to start the worker. Can't you just create the `CPUMonitorWorker`, call on `doWork` and then `verify`?

asserting that the worker starts has very limited value, but the 1s delay in every test is not really desirable.

nshuba: It's been a while, but I think the reason I didn't use `CPUMonitorWorker` directly is because then we'd need to provide `WorkerParameters`, which has a restricted constructor. Working around that seemed like an overkill based on [this StrackOverflow thread](https://stackoverflow.com/questions/52422361/how-can-create-workerparameters).

I agree that a delay is not great, but I think that in most cases, the `verify` will trigger much earlier than 1s.

Let me know if I've missed something about creating a `CPUMonitorWorker`!

malmstein:
@nshuba @aitorvs  is this still valid? are you planning on working on this?

PR #2493, opened 2022-11-04
CT7–Crash after restoring default protection

anikiki:
Task/Issue URL: https://app.asana.com/0/0/1203286669056489/f

### Description
Updated the way we restart the vpn service in an attempt to reduce the `android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException` exceptions.

### Steps to test this PR

**Note**: it's very difficult to reproduce the crash by using the app. See more details in [Asana ](https://app.asana.com/0/0/1203286669056489/f)

- [x] Fresh install from this branch.
- [x] Enable AppTP from Settings.
- [x] Go to `All Apps` screen (`App Tracking Protection` screen -> tap on `Please Report an issue` -> `Recent Apps` screen -> tap on `Show All Apps`) and disable a large number of apps,  then tap on the overflow menu and choose `Restore Default Protected Apps`.
- [x] You should see how the vpn service restarts (look at the key icon in the status bar).
- [x] The app should not crash.

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ui/TrackingProtectionExclusionListActivity.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ui/TrackingProtectionExclusionListActivity.kt
@@ -264,7 +264,6 @@ class TrackingProtectionExclusionListActivity :
 
     override fun onDefaultProtectionRestored() {
         viewModel.restoreProtectedApps()
-        restartVpn()
         Snackbar.make(shimmerLayout, getString(R.string.atp_ExcludeAppsRestoreDefaultSnackbar), Snackbar.LENGTH_LONG)
             .show()
     }
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/heartbeat/VpnServiceHeartbeat.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/heartbeat/VpnServiceHeartbeat.kt
@@ -24,6 +24,7 @@ import com.duckduckgo.di.scopes.VpnScope
 import com.duckduckgo.mobile.android.vpn.service.VpnServiceCallbacks
 import com.duckduckgo.mobile.android.vpn.state.VpnStateMonitor.VpnStopReason
 import com.duckduckgo.mobile.android.vpn.state.VpnStateMonitor.VpnStopReason.ERROR
+import com.duckduckgo.mobile.android.vpn.state.VpnStateMonitor.VpnStopReason.RESTART
 import com.duckduckgo.mobile.android.vpn.state.VpnStateMonitor.VpnStopReason.REVOKED
 import com.duckduckgo.mobile.android.vpn.state.VpnStateMonitor.VpnStopReason.SELF_STOP
 import com.duckduckgo.mobile.android.vpn.state.VpnStateMonitor.VpnStopReason.UNKNOWN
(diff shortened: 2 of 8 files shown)


aitorvs (commented):
I know it is in draft...QQ why not just killing the process and let the `VpnStateMonitorService` kick in ? Did you find that way less reliable?

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ui/TrackingProtectionExclusionListActivity.kt
```diff
@@ -264,7 +264,6 @@ class TrackingProtectionExclusionListActivity :
 
     override fun onDefaultProtectionRestored() {
         viewModel.restoreProtectedApps()
-        restartVpn()
```
anikiki, line 267: Removed the `restartVpn()` call as `viewModel.restoreProtectedApps()` already restarts the vpn.

anikiki:
> I know it is in draft...QQ why not just killing the process and let the `VpnStateMonitorService` kick in ? Did you find that way less reliable?

I tried both ways, but I'm more confident attempting a start in `onDestroy` as well. I added more into in the Asana task. If we see a decrease in the number of crashes I'm inclined to remove it an let only `VpnStateMonitorService` handle this and see what happens.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
```diff
@@ -466,7 +478,7 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope() {
             TRIM_MEMORY_RUNNING_MODERATE -> deviceShieldPixels.vpnMemoryRunningModerate(memoryData)
             TRIM_MEMORY_RUNNING_LOW -> deviceShieldPixels.vpnMemoryRunningLow(memoryData)
             TRIM_MEMORY_RUNNING_CRITICAL -> deviceShieldPixels.vpnMemoryRunningCritical(memoryData)
-            else -> null /* noop */
+            else -> {} // no-op
```
anikiki, line 481: Just removed the `null` as it wasn't needed.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldReminderNotificationScheduler.kt
```diff
@@ -67,6 +67,7 @@ class DeviceShieldReminderNotificationScheduler @Inject constructor(
         vpnStopReason: VpnStopReason
     ) {
         when (vpnStopReason) {
+            VpnStopReason.RESTART -> {} // no-op
```
anikiki, line 70: I don't think we need to do anything here when we do a restart, let me know if it's not the case.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/heartbeat/VpnServiceHeartbeat.kt
```diff
@@ -74,6 +75,7 @@ class VpnServiceHeartbeat @Inject constructor(
         Timber.v("onVpnStopped called")
         when (vpnStopReason) {
             ERROR -> Timber.v("HB monitor: sudden vpn stopped $vpnStopReason")
+            RESTART -> Timber.v("HB monitor: restarting vpn $vpnStopReason")
```
anikiki, line 78: I don't think we need to do anything here when we do a restart, let me know if it's not the case.

aitorvs (approved):
Awesome work! works as expected

PR #2538, opened 2022-11-18
Enable passing release number and release notes through params/options for release lane

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203398448486630/f

### Description
This PR allows passing the following as options for the release lane:
- release_notes -> Actual custom release notes
- notes_type -> "KEEP EXISTING", "CUSTOM" and then others
- release_number -> new release version number

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -154,79 +154,133 @@ platform :android do
     # This dependency could be removed with a little more time to tidy up this script to do the branching/merging manually.
 
     desc "Create new release"
-    lane :release do
+    lane :release do |options|
 
         ensure_git_status_clean
         ensure_git_branch( branch: 'develop' )
 
-        newVersion = determine_version_number()
-        releaseNotes = determine_release_notes()
+        options_release_number = options[:release_number]
+        options_release_notes = options[:release_notes]
+        options_notes_type = options[:notes_type]
 
-        if UI.confirm("Are you sure you're happy with this release?\n\nVersion=#{newVersion}\nRelease Notes:\n#{releaseNotes}\n")
-          UI.success "Creating release branch for release/#{newVersion}"
+        newVersion = determine_version_number(
+            release_number: options_release_number
+        )
+        releaseNotes = determine_release_notes(
+            release_notes: options_release_notes,
+            notes_type: options_notes_type
+        )
 
-          sh "git flow release start #{newVersion}"
+        isInteractiveMode = options_release_number == nil || (options_notes_type == nil && options_release_notes == nil) || (options_notes_type == "CUSTOM" && options_release_notes == nil)
 
-          File.open('../app/version/version.properties', 'w') do |file|
-            file.write("VERSION=#{newVersion}")
-          end
+        # For interactive flows, we want to confirm the data inputted by the user before proceeding
+        if !isInteractiveMode
+            UI.message("This are the release information:\n\nVersion=#{newVersion}\nRelease Notes:\n#{releaseNotes}\n")
+            do_create_release_branch(newVersion: newVersion, releaseNotes: releaseNotes)
+            do_create_and_push_tags(newVersion: newVersion)
+        else
+            if UI.confirm("Are you sure you're happy with this release?\n\nVersion=#{newVersion}\nRelease Notes:\n#{releaseNotes}\n")
+                UI.success "Creating release branch for release/#{newVersion}"
+                do_create_release_branch(newVersion: newVersion, releaseNotes: releaseNotes)
+                if UI.confirm(text:"If you have any other changes to make to the release branch, do them now. Enter `y` when ready to create and push tags")
+                    do_create_and_push_tags(newVersion: newVersion)
+                else
+                    UI.error errorMessageCancelled
+                end
+            else
+                UI.error errorMessageCancelled
+            end
+        end
+    end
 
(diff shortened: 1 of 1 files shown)


karlenDimla:
* develop
  * **PR #2538** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2538" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2538?utm_source=stack-comment).

fastlane/Fastfile
```diff
@@ -154,13 +154,18 @@ platform :android do
     # This dependency could be removed with a little more time to tidy up this script to do the branching/merging manually.
 
     desc "Create new release"
-    lane :release do
+    lane :release do |option|
```
CDRussell, line 157:
```suggestion
    lane :release do |options|
```

fastlane/Fastfile
```diff
@@ -202,31 +207,52 @@ platform :android do
         end
 
 
+        private_lane :determine_release_notes do |options|
+            notes_type = options[:notes_type]
+            custom_release_notes = options[:release_notes]
```
CDRussell, line 212: if you provide custom release notes, it would be nice if that automatically set the `notes_type` to custom

fastlane/Fastfile
```diff
@@ -154,13 +154,18 @@ platform :android do
...
         ensure_git_status_clean
         ensure_git_branch( branch: 'develop' )
 
-        newVersion = determine_version_number()
-        releaseNotes = determine_release_notes()
+        newVersion = determine_version_number(
+            release_number: options[:release_number]
+        )
+        releaseNotes = determine_release_notes(
+            release_notes: options[:release_notes],
```
CDRussell, line 166: Can you provide release notes which are multi-line using this mechanism?

CDRussell (changes requested):

fastlane/Fastfile
CDRussell: This change is required, otherwise running `bundle exec fastlane release` will break complaining about it (as it tries to access non-existent option**s**

karlenDimla: You are right. I mistyped this one. Good catch!

fastlane/Fastfile
karlenDimla: At least from jenkins, I tried to set mutline params and passed multiline values and it works as expected.

fastlane/Fastfile
```diff
@@ -154,79 +154,132 @@ platform :android do
...
+        )
+        releaseNotes = determine_release_notes(
+            release_notes: options_release_notes,
+            notes_type: options_notes_type
+        )
+
+        # We will prompt user input if no release_number is passed OR no notes_type is passed.
+        hasPromptedForUserInput = options_release_number == nil || (options_notes_type == nil && options_notes_type == nil)
+
+        if !hasPromptedForUserInput
```
karlenDimla, line 177: @CDRussell I had to do additional changes to bypass this confirmation when params are passed.

fastlane/Fastfile
```diff
@@ -154,79 +154,133 @@ platform :android do
...
+        newVersion = determine_version_number(
+            release_number: options_release_number
+        )
+        releaseNotes = determine_release_notes(
+            release_notes: options_release_notes,
+            notes_type: options_notes_type
+        )
+
+        # We will prompt user input if no release_number is passed OR no notes_type is passed.
+        hasPromptedForUserInput = options_release_number == nil || (options_notes_type == nil && options_notes_type == nil)
```
CDRussell, line 175: `(options_notes_type == nil && options_notes_type == nil)`

should one of those be `options_release_notes`?

fastlane/Fastfile
```diff
@@ -154,79 +154,133 @@ platform :android do
...
-        end
-
         end
 
+        private_lane :determine_release_notes do |options|
+            notes_type = options[:notes_type]
+            custom_release_notes = options[:release_notes]
+            existingReleaseNotes = File.read(releaseNotesFileBody)
+
+            # This maps to the old flow when no options are being passed to determine_release_notes
```
CDRussell, line 232: "old flow" is perhaps misleading; i think it's less of an old vs new type scenario, and more that we want to support both options.

CDRussell:
@karlenDimla gave it a quick scan and suggested a change. haven't had time to do a thorough review; will do that today or tomorrow if that's cool. LMK if you need it reviewed sooner.

karlenDimla:
@CDRussell Only got time for this today. I tested it as much as I can here https://github.com/karlenDimla/FastLaneTest-Android/tree/main. You can see the new commits (but don't mean much)

CDRussell (approved):
This is working great! 🎉

Summarizing what we discussed on the call, optional suggestions:

- rename `hasPromptedForUserInput` to `interactiveMode` or such
- ensure it is either fully interactive, or fully non-interactive (don't allow it to prompt for something, but then not offer a "are you sure" check later)

fastlane/Fastfile
CDRussell: For posterity, this is also fine from local command line.

`fastlane release release_notes:"`  would be enough to trigger the multi-line input, and ending with `"` would allow the multi-line release notes to be added.

Thanks for the suggestion @karlenDimla.

PR #2546, opened 2022-11-22
ADS DaxTextView: Support for android:textColor

malmstein:
Task/Issue URL: https://app.asana.com/0/1202857801505092/1203418290512953

### Description

- Support the textColor attribute from Android
- When textType and textColor are both set, textType will take priority.

### DIFF
--- a/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextView.kt
+++ b/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextView.kt
@@ -20,7 +20,7 @@ import android.content.Context
 import android.util.AttributeSet
 import androidx.core.content.ContextCompat
 import com.duckduckgo.mobile.android.R
-import com.duckduckgo.mobile.android.ui.view.getColorFromAttr
+import com.duckduckgo.mobile.android.ui.view.TypedArrayUtils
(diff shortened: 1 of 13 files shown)


common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextView.kt
```diff
@@ -40,33 +40,46 @@ constructor(
                 0,
             )
 
-        val typographyType = if (typedArray.hasValue(R.styleable.DaxTextView_typography)) {
-            Type.from(typedArray.getInt(R.styleable.DaxTextView_typography, 0))
+        val typographyTypography = if (typedArray.hasValue(R.styleable.DaxTextView_typography)) {
```
nalcalag, line 43: I think there is a typo in the variable name

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextView.kt
```diff
@@ -114,22 +127,22 @@ constructor(
         }
     }
 
-    enum class TextColor {
+    enum class Type {
```
nalcalag, line 130: `TextType`?

malmstein:
@nalcalag thanks for the review, I've added a missing lint rule too

nalcalag (approved):
Works as expected 💯

PR #2604, opened 2022-12-06
Move unnecessary public APIs to vpm-impl

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203491694304847/f

### Description
Some public APIs are no longer used and should not be in `vpn-api` module.

Move them inside `vpn-impl` so they remain private

### Steps to test this PR
Smoke AppTP tests


karlenDimla (approved):
LGTM.

Just a few non-blocking questions:
- VpnDetector -> seems like this will be useful for ANY VPN feature, not only AppTP.
- VpnSharedPreferencesProvider -> seems also useful to get SharedPrefs for any VPN feature.
- Not so sure about the SurveyDownloader changes. I am not sure I have anought context why we are changing that.

PR #2608, opened 2022-12-07
Bypass auth prompt when no credentials stored

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203320214644536/f

### Description
Allows access to the `AutofillManagementActivity` without authenticating when there are no saved logins. So the decision tree on what happens when you access that activity is now:

```mermaid
flowchart LR
    start(Launch activity) --> deviceAuth{Has valid device auth}
   deviceAuth-- No ---showDisabled(((Disabled state)))
   deviceAuth-- Yes ---hasLogins{>= 1 logins}
   hasLogins-- No ---showPlaceholder(((Placeholder state, no auth)))
   hasLogins-- Yes ---gracePeriod{In auth grace period}
   gracePeriod-- Yes ---showLogins(((Show logins)))
   gracePeriod-- No ---authenticate
   authenticate{Authentication challenge}
   authenticate-- Success ---showLogins
   authenticate-- Failure/cancel ---exitActivity(((Exit activity)))
```

#### Device auth disabled

- [x] Set device auth to disabled by removing all PIN/Password/Fingerprint protection
- [x] Access the cred management activity
- [x] Verify you see the `disabled state`

#### No logins stored

- [x] Enable device auth
- [x] Enter creds management activity
- [x] Verify you are not prompted to authenticate
- [x] Verify you can see the toggle to enable/disable 
- [x] Verify you see the empty state placeholder

#### Logins stored

- [x] Store a login credential
- [x] Enter creds management activity
- [x] Verify you are prompted to authenticate
- [x] Cancel; verify you are exited from the activity
- [x] Re-enter, and authenticate
- [x] Verify you can see the credential list

#### Grace period
- [x] Re-enter the creds management screen within 15s of last successful authentication
- [x] Verify you are not prompted to authenticate
- [x] Verify you can see the credential list
- [x] Wait > 15s and re-enter; verify you are now prompted to authenticate again


CDRussell:
* develop
  * **PR #2608** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2608" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2608?utm_source=stack-comment).

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillSettingsViewModel.kt
```diff
@@ -225,7 +243,7 @@ class AutofillSettingsViewModel @Inject constructor(
 
     data class ViewState(
         val autofillEnabled: Boolean = true,
-        val logins: List<LoginCredentials> = emptyList(),
+        val logins: List<LoginCredentials>? = null,
```
CDRussell, line 246: Because of the new "empty state" placeholder, it's important to be able to distinguish between "we have no saved credentials" and "we don't know yet how many there are".

Making this `nullable` supports the case whereby other properties of the view state have been initialised before we have the list of credentials.

CDRussell:
cc @THISISDINOSAUR in case the diagram is useful to compare against your thinking on how it should flow

autofill/autofill-impl/src/main/AndroidManifest.xml
```diff
@@ -21,7 +21,7 @@
         <activity
             android:name="com.duckduckgo.autofill.ui.credential.management.AutofillManagementActivity"
             android:configChanges="orientation|screenSize"
-            android:windowSoftInputMode="stateVisible|adjustResize"
```
CDRussell, line 24: I'm unsure why we had this here; it looks to have been added for editing credentials but I see that working fine without it. 

I'm removing it because it causes the keyboard to automatically show when entering the credential list on some devices (where it doesn't make sense to present a keyboard)

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillManagementActivity.kt
```diff
@@ -107,18 +109,14 @@ class AutofillManagementActivity : DuckDuckGoActivity() {
     }
 
     private fun launchDeviceAuth() {
-        if (deviceAuthenticator.hasValidDeviceAuthentication()) {
```
CDRussell, line 110: Moving this check to the `ViewModel` instead of the `Activity`

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementListMode.kt
```diff
@@ -100,7 +102,9 @@ class AutofillManagementListMode : DuckDuckGoFragment(R.layout.fragment_autofill
             viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
                 viewModel.viewState.collect { state ->
                     binding.enabledToggle.quietlySetIsChecked(state.autofillEnabled, globalAutofillToggleListener)
-                    credentialsListUpdated(state.logins)
+                    state.logins?.let {
```
CDRussell, line 105: if `state.logins` is null here, we still don't know if we have credentials or not, so ignore it until we do know for sure

autofill/autofill-impl/src/main/AndroidManifest.xml
karlenDimla: You are right. This seems to be not necessary.

karlenDimla: Without it, the previously reported issue seems to be fixed still.

PR #2624, opened 2022-12-10
Add TDS type exception support

joshliebe:
Task/Issue URL: https://app.asana.com/0/0/1203771742633842/f

### Description
- Detects the MIME type from the `Accept`/`X-Requested-With` headers in requests in order to filter them out in TDS.
- Falls back to to using `MimeTypeMap` if we can't detect the type from the headers.

### Steps to test this PR
- Smoke test and code review.


app/src/androidTest/resources/reference_tests/domain_matching_tests.json
```diff
@@ -53,10 +53,7 @@
                 "siteURL": "https://ignore.test/",
                 "requestURL": "https://bad.third-party.site/",
                 "requestType": "script",
-                "expectAction": "block",
-                "exceptPlatforms": [
-                    "android-browser"
-                ]
+                "expectAction": "block"
```
marcosholgado, line 56: Remember to open a PR against the ref tests repo :)

app/src/main/java/com/duckduckgo/app/trackerdetection/UrlToTypeMapper.kt
```diff
@@ -0,0 +1,87 @@
...
+
+interface UrlToTypeMapper {
+    fun map(
+        url: String,
+        requestHeaders: Map<String, String>,
+    ): String?
+}
+
+@ContributesBinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
```
marcosholgado, line 33: I don't think we need a single instance here because there is no state in this class.

marcosholgado (approved):
LGTM!

app/src/androidTest/resources/reference_tests/domain_matching_tests.json
joshliebe: PR opened: https://github.com/duckduckgo/privacy-reference-tests/pull/73

PR #2612, opened 2022-12-08
Translations for autofill's new auth flow

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203508506791934/f

### Description
Translations for autofill empty list placeholder text

### Steps to test this PR

- [ ] Visit credential management screen when there are no saved logins
- [ ] Verify text is correct

PR #2631, opened 2022-12-13
Add ability to manually create logins from credential management screen.

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203522700191989/f

### Description
Adds ability to manually create logins from the credential management screen.

#### From empty login state
- [x] Delete all existing logins
- [x] Visit credential management screen; verify not prompted to authenticate
- [x] Click the ➕ button in the action bar
- [x] Verify you see the "manual creation" screen, and that all fields are editable
- [x] Verify that the ✔️ button is **not** enabled before you've entered any data
- [x] Add some text to a field, verify the ✔️ appears. Remove the text and verify it disappears again.
- [x] Repeat the last step for each individual text field, verifying that the ✔️ only appears when some data has been entered.
- [x] Enter some details across the fields and tap the ✔️ button to save
- [x] Verify you then see the "view mode" for the newly added credential
- [x] Navigate back; verify that you are not prompted to auth, and that you can see the newly created cred in the list view

#### Navigating back mid-creation
- [x] Start manually adding a credential, adding some data to the fields
- [x] Instead of saving, hit the back button to navigate away
- [x] Verify login is discarded (no confirmation prompt)
- [x] Verify the back button returns to being a left arrow and not ✖️

#### Backgrounding the app mid-creation
- [x] Start manually adding a credential, adding some data to the fields
- [x] Instead of saving, hit the home button
- [x] Return to the app; verify you are prompted to auth. 
- [x] Verify that successful auth returns you to the screen you were editing with data intact. (auth failure kicks you out and discards your mid-creation login)
- [x] Verify the back button is an ✖️ and not a left arrow

##### Disabled mode overlays list view
- [x] Open list mode
- [x] Open credential
- [x] Enter edit mode
- [x] Background app
- [x] Remove device auth
- [x] Go back to app
- [x] Verify disabled mode is rendered correctly (e.g., not as an overlay on the list)

##### Navigating back mid-creation
- updated test case **Navigating back mid-creation** with a new check

##### Backgrounding the app mid-creation
- updated test case **Backgrounding the app mid-creation** with a new check

##### On edit text change watcher not working when directly jumped into edit mode

- [x] From management list screen, use overflow to choose to edit
- [x] Verify the save button (✔️) is hidden by default
- [x] Verify it appears when you edit any data, and disappears when you restore the text to what it was originally
- [x] Repeat the above, but from navigating from list -> credential -> edit mode and verify it still works as expected

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillManagementActivity.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillManagementActivity.kt
@@ -40,11 +40,15 @@ import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewMode
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.Command.LaunchDeviceAuth
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.Command.ShowCredentialMode
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.Command.ShowDisabledMode
+import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.Command.ShowListMode
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.Command.ShowLockedMode
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.Command.ShowUserPasswordCopied
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.Command.ShowUserUsernameCopied
+import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.CredentialMode.Disabled
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.CredentialMode.EditingExisting
 import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.CredentialMode.EditingNewEntry
+import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.CredentialMode.ListMode
+import com.duckduckgo.autofill.ui.credential.management.AutofillSettingsViewModel.CredentialMode.Locked
 import com.duckduckgo.autofill.ui.credential.management.viewing.AutofillManagementCredentialsMode
 import com.duckduckgo.autofill.ui.credential.management.viewing.AutofillManagementDisabledMode
 import com.duckduckgo.autofill.ui.credential.management.viewing.AutofillManagementListMode
@@ -105,7 +109,7 @@ class AutofillManagementActivity : DuckDuckGoActivity() {
                 viewModel.onViewCredentials(it, true)
             }
         } else {
-            showListMode()
+            viewModel.onShowListMode()
         }
     }
 
@@ -149,6 +153,7 @@ class AutofillManagementActivity : DuckDuckGoActivity() {
             is ShowCredentialMode -> showCredentialMode(command.isLaunchedDirectly)
             is ShowUserUsernameCopied -> showCopiedToClipboardSnackbar("Username")
             is ShowUserPasswordCopied -> showCopiedToClipboardSnackbar("Password")
+            is ShowListMode -> showListMode()
             is ShowDisabledMode -> showDisabledMode()
             is ShowLockedMode -> showLockMode()
             is LaunchDeviceAuth -> launchDeviceAuth()
@@ -193,7 +198,10 @@ class AutofillManagementActivity : DuckDuckGoActivity() {
         if (credentialModeLaunchedDirectly()) {
             finish()
         } else {
-            supportFragmentManager.popBackStack()
+            resetToolbar()
+            supportFragmentManager.commit {
+                replace(R.id.fragment_container_view, AutofillManagementListMode.instance(), TAG_ALL_CREDENTIALS)
+            }
         }
     }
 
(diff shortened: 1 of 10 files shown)


CDRussell:
* develop
  * **PR #2608** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2608" target="_blank">Graphite</a> 
    * **PR #2631** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2631" target="_blank">Graphite</a>  👈
    * **PR #2612** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2612" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2631?utm_source=stack-comment).

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillManagementActivity.kt
```diff
@@ -250,7 +246,8 @@ class AutofillManagementActivity : DuckDuckGoActivity() {
 
     override fun onBackPressed() {
         when (viewModel.viewState.value.credentialMode) {
-            is Editing -> viewModel.onCancelEditMode()
+            is EditingExisting -> viewModel.onCancelEditMode()
+            is EditingNewEntry -> viewModel.onCancelManualCreation()
```
CDRussell, line 250: `Editing` is now an abstract class; the concrete modes are now `EditingExisting` and `EditingNewEntry` which are both types of `Editing` modes

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/AutofillSettingsViewModel.kt
```diff
@@ -62,6 +67,9 @@ class AutofillSettingsViewModel @Inject constructor(
     private val _commands = MutableStateFlow<List<Command>>(emptyList())
     val commands: StateFlow<List<Command>> = _commands
 
+    private val _commandsCredentialView = MutableStateFlow<List<CredentialModeCommand>>(emptyList())
```
CDRussell, line 70: I found it necessary to differentiate between the existing `commands` that the `AutofillManagementActivity` might consume vs commands that were expected to be consumed by the fragment only. 

Re-using the same commands is error prone, as both activity and fragment could try to consume it but only one would get it. Adding specific commands that only the credential view fragment would consume solves this. (I don't _love_ the approach, but it works)

karlenDimla:
I found a few small bugs:
- In the second test scenario above when navigating back from edit page, the back button remains an x instead of left arrow.
- In the third test scanario above when returning to the app after pressing home, the back button becomes a left arrow instead of x (which is the expected back button icon whenever in edit mode screen).

These are probaly bugs caused in the previous refactoring that I missed to notice when testing last time.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
```diff
@@ -125,13 +137,27 @@ class AutofillManagementCredentialsMode : DuckDuckGoFragment(R.layout.fragment_a
...
+    private fun startEditTextWatchers() {
+        val initialState = binding.currentTextState()
+        viewModel.allowSaveInEditMode(false)
+        binding.watchSaveState(saveStateWatcher) {
+            val currentState = binding.currentTextState()
+
+            val changed = currentState != initialState
+            val empty = currentState.isEmpty()
+
+            viewModel.allowSaveInEditMode(!empty && changed)
```
karlenDimla, line 153: Is it necessary to check if the data has changed? From what I see, we only hide the check icon when all fields are empty otherwise we show it. Even on edit mode, we show it right away even if there were no changes.

karlenDimla: Ah okay, now I noticed that the changed is only correctly checked when:
- List mode -> click on item -> credential mode -> edit.

However, the check is not taking effect for:
- List mode -> context menu on item -> edit

CDRussell:
@karlenDimla i've updated the PR description with additional checks which encompass the bugs you found before.

karlenDimla:
Found another bug:
This scenario is broken
- Save new credentials by opening the site and saving login
- Click VIEW in the snackbar

Expected:
- Open view mode of credential

Actual:
- Empty view mode where credentials are not prefilled.

karlenDimla:
@CDRussell reviewed this an it's almost there. The only minor bug I found is that in the scenario mentioned here https://github.com/duckduckgo/Android/pull/2631#issuecomment-1355218911 everything is there except for the context menu allowing the user to edit/delete the credential.

CDRussell:
> @CDRussell reviewed this an it's almost there. The only minor bug I found is that in the scenario mentioned here [#2631 (comment)](https://github.com/duckduckgo/Android/pull/2631#issuecomment-1355218911) everything is there except for the context menu allowing the user to edit/delete the credential.
> 
> Debugging: seems like in onPrepareMenu of AutofillManagementCredentialsMode we get ListMode as viewModel.viewState.value.credentialMode.

Sorry, wasn't ready for the re-review yet as I'm still making changes. Will see if this is still an issue before tagging for re-review.

karlenDimla (approved):
Works as expected in all scenarios I know! 🥳

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
```diff
@@ -224,25 +256,39 @@ class AutofillManagementCredentialsMode : DuckDuckGoFragment(R.layout.fragment_a
...
+        }
+
         viewModel.viewState
             .flowWithLifecycle(viewLifecycleOwner.lifecycle, Lifecycle.State.STARTED)
             .onEach { state ->
                 when (state.credentialMode) {
                     is Viewing -> {
                         populateFields(state.credentialMode.credentialsViewed)
                         showViewMode(state.credentialMode.credentialsViewed)
+                        invalidateMenu()
```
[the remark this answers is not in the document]
CDRussell: Will address this in the bigger refactor of all of this (to avoid triggering a further round of testing of all the scenarios now)

PR #2637, opened 2022-12-13
Avoid starting VPN twice

nshuba:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203529428649444/f

#### Background
Users report that AppTP causes all protected apps to lose connectivity. Reproducing this issue in-house revealed that sometimes two TUN interfaces are created. Upon further inspection, we found the root cause to be the system calling `onStartCommand` twice (even though we call `startForegroundService` once).

#### Fix
The fix here is to detect when we are already in the middle of creating a VPN and not start creating another one.

### Steps to test this PR
- [x] Smoke tests for AppTP
- [x] Attempt to repro the bug: 
  1. Open an app and then force close it
  2. Disable protection for it
  3. Re-enable protection
  4. Rinse and repeat ~10 times
- [x] Check logs for TUN errors - none should appear
- [x] Check logs for the "VPN is already being started, abort" message - it should appear, meaning that we reached a condition where `onStartCommand` was called twice and the fix is working

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
@@ -98,6 +98,8 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope() {
 
     private var restartRequested = false
 
+    private val startVpnLock = Object()
+
     private val isInterceptDnsTrafficEnabled by lazy {
         appTpFeatureConfig.isEnabled(AppTpSetting.InterceptDnsTraffic)
     }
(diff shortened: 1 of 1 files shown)


app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
```diff
@@ -222,8 +224,18 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope() {
...
         logcat { "VPN log: Starting VPN" }
 
-        // We need to rethink how to log this state. This will likely change.
-        vpnServiceStateStatsDao.insert(VpnServiceStateStats(state = ENABLING))
+        synchronized(startVpnLock) {
+            val currStateStats = vpnServiceStateStatsDao.getLastStateStats()
+            if (currStateStats?.state == ENABLING) {
+                // Sometimes onStartCommand gets called twice - this is a safety rail against that
+                logcat(LogPriority.WARN) { "VPN is already being started, abort" }
+                return@withContext
```
aitorvs, line 232: Can we add a pixel here to get a sense how often this was happening?
As this is surely source for breakage when it happens, it'd be good to know how often it actually happened (?)

aitorvs (approved):

PR #2672, opened 2022-12-21
Fix setting state back to DISABLED when vpn fails to start

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203586092184943/f

### Description
This PR includes resetting the VPN state to DISABLED when any error occurs before the VPN successfully starts.

### Steps to test this PR

Smoke test AppTP

_Attempt repro 2 tun interface bug_
- [x] Open an app and then force close it
- [x] Disable protection for it
- [x] Re-enable protection
- [x] Rinse and repeat ~10 times
- [x] See "VPN is already being started, abort" in the logs

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
@@ -37,6 +37,7 @@ import com.duckduckgo.mobile.android.vpn.dao.VpnServiceStateStatsDao
 import com.duckduckgo.mobile.android.vpn.feature.AppTpFeatureConfig
 import com.duckduckgo.mobile.android.vpn.feature.AppTpSetting
 import com.duckduckgo.mobile.android.vpn.integration.VpnNetworkStackProvider
+import com.duckduckgo.mobile.android.vpn.model.VpnServiceState.ENABLED
 import com.duckduckgo.mobile.android.vpn.model.VpnServiceState.ENABLING
 import com.duckduckgo.mobile.android.vpn.network.VpnNetworkStack.VpnTunnelConfig
 import com.duckduckgo.mobile.android.vpn.network.util.asRoute
@@ -225,7 +226,7 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope() {
 
         synchronized(startVpnLock) {
             val currStateStats = vpnServiceStateStatsDao.getLastStateStats()
-            if (currStateStats?.state == ENABLING) {
+            if (currStateStats?.state == ENABLING || currStateStats?.state == ENABLED) {
                 // Sometimes onStartCommand gets called twice - this is a safety rail against that
                 logcat(LogPriority.WARN) { "VPN is already being started, abort" }
                 return@withContext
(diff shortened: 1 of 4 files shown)


karlenDimla:
* develop
  * **PR #2672** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2672" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2672?utm_source=stack-comment).

aitorvs:
LGTM! I manage to repro the issue with two VPN tun(s) created and we still handle that well.

@nshuba, can you take a second pass here, given that this code is pretty critical

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
```diff
@@ -232,8 +231,10 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope() {
                 return@withContext
             }
 
-            // We need to rethink how to log this state. This will likely change.
-            vpnServiceStateStatsDao.insert(VpnServiceStateStats(state = ENABLING))
+            vpnServiceCallbacksPluginPoint.getPlugins().forEach {
+                logcat { "VPN log: onVpnStarting ${it.javaClass} callback" }
+                it.onVpnStarting(this)
```
nshuba, line 236: Since `it.onVpnStarting(this)` now changes the state within a coroutine, isn't it now possible that another thread enters this critical section and the sate hasn't yet changed to `ENABLING`? It's probably ok for now, but I think it's still a race condition and we need to further re-think how we keep track of states.

aitorvs: In principle I think that was also the case before. `Dispatchers.IO` is backed by a thread pool. 
But yeah, I think we need to about, also because we should not use `Dispatchers.IO` anymore.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/stats/VpnServiceStateLogger.kt
```diff
@@ -60,7 +60,20 @@ class VpnServiceStateLogger @Inject constructor(
     private val dispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
     private val job = ConflatedJob()
 
+    override fun onVpnStarting(coroutineScope: CoroutineScope) {
+        job += coroutineScope.launch(dispatcher) {
+            logcat { "VpnServiceStateLogger, new state ENABLING" }
+            vpnDatabase.vpnServiceStateDao().insert(VpnServiceStateStats(state = VpnServiceState.ENABLING))
+        }
```
nshuba, line 67: Synced with @aitorvs and we agreed that this change will fix the race condition I mentioned above:
```suggestion
        logcat { "VpnServiceStateLogger, new state ENABLING" }
        vpnDatabase.vpnServiceStateDao().insert(VpnServiceStateStats(state = VpnServiceState.ENABLING))
```

This way, we don't switch threads when starting the VPN.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/TrackerBlockingVpnService.kt
```diff
@@ -232,8 +231,10 @@ class TrackerBlockingVpnService : VpnService(), CoroutineScope by MainScope() {
                 return@withContext
             }
 
-            // We need to rethink how to log this state. This will likely change.
```
nshuba, line 235: Looks like I can't add suggestions that are too far from code changes, but here is another issue that @aitorvs and I uncovered: on line 228 - we need to change `if (currStateStats?.state == ENABLING) {` to `if (currStateStats?.state == ENABLING || currStateStats?.state == ENABLED) {`. This is an improvement on my change in https://github.com/duckduckgo/Android/pull/2637.

nshuba (commented):
Added some suggestions after syncing with @aitorvs and tested with the changes applied. Reproduced the two tuns bug and verified that things work as expected.

karlenDimla:
@aitorvs @nshuba Thanks for the review and good finds! I agree with the points you mentioned and I've updated the PR.

aitorvs (approved):
LGTM

PR #2673, opened 2022-12-21
Improved parsing for data URL scheme.

anikiki:
Task/Issue URL: https://app.asana.com/0/1200905986587319/1203594326324933/f

### Description
Updated the data URI parsing.

### Steps to test this PR

- [x] Install from this branch and download a blob. I.e https://dopiaza.org/tools/datauri/examples/index.php (just smoke test example, we have tests covering this)

### DIFF
--- a/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DataUriParser.kt
+++ b/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DataUriParser.kt
@@ -26,7 +26,18 @@ import okio.ByteString.Companion.decodeBase64
 class DataUriParser @Inject constructor() {
 
     fun generate(url: String): ParseResult {
-        val result = MIME_TYPE_REGEX.find(url)
+        val offset = url.indexOf(',')
+        if (offset == -1) {
+            return ParseResult.Invalid
+        }
+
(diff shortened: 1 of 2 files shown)


malmstein (commented):
Shouldn't we have a test against this specific case?

anikiki:
> Shouldn't we have a test against this specific case?

We have tests. This is just removing the pattern matching on large data.

downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DataUriParser.kt
```diff
@@ -26,15 +26,25 @@ import okio.ByteString.Companion.decodeBase64
...
-        val result = MIME_TYPE_REGEX.find(url)
+        val offset = url.indexOf(',')
+        if (offset == -1) {
+            return ParseResult.Invalid
+        }
+
+        val data = url.substring(offset + 1)
+        if (data.isEmpty()) {
+            return ParseResult.Invalid
+        }
```
joshliebe, line 37: I think the unit test that's already there catches the first `ParseResult.Invalid` case, should we not add second test for this case?

joshliebe (approved):
Looks good and works as expected! Just added one comment.

PR #2699, opened 2023-01-09
Add a fastlane check to release notes length

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1202728801491382/f

### Description
Adds check in fastlane to ensure release notes aren't bigger than max supported length in Play Store.

The check is triggered from the following lanes:
- `deploy_playstore` (which will call `update_fastlane_release_notes`, which calls `release_notes_playstore`)
- `hotfix-start` (which calls `determine_release_notes`)
- `release` (which calls `determine_release_notes`)

This means Fastlane will fail if the release notes are too long, when you are attempting:
- to construct a release locally using `fastlane release`
- to construct a release using CI (using `fastlane release` with CI providing the release notes)
- to deploy to the Play Store from CI (using `fastlane deploy_playstore`)
- to start a new hotfix (using `fastlane hotfix-start`)

which I think covers all the bases.

### Steps to test this PR

**For ease of testing**
- Change line 231 to make the lane publicly callable `private_lane` --> `lane`

##### No parameters
- [x] Call `fastlane determine_release_notes` with no parameters. Choose option 1. Verify it finishes successfully
- [x] Call `fastlane determine_release_notes` with no parameters. Choose option 2 and enter short notes. Verify it finishes successfully
- [x] Call `fastlane determine_release_notes` with no parameters. Choose option 3. Verify it finishes successfully

##### With release notes params provided
- [x] Run `fastlane determine_release_notes release_notes:hello`. Verify it finishes successfully

#### Testing when notes are above limit

Modify line 23 to have a really small max length (e.g., 5)

##### No parameters
- [x] Call `fastlane determine_release_notes` with no parameters. Choose option 1. Verify it fails.
- [x] Call `fastlane determine_release_notes` with no parameters. Choose option 2 and longer notes than the max length. Verify it fails.
- [x] Call `fastlane determine_release_notes` with no parameters. Choose option 3. Verify it fails.

##### With release notes params provided
- [x] Run `fastlane determine_release_notes release_notes:HelloWorld`. Verify it fails

#### QA for the other changes
- I wouldn't recommend running `release` or `deploy_playstore` manually to test this out, as it will require stubbing out various lines to make it safe to call
- Instead, recommend sense-checking the changes look good

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -20,6 +20,7 @@
     releaseNotesFileHiring = "../app/version/release-notes-hiring"
     releaseNotesFileSuffix = "../app/version/release-notes-suffix"
     releaseNotesLocales= ["en-US", "en-GB", "en-CA"]
+    releaseNotesMaxLength = 500
     errorMessageCancelled = "Release cancelled 😢"
     asanaBridgeInstallationProblem = "Android Asana Release Bridge not installed or configured correctly - see https://app.asana.com/0/0/1203116937958001/f for instructions"
 
@@ -35,7 +36,10 @@ platform :android do
     releaseNotesHiring = File.read(releaseNotesFileHiring)
 
     formatted = "#{releaseNotesBodyHeader}\n#{releaseNotesBody}#{releaseNotesHiring}\n\n#{releaseNotesSuffixHeader}\n#{releaseNotesSuffix}"
+
+    validateReleaseNotes(releaseNotes: formatted)
     UI.message("\n#{formatted}")
+
     formatted
 
   end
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -265,6 +269,17 @@ platform :android do
...
 
             end
+
+            validateReleaseNotes(releaseNotes: releaseNotes)
+            releaseNotes
+        end
+
+        desc "Validates the release notes to ensure they are suitable for the Play Store"
+        private_lane :validateReleaseNotes do |options|
+            releaseNotesLength = options[:releaseNotes].length
```
CDRussell, line 279: using `"".length` might not give us the exact match the Play Store uses (500 unicode characters) but given the release notes will be English, that should be close enough in most cases

fastlane/Fastfile
```diff
@@ -265,6 +269,17 @@ platform :android do
                 end
 
             end
+
+            validateReleaseNotes(releaseNotes: releaseNotes)
```
karlenDimla, line 273: NIT: we are setting `500` to be used here. The thing is it is possible that the release notes is 500 chars but it will fail in `release_notes_playstore` since we are adding more strings to the release notes there.

I think we can keep this check since it allows us to fail early but i think the more important check is in  `release_notes_playstore`

karlenDimla (approved):
I just have one nit.

Overall, I think this works as fastlane will fail if the release notes is greater than 500 before we actually deploy to play store which is what exactly we are trying to solve.

PR #2708, opened 2023-01-11
Remove AppTP waitlist notification channel

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203701902138004/f

### Description
Remove the unused AppTP waitlist notification channel

### Steps to test this PR

_Before_
- [ ] install application from develop
- [ ] long press on app icon and go to App info -> notifications
- [ ] "app Tracking Protection waitlist" notification channel should appear

_After_
- [ ] install application from develop
- [ ] long press on app icon and go to App info -> notifications
- [ ] "app Tracking Protection waitlist" notification channel should NOT appear

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/notification/NotificationRegistrar.kt
+++ b/app/src/main/java/com/duckduckgo/app/notification/NotificationRegistrar.kt
@@ -37,7 +37,6 @@ import com.duckduckgo.app.settings.db.SettingsDataStore
 import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.appbuildconfig.api.AppBuildConfig
 import com.duckduckgo.di.scopes.AppScope
-import com.duckduckgo.mobile.android.vpn.R as VpnR
 import com.squareup.anvil.annotations.ContributesMultibinding
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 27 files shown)


app/src/main/java/com/duckduckgo/app/notification/NotificationRegistrar.kt
```diff
@@ -65,7 +64,6 @@ class NotificationRegistrar @Inject constructor(
         const val PrivacyProtection = 101
         const val Article = 103 // 102 was used for the search notification hence using 103 moving forward
         const val AppFeature = 104
-        const val EmailWaitlist = 106 // 105 was used for the UOA notification
```
karlenDimla, line 68: @aitorvs QQ: Does this mean that going forward we can't use 106 as the notif id for backwards compatibility? Do we need to note that anywhere?

aitorvs: Ah good catch. Will add that as a comment. 
We should probably make this more bullet proof than comments. Will prob create a separate task for that.

karlenDimla (commented):
Is it expected that when updating from develop -> this branch that the appTP waitlist doesn't disappear? It works as expected when doing a fresh install of this branch.

aitorvs:
> Is it expected that when updating from develop -> this branch that the appTP waitlist doesn't disappear? It works as expected when doing a fresh install of this branch.

I thought they stay for ever but it seems they can be remove by calling https://developer.android.com/reference/android/app/NotificationManager#deleteNotificationChannel(java.lang.String)

Will update PR

aitorvs:
@karlenDimla, https://github.com/duckduckgo/Android/pull/2708/commits/bb7bbd782d81047f96ea3455c943e75354c41f04 is the best I could do :/

karlenDimla (approved):
Approach works 👍 

Tested update and fresh install both removes the appTP notifs.

PR #2741, opened 2023-01-18
Implementation task: Prerequisites

anikiki:
Task/Issue URL: https://app.asana.com/0/0/1203713734759235/f

### Description
Increased target_sdk to 33. Added android.permission.POST_NOTIFICATIONS to the manifest. Updated the code to use NotificationManagerCompat instead of NotificationManager.

### Steps to test this PR
No testing required.

### DIFF
--- a/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultFileDownloadNotificationManager.kt
+++ b/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultFileDownloadNotificationManager.kt
@@ -16,7 +16,6 @@
 
 package com.duckduckgo.downloads.impl
 
-import android.app.NotificationManager
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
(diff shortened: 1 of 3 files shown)


downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultFileDownloadNotificationManager.kt
```diff
@@ -52,7 +52,7 @@ private const val SUMMARY_ID = 0
 )
 @SingleInstanceIn(AppScope::class)
 class DefaultFileDownloadNotificationManager @Inject constructor(
-    private val notificationManager: NotificationManager,
+    private val notificationManager: NotificationManagerCompat,
```
malmstein, line 55: in `DeviceShieldAlertNotificationBuilder`, `NotificationModule` we use `NotificationManager` should we replace them there too?

malmstein (approved):
Added a comment about `NotificationManager` It doesn't seem like we need to replace them now, so happy to move on.

downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultFileDownloadNotificationManager.kt
anikiki: Good point! Added https://app.asana.com/0/0/1203771987668114/f to take care of it as a separate PR.

PR #2744, opened 2023-01-19
Sync native library as submodule

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203781131138682/f 

### Description
- SyncNativeLib kotlin class that provides methods to consume ddgcrypto Library
- Calls from kotlin to native code is done via JNI interfaces
- JNI interface implementation

Disclaimer: We are now using a copy of what is inside https://github.com/duckduckgo/BrowserServicesKit/tree/brindy/ddg-sync/Sources/DDGSync. The plan is to move lib module to its own repository where the Sync c library will be shared between clients of different platforms.

### Steps to test this PR
No tests are required on this PR, we will test the library in future PRs when we introduce the different setup flows.

Right now, ensuring Android Tests pass is enough.

### UI changes
| Before  | After |
| ------ | ----- |
!(Upload before screenshot)|(Upload after screenshot)|

PR #2740, opened 2023-01-18
Sync encrypted store for ids

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203773736913472/f 

### Description
This PR introduces encrypted storage for sync. Currently only stored some identifiers.

Reviewer will find:
- Encrypted storage where we will store userId, deviceId and deviceName
- Adds :store module for sync
- We persist generated ids into the store

### Steps to test this PR

_Feature 1_
- [x] fresh install (internal build)
- [x] go to sync dev settings
- [x] see generated Ids
- [x] Ids are persisted. Ensure killing the app and visiting the screen shows the same Ids

### UI changes
| Before  | After |
| ------ | ----- |
!(Upload before screenshot)|(Upload after screenshot)|

PR #2737, opened 2023-01-18
Sync settings test screen skeleton

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203771840830191/f 

### Description
Creates the skeleton and the first version of the Sync dev settings Test screen.
Includes:
- submodules structure
- initial activity and layout (only visible to internal builds)
- shows some device data (this will change a bit in future PRs)

### Steps to test this PR

_Feature 1_
- [x] Access settings 
- [x] If internal build, ensure option is present and accessing you see userId, deviceId and deviceName
- [x] If not internals, ensure option is not present

### UI changes
| Before  | After |
| ------ | ----- |
!(Upload before screenshot)|(Upload after screenshot)|

PR #2759, opened 2023-01-23
Create account user flow

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203804027992477/f 

### Description
PR adds to the internal setting screen options to allow user to create a sync account.

Some addons for testing purposes:
- show errors as toast (testing purposes)
- Print data related to the account created
- Store recovery code (testing purposes)
- Reset clears all data store (testing purposes)

### Steps to test this PR

_Feature 1_
- [x] install app
- [x] Go to internal sync settings
- [x] Create account
- [x] ensure account details are shown
- [x] Create account
- [x] ensure "duplicate user" error appears (toast message)
- [x] Click "reset"
- [x] Ensure account info removed
- [x] Now you can create another account

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
@@ -107,7 +107,7 @@ constructor(
 
     private fun isSignedIn() = !syncStore.primaryKey.isNullOrEmpty()
 
-    class Adapters {
+    private class Adapters {
         companion object {
             private val moshi = Moshi.Builder().build()
             val recoveryCodeAdapter: JsonAdapter<RecoveryCode> =
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
@@ -35,10 +35,7 @@ interface SyncRepository {
 }
 
 @ContributesBinding(AppScope::class)
-@SingleInstanceIn(AppScope::class)
-class AppSyncRepository
-@Inject
-constructor(
+class AppSyncRepository @Inject constructor(
     private val syncDeviceIds: SyncDeviceIds,
     private val nativeLib: SyncLib,
     private val syncApi: SyncApi,
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
@@ -28,7 +28,7 @@ import javax.inject.Inject
 import timber.log.Timber
 
 interface SyncRepository {
-    fun createAccount(): Result
+    fun createAccount(): Result<Boolean>
     fun getAccountInfo(): AccountInfo
     fun storeRecoveryCode()
     fun removeAccount()
@@ -45,7 +45,7 @@ constructor(
     private val syncStore: SyncStore,
 ) : SyncRepository {
 
-    override fun createAccount(): Result {
+    override fun createAccount(): Result<Boolean> {
         val userId = syncDeviceIds.userId()
         val deviceId = syncDeviceIds.deviceId()
         val deviceName = syncDeviceIds.deviceName()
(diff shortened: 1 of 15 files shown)


cmonfortep:
* develop
  * **PR #2737** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2737" target="_blank">Graphite</a> 
    * **PR #2740** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2740" target="_blank">Graphite</a> 
      * **PR #2744** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2744" target="_blank">Graphite</a> 
        * **PR #2759** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2759" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2759?utm_source=stack-comment).

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncServiceRemote.kt
```diff
@@ -0,0 +1,99 @@
...
+                    token = token,
+                    user_id = userId,
+                ),
+            )
+        }
+    }
+
+    private fun <T, R> onSuccess(
+        response: Response<T?>,
+        onSuccess: (T?) -> Result<R>,
```
cmonfortep, line 75: Related to your questions about this lambda, maybe it's clear checking the usage here: https://github.com/duckduckgo/Android/pull/2773/files#diff-080e21d07cb22aa1538a99beb48a2c59e716e6e384244b598beb8d5b4af4f9f2R91

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
```diff
@@ -0,0 +1,140 @@
...
+
+interface SyncRepository {
+    fun createAccount(): Result<Boolean>
+    fun getAccountInfo(): AccountInfo
+    fun storeRecoveryCode()
+    fun removeAccount()
+}
+
+@ContributesBinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
```
aitorvs, line 38: Given that this is stateless, does it need to be a singleton?

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
```diff
@@ -0,0 +1,140 @@
...
+    data class Success<out T>(val data: T) : Result<T>()
+    data class Error(val code: Int = -1, val reason: String) : Result<Nothing>()
+
+    override fun toString(): String {
+        return when (this) {
+            is Success<*> -> "Success[data=$data]"
+            is Error -> "Error[exception=$code, $reason]"
+        }
+    }
+}
```
aitorvs, line 140: Kotlin stdlib already has a `Result` that has more/better functionality. Do we need our own? I would remove it

aitorvs (approved):
Left couple code change suggestions.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
cmonfortep: I would like to replace it as well. The problem I have with kotlin `Result` is that `failure` only accepts a `Throwable`. I would like to be able to add any object as possible failure and not only `Throwables`. what do you think?

aitorvs: I guess I am not clear why we want to be so specific and propagate the `ErrorResponse` errors. The error string is not localized anyway so we can't really use it.
I wonder if we're over-designing this atm.

No strong opinions

cmonfortep: I agree with you.

To give more context, this is just the current approach to allow me to show an error as a Toast in the UI. This is just a convenient way to propagate them into the UI. We are not using those errors for other purposes, and it's ok to show them as they come (no localization) from backend or other internal logic.

➡️ In the future, this will change for sure. We will only propagate errors that need to land on the UI (maybe none of them), and we will not use the current entity of `code`:`message`. Maybe at that point will make sense to drop `Result`.

PR #2776, opened 2023-01-26
Android Bug: Undo doesn't work when deleting downloaded files

anikiki:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1203760457164434/f

### Description
Updated the downloads repository to not handle any physical file deletion as this is done in the view model.

### Steps to test this PR

Reproduce the issue (delete all files scenario + delete individual file scenario)
- [ ] Install from develop.
- [ ] Open the app.
- [ ] Download a file.
- [ ] Go to the "Downloads" screen and tap on "Delete All" from the overflow menu.
- [ ] See the toast and tap on "Undo".
- [ ] Notice the file is back in the list, however when you tap on it it disappears as it doesn't exist on the disk.
- [ ] Download another file.
- [ ] Go to the "Downloads" screen and tap on the menu associated to the file item and select "Delete".
- [ ] See the toast and tap on "Undo".
- [ ] Notice the file is back in the list, however when you tap on it it disappears as it doesn't exist on the disk.

Test the fix (delete all files scenario + delete individual file scenario)
- [ ] Install from this branch.
- [ ] Open the app.
- [ ] Download a file.
- [ ] Go to the "Downloads" screen and tap on "Delete All" from the overflow menu.
- [ ] See the toast and tap on "Undo".
- [ ] Notice the file is back in the list, and it exists on the disk.
- [ ] Download another file.
- [ ] Go to the "Downloads" screen and tap on the menu associated to the file item and select "Delete".
- [ ] See the toast and tap on "Undo".
- [ ] Notice the file is back in the list, and it exists on the disk.

### DIFF
--- a/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultDownloadsRepository.kt
+++ b/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultDownloadsRepository.kt
@@ -24,7 +24,6 @@ import com.duckduckgo.downloads.store.DownloadStatus
 import com.duckduckgo.downloads.store.DownloadsDatabase
 import com.squareup.anvil.annotations.ContributesBinding
 import dagger.SingleInstanceIn
-import java.io.File
 import javax.inject.Inject
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
@@ -57,9 +56,6 @@ class DefaultDownloadsRepository @Inject constructor(
     }
 
     override suspend fun delete(downloadId: Long) {
-        downloadsDatabase.downloadsDao().getDownloadItem(downloadId)?.let {
-            File(it.filePath).delete()
-        }
         downloadsDatabase.downloadsDao().delete(downloadId)
         urlFileDownloadCallManager.remove(downloadId)
     }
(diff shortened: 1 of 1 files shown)


downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultDownloadsRepository.kt
```diff
@@ -57,9 +56,6 @@ class DefaultDownloadsRepository @Inject constructor(
     }
 
     override suspend fun delete(downloadId: Long) {
-        downloadsDatabase.downloadsDao().getDownloadItem(downloadId)?.let {
```
anikiki, line 60: This was added when we removed the DownloadManager. File deletion is already handled in the viewmodel.

downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DefaultDownloadsRepository.kt
```diff
@@ -71,7 +67,6 @@ class DefaultDownloadsRepository @Inject constructor(
 
     override suspend fun deleteAll() {
         downloadsDatabase.downloadsDao().getDownloads().forEach {
-            File(it.filePath).delete()
```
anikiki, line 74: Same as above.

cmonfortep (approved):
LGTM

PR #2787, opened 2023-01-30
Remove bad health unused code

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203851151866051/f

### Description
Remove the bad health monitoring as it's no longer needed/used

### Steps to test this PR
Smoke tests for AppTP and code review


aitorvs:
* develop
  * **PR #2787** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2787" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2787?utm_source=stack-comment).

karlenDimla (approved):
LGTM and smoke test worked Ok  👍 

Note: I see that we are deleting AppHealthDatabase. I assume that it wouldn't be dropped on the users who the DB on their devices? Is that correct and expected?

aitorvs:
> LGTM and smoke test worked Ok 👍
> 
> Note: I see that we are deleting AppHealthDatabase. I assume that it wouldn't be dropped on the users who the DB on their devices? Is that correct and expected?

Yeah, this was on purpose. I had checked and that DB will ever had a handful of rows so it seemed like a good enough trade-off given than clearing the table would require a migration.

I think anyway we should look internally how we ensure cleanup when we retire DBs. Don't have a great idea tho 🤔

PR #2813, opened 2023-02-03
Change the way we store URLs in autofill

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203892213154867/f 

### Description
This PR changes the way that we store URls in autofill to only store host:port

### Steps to test this PR

- [x] Go to `https://fill.dev/form/login-simple`
- [x] Login and store the credentials
- [x] Check that the website URL that was store is `fill.dev`
- [x] Check the tests to make sure we also store the port when it exists and also the full host when it contains a subdomain(s)

### DIFF
--- a/autofill/autofill-api/src/main/java/com/duckduckgo/autofill/api/urlmatcher/AutofillUrlMatcher.kt
+++ b/autofill/autofill-api/src/main/java/com/duckduckgo/autofill/api/urlmatcher/AutofillUrlMatcher.kt
@@ -23,5 +23,11 @@ interface AutofillUrlMatcher {
         savedSite: ExtractedUrlParts,
     ): Boolean
 
+    /**
+     * This method tries to clean up a raw URL.
+     * @return a `String` containing host:port for the given raw URL.
+     */
+    fun cleanRawUrl(rawUrl: String): String
+
     data class ExtractedUrlParts(val eTldPlus1: String?, val subdomain: String?)
 }
--- a/app/src/test/java/com/duckduckgo/app/global/UriExtensionTest.kt
+++ b/app/src/test/java/com/duckduckgo/app/global/UriExtensionTest.kt
@@ -257,14 +257,6 @@ class UriExtensionTest {
     }
 
     @Test
-    fun whenUriExtractSchemeAndDomainThenReturnOnlySchemeAndDomain() {
-        assertEquals("https://www.foo.com", "https://www.foo.com/path/to/foo?key=value".extractSchemeAndDomain())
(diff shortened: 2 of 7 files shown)


marcosholgado:
* develop
  * **PR #2813** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2813" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2813?utm_source=stack-comment).

CDRussell:
When manually editing a URL, it does not strip the scheme (or anything) like it does when it's automatically saved.

marcosholgado:
> When manually editing a URL, it does not strip the scheme (or anything) like it does when it's first saved. The editing behaviour should match the new saving behaviour re URL stripping.

👍  sorted :D

autofill/autofill-api/src/main/java/com/duckduckgo/autofill/api/urlmatcher/AutofillUrlMatcher.kt
```diff
@@ -20,5 +20,11 @@ interface AutofillUrlMatcher {
     fun extractUrlPartsForAutofill(originalUrl: String): ExtractedUrlParts
     fun matchingForAutofill(visitedSite: ExtractedUrlParts, savedSite: ExtractedUrlParts): Boolean
 
+    /**
+     * This method tries to clean up a raw URL. If the input is not a URL we then store as it comes
+     * @return a `String` which has been cleaned up if it is a URL, or through pass otherwise.
```
CDRussell, line 25: nit: not sure what this means: "or through pass otherwise"

CDRussell:
I think there is a missing part of the URL clean up: handling a website which **is not** prefixed with a scheme. for example, if the user enters:

- `https://example.com/path/to/foo` that gets cleaned up to `example.com` 👍
- `example.com/path/to/foo` ➡️ `example.com/path/to/foo` ❌  this doesn't currently get cleaned up, but should IMO

I think both of these inputs should resolve to the same final output (`example.com`).

marcosholgado:
> I think there is a missing part of the URL clean up: handling a website which **is not** prefixed with a scheme. for example, if the user enters:
> 
> * `https://example.com/path/to/foo` that gets cleaned up to `example.com` 👍
> * `example.com/path/to/foo` ➡️ `example.com/path/to/foo` ❌  this doesn't currently get cleaned up, but should IMO
> 
> I think both of these inputs should resolve to the same final output (`example.com`).

Oki so... I was overcomplicating this thing way too much. So yes, you are correct and this should work now. More context in https://app.asana.com/0/488551667048375/1203892213154867/f

CDRussell (approved):

PR #2818, opened 2023-02-07
Create the feature flag code generator

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203898052213031/f

### Description
See [asana task](https://app.asana.com/0/1202552961248957/1203898052213029/f) for full background.

The PR adds a code generator for (remote) feature flags. 

For instance, the following remote feature:

```json
{
  "myFeatureName": {
    "exceptions": [...],
    "state": "enabled",
    "settings": {...},
    "features": {
      "remoteFeatureName": {
        "state": "disabled"
      }
    }
    "hash": "..."
  }
}
```

can be represented with the following code:

```kotlin
@ContributesRemoteFeature(
    scope = AppScope::class,
    featureName = "name",
//    settingsStore = MyFeatureFeatureSettingStore::class,
//    exceptionsStore = MyFeatureFeatureExceptionStore::class,
)
class MyRemoteFeature {
  @Toggle.DefaultValue(false)
  fun self(): Toggle

  @Toggle.DefaultValue(false)
  fun remoteFeatureName(): Toggle
}

class MyFeatureFeatureToggleStore : Toggle.Store {...}

class MyFeatureFeatureSettingStore : RemoteFeatureSettingsStore {...}

class MyFeatureFeatureExceptionStore : RemoteFeatureExceptionStore {...}

```

Where:
* the `MyRemoteFeature` will be the class used to check the state of the flags
* the `MyRemoteFeature` shall be annotated with `FeatureName` that will specify the name of the global feature
* the `MyRemoteFeature` shall be annotated with `ContributesRemoteFeature` for the code to be generated and shall specify
  * Dagger scope
  * `settingsStore` class that shall implement `RemoteFeatureSettingsStore` interface (Optional)
  * `exceptionStore` class that shall implement `RemoteFeatureExceptionStore` interface (Optional)
  * _NOTE: settings and exceptions store are optional. If not provided, neither will be stored_
* all methods in `MyRemoteFeature` shall return `Toggle`
* the `MyRemoteFeature` class shall contain one method called `self()`
  * this method will return the state of the global `featureName`
* the `MyRemoteFeature` may content other methods, dubbed after the sub-feature name that will also return `Toggle`

The codegen adds validation checks to ensure the correctness of the feature interfaces

After, to use the feature one can do the following:

```kotlin
private val feature: MyRemoteFeature

feature.self().isEnabled()
feature.self().setEnabled(state)

feature.remoteFeatureName().isEnabled()
feature.remoteFeatureName().setEnabled(state)
```

### How does this work?
The magic is done in two steps:
* `ContributesRemoteFeatureCodeGenerator` class is triggered by interfaces annotated with `ContributesRemoteFeature` that represent the feature. 
  * generates and contributes a `PrivacyFeaturePlugin` that will do the proper parsing for the feature
  * generates a dynamic proxy for the feature flag interface. This proxy intercepts all interface function calls and then checks the state of the toggles in the `Toggle.Store`
* finally, the proxy implementation leaves in `FeatureToggles` class

### Steps to test this PR

_Test App/Autofill_
- [x] smoke test the app and autofill to ensure nothing changed

### DIFF
--- a/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureSettings.kt
+++ b/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureSettings.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2023 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.duckduckgo.feature.toggles.api
+
+object FeatureSettings {
+    interface Store {
+        fun store(
+            featureName: String,
+            jsonString: String,
+        )
+    }
+
+    val EMPTY_STORE = object : Store {
+        override fun store(featureName: String, jsonString: String) {}
+    }
+}
--- a/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureSettings.kt
+++ b/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureSettings.kt
@@ -18,12 +18,11 @@ package com.duckduckgo.feature.toggles.api
 object FeatureSettings {
     interface Store {
         fun store(
-            featureName: String,
             jsonString: String,
         )
     }
 
     val EMPTY_STORE = object : Store {
-        override fun store(featureName: String, jsonString: String) {}
+        override fun store(jsonString: String) {}
     }
 }
--- a/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
+++ b/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
@@ -38,7 +38,19 @@ class FeatureToggles private constructor(
         fun store(store: Toggle.Store) = apply { this.store = store }
         fun appVersionProvider(appVersionProvider: () -> Int) = apply { this.appVersionProvider = appVersionProvider }
         fun featureName(featureName: String) = apply { this.featureName = featureName }
-        fun build() = FeatureToggles(this.store!!, appVersionProvider, featureName!!)
+        fun build(): FeatureToggles {
+            val missing = StringBuilder()
+            if (this.store == null) {
+                missing.append("store")
+            }
+            if (this.featureName == null) {
+                missing.append(", featureName ")
+            }
+            if (missing.isNotBlank()) {
+                throw IllegalArgumentException("This following parameters can't be null: $missing")
+            }
+            return FeatureToggles(this.store!!, appVersionProvider, featureName!!)
+        }
     }
 
     fun <T> create(toggles: Class<T>): T {
--- a/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
+++ b/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2023 DuckDuckGo
+ *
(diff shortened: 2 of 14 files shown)


aitorvs:
* develop
  * **PR #2818** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2818" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2818?utm_source=stack-comment).

aitorvs:
This is draft but close to final. I think the only thing missing is adding support for minVersion when checking `isEnabled` and ensure tests pass

CDRussell:
I'm excited for this to land! Even though draft, gave it a quick run through to test re Autofill capabilities and looking good. 

Can provide a proper review when it's ready, but overall this is definitely taking us in the right direction. 👍

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -0,0 +1,144 @@
...
+    private val defaultValue: Boolean,
+    private val appVersionProvider: () -> Int,
+) : Toggle {
+    override fun isEnabled(): Boolean {
+        store.get(key)?.let { state ->
+            return state.enable && appVersionProvider.invoke() >= (state.minSupportedVersion ?: 0)
+        } ?: return defaultValue
+    }
+
+    override fun setEnabled(state: Toggle.State) {
```
CDRussell, line 141: did you still want to make this more clearly related to overriding? thinking on it since i asked originally, I don't think it's necessary; originally thought it was taking in a boolean but taking in a state should make it obvious enough what the options are

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/feature/plugin/AutofillRemoteFeature.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.duckduckgo.feature.toggles.api.features.AutofillFeature
+
+@ContributesRemoteFeature(
+    scope = AppScope::class,
+    boundType = AutofillFeature::class,
+    toggleStore = AutofillFeatureToggleStore::class,
+    settingsStore = AutofillFeatureSettingStore::class,
+    exceptionsStore = AutofillFeatureExceptionStore::class,
+)
+interface AutofillRemoteFeature
```
CDRussell, line 30: It's not clear to me, but I don't think this interface is used externally for anything. Can we make that more obvious that's it a dummy/placeholder (if it is indeed just that) by:

- marking it as `private`
- calling it something like `Unused` or something
- updating the example in `ContributesRemoteFeature`

or maybe i'm missing something about it and it is important? 

Related to this, is it not possible to ditch this interface altogether and move all of the properties straight onto the actual feature interface? i.e., instead of defining them all here like this, could this `@ContributesRemoteFeature` annotation replace `@FeatureName("autofill")`?

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
aitorvs: I realised that the `setEnabled` is also (rightly) called from codegen, where we won't override anything.
I don't think we have great options here. We'll have to leave it to devs to understand they should never call this method from prod code

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/feature/plugin/AutofillRemoteFeature.kt
aitorvs: That's now renamed to `private interface UnusedAutofillRemoteFeatureCodegenTrigger`

feature-toggles/feature-toggles-impl/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
     implementation project(path: ':di')
     implementation project(path: ':common')
     implementation project(path: ':feature-toggles-api')
+    implementation project(path: ':feature-toggles-test')
```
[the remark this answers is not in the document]
aitorvs: Yes

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureSettings.kt
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.duckduckgo.feature.toggles.api
+
+object FeatureSettings {
+    interface Store {
+        fun store(
+            featureName: String,
```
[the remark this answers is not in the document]
aitorvs: Removed

marcosholgado (approved):
Ship it!

PR #2824, opened 2023-02-08
Consider ports for autofill matching and for displaying logins

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203892213154868/f

### Description
Considers port numbers when comparing urls, and displays port number in the UI if set.

#### Testing when saved site contains a port
- [x] Add a login for `fill.dev`
- [x] Visit `fill.dev/form/login-simple` and verify it offers to autofill as normal
- [x] Open credential management screen; verify it is listed as a `suggested` site
- [x] Edit the login and append a port to the URL (e.g., `fill.dev:9000`)
- [x] Verify the port number is visible in the list item url (assuming a title isn't given)
- [x] Visit the login page again (refreshing if required); verify you are **not prompted** to autofill (as ports don't match)
- [x] Open credential management screen; verify it is **not** listed as a `suggested` site

#### Testing when visited site contains a port
This is harder, unless you happen to know a publicly accessible website which uses a non-standard web port. If not, simulate the test by hardcoding a URL for the current site when launching creds management screen. 
- [x] `return AutofillManagementActivity.intentShowSuggestion(context, "fill.dev:9000")` on `AutofillManagementActivity`, line 346
- [x] Open credential management screen; verify it is listed as a `suggested` site

### DIFF
--- a/autofill/autofill-store/src/main/java/com/duckduckgo/autofill/store/urlmatcher/AutofillDomainNameUrlMatcher.kt
+++ b/autofill/autofill-store/src/main/java/com/duckduckgo/autofill/store/urlmatcher/AutofillDomainNameUrlMatcher.kt
@@ -27,17 +27,19 @@ import timber.log.Timber
 class AutofillDomainNameUrlMatcher @Inject constructor() : AutofillUrlMatcher {
 
     override fun extractUrlPartsForAutofill(originalUrl: String?): ExtractedUrlParts {
-        if (originalUrl == null) return ExtractedUrlParts(null, null)
+        if (originalUrl == null) return unextractable()
 
         val normalizedUrl = originalUrl.normalizeScheme()
         return try {
-            val eTldPlus1 = normalizedUrl.toHttpUrl().topPrivateDomain()
+            val httpUrl = normalizedUrl.toHttpUrl()
+            val eTldPlus1 = httpUrl.topPrivateDomain()
             val domain = originalUrl.extractDomain()
+            val port = httpUrl.port
             val subdomain = determineSubdomain(domain, eTldPlus1)
-            ExtractedUrlParts(eTldPlus1, subdomain)
+            ExtractedUrlParts(eTldPlus1, subdomain, port)
         } catch (e: IllegalArgumentException) {
             Timber.w("Unable to parse e-tld+1 from $originalUrl")
-            ExtractedUrlParts(null, null)
+            unextractable()
         }
     }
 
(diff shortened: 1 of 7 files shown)


CDRussell:
* develop
  * **PR #2824** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2824" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2824?utm_source=stack-comment).

autofill/autofill-store/src/main/java/com/duckduckgo/autofill/store/urlmatcher/AutofillDomainNameUrlMatcher.kt
```diff
@@ -27,17 +27,19 @@ import timber.log.Timber
...
-        if (originalUrl == null) return ExtractedUrlParts(null, null)
+        if (originalUrl == null) return unextractable()
 
         val normalizedUrl = originalUrl.normalizeScheme()
         return try {
-            val eTldPlus1 = normalizedUrl.toHttpUrl().topPrivateDomain()
+            val httpUrl = normalizedUrl.toHttpUrl()
+            val eTldPlus1 = httpUrl.topPrivateDomain()
             val domain = originalUrl.extractDomain()
+            val port = httpUrl.port
```
marcosholgado, line 37: Not sure about the consequences of this but getting the port from `HttpUrl` will always returns a port. If this is only used for matching then it's fine but if at some points it leaks into title, etc we might show ports like 80. Getting the port from URI will return -1 if there is no port and it won't infer 80, etc.

marcosholgado (approved):
LGTM. Left a comment just in case we need to be careful about it. Secondarily, should ports be taken in consideration when sorting? I added fill.dev with 9000, 100 and 450 in that order and that's the order they show up.

autofill/autofill-store/src/main/java/com/duckduckgo/autofill/store/urlmatcher/AutofillDomainNameUrlMatcher.kt
CDRussell: I think it's fine here because it is just used for matching, and elsewhere for display (currently `cleanRawUrl`) it handles it differently to omit _implicit_ ports.

CDRussell:
>should ports be taken in consideration when sorting

Makes sense 👍.  I'll bring that change into the next PR.

PR #2874, opened 2023-02-20
Improve the way we choose the favicon placeholder letter for autofill screens

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1203994780181874/f

### Description
Updates the favicon placeholder used when we can't get a real favicon in the autofill management screens. 

### Steps to test this PR

Note, favicons won't show immediately for a manually added site, so you will typically see the placeholder in these scenarios (they'll be cached after next visiting that site so it will show thereafter).

For each of the following; verify the favicon placeholder is correct in **both** management list screen and when clicking into the details for that credential.

- [x] Add a login with only a text-based title (e.g. "example); verify the first letter of the title is used "e" for placeholder
- [x] Add a login with only a numerical title (e.g., 192); verify the `#` placeholder is used
- [x] Add a login with only a simple domain (e.g., `example.com`; verify `E` used as placeholder
- [x] Add a login with only a subdomain (e.g., `foo.example.com`; verify first letter of E-TLD+1 used as placeholder (e.g., "E")
- [x] Add a login with both title and url; verify the first letter of the title is used

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
@@ -29,6 +29,7 @@ import com.duckduckgo.app.browser.certificates.rootstore.IsrgRootX1
 import com.duckduckgo.app.browser.certificates.rootstore.IsrgRootX2
 import java.io.InputStream
 import java.security.cert.X509Certificate
+import java.util.concurrent.TimeUnit.MILLISECONDS
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.tls.HandshakeCertificates
@@ -42,6 +43,7 @@ class GlobalGlideModule : AppGlideModule() {
         registry: Registry,
     ) {
         val okHttpClientBuilder = OkHttpClient.Builder()
+            .connectTimeout(700, MILLISECONDS) // rather than 10 seconds
             .addInterceptor { chain: Interceptor.Chain ->
                 val request = chain.request()
 
--- a/app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
@@ -29,7 +29,6 @@ import com.duckduckgo.app.browser.certificates.rootstore.IsrgRootX1
 import com.duckduckgo.app.browser.certificates.rootstore.IsrgRootX2
 import java.io.InputStream
 import java.security.cert.X509Certificate
-import java.util.concurrent.TimeUnit.MILLISECONDS
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.tls.HandshakeCertificates
@@ -43,7 +42,6 @@ class GlobalGlideModule : AppGlideModule() {
         registry: Registry,
     ) {
         val okHttpClientBuilder = OkHttpClient.Builder()
-            .connectTimeout(700, MILLISECONDS) // rather than 10 seconds
             .addInterceptor { chain: Interceptor.Chain ->
                 val request = chain.request()
 
--- a/app/src/main/java/com/duckduckgo/app/global/view/FaviconImageView.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/view/FaviconImageView.kt
@@ -36,8 +36,9 @@ import okio.ByteString.Companion.encodeUtf8
 fun ImageView.loadFavicon(
     file: File,
     domain: String,
+    placeholder: String? = null,
 ) {
-    val defaultDrawable = generateDefaultDrawable(this.context, domain)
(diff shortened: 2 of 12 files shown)


CDRussell:
* develop
  * **PR #2874** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2874" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2874?utm_source=stack-comment).

app/src/main/java/com/duckduckgo/app/global/view/FaviconImageView.kt
```diff
@@ -69,15 +69,27 @@ fun ImageView.loadDefaultFavicon(domain: String) {
...
     domain: String,
+    overridePlaceholderCharacter: String? = null,
 ): Drawable {
     return object : Drawable() {
         private val baseHost: String = domain.toUri().baseHost ?: ""
 
-        private val letter
-            get() = baseHost.firstOrNull()?.toString()?.uppercase(Locale.getDefault()) ?: ""
+        private val letter: String
+            get() {
```
CDRussell, line 87: This is the crux of the change: being able to override the default favicon's default logic to extract the placeholder character from the host and instead allowing us to pass in exactly what should be shown (while still benefitting from all of the other stuff it's doing)

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
```diff
@@ -354,26 +363,47 @@ class AutofillManagementCredentialsMode : DuckDuckGoFragment(R.layout.fragment_a
         invalidateMenu()
     }
 
+    private suspend fun preloadFavicon(credentials: LoginCredentials) {
+        withContext(dispatchers.io()) {
+            val size = resources.getDimensionPixelSize(dimen.toolbarIconSize)
+            val placeholder = generateDefaultFavicon(credentials, size)
+            withContext(dispatchers.main()) {
+                getActionBar()?.setLogo(BitmapDrawable(resources, placeholder))
```
CDRussell, line 371: `BitmapDrawable(resources, placeholder)`--- Maybe this should be done outside of the main thread? It should only be the `actionBar()?.setLogo()` call that requires `main`.

e.g., 

```
private suspend fun preloadFavicon(credentials: LoginCredentials) {
        withContext(dispatchers.io()) {
            val size = resources.getDimensionPixelSize(dimen.toolbarIconSize)
            val placeholder = generateDefaultFavicon(credentials, size)
            val bitmapDrawable = BitmapDrawable(resources, placeholder)
            withContext(dispatchers.main()) {
                getActionBar()?.setLogo(bitmapDrawable)
            }
        }
    }
```

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
```diff
@@ -354,26 +363,47 @@ class AutofillManagementCredentialsMode : DuckDuckGoFragment(R.layout.fragment_a
         invalidateMenu()
     }
 
+    private suspend fun preloadFavicon(credentials: LoginCredentials) {
```
CDRussell, line 366: when i first read `preload` i was thinking it was pre-emptively trying to fetch the favicon from the web... maybe we could rename this function to `showPlaceholderFavicon` ?

app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
```diff
@@ -42,6 +43,7 @@ class GlobalGlideModule : AppGlideModule() {
         registry: Registry,
     ) {
         val okHttpClientBuilder = OkHttpClient.Builder()
+            .connectTimeout(700, MILLISECONDS) // rather than 10 seconds
```
CDRussell, line 46: I don't think this is needed. I tried removing it, and couldn't find a bad side effect of doing so. I think that we now show the placeholder first, before trying to load the _real_ favicon renders this unnecessary.

(I'm also a bit worried it brings in unwanted behaviour elsewhere in the app, so keen to kill it if we can)

CDRussell (commented):
@marcosholgado can you confirm if we really need that new timeout to be added? I'm hoping not.

- If that can be killed, consider this approved (I can't approve since I opened the PR).
- If we do need that, I can do a more thorough testing across elsewhere in the app that uses favicons first.

Let me know

app/src/main/java/com/duckduckgo/app/global/image/GlobalGlideModule.kt
marcosholgado: You are correct, this is a leftover from my testing. It should not be a problem, just will timeout at some point but same as it has been so far.

PR #2902, opened 2023-03-01
fixes lint rule and suppress non di usages

cmonfortep:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1204091058652565/f 

### Description
Fixes lint rule.

### Steps to test this PR

_Feature 1_
- [x] remove suppress from one of the current usages
- [x] execute lint
- [x] ensure lint fails

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ManageAppsProtectionViewModel.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ManageAppsProtectionViewModel.kt
@@ -16,6 +16,7 @@
 
 package com.duckduckgo.mobile.android.vpn.apps
 
+import android.annotation.SuppressLint
 import androidx.lifecycle.*
 import com.duckduckgo.anvil.annotations.ContributesViewModel
 import com.duckduckgo.app.global.DispatcherProvider
@@ -40,6 +41,7 @@ import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.launch
 
+@SuppressLint("NoLifecycleObserver") // we don't use DI here
 @ContributesViewModel(ActivityScope::class)
 class ManageAppsProtectionViewModel @Inject constructor(
     private val excludedApps: TrackingProtectionAppsRepository,
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ManageAppsProtectionViewModel.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ManageAppsProtectionViewModel.kt
@@ -41,7 +41,7 @@ import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.launch
 
-@SuppressLint("NoLifecycleObserver") // we don't use DI here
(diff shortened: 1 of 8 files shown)


cmonfortep:
* develop
  * **PR #2902** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2902" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2902?utm_source=stack-comment).

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/apps/ManageAppsProtectionViewModel.kt
```diff
@@ -40,6 +41,7 @@ import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.launch
 
+@SuppressLint("NoLifecycleObserver") // we don't use DI here
```
cmonfortep, line 44: unsure if we prefer suppress or replace `DefaultLifecycleObserver` by `MainProcessLifecycleObserver`... I think better to suppress since we are not observing app lifecycle.

aitorvs: Suppress yup

app/src/main/java/com/duckduckgo/app/browser/PulseAnimation.kt
```diff
@@ -30,6 +31,7 @@ import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleOwner
 import com.duckduckgo.mobile.android.ui.view.setAllParentsClip
 
+@SuppressLint("NoLifecycleObserver") // we don't use DI here
```
aitorvs, line 34: The comment is kinda wrong, the justification is not that `we don't use DI`. Is that we're not observing the app lifecycle.
Same for other comments

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
```diff
@@ -66,6 +67,7 @@ import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import timber.log.Timber
 
+@SuppressLint("NoLifecycleObserver") // we don't use DI here
```
aitorvs, line 70: I don't see where the `SettingsViewModel` observers the activity lifecycle. Maybe we need to fix that?

app/src/main/java/com/duckduckgo/app/browser/PulseAnimation.kt
cmonfortep: true. I was thinking about that when writing my previous comment 👍

aitorvs (commented):
Left some comments but haven't tested yet. Will test tomorrow

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
cmonfortep: mmm...you are right...not sure what's this doing here. I see `onStop` cancels `deviceShieldStatePollingJob`, which is always null. Gonna remove it, but please review it because probably related to AppTp old feature?

aitorvs (approved):
Nice, thanks for taking care of this

PR #2910, opened 2023-03-03
Fix: Deny request when system permissions are not allowed

nalcalag:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204096490114623/f

### Description
Deny permission request from website when system permissions haven't been allowed

### Steps to test this PR

- Uninstall previous version of Android app if already have permission configurations (or just clear storage on Settings)
- Install from this branch
- Go to `www.mictest.cc`
- Tap on "Test my mic" button -> Rationale dialog appears
- Tap on "Allow" option -> System permission prompt appears
- Select "Don't Allow" option -> Snackbar appears
- [x] Check request is denied after Snackbar is hidden (you'll see a message that mic can't be used within the website)
- Reload the page and tap on "Test my mic" button **again** -> Rationale dialog appears
- Tap on "Allow" option **again** -> System permission prompt appears
- Select "Don't Allow" option -> Open Settings dialog appears
- [x] Check request have been denied

### DIFF
--- a/site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
+++ b/site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
@@ -35,11 +35,9 @@ import com.google.android.material.snackbar.BaseTransientBottomBar.BaseCallback
 import com.google.android.material.snackbar.Snackbar
 import com.google.android.material.snackbar.Snackbar.SnackbarLayout
 import com.squareup.anvil.annotations.ContributesBinding
-import dagger.SingleInstanceIn
 import javax.inject.Inject
 
 @ContributesBinding(FragmentScope::class)
-@SingleInstanceIn(FragmentScope::class)
 class SitePermissionsDialogActivityLauncher @Inject constructor(
     private val systemPermissionsHelper: SystemPermissionsHelper,
     private val sitePermissionsRepository: SitePermissionsRepository,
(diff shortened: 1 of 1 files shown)


nalcalag:
* develop
  * **PR #2910** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2910" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2910?utm_source=stack-comment).

site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
```diff
@@ -230,14 +229,23 @@ class SitePermissionsDialogActivityLauncher @Inject constructor(
...
-                onPermissionAllowed()
-            }
+            setAction(R.string.sitePermissionsDeniedSnackBarAction) { onPermissionAllowed() }
+            addCallback(
+                object : Snackbar.Callback() {
+                    override fun onDismissed(
+                        transientBottomBar: Snackbar?,
+                        event: Int,
+                    ) {
+                        if (event == BaseCallback.DISMISS_EVENT_TIMEOUT) { sitePermissionRequest.deny() }
```
CDRussell, line 239: This is working as expected. 

Minor nit: do we need to specifically check for a timeout event, or could we just call `sitePermissionRequest.deny()` when `onDismissed()` is called? (thinking that while right now those snackbars are not user-dismissable, if we changed that elsewhere in the code we'd suddenly miss that use case and wouldn't call deny if the user swiped the snackbar away)

CDRussell (approved):

site-permissions/site-permissions-impl/src/main/java/com/duckduckgo/site/permissions/impl/SitePermissionsDialogActivityLauncher.kt
nalcalag: @CDRussell this is on purpose because the user can select the onAction button in the snackbar which will launch the system permissions dialogs again. Therefore, if I `deny()` the permission in `onDismissed()` I won't be able to grant it after user selects allowing system permission in this second attempt.

CDRussell: Makes sense!

PR #2912, opened 2023-03-03
Remove location disabled pixel when there's not dialog prompt

nalcalag:
Task/Issue URL: https://app.asana.com/0/414730916066338/1204082334468711/f

### Description
We don't send `location_disabled` pixel anymore when the permission has been revoked forever by the user

### Steps to test this PR
- Uninstall previous version of Android app if already have permission configurations (or just clear storage on Settings)
- Install form this branch
- Navigate to maps.google.com
- Tap on "Your Location 🎯" button -> Dialog prompt appears
- Tap on "Enable" option -> System permission prompt appears
- Select "Don't Allow" option
- [x] Check pixel `m_pc_s_l_d` is fired
- Tap on "Your Location 🎯" button **again** -> Dialog prompt appears
- Tap on "Enable" option **again**-> System permission prompt appears
- Select "Don't Allow" option **again**
- [x] Check pixel `m_pc_s_l_d` is fired **again**
- Tap on "Your Location 🎯" button **(for the 3rd time)**
- [x] Check pixel `m_pc_s_l_d` is **NOT** fired

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -1141,7 +1141,7 @@ class BrowserTabFragment :
     ) {
         if (locationPermissionsHaveNotBeenGranted()) {
             if (deniedForever) {
-                viewModel.onSystemLocationPermissionDeniedOneTime()
+                viewModel.onSystemLocationPermissionDeniedForever()
             } else {
                 showSystemLocationPermissionDialog(domain)
             }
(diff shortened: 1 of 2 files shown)


CDRussell (approved):
The naming of `onSystemLocationPermissionDeniedTwice` is a little confusing, especially since it is essentially just `onSystemLocationPermissionDeniedForever` but with an extra pixel call.

I think it's the `twice` part that i dislike the most, because it might not actually be the second time you've denied the request that this gets called (if you hit "maybe later" a few times on the system dialog)

But i also don't have any good suggestion for improvement! 😅 So happy to merge as is if you also can't think of way of naming this better.

nalcalag:
> The naming of `onSystemLocationPermissionDeniedTwice` is a little confusing, especially since it is essentially just `onSystemLocationPermissionDeniedForever` but with an extra pixel call.
> 
> I think it's the `twice` part that i dislike the most, because it might not actually be the second time you've denied the request that this gets called (if you hit "maybe later" a few times on the system dialog)
> 
> But i also don't have any good suggestion for improvement! 😅 So happy to merge as is if you also can't think of way of naming this better.

This method is called the second time users don't allow permissions from the system permissions prompt so it's denied forever. It's true that you can deny more times from the rationale dialog prior the system one so I understand your confusion.
This flow will be refactored when migrating Location dialogs to match updated mic&camera ones, so let's keep it like this for the moment 👍

PR #2914, opened 2023-03-06
Fix DeviceShieldReminderNotificationScheduler

karlenDimla:
Task/Issue URL: 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldReminderNotificationScheduler.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldReminderNotificationScheduler.kt
@@ -28,6 +28,8 @@ import androidx.work.WorkManager
 import com.duckduckgo.app.di.AppCoroutineScope
 import com.duckduckgo.app.global.DispatcherProvider
 import com.duckduckgo.di.scopes.VpnScope
+import com.duckduckgo.mobile.android.vpn.AppTpVpnFeature
+import com.duckduckgo.mobile.android.vpn.VpnFeaturesRegistry
 import com.duckduckgo.mobile.android.vpn.feature.removal.VpnFeatureRemover
 import com.duckduckgo.mobile.android.vpn.service.TrackerBlockingVpnService
 import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationWorker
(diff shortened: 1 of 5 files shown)


karlenDimla:
* develop
  * **PR #2914** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2914" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2914?utm_source=stack-comment).

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldReminderNotificationScheduler.kt
```diff
@@ -81,23 +86,34 @@ class DeviceShieldReminderNotificationScheduler @Inject constructor(
...
+        vpnFeaturesRegistry.getLastRegistryChange().first == AppTpVpnFeature.APPTP_VPN.featureName &&
+            !vpnFeaturesRegistry.getLastRegistryChange().second
+
+    private fun tryShowAndScheduleAppTPNotifications() {
+        if (isStoppedDueToAppTPDisabled()) {
+            showImmediateReminderNotification()
+            scheduleReminderForTomorrow()
+        }
+    }
+
```
aitorvs, line 106: Apologies if i sound like a broken record 😬 but I am not yet convinced of this change.

I think whether or not a notification is shown should be govern by a plugin sending the content of the notification.

Possible idea
* the `DeviceShieldReminderNotificationScheduler` owns a plugin point 
* plugins would contribute the content
* `DeviceShieldReminderNotificationScheduler` knows what's the latest change because it can listen for changes to the `vpnFeaturesRegistry` as we already expose a `Flow`
* on `onVpnStopped` it asks for content to the plugin for a given feature name
* IF content returned THEN show notification ELSE noop()

karlenDimla: RIght. I think there's 2 concepts that I am going back and forth here.
1. Each VPNServiceCallback knows if they should execute an action for a VPN Feature or not.
2. Having the provider for a content of notifications (revoke, disabled) depending on the feature.

I agree that having 2 is the longer term approach as you suggested and probably what I had in this draft is more of a quick fix. Will try to find time to implement what you are suggesting.

PR #2915, opened 2023-03-06
Fix privacy workflow conditions

joshliebe:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1204116417359303/f

### Description
This PR updates the privacy workflow to run for these conditions:
- A branch with the "content scope scripts" label is pushed.
- The nightly runs.
- Manually (On failure, a task is not created in this case).

### Steps to test this PR
- [ ] https://github.com/duckduckgo/Android/actions/runs/4343261076 should be green.

### DIFF
--- a/.github/workflows/privacy.yml
+++ b/.github/workflows/privacy.yml
@@ -54,7 +54,7 @@ jobs:
           path: android-tests-report.zip
 
       - name: Create Asana task when workflow failed
-        if: ${{ failure() }}
+        if: ${{ failure() && github.event_name != 'workflow_dispatch' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/privacy.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           path: android-tests-report.zip
 
       - name: Create Asana task when workflow failed
-        if: ${{ failure() }}
+        if: ${{ failure() && github.event_name != 'workflow_dispatch' }}
```
CDRussell, line 57: Checking my understanding here, does this just mean we _don't_ create an Asana task if the job was manually created?

PR #2916, opened 2023-03-06
Replace usages of MainScope with more modern equivalents

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203802507890823/f

### Description
Older code had used `MainScope` directly to get a `CoroutineScope`, because it pre-dated AndroidX providing coroutine support into most of the Android components (e.g., lifecycle-aware `lifecycleScope`).

In most places, direct use of `MainScope` can be replaced with `lifecycleScope` where it is available.

There are a few subtle differences (but should be positive, or at least neutral, changes):

- `lifecycleScope` coroutines will be cancelled appropriately with lifecycle changes. With `MainScope`, that wasn't the case and we should have been cleaning up after ourselves manually e.g., in `onDestroy()` but never were.
- `MainScope` uses `Dispatchers.Main` by default, whereas `lifecycleScope` uses `Dispatchers.Main.immediate`. The latter should result in work launched on the main dispatcher, when already on the main thread, to happen immediately instead of being queued for execution shortly. In reality, this is likely entirely negligible. 

### Steps to test this PR

Mostly smoke testing around a few key areas:
- [x] Onboarding (making sure onboarding shown first time in the app and then the browser tab activity thereafter)
- [x] In `BrowserTabActivity`:
    - [x] long press on a link to open a tab in the background
    - [x] opening an external link works (e.g., write a hyperlink in a text message, click it and select our browser)
    - [x] add the widget when you have no favourites, test the "favourites onboarding" flow
    - [x] opening a favourite from widget works 
- [x] FireDialog works as expected

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -27,6 +27,7 @@ import android.view.KeyEvent
 import android.view.View
 import android.widget.Toast
 import androidx.annotation.VisibleForTesting
+import androidx.lifecycle.lifecycleScope
 import androidx.webkit.ServiceWorkerClientCompat
 import androidx.webkit.ServiceWorkerControllerCompat
 import androidx.webkit.WebViewFeature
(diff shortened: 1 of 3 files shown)


CDRussell:
* develop
  * **PR #2916** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2916" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2916?utm_source=stack-comment).

cmonfortep (approved):
LGTM 👍

PR #2918, opened 2023-03-06
Connected devices to sync account

cmonfortep:
Task/Issue URL: https://app.asana.com/0/0/1204039315134729/f

### Description
This PR adds functionality to show connected devices and allow do logout on them

### Steps to test this PR

_Feature 1_
- [ ] go to sync settings
- [ ] create account
- [ ] you should see now all connected devices
- [ ] you can logout the device from there

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
@@ -37,6 +37,7 @@ interface SyncRepository {
     fun logout(): Result<Boolean>
     fun deleteAccount(): Result<Boolean>
     fun latestToken(): String
+    fun getConnectedDevices(): Result<List<ConnectedDevice>>
 }
 
 @ContributesBinding(AppScope::class)
@@ -191,6 +192,30 @@ class AppSyncRepository @Inject constructor(
         return syncStore.token ?: ""
     }
 
+    override fun getConnectedDevices(): Result<List<ConnectedDevice>> {
+        val token = syncStore.token.takeUnless { it.isNullOrEmpty() }
+            ?: return Result.Error(reason = "Token Empty")
+
+        return when (val result = syncApi.getDevices(token)) {
+            is Result.Error -> {
+                Timber.i("SYNC getDevices failed $result")
+                result
+            }
+
+            is Result.Success -> {
+                return Result.Success(
+                    result.data.map {
+                        ConnectedDevice(
+                            thisDevice = syncStore.deviceId == it.device_id,
+                            deviceName = it.device_name,
+                            deviceId = it.device_id,
+                        )
+                    },
+                )
+            }
+        }
+    }
+
     private fun isSignedIn() = !syncStore.primaryKey.isNullOrEmpty()
 
     private class Adapters {
@@ -214,6 +239,12 @@ data class AccountInfo(
 data class RecoveryCode(
     val primaryKey: String,
     val userID: String,
+)
+
+data class ConnectedDevice(
+    val thisDevice: Boolean = false,
+    val deviceName: String,
+    val deviceId: String,
 )
 
 sealed class Result<out R> {
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
@@ -25,8 +25,8 @@ import com.duckduckgo.sync.store.SyncStore
 import com.squareup.anvil.annotations.ContributesBinding
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
-import timber.log.Timber
 import javax.inject.*
+import timber.log.Timber
 
 interface SyncRepository {
     fun createAccount(): Result<Boolean>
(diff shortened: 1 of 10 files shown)


cmonfortep:
* develop
  * **PR #2918** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2918" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2918?utm_source=stack-comment).

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ui/SyncInitialSetupActivity.kt
```diff
@@ -86,5 +87,15 @@ class SyncInitialSetupActivity : DuckDuckGoActivity() {
         binding.deviceNameTextView.text = viewState.deviceName
         binding.primaryKeyTextView.text = viewState.primaryKey
         binding.secretKeyTextView.text = viewState.secretKey
+        binding.connectedDevicesList.removeAllViews()
+        if (viewState.isSignedIn) {
```
cmonfortep, line 91: All this is a bit meh, but just for testing and demo.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
```diff
@@ -34,9 +34,10 @@ interface SyncRepository {
     fun getAccountInfo(): AccountInfo
     fun storeRecoveryCode()
     fun removeAccount()
-    fun logout(): Result<Boolean>
+    fun logout(deviceId: String = ""): Result<Boolean>
```
aitorvs, line 37: I would remove the default value for `deviceId`. 
```suggestion
    fun logout(deviceId: String): Result<Boolean>
```

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
```diff
@@ -152,19 +154,29 @@ class AppSyncRepository @Inject constructor(
...
-    override fun logout(): Result<Boolean> {
+    override fun logout(deviceId: String): Result<Boolean> {
         val token = syncStore.token.takeUnless { it.isNullOrEmpty() }
             ?: return Result.Error(reason = "Token Empty")
-        val deviceId = syncStore.deviceId.takeUnless { it.isNullOrEmpty() }
-            ?: return Result.Error(reason = "Device Id Empty")
+
+        val logoutThisDevice = deviceId.isEmpty() || deviceId == syncStore.deviceId
+
+        val deviceId = if (logoutThisDevice) {
```
malmstein, line 163: deviceId is either `syncStore.deviceId` or we return Result.Error right? You can simplify all that logic here and also enforce `deviceId` to not be empty when passed into this method.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
```diff
@@ -191,6 +204,30 @@ class AppSyncRepository @Inject constructor(
         return syncStore.token ?: ""
     }
 
+    override fun getConnectedDevices(): Result<List<ConnectedDevice>> {
+        val token = syncStore.token.takeUnless { it.isNullOrEmpty() }
```
malmstein, line 208: this is becoming a pattern, checking for a value in the store then returning an Error if it doesn't exist. How do you feel about moving all that logic to the store itself?

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ui/SyncInitialSetupActivity.kt
malmstein: as long we clean it I don't mind for now.

sync/sync-impl/src/test/java/com/duckduckgo/sync/impl/AppSyncRepositoryTest.kt
```diff
@@ -192,6 +197,20 @@ class AppSyncRepositoryTest {
         verify(syncStore, times(0)).clearAll()
     }
 
+    @Test
+    fun whenLogoutRemoteDeviceSucceedsThenReturnSuccessButDoNotRemoveLocalData() {
```
malmstein, line 201: since you have `deviceId` as a parameter that can be empty, there's a test missing when `logout` is called without `deviceId`.

malmstein (approved):
there's a test missing and I've added a suggestion for an emerging pattern. we can discuss that offline and make the changes in other PR.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
cmonfortep: Good point about being pattern (also in SyncApi and other parts). Probably something to extract, but it should stay somehow in the same class. That check is more related to the action the user is taking than to the value in the store. Inside the store, the empty token is a valid state, so I wouldn't move it there.

However, I don't know if we will need that level of error granularity in the final UI, so probably most of them will be removed.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncRepository.kt
cmonfortep: it can be syncStore.deviceId or another deviceId (logging out a remote device)

sync/sync-impl/src/test/java/com/duckduckgo/sync/impl/AppSyncRepositoryTest.kt
cmonfortep: I've removed the default value, and now `deviceId` is required 👍

PR #2935, opened 2023-03-08
Create pluginpoint for reminder notifs

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204088747317856/f

### Description
This PR includes introducing content plugins for VPN reminder notifications. See more in task for info.

### Steps to test this PR

_Disable AppTP_
- [x] Install build
- [x] Enable AppTP
- [x] Disable AppTP
- [x] Verify that disabled notif is shown

_Schedule notif_
- [x] Modify `DeviceShieldReminderNotificationScheduler` line `160` to be `SECONDS` instead of `HOURS`.
- [x] Install build
- [x] Enable AppTP
- [x] Disable AppTP
- [x] Remove existing disabled notification
- [x] Remove app from background (Show all apps and then remove DDG)
- [x] Wait for 20 seconds
- [x] Verify that disabled notification is shown again

_Update App_
- [x] Install develop
- [x] Enable AppTP
- [x] Disable AppTP
- [x] Verify that disabled notif is shown

_Revoke Notification_
- [x] enable AppTP
- [x] install another 3rd party VPN app
- [x] enable the 3rd party VPN app
- [x] verify AppTP is disabled and revoke notification appears
- [x] tap on revoke notification
- [x] verify it links to AppTP activity screen

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpDisabledContentPlugin.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpDisabledContentPlugin.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2023 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.mobile.android.vpn.service.notification
+
+import android.content.Context
+import androidx.core.app.NotificationCompat
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.mobile.android.vpn.R
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin.NotificationContent
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin.NotificationPriority
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin.NotificationPriority.NORMAL
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin.Type
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin.Type.DISABLED
+import com.duckduckgo.mobile.android.vpn.ui.notification.NotificationActionReportIssue
+import com.squareup.anvil.annotations.ContributesMultibinding
+import javax.inject.Inject
+
+@ContributesMultibinding(AppScope::class)
+class AppTpDisabledContentPlugin @Inject constructor(
+    private val context: Context,
+) : VpnReminderNotificationContentPlugin {
+    private val notificationPendingIntent by lazy { getAppTPNotificationPressIntent(context) }
+    private val notificationActionPendingIntent by lazy { getAppTPStartIntent(context) }
+
+    override fun getContent(): NotificationContent? {
+        return NotificationContent(
+            isSilent = true,
+            shouldAutoCancel = null,
+            customViewLayout = R.layout.notification_device_shield_disabled,
+            onNotificationPressIntent = notificationPendingIntent,
+            notificationAction = listOf(
+                NotificationCompat.Action(
+                    R.drawable.ic_vpn_notification_24,
+                    context.getString(R.string.atp_EnableCTA),
+                    notificationActionPendingIntent,
+                ),
+                NotificationActionReportIssue.reportIssueNotificationAction(context),
+            ),
+        )
+    }
+
+    override fun getPriority(): NotificationPriority {
+        return NORMAL
+    }
+
+    override fun getType(): Type = DISABLED
+}
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpRevokedContentPlugin.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpRevokedContentPlugin.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2023 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.mobile.android.vpn.service.notification
+
+import android.content.Context
+import androidx.core.app.NotificationCompat.Action
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.mobile.android.vpn.R
+import com.duckduckgo.mobile.android.vpn.R.drawable
+import com.duckduckgo.mobile.android.vpn.R.string
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin.NotificationContent
+import com.duckduckgo.mobile.android.vpn.service.VpnReminderNotificationContentPlugin.NotificationPriority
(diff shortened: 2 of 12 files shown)


karlenDimla:
* develop
  * **PR #2935** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/2935" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/2935?utm_source=stack-comment).

app-tracking-protection/vpn-api/src/main/java/com/duckduckgo/mobile/android/vpn/VpnFeaturesRegistry.kt
```diff
@@ -63,6 +63,8 @@ interface VpnFeaturesRegistry {
     fun registryChanges(): Flow<Pair<String, Boolean>>
 
     fun getRegisteredFeatures(): List<VpnFeature>
+
+    fun getLastRegistryChange(): Pair<String, Boolean>
```
karlenDimla, line 67: The implementations of `VpnStaticNotificationContentPlugin` wouldn't have a good place to subscribe to the  Flow returned by `registryChanges` (returns hot flow). We need to be able to get whatever is the last known registryChange  when isActive is called.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldAlertNotificationBuilder.kt
```diff
@@ -51,15 +51,6 @@ object DeviceShieldAlertNotificationBuilderModule {
 
 interface DeviceShieldAlertNotificationBuilder {
 
-    fun buildReminderNotification(
```
karlenDimla, line 54: My ultimate goal is to get rid of this class and have each notification have its own VpnStaticNotificationContentPlugin implementation.

app-tracking-protection/vpn-api/src/main/java/com/duckduckgo/mobile/android/vpn/service/VpnStaticNotificationContentPlugin.kt
```diff
@@ -0,0 +1,80 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.mobile.android.vpn.service
+
+import android.app.PendingIntent
+import androidx.annotation.LayoutRes
+import androidx.core.app.NotificationCompat
+
+interface VpnStaticNotificationContentPlugin {
```
aitorvs, line 23: What does it mean `Static` notification?
Should we just call it `VpnReminderNotificationContentPlugin` ?

app-tracking-protection/vpn-api/src/main/java/com/duckduckgo/mobile/android/vpn/VpnFeaturesRegistry.kt
aitorvs: Can we do `registryChanges().last()` instead? I still don't see why we have to introduce the `getLastRegistryChange()` method having the `Flow` already

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/VpnFeaturesRegistryImpl.kt
```diff
@@ -94,6 +94,10 @@ internal class VpnFeaturesRegistryImpl(
         return registeredFeatures().keys.map { VpnFeature { it } }
     }
 
+    override fun getLastRegistryChange(): Pair<String, Boolean> {
+        return _registry.asStateFlow().value
+    }
+
```
aitorvs, line 100: Ditto, I think we can remove it

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpDisabledContentPlugin.kt
```diff
@@ -0,0 +1,91 @@
...
+            ),
+        )
+    }
+
+    override fun getPriority(): NotificationPriority {
+        return NORMAL
+    }
+
+    override fun isActive(): Boolean {
+        return vpnFeaturesRegistry.getLastRegistryChange().isAppTpDisabled()
```
aitorvs, line 71: Ditto, why not `vpnFeaturesRegistry.registryChanges().last()`.
We need to do a `runBlocking` probably, but at least is explicit

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldReminderNotificationScheduler.kt
```diff
@@ -27,39 +27,48 @@ import androidx.work.PeriodicWorkRequestBuilder
...
+    private val vpnFeaturesRegistry: VpnFeaturesRegistry,
+    private val vpnStaticNotificationContentPluginPoint: PluginPoint<VpnStaticNotificationContentPlugin>,
 ) : VpnServiceCallbacks {
 
     override fun onVpnStarted(coroutineScope: CoroutineScope) {
-        scheduleUndesiredStopReminderAlarm()
-        cancelReminderForTomorrow()
-        hideReminderNotification()
-        enableReminderReceiver()
+        if (vpnFeaturesRegistry.isFeatureRegistered(AppTpVpnFeature.APPTP_VPN)) {
```
aitorvs, line 65: Why do we need to check AppTP feature specifically if we added the plugin point?

aitorvs (commented):
Some comments. I still need to re-check some things on the `DeviceShieldReminderNotificationScheduler`. It seems like an odd mix now. But I have to admit that I have just glanced through that class for now

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldReminderNotificationScheduler.kt
karlenDimla: We probably don't need to. Ill check this tom.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpDisabledContentPlugin.kt
karlenDimla: When I tried this before, the flow doesn't emit any values. I can check again tomorrow - maybe I missed something then.

app-tracking-protection/vpn-api/src/main/java/com/duckduckgo/mobile/android/vpn/service/VpnStaticNotificationContentPlugin.kt
karlenDimla: Reminder used to pertain to Disabled notif before so i tried to stay away from the term. But now that I removed anything called "reminder notif", can probably use that.

I just thought of static since unlike the Enabled notif, these notification contents are not designed to change.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/notification/DeviceShieldReminderNotificationScheduler.kt
karlenDimla: @aitorvs SO I checked this, this part of the code doesn't access any content specifically. It just schedules them.

I think I can remove this since:
- scheduleUndesiredStopReminderAlarm() - The worker schedule here checks if there's an active content plugin anyway.
- cancelReminderForTomorrow() - should work regardless of what feature is started.
- hideReminderNotification() - should work regardless of what feature is started.
-  enableReminderReceiver() - should work regardless of what feature is started.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpDisabledContentPlugin.kt
```diff
@@ -0,0 +1,69 @@
...
+            ),
+        )
+    }
+
+    override fun getPriority(): NotificationPriority {
+        return NORMAL
+    }
+
+    override fun isActive(): Boolean {
+        return !vpnFeaturesRegistry.isFeatureRegistered(AppTpVpnFeature.APPTP_VPN)
```
aitorvs, line 65: Why not returning `true` instead, or even better, removing the `isActive()` method altogether until we have a second plugin (?)
This slightly changes the logic we had before, it's only necessary when we have more than one (DISABLED) plugin and it can be the source of empty notifications (as we've seen internally) or in this case, no notification

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/service/notification/AppTpRevokedContentPlugin.kt
```diff
@@ -0,0 +1,71 @@
...
+        )
+    }
+
+    override fun getPriority(): NotificationPriority {
+        return NORMAL
+    }
+
+    override fun isActive(): Boolean {
+        return !vpnFeaturesRegistry.isFeatureRegistered(AppTpVpnFeature.APPTP_VPN)
+    }
```
aitorvs, line 68: ditto, I think we need to remove.
I have tested the revoke notification and doesn't show. Returning `true` here fixes the issue

aitorvs (commented):
We need to remove the `isActive()` method call. In addition to the proposed test cases I tested the revoked notification and doesn't show consistently.
Returning `true` in `isActive` fixes the issue.

This is the race we've also seen internally in other places.

aitorvs (approved):
LGTM, also tested revoke and now it works

PR #3012, opened 2023-03-28
Translations for [Experiment] Increase retention through AppTP promotions

anikiki:
Task/Issue URL: https://app.asana.com/0/0/1204270630225683/f

### Description
Added copy for AppTP retention experiment.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -288,6 +288,12 @@
     <!-- Notification Channel Names -->
     <string name="notificationChannelTutorials">Tutorials</string>
 
+    <!-- AppTP Retention Experiment Notifications-->
+    <string name="enableAppTpNotificationEasyStepTitle">One easy step for app privacy</string>
+    <string name="enableAppTpNotificationEasyStepDescription">Enable App Tracking Protection to block trackers hiding in apps, even when you aren’t using your phone.</string>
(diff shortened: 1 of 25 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -288,6 +288,12 @@
     <!-- Notification Channel Names -->
     <string name="notificationChannelTutorials">Tutorials</string>
 
+    <!-- AppTP Retention Experiment Notifications-->
+    <string name="enableAppTpNotificationEasyStepTitle">One easy step for app privacy</string>
+    <string name="enableAppTpNotificationEasyStepDescription">Enable App Tracking Protection to block trackers hiding in apps, even when you aren’t using your phone.</string>
```
CDRussell, line 293: I'm not sure if this will cause any problems or not, but flagging for your thoughts:
- `aren’t`  - i don't know if that's the right character to use `’`
- elsewhere we use `\'`

CDRussell (approved):
Approving, but looking to you to confirm no problems using that `’` character

app/src/main/res/values/strings.xml
anikiki: We use either of them from what I see. I always take the text with copy/paste to avoid any errors. In the past I asked this question and didn't get a compelling answer over which one to use.

PR #3034, opened 2023-04-05
Icon colors based on tracker name

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203581260317501/f

### Description
When we don't have the icon of the tracking company we create an icon on
the fly with the first letter but fixed background color.

This PR changes the logic so that the color is picked from a color
pallete and calculated directly from the company name text

### Steps to test this PR
Smoke tests for AppTP

_Icon colors_
- [x] enable AppTP
- [x] install termux app
- [x] execute `curl -v <domain> &` for the following domains, ie. `https://logs.datadoghq.com` and `https://mobile.adsafeprotected.com`
- [x] go to AppTP activity screen
- [x] verify icons have colors
- [x] go to AppTP detail screen (tap on the termux entry)
- [x] verify icons have the same color
- [x] close and relaunch the app
- [x] verify the colors continue to be the same in above screens

### DIFF
--- a/common-ui/build.gradle
+++ b/common-ui/build.gradle
@@ -46,7 +46,6 @@ dependencies {
     implementation AndroidX.constraintLayout
     implementation AndroidX.recyclerView
     implementation AndroidX.lifecycle.viewModelKtx
-    implementation Square.okio
     // just to get the dagger annotations
     implementation Google.dagger
 
--- a/common-ui/build.gradle
+++ b/common-ui/build.gradle
@@ -46,6 +46,7 @@ dependencies {
     implementation AndroidX.constraintLayout
     implementation AndroidX.recyclerView
     implementation AndroidX.lifecycle.viewModelKtx
+    implementation Square.okio
     // just to get the dagger annotations
     implementation Google.dagger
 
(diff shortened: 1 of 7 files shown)


aitorvs:
* develop
  * **PR #3034** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3034" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3034?utm_source=stack-comment).

nalcalag:
Maybe I'm wrong, but as far as I can see the custom view `TextDrawable` is not a component defined in the design system and it's only used in AppTP. Therefore, I would suggest to move it to `vpn-impl` module.
The main reason it's that we are adding a dependency only because this feature which will be added to all modules we are using `common-ui`.

aitorvs:
> Maybe I'm wrong, but as far as I can see the custom view `TextDrawable` is not a component defined in the design system and it's only used in AppTP. Therefore, I would suggest to move it to `vpn-impl` module. The main reason it's that we are adding a dependency only because this feature which will be added to all modules we are using `common-ui`.

I can move it yup. Didn't touch it because it was originally in there.
FWIW okio is already added as a dependency in the app (directly and indirectly), so moving or not moving doesn't change a thing on that front.

common-ui/build.gradle
```diff
@@ -46,6 +46,7 @@ dependencies {
     implementation AndroidX.constraintLayout
     implementation AndroidX.recyclerView
     implementation AndroidX.lifecycle.viewModelKtx
+    implementation Square.okio
```
nalcalag, line 49:
```suggestion
```
We don't need this anymore 👍

karlenDimla (approved):
Tested and logic make sense.

nalcalag (approved):
Thanks for moving the custom view `TextDrawable` to `vpn-impl` method 🙏

PR #3040, opened 2023-04-05
AppTP: Change endpoint for blocklist to pull new format

laghee:
Task/Issue URL: https://app.asana.com/0/0/1204239971719179/f

### Description
Now that we're using the cumulative blocklist for AppTP, we've decided to use a new endpoint to support the new list format and any future updates. We'll maintain the old endpoint, without updates, for users on earlier versions.

### Steps to test this PR
The endpoint is up & running with the old (current) list, but in the new format (which I'll alter today to include entity scores).

To test:
Install app from `develop`, enable AppTP & let block some trackers
Update app from this branch, verify that database transitions smoothly without trying to migrate to a new version

### DIFF
--- a/app-tracking-protection/vpn-store/src/main/java/com/duckduckgo/mobile/android/vpn/trackers/AppTracker.kt
+++ b/app-tracking-protection/vpn-store/src/main/java/com/duckduckgo/mobile/android/vpn/trackers/AppTracker.kt
@@ -105,7 +105,7 @@ data class JsonAppBlockingList(
 
 class JsonAppTracker(
     val owner: TrackerOwner,
-    val app: TrackerApp,
+    // val app: TrackerApp,
     @field:Json(name = "default")
     val defaultAction: String? = null,
 )
@@ -120,6 +120,7 @@ data class TrackerOwner(
     val displayName: String,
 )
 
+@Deprecated("This obj is no longer used. Stays to avoid db migration, SQLite doesn't allow renaming columns")
 data class TrackerApp(
     val score: Int,
     val prevalence: Double,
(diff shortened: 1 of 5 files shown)


laghee:
Just realized I probably need to regenerate the test `json` here to include entity scores as well.

app-tracking-protection/vpn-store/src/main/java/com/duckduckgo/mobile/android/vpn/trackers/AppTracker.kt
```diff
@@ -105,7 +105,7 @@ data class JsonAppBlockingList(
 
 class JsonAppTracker(
     val owner: TrackerOwner,
-    val app: TrackerApp,
+    // val app: TrackerApp,
```
aitorvs, line 108: remove?
```suggestion
```

aitorvs (commented):
Added one nit. Still need to test but we'll get to the test tomorrow

app-tracking-protection/vpn-store/src/test/resources/full_app_trackers_blocklist.json
aitorvs, line 1: This file seems to be different than the one we have in the `/android-tds.json` endpoint. Should it not be the same?

laghee: There's an open PR there to revert the entities to have scores & the version is different, but the rest should be the same. I'll ping Konrad to get that merged asap.

laghee: Ah, no -- this was messed up, sorry. Fixed the list (again) -- should also match the endpoint now.

aitorvs (approved):
Tested app update from develop and fresh install. LGTM

PR #3071, opened 2023-04-17
Detect app install source package

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204413310429877/f

### Description
Detects the app installation source package ID.

### Testing steps
See https://app.asana.com/0/488551667048375/1204413310429877/f


marcosholgado (approved):
LGTM. Could be on its own module rather than app but no strong opinion.

CDRussell:
>Could be on its own module rather than app but no strong opinion.

Good suggestion. I have a follow-up change to make to this in a few weeks, so I can pull it out alongside that change.

PR #3074, opened 2023-04-18
Check table and column exist before executing query

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204397066248837/f 

### Description
This PR checks if the cookies table and is_httponly column exist before executing the query.

### Steps to test this PR

- [x] Install in emulator with API 23 which will use an old webview version and won't have the is_httponly column
- [x] Launch the app and go to any website
- [x] Send the app to the background
- [x] App should not register a crash pixel
- [x] All tests should pass

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/referencetests/FirstPartyCookiesReferenceTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/referencetests/FirstPartyCookiesReferenceTest.kt
@@ -27,7 +27,6 @@ import com.duckduckgo.app.fire.FireproofRepository
 import com.duckduckgo.app.fire.WebViewDatabaseLocator
 import com.duckduckgo.app.global.DefaultDispatcherProvider
 import com.duckduckgo.app.privacy.db.UserAllowListRepository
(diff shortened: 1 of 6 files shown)


marcosholgado:
* develop
  * **PR #3074** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3074" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3074?utm_source=stack-comment).

CDRussell:
Querying the behaviour i'm seeing to ask if it's intended:
- on API 23, when `FirstPartyCookiesModifierWorker` runs, it can't make the update so we request that the work is retried `Result.retry()`
- I see there is an attempt to cancel all work with that tag, but I'm not seeing it ever get cancelled. When i continually background/foreground the app, it keeps trying to run that code and flagging that it should be tried again (but that will never succeed?)

CDRussell (approved):
As discussed elsewhere, treating the flagged scenario as a `success` rather than a `retry` makes sense. Approving for when that change is made.

PR #3077, opened 2023-04-19
Fix cancelled callback on dialogs

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204423592674026/f 

### Description
Fixes asking users multiple times for feedback when they dismiss the dialog.

Issue: `onDialogCancelled` was only triggered when user clicks back button, not if they click outside the container.

### Steps to test this PR

_Feature 1_
- [x] Apply patch attached in asana task
- [x] Fresh install
- [x] you should see the feedback dialog
- [x] click outside the dialog
- [x] dialog will hide
- [x] send app to background and bring it back to foreground
- [x] ensure dialog doesn't show again

_Feature 2_
- [x] Apply patch attached in asana task
- [x] Fresh install
- [x] you should see the feedback dialog
- [x] click back button
- [x] dialog will hide
- [x] send app to background and bring it back to foreground
- [x] ensure dialog doesn't show again

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/dialog/TextAlertDialogBuilder.kt
+++ b/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/dialog/TextAlertDialogBuilder.kt
@@ -111,11 +111,9 @@ class TextAlertDialogBuilder(val context: Context) : DaxAlertDialog {
             .setCancelable(isCancellable)
             .apply {
                 if (isCancellable) {
-                    setOnKeyListener(
-                        BackKeyListener {
-                            listener.onDialogCancelled()
-                        },
-                    )
+                    setOnCancelListener {
+                        listener.onDialogCancelled()
+                    }
                 }
             }
         dialog = dialogBuilder.create()


cmonfortep:
* develop
  * **PR #3077** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3077" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3077?utm_source=stack-comment).

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/dialog/TextAlertDialogBuilder.kt
```diff
@@ -111,11 +111,9 @@ class TextAlertDialogBuilder(val context: Context) : DaxAlertDialog {
...
             .apply {
                 if (isCancellable) {
-                    setOnKeyListener(
-                        BackKeyListener {
-                            listener.onDialogCancelled()
-                        },
-                    )
+                    setOnCancelListener {
+                        listener.onDialogCancelled()
+                    }
```
nalcalag, line 116: Wondering if what David wanted here was setting the back key in case it's not cancellable so the experience is not that intrusive 🤔 
```suggestion
             if (!isCancellable) {       
                    setOnKeyListener(
                        BackKeyListener {
                            listener.onDialogCancelled()
                        },
                     )
```

nalcalag (approved):
Works as expected 💯 
Firing a question for improving the experience as I think it was the expectation adding this piece of code which is not a blocker for this bug fix at all.

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/dialog/TextAlertDialogBuilder.kt
cmonfortep: I've created https://app.asana.com/0/1202857801505092/1204448904099019/f to capture the question. We can review it, but not blocking the fix.

PR #3095, opened 2023-04-24
Ensure we catch ActivityNotFoundExceptions when opening external app links

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204454121220459/f

### Description
When we detect users are navigating to a URL that could be consumed by a native app, we offer a snackbar with an option to launch that app. However, since we don't control the intent that will be used, there is a possibility our app will crash when we use that intent with `startActivity` due to an `ActivityNotFoundException` (which is happening according to our crash data)

### Steps to test this PR

- [x] Search for `uber play store`
- [x] Click on the Play Store result from the SERP
- [x] Click on the SnackBar action item; verify it launches ok
- [x] To really test it, use a fake intent in the `BrowserTabFragment.openAppLink()` function like this, `Intent("com.example.foo/com.bar.ActivityThatDoesNotExist")` and make sure the app doesn't crash when you click on the snackbar action button

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -1375,16 +1375,24 @@ class BrowserTabFragment :
         if (appLink.appIntent != null) {
             appLink.appIntent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
             try {
-                startActivity(appLink.appIntent)
+                startActivityOrQuietlyFail(appLink.appIntent)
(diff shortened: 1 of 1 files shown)


CDRussell:
* develop
  * **PR #3095** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3095" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3095?utm_source=stack-comment).

joshliebe (approved):
Nice, LGTM

PR #3120, opened 2023-05-02
add crash log for shouldoverride method

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204513195956997/f 

### Description
Add crashlog for crash events inside webview shouldOverride method

### Steps to test this PR

_Feature 1_
- [x] Add a crash event inside shouldOverride webview method
- [x] ensure it's captured
- [x] ensure log is send

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
@@ -30,6 +30,8 @@ import androidx.test.annotation.UiThreadTest
 import androidx.test.filters.SdkSuppress
 import androidx.test.platform.app.InstrumentationRegistry
 import com.duckduckgo.adclick.api.AdClickManager
+import com.duckduckgo.anrs.api.CrashLogger
+import com.duckduckgo.anrs.api.CrashLogger.Crash
 import com.duckduckgo.app.CoroutineTestRule
 import com.duckduckgo.app.accessibility.AccessibilityManager
 import com.duckduckgo.app.browser.certificates.rootstore.TrustedCertificateStore
@@ -94,6 +96,7 @@ class BrowserWebViewClientTest {
     private val autoconsent: Autoconsent = mock()
     private val contentScopeScripts: ContentScopeScripts = mock()
     private val pixel: Pixel = mock()
+    private val crashLogger: CrashLogger = mock()
 
     @UiThreadTest
     @Before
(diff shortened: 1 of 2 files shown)


cmonfortep:
* develop
  * **PR #3120** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3120" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3120?utm_source=stack-comment).

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
```diff
@@ -250,6 +254,14 @@ class BrowserWebViewClientTest {
         verify(browserAutofillConfigurator).configureAutofillForCurrentPage(webView, null)
     }
 
+    @Test
+    fun whenShouldOverrideThrowsExceptionThenRecordException() = runTest {
+        val exception = RuntimeException()
+        whenever(specialUrlDetector.determineType(initiatingUrl = any(), uri = any())).thenThrow(exception)
+        testee.shouldOverrideUrlLoading(webView, "")
+        verify(crashLogger).logCrash(Crash(shortName = "m_webview_shouldOverride", t = exception))
```
joshliebe, line 262:
```suggestion
        verify(crashLogger).logCrash(Crash(shortName = "m_webview_should_override", t = exception))
```

joshliebe (approved):
LGTM (There's just a typo in the test)

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -213,7 +215,7 @@ class BrowserWebViewClient @Inject constructor(
                 }
             }
         } catch (e: Throwable) {
-            Timber.e(e.localizedMessage)
+            crashLogger.logCrash(CrashLogger.Crash(shortName = "m_webview_should_override", t = e))
```
marcosholgado, line 218: Do we want a different shortName? Is there any value? The stacktrace should point to here anyways right?

cmonfortep: Probably no big difference at the moment, but I think putting a different `shortName` for captured crashes may have some benefits.

Pixel name will still be `APPLICATION_CRASH_GLOBAL` since we are using `logCrash`.

For the `shortName`, we could also use `APPLICATION_CRASH_GLOBAL`, but I don't think 100% correct for something we've captured. Having different SN allows you to filter or group (again, we are not doing it now, but it could be a usecase in the future). If at some point we decide to remove that pixel from the global crash, and fire it as a pixel, that will be the pixel name, so we can follow the trend.

PR #3145, opened 2023-05-08
Remove redundant onboarding logic in Autofill

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1204099438373202/f

### Description
Initially, in the dialog where we ask if you want to save a login, we had a special "onboarding" title and subtitle for the dialog which would only show if you'd never saved a login before. 

Later, we decided to always show the longer-form "onboarding" text, but left some code and strings lying around in case we decided against it. At the time, we noted we should come back and clean up if it looks like it wasn't coming back any time soon. This is that clean-up.

### Steps to test this PR

- [x] Visit https://fill.dev/form/login-simple
- [x] Enter username and password and hit login button
- In the prompt that shows offering to save credentials, 
    - [x] verify that the title is `Do you want DuckDuckGo to save your Login?`
    - [x] verify the subtitle is `Logins are stored securely on this device only, and can be managed from the Logins menu in Settings.`
- [x] Repeat this test saving another login (different username+password)
- [x] Verify title and subtitle in dialog are the same as before


joshliebe (approved):
LGTM!

PR #3204, opened 2023-05-25
Enable sync for internal

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204676817319311/f 

### Description
Enables, and makes feature visible, to internal flavor (aka. internal users)

### Steps to test this PR

_Feature 1_
- [ ] checkout branch
- [ ] `./gradlew installPD` should not have sync feature visible
- [ ] `./gradlew installID` should have sync feature visible

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/sync/FakeDeviceSyncState.kt
+++ b/app/src/main/java/com/duckduckgo/app/sync/FakeDeviceSyncState.kt
@@ -26,8 +26,7 @@ import javax.inject.*
  */
 @ContributesBinding(
     scope = AppScope::class,
-    // mitigation for https://app.asana.com/0/414730916066338/1204248963638410/f
-    priority = ContributesBinding.Priority.HIGHEST,
+    priority = ContributesBinding.Priority.NORMAL,
 )
 class FakeDeviceSyncState @Inject constructor() : DeviceSyncState {
     override fun isFeatureEnabled(): Boolean = false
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/sync/FakeDeviceSyncState.kt
```diff
@@ -26,8 +26,7 @@ import javax.inject.*
  */
 @ContributesBinding(
     scope = AppScope::class,
-    // mitigation for https://app.asana.com/0/414730916066338/1204248963638410/f
-    priority = ContributesBinding.Priority.HIGHEST,
+    priority = ContributesBinding.Priority.NORMAL,
```
aitorvs, line 29: Why do we need the `FakeDeviceSyncState`. Can we remove it and use `AppBuildConfig` in `AppDeviceSyncState` to enable it only for internal builds?

cmonfortep: Currently, we use `internalImplementation` for sync modules. We need this fake so we can compile other flavors.

That being said, even though I'm not confident, we will move to `implementation` this week, and test if that causes any issues when releasing into other stores. If no issues, we will remove the fake class.

malmstein (approved):
nice, 🚢

PR #3216, opened 2023-05-29
ADS: Allow set endIcon programmatically on TextInput component

nalcalag:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204667366884988/f

### Description
Added `fun setEndIcon(endIconRes: Int?, contentDescription: String? = null)` method to TextInput interface so we can edit it programmatically and not only from the xml file

### Steps to test this PR (Optional)

- Install from this branch
- Go to Settings > App Components Design Preview
- Select Text Input tab
- [ ] Check there is an end icon set on the 3rd item (Non-editable text with end icon)

### DIFF
--- a/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextInput.kt
+++ b/common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextInput.kt
@@ -55,9 +55,11 @@ interface TextInput {
     fun addTextChangedListener(textWatcher: TextWatcher)
     fun removeTextChangedListener(textWatcher: TextWatcher)
     fun setEndIcon(
-        @DrawableRes endIconRes: Int?,
+        @DrawableRes endIconRes: Int,
         contentDescription: String? = null,
     )
+
+    fun removeEndIcon()
 
     fun onAction(actionHandler: (Action) -> Unit)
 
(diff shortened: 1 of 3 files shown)


nalcalag:
* develop
  * **PR #3216** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3216" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3216?utm_source=stack-comment).

common-ui/src/main/res/layout/component_text_input_view.xml
```diff
@@ -58,8 +58,7 @@
             android:layout_marginTop="@dimen/keyline_3"
             android:hint="Non-editable text with end icon"
             android:text="This is not editable."
-            app:editable="false"
-            app:endIcon="@drawable/ic_copy" />
```
malmstein, line 62: why is this removed?

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextInput.kt
```diff
@@ -52,6 +54,10 @@ interface TextInput {
 
     fun addTextChangedListener(textWatcher: TextWatcher)
     fun removeTextChangedListener(textWatcher: TextWatcher)
+    fun setEndIcon(
+        @DrawableRes endIconRes: Int?,
```
malmstein, line 58: I don’t think we should allow this to be nullable

malmstein (approved):
Couple comments but nothing blocking, just the nullable piece that I think we should change

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextInput.kt
nalcalag: it's to remove the icon. Another approach could be to create a different method to `hideEndIcon()` but I think this is simpler.

common-ui/src/main/res/layout/component_text_input_view.xml
nalcalag: Good catch! `endIcon` is set programmatically with the new method. But I will bring back `editable="false"`

nalcalag: Actually, `editable="false"` is still there so the change it's just show endIcon programmatically so there is a use case in the preview which is intended.

common-ui/src/main/java/com/duckduckgo/mobile/android/ui/view/text/DaxTextInput.kt
nalcalag: Talk to @malmstein and we agreed on add `removeEndIcon()` so `endIconRes` is not nullable anymore

PR #3224, opened 2023-05-31
Measure impact of events in retention

nalcalag:
Task/Issue URL: https://app.asana.com/0/0/1204601774518723/f

### Description
Implemented new pixel which will send as parameters app events performed prior to the specified date. 

### Steps to test this PR

- smoke test

Asana task: https://app.asana.com/0/0/1204601774518724/f


app/build.gradle
```diff
@@ -346,6 +346,7 @@ dependencies {
     implementation "androidx.work:work-multiprocess:_"
     androidTestImplementation AndroidX.work.testing
     implementation AndroidX.work.rxJava2
+    testImplementation AndroidX.work.testing
```
nalcalag, line 349: FYI This is not needed anymore, I will remove it

statistics/src/main/java/com/duckduckgo/app/statistics/api/featureusage/FeatureSegmentType.kt
```diff
@@ -0,0 +1,30 @@
+/*
```
marcosholgado, line 1: Overall would have been nicer to extract the segment stuff into its own module

statistics/src/main/java/com/duckduckgo/app/statistics/api/featureusage/FeatureSegmentsPixelSender.kt
```diff
@@ -0,0 +1,50 @@
...
+package com.duckduckgo.app.statistics.api.featureusage
+
+import com.duckduckgo.app.statistics.api.RefreshRetentionAtbPlugin
+import com.duckduckgo.browser.api.UserBrowserProperties
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesMultibinding
+import javax.inject.Inject
+
+@ContributesMultibinding(AppScope::class)
+class FeatureSegmentsPixelSender @Inject constructor(
```
marcosholgado, line 26: Name is kinda weird, this implements `RefreshRetentionAtbPlugin` which is not related with pixels at all. Perhaps something like `SegmentsRefreshRetentionAtbPlugin` ?

app/src/main/java/com/duckduckgo/app/featureusage/FeatureSegmentsManagerImpl.kt
```diff
@@ -0,0 +1,112 @@
+/*
```
marcosholgado, line 1: As with the other comment, everything in this package is specific to the segments, not sure why it has to exist in the app module?

app/src/main/java/com/duckduckgo/app/featureusage/FeatureSegmentsManagerImpl.kt
```diff
@@ -0,0 +1,112 @@
...
+import com.duckduckgo.app.statistics.pixels.Pixel
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+
+@ContributesBinding(AppScope::class)
+class FeatureSegmentManagerImpl @Inject constructor(
+    private val featureSegmentsDataStore: FeatureSegmentsDataStore,
+    private val pixel: Pixel,
+) : FeatureSegmentsManager {
```
marcosholgado, line 42: This class is implementing the interface from the statistics module but it doesn't require anything specific from the app module?

statistics/src/main/java/com/duckduckgo/app/statistics/api/featureusage/FeatureSegmentType.kt
nalcalag: I thought about it but this is temporal, it will be live around 7-14 days

marcosholgado: Even if is temporary, the setup for a new module is trivial so there are no good reasons for not doing it right and instead add tech debt because we don't know if we will have to change the experiment or keep it for longer, etc. Plus when the time comes to delete the code you can delete the module altogether rather than searching for all the pieces.

statistics/src/main/java/com/duckduckgo/app/statistics/api/featureusage/FeatureSegmentsPixelSender.kt
nalcalag: I followed the same pattern we have for `FeatureRetentionPixelSender` and `DeviceShieldRetentionPixelSender` which have similar function and implement the same interface so I would recommend to leave this name to have consistency

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/VpnFeaturesRegistryImpl.kt
```diff
@@ -44,6 +47,7 @@ internal class VpnFeaturesRegistryImpl(
             putString(feature.featureName, UUID.randomUUID().toString())
         }
         vpnServiceWrapper.restartVpnService(forceRestart = true)
+        featureSegmentsManager.addUserToFeatureSegment(FeatureSegmentType.APP_TP_ENABLED)
```
aitorvs, line 50: @nalcalag I don't think this is the way to proceed. I know this is temporary code, but it's cluttering a lot of places we don't want it to clutter that will make difficult its removal later on

aitorvs (dismissed):

app/src/main/java/com/duckduckgo/app/featureusage/FeatureSegmentsManagerImpl.kt
nalcalag: I wouldn't create a module because it's going to be removed quite soon but probably makes sense to move it to statistics as everything is related to set and send the new pixel 👍

aitorvs:
@nalcalag, I am available to discuss the approach on the PR. This code is temporary but it is all over the place and I would like to understand if there's a better way that will allow its removal more easily.

app/src/main/java/com/duckduckgo/app/featureusage/FeatureSegmentsManagerImpl.kt
nalcalag: As I mentioned before, I will move it to `statistics` module

nalcalag:
@aitorvs I updated the description with the 10 defined events I need to send as pixel parameters so you can have a better understanding.
I understand your concern that's why I just added a line of code in all those places and extract the logic in the Manager class. However, happy to sync and check if we can find a less noisy way for enabling AppTP or other events.

aitorvs (commented):
All discussed, all good!

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
```diff
@@ -207,6 +211,7 @@ class TabSwitcherActivity : DuckDuckGoActivity(), TabSwitcherListener, Coroutine
             settingsDataStore = settingsDataStore,
             userEventsStore = userEventsStore,
             appCoroutineScope = appCoroutineScope,
+            featureSegmentsManager = featureSegmentsManager,
```
malmstein, line 214: is this needed?

malmstein (changes requested):
Setting the browser as default from Settings should also count, it doesn’t now. This is only working for internal builds, is that the intention?

malmstein (approved):
Talked and tested offline, looks good!

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
nalcalag: yes, we create an instance of the `FireDialog` here which needs the `featureSegmentsManager` as an argument

PR #3244, opened 2023-06-06
Optimize extractQuery function

abhishekdubey331:
### Description

- Refactor the extractQuery function to improve readability and performance. 
- Use the Elvis operator (?:) for early return and simplify null or blank checks with isNullOrBlank(). 
- Introduce a when block for condition evaluation. Update logging to remove unnecessary non-null checks.

### Steps to test this PR

Smoke test the app


abhishekdubey331:
@aitorvs It would be very much helpful if you can comment what was the issue with this pull request.

aitorvs:
Hey there, sure. There was no issue with the original code, and so we internally don't refactor code unless there's a clear reason behind it.

PR #3247, opened 2023-06-08
Make VpnFeaturesRegistryImpl calls async

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204763839445752/f

### Description
The `VpnFeaturesRegistry` functionality is backed by a multiprocess shared preferences but all its API is synchronous and called from various different threads, including the UI thread.

This PR:
* moves its API to be async using `suspend` and offload them to `io` dispatcher
* refactors the calling side when necessary

### Steps to test this PR

Aside from CI, perform the following sanity testing

- [x] smoke test AppTP and VPN
- [x] NP cohort should be assigned correctly
- [x] AppTP/VPN should work correctly, on/off, off/on, on/on
- [x] verify clean install and AppTP onboarding
- [x] verify clean install and VPN onboarding
- [x] verify AppTP / VPN notifications work correctly, on/off, off/on, on/on, off/off
- [x] verify re-key
- [x] verify VPN conflicts UX, during AppTP onboarding, AppTP activity screen and VPN management screen
- [x] verify `m_atp_ev_enabled_on_search_d/c`, `m_atp_ev_disabled_on_search_d/c`, `m_atp_ev_enabled_on_launch_c/d` and `m_atp_ev_disabled_on_launch_c/d` are sent correctly
- [x] verify AppTP quick setting (aka tile) works properly
- [x] verify AppTP can be enabled from disabled notification

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt
@@ -103,6 +103,7 @@ import com.duckduckgo.windows.api.ui.WindowsScreenWithEmptyParams
 import com.duckduckgo.windows.api.ui.WindowsWaitlistScreenWithEmptyParams
 import com.google.android.material.snackbar.Snackbar
 import javax.inject.Inject
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import timber.log.Timber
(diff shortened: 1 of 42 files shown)


app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt
```diff
@@ -294,6 +294,7 @@ class SettingsActivity : DuckDuckGoActivity() {
     private fun observeViewModel() {
         viewModel.viewState()
             .flowWithLifecycle(lifecycle, Lifecycle.State.RESUMED)
+            .distinctUntilChanged()
```
karlenDimla, line 297: @aitorvs  do we need this? The flow already doesn't emit duplicates, right?

aitorvs: We don't strictly need it, no strong opinions, I can remove it.
The only benefit is that without looking at the VM implementation it's clear we don't trigger any logic upon same emissions (which may not always be the case if we change something in the VM)

Lmk if you want me to remove it

karlenDimla (approved):
LGTM

app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt
karlenDimla: Let's keep it.

PR #3269, opened 2023-06-21
Retention Drivers Survey

nalcalag:
Task/Issue URL: https://app.asana.com/0/1201807753394693/1204669492338873

### Description
Two new features for survey feature:

- We also notify the user about an available survey through a notification
- New 2 url parameters have been added: `da` and `src`

### Steps to test this PR
ℹ️ For all tests, in `SurveyService` change endpoint for survey() to `https://ddg-sandbox.s3.amazonaws.com/survey/v2/survey-mobile.json`
ℹ️ For changing survey's `installationDay` go to `App Inspector` and from the table `survey` edit `daysInstalled` value

_Survey notification_
- Fresh install
- When navigate to browser screen:
- [x] Check in-app cta is showing
- [x] Check you receive the notification 5 seconds later

_Access survey from notification_
- Fresh install
- When navigate to browser screen a notification will show in 5 seconds
- Tap on survey notification
- [x] Check it navigates you to Survey screen

_Finish Survey accessing from notification_
- Fresh install
- When navigate to browser screen a notification will show in 5 seconds
- Tap on survey notification
- Complete survey
- [ ] Check survey cta and notification are dismiss

_Finish Survey accessing from CTA_
- Fresh install
- When navigate to browser screen the survey cta will appear
- Tap on survey CTA
- Complete survey
- [ ] Check survey cta and notification are dismiss

_Survey _installationDay_ != 0_
- Fresh install
- Navigate to browser screen
- [ ] Check in-app cta is NOT showing
- [ ] Check you DON'T receive the notification 5 seconds later
- [ ] Check you receive the notification in day `installationDay` _(optional)_

_Survey _installationDay_ == -1 (survey targeting all users)_
- Fresh install
- Navigate to browser screen
- [ ] Check in-app cta is showing
- [ ] Check you receive the notification 5 seconds later

_Survey _installationDay_ == null (survey targeting users who installed the app >= 30 days)_
- Fresh install
- Navigate to browser screen
- **IF install day >= 30**
- [ ] Check in-app cta is showing
- [ ] Check you receive the notification 5 seconds later
- **IF install day < 30**
- [ ] Check in-app cta is NOT showing
- [ ] Check you DON'T receive the notification 5 seconds later


nalcalag:
* develop
  * **PR #3269** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3269" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3269?utm_source=stack-comment).

app/src/main/java/com/duckduckgo/app/notification/AndroidNotificationScheduler.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 DuckDuckGo
+ * Copyright (c) 2023 DuckDuckGo
```
aitorvs, line 2: Nit, can we undo this change?

app/src/main/java/com/duckduckgo/app/survey/api/SurveyDownloader.kt
```diff
@@ -86,7 +87,9 @@ class SurveyDownloader @Inject constructor(
             }
 
             newSurvey?.let {
-                surveyDao.insert(newSurvey)
+                if (surveyRepository.isUserEligibleForSurvey(newSurvey)) {
+                    surveyDao.insert(newSurvey)
```
aitorvs, line 91: Seems odd to have a survey repository and then have to also go through the DAO

app/src/main/java/com/duckduckgo/app/survey/api/SurveyRepository.kt
```diff
@@ -0,0 +1,91 @@
...
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import java.util.*
+import javax.inject.Inject
+
+interface SurveyRepository {
+    fun isUserEligibleForSurvey(survey: Survey): Boolean
+    fun remainingDaysForShowingSurvey(survey: Survey): Long
+    fun getScheduledSurvey(): Survey
+    fun clearSurveyNotification()
```
aitorvs, line 33: This is confusing, we have a `clearSurveyNotification` but I can't see in this API where the notification is created

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt
```diff
@@ -146,13 +149,21 @@ class SurveyActivity : DuckDuckGoActivity() {
...
             return intent
         }
 
         const val SURVEY_EXTRA = "SURVEY_EXTRA"
+        const val SURVEY_SOURCE_EXTRA = "SURVEY_SOURCE_EXTRA"
+
+        enum class SurveySource {
+            PUSH,
+            IN_APP,
+        }
```
aitorvs, line 166: Now that we were touching this, it'd have been great to move it to use the `GlobalActivityStarter`

app/src/main/java/com/duckduckgo/app/usage/app/AppDaysUsedDao.kt
```diff
@@ -30,6 +30,9 @@ abstract class AppDaysUsedDao {
     @Query("SELECT COUNT(*) from app_days_used WHERE date > :isoDate")
     abstract fun getNumberOfDaysAppUsedSince(isoDate: String): Long
 
-    @Insert(onConflict = OnConflictStrategy.IGNORE)
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
```
aitorvs, line 33: This is changing the previous behavior, was it a bug before?

app/src/test/java/com/duckduckgo/app/survey/ui/SurveyViewModelTest.kt
```diff
@@ -64,22 +68,31 @@ class SurveyViewModelTest {
...
-
-        testee = SurveyViewModel(
-            mockSurveyDao,
-            mockStatisticsStore,
-            mockAppInstallStore,
-            mockAppBuildConfig,
-            coroutineTestRule.testDispatcherProvider,
-        )
-        testee.command.observeForever(mockCommandObserver)
+        runTest {
```
aitorvs, line 81: Why is everything under `runTest`? I guess this is only because of `getLastActiveDay` and probably we should use `runBlocking` instead

app/src/main/java/com/duckduckgo/app/usage/app/AppDaysUsedDao.kt
nalcalag: It wasn't a bug before because we only had `date` column. However, the new approach with the second column needs REPLACE strategy as we need to update `previous_date` with the latest value.

app/src/main/java/com/duckduckgo/app/survey/api/SurveyRepository.kt
nalcalag: I understand your concern. There were few options on how to approach this:

1. Inject `notificationManager` to SurveyViewModel directly so I can clear it from there

- It could work, but I believe it's better to have it in the Repository in case we want to perform different actions in the future, such as launching a pixel or interacting with other survey feature classes. This will ensure consistency if the function is called from different places in the future as well.

2. Add the function to the SurveyNotificationScheduler API and inject `notificationManager` there

- Because the class is a Scheduler instead a notification manager I thought adding the function here could be a bit confusing. However, it maybe makes more sense.

3. Inject `notificationManager` in the Repository like I did

- I went for this approach because as I explained in point 1, it's a way to unify the same behaviour for clearing survey notification from different places in the code. Also, it would be easier if we want to add a pixel or put survey in their own module in the future as a public API. However, it could also scale if I add the fun in the SurveyNotificationScheduler if you think it makes more sense @aitorvs . 

As a summery, I didn't add `clearNotification()` to the `NotificationScheduler` because it was a bit confusing for me as it is a Scheduler instead a Manager. But on the other hand, as you mentioned, it is also confusing to have the clearNotification function in the Repository and not having the create one. So I would agree on moving it to the NotificationScheduler if you think it's a better approach.

aitorvs: No strong opinions, javadocs work as well. At least to understand what are we clearing and where it originates

malmstein: @nalcalag  check https://github.com/duckduckgo/Android/blob/develop/saved-sites/saved-sites-api/src/main/java/com/duckduckgo/savedsites/api/SavedSitesRepository.kt for api docs

app/src/main/java/com/duckduckgo/app/survey/api/SurveyDownloader.kt
nalcalag: I do it through the Repository to have an extra data layer for Survey. The way I always worked was not letting ViewModels, Presenters, etc have the Dao directly injected but having the feature UseCase or Repository as an intermediate. I know the Dao is still injected in some classes as I only added here the functions needed for the new implementation or small refactor I made. 
The idea would be to refactor survey feature and use the Repository as a public Api instead accessing the Dao directly. Does it make sense @aitorvs ?

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt
nalcalag: Follow-up task -> https://app.asana.com/0/0/1204888830157412/f

app/src/main/java/com/duckduckgo/app/usage/app/AppDaysUsedDao.kt
```diff
@@ -30,6 +30,9 @@ abstract class AppDaysUsedDao {
...
     abstract fun insert(appUsedEntity: AppDaysUsedEntity)
+
+    @Query("SELECT * FROM app_days_used ORDER BY date DESC LIMIT 1")
```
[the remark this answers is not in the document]
nalcalag: @malmstein could you write the Query you are suggesting? I'm not getting what you mean, sorry. 
`date` is the @PrimaryKey so I just order by latest so I get the one I want, but I could be wrong and could be a better way.

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt
```diff
@@ -99,6 +101,7 @@ class SurveyActivity : DuckDuckGoActivity() {
     private fun loadSurvey(url: String) {
         binding.progress.show()
         webView.loadUrl(url)
+        Timber.d("Survey URL: $url")
```
malmstein, line 104: should this be removed?

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyViewModel.kt
```diff
@@ -52,12 +57,21 @@ class SurveyViewModel @Inject constructor(
...
-    fun start(survey: Survey) {
+    fun start(survey: Survey, source: SurveySource) {
         val url = survey.url ?: return
         this.survey = survey
-        command.value = Command.LoadSurvey(addSurveyParameters(url))
+        this.source = source
+        viewModelScope.launch {
+            lastActiveDay = when (source) {
+                SurveySource.IN_APP -> appDaysUsedRepository.getLastActiveDay()
+                SurveySource.PUSH -> appDaysUsedRepository.getPreviousActiveDay() ?: appDaysUsedRepository.getLastActiveDay()
```
malmstein, line 71: how about `getPreviousActiveDay` returns  `getLastActiveDay` if null, so this logic in not in the ViewModel?

app/src/main/java/com/duckduckgo/app/usage/app/AppDaysUsedDao.kt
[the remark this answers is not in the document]
malmstein: nevermind, the format fixes it.

malmstein (approved):
looks good, once you have the things we’ve discussed offline we can merge this in.

app/src/main/java/com/duckduckgo/app/survey/notification/SurveyAvailableNotification.kt
```diff
@@ -0,0 +1,120 @@
...
+        val deferred = coroutineScope.async(dispatcherProvider.io()) {
+            schedulableNotification.buildSpecification()
+        }
+        return runBlocking {
+            deferred.await()
+        }
+    }
+
+    override fun getLaunchIntent(): PendingIntent? {
+        val intent = SurveyActivity.intent(context, surveyRepository.getScheduledSurvey(), SurveySource.PUSH)
```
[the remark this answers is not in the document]
nalcalag: Yeah, I think it should. I will add it the pixel

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyViewModel.kt
nalcalag: I wouldn't like to add this logic in the Repository because getting lastDay instead, is specific to this scenario. In other scenarios (if any) we would like to return null so we know there isn't a previous date. Does it make sense?

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt
nalcalag: yes, I will remove it as I added it for debug purposes and forgot to remove it.

app/src/main/java/com/duckduckgo/app/survey/ui/SurveyViewModel.kt
malmstein: no strong opinions, sounds good!

app/src/main/java/com/duckduckgo/app/global/db/AppDatabase.kt
```diff
@@ -610,6 +610,12 @@ class MigrationsProvider(val context: Context, val settingsDataStore: SettingsDa
         }
     }
 
+    private val MIGRATION_47_TO_48: Migration = object : Migration(47, 48) {
+        override fun migrate(database: SupportSQLiteDatabase) {
+            database.execSQL("ALTER TABLE `app_days_used` ADD COLUMN `previous_date` TEXT")
```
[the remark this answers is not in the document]
nalcalag: We don't require a default value. I'll leave it as null since the parameter is optional, ensuring that no crashes will happen as a result.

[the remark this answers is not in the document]
nalcalag: There are more migration with this pattern like `1_TO_2`, `2_TO_3`, `13_TO_14`, `46_TO_47`

PR #3278, opened 2023-06-26
Add content scope scripts feature toggle

joshliebe:
Task/Issue URL: https://app.asana.com/0/488551667048375/1204659079168053/f

### Description
This PR adds the ability to enable/disable Content Scope Scripts via remote config.

### Steps to test this PR
- [x] Go to https://privacy-test-pages.glitch.me/privacy-protections/fingerprinting/
- [x] Verify: `navigator.getBattery() - (object) - { "charging": true, "chargingTime": 0, "level": 1 }`
- [x] Override the remote config URL with the JSON blob linked in the task
- [x] Restart the app
- [x] Go to https://privacy-test-pages.glitch.me/privacy-protections/fingerprinting/
- [x] Verify: `navigator.getBattery() - (object) - { "charging": false, "chargingTime": null, "level": 1 }` (Or the actual battery status if using a physical device)

### DIFF
--- a/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/RealContentScopeScripts.kt
+++ b/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/RealContentScopeScripts.kt
@@ -34,6 +34,7 @@ import javax.inject.Inject
 
 interface CoreContentScopeScripts {
     fun getScript(): String
+    fun isEnabled(): Boolean
 }
 
 @SingleInstanceIn(AppScope::class)
@@ -45,6 +46,7 @@ class RealContentScopeScripts @Inject constructor(
     private val appBuildConfig: AppBuildConfig,
     private val unprotectedTemporary: UnprotectedTemporary,
     private val fingerprintProtectionManager: FingerprintProtectionManager,
+    private val contentScopeScriptsFeature: ContentScopeScriptsFeature,
 ) : CoreContentScopeScripts {
 
     private var cachedContentScopeJson: String = getContentScopeJson("", emptyList())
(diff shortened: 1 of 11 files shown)


joshliebe:
* develop
  * **PR #3278** <a href="https://app.graphite.dev/github/pr/duckduckgo/android/3278" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/android/3278?utm_source=stack-comment).

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/RealContentScopeScripts.kt
```diff
@@ -34,6 +34,7 @@ import javax.inject.Inject
 
 interface CoreContentScopeScripts {
     fun getScript(): String
+    fun isEnabled(): Boolean
```
marcosholgado, line 37: I have mixed feelings here. From one side we don't really need the `isEnabled()` interface in more places because we already have `contentScopeScriptsFeature.self().isEnabled()` which is the same thing, so this is a pass through.

On the other hand, not having this interface means adding an extra dependency in places like `BrowserWebClient`

I think we need to change the API and make this an implementation detail of it.

Rather than having `fun getScript(): String` in the `ContentScopeScripts` API. I think we should change it to `fun `injectContentScopeScripts(webview: WebView)` or similar and move all the logic to check if the toggle is enabled or not inside the implementation. That way we wouldn't return the script but rather directly inject it (or not) in the given webview instance.

marcosholgado (changes requested):
Have been mulling over something that was a bit annoying, left a comment with something that I think we need to change.

marcosholgado (approved):
LGTM!

PR #3291, opened 2023-06-29
Github Action: Update Content scope and Reference Tests

malmstein:
Task/Issue URL: 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/.github/workflows/update-content-scope.yml
+++ b/.github/workflows/update-content-scope.yml
@@ -11,9 +11,10 @@ concurrency:
 
 jobs:
   check-ref-tests:
-    name: Check Content Scope Scripts
+    name: Check Content Scope scripts
     runs-on: ubuntu-latest
-
+    outputs:
+      outcome: ${{ steps.update-check.outputs.outcome }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
@@ -32,9 +33,11 @@ jobs:
         run: ncu @duckduckgo/content-scope-scripts --errorLevel 2
 
   update-ref-tests:
-    - name: Update content scope scripts
-      if: ${{ steps.update-check.outcome == 'failure' }}
+    - name: Update Content Scope scripts
+      if: ${{ needs.check-ref-tests.outputs.outcome == 'failure' }}
       needs: check-ref-tests
+      outputs:
+        version: ${{ steps.latest-release.outputs.version }}
       steps:
         - name: Update content scope scripts
           run: ncu @duckduckgo/content-scope-scripts -u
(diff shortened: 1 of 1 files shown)


marcosholgado (changes requested):
Both workflows fail:

The workflow is not valid. .github/workflows/update-ref-tests.yml (Line: 51, Col: 12): Job 'find-latest-version' depends on unknown job 'update-ref-tests'. .github/workflows/update-ref-tests.yml (Line: 68, Col: 13): Job 'process-update' depends on unknown job 'update-ref-tests'.

The workflow is not valid. .github/workflows/update-content-scope.yml (Line: 68, Col: 13): Job 'process-update' depends on unknown job 'update-content-scope-scripts'.

aitorvs:
@malmstein the only thing we need here is to add the version to the Asana task. Is all the refactor this PR does required just to add a version in the asana task?
If it is not required, please remove all the refactoring

.github/workflows/update-ref-tests.yml
```diff
@@ -64,29 +89,47 @@ jobs:
...
             If tests failed check out https://app.asana.com/0/0/1203766026095653/f for further information on what to do next.
             
             See ${{ steps.create-pr.outputs.pull-request-url }}
+          action: 'create-asana-task'
+
+      - name: Add Asana task to Release Board project when workflow succeeded
+        uses: malmstein/github-asana-action@master
+        with:
+          asana-pat: ${{ secrets.GH_ASANA_SECRET }}
+          asana-project: '488551667048375'
```
marcosholgado, line 111: All these should be a secret

.github/workflows/update-content-scope.yml
```diff
@@ -64,29 +89,47 @@ jobs:
...
             If tests failed check out https://app.asana.com/0/1202561462274611/1203986899650836/f for further information on what to do next.
             
             See ${{ steps.create-pr.outputs.pull-request-url }}
+          action: 'create-asana-task'
+
+      - name: Add Asana task to Release Board project when workflow succeeded
+        uses: malmstein/github-asana-action@master
+        with:
+          asana-pat: ${{ secrets.GH_ASANA_SECRET }}
+          asana-project: '488551667048375'
```
marcosholgado, line 111: Should be a secret

.github/workflows/update-ref-tests.yml
malmstein: agreed to move use vars instead of secrets, we don’t need to encrypt these

.github/workflows/update-content-scope.yml
malmstein: ditto

.github/workflows/update-content-scope.yml
```diff
@@ -32,27 +32,37 @@ jobs:
         run: ncu @duckduckgo/content-scope-scripts --errorLevel 2
 
       - name: Update content scope scripts
-        if: ${{ steps.update-check.outcome == 'failure' }}
+        if: ${{ steps.update-check.outputs.outcome == 'failure' }}
```
marcosholgado, line 35: `output` is not correct, it should only be `outcome`, same for all the other steps in both workflows when you refer to the whole step failing.

.github/workflows/update-ref-tests.yml
```diff
@@ -63,30 +73,52 @@ jobs:
...
+        if: ${{ steps.update-check.outputs.outcome == 'failure' }}
+        uses: malmstein/github-asana-action@master
+        with:
+          asana-pat: ${{ secrets.GH_ASANA_SECRET }}
+          asana-project: ${{ vars.GH_ANDROID_RELEASE_BOARD_PROJECT_ID }}
+          asana-section: ${{ vars.GH_ANDROID_RELEASE_BOARD_PR_SECTION_ID }}
+          asana-task-id: ${{ steps.create-task.outputs.taskId }}
+          action: 'add-task-asana-project'
+
+      - name: Update PR description with Asana task
```
marcosholgado, line 103: This step failed with

```
calling add-task-pr-description
/home/runner/work/_actions/malmstein/github-asana-action/master/node_modules/@octokit/request/dist-node/index.js:122
      const error = new import_request_error.RequestError(toErrorMessage(data), status, {
```

See error in https://github.com/duckduckgo/Android/actions/runs/5452587649/jobs/9920294125

.github/workflows/update-content-scope.yml
malmstein: It’s failing because there is no PR to modify. You executed the action manually, not via a PR. The if clause is correct

.github/workflows/update-ref-tests.yml
malmstein: Same, there’s no PR to update, that’s why the step fails.

.github/workflows/update-content-scope.yml
marcosholgado: This workflow is executed nightly or manually, I don't see why we would need a PR to be able to execute it? The if clause can never work because outcome is not part of the outputs of a step, if you check the diff you will see it didn't have that before.

malmstein: when you are right, you are right

.github/workflows/update-ref-tests.yml
malmstein: when you are right, you are right

marcosholgado (approved):
👍

PR #3360, opened 2023-07-19
Restrict dialogs showing when triggered from inactive tabs

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205077297202827/f

### Description
Restricts background tabs from showing user-facing dialogs (which would otherwise appear over the active tab)

### Steps to test this PR

_Testing Javascript-driven dialogs_
- [x] Visit https://jsfiddle.net/cdrussell/xwqbmtgL/35/show
- [x] Tap on some of the buttons, both the "immediately" ones and the delayed ones **while staying on the same tab**
- [x] Ensure all the dialogs appear as they normally would (the rough UI for these dialogs is nothing new in this PR)
- [x] Now tap on one of the "after 5 seconds" buttons and quickly switch to another tab. 
- [x] Wait >5 seconds
- [x] Verify the dialog doesn't show over the new tab

_Testing native-driven dialogs_
- [x] Visit https://jsfiddle.net/cdrussell/fmey4cgt/show
- [x] Tap on the "exploit me" button
- [x] Verify no dialog shows over the new tab that is launched

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserChromeClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserChromeClient.kt
@@ -163,15 +163,7 @@ class BrowserChromeClient @Inject constructor(
         url: String,
         message: String,
         result: JsResult,
-    ): Boolean {
-        if (webViewClientListener?.isActiveTab() == true) {
-            return false
-        }
-
-        Timber.v("onJsAlert called but is not the active tab; suppressing dialog")
-        result.cancel()
-        return true
-    }
+    ): Boolean = shouldSuppressJavascriptDialog(result)
 
     /**
      * Called when a site's javascript tries to create a javascript prompt dialog
@@ -183,15 +175,7 @@ class BrowserChromeClient @Inject constructor(
         message: String?,
         defaultValue: String?,
         result: JsPromptResult,
-    ): Boolean {
-        if (webViewClientListener?.isActiveTab() == true) {
-            return false
-        }
-
-        Timber.v("onJsPrompt called but is not the active tab; suppressing dialog")
-        result.cancel()
-        return true
-    }
+    ): Boolean = shouldSuppressJavascriptDialog(result)
 
     /**
      * Called when a site's javascript tries to create a javascript confirmation dialog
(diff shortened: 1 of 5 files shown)


CDRussell:
* develop
  * **PR #3360** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3360" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3360?utm_source=stack-comment).

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1299,6 +1299,11 @@ class BrowserTabFragment :
     }
 
     private fun askSiteLocationPermission(domain: String) {
+        if (!isActiveTab) {
+            Timber.v("Will not launch a dialog for an inactive tab")
+            return
+        }
+
```
nalcalag, line 1306: NIT: I would suggest moving this block of code (and the same 5 ones) to the ViewModel so it can be checked before getting here and tested independently so we don't rely on the fragment for checking it.

app/src/main/java/com/duckduckgo/app/browser/BrowserChromeClient.kt
```diff
@@ -154,6 +154,64 @@ class BrowserChromeClient @Inject constructor(
...
+        message: String,
+        result: JsResult,
+    ): Boolean {
+        if (webViewClientListener?.isActiveTab() == true) {
+            return false
+        }
+
+        Timber.v("onJsAlert called but is not the active tab; suppressing dialog")
+        result.cancel()
+        return true
```
nalcalag, line 173: NIT: I would extract this code in a private function so you can reuse it in all callbacks.

nalcalag (approved):
It works great! 💯  Just a couple of NIT suggestions but nothing blocking!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
CDRussell: I think there's a small risk in moving it to the view model that it's more possible when the view model checks, it is the active tab, but there's some small async work happens before the dialog is about to be shown and by then, it's no longer the active tab. Keeping it right at the point of showing the dialog minimises that chance.

Would rather leave them here for now.

PR #3361, opened 2023-07-19
Assemble issue after gradle upgrade - noticed in Jenkins, can be reproduced locally

anikiki:
Task/Issue URL: https://app.asana.com/0/1200905986587319/1204834909669704/f

### Description
The `httpsupgrade` package from the `app` module was extracted to its own module called `httpsupgrade` (api, impl, store). Thanks @aitorvs for making the changes.

### Steps to test this PR

Test fresh install and upgrade (there is a db migration and we need to make sure there are no issues).
Also see https://app.asana.com/0/1200905986587319/1204834909669704/f


aitorvs (approved):
Tested fresh install and upgrade. All good!

PR #3368, opened 2023-07-20
fix gradle deamon memory struggling

kostasfilios:
### Description
I noticed a big memory usage from android studio after some builds, 
thourgh VisualVM i noticed that Gradle Deamon stays idle for a long time after build and keeps also reference to 
org.jetbrains.kotlin.daemon.KotlinCompileDaemon. 
Each one keeps up to 4gb ram 

### Steps to test this PR
Try multiple build and each time you can notice that at build time studio requires up to 14gb ram but after process finish 
returns to normal memory usage


cmonfortep:
Closing as per contribution guidelines. Contributions should be related to an accepted issue (label Help wanted). Thanks.

PR #3388, opened 2023-07-28
Fix internal dependencies and add gradle checks

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205153634789947/f 

### Description
This PR fixes a wrong use of an internal module that was using implementation instead of internalImplementation. It also adds a check so it doesn't happen again.

### Steps to test this PR

- [x] Using playDebug sync gradle, no error should be shown
- [x] Change any internal dependency to use `implementation` instead of `internalImplementation`
- [x] Sync gradle and you should see an error
- [x] Changing to internalDebug should always build regardless of using `implementation` or `internalImplementation` but the error will be picked up during CI.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,19 +83,21 @@ subprojects {
     }
 
     def projectPath = path
-    configurations.all {
+    configurations.configureEach {
         if (name == "compileClasspath" || name.endsWith("CompileClasspath")) {
             incoming.beforeResolve {
                 for (dependency in dependencies) {
                     if (dependency instanceof ProjectDependency) {
                         def dependencyPath = dependency.dependencyProject.path
                         if (dependencyPath == projectPath) continue
-                        // vpn-internal is part of vpn-impl
-                        if (projectPath.endsWith(':vpn-internal') && dependencyPath.endsWith(':vpn-impl')) continue
-                        // network-protection-internal is part of network-protection-impl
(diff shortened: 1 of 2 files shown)


marcosholgado:
* develop
  * **PR #3388** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3388" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3388?utm_source=stack-comment).

build.gradle
```diff
@@ -83,19 +83,21 @@ subprojects {
...
-    configurations.all {
+    configurations.configureEach {
         if (name == "compileClasspath" || name.endsWith("CompileClasspath")) {
             incoming.beforeResolve {
                 for (dependency in dependencies) {
                     if (dependency instanceof ProjectDependency) {
                         def dependencyPath = dependency.dependencyProject.path
                         if (dependencyPath == projectPath) continue
-                        // vpn-internal is part of vpn-impl
-                        if (projectPath.endsWith(':vpn-internal') && dependencyPath.endsWith(':vpn-impl')) continue
```
marcosholgado, line 94: I removed all the specific rules so internal can depend on impl and generalized it.

aitorvs (approved):

PR #3421, opened 2023-08-05
Add maestro tests for autofill flows

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203874354762345/f

### Description
Adds maestro tests for main autofill flows.

Note, only a subset of the flows are suitable for a general smoke test as at a certain point it relies on the device/emulator having a PIN/Password set. 

`.maestro/autofill/smoke.yaml`: suitable for general purpose smoke testing (will ensure login option appears in the overflow)
`.maestro/autofill/0_all.yaml`: suitable only if the device has device-level authentication configured

#### General purpose smoke test
- [x] Run `maestro test .maestro/autofill/smoke.yaml `
- [x] Should all pass

#### With device-auth set
- [x] Ensure you have set a system-level PIN/Password
- [x] Run `maestro test .maestro/autofill/0_all.yaml`
- [x] Should all pass

### DIFF
--- a/.maestro/autofill/0_all.yaml
+++ b/.maestro/autofill/0_all.yaml
@@ -1,0 +1,12 @@
+appId: com.duckduckgo.mobile.android
+name: Autofill - Run all tests
+---
+- launchApp:
+    clearState: true
+- runFlow: ../shared/onboarding.yaml
+- runFlow: 1_autofill_shown_in_overflow.yaml
+
+  # Everything below requires a device has device-level authentication set (e.g., a PIN/Password etc...)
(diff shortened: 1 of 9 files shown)


CDRussell:
@aitorvs, I have some thoughts around smoke testing and how maestro tests could be structured etc... that i can discuss with you next week. But finished tidying up the maestro tests for autofill that I'd hadn't pushed before. I plan to add them in using this PR, then better make sure they're run during smoke testing separately next week.

aitorvs (approved):

PR #3423, opened 2023-08-07
Sync: M2Pixels

malmstein:
Task/Issue URL:  https://app.asana.com/0/488551667048375/1204998012054462/f

### Description
This PR adds the necessary pixels for Sync M2

### Steps to test this PR

_DAUs_
Modify FeatureRetentionPixelSender so it sends pixels even if they same day was already sent.
Remove the if statement in line 63

- [ ] Open app and enable sync
- [ ] Force close the app and open it again
- [ ] Verify m_browser_feature_daily_active_user_d is sent with parameter sync = 1

_Daily Success Rate_
Modify SyncDailyReportingWorkerScheduler so it schedules the Worker every minute instead of every 24 hours
Modify line 45 in SyncStatsRepository so it doesn’t filter by sync operations of the day before

 - [ ]Open app and enable sync
- [ ] Do some changes in syncable objects so sync operations are triggered
 - [ ]Force close the app and open it again
- [ ] Verify m_sync_daily_success_rate is sent
- [ ] Verify m_sync_daily_attempts is sent

_Encryption_
Modify RealSyncCrypto so it always fails when encrypting / decrypting

- [ ] Open app and enable sync
- [ ] Do some changes in syncable objects so sync operations are triggered
- [ ] Verify m_sync_encrypt_failure is sent

_Decryption (needs two devices)_

- [ ] Open app and add some bookmarks
- [ ] Make some changes in first device
- [ ] Join the first device’s account
- [ ] Verify m_sync_decrypt_failure is sent

### DIFF
--- a/saved-sites/saved-sites-impl/src/main/java/com/duckduckgo/savedsites/impl/sync/algorithm/SavedSitesSyncPersisterAlgorithm.kt
+++ b/saved-sites/saved-sites-impl/src/main/java/com/duckduckgo/savedsites/impl/sync/algorithm/SavedSitesSyncPersisterAlgorithm.kt
@@ -43,7 +43,7 @@ interface SavedSitesSyncPersisterAlgorithm {
     fun processEntries(
         bookmarks: SyncBookmarkEntries,
         conflictResolution: SyncConflictResolution,
-    ): SyncMergeResult<Boolean>
+    ): SyncMergeResult
 }
 
 @ContributesBinding(AppScope::class)
@@ -58,7 +58,9 @@ class RealSavedSitesSyncPersisterAlgorithm @Inject constructor(
     override fun processEntries(
         bookmarks: SyncBookmarkEntries,
         conflictResolution: SyncConflictResolution,
-    ): SyncMergeResult<Boolean> {
+    ): SyncMergeResult {
+        var orphans = false
+
         val processIds: MutableList<String> = mutableListOf()
         val allResponseIds = bookmarks.entries.filterNot { it.deleted != null }.map { it.id }
         val allFolders = bookmarks.entries.filter { it.isFolder() }.filterNot { it.id == SavedSitesNames.FAVORITES_ROOT }
@@ -106,9 +108,12 @@ class RealSavedSitesSyncPersisterAlgorithm @Inject constructor(
         }
 
         val unprocessedIds = allResponseIds.filterNot { processIds.contains(it) }
-        Timber.d("Sync-Feature: there are ${unprocessedIds.size} items orphaned $unprocessedIds")
+        if (unprocessedIds.isNotEmpty()) {
+            orphans = true
+            Timber.d("Sync-Feature: there are ${unprocessedIds.size} items orphaned $unprocessedIds")
+        }
 
-        return SyncMergeResult.Success(true)
+        return SyncMergeResult.Success(orphans = orphans)
     }
 
     private fun processFolder(
--- a/saved-sites/saved-sites-impl/src/main/java/com/duckduckgo/savedsites/impl/sync/algorithm/SavedSitesSyncPersisterAlgorithm.kt
+++ b/saved-sites/saved-sites-impl/src/main/java/com/duckduckgo/savedsites/impl/sync/algorithm/SavedSitesSyncPersisterAlgorithm.kt
@@ -61,7 +61,7 @@ class RealSavedSitesSyncPersisterAlgorithm @Inject constructor(
     ): SyncMergeResult {
         var orphans = false
 
-        val processIds: MutableList<String> = mutableListOf()
+        val processIds: MutableList<String> = mutableListOf(SavedSitesNames.BOOKMARKS_ROOT)
         val allResponseIds = bookmarks.entries.filterNot { it.deleted != null }.map { it.id }
         val allFolders = bookmarks.entries.filter { it.isFolder() }.filterNot { it.id == SavedSitesNames.FAVORITES_ROOT }
         val allFolderIds = allFolders.map { it.id }
@@ -79,6 +79,9 @@ class RealSavedSitesSyncPersisterAlgorithm @Inject constructor(
         // check all children, the ones that are not in allFolders don't have a parent
         val foldersWithoutParent = allFolderIds.filterNot { allChildren.contains(it) }
         foldersWithoutParent.forEach { folderId ->
+            if (repository.getFolder(folderId) != null) {
+                processIds.add(folderId)
+            }
             processFolder(folderId, "", bookmarks.entries, bookmarks.last_modified, processIds, conflictResolution)
         }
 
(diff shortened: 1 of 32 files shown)


malmstein:
* develop
  * **PR #3423** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3423" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3423?utm_source=stack-comment).

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixels.kt
```diff
@@ -0,0 +1,118 @@
...
+        )
+        pixel.fire(
+            SyncPixelName.SYNC_DAILY_ATTEMPTS,
+            mapOf(
+                SyncPixelParameters.ATTEMPTS to dailyStats.attempts.toString(),
+            ),
+        )
+    }
+
+    override fun fireMergeConflictPixel(feature: Feature) {
```
cmonfortep, line 60: Proposal to accept `String` instead of `Feature`, so we can directly use `SyncableType.field`.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixels.kt
```diff
@@ -0,0 +1,118 @@
...
+}
+
+object SyncPixelParameters {
+    const val ATTEMPTS = "attempts"
+    const val RATE = "rate"
+    const val FEATURE = "feature"
+}
+
+object SyncPixelValues {
+    sealed class Feature {
```
cmonfortep, line 113: Proposal to use instead `SyncableType.field`

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixels.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.duckduckgo.sync.impl.stats.SyncStatsRepository
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+
+interface SyncPixels {
+    fun fireStatsPixel()
+
+    fun fireMergeConflictPixel(feature: Feature)
+
+    fun fireOrphanPresentPixel(feature: Feature)
```
cmonfortep, line 31: This method seems to be unused.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixels.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.sync.impl.pixels.SyncPixelValues.Feature
+import com.duckduckgo.sync.impl.stats.SyncStatsRepository
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+
+interface SyncPixels {
+    fun fireStatsPixel()
+
+    fun fireMergeConflictPixel(feature: Feature)
```
cmonfortep, line 29: This one seems to be unused.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/stats/SyncStatsRepository.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun getDailyStats(): DailyStats
+}
+
+data class DailyStats(
+    val attempts: Int,
+    val successRate: Double,
+)
+
+class RealSyncStatsRepository @Inject constructor(private val syncStateRepository: SyncStateRepository) : SyncStatsRepository {
+    override fun getDailyStats(): DailyStats {
```
cmonfortep, line 43: To confirm I understand this pixel:

I assume this pixel will report metrics about sync for each day. So the goal here is to know the attempts / rate / other data for a specific day.

I see we rely on work manager to trigger the pixel every 24h. I believe that will not give us what we want. Imagine workmanager triggers the worker every 24h at 9am. Then filtering by today will give us the attempts so far until that moment, but next day we will lose anything that happened the previous day between 9am until EoD. Is my assumption correct?

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncDailyReportingWorker.kt
```diff
@@ -0,0 +1,79 @@
...
+    boundType = MainProcessLifecycleObserver::class,
+)
+class SyncDailyReportingWorkerScheduler @Inject constructor(
+    private val workManager: WorkManager,
+) : MainProcessLifecycleObserver {
+
+    override fun onCreate(owner: LifecycleOwner) {
+        super.onCreate(owner)
+        Timber.v("Scheduling sync daily reporting worker")
+        val workerRequest = PeriodicWorkRequestBuilder<SyncDailyReportingWorker>(24, HOURS)
```
cmonfortep, line 69: there's a question about scheduling every 24h here and then filtering by `today()`.

cmonfortep (commented):
Took a first pass, left some comments we discussed offline.

Also discussed offline, to add a failure pixel that is triggered for each feature when sync fails. Until syncstaterepository doesn't support multiple features, it's the easiest approach to have some visibility if a feature fails to sync.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/stats/SyncStatsRepository.kt
malmstein: I had this right in the API javadoc but not in the code. What we really want to send is yesterday’s statistics, not today. Your assumption is correct, we can’t rely on WorkManager to give us today’s data, but we can use WorkManager to schedule every 24 hours, and then get the data from the day before. That way we know that all data from the previous day has been properly stored.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncDailyReportingWorker.kt
malmstein: answered above

saved-sites/saved-sites-impl/src/main/java/com/duckduckgo/savedsites/impl/sync/algorithm/SavedSitesSyncPersisterAlgorithm.kt
```diff
@@ -58,7 +58,9 @@ class RealSavedSitesSyncPersisterAlgorithm @Inject constructor(
     override fun processEntries(
         bookmarks: SyncBookmarkEntries,
         conflictResolution: SyncConflictResolution,
-    ): SyncMergeResult<Boolean> {
+    ): SyncMergeResult {
+        var orphans = false
```
cmonfortep, line 62: We discussed this one over zoom:

Notes:
- To detect orphans we rely on `processIds` and `allResponseIds`.
- Any id from payload which is not present in `processIds` is an orphan.
- However our implementation is able to process all items on a payload.
- We defined an orphan as a savedsite (bokmark or folder) from payload, without a parent in the payload AND not present on the local store.
- For those cases, even if we apply our fallback logic (adding them as childs of root folder), we should not add their ids to `processIds`

cmonfortep (commented):
Code LGTM, and I only have a comment related to how detect orphans we already discussed over zoom. I left my notes in the comment.

cmonfortep (approved):
LGTM

Thanks for pairing and adding the orphan tests. Tested and works as described. 👍

PR #3438, opened 2023-08-09
Maestro appTP test: make vpn dialog handling optional

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1205234542811814/f

### Description
Makes it so that the VPN permission dialog step is optional. 
- Otherwise, this appTP test will only pass once and require the APK to be uninstalled and reinstalled before the test will run again.
- Even if the dialog doesn't show (because permission was accepted before on a previous test run), the later assertions would catch a problem if app TP couldn't be enabled.

### Steps to test this PR

- [x] Fresh install app
- [x] Run `maestro test .maestro/app_tp/app_tp_onboarding.yaml`
- [x] Ensure tests pass
- [x] Run again and verify they still pass 2nd time around

### DIFF
--- a/.maestro/app_tp/app_tp_onboarding.yaml
+++ b/.maestro/app_tp/app_tp_onboarding.yaml
@@ -32,8 +32,9 @@ tags:
 - tapOn:
     text: "Enable App Tracking Protection"
     index: 1
-- assertVisible: "Connection request"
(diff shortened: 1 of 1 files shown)


malmstein (approved):
works the charm, thanks @CDRussell !

PR #3439, opened 2023-08-10
Update DaxTextInput to improve site breakage form

laghee:
Task/Issue URL: https://app.asana.com/0/488551667048375/1203532493649164/f

### Description
To facilitate easier reporting in order to improve our breakage signal:

- full edittext is now clickable rather than just the end icon
- neither breakage category nor text description is required for submission anymore
- category dialog list has been randomized

Plus added 2 additional requested updates to DaxTextInput:

- optional truncation with ellipses of single-line fields
- optional form mode for multi-line fields of minimum 3 lines

### Steps to test this PR
- [x] Load any website & tap either duck foot or menu item for Report Broken Site
- [x] Verify that it's possible to immediately send a blank report with only metadata
- [x] Verify that tapping on either "Describe what happened" or the down carat brings up the radio selection dialog
- [x] Verify that new options appear correctly in the internal component design preview


laghee:
Morning, @malmstein! I spent quite a while banging my head against it, but can't get the truncate function to work. Seems like there's potentially [some kind of existential issue with using `ellipsize` in a `TextInput`](https://stackoverflow.com/questions/5199712/android-edittext-doesnt-get-ellipsized?noredirect=1&lq=1), which I guess makes some sense. If there's something else you think I could try, I'm happy to do so. Alternatively, if you think we should just back that part out for now, I can do that too. Also, wasn't sure if there was value in adding a `formMode` example in the internal component preview?

malmstein:
@laghee thanks for taking care of this!

I’ve refactor the input type to allow for form mode, instead of using a boolean.
re the truncation, I wouldn’t worry about it now, and take care of the rest.

malmstein (approved):
:shipit:

PR #3463, opened 2023-08-16
Minimise use of default() dispatcher

aitorvs:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1204763838216666/f

### Description
Use `IO` dispatcher over `default` dispatcher.
Rename the `default` dispatcher to `computation`

### Steps to test this PR
Smoke tests for every feature
https://github.com/duckduckgo/Android/actions/runs/5896386419/job/15994154084 passes

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/tracker_activity/DeviceShieldActivityFeedFragment.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/tracker_activity/DeviceShieldActivityFeedFragment.kt
@@ -87,7 +87,8 @@ class DeviceShieldActivityFeedFragment : DuckDuckGoFragment() {
             adapter = trackerFeedAdapter
         }
 
-        lifecycleScope.launch(dispatcherProvider.io()) {
+        // let calling suspend fun define the thread
+        lifecycleScope.launch {
             viewModel.getMostRecentTrackers(
                 TimeWindow(
                     config.timeWindow.toLong(),
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/tracker_activity/DeviceShieldActivityFeedFragment.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/tracker_activity/DeviceShieldActivityFeedFragment.kt
@@ -28,6 +28,7 @@ import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.flowWithLifecycle
 import androidx.lifecycle.lifecycleScope
 import com.duckduckgo.anvil.annotations.InjectWith
+import com.duckduckgo.app.global.DispatcherProvider
 import com.duckduckgo.app.global.DuckDuckGoFragment
 import com.duckduckgo.app.global.FragmentViewModelFactory
 import com.duckduckgo.di.scopes.FragmentScope
@@ -58,6 +59,9 @@ class DeviceShieldActivityFeedFragment : DuckDuckGoFragment() {
     @Inject
     lateinit var trackerFeedAdapter: TrackerFeedAdapter
 
+    @Inject
+    lateinit var dispatcherProvider: DispatcherProvider
+
     private val viewModel: DeviceShieldActivityFeedViewModel by bindViewModel()
     private lateinit var binding: ViewDeviceShieldActivityFeedBinding
 
@@ -83,7 +87,7 @@ class DeviceShieldActivityFeedFragment : DuckDuckGoFragment() {
             adapter = trackerFeedAdapter
         }
 
-        lifecycleScope.launch {
+        lifecycleScope.launch(dispatcherProvider.io()) {
             viewModel.getMostRecentTrackers(
                 TimeWindow(
                     config.timeWindow.toLong(),
@@ -96,7 +100,7 @@ class DeviceShieldActivityFeedFragment : DuckDuckGoFragment() {
                 }
         }
 
-        lifecycleScope.launch {
+        lifecycleScope.launch(dispatcherProvider.io()) {
             viewModel.commands()
                 .flowWithLifecycle(lifecycle, Lifecycle.State.STARTED)
                 .collectLatest { processCommands(it) }
--- a/app/src/main/java/com/duckduckgo/app/about/AboutDuckDuckGoViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/about/AboutDuckDuckGoViewModel.kt
@@ -68,7 +68,7 @@ class AboutDuckDuckGoViewModel @Inject constructor(
     fun viewState(): Flow<ViewState> = viewState.onStart {
         val variant = variantManager.getVariant()
 
-        viewModelScope.launch(dispatcherProvider.io()) {
+        viewModelScope.launch {
             viewState.emit(
                 currentViewState().copy(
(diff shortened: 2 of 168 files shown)


aitorvs:
* develop
  * **PR #3463** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3463" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3463?utm_source=stack-comment).

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/tracker_activity/DeviceShieldActivityFeedFragment.kt
```diff
@@ -83,6 +87,7 @@ class DeviceShieldActivityFeedFragment : DuckDuckGoFragment() {
             adapter = trackerFeedAdapter
         }
 
+        // let calling suspend fun define the thread
```
CDRussell, line 90: It's not clear why you'd keep `lifecycleScope.launch` here instead of choosing a dispatcher... Can you elaborate?

app/src/main/java/com/duckduckgo/app/about/AboutDuckDuckGoViewModel.kt
```diff
@@ -43,6 +44,7 @@ class AboutDuckDuckGoViewModel @Inject constructor(
     private val appBuildConfig: AppBuildConfig,
     private val variantManager: VariantManager,
     private val pixel: Pixel,
+    private val dispatcherProvider: DispatcherProvider,
```
CDRussell, line 47: FYI: unused in this class

app/src/main/java/com/duckduckgo/app/launch/LaunchBridgeActivity.kt
```diff
@@ -42,6 +42,7 @@ class LaunchBridgeActivity : DuckDuckGoActivity() {
 
         configureObservers()
 
+        // determineViewToShow() does things with livedata, can't be on IO dispatcher
         lifecycleScope.launch { viewModel.determineViewToShow() }
```
CDRussell, line 46: Two thoughts about this:

1. `viewModel.determineViewToShow()` is `suspend`, so it could handle threading requirements internally meaning this code that calls it shouldn't have to be concerned about how it is called
2. if not wanting to touch `determineViewToShow()` to make 1) true, should we at least be explicit in saying `lifecycleScope.launch(dispatchers.main())` if that is what is needed. i know that functionally it's the same as `lifecycleScope.launch {}` since that defaults to main, but I also dislike it because you need to remember that's the case and would prefer us to be more explicit to help readability if we can.

request-filterer/request-filterer-impl/src/main/java/com/duckduckgo/request/filterer/impl/RequestFiltererImpl.kt
```diff
@@ -81,6 +81,7 @@ class RequestFiltererImpl @Inject constructor(
         currentPage = url
         hasTimeElapsed = false
         if (job?.isActive == true) job?.cancel()
+        // scope is created with context dispatchers.io()
```
CDRussell, line 84: Rather than adding a comment, we could be explicit in using `dispatchers.io()` here. (no strong feelings)

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/tracker_activity/DeviceShieldActivityFeedFragment.kt
aitorvs: I can revert this change for now if you feel strongly. 
After going though all these changes, talking to some ppl and thinking about it for a long time, my high level rational behind for me would be:
* only functions/APIs that (they themselves) perform IO are declared as `suspend` and they specify the `dispatcher`
* any other function (that doesn't do themselves IO) and just call those APIs should be imperative and use `launch` without specifying any dispatcher (ie. we do not mark a function as `suspend` just because it needs to call a suspend function)
  * _Note: I know we currently have cases where we do that (marking a fun as `suspend` just because it calls another `suspend`) when we want to block on the result. I really want to thoroughly review those cases because I am currently not sure how much of it is "need it" vs how much of it is just "code smell"._

Tbf, our (current) usage of `suspend` is a bit "crafty". So this PR is an imperfect first attempt just focused on improving lock contention. After that my proposal would be to discuss how `suspend` should be used in the code base, and enforce as much as we can via static analysis.

request-filterer/request-filterer-impl/src/main/java/com/duckduckgo/request/filterer/impl/RequestFiltererImpl.kt
aitorvs: Not sure I understand/agree with this logic. `scope` is a private property in an impl class. I don't think we need to start guarding mis-usages of private properties/funs inside the same class. Seems unnecessary.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/ui/tracker_activity/DeviceShieldActivityFeedFragment.kt
CDRussell: I'm not sure on the best answer here, so will first instead try give my (still forming) thoughts.

Looking back at the Asana task description, there is this "Possibly lint check to enforce explicitly picking a context when using launch" which I guess is what it all comes down to. _is it ever valid / good practice to launch without specifying a dispatcher?_. 

- If you'd asked me a month ago, I'd have said "no" 
- but since then I've noticed it feels a bit strange to be specifying a dispatcher (typically now `io`) in a place where it doesn't directly do IO, but instead is just launching because it calls a `suspend` function (your 2nd bullet point).

On the other hand, if you don't specify the dispatcher you will sometimes be using the `default` dispatchers (which this PR is trying to avoid) and sometimes be using the `main` dispatcher depending on which scope you're using. In a perfect world, all `suspend` functions doing something intensive would internally take care of ensuring it uses `io` dispatcher (ideal in my opinion, anyway) so that it shouldn't matter how it was called, but that's definitely not the case in the codebase yet and we do have to be very careful right now about how the coroutine was launched. 


So back to _is it ever valid / good practice to launch without specifying a dispatcher?_

- I think **no** when launching on a scope that would then use the `default` (which we're trying to avoid)
- Technically it would be fine to launch on an activity's `lifecycleScope` or `viewModelScope` without specifying since it wouldn't use `default`.
- But wondering if it's just simpler to be explicit at all times, rather than allowing it sometimes for some scopes but not others. It's not hard to be deliberate and specify the `main()` dispatcher if that's what you're wanting.

So no, I don't feel strongly and indeed I'm still trying to form an opinion on some of this.

request-filterer/request-filterer-impl/src/main/java/com/duckduckgo/request/filterer/impl/RequestFiltererImpl.kt
CDRussell: It's more that you felt the need to add a comment at all, which indicates the line of code itself isn't self-explanatory. To make the line of code clearer in its intent, you can either comment it (like you have) or explicitly add the dispatcher which is more self-documenting.

Doesn't matter too though; the comment also makes the intent of the line clear.

Issue #3250, opened 2023-06-09
[Bug] there isn't full screen support for youtube videos.


minischoolX:
### Describe the bug

hello everyone!

I use DuckDuckGo Android for my mobile phone. I simply love it. However there is one annoying issue I encounter every single time I browse youtube.

### How to Reproduce

<strong>How to Reproduce *</strong>
Steps to reproduce the behaviour

    1. Go to 'youtube.com' or 'm.youtube.com'
    2. Click on any video to watch
    3. try to enter full screen by clicking box icon inside video placeholder
    4. See error (we indeed enter full-screen mode but it isn't the expected behaviour of full-screen. 
       what I mean here is in other browsers like chrome automatically switches to landscape mode, 
       which isn't the case with our browser)

### Expected behavior

<strong>Expected behaviour *</strong>

While entering full-screen mode the video shall automatically rotate to landscape mode. which is default behaviour of the website.

### Environment

```markdown
- DDG App Version: Latest and probably all previous versions
- Device: Redmi S2 and assuming we could reproduce this on all devices.
- OS: Android 12 and assuming we could reproduce this on all OS.
```

growabeard:
This seems to be fixed, can it be closed?

PR #3478, opened 2023-08-18
Request landscape orientation when going full screen

mtsyh:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/3250

### Steps to test this PR
* Open a YouTube video
* After playback starts, tap the full screen button
* Observe full screen playback **in landscape mode** rather than portrait

### Notes
* Unfortunately, Android seems to have removed [the onShowCustomView API
](https://developer.android.com/reference/android/webkit/WebChromeClient#onShowCustomView(android.view.View,%20int,%20android.webkit.WebChromeClient.CustomViewCallback)) which allowed the WebView to specify full screen orientation. In that case, we might want to force landscape, as in this PR, because our browser orientation seems to be locked in portrait.
* I left the code which allows the WebView to specify the orientation, in case someone had feedback on another way we could find the requested orientation. If there is no longer any way to do this on Android, we can remove this code and just set landscape in `toggleFullScreen` without piping through `BrowserViewState`.

### UI changes
| Before  | After |
| ------ | ----- |
Before - Portrait|After - Landscape|

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserChromeClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserChromeClient.kt
@@ -16,6 +16,7 @@
 
 package com.duckduckgo.app.browser
 
+import android.content.pm.ActivityInfo
 import android.graphics.Bitmap
 import android.net.Uri
 import android.os.Message
@@ -47,6 +48,7 @@ class BrowserChromeClient @Inject constructor(
 
     override fun onShowCustomView(
         view: View,
+        /* requestedOrientation: Int, */
         callback: CustomViewCallback?,
     ) {
         Timber.d("on show custom view")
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserChromeClient.kt
```diff
@@ -56,7 +58,7 @@ class BrowserChromeClient @Inject constructor(
         }
 
         customView = view
-        webViewClientListener?.goFullScreen(view)
```
mtsyh, line 59: The commented code here demonstrates what was possible with the previous API. Since there isn't currently a way to retrieve `requestedOrientation` from the WebView we default to `SCREEN_ORIENTATION_LANDSCAPE`.

cmonfortep:
Sorry for the late response here. We are currently checking staled PRs, and we are closing this one as per our [contributions guidelines](https://github.com/duckduckgo/Android/blob/develop/TRANSLATIONS.md#improve-language-translations), we only accept contributions for accepted issues and labeled as Help Wanted. Thanks

PR #3481, opened 2023-08-20
Add AppTP connectivity handler

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205307434865662/f

### Description
Add an AppTP-sspecific connectivity handler

### Steps to test this PR
_(Use an emulator)_
- [x] Modify the `hasDeviceConnectivity()` method to always return true
- [x] install internal build from this branch
- [x] enable AppTP
- [x] filter logs by`NetworkConnectivityHealthHandler`
- [x] Disable the WIFI of your computer (this will cause connectivity issues detected as VPN connectivity loss)
- [x] verify AppTP is restarted in the third attempt

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/app/tracking/AppTPVpnConnectivityLossListener.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/app/tracking/AppTPVpnConnectivityLossListener.kt
@@ -16,6 +16,9 @@
 
 package com.duckduckgo.mobile.android.app.tracking
 
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.core.content.edit
 import com.duckduckgo.app.global.DispatcherProvider
 import com.duckduckgo.di.scopes.VpnScope
 import com.duckduckgo.mobile.android.vpn.feature.AppTpFeatureConfig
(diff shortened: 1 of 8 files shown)


aitorvs:
* develop
  * **PR #3481** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3481" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3481?utm_source=stack-comment).

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/app/tracking/AppTPVpnConnectivityLossListener.kt
```diff
@@ -0,0 +1,91 @@
...
+            }
+        }
+    }
+
+    override fun onVpnConnectivityLoss(coroutineScope: CoroutineScope) {
+        coroutineScope.launch {
+            if (isActive() && connectivityLostEvents++ > 1) {
+                logcat(LogPriority.WARN) { "AppTP detected connectivity loss, re-configuring..." }
+                connectivityLostEvents = 0
+                appTrackingProtection.restart()
```
aitorvs, line 67: Should I do `Runtime.getRuntime().exit(0)` instead?

karlenDimla: What advantage to we get from killing the process? Wouldn't this potentially cause last state to be DISABLED with unknown reason which will cause always on to show more?

karlenDimla (approved):
LGTM

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/app/tracking/AppTPVpnConnectivityLossListener.kt
aitorvs: Kill process would ensure all state is freed, for instance memory. We don't have evidence of eg. mem leaks, but if we did process kill would get rid of them.

Another Q, do you think we should put a limit to re-tries? ie. if we try to reconnect 3-4 times without success we basically give up and let user re-enable? If we can recover, it may be indicative of some issue we can't really solve without stopping the VPN service (?)

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/app/tracking/AppTPVpnConnectivityLossListener.kt
```diff
@@ -47,24 +50,37 @@ class AppTPVpnConnectivityLossListener @Inject constructor(
     private val appTrackingProtection: AppTrackingProtection,
     private val appTpFeatureConfig: AppTpFeatureConfig,
     private val dispatcherProvider: DispatcherProvider,
+    private val context: Context,
 ) : VpnConnectivityLossListenerPlugin, VpnServiceCallbacks {
 
+    private val preferences: SharedPreferences by lazy { context.getSharedPreferences(FILENAME, Context.MODE_PRIVATE) }
```
karlenDimla, line 56: NIT : we have way too much sharedprefs. But that is not an issue for this PR

PR #3531, opened 2023-09-12
Add AppTP connectivity pixels

nshuba:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205295861136336/f

### Description
* Add pixel for when we can't stop the VPN
* Update AppTP disabled pixel to queue (in case we lost connectivity)

### Steps to test this PR
Mostly code review as this is difficult to test. One way to test the VPN stop timeout pixel is to create a local version of NetGuard with a forced infinite loop and then try to stop AppTP.

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/integration/NgVpnNetworkStack.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/integration/NgVpnNetworkStack.kt
@@ -26,6 +26,7 @@ import com.duckduckgo.mobile.android.vpn.apps.TrackingProtectionAppsRepository
 import com.duckduckgo.mobile.android.vpn.feature.AppTpLocalFeature
 import com.duckduckgo.mobile.android.vpn.network.VpnNetworkStack
 import com.duckduckgo.mobile.android.vpn.network.VpnNetworkStack.VpnTunnelConfig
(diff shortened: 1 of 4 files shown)


aitorvs (changes requested):
I am not sure about this change plus we don't need it anymore. It is covered by the connection checks we re added. 
I don't really want to over complicate this logic "just in case" atm.
I would propose dropping this change altogether.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/integration/NgVpnNetworkStack.kt
```diff
@@ -210,18 +212,27 @@ class NgVpnNetworkStack @Inject constructor(
             }
 
             var thread = tunnelThread
+            var remainingAttempts = 5
             while (thread != null && thread.isAlive) {
+                if (remainingAttempts <= 0) {
+                    logcat { "Could not stop tunnel thread" }
+                    deviceShieldPixels.reportTunnelThreadStopTimeout()
+                    break
+                }
```
aitorvs, line 221: _(See main comment)_ I feel strongly about not touching this file. ie.

This change doesn't really do anything other than `break`ing and leaving the thread hanging. So not really fixing anything and eventually things will break but in a different place much harder to detect.

nshuba:
As discussed offline, we decided to keep both pixels, but not `break` out of the loop that waits for the VPN thread to stop.

aitorvs (approved):
LGTM! thanks a lot for this

PR #3551, opened 2023-09-18
Plugin point for pixels which get data stripped

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205520159940704/f 

### Description
Introduces a plugin point for defining pixels which should have their ATB and app version removed. These pixels are already defined, but defining them via the plugin point approach allows for more flexibility in _where_ the pixels are defined, allowing them to be created in other modules but still contribute towards this list which the `AtbAndAppVersionPixelRemovalInterceptor` will use.

### Steps to test this PR

- [ ] Take one of the pixels in the list and trigger the scenario where it will be used
  - e.g., `AppPixelName.EMAIL_USE_ADDRESS`
   - log into Email Protectio 
   - visit https://fill.dev/form/registration-email
   - tap on the email field
   - choose to use the personal duck address
   - verify in the logs or with a breakpoint that this is picked up as a special case and ATP/app version are removed by `AtbAndAppVersionPixelRemovalInterceptor` (e.g., add `Timber.d("Removing ATB and app version from pixel $pixel")` at line 49)

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/api/AtbAndAppVersionPixelRemovalInterceptor.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/api/AtbAndAppVersionPixelRemovalInterceptor.kt
@@ -37,8 +37,8 @@ class AtbAndAppVersionPixelRemovalInterceptor @Inject constructor(
     private val pixelsPlugin: PluginPoint<PixelRequiringDataCleaningPlugin>,
 ) : Interceptor, PixelInterceptorPlugin {
 
-    val pixels: List<String> by lazy {
-        pixelsPlugin.getPlugins().flatMap { it.names() }.distinct()
+    val pixels: Set<String> by lazy {
+        pixelsPlugin.getPlugins().flatMap { it.names() }.toSet()
     }
 
     override fun intercept(chain: Interceptor.Chain): Response {
--- a/app/src/main/java/com/duckduckgo/app/global/api/AtbAndAppVersionPixelRemovalInterceptor.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/api/AtbAndAppVersionPixelRemovalInterceptor.kt
@@ -16,9 +16,10 @@
 
 package com.duckduckgo.app.global.api
 
-import androidx.annotation.VisibleForTesting
 import com.duckduckgo.app.global.AppUrl
+import com.duckduckgo.app.global.plugins.PluginPoint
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/duckduckgo/app/global/api/AtbAndAppVersionPixelRemovalInterceptor.kt
```diff
@@ -55,11 +63,15 @@ class AtbAndAppVersionPixelRemovalInterceptor @Inject constructor() : Intercepto
...
 
-    companion object {
-        // list of pixels (pixel name or prefix) for which we'll remove the ATB and App version information
-        @VisibleForTesting
-        internal val pixels = listOf(
+@ContributesMultibinding(
+    scope = AppScope::class,
+    boundType = PixelRequiringDataCleaningPlugin::class,
+)
+object PixelInterceptorPixelsRequiringDataCleaning : PixelRequiringDataCleaningPlugin {
```
CDRussell, line 72: The plan is that most of these pixels can be moved to another module. However, is also dependent on another branch so will move them after this PR lands.

app/src/main/java/com/duckduckgo/app/global/api/AtbAndAppVersionPixelRemovalInterceptor.kt
```diff
@@ -32,7 +33,14 @@ import okhttp3.Response
     scope = AppScope::class,
     boundType = PixelInterceptorPlugin::class,
 )
-class AtbAndAppVersionPixelRemovalInterceptor @Inject constructor() : Interceptor, PixelInterceptorPlugin {
+class AtbAndAppVersionPixelRemovalInterceptor @Inject constructor(
+    private val pixelsPlugin: PluginPoint<PixelRequiringDataCleaningPlugin>,
+) : Interceptor, PixelInterceptorPlugin {
+
+    val pixels: List<String> by lazy {
+        pixelsPlugin.getPlugins().flatMap { it.names() }.distinct()
```
aitorvs, line 41: nit: `distinct()` converts to a set and then back to a list. The second (to list) conversion is not really necessary. We can just `toSet()` instead

aitorvs (approved):
Left a nit. LGTM

CDRussell:
Thanks for the review @aitorvs. Adding you as assignee.

I've changed to using `toSet()` too. Will merge once CI checks pass.

PR #3554, opened 2023-09-19
Move EmailManager's isSignedIn() init to worker thread

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205528656118798/f 

### Description
Moves an expensive call to a worker thread as it was a source of ANRs, as it was being called from main thread and immediately upon process creation.

### Steps to test this PR

- QA optional
- [optional]: Check Email Protection still working as expecting, both when signed in and not (e.g., check `Generate Duck Address` shows in the overflow when logged in)


nalcalag (approved):
Tested and works as expected 💯

PR #3576, opened 2023-09-22
Fix issues with bookmarks search

joshliebe:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205531244553377/f

### Description
- Fixes an issue where favorites would disappear when searching.
- Also fixes an issue where folders would display incorrectly when searching.

## Steps to test this PR

- [x] Add some favorites, folders and bookmarks
- [ ] Verify that the favorites, folders and bookmarks are filtered correctly when searching

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/bookmarks/ui/FavoritesAdapter.kt
+++ b/app/src/main/java/com/duckduckgo/app/bookmarks/ui/FavoritesAdapter.kt
@@ -60,6 +60,7 @@ class FavoritesAdapter(
 
     fun setItems(
         favoriteItems: List<FavoriteItem>,
+        filteringMode: Boolean = false,
     ) {
         val generatedList = generateNewList(favoriteItems)
         val diffCallback = DiffCallback(old = this.favoriteItems, new = generatedList)
@@ -67,7 +68,7 @@ class FavoritesAdapter(
         this.favoriteItems.clear().also { this.favoriteItems.addAll(generatedList) }
         diffResult.dispatchUpdatesTo(this)
 
-        if (favoriteItems.isEmpty()) {
+        if (filteringMode || favoriteItems.isEmpty()) {
             return
         }
     }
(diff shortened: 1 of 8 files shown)


joshliebe:
* develop
  * **PR #3576** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3576" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3576?utm_source=stack-comment).

CDRussell:
Nit: it's weird that the copy we show for no matching bookmarks or favorites is obviously the empty state placeholder rather than something like "No matching favorites/bookmarks found":

- No favorites added yet
- No boomarks added yet

it would be ideal to treat that copy as a bug that could also be improved. Either as part of this PR or capturing as a LHF follow-up.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/FavoritesAdapter.kt
```diff
@@ -60,14 +60,15 @@ class FavoritesAdapter(
...
+        filteringMode: Boolean = false,
     ) {
         val generatedList = generateNewList(favoriteItems)
         val diffCallback = DiffCallback(old = this.favoriteItems, new = generatedList)
         val diffResult = DiffUtil.calculateDiff(diffCallback)
         this.favoriteItems.clear().also { this.favoriteItems.addAll(generatedList) }
         diffResult.dispatchUpdatesTo(this)
 
-        if (favoriteItems.isEmpty()) {
+        if (filteringMode || favoriteItems.isEmpty()) {
```
CDRussell, line 71: this whole `if` can be removed, right? it's guarding against nothing; it used to do an _early return_ to prevent some other code from running i'm guessing but is redundant now

CDRussell (approved):

joshliebe:
> Nit: it's weird that the copy we show for no matching bookmarks or favorites is obviously the empty state placeholder rather than something like "No matching favorites/bookmarks found":
> 
> * No favorites added yet
> * No boomarks added yet
> 
> it would be ideal to treat that copy as a bug that could also be improved. Either as part of this PR or capturing as a LHF follow-up.

Agreed, I think we should ship this one and we can update the copy separately. I'll make a task.

app/src/main/java/com/duckduckgo/app/bookmarks/ui/FavoritesAdapter.kt
joshliebe: You're right, that doesn't need to be there anymore, I'll remove 👍

PR #3615, opened 2023-10-03
Support alternative Dashboard layout

shakyShane:
Task/Issue URL: https://app.asana.com/0/0/1205648391652872/f

### Description

The updated Dashboard will have the following new interface, designed to allow platforms to opt-in to additional UI features/layouts in a unified way. For now just `primaryScreen.layout` is supported to enable an experiment to increase adoption of the protections toggle

```ts
/**
 * Call this function to provide the dashboard with RemoteFeatureSettings
 */ 
declare function onChangeFeatureSettings(settings: RemoteFeatureSettings): void

/**
 * This describes the payload for feature settings
 */
export interface RemoteFeatureSettings {
  primaryScreen?: PrimaryScreen;
}

export interface PrimaryScreen {
  /**
   * A string to represent different screen layouts
   */
  layout: "default" | "highlighted-protections-toggle";
}
```

To provide different `layout` values, the PR includes support for a new Privacy Configuration key, `privacyDashboard`

- Example PR adding the dashboard to privacy config the https://github.com/duckduckgo/privacy-configuration/pull/1349
- Example PR enabling the experiment https://github.com/duckduckgo/privacy-configuration/pull/1350

So, the experiment will be enabled remotely, and then the `PrivacyDashboardViewModel` will use the values from the privacy config to choose which layout the dashboard should show.

toggle exeriment

## Pixel Params

This PR also adds the param `dashboard_highlighted_toggle`, to the following pixels

- mp_wla (adding to the allowlist, pressing the toggle)
- mp_wlr (removing from the allowlist)
- m_bsr (feedback form is submitted)
- Privacy Triage: https://app.asana.com/0/0/1205648391652879/f

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Normal  | Experiment |
| ------ | ----- |
Screenshot_20231004_094645|Screenshot_20231004_094914


app/src/main/java/com/duckduckgo/app/brokensite/BrokenSiteActivity.kt
```diff
@@ -165,6 +167,7 @@ class BrokenSiteActivity : DuckDuckGoActivity() {
         private const val CONSENT_MANAGED_EXTRA = "CONSENT_MANAGED_EXTRA"
         private const val CONSENT_OPT_OUT_FAILED_EXTRA = "CONSENT_OPT_OUT_FAILED_EXTRA"
         private const val CONSENT_SELF_TEST_FAILED_EXTRA = "CONSENT_SELF_TEST_FAILED_EXTRA"
+        private const val BOOLEAN_PARAMS = "BOOLEAN_PARAMS"
```
shakyShane, line 170: sending a list of params here that will end up like 

"<param>=true"

Does this way because this data is send through the Intent and I didn't know how to serialize a map through it :)

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -157,6 +161,33 @@ class PrivacyDashboardHybridViewModel @Inject constructor(
...
+
+    enum class LayoutType(val value: String) {
+        DEFAULT("default"),
+        HIGHLIGHTED_PROTECTIONS_TOGGLE("highlighted-protections-toggle");
+    }
+
+    data class PrimaryScreenSettings(
+        val layout: String = LayoutType.DEFAULT.value
+    )
+
```
shakyShane, line 190: The dashboard just has an alternative layout - it knows nothing about experiments etc. Its the job of this view model to determine if we're in an experiment group and then apply the correct settings each time the dashboard is open

shakyShane: The string values here match the documented strings in the updated Privacy Dashboard

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -157,6 +161,36 @@ class PrivacyDashboardHybridViewModel @Inject constructor(
...
+                }
+
+                return RemoteFeatureSettingsViewState(
+                    primaryScreen = PrimaryScreenSettings(
+                        layout = layout,
+                    ),
+                )
+            }
+        }
+    }
```
aitorvs, line 182: I would remove the `fromSettings()` (see also more changes below)
```suggestion
data class RemoteFeatureSettingsViewState(val primaryScreen: PrimaryScreenSettings = PrimaryScreenSettings())
```

The reason is that `isEnabled()` checks on a feature flag may require IO, so we should do them off the main thread.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -186,6 +223,22 @@ class PrivacyDashboardHybridViewModel @Inject constructor(
...
+        // otherwise, send the pixel param
+        return listOf(PixelParameter.PROTECTION_TOGGLE_EXP)
+    }
+
+    private fun pixelParamMap(): Map<String, String> {
+        return pixelParamList().associateWith { true.toString() }
+    }
+
     private suspend fun updateSite(site: Site) {
         withContext(dispatcher.main()) {
```
aitorvs, line 243: This change goes alongside the one above in `RemoteFeatureSettingsViewState` and the one below this one in the same method (GH is stupid enough to not allow suggestions in code that isn't modified, so I needed to split it)
 ```suggestion
        suspend fun createRemoteFeatureSettingsViewState() = withContext(dispatcher.io()) {
            val layout = if (privacyDashboardRemoteFeature.protectionToggleHighlight().isEnabled()) {
                LayoutType.HIGHLIGHTED_PROTECTIONS_TOGGLE.value
            } else {
                LayoutType.DEFAULT.value
            }

            return@withContext RemoteFeatureSettingsViewState(
                primaryScreen = PrimaryScreenSettings(
                    layout = layout,
                ),
            )
        }
        val remoteFeatureSettings = createRemoteFeatureSettingsViewState()
        withContext(dispatcher.main()) {
```

File: privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt [hunk 5]
```diff
@@ -194,6 +247,7 @@ class PrivacyDashboardHybridViewModel @Inject constructor(
                     requestData = requestDataViewStateMapper.mapFromSite(site),
                     protectionStatus = protectionStatusViewStateMapper.mapFromSite(site),
                     cookiePromptManagementStatus = autoconsentStatusViewStateMapper.mapFromSite(site),
+                    remoteFeatureSettings = RemoteFeatureSettingsViewState.fromSettings(privacyDashboardRemoteFeature),
```
aitorvs, line 250: Ditto
```suggestion
                    remoteFeatureSettings = remoteFeatureSettings,
```

aitorvs (commented):
I just looked at the usage of the feature flags.
My only comment is try to do the `isEnabled` checks off the main thread.
Haven't reviewed the other parts.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardHybridViewModel.kt
```diff
@@ -186,14 +207,44 @@ class PrivacyDashboardHybridViewModel @Inject constructor(
...
+        return listOf(PixelParameter.DASHBOARD_TOGGLE_HIGHLIGHT)
+    }
+
+    private fun pixelParamMap(): Map<String, String> {
+        return pixelParamList().associateWith { true.toString() }
+    }
+
     private suspend fun updateSite(site: Site) {
+        suspend fun createRemoteFeatureSettingsViewState() = withContext(dispatcher.io()) {
+            val layout = if (privacyDashboardRemoteFeature.highlightedProtectionsToggle().isEnabled()) {
```
cmonfortep, line 228: Related to the locale check, if you want to show the new layout to english speaker users, this is a good place to add the check. 

`if (privacyDashboardRemoteFeature.highlightedProtectionsToggle().isEnabled()
            && Locale.getDefault().language == Locale.ENGLISH.language
        )`

If you go for this approach, in the `pixelParamList` method. I believe you can just do: 
```
    private fun pixelParamList(): List<String> {
        val viewState = viewState.value ?: return emptyList()
        if (viewState.remoteFeatureSettings.primaryScreen.layout == LayoutType.DEFAULT.value) return emptyList()
        // otherwise, send the pixel param
        return listOf(DASHBOARD_TOGGLE_HIGHLIGHT)
    }
```

cmonfortep (commented):
Overall looks great, I've just posted some questions about Locale logic that are not clear.

PR #3662, opened 2023-10-12
Fix DRM permissions

nshuba:
Task/Issue URL: https://app.asana.com/0/0/1205710253534934/f

### Description
* Make sure we deny DRM explicitly to avoid website hangs
* If users disable protections, DRM should be allowed
* Similarly, if we temporarily disable all protections for a site via `unprotectedTemporary`, DRM should be allowed

### Steps to test this PR
- [x] Try to load https://shaka-player-demo.appspot.com/support.html - it should be blank
- [x] Load https://shaka-player-demo.appspot.com/support.html - it should now show available DRMs and `com.widevine.alpha` should be `null`
- [x] Disable protections for the site and reload it. `com.widevine.alpha` should now be a non-null object
- [x] Re-enable protections and override PrivacyConfig to add `shaka-player-demo.appspot.com` to `eme` exceptions. You can use this one (version may need an update): https://jsonblob.com/api/jsonBlob/1160261143960084480
- [x] Load https://shaka-player-demo.appspot.com/support.html - `com.widevine.alpha` should be a non-null object
- [x] Update the config and add `shaka-player-demo.appspot.com` to `unprotectedTemporary`
- [x] Load https://shaka-player-demo.appspot.com/support.html - `com.widevine.alpha` should be a non-null object

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserChromeClientTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserChromeClientTest.kt
@@ -209,6 +209,7 @@ class BrowserChromeClientTest {
         testee.onPermissionRequest(mockPermission)
 
         verify(mockPermission, never()).grant(any())
+        verify(mockPermission).deny()
     }
 
     @ExperimentalCoroutinesApi
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserChromeClientTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserChromeClientTest.kt
@@ -78,7 +78,7 @@ class BrowserChromeClientTest {
         mockFileChooserParams = mock()
         testee.webViewClientListener = mockWebViewClientListener
         webView = TestWebView(getInstrumentation().targetContext)
-        whenever(mockDrm.getDrmPermissionsForRequest(any(), any())).thenReturn(arrayOf())
+        whenever(mockDrm.isDrmAllowedForUrl(any())).thenReturn(false)
         mockSitePermissionsManager.stub { onBlocking { getSitePermissionsAllowedToAsk(any(), any()) }.thenReturn(arrayOf()) }
     }
 
@@ -192,7 +192,7 @@ class BrowserChromeClientTest {
         val mockPermission: PermissionRequest = mock()
         whenever(mockPermission.resources).thenReturn(permissions)
         whenever(mockPermission.origin).thenReturn("https://open.spotify.com".toUri())
-        whenever(mockDrm.getDrmPermissionsForRequest(any(), any())).thenReturn(permissions)
+        whenever(mockDrm.isDrmAllowedForUrl(any())).thenReturn(true)
         testee.onPermissionRequest(mockPermission)
 
         verify(mockPermission).grant(arrayOf(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID))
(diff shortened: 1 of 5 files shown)


privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/Drm.kt
```diff
@@ -24,10 +24,9 @@ interface Drm {
      * @return an `Array<String>` if the given [url] is in the eme list and an empty array
```
marcosholgado, line 24: We need to change the documentation to match the new method

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/drm/RealDrm.kt
```diff
@@ -40,25 +38,12 @@ class RealDrm @Inject constructor(
...
-        if (shouldEnableDrmForUri(url.toUri())) {
-            resources.find { (it == PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID) }?.let {
-                perms.add(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID)
-            }
-        }
-        return perms.toTypedArray()
-    }
-
-    private fun shouldEnableDrmForUri(uri: Uri): Boolean {
+    override fun isDrmAllowedForUrl(url: String): Boolean {
```
marcosholgado, line 41: nit: you could pass a uri instead which is what you are getting from `onPermissionRequest` and save a few parsings here and there.

privacy-config/privacy-config-impl/src/test/java/com/duckduckgo/privacy/config/impl/features/drm/RealDrmTest.kt
```diff
@@ -48,51 +47,24 @@ class RealDrmTest {
...
-        val url = "https://open.spotify.com"
-
-        val value = testee.getDrmPermissionsForRequest(url, permissions)
-
-        assertEquals(0, value.size)
+        assertTrue(testee.isDrmAllowedForUrl(url))
     }
 
     @Test
     fun whenGetDrmPermissionsForRequestIfFeatureIsEnabledAndDomainIsNotInExceptionsListThenNoPermissionsAreReturned() {
```
marcosholgado, line 55: All the test names need to be changed to match what the new test is doing

app/src/androidTest/java/com/duckduckgo/app/browser/BrowserChromeClientTest.kt
```diff
@@ -78,7 +78,7 @@ class BrowserChromeClientTest {
         mockFileChooserParams = mock()
         testee.webViewClientListener = mockWebViewClientListener
         webView = TestWebView(getInstrumentation().targetContext)
-        whenever(mockDrm.getDrmPermissionsForRequest(any(), any())).thenReturn(arrayOf())
+        whenever(mockDrm.isDrmAllowedForUrl(any())).thenReturn(false)
```
marcosholgado, line 81: In some of these tests we are validating that grant() was never called, i.e. `verify(mockPermission, never()).grant(any())`. Now that we are denying them we should also validate that deny() was called too.

marcosholgado (approved):
LGTM!

Pre-approved and I've added a few comments to improve some tests and change the docs etc to match the new logic.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/drm/RealDrm.kt
nshuba: Good point. I tried going this route but it required changing gradle dependencies in `privacy-config-api`, so decided it wasn't worth it since we're likely to remove this anyway once we implement the consent prompt.

PR #3664, opened 2023-10-13
Add support for privacy config v4

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1205690187399816/f 

### Description
Add support for remote config v4

### Steps to test this PR

_Update_
- [ ] Install from develop
- [ ] Check the DB for the different features (privacy_config.db, autoconsent, autofill, cookies, etc). Go to the different exceptions tables and seen that the column for reasons has content
- [ ] Update from this branch
- [ ] The app should not crash
- [ ] Check the DBs and the reason column should now be null but domains should still be there.

_New install_
- [ ] The app should not crash
- [ ] Check the DBs and the reason column should now be null but domains should still be there.

_Tests_
- [ ] Privacy Tests should pass, see https://github.com/duckduckgo/Android/actions/runs/6510081560

### DIFF
--- a/ad-click/ad-click-impl/src/main/java/com/duckduckgo/adclick/impl/AdClickAttributionFeature.kt
+++ b/ad-click/ad-click-impl/src/main/java/com/duckduckgo/adclick/impl/AdClickAttributionFeature.kt
@@ -16,11 +16,13 @@
 
 package com.duckduckgo.adclick.impl
 
+import com.duckduckgo.feature.toggles.api.FeatureExceptions.FeatureException
+
 data class AdClickAttributionFeature(
     val state: String,
     val minSupportedVersion: Int?,
     val settings: AdClickAttributionSettings,
-    val exceptions: List<AdClickAttributionException>,
+    val exceptions: List<FeatureException>,
 )
 
 data class AdClickAttributionSettings(
@@ -41,5 +43,3 @@ data class AdClickAttributionAllowlist(
     val blocklistEntry: String?,
     val host: String?,
 )
-
-data class AdClickAttributionException(val domain: String, val reason: String)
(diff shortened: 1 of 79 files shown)


marcosholgado:
* develop
  * **PR #3664** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3664" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3664?utm_source=stack-comment).

ad-click/ad-click-impl/src/main/java/com/duckduckgo/adclick/impl/AdClickAttributionFeature.kt
```diff
@@ -16,11 +16,13 @@
...
 package com.duckduckgo.adclick.impl
 
+import com.duckduckgo.feature.toggles.api.FeatureExceptions.FeatureException
+
 data class AdClickAttributionFeature(
     val state: String,
     val minSupportedVersion: Int?,
     val settings: AdClickAttributionSettings,
-    val exceptions: List<AdClickAttributionException>,
+    val exceptions: List<FeatureException>,
```
aitorvs, line 25: Nit: to be clear, this is fine. 
It feels odd for now because all these features (ad-click) etc related to the `privacy-config` module but take the model from the `feature-toggles` module.
No need to change, just a comment

privacy-config/privacy-config-api/build.gradle
```diff
@@ -31,6 +31,7 @@ kotlin {
 }
 
 dependencies {
+    implementation project(path: ':feature-toggles-api')
```
aitorvs, line 34: Yeah, that's what I was saying before. Maybe we want a `PrivacyConfigException` instead of using the `FeatureException`.
No strong opinions tbh. I think long term we want to migrate everything to use the feature-toggles module

aitorvs (approved):
lgtm. Tested and also worked as expected.

PR #3804, opened 2023-11-06
Integrate fetaure flags with remote variant manager

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205644869911730/f

### Description
TODO

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app-build-config/app-build-config-api/src/main/java/com/duckduckgo/appbuildconfig/api/AppBuildConfig.kt
+++ b/app-build-config/app-build-config-api/src/main/java/com/duckduckgo/appbuildconfig/api/AppBuildConfig.kt
@@ -32,6 +32,11 @@ interface AppBuildConfig {
     val model: String
     val deviceLocale: Locale
     val isDefaultVariantForced: Boolean
+
+    /**
+     * You should call [variantName] in a suspend function
+     */
+    val variantName: String?
 }
 
 enum class BuildFlavor {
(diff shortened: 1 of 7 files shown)


feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -182,10 +191,27 @@ internal class ToggleImpl constructor(
...
+
+    private fun Toggle.State.isVariantTreated(variant: String?): Boolean {
+        // if no variants a present, we consider always treated
+        if (this.targets.isEmpty()) {
+            return true
+        }
+        // if variants present BUT no variant has been assigned yet, consider not treated
+        if (variant == null) {
+            return false
+        }
```
nalcalag, line 205: I think I confused you when you asked me earlier. This should return `true`. 
If the variant hasn't been assigned when checking feature flag (experiment), we should consider it as treated and add the logic to assign the default one. Does it make sense?

app-build-config/app-build-config-api/src/main/java/com/duckduckgo/appbuildconfig/api/AppBuildConfig.kt
```diff
@@ -32,6 +32,11 @@ interface AppBuildConfig {
     val model: String
     val deviceLocale: Locale
     val isDefaultVariantForced: Boolean
+
+    /**
+     * You should call [variantName] in a suspend function
+     */
+    val variantName: String?
```
nalcalag, line 39: `getVariantKey()` is synchronous so this is not needed

nalcalag (approved):
Tested and works as expected 💯

PR #3857, opened 2023-11-13
Report VPN events

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205896793620608/f

### Description
See Asana task

### Steps to test this PR
See `Test` section in Asana task


aitorvs:
* develop
  * **PR #3857** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3857?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3857?utm_source=stack-comment).

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/NetworkProtectionPixelNames.kt
```diff
@@ -46,8 +49,11 @@ enum class NetworkProtectionPixelNames(
     NETP_WG_ERROR_CANT_START_WG_BACKEND_DAILY("m_netp_ev_wireguard_error_cannot_start_wireguard_backend_d", enqueue = true),
     NETP_WG_ERROR_FAILED_TO_LOAD_WG_LIBRARY("m_netp_ev_wireguard_error_unable_to_load_wireguard_library_c", enqueue = true),
     NETP_WG_ERROR_FAILED_TO_LOAD_WG_LIBRARY_DAILY("m_netp_ev_wireguard_error_unable_to_load_wireguard_library_d", enqueue = true),
-    NETP_LATENCY_REPORT("m_netp_ev_latency_c"),
-    NETP_REPORT_POOR_LATENCY_DAILY("m_netp_ev_poor_latency_d"),
+    NETP_REPORT_TERRIBLE_LATENCY("m_netp_ev_terrible_latency_c"),
```
karlenDimla, line 52: Should terrible and poor be ennqueued? Not sure about the likelihood of failing to send this pixel if the device connectivity is bad?

aitorvs: I think it's fine not to enqueue them (we can revisit once we see the first metrics). The reasoning is:
* if we're able to measure latency pixels should go through
* even if they don't we compute a running geometric average, and the weight on the last value is such that the average is not extremely responsive to it, so the poor connectivity will take several good measurements to go back to be GOOD

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
```diff
@@ -0,0 +1,165 @@
...
+    private val networkProtectionState: NetworkProtectionState,
+    private val currentNetworkState: CurrentNetworkState,
+) : VpnServiceCallbacks {
+
+    private val job = ConflatedJob()
+    private val failureReported = AtomicBoolean(false)
+
+    override fun onVpnStarted(coroutineScope: CoroutineScope) {
+        job += coroutineScope.launch {
+            failureReported.set(false)
```
karlenDimla, line 60: Why are we reseting failureReported in onVpnStarted only?

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
```diff
@@ -0,0 +1,165 @@
...
+                        // skip if previously reported
+                        if (!failureReported.getAndSet(true)) {
+                            pixels.reportTunnelFailure()
+                        } else {
+                            logcat { "Last handshake was already reported, skipping" }
+                        }
+                    } else {
+                        if (failureReported.getAndSet(false)) {
+                            logcat(WARN) { "Recovered from tunnel failure" }
+                            pixels.reportTunnelFailureRecovered()
```
karlenDimla, line 99: If after `reportTunnelFailure` the device loses connectivity, we are going to report `reportTunnelFailureRecovered`. I don’t think that is a correct signal. Not sure how edge case-y that is though.

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
aitorvs: Is not only reset here, see the calls to `ailureReported.getAndSet`.

where else would you reset it?

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
aitorvs: Good catch, I guess the `else` condition is rather `else if (diff.seconds.inWholeMinutes <= 2)`

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
karlenDimla: In the scenario where user has apptp and netp enabled. User disables netp after experiencing tunnel failure (`failureReported` will be set to true). If the user re-enables Netp (handshake diff will be less than threshold), do we expect `reportTunnelFailureRecovered` to be emitted?

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
karlenDimla: I am not sure where `2` came from. But yes i think we should check that the diff still to not include the scenario above.

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
karlenDimla: or maybe just at line `82`?

karlenDimla (approved):

PR #3859, opened 2023-11-13
Migrate geoswitching to locations

karlenDimla:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205842615840144/f

### Description
Migrate geoswitching to use /locations instead

### Steps to test this PR
https://app.asana.com/0/0/1205937644199354/f

### DIFF
--- a/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/settings/geoswitching/NetpEgressServersProvider.kt
+++ b/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/settings/geoswitching/NetpEgressServersProvider.kt
@@ -65,8 +65,8 @@ class RealNetpEgressServersProvider @Inject constructor(
             ServerLocation(
                 countryCode = it.countryCode,
                 countryName = it.countryName,
-                cities = it.cities.sorted(),
+                cities = it.cities,
             )
-        }.sortedBy { it.countryName }
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


karlenDimla:
* develop
  * **PR #3859** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3859?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3859?utm_source=stack-comment).

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/settings/geoswitching/NetpEgressServersProvider.kt
```diff
@@ -63,34 +65,8 @@ class RealNetpEgressServersProvider @Inject constructor(
             ServerLocation(
                 countryCode = it.countryCode,
                 countryName = it.countryName,
-                cities = it.cities,
+                cities = it.cities.sorted(),
```
karlenDimla, line 68: I moved the sorting to when the serverlocations are being retrieved instead of when it is stored as I am not sure if the sorting is preserved when inserting in the room db. Thoughts?

jbrendel: Supposedly, room DB doesn't guarantee that things are return in the same order in which they were inserted. So, this seems to be safer. And since we are talking about a small list of items, I think your approach is fine.

aitorvs: nit: @karlenDimla, the sort logic seems to be logic that should rather be in the UI layer, not in the domain layer.
Sorting the cities like that is fine, but this is not really about room, it's just about SQL. If you want a certain ordering to be returned it's best to put that in the SQL query.

aitorvs (approved):
LGTM

PR #3866, opened 2023-11-15
Update to Kotlin 1.9.20

joshliebe:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205249419591210/f

### Description

Changes:
- `kotlin_version` = `1.8.10` → `1.9.20`
-  `anvil_version` = `2.4.4` → `2.4.8`
- `ksp_version` = `1.8.10-1.0.9` → `1.9.20-1.0.14`
- `compile_sdk` = `33` → `34`'
- `version.androidx.room` = `2.5.0` → `2.6.0`
- `version.robolectric` = `4.9.2` → `4.11.1`

### Steps to test this PR
- [ ] Smoke test

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpRemoteFeatures.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpRemoteFeatures.kt
@@ -71,11 +71,13 @@ class AppTpRemoteFeaturesStore @Inject constructor(
     private val stateAdapter: JsonAdapter<State> = moshi.newBuilder().add(KotlinJsonAdapterFactory()).build().adapter(State::class.java)
     private val listener = OnSharedPreferenceChangeListener { preferences, key ->
         coroutineScope.launch(dispatcherProvider.io()) {
-            val state = preferences.load(key)
-            if (state == null) {
-                togglesCache.remove(key)
-            } else {
-                togglesCache[key] = state
+            key?.let {
+                val state = preferences.load(key)
+                if (state == null) {
+                    togglesCache.remove(key)
+                } else {
+                    togglesCache[key] = state
+                }
             }
         }
     }
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
@@ -39,7 +39,6 @@ import com.duckduckgo.common.utils.extensions.html
 import com.duckduckgo.di.scopes.FragmentScope
 import javax.inject.Inject
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.channels.BroadcastChannel
 import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.launch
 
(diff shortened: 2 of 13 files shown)


joshliebe:
* develop
  * **PR #3866** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3866?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/3866?utm_source=stack-comment).

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
```diff
@@ -71,8 +70,7 @@ class WelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome)
     private var typingAnimation: ViewPropertyAnimatorCompat? = null
     private var welcomeAnimationFinished = false
 
-    // we use a BroadcastChannel because we don't want to emit the last value upon subscription
```
CDRussell, line 74: This comment seems important; it says we don't want to emit the last value... but will `replay = 1` do just that?

build.gradle
```diff
@@ -3,15 +3,15 @@
...
-        anvil_version = '2.4.4'
-        ksp_version = '1.8.10-1.0.9'
+        anvil_version = '2.4.8'
+        ksp_version = '1.9.20-1.0.14'
         gradle_plugin = '8.1.2' // When updating, also update lint_version
         lint_version = '31.1.2' // This value must always be gradle_plugin + 23
         min_sdk = 23
-        target_sdk = 33
-        compile_sdk = 33
+        target_sdk = 34
```
CDRussell, line 13: are we sure we want to change `target_sdk` in here? have you confirmed if this will bring any new behaviours and if they are all desired? 

If not, we presumably can defer changing the target sdk while still updating everything else?

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpRemoteFeatures.kt
```diff
@@ -71,11 +71,13 @@ class AppTpRemoteFeaturesStore @Inject constructor(
     private val stateAdapter: JsonAdapter<State> = moshi.newBuilder().add(KotlinJsonAdapterFactory()).build().adapter(State::class.java)
     private val listener = OnSharedPreferenceChangeListener { preferences, key ->
         coroutineScope.launch(dispatcherProvider.io()) {
-            val state = preferences.load(key)
-            if (state == null) {
-                togglesCache.remove(key)
-            } else {
-                togglesCache[key] = state
+            key?.let {
```
CDRussell, line 74: I think we might need to handle the case when `key` is null. According to the docs, that signifies that the preferences were cleared and so perhaps the appropriate response is to clear the toggles cache. 

Not sure if there are gotchas to doing that here; @aitorvs thoughts?

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
joshliebe: I added the comment back, changed it to:
```
// we use replay = 0 because we don't want to emit the last value upon subscription
private val events = MutableSharedFlow<WelcomePageView.Event>(replay = 0, extraBufferCapacity = 1)
```

CDRussell: Looks good

build.gradle
joshliebe: Changed to 33 ✅

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpRemoteFeatures.kt
CDRussell: I think this would only matter if we called `clear` on the preferences so i wouldn't block on this, especially since we weren't handling that case before either from what i can see.

CDRussell:
LGTM.

Tested across Android APIs 34, 30 and 24; all working as expected.

CDRussell (approved):

PR #3874, opened 2023-11-16
Fix pixels related to protections toggle on Broken Site screen

lmac012:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205947853089389/f

### Description

In the BrokenSiteViewModel, the onProtectionsToggled() method is invoked right after the view is initialized, which leads to sending m_broken_site_allowlist_remove pixel even though the user didn't interact with the protections toggle.

#### Pixels are not fired without user interaction with the protections toggle (this is the fix)
- [x] Go to the site breakage report screen
- [x] Verify that m_broken_site_allowlist_remove / m_broken_site_allowlist_add pixels weren't fired until you use the protections toggle.

#### Verify pixels are fired when toggling protections (existing behavior that shouldn't change)
 - [x] Go to the site breakage report screen
 - [x] Toggle protections off/on
 - [x] Verify that m_broken_site_allowlist_add pixel is fired on turning protections OFF
 - [x] Verify that m_broken_site_allowlist_remove pixel is fired on turning protections ON

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/brokensite/view/SiteProtectionsToggle.kt
+++ b/app/src/main/java/com/duckduckgo/app/brokensite/view/SiteProtectionsToggle.kt
@@ -43,7 +43,9 @@ class SiteProtectionsToggle @JvmOverloads constructor(
 
     fun setOnProtectionsToggledListener(listener: (Boolean) -> Unit) {
         binding.protectionsSwitch.setOnCheckedChangeListener { _, isChecked ->
-            listener.invoke(isChecked)
+            if (isChecked != areProtectionsEnabled()) {
+                listener.invoke(isChecked)
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/brokensite/view/SiteProtectionsToggle.kt
```diff
@@ -80,4 +82,11 @@ class SiteProtectionsToggle @JvmOverloads constructor(
             }
         }
     }
+
+    private fun areProtectionsEnabled(): Boolean = when (state) {
+        SiteProtectionsState.ENABLED -> true
+        SiteProtectionsState.DISABLED,
+        SiteProtectionsState.DISABLED_BY_REMOTE_CONFIG,
+        -> false
+    }
```
anikiki, line 91: This can be simplified (unless there is a reason for the when):
```
private fun areProtectionsEnabled(): Boolean {
    return state == SiteProtectionsState.ENABLED
}
```

anikiki (approved):

app/src/main/java/com/duckduckgo/app/brokensite/view/SiteProtectionsToggle.kt
lmac012: The only reason I chose to use when here (and in general with enums / sealed classes) is the extra safety in case e.g. we add more states later -> With `when`, a developer will be forced by the compiler to edit this method, with `if`, the logic here may silently become incorrect.

PR #3870, opened 2023-11-15
Implement share web api with css

marcosholgado:
Task/Issue URL: https://app.asana.com/0/0/1205583865502218/f 

### Description
This PR adds support for the share web API through c-s-s

### Steps to test this PR

- [x] Use `https://jsonblob.com/api/1169732359619993600` as the remote config
- [x] Go to `https://mdn.github.io/dom-examples/web-share/`
- [x] Click on Share MDN
- [x] A share sheet should show up 
- [x] Dismiss it
- [x] Web should say Error: DataError: User aborted
- [x] Click again
- [x] Copy to clipboard 
- [x] Web should say MDN shared successfully

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205966966963452

PR #3877, opened 2023-11-17
Add support for permissions API

nshuba:
Task/Issue URL: https://app.asana.com/0/0/1205860617093733/f

### Description
Add support for [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query)

### Steps to test this PR

First, update your local copy of contentScope.js to include changes in https://github.com/duckduckgo/content-scope-scripts/pull/820

_Session-based permissions: deny_
- [x] Go to https://nshuba.github.io/test-site/
- [x] Click "Query Permission" and verify that the printed state is 'prompt'
- [x] Click "Request Camera" and deny the permission
- [x] Click "Query Permission" and verify that the printed state is 'prompt'. 
  * **Note**: ideally this should be "denied", but we don't store denials on-disk. 
  * We can revisit how to better handle this in https://app.asana.com/0/1200396763693192/1205944992235904/f

_Session-based permissions: allow_
- [x] Open a new tab and navigate to the test site
- [x] Click "Query Permission" and verify that the printed state is 'prompt'
- [x] Click "Request Camera" and allow the permission
- [x] Click "Query Permission" and verify that the printed state is 'granted'
- [x] From the dropdown, change "camera" to "microphone"
- [x] Click "Query Permission" and verify that the printed state is 'prompt'
- [x] Click "Request Microphone" and allow the permission
- [x] Click "Query Permission" and verify that the printed state is 'granted'

_Permanent permissions_
- [x] Open a new tab and navigate to the test site
- [x] Click "Query Permission" and verify that the printed state is 'prompt'
- [x] Go to Settings --> Permissions --> Site Permissions
- [x] Change the Camera permission to Allow and the Microphone permission to Deny
- [x] Go back to the tab you just opened and refresh
- [x] Click "Query Permission" and verify that the printed state is 'granted'
- [x] From the dropdown, change "camera" to "microphone"
- [x] Click "Query Permission" and verify that the printed state is 'denied'

_Unsupported_
- [x] From the dropdown, change "microphone" to "bluetooth"
- [x] Click "Query Permission" and verify that the printed state is 'denied'

### UI changes
None

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -1283,6 +1283,7 @@ class BrowserTabFragment :
             )
 
             is Command.WebViewError -> showError(it.errorType, it.url)
+            is Command.OnPermissionsQueryResponse -> contentScopeScripts.onResponse(it.jsCallbackData)
             else -> {
                 // NO OP
             }
@@ -2129,19 +2130,7 @@ class BrowserTabFragment :
     }
 
     private fun permissionsQuery(featureName: String, method: String, id: String, data: JSONObject) {
-        val url = runBlocking(dispatchers.main()) {
-            webView?.url
-        }
-
-        val permissionState = webChromeClient.getPermissionsQueryResponse(url, tabId, data.getString("name"))
-        val response = JsCallbackData(
-            JSONObject("""{ "state":"$permissionState"}"""),
-            featureName,
-            method,
-            id,
-        )
-
-        contentScopeScripts.onResponse(response)
+        viewModel.onPermissionsQuery(featureName, method, id, data)
     }
 
     private fun configureWebViewForAutofill(it: DuckDuckGoWebView) {
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2112,7 +2112,7 @@ class BrowserTabFragment :
                     override fun process(featureName: String, method: String, id: String, data: JSONObject) {
                         when (method) {
                             "webShare" -> webShare(featureName, method, id, data)
-                            "permissionsQuery" -> permissionsQuery(featureName, method, id, data)
+                            "permissionsQuery" -> viewModel.onPermissionsQuery(featureName, method, id, data)
                             else -> {
                                 // NOOP
                             }
@@ -2127,10 +2127,6 @@ class BrowserTabFragment :
 
     private fun webShare(featureName: String, method: String, id: String, data: JSONObject) {
         webShareRequest.launch(JsCallbackData(data, featureName, method, id))
-    }
-
-    private fun permissionsQuery(featureName: String, method: String, id: String, data: JSONObject) {
-        viewModel.onPermissionsQuery(featureName, method, id, data)
     }
 
     private fun configureWebViewForAutofill(it: DuckDuckGoWebView) {
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2111,6 +2111,7 @@ class BrowserTabFragment :
                     override fun process(featureName: String, method: String, id: String, data: JSONObject) {
                         when (method) {
                             "webShare" -> webShare(featureName, method, id, data)
+                            "permissionsQuery" -> permissionsQuery(featureName, method, id, data)
                             else -> {
                                 // NOOP
                             }
@@ -2125,6 +2126,22 @@ class BrowserTabFragment :
 
     private fun webShare(featureName: String, method: String, id: String, data: JSONObject) {
         webShareRequest.launch(JsCallbackData(data, featureName, method, id))
+    }
+
+    private fun permissionsQuery(featureName: String, method: String, id: String, data: JSONObject) {
+        val url = runBlocking(dispatchers.main()) {
+            webView?.url
+        }
+
+        val permissionState = webChromeClient.getPermissionsQueryResponse(url, tabId, data.getString("name"))
+        val response = JsCallbackData(
+            JSONObject("""{ "state":"$permissionState"}"""),
+            featureName,
+            method,
+            id,
+        )
+
+        contentScopeScripts.onResponse(response)
     }
 
     private fun configureWebViewForAutofill(it: DuckDuckGoWebView) {
--- a/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
+++ b/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
@@ -57,7 +57,10 @@ class ContentScopeScriptsJsMessaging @Inject constructor(
     override val secret: String = coreContentScopeScripts.secret
     override val allowedDomains: List<String> = emptyList()
 
-    private val handlers: List<JsMessageHandler> = listOf(WebShareHandler())
+    private val handlers: List<JsMessageHandler> = listOf(
+        WebShareHandler(),
+        PermissionsQueryMessage(),
+    )
 
     @JavascriptInterface
     override fun process(message: String, secret: String) {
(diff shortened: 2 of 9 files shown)


node_modules/@duckduckgo/content-scope-scripts/build/android/contentScope.js
nshuba, line 1: Will update via `package.json` once https://github.com/duckduckgo/content-scope-scripts/pull/820/ is merged

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
```diff
@@ -48,13 +54,14 @@ class ContentScopeScriptsJsMessaging @Inject constructor(
 
     private lateinit var webView: WebView
     private lateinit var jsMessageCallback: JsMessageCallback
+    private var tabId: String? = null
```
aitorvs, line 57: @marcosholgado I know you're working on js messageing too.
Can you take al look at this. I have very limited context but this `var tabId` smells a bit. It seems very specific/hacky to be later part of the public API

node_modules/@duckduckgo/content-scope-scripts/build/android/contentScope.js
aitorvs: Any reason why this is not in a different PR?

site-permissions/site-permissions-api/src/main/java/com/duckduckgo/site/permissions/api/SitePermissionsManager.kt
```diff
@@ -20,6 +20,17 @@ import android.webkit.PermissionRequest
...
+    fun isPermissionAllowedToAsk(
+        url: String,
+        permission: String
+    ): Boolean
+
+    fun isPermissionGranted(
+        url: String,
+        tabId: String,
+        permission: String
+    ): Boolean
```
aitorvs, line 32: In the Android team we have a process for adding new public API surface. More so give this is contribution outside the functional team. Please check [this task](https://app.asana.com/0/1202561462274611/1205060262189411/f)

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
marcosholgado: Agreed. All the logic in the `PermissionsQueryMessage` should be moved to the BrowserTabViewModel and get there via the BrowserTabFragment. 

The idea being that your message handler just calls `jsMessageCallback.process(method)` and everything is delegated to the tab fragment and view model. 

You can have a look at https://github.com/duckduckgo/Android/pull/3870 which should be merged soon and does something similar. That PR also standardises a bit more the JsMessageCallback so all the data received in the message is passed along (which wasn't needed at first). That means all the API changes made here can be rolled back too.

js-messaging/js-messaging-api/src/main/java/com/duckduckgo/js/messaging/api/JsMessaging.kt
```diff
@@ -25,7 +25,7 @@ interface JsMessaging {
     /**
      * Method to register the JS interface to the webView instance
      */
-    fun register(webView: WebView, jsMessageCallback: JsMessageCallback?)
+    fun register(webView: WebView, jsMessageCallback: JsMessageCallback?, tabId: String? = null)
```
marcosholgado, line 28: This should be rolled back

js-messaging/js-messaging-api/src/main/java/com/duckduckgo/js/messaging/api/JsMessaging.kt
```diff
@@ -83,7 +83,7 @@ interface JsMessageHandler {
      * This method processes a [JsMessage] and can return a JsRequestResponse to reply to the message if needed
      * @return `JsRequestResponse` or `null`
      */
-    fun process(jsMessage: JsMessage, secret: String, webView: WebView, jsMessageCallback: JsMessageCallback): JsRequestResponse?
+    fun process(jsMessage: JsMessage, secret: String, jsMessageInfo: JsMessageInfo): JsRequestResponse?
```
marcosholgado, line 86: Once the PR I mentioned above is merged, get those changes instead of these.

js-messaging/js-messaging-api/src/main/java/com/duckduckgo/js/messaging/api/JsMessaging.kt
```diff
@@ -109,6 +109,15 @@ data class JsMessage(
     val id: String?,
 )
 
+/**
+ * Used to keep additional useful information when passing a [JsMessage] to a [JsMessageHandler]
+ */
+data class JsMessageInfo(
```
marcosholgado, line 115: This is not needed once you get the changes from the other PR.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/messaging/SubscriptionMessagingInterface.kt
```diff
@@ -26,6 +26,7 @@ import com.duckduckgo.js.messaging.api.JsMessage
 import com.duckduckgo.js.messaging.api.JsMessageCallback
 import com.duckduckgo.js.messaging.api.JsMessageHandler
 import com.duckduckgo.js.messaging.api.JsMessageHelper
+import com.duckduckgo.js.messaging.api.JsMessageInfo
```
marcosholgado, line 29: All of these changes should also be rolled back

site-permissions/site-permissions-api/src/main/java/com/duckduckgo/site/permissions/api/SitePermissionsManager.kt
marcosholgado: Yup, it would be great to discuss what are the specific needs to figure out what's the best option . 

For example, I get the difference here is that we don't get a PermissionRequest but directly a string with the actual permission, it is however not clear to me if we can get more than one permission on the same message. I'm assuming the answer is yes which then comes back to the c-s-s implementation of this but the PR is not the place to discuss that.

marcosholgado (changes requested):

node_modules/@duckduckgo/content-scope-scripts/build/android/contentScope.js
nshuba: I saw Marcos doing something similar, so thought I'd make testing easier as well. Reverted and added as a step in testing instead.

site-permissions/site-permissions-api/src/main/java/com/duckduckgo/site/permissions/api/SitePermissionsManager.kt
nshuba: Sounds good. Reworked this based on the update codebase and submitted change proposal: https://app.asana.com/0/1202552961248957/1205981799209335/f. Once we settle on it, I can add more tests for PermissionsManager.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2131,6 +2132,22 @@ class BrowserTabFragment :
         webShareRequest.launch(JsCallbackData(data, featureName, method, id))
     }
 
+    private fun permissionsQuery(featureName: String, method: String, id: String, data: JSONObject) {
+        val url = runBlocking(dispatchers.main()) {
+            webView?.url
+        }
```
aitorvs, line 2138: Why do we need this?
`runBlocking` inside an imperative function feels odd. We're doing something odd here.

marcosholgado: Agreed, overall this is kinda weird, it als calls the web chrome client when that's not needed. Instead of that I'd move the logic from the WebChromeClient to the TabViewModel, there you can check the URL by accessing `site.url` rather than jumping on the main thread. The view model can then return a command containing the jscallbackdata so contentScopeScripts can call onResponse with it.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2115,6 +2116,7 @@ class BrowserTabFragment :
                     override fun process(featureName: String, method: String, id: String, data: JSONObject) {
                         when (method) {
                             "webShare" -> webShare(featureName, method, id, data)
+                            "permissionsQuery" -> permissionsQuery(featureName, method, id, data)
```
marcosholgado, line 2119: You can directly call `viewModel.onPermissionsQuery(featureName, method, id, data)` here

PR #3882, opened 2023-11-17
Address Bar Spoofing Test Cases + Remediation

not-a-rootkit:
Task/Issue URL: https://app.asana.com/0/0/1205916690837650/f
### Description
The browser implementation of the address bar updating functionality is currently vulnerable to an Address Bar Spoofing vulnerability. In this PR I integrate all of our known address bar spoofing tests, including a fix for the basic auth address bar spoofing issues where someone can hide the domain portion of the address bar by prepending fake basic auth credentials to the domain. 

In the fix, I simply strip the username/password fields from the address bar which forces the browser to prompt for username/password. This is consistent with our other browser implementations on macOS and iOS.

### Steps to test this PR

_Security Tests_
- [x] run `maestro test .maestro/security_tests/0_all.yaml` 
- [x] ensure all tests are passing 

_Basic Auth_
- [x] visit a page that actually requires basic auth: https://notarootkit.com
- [x] check that the username/password prompt is still raised 
- [x] visit a page that doesn't require basic auth but contains "@" in the address bar: https://medium.com/@duckduckgo
- [x] ensure the address bar value is correctly displayed and not truncated

### Primary Questions:
1. How often do we want these tests to run? 
2. What configuration is required to get them to run at our required schedule?
3. What happens if the tests fail?
4. Are we comfortable with the fix?


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1440,6 +1440,9 @@ class BrowserTabViewModel @Inject constructor(
 
         return if (duckDuckGoUrlDetector.isDuckDuckGoQueryUrl(url)) {
             duckDuckGoUrlDetector.extractQuery(url) ?: url
+        } else if (url.contains("@")) {
```
CDRussell, line 1443: i'm concerned this might affect more URLs than intended; won't this also affect a url like usernames in twitter....
- https://twitter.com/@username
- https://twitter.com/username

in the case of twitter, these are treated as the same, but presume not all sites going to do that, and sometimes the `@` is valid to be included in the URL?

not-a-rootkit: Good spot, I forget '@' is allowed in the URI path of URLs. I tested https://medium.com/@duanevalz and the navigations aren't blocked but the address bar is just truncated. I'll use the java URI parser to extract the basic auth component instead.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1435,13 +1435,26 @@ class BrowserTabViewModel @Inject constructor(
...
+        val userInfo = uri.userInfo
+
+        if (userInfo != null) {
+            val queryStr = uri.rawQuery?.let { "?$it" } ?: ""
+            val uriFragment = uri.fragment?.let { "#$it" } ?: ""
+            return "${uri.scheme}://${uri.host}${uri.path}$queryStr$uriFragment"
+        }
+
+        return url
+    }
```
aitorvs, line 1449: This change affects the URL bar so I'd expect a very comprehensive set of unit tests (not only them maestro ones) to make sure we're covering decent amount of cases

marcosholgado:
Are the maestro test executed at all? If they are not part of any workflows they will never run.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
not-a-rootkit: Thanks @aitorvs - sounds good to me, I'll get those unit tests written up 👍🏼

not-a-rootkit:
Thanks for taking a look folks.

@marcosholgado - I’ve added these to run at the same intervals as Privacy Tests. Do we think this is a reasonable choice?

In addition to these I’ve added some fairly extensive unit tests for expected success and failure cases of basic auth extraction. Please let me know if you spot any cases that are missing.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1479,13 +1480,27 @@ class BrowserTabViewModel @Inject constructor(
...
+
+        if (userInfo != null) {
+            val queryStr = uri.rawQuery?.let { "?$it" } ?: ""
+            val uriFragment = uri.fragment?.let { "#$it" } ?: ""
+            val portStr = if (uri.port != -1) ":${uri.port}" else ""
+            return "${uri.scheme}://${uri.host}$portStr${uri.path}$queryStr$uriFragment"
+        }
+
+        return url
+    }
```
joshliebe, line 1495: Casting the url to a `URI` can throw an exception. 
We also generally use the `toUri()` extension function instead of `URI()`.

I would replace this function with:
```
    fun stripBasicAuthFromUrl(url: String): String {
        try {
            val uri = url.toUri()
            val userInfo = uri.userInfo

            if (userInfo != null) {
                val queryStr = uri.rawQuery?.let { "?$it" } ?: ""
                val uriFragment = uri.fragment?.let { "#$it" } ?: ""
                val portStr = if (uri.port != -1) ":${uri.port}" else ""
                return "${uri.scheme}://${uri.host}$portStr${uri.path}$queryStr$uriFragment"
            }
        } catch(e: Exception) {
            return url
        }
        return url
    }
```

You should also add a test for the case where it throws an exception.

.maestro/security_tests/0_all.yaml
```diff
@@ -0,0 +1,14 @@
+appId: com.duckduckgo.mobile.android
+tags:
+    - securityTest
```
joshliebe, line 3: Do we need this file? 

Since this is tagged with `securityTest` and all the tests (1-7) are also tagged with `securityTest`, does that not mean that everything will run twice in `end-to-end.yml`?

.github/workflows/end-to-end.yml
```diff
@@ -74,6 +74,17 @@ jobs:
...
+        if: always()
+        uses: mobile-dev-inc/action-maestro-cloud@v1.4.1
+        with:
+          api-key: ${{ secrets.MOBILE_DEV_API_KEY }}
+          name: ${{ github.sha }}
+          app-file: apk/release.apk
+          android-api-level: 30
+          workspace: .maestro
+          include-tags: securityTest
+
```
joshliebe, line 87: I ran the end-to-end test and it fails: https://github.com/duckduckgo/Android/actions/runs/7106164156/job/19344978847

Screenshot 2023-12-05 at 20 50 29
<br/><br/>
I tried running it locally and I had the same issue, this pop-up doesn't allow "run" to be clicked:
Screenshot 2023-12-05 at 20 46 22

joshliebe (changes requested):
Tested and it seems to be working as expected. However, the Maestro tests aren't passing and I had a few other issues that should be addressed before we can merge.

.maestro/security_tests/0_all.yaml
not-a-rootkit: You're right - everything would run twice this way. I was using it in testing to speed it up so I wouldn't have to repeatedly run through the onboarding process for each test. I've removed it.

.maestro/security_tests/3_-_AddressBarSpoof,_appschemes.yaml
```diff
@@ -0,0 +1,30 @@
+appId: com.duckduckgo.mobile.android.debug
```
joshliebe, line 1: Looking at https://github.com/duckduckgo/Android/actions/runs/7126639716, it only ran the first two tests.
You need to set the `appId` to `com.duckduckgo.mobile.android` for tests 3-7.

joshliebe (approved):
LGTM! Thanks 👍

PR #3856, opened 2023-11-13
Ensure autofill disable prompt disabled no matter how a credential is saved

CDRussell:
Task/Issue URL: https://app.asana.com/0/1203822806345703/1205939031559534/f 

### Description
Change the way the Autofill decline counter works to ensure it is disabled when credentials are added regardless of how they were added. 

Previously, this required the decline counter to be manually disabled at the point in the code that was adding a credential (e.g., when handing the result of save credential prompt) but wasn't handling the other places that a credential could be created (sync + manually created). 

This PR introduces a dedicated observer on the credential count that will automatically disable the decline counter if it detects one or more credentials exist.

### Steps to test this PR

Logcat filter: `package:mine & (message~:"CredentialEverPopulatedObserver" | message~:"Permanently disabling Autofill decline counter" | message~:"User declined to save credentials" | message~:"Determined if decline counter should be active")`

#### Decline counter disabled when manually adding a credential
- [x] Fresh install (`internal`)
- [x] Visit Overflow->Logins
- [x] Click the ➕ and add details for a login. tap the ✔️ to save
- [x] Verify you see the following in the logs:

```
Permanently disabling Autofill decline counter
CredentialEverPopulatedObserver stopped
```

#### Decline counter disabled from sync
- [x] Fresh install (`internal`)
- [x] Go to sync, and sync saved logins from elsewhere
- [x] Verify you see the following in the logs:

```
Permanently disabling Autofill decline counter
CredentialEverPopulatedObserver stopped
```

#### Decline counter disabled from saving a login
- [x] Fresh install (`internal`)
- [x] Visit https://fill.dev/form/login-simple. Enter username and password (>= 4 chars). Tap login button
- [x] Save when prompted
- [x] Verify you see the following in the logs:

```
Permanently disabling Autofill decline counter
CredentialEverPopulatedObserver stopped
```

#### Ensure decline counter still works
- [x] Fresh install (`internal`)
- [x] Visit https://fill.dev/form/login-simple. Enter username and password (>= 4 chars). Tap login button. **Don't save**.
- [x] Login on another page( e.g., https://privacy-test-pages.glitch.me/autofill/form-submission.html). **Don't save.**
- [x] Verify you see the prompt to "Keep saving logins"

PR #3961, opened 2023-12-04
Revert "Ensure autofill disable prompt disabled no matter how a credential is saved"

CDRussell:
Task/Issue URL: https://app.asana.com/0/0/1206091965606266/f 

### Description
Reverts https://github.com/duckduckgo/Android/pull/3856 which added an autofill database observer so it could detect when a credential was added and disabled the UX where a user would be asked to disable autofill. 

The change caused an uptick in `net.sqlcipher.database.SQLiteException`. While this change didn't change anything in the database layer, it seems to have exacerbated an underlying issue. The benefits of the change don't justify the increased number of exceptions, so **reverting this change** is best for now, then can investigate more about the root cause. 

### Steps to test this PR

- [ ] Install from `develop`
- [ ] Add a login
- [ ] Visit logins and make sure login is still accessible


cmonfortep (approved):
👍

PR #3966, opened 2023-12-05
Update autofill copy, change logins->passwords globally

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205931258794166/f 

### Description
Updates our autofill-related terminology around saved logins... these will now be referred to as `passwords` instead of `logins`. 

Several preceding PRs have made similar incremental changes already, and this one should take care of the remaining loose ends. 
- Most of the footprint of change is caused by removing existing (now outdated) translations. 
- New translations will follow separately.

### Steps to test this PR
This is QA-optional, as you should be able to assess the change from the PR diff. If you do want to do extra testing:

- [ ] Smoke test autofill functionality (now called `Passwords` in menu)
- [ ] Make sure there are no remaining references to `Logins` in the UI


CDRussell:
For some reason it still shows as "Some checks haven't completed yet" and "waiting for status to be reported" even though they are all done.

Bypassing the "wait for checks to pass" protection since it seems to be kinda broken right now; when I look into each check it's all green.

PR #4021, opened 2023-12-18
feat: ukrainian-localization

sliderzxc:
Just translated all strings to ukainian, i'm native ukrainian speaker and becase the translate is very goog

A lot of people from Ukraine don't know English and as you know, now russia attacks Ukraine, and ukrainian don't want to use russian but they want to understand everything in application, and because i created this localization

I didn't change anything in UI, i just created values-uk/strings.xml and all
Thank you, i hope that you accept this pull request and ukrainian will be in the following versions of your great application

### DIFF
--- a/app/src/main/res/values-uk/strings.xml
+++ b/app/src/main/res/values-uk/strings.xml
@@ -1,0 +1,795 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (c) 2022 DuckDuckGo
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


marcosholgado:
Thanks for opening this PR. We are currently not accepting any translations for new languages. See https://github.com/duckduckgo/Android/blob/develop/TRANSLATIONS.md

`Please note we are not looking for help with other languages at this time; only the languages listed above. We may open up requests in future for additional languages.`

I'll pass this to the team so it can be evaluated though. Gonna close this for now.

Note: *If* we were accepting new languages, note that there are multiple string files that will also require translation.

PR #4027, opened 2023-12-19
Support screen lock/unlock API

nshuba:
Task/Issue URL: https://app.asana.com/0/0/1206163779688838/f

### Description
Support [screen.lock](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock) and [screen.unlock](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/unlock) APIs

### Steps to test this PR
Before testing: apply a patch to C-S-S from https://github.com/duckduckgo/content-scope-scripts/pull/875

_Typical use case_
- [x] Open any video on YouTube and tap the full screen button - orientation should remain in portrait
- [x] Override PrivacyConfig to add an `"screenLock": "enabled"` to `webCompat` and ` "screenLock": {"state": "enabled"}` to `androidBrowserConfig`
- [x] Reload the YouTube page and press the full screen button again - orientation should switch to landscape
- [x] Use YouTube's button to exit full screen - orientation should switch back to portrait
  * Note: if you have auto-rotate ON, you just unlocked your phone, and it stayed flat on a desk - you may not have a default orientation set - in this case the orientation will remain in landscape upon full screen exit (this is the same behavior as in Chrome). If you orient your phone to portrait and try the exercise again, it should flip back automatically.
- [x] Test that disabling either the `webCompat` or the corresponding `androidBrowserConfig` disables the feature

  
_Web Share_
- [x] Smoke test web share as per https://github.com/duckduckgo/Android/pull/4027#discussion_r1433323915
  
 _Bonus_
 - [x]  Go to https://nshuba.github.io/test-site/screen.html and click Lock Orientation
 - [x] Text box should contain error indicating we need to be full screen before calling lock()
 - [x] Click Request Full Screen, select "landscape" from the dropbox and then click Lock Orientation
 - [x] Orientation should switch to landscape
 - [x] Test that the other orientation locks behave as expected (https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock)
 - [x] In the end, lock the orientation to any landscape or portrait-secondary and exit full screen (drag from the bottom and click "back")
 - [x] Orientation should change back to regular portrait


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2146,6 +2149,8 @@ class BrowserTabFragment :
                         when (method) {
                             "webShare" -> if (id != null && data != null) { webShare(featureName, method, id, data) }
                             "permissionsQuery" -> if (id != null && data != null) { viewModel.onPermissionsQuery(featureName, method, id, data) }
+                            "screenLock" -> if (id != null && data != null) { screenLock(featureName, method, id, data) }
```
marcosholgado, line 2152: I believe is time to move this when to the view model ala `SubscriptionsWebViewActivity`

app/src/main/java/com/duckduckgo/app/global/view/ActivityExtension.kt
```diff
@@ -55,7 +57,13 @@ fun Context.fadeTransitionConfig(): Bundle? {
     return config.toBundle()
 }
 
+@Synchronized
```
marcosholgado, line 60: Why the @Syncronized in all these methods?

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Updates the activity's orientation based on provided JS data
+     *
+     * @return response data
+     */
+    fun updateOrientation(data: JsCallbackData, activity: FragmentActivity?): JsCallbackData {
+        val response = if (activity == null) {
+            NO_ACTIVITY_ERROR
+        } else {
+            val requestedOrientation = data.params.getString("orientation")
```
marcosholgado, line 36: This can crash the app if `orientation` doesn't exist.

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.json.JSONObject
+
+class JsOrientationHandler {
+
+    /**
+     * Updates the activity's orientation based on provided JS data
+     *
+     * @return response data
+     */
+    fun updateOrientation(data: JsCallbackData, activity: FragmentActivity?): JsCallbackData {
```
marcosholgado, line 32: Rather than passing the activity you could pass the result of `activity.requestJsOrientationChange(matchedOrientation)` which is safer.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2162,6 +2167,15 @@ class BrowserTabFragment :
         webShareRequest.launch(JsCallbackData(data, featureName, method, id))
     }
 
+    private fun screenLock(featureName: String, method: String, id: String, data: JSONObject) {
+        val returnData = JsOrientationHandler().updateOrientation(JsCallbackData(data, featureName, method, id), activity)
```
marcosholgado, line 2171: You want to inject the JsOrientationHandler rather than creating a new instance every time not only because of that but for testing purposes.

app/src/main/java/com/duckduckgo/app/global/view/ActivityExtension.kt
nshuba: This was more of a concern earlier when the JavaBridge thread would try to change the orientation and the Main thread would enter/exit full screen (locking orientation is not allowed when the page is not full screen). With the updated code, both pathways now end up on Main, so I removed the @synchronized. Do you think we should add something like @MainThread here instead? Basically, I am trying to protect against a scenario where the user gets locked in some orientation even after exiting full screen.

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
nshuba: Yes, my JS-side unit tests uncovered this too! 😄  Will fix in the next round 👍

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
nshuba: Thanks, this is one of those parts I was iffy about, but couldn't come up with anything better 😅  Could you expand your suggestion a bit more? This method determines what value we should pass to `requestJsOrientationChange` and determines error cases as it goes, so I am not sure how we can pass it the result of `requestJsOrientationChange`? Alternatively, we could have `requestJsOrientationChange` handle different error cases and return appropriate codes, but that doesn't seem right either...

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
nshuba: For sure, this is one of those ugly-draft things. Will tackle it together with unit tests. I'd like to settle on what `JsOrientationHandler` looks like first 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -707,6 +710,8 @@ class BrowserTabFragment :
             contentScopeScripts.onResponse(it)
         }
 
+    private val jsOrientationHandler = JsOrientationHandler()
```
aitorvs, line 713: Please use DI for this.

nshuba: Yep yep - please see https://github.com/duckduckgo/Android/pull/4027#discussion_r1433339787

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3061,7 +3082,19 @@ class BrowserTabViewModel @Inject constructor(
...
+
+    private fun screenLock(featureName: String, method: String, id: String, data: JSONObject) {
+        viewModelScope.launch(dispatchers.main()) {
+            command.value = ScreenLock(JsCallbackData(data, featureName, method, id))
+        }
+    }
+
+    private fun screenUnlock() {
+        viewModelScope.launch(dispatchers.main()) {
+            command.value = ScreenUnlock
```
aitorvs, line 3097: Can we feature flag both screenLock and screenUnlock using a kill-switcg in `AndroidBrowserConfig`?
Something like.

```kotlin
    private fun screenUnlock() {
            if (andoidBrowserConfig.screenLock().isEnabled()) {
                withContext(dispatchers.main()) {
                    command.value = ScreenUnlock
                }
            }
        }
    }
```

I think this change is high potential to cause a regression and I want to avoid having to hotfix anything like we had recently with similar/related changes.

app/src/main/java/com/duckduckgo/app/global/view/ActivityExtension.kt
```diff
@@ -64,7 +71,20 @@ fun FragmentActivity.toggleFullScreen() {
     window.decorView.systemUiVisibility = newUiOptions
 }
 
+fun FragmentActivity.isFullScreen(): Boolean {
```
aitorvs, line 74: This should be `private`

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
aitorvs: It feels the only intention to `JsOrientationHandler` class to extract this code our of the `BrowserTabFragment`. But otherwise it's an implementation detail of `BrowserTabFragment`, always to be instantiated by it not to leak any memory.

However having `JsOrientationHandler` as a separate class, in a separate package that accepts a generic activity as a parameter is the wrong contract, for the same above reason and because it sets easy precedent for others to copy-paste anti-patterns.

To avoid leaking what I would do is:
* Move `JsOrientationHandler` inside the `BrowserTabFragment.kt` file as a private class
* (forget about my comment about using DI, ie. instantiate as `private val jsOrientationHandler = JsOrientationHandler()` and leave a comment as to why
* change the contract of `updateOrientation` to be
```kotlin
fun updateOrientation(data: JsCallbackData, browserTabFragment: BrowserTabFragment)
```

app/src/main/java/com/duckduckgo/app/global/view/ActivityExtension.kt
```diff
@@ -64,7 +71,20 @@ fun FragmentActivity.toggleFullScreen() {
...
 }
+
+fun FragmentActivity.requestJsOrientationChange(newOrientation: JavaScriptScreenOrientation): Boolean {
+    if (!isFullScreen()) {
+        return false
+    }
+
+    requestedOrientation = newOrientation.nativeValue
+    return true
+}
```
aitorvs, line 90: There is no need to expose `JavaScriptScreenOrientation`. Pass an int instead to make it generic
```suggestion
fun FragmentActivity.requestJsOrientationChange(newOrientation: Int): Boolean {
    if (!isFullScreen()) {
        return false
    }

    requestedOrientation = newOrientation
    return true
}
```

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
```diff
@@ -0,0 +1,75 @@
...
+        val response = if (activity == null) {
+            NO_ACTIVITY_ERROR
+        } else {
+            val requestedOrientation = data.params.getString("orientation")
+            val matchedOrientation = JavaScriptScreenOrientation.values().find { it.jsValue == requestedOrientation }
```
aitorvs, line 37:
```suggestion
            val matchedOrientation = JavaScriptScreenOrientation.entries.find { it.jsValue == requestedOrientation }
```

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
```diff
@@ -0,0 +1,75 @@
...
+        const val EMPTY = """{}"""
+        const val NOT_FULL_SCREEN_ERROR = """{"failure":{"name":"InvalidStateError","message":
+            "The page needs to be fullscreen in order to call screen.orientation.lock()"}}"""
+        const val TYPE_ERROR = """{"failure":{"name":"TypeError","message":
+            "Failed to execute 'lock' on 'ScreenOrientation': The provided value '%s' is not a valid enum value of type OrientationLockType."}}"""
+        const val NO_ACTIVITY_ERROR = """{"failure":{"name":"InvalidStateError","message":"The page is not tied to an activity"}}"""
+    }
+}
+
+enum class JavaScriptScreenOrientation(val jsValue: String, val nativeValue: Int) {
```
aitorvs, line 66: See other comments and as a result, make this `private enum class` and be it a subclass of `JsOrientationHandler`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
nshuba: We will have a feature flag on the C-S-S side: https://github.com/duckduckgo/content-scope-scripts/blob/nastia/screen_lock/src/features/web-compat.js#L114 so it won't send messages for either screenLock nor screenUnlock. Do you think we need an additional feature flag on the native side?

app/src/main/java/com/duckduckgo/app/global/view/ActivityExtension.kt
nshuba: Won't that allow any int to be passed? Don't we want the values to be restricted to allowed orientation values? I could add some sanity checks, but that will add another possible error here...

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
nshuba: It's true - the intention of `JsOrientationHandler` was to keep that code outside of `BrowserTabFragment`. We could just plop it back in as you suggest, Aitor. 

Another idea I've been thinking about is creating a `FragmentActivity` extension inside the `com.duckduckgo.app.browser` package so it's not as exposed and move the logic of `updateOrientation` to there.

What do you guys think? I have no strong opinions here.

app/src/main/java/com/duckduckgo/app/global/view/ActivityExtension.kt
aitorvs: Your API is a bit convoluted/odd. 
* naming: what you're doing is mapping js string orientation to android screen (activity info) orientation, yet the enum is called `JavaScriptScreenOrientation`
* the extension function that requests the orientation is called `requestJsOrientationChange` yet it doesn't have anything to do with JS, its Android
* the extension function depends on a type (`JavaScriptScreenOrientation`) defined in a very different place

What I would do
* rename `requestJsOrientationChange` to `requestOrientationChangeFromJsOrientation`
* signature be the one below and perform the checks inside it
```kotlin
fun FragmentActivity.requestJsOrientationChange(newOrientation: String): Result<Boolean>
```
* Modify `updateOrientation` it just calls `requestJsOrientationChange(requestedOrientation)` and then handles the `Result<Boolean>` same way you handle now the errors

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
aitorvs: My main point is that this is an implementation detail of `BrowserTabFragment`, not of any other generic type. So we should not create public API surface on generic types when we can't/should not use them

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
aitorvs: I'd like to have a kill-switch in native. We've had issues with feature flags in JS before.

app/src/main/java/com/duckduckgo/app/browser/orientation/JsOrientationHandler.kt
nshuba: Makes sense - addressed

app/src/main/java/com/duckduckgo/app/global/view/ActivityExtension.kt
nshuba: True, it was a bit funky. I re-named and re-architected things a bit. Seems cleaner now, let me know what you think.

aitorvs (approved):
LGTM!

Note: haven't tested it manually but Nastia and Ana did AFAIK. Feel free to merge if that's the case.

anikiki:
Tested and works as expected! :muscle:

PR #4033, opened 2023-12-20
Android version filter for experiments

nalcalag:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206193424728741/f

### Steps to test this PR

- Change `const val PRIVACY_REMOTE_CONFIG_URL` to `https://www.jsonblob.com/api/1168612786158034944`
- There is a log in `VariantManagerImpl` _Line 136_ when a new variant allocation is made. This way, you can check if the variant has been allocated correctly

_Android version filter_
- Add the API version of your device you are testing with in the `filters` parameter.in the JSONBlob endpoint e.g.
Screenshot 2023-12-21 at 11 57 51

- Fresh install
- [x] Check the experimental variant is allocated correctly

_Android version filter doesn't match_
- Add a different API version from your device you are testing with in the `filters` parameter in the JSONBlob endpoint e.g.
Screenshot 2023-12-21 at 11 57 51

- Fresh install
- [x] Check the default variant is allocated

_No filters_
- Remove filters parameter or leave it empty
- Fresh install
- [x] Check the experimental variant is allocated correctly

_Filters combined 1_
- Add the API version of your device you are testing with in the `filters` parameter.in the JSONBlob endpoint e.g.
Screenshot 2023-12-21 at 11 57 51

- Add a different locale filter which doesn't match your testing device. _e.g._ `"locale": ["en_US"]`
- Fresh install
- [ ] Check the default variant is allocated

_Filters combined 2_
- Add the API version of your device you are testing with in the `filters` parameter.in the JSONBlob endpoint e.g.
Screenshot 2023-12-21 at 11 57 51

- Add the locale which matched your testing device.
- Fresh install
- [x] Check the experimental variant is allocated correctly

_Filters combined 3_
- Add a different API version from your device you are testing with in the `filters` parameter in the JSONBlob endpoint _e.g._
Screenshot 2023-12-21 at 11 57 51

- Add a different locale filter which doesn't match your testing device. _e.g._ `"locale": ["en_US"]`
- Fresh install
- [x] Check the default variant is allocated

---------------
Improved steps in this [comment](https://github.com/duckduckgo/Android/pull/4033#pullrequestreview-1829408345) by @marcosholgado

### DIFF
--- a/experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
+++ b/experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
@@ -19,7 +19,7 @@ package com.duckduckgo.experiments.api
 data class VariantConfig(
     val variantKey: String,
     val weight: Double? = 0.0,
-    val filters: VariantFilters = VariantFilters(),
+    val filters: VariantFilters? = null,
 )
 
 data class VariantFilters(
--- a/experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
+++ b/experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
@@ -24,4 +24,5 @@ data class VariantConfig(
 
 data class VariantFilters(
     val locale: List<String>? = null,
+    val androidVersion: List<String>? = null,
 )
--- a/experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
+++ b/experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
@@ -19,10 +19,10 @@ package com.duckduckgo.experiments.api
 data class VariantConfig(
     val variantKey: String,
     val weight: Double? = 0.0,
-    val filters: VariantFilters? = null,
+    val filters: VariantFilters = VariantFilters(),
 )
 
 data class VariantFilters(
-    val locale: List<String>? = null,
-    val androidVersion: List<String>? = null,
+    val locale: List<String> = emptyList(),
+    val androidVersion: List<String> = emptyList(),
 )
--- a/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/ExperimentVariantRepository.kt
+++ b/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/ExperimentVariantRepository.kt
@@ -48,6 +48,7 @@ class ExperimentVariantRepositoryImpl @Inject constructor(
                     key = it.variantKey,
                     weight = it.weight,
                     localeFilter = it.filters?.locale.orEmpty(),
+                    androidVersionFilter = it.filters?.androidVersion.orEmpty(),
                 ),
             )
         }
--- a/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/ExperimentVariantRepository.kt
+++ b/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/ExperimentVariantRepository.kt
@@ -47,8 +47,8 @@ class ExperimentVariantRepositoryImpl @Inject constructor(
                 ExperimentVariantEntity(
                     key = it.variantKey,
                     weight = it.weight,
-                    localeFilter = it.filters.locale,
-                    androidVersionFilter = it.filters.androidVersion,
+                    localeFilter = it.filters?.locale.orEmpty(),
+                    androidVersionFilter = it.filters?.androidVersion.orEmpty(),
                 ),
             )
         }
--- a/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/ExperimentVariantRepository.kt
+++ b/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/ExperimentVariantRepository.kt
@@ -47,8 +47,8 @@ class ExperimentVariantRepositoryImpl @Inject constructor(
                 ExperimentVariantEntity(
                     key = it.variantKey,
                     weight = it.weight,
-                    localeFilter = it.filters?.locale.orEmpty(),
-                    androidVersionFilter = it.filters?.androidVersion.orEmpty(),
+                    localeFilter = it.filters.locale,
+                    androidVersionFilter = it.filters.androidVersion,
                 ),
             )
         }
--- a/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
+++ b/experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
@@ -107,18 +107,14 @@ class VariantManagerImpl @Inject constructor(
 
         var matchesLocaleFilter = true
         var matchesAndroidVersionLocale = true
-        when {
-            entity.localeFilter.isNotEmpty() -> {
-                val userLocale = Locale.getDefault()
-                matchesLocaleFilter = entity.localeFilter.contains(userLocale.toString())
-            }
-            entity.androidVersionFilter.isNotEmpty() -> {
-                val userAndroidVersion = appBuildConfig.sdkInt.toString()
-                matchesAndroidVersionLocale = entity.androidVersionFilter.contains(userAndroidVersion)
-            }
-            else -> {
-                return { noFilter() }
-            }
+
+        if (entity.localeFilter.isNotEmpty()) {
+            val userLocale = Locale.getDefault()
+            matchesLocaleFilter = entity.localeFilter.contains(userLocale.toString())
+        }
+        if (entity.androidVersionFilter.isNotEmpty()) {
+            val userAndroidVersion = appBuildConfig.sdkInt.toString()
+            matchesAndroidVersionLocale = entity.androidVersionFilter.contains(userAndroidVersion)
         }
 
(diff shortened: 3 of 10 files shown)


nalcalag:
* `develop`
  * **PR #4033** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4033?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/4033?utm_source=stack-comment).

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
```diff
@@ -106,12 +106,20 @@ class VariantManagerImpl @Inject constructor(
         if (entity.key == "sc" || entity.key == "se") {
             return { isSerpRegionToggleCountry() }
         }
-        if (entity.localeFilter.isEmpty()) {
-            return { noFilter() }
+
+        var matchesLocaleFilter = true
```
marcosholgado, line 110: This can all be simplified to a couple of lines

```
val matchesLocale = entity.localeFilter.contains(Locale.getDefault().toString())
val matchesAndroidVersion = entity.androidVersionFilter.contains(appBuildConfig.sdkInt.toString())
```

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
```diff
@@ -106,12 +106,20 @@ class VariantManagerImpl @Inject constructor(
...
+        var matchesAndroidVersionLocale = true
```
marcosholgado, line 111: This name is confusing, the filter is only for the version, not for the locale.

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
```diff
@@ -106,12 +106,20 @@ class VariantManagerImpl @Inject constructor(
...
+            matchesLocaleFilter = entity.localeFilter.contains(userLocale.toString())
+        }
+        if (entity.androidVersionFilter.isNotEmpty()) {
+            val userAndroidVersion = appBuildConfig.sdkInt.toString()
+            matchesAndroidVersionLocale = entity.androidVersionFilter.contains(userAndroidVersion)
         }
 
-        val userLocale = Locale.getDefault()
-        return { entity.localeFilter.contains(userLocale.toString()) }
+        return { matchesLocaleFilter && matchesAndroidVersionLocale }
```
marcosholgado, line 122: High-level I don't think this scales nicely. I think is ok for now but we need to think this through. cc/ @aitorvs

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
```diff
@@ -106,12 +106,20 @@ class VariantManagerImpl @Inject constructor(
         if (entity.key == "sc" || entity.key == "se") {
             return { isSerpRegionToggleCountry() }
         }
-        if (entity.localeFilter.isEmpty()) {
-            return { noFilter() }
+
```
marcosholgado, line 109: All this logic needs testing. We cannot write critical code without creating tests for it.

marcosholgado (changes requested):
Needs tests, code can be simplified and would like to talk about how the filters can properly scale in the future.

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
nalcalag: You are right, the problem is that filters are set in a private function within the class. I suppose I can create another layer for experiment filters and test it properly.

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
nalcalag: yes, it doesn't. We'll need to include every filter here when added. When adding next filter I can probably create an array of filters and just check if all of them are true at this point.

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
nalcalag: oops! I'll remove the Locale part

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
nalcalag: It needs to be true by default though

marcosholgado: gotcha

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
nalcalag: I think creating a new class to manage filters is a better approach that scales much better. We'll be adding more, like country or language, so I think it's totally worth creating the new class now and not overcomplicating the Test class

experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
```diff
@@ -24,4 +24,5 @@ data class VariantConfig(
 
 data class VariantFilters(
     val locale: List<String>? = null,
+    val androidVersion: List<String>? = null,
```
aitorvs, line 27: I don't get why we need nullable lists? a null list should rather be an empty list.
IIRC JSON (de)serialization returns empty list when not present

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/ExperimentVariantRepository.kt
```diff
@@ -48,6 +48,7 @@ class ExperimentVariantRepositoryImpl @Inject constructor(
                     key = it.variantKey,
                     weight = it.weight,
                     localeFilter = it.filters?.locale.orEmpty(),
+                    androidVersionFilter = it.filters?.androidVersion.orEmpty(),
```
aitorvs, line 51: Ditto, if `locale` and `androidVersion` are not nullable we can ditch the `?.orEmtpy()`

experiments/experiments-impl/src/main/java/com/duckduckgo/experiments/impl/VariantManagerImpl.kt
nalcalag: I changed the implementation to scale better in the end

experiments/experiments-api/src/main/java/com/duckduckgo/experiments/api/VariantConfig.kt
nalcalag: Sure, I was duplicating the behaviour from locale filter but makes sense to make it not nullable as I don't see the case we want to return a null here instead an empty list

marcosholgado (changes requested):
The remote config is not getting stored anymore so variants are never assigned.

marcosholgado (approved):
LGTM. 

The test steps are not accurate and difficult to follow, the images of the json are super misleading and is the same image repeated over and over. I would have used just simple code showing exactly what to put.

*Pre Steps*
- [ ] Change `const val PRIVACY_REMOTE_CONFIG_URL` to `https://www.jsonblob.com/api/1168612786158034944`
- [ ] There is a log in `VariantManagerImpl` _Line 136_ when a new variant allocation is made. This way, you can check if the variant has been allocated correctly
- [ ] Before every test you need to delete the DuckDuckGo folder from downloads

_Android version filter_
- [ ] Change JSON blob to the one below in the variant `ma` and replace `YourDeviceAPIHere` with your current device API.
```
{
        "desc": "Control group for notification permissions experiment",
        "variantKey": "ma",
        "weight": 1,
        "filters": {
          "androidVersion": [
            "34",
            "33",
            "YourDeviceAPIHere"
          ]
        }
      }
```
- [ ] Delete the DuckDuckGo folder from Downloads
- [ ] Fresh install
- [ ] Check the experimental variant is allocated correctly

_Android version filter doesn't match_
- [ ] Change JSON blob to the one below in the variant `ma` and remove your device API from the filter list.
```
{
        "desc": "Control group for notification permissions experiment",
        "variantKey": "ma",
        "weight": 1,
        "filters": {
          "androidVersion": [
            "34",
            "33",
          ]
        }
      }
```
- [ ] Delete the DuckDuckGo folder from Downloads
- [ ] Fresh install
- [ ] Check the default variant is allocated

_No filters_
- [ ] Remove filters parameter or leave them empty from all the variants
- [ ] Delete the DuckDuckGo folder from Downloads
- [ ] Fresh install
- [ ] Check the experimental variant is allocated correctly

etc, etc, etc

nalcalag:
Thanks @marcosholgado ! I totally agree, I added the link to your comment in the steps section

PR #4036, opened 2023-12-21
Experiment: Remove notifications prompt for Android 13+

nalcalag:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206161927270269/f

### Description
Android experiment implementation where the experimental group won't see notifications permissions prompt when app is installed.

### Steps to test this PR
_Pre Steps_
- [x] Change const val `PRIVACY_REMOTE_CONFIG_URL` to `https://www.jsonblob.com/api/1168612786158034944`
- [x] There is a log in `VariantManagerImpl` _Line 136_ when a new variant allocation is made. This way, you can check if the variant has been allocated correctly
- [x] Before every test you need to delete the DuckDuckGo folder from downloads

_Experimental variant_
- [x] Change JSON blob to the one below in the variant `mg` and make sure you have `androidVersion` filter set to 33 and 34.
```
{
        "desc": "No notifications permissions prompt experimental group",
        "variantKey": "mg",
        "weight": 1,
        "filters": {
          "androidVersion": [
            "34",
            "33"
          ]
        }
}
```
- [x] Delete the DuckDuckGo folder from Downloads
- [x] Fresh install
- [x] Check experimental variant is assigned
- [x] Check you don't see notifications permissions prompt when first install

_Control variant_
- [x] Change JSON blob to the one below in the variant `mf` and make sure you have `androidVersion` filter set to 33 and 34.
```
{
        "desc": "No notifications permissions prompt experimental group",
        "variantKey": "mf",
        "weight": 1,
        "filters": {
          "androidVersion": [
            "34",
            "33"
          ]
        }
}
```
- [x] Delete the DuckDuckGo folder from Downloads
- [x] Fresh install
- [x] Check control variant is assigned
- [x] Check you see notifications permissions prompt when first install

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
@@ -61,7 +61,7 @@ class WelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome) 
         // displayed on top of the onboarding.
         if (view?.windowVisibility == View.VISIBLE) {
             // Nothing to do at this point with the result. Proceed with the welcome animation.
-            scheduleWelcomeAnimation(ANIMATION_DELAY_AFTER_NOTIFICATIONS_PERMISSIONS_HANDLED)
+            scheduleTypingAnimation()
         }
     }
 
@@ -74,7 +74,7 @@ class WelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome) 
     private val events = MutableSharedFlow<WelcomePageView.Event>(replay = 0, extraBufferCapacity = 1)
 
     private val welcomePageViewModel: WelcomePageViewModel by lazy {
-        ViewModelProvider(this, viewModelFactory).get(WelcomePageViewModel::class.java)
+        ViewModelProvider(this, viewModelFactory)[WelcomePageViewModel::class.java]
     }
 
     private val binding: ContentOnboardingWelcomeBinding by viewBinding()
@@ -86,6 +86,7 @@ class WelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome) 
         super.onViewCreated(view, savedInstanceState)
 
         configureDaxCta()
+        scheduleWelcomeAnimation()
         setSkipAnimationListener()
 
         lifecycleScope.launch {
@@ -94,8 +95,16 @@ class WelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome) 
                 .flowOn(dispatcherProvider.io())
                 .collect(::render)
         }
+    }
 
-        requestNotificationsPermissions()
+    private fun scheduleWelcomeAnimation() {
+        welcomeAnimation = ViewCompat.animate(binding.welcomeContent as View)
+            .alpha(MIN_ALPHA)
+            .setDuration(ANIMATION_DURATION)
+            .setStartDelay(ANIMATION_DELAY)
+            .withEndAction {
+                requestNotificationsPermissions()
+            }
     }
 
     private fun requestNotificationsPermissions() {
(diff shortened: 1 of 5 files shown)


nalcalag:
* `develop`
  * **PR #4036** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4036?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/4036?utm_source=stack-comment).

marcosholgado:
Sharing the same jsonblob with the other PR is making this pretty confusing. We should be using a different one configured for each PR.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
```diff
@@ -95,17 +94,23 @@ class WelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome)
...
-            requestPermission.launch(Manifest.permission.POST_NOTIFICATIONS)
+            event(WelcomePageView.Event.OnNotificationPermissionsRequested)
         } else {
             scheduleWelcomeAnimation()
         }
     }
 
+    @SuppressLint("InlinedApi")
+    private fun showNotificationsPermissionsPrompt() {
+        requestPermission.launch(Manifest.permission.POST_NOTIFICATIONS)
```
marcosholgado, line 111: This method only works in TIRAMISU, we should be checking that here. I know if I follow the chain up, it's checked there but it doesn't stop anyone to directly call this without knowing what it does internally.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
```diff
@@ -95,17 +94,23 @@ class WelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome)
...
         }
+
+        requestNotificationsPermissions()
     }
 
-    @SuppressLint("InlinedApi")
     private fun requestNotificationsPermissions() {
         if (appBuildConfig.sdkInt >= android.os.Build.VERSION_CODES.TIRAMISU) {
-            requestPermission.launch(Manifest.permission.POST_NOTIFICATIONS)
+            event(WelcomePageView.Event.OnNotificationPermissionsRequested)
```
marcosholgado, line 103: I don't follow why we have this new event? Can't we just check here if the variant is assigned then request permissions otherwise do the welcome animation?

It looks like this new event throws more events to basically end up doing the same thing that we already have here?

marcosholgado (changes requested):
This PR doesn't work with the steps in the description. The androidVersion filter doesn't exist so it will never be able to filter by version. 

Regardless of that, if I remove the filter I never get any variants assigned.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt
nalcalag: We need to create an Event to check if the experiment is enabled and the user variant on the ViewModel.
The welcome page has a different pattern which we don't use in other fragments, but I prefer to follow it for consistency.

marcosholgado (approved):
@nalcalag  overall looks good to me, there's just one thing. 

Even though I have the variant mg and notifications are disabled, the notifications pixels `mnot_s_cd` and `mnot_s_pp` are still sent. The notifications are not shown as expected but we are registering them as seen and sending the pixels which we probably shouldn't.

I would probably guard in the `AndroidNotificationScheduler` the `scheduleUnusedAppNotifications` method so it's only executed if the notifications permission has been granted.

nalcalag:
@marcosholgado Interesting catch! 💯 I would probably handle that as a follow-up task in a different PR, so it doesn't get deleted when removing the experiment. This logic should stay in the code for possible future implementations or experiments. Thanks a lot for the discovery!

PR #4070, opened 2024-01-11
Add pixel for page load time

CrisBarreiro:
Task/Issue URL: https://app.asana.com/0/0/1206297995552832/f 

### Description
See title

### Steps to test this PR
_1_
- [x] Update existing version of the app
- [x] Check app starts normally

_2_
- [x] Go to OfflinePixelScheduler and reduce time to 15 minutes to make testing easier
- [x] Open wikipedia.org
- [x] Inspect app db and check there's a new entry in page_loaded_pixel_entity
- [x] Check no pixel is fired immediately
- [x] Check pixel is fired later on

_3_
- [x] Open example.com
- [x] Inspect app db and check there's a new entry in page_loaded_pixel_entity

_4_
- [x] Turn data and WiFi off
- [x] Try to load wikipedia
- [x] Wait for 30 seconds
- [x] Enable data and/or Wifi
- [x] Try to load wikipedia again
- [x] Inspect app db and check there's a new entry in page_loaded_pixel_entity and elapsed time is not 30+ seconds

### UI changes
No UI changes


app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -79,10 +90,13 @@ class BrowserWebViewClient @Inject constructor(
     private val pixel: Pixel,
     private val crashLogger: CrashLogger,
     private val jsPlugins: PluginPoint<JsInjectorPlugin>,
+    private val currentTimeProvider: CurrentTimeProvider,
```
aitorvs, line 93: Is it worth having this as separate class rather than just a `private fun`?

app/src/main/java/com/duckduckgo/app/browser/pageloadpixel/PageLoadedHandler.kt
```diff
@@ -0,0 +1,68 @@
...
+    "nytimes.com",
+    "reddit.com",
+    "twitch.tv",
+    "twitter.com",
+    "wikipedia.org",
+    "weather.com",
+)
+
+interface PageLoadedHandler {
+    operator fun invoke(url: String, start: Long, end: Long)
```
aitorvs, line 43: nit: I've always found confusing using `invoke` operator. Why not just a `fun`?

aitorvs (approved):
Left couple comments

PR #4085, opened 2024-01-17
Handle subs across platforms

marcosholgado:
Task/Issue URL: https://app.asana.com/0/488551667048375/1205557379404217/f 

### Description
See task

### Steps to test this PR
See task

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/settings/views/ItrSettingViewModel.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/settings/views/ItrSettingViewModel.kt
@@ -63,6 +63,8 @@ class ItrSettingViewModel(
             subscriptions.getEntitlementStatus(ITR_PRODUCT_NAME).also {
                 if (it.isSuccess) {
                     _viewState.emit(viewState.value.copy(hasSubscription = it.getOrDefault(NotFound) == Found))
+                } else {
+                    _viewState.emit(viewState.value.copy(hasSubscription = false))
                 }
             }
         }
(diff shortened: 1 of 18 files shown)


marcosholgado:
* `develop`
  * **#4085** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4085?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

 Join @marcosholgado and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/settings/views/ItrSettingViewModel.kt
```diff
@@ -63,6 +63,8 @@ class ItrSettingViewModel(
             subscriptions.getEntitlementStatus(ITR_PRODUCT_NAME).also {
                 if (it.isSuccess) {
                     _viewState.emit(viewState.value.copy(hasSubscription = it.getOrDefault(NotFound) == Found))
+                } else {
+                    _viewState.emit(viewState.value.copy(hasSubscription = false))
```
karlenDimla, line 67: Consider this scenario:
- User attempts to enable NetP (Connecting…)
- User returns to quickly Settings page
- Most likely the API calls would fail since we are changing the network tun.
- Netp successfully connects

we will get failure for the subscriptions.getEntitlementStatus call and we will hide the entry. We can only show it if the user goes in and out of the settings page again. Not 100% sure what the best approach here.

marcosholgado: Caught up on MM about it. This is a known issue for both ITR and PIR that will get solved in the next project when plugin in the correct screens (right now just shows a toast).

PR #4096, opened 2024-01-19
Privacy Protections Popup UI updates + "Don't Show Again"

lmac012:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206391673671159/f

#### Shows "Don't Show Again" option instead of "Dismiss" on subsequent appearences
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
- [x] Verify that the popup is visible and dismiss it
- [x] Navigate to a different webpage
- [x] Refresh
- [x] Verify that the popup is visible and contains "Don't Show Again" button instead of "Dismiss"
- [x] Tap "Don't Show Again" to dismiss the popup
- [x] Navigate to a different webpage
- [x] Refresh
- [x] Verify that the popup is not shown

#### Shows blue, animated highlight over the shield icon
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
- [x] Verify that the popup is visible, and there is a blue, animated highlight over the shield icon

#### Popup buttons are stacked vertically when they don't fit side by side
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
- [x] Verify that the popup is shown. Unless it's a very small phone, popup buttons should be laid out next to each other horizontally
- [x] Increase font/display size in system settings and repeat previous steps 
- [x] Verify that the popup falls back to layout with buttons stacked vertically once there is not enought space to display them horizontally

#### Popup stretches to the entire width of the screen on small screens
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
- [x] Verify that the popup covers entire screen width (with small, 4dp side margins) 
- [x] Rotate the screen to landscape
- [x] Verify that the popup wraps its content

#### Clicking on the highlighted dashboard icon opens the Privacy Dashboard
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
- [x] Verify that the popup is visible
- [x] Click on the shield icon while the popup is visible
- [x] Verify that the the Privacy Dashboard is opened
- [x] Navigate back to the browser
- [x] Verify that the popup is no longer visible

#### Tapping on the address bar dismisses the popup
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
- [x] Verify that the popup is visible and tap on the address bar above the popup
- [x] Verify that the popup has been dismissed

### UI changes
| Before  | After |
| ------ | ----- |
[...]

PR #4090, opened 2024-01-18
Privacy Protections Popup trigger adjustments

lmac012:
Task/Issue URL: https://app.asana.com/0/0/1206380388310627/f

#### Feature flag controls the popup
- [x] Remove @InternalAlwaysEnabled annotation from PrivacyProtectionsPopupFeature::self() and install internal build
- [x] Verify that you can't trigger the popup (feature flag is disabled in current config).
- [x] Override the privacy config with one that has the privacyProtectionsPopup flag set to true (e.g., https://jsonblob.com/api/1197271020888514560)
- [x] Verify that the popup appears on page refresh

#### Fire button does not make the popup re-appear if protections toggle was used
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Toggle protections off and on again (using the Privacy Dashboard or options menu item)
- [x] Refresh the page 
- [x] Verify there is no popup
- [x] Use the fire button
- [x] Navigate to the same site as before
- [x] Refresh
- [x] Verify there is no popup

#### Using protections toggle suppresses popup on all sites for 30 days
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Disable protections in any of the following ways:
   - Use the toggle on the Privacy Dashboard
   - Use the toggle on the Broken Site screen
   - Tap "Disable Protections" in the options menu
- [x] Refresh
- [x] Verify that popup does not appear
- [x] Navigate to a different site
- [x] Refresh
- [x] Verify that the popup does not appear
- [x] Change the date in system settings and move it at least 30 days forward
- [x] Navigate to the same site as before
- [x] Refresh
- [x] Verify that the popup appears on screen

#### Popup is shown again for the same domain after 48 hours
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
- [x] Dismiss popup 
- [x] Refresh
- [x] Verify that popup does not appear
- [x] Change the date in system settings and move it at least 2 days forward
- [x] Navigate to the same site as before
- [x] Refresh
- [x] Verify that the popup appears on screen

PR #3979, opened 2023-12-07
Privacy Protections Popup

lmac012:
Task/Issue URL: https://app.asana.com/0/1205648422731273/1206122343139850/f

### Description

This feature adds popup prompting user to try disabling protections when they refresh the page (which may be a result of observed site breakage). In this PR it is enabled only for debug builds, a feature flag and experiment will be set up in a following / stacked PR. 

### Steps to test this PR
The feature persists some data to local database - I recommend using the fire button before testing each scenario.

#### Shows popup on user-generated page refresh
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page (either from options menu or by pull-to-refresh)
- [x] Verify that the popup appears on screen
- [x] Dismiss popup (either by touch outside or the dismiss button)
- [x] Refresh again
- [x] Verify that the popup does not appear this time
- [x] Open the same page in new tab
- [x] Refresh
- [x] Verify that the popup does not appear this time
- [x] Navigate to a different site in the same tab
- [x] Refresh
- [x] Verify that the popup appears on screen

#### Protections can be disabled directly from popup
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page (either from options menu or by pull-to-refresh)
- [x] Tap "Turn Off Protections" on the popup
- [x] Verify that the popup disappeared and website reloaded with protections disabled

#### Popup is not shown when protections are already disabled
- [x] Navigate to a webpage that has protections disabled e.g. strava.com
- [x] Refresh
- [x] Verify that the popup does not appear

#### Popup is not shown for sites under DDG domain
- [x] Type something into search, e.g. "weather"
- [x] Refresh
- [x] Verify that the popup does not appear

#### Using protections toggle suppresses popup on all sites for 2 weeks
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Disable protections in any of the following ways:
   - Use the toggle on the Privacy Dashboard
   - Use the toggle on the Broken Site screen
   - Tap "Disable Protections" in the options menu
- [x] Refresh
- [x] Verify that popup does not appear
- [x] Navigate to a different site
- [x] Refresh
- [x] Verify that the popup does not appear
- [x] Change date in system settings and move it at least 14 days forward
- [x] Navigate to the same site as before
- [x] Refresh
- [x] Verify that the popup appears on screen

#### Popup is shown again for the same domain after 24 hours
- [x] Navigate to a webpage, eg. www.wikipedia.org
- [x] Refresh the page 
[...]

PR #4102, opened 2024-01-22
Handle daily and unique pixels in the statistics module

lmac012:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1206347783995803/f

### Description

This PR extends the statistics module API to allow sending unique and daily pixels.

### Steps to test this PR

- [ ] PR checks pass

### No UI changes


lmac012:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4102?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `develop`
  * **#3979** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/3979?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **#4090** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4090?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **#4096** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **#4102** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4102?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

 Join @lmac012 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

statistics/src/main/java/com/duckduckgo/app/statistics/store/TimeProvider.kt
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.duckduckgo.app.statistics.store
+
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import java.time.Instant
+import javax.inject.Inject
+
+interface TimeProvider {
```
CrisBarreiro, line 24: There's already `CurrentTimeProvider` in `common-utils`, would it be possible to agree on a common way to get a timestamp (`Instant.now` vs `SystemClock.elapsedRealtime` vs `System.currentTimeMillis`) and not have different pixels getting timestamps from different classes?

statistics/src/main/java/com/duckduckgo/app/statistics/pixels/Pixel.kt
```diff
@@ -84,16 +88,20 @@ interface Pixel {
         const val FIRE_ANIMATION_NONE = "fann"
     }
 
+    enum class PixelType { DEFAULT, DAILY, UNIQUE }
```
CrisBarreiro, line 91: I think it would be good to document here the differences between each PixelType

anrs/anrs-impl/src/main/java/com/duckduckgo/app/anr/AnrOfflinePixelSender.kt
```diff
@@ -45,7 +46,8 @@ class AnrOfflinePixelSender @Inject constructor(
                         ANR_WEBVIEW_VERSION to it.webView,
                     ),
                     mapOf(),
-                ).doOnComplete {
+                    DEFAULT,
+                ).ignoreElement().doOnComplete {
```
CrisBarreiro, line 50: Why do we need the `ignoreElement` here?

statistics/src/main/java/com/duckduckgo/app/statistics/api/PixelSender.kt
```diff
@@ -163,4 +187,25 @@ class RxPixelSender @Inject constructor(
     private fun getAtbInfo() = statisticsDataStore.atb?.formatWithVariant(variantManager.getVariantKey()) ?: ""
 
     private fun getDeviceFactor() = deviceInfo.formFactor().description
+
+    private suspend fun shouldFirePixel(
+        pixelName: String,
+        type: PixelType,
+    ): Boolean =
+        when (type) {
+            DEFAULT -> true
```
anikiki, line 196: I believe using `DEFAULT` doesn't provide much clarity for this type. Could we consider using terms like `ADHOC` or `ON_DEMAND` or something else that gives a more meaningful description?

statistics/src/main/java/com/duckduckgo/app/statistics/store/TimeProvider.kt
lmac012: I actually looked for existing time provider implementations before adding this one and noticed that there is `CurrentTimeProvider` in `common-utils`. I didn't use it, because it doesn't return timestamp in the conventional sense (a specific point on the timeline), but time since the system has booted, that can't be converted into a date.

While `Instant.now()` and `System.currentTimeMillis()` are based on the same clock, and the difference between those methods comes down to return type, `SystemClock.elapsedRealtime()` serves a different purpose.

I agree that we should have a standarized way obtaining time (that covers both wall clock and time since boot), and not have to create some kind of `TimeProvider` interface in every module, but I left it out of scope of this PR.

CrisBarreiro: Right, my bad, in my case all options were interchangeable since I wanted to compute the difference between 2 timestamps, but of course they serve different purposes

anrs/anrs-impl/src/main/java/com/duckduckgo/app/anr/AnrOfflinePixelSender.kt
lmac012: the `sendPixel()` method now returns `Single<SendPixelResult>` instead of `Completable`, but in this particular case, we don't care about `SendPixelResult` and just want to know when the stream completes successfully, so ignoring that element converts `Single` back to `Completable`.

statistics/src/main/java/com/duckduckgo/app/statistics/api/PixelSender.kt
lmac012: I don't feel strongly about this naming, in fact I also don't think `DEFAULT` is a particularly good name, but I was out of ideas. How about `COUNT` to match pixel naming convention that is already used in some modules? https://github.com/duckduckgo/Android/blob/dbae7fef6f47f31c7263e064469f2077675d308e/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/NetworkProtectionPixelNames.kt#L29 ?

statistics/src/main/java/com/duckduckgo/app/statistics/store/StatisticsDatabaseModule.kt
```diff
@@ -0,0 +1,51 @@
...
+@ContributesTo(AppScope::class)
+class StatisticsDatabaseModule {
+    @Provides
+    @SingleInstanceIn(AppScope::class)
+    fun provideStatisticsDatabase(context: Context): StatisticsDatabase =
+        Room
+            .databaseBuilder(
+                context = context,
+                klass = StatisticsDatabase::class.java,
+                name = "statistics.db",
```
anikiki, line 37: We need a different name for this DB, maybe `pixels.db` as we want to move away from "statistics".

statistics/src/main/java/com/duckduckgo/app/statistics/api/PixelSender.kt
anikiki: Yes, COUNT makes sense.

anikiki (approved):
LGTM :+1: Only a few comments about naming.

PR #4104, opened 2024-01-22
Optimize tracker evaluation

CrisBarreiro:
Task/Issue URL: https://app.asana.com/0/72649045549333/1206301192455389 

### Description
Optimize tracker evaluation by avoiding back and forth conversion of String to Uri and unneeded precautions for edge cases. 

Note: The methods I optimized are used in several places, where we can't be sure the URL is correctly formed (or can only be sure for some parameters but not others), and therefore we need to keep the old implementation as well as create several variants to make sure nothing breaks while optimizing whenever possible. Further optimizations might be possible but they would require a larger investment

A follow-up PR will setup an A/B test

### Steps to test this PR

_Feature 1_
- [ ] Smoke test

### UI changes
No UI changes

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
@@ -52,6 +52,7 @@ import com.duckduckgo.app.browser.navigation.safeCopyBackForwardList
 import com.duckduckgo.app.browser.pageloadpixel.PageLoadedHandler
 import com.duckduckgo.app.browser.print.PrintInjector
 import com.duckduckgo.app.global.model.Site
+import com.duckduckgo.app.pixels.remoteconfig.OptimizeTrackerEvaluationRCWrapper
 import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.autoconsent.api.Autoconsent
 import com.duckduckgo.autofill.api.BrowserAutofill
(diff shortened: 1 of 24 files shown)


CrisBarreiro:
* `develop`
  * **#4104** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4104?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

 Join @CrisBarreiro and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

aitorvs (approved):
This looks OK to me. Just to make sure I understand, after we've verified this, do we plan to delete all the duplicated methods?

CrisBarreiro:
> This looks OK to me. Just to make sure I understand, after we've verified this, do we plan to delete all the duplicated methods?

Not all of them. For some methods, I created a new duplicate implementation which will go away once we've verified this. However, in some other cases, I had to create more than one new implementation (see [fun evaluate](https://github.com/duckduckgo/Android/blob/94cdbcd705f7c78b7c7860cb4a50a01ef7322299/app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDetector.kt#L52) or [fun matches](https://github.com/duckduckgo/Android/blob/94cdbcd705f7c78b7c7860cb4a50a01ef7322299/app/src/main/java/com/duckduckgo/app/trackerdetection/Client.kt#L54)). This is because I've only optimized URLs coming from the webview and request themselves where I could be sure they were well-formed in a safe amount of time, and also where I thought I could quickly verify my hypothesis. Ideally, we should be able to expand this to URLs coming from cloacked host detection, which would allow us to get rid of all the duplicate implementations, but that's something that would need to be scoped in a new project (likely https://app.asana.com/0/715106103902962/1206211280846299)

PR #4123, opened 2024-01-26
Fix re-key

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206444028232342/f

### Description
Re-key was not working as expected because when we restart the tunnel we always re-register.

Note: This PR also reduces the amount of models to represent the WG configuration to basically one.

There are now two interfaces related to the tunnel, ie `WgTunnel` and `WgTunnelConfig`
* `WgTunnel` exposes APIs to create the tunnel config and (optionally) save it
* `WgTunnelConfig` allows to read the saved tunnel config and also clear it.

This is mainly done because the UI (browser process) needs read access to tunnel config, while the VPN process is the one that needs access to create the tunnel.

### Steps to test this PR

_1_
- [x] enable VPN
- [x] smoke tests VPN and AppTP in their different configurations
- [x] verify VPN management screen displays right information
- [x] verify geo-switching works as expected
- [x] verify 'netp dev settings' display right information
- [x] verify 'exclusion list' feature works as expected
- [x] verify 'exclude local networks' setting works as expected

_2_
- [x] Filter logcat by `Wireguard configuration`
- [x] enable only the VPN (AppTP disabled) and note the configuration (config 1)
- [x] force close the app and re-open
- [x] verify the configuration (config 2) is identical to the previous one
- [x] disable the VPN and re-enable
- [x] verify the configuration (config 3) is different than previous ones above (config 1 and 2)
- [x] enable both VPN and AppTP and take note of the configuration (config 4)
- [x] snooze the VPN and reconnect
- [x] verify the configuration (config 4) is different than config 3 above
- [ ] go to NetP dev settings and force rekey
- [ ] verify the configuration (config 5) is different than config 4 above (you may need to force rekey 1+ times)

_3_
- [x] Filter logcat by `Wireguard configuration`
- [x] enable only VPN and take note of the configuration (config 1)
- [x] go to NetP dev settings
- [x] tap on `Exclude all System apps`
- [ ] verify the configuration (config 2) is the same as config 1 above (rinse and repeat 1+ times)
- [ ] Enable AppTP
- [ ] Repeat from step 3 above

_4_
- [x] Enable VPN
- [x] Change location using geo-switching feature
- [x] verify VPN re-connects to different server

_5_
- [x] Install app from develop
- [x] Enable VPN
- [x] Build this PR branch, update the app and re-launch
- [x] Verify VPN is re-enabled automatically
- [x] Verify VPN works as expected
- [x] Repeat from step 1 and in step 2 enable both AppTP and VPN


aitorvs:
* `develop`
  * **#4123** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4123?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

network-protection/network-protection-impl/src/main/java/com/wireguard/config/InetAddresses.kt
```diff
@@ -37,8 +37,8 @@ object InetAddresses {
             |2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)
             (\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d
             |1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d
-            |[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?)|((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9]
-            [0-9]?))${'$'}
+            |[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))
```
karlenDimla, line 40: @aitorvs why a new line?

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/WgVpnNetworkStack.kt
```diff
@@ -53,56 +49,44 @@ class WgVpnNetworkStack @Inject constructor(
...
-                )
-                clientInterface = ClientInterface(
-                    tunnelCidrSet = wgTunnelData!!.tunnelAddress.toCidrString(),
-                )
-            }
+            wgConfig = networkProtectionRepository.get().wireguardConfig ?: fetchWireguardConfigAndCacheIt()
+            logcat { "Wireguard configuration:\n$wgConfig" }
+
+            // refresh WG config
+            networkProtectionRepository.get().wireguardConfig = wgConfig
```
karlenDimla, line 77: NIT:  we only do this in fetchWireguardConfigAndCacheIt()? Seems unnecessary to do all the time.

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/WgVpnNetworkStack.kt
```diff
@@ -132,17 +116,18 @@ class WgVpnNetworkStack @Inject constructor(
...
         }
 
         val result = wgProtocol.get().startWg(
             tunfd,
-            wgTunnelData!!.userSpaceConfig.also {
+            wgConfig!!.toWgUserspaceString().also {
                 logcat { "WgUserspace config: $it" }
             },
-            pcapConfig = netPDefaultConfigProvider.pcapConfig(),
+            // pcapConfig = netPDefaultConfigProvider.pcapConfig(),
```
karlenDimla, line 129: @aitorvs  do we not support this anymore?

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/WgVpnNetworkStack.kt
aitorvs: Ah good call, this actually would allow me to move the `lastPrivateKeyUpdateTimeInMillis` inside the `Config`. Which is something I was thinking about doing instead of having to maintain it separately

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgTunnel.kt
aitorvs, line 1: I don't really like having `WgTunnel` and `WgTunnelConfig` but the issue is that we need access from both the app scope and the vpn scope

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgServerApi.kt
```diff
@@ -88,7 +87,7 @@ class RealWgServerApi @Inject constructor(
         serverName = server.name,
         publicKey = server.publicKey,
         publicEndpoint = server.extractPublicEndpoint(),
-        address = allowedIPs.joinToString(","),
+        address = allowedIPs.first(),
```
karlenDimla, line 90: @aitorvs will there never be a scenario that we will have more than 1 here?
If so, maybe better to pass the list?

aitorvs: Tbh I am not sure what happens if we add more than one address to the `tun`. In principle it should just work but we'd need to test before.
Added the `first()` to avoid issues in the event that BE (wrongly) sends more than one address here.

This is actually a bit more convoluted because of why `allowsIPs` means `tun` address for the client. We can discuss offline

karlenDimla (approved):
LGTM and works as expected

PR #4126, opened 2024-01-29
Amend recover copy

marcosholgado:
Task/Issue URL: https://app.asana.com/0/0/1206454043535976/f 

### Description
Amend copy

### Steps to test this PR
N/A

PR #4121, opened 2024-01-25
Add itr screen

marcosholgado:
Task/Issue URL: https://app.asana.com/0/0/1206414641255151/f 

### Description
See tasl

### Steps to test this PR
See task

PR #4117, opened 2024-01-25
Add Pir screen

marcosholgado:
Task/Issue URL: https://app.asana.com/0/0/1206414641255152/f 

### Description
See task

### Steps to test this PR
See task

PR #4127, opened 2024-01-29
Check for etld in allowed domains

marcosholgado:
Task/Issue URL: https://app.asana.com/0/0/1206093040131785/f 

### Description
See task

### Steps to test this PR
See task

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/messaging/ItrMessagingInterface.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/messaging/ItrMessagingInterface.kt
@@ -20,6 +20,7 @@ import android.webkit.JavascriptInterface
 import android.webkit.WebView
 import androidx.core.net.toUri
 import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.common.utils.extensions.toTldPlusOne
 import com.duckduckgo.di.scopes.ActivityScope
 import com.duckduckgo.js.messaging.api.JsCallbackData
 import com.duckduckgo.js.messaging.api.JsMessage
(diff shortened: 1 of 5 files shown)


marcosholgado:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4127?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `develop`
  * **#4117** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4117?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **#4121** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4121?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **#4126** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **#4127** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4127?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

 Join @marcosholgado and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/messaging/ItrMessagingInterface.kt
```diff
@@ -61,11 +61,11 @@ class ItrMessagingInterface @Inject constructor(
         try {
             val adapter = moshi.adapter(JsMessage::class.java)
             val jsMessage = adapter.fromJson(message)
-            val domain = runBlocking(dispatcherProvider.main()) {
-                webView.url?.toUri()?.host
+            val url = runBlocking(dispatcherProvider.main()) {
```
CrisBarreiro, line 64: Since we're using the URI from the webview, I'd suggest keeping the conversion to URI and not passing strings along the way. That way, `String.toTldPlusOne()` doesn't need to rely on `UriString.host(this)`, which is somewhat expensive, and could just do this.host instead. Probably not super relevant in this case, but I'm trying to remove String URIs when possible as part of the performance initiative, as the method I mentioned before introduces a significant toll when called a lot of times

marcosholgado: Makes sense. Changed to use the url string directly.

marcosholgado: FYI I'm going back to this here because the host is needed to calculate the eTLD+1 otherwise the result includes the path too. What's not needed is the reparsing to URI inside the toTldPlusOne method

PR #4139, opened 2024-01-31
Implement failure recovery

karlenDimla:
Task/Issue URL: https://app.asana.com/0/0/1206063664298512/f 

### Description
See attached description

### Steps to test this PR
https://app.asana.com/0/0/1206483769461553/f

### DIFF
--- a/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgServerApi.kt
+++ b/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgServerApi.kt
@@ -19,7 +19,6 @@ package com.duckduckgo.networkprotection.impl.configuration
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.di.scopes.VpnScope
 import com.duckduckgo.networkprotection.impl.configuration.WgServerApi.Mode
-import com.duckduckgo.networkprotection.impl.configuration.WgServerApi.Mode.Default
 import com.duckduckgo.networkprotection.impl.configuration.WgServerApi.Mode.FailureRecovery
 import com.duckduckgo.networkprotection.impl.configuration.WgServerApi.WgServerData
 import com.duckduckgo.networkprotection.impl.di.UnprotectedVpnControllerService
@@ -40,12 +39,15 @@ interface WgServerApi {
 
     suspend fun registerPublicKey(
         publicKey: String,
-        mode: Mode = Default,
+        mode: Mode? = null,
     ): WgServerData?
 
     sealed class Mode {
-        data object Default : Mode()
-        data class FailureRecovery(val currentServer: String) : Mode()
+        data class FailureRecovery(val currentServer: String) : Mode() {
+            override fun toString(): String {
+                return "failureRecovery"
+            }
+        }
     }
 }
 
@@ -58,7 +60,7 @@ class RealWgServerApi @Inject constructor(
 
     override suspend fun registerPublicKey(
         publicKey: String,
-        mode: Mode,
+        mode: Mode?,
     ): WgServerData? {
         // This bit of code gets all possible egress servers which should be order by proximity, caches them for internal builds and then
         // returns the closest one or null if list is empty
@@ -76,7 +78,7 @@ class RealWgServerApi @Inject constructor(
 
         val userPreferredLocation = netNetpEgressServersProvider.updateServerLocationsAndReturnPreferred()
         val registerKeyBody = if (mode is FailureRecovery) {
-            RegisterKeyBody(publicKey = publicKey, server = mode.currentServer, mode = MODE_FAILURE)
+            RegisterKeyBody(publicKey = publicKey, server = mode.currentServer, mode = mode.toString())
         } else if (selectedServer != null) {
             RegisterKeyBody(publicKey = publicKey, server = selectedServer)
         } else if (userPreferredLocation != null) {
@@ -136,10 +138,6 @@ class RealWgServerApi @Inject constructor(
 
         val city: String? by attributes
         val country: String? by attributes
-    }
-
-    companion object {
-        private const val MODE_FAILURE = "failureRecovery"
     }
 }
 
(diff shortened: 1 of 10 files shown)


karlenDimla:
* `develop`
  * **#4139** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4139?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

 Join @karlenDimla and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgServerApi.kt
```diff
@@ -35,7 +38,15 @@ interface WgServerApi {
...
 
-    suspend fun registerPublicKey(publicKey: String): WgServerData?
+    suspend fun registerPublicKey(
+        publicKey: String,
+        mode: Mode = Default,
+    ): WgServerData?
+
+    sealed class Mode {
+        data object Default : Mode()
+        data class FailureRecovery(val currentServer: String) : Mode()
```
aitorvs, line 48: Why not overload the `toString` method (like below) so that we don't need to do the mapping between the model and the `MODE_FAILURE` constant and we just do `toString()` on it?

```suggestion
        data class FailureRecovery(val currentServer: String) : Mode() {
            override fun toString(): String {
                return "failureRecovery"
            }
        }
```

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgServerApi.kt
```diff
@@ -117,6 +137,10 @@ class RealWgServerApi @Inject constructor(
         val city: String? by attributes
         val country: String? by attributes
     }
+
+    companion object {
+        private const val MODE_FAILURE = "failureRecovery"
+    }
```
aitorvs, line 143: Ditto above, I think we can remove

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgServerApi.kt
```diff
@@ -35,7 +38,15 @@ interface WgServerApi {
         val gateway: String,
     )
 
-    suspend fun registerPublicKey(publicKey: String): WgServerData?
+    suspend fun registerPublicKey(
+        publicKey: String,
+        mode: Mode = Default,
```
aitorvs, line 43: I think `mode` should be nullable rather than forcing a value on it and then have a `Default` mode that is (and will never) be used.
```suggestion
        mode: Mode? = null,
```

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgServerApi.kt
```diff
@@ -61,11 +75,17 @@ class RealWgServerApi @Inject constructor(
             }
 
         val userPreferredLocation = netNetpEgressServersProvider.updateServerLocationsAndReturnPreferred()
-        val registerKeyBody = if (selectedServer != null) {
+        val registerKeyBody = if (mode is FailureRecovery) {
+            RegisterKeyBody(publicKey = publicKey, server = mode.currentServer, mode = MODE_FAILURE)
+        } else if (selectedServer != null) {
```
aitorvs, line 80: Ditto, I thin we should make `mode` nullable and edit this logic accordingly

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/pixels/WireguardHandshakeMonitor.kt
aitorvs, line 1: I don't think the failure recovery logic should be inside the handshake monitor. The monitor was very simple, now it's kind of not.

I'd rather create a `WireguardHandshakeMonitor.Listener` class like eg. 

```kotlin
class Listener {
  fun onTunnelFailureDetected(lastHandshakeEpocSeconds: Long)
}
```

And decouple whatever logic out of the monitor

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgVpnControllerService.kt
```diff
@@ -126,7 +126,7 @@ interface WgVpnControllerService {
     suspend fun getEligibleLocations(): List<EligibleLocation>
 }
 
-const val NETP_ENVIRONMENT_URL = "https://controller.netp.duckduckgo.com"
+const val NETP_ENVIRONMENT_URL = "https://staging1.netp.duckduckgo.com"
```
aitorvs, line 129: Revert this change. Also we can set the environment from internal dev settings.

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/configuration/WgTunnel.kt
```diff
@@ -50,15 +52,15 @@ interface WgTunnel {
...
+    suspend fun createWgConfig(keyPair: KeyPair? = null, fromFailure: Boolean = false): Result<Config>
 
     /**
      * Creates a new wireguard [Config], returns it and stores it internally.
      * If one exists, updates it and returns it and also stores it internally.
      *
      * @param keyPair is the private/public key [KeyPair] to be used in the wireguard [Config]
      */
-    suspend fun createAndSetWgConfig(keyPair: KeyPair? = null): Result<Config>
+    suspend fun createAndSetWgConfig(keyPair: KeyPair? = null, fromFailure: Boolean = false): Result<Config>
```
aitorvs, line 63: Tbf, I don't like this `fromFailure` boolean value, In general I don't like flag-arguments. There's plenty of literature around that, one post I like about the topic is [this one](https://martinfowler.com/bliki/FlagArgument.html).

I have other proposals, let's discuss 1on1

aitorvs (approved):
Tested and works as expected

PR #4160, opened 2024-02-06
Add test to verify bucket assignment

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206514199358858/f

### Description
Integration test to test incremental rollout bucket allocation

### Steps to test this PR
Test shall pass

### DIFF
--- a/feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/codegen/ContributesRemoteFeatureCodeGeneratorBucketAssignmentTest.kt
+++ b/feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/codegen/ContributesRemoteFeatureCodeGeneratorBucketAssignmentTest.kt
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2023 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/codegen/ContributesRemoteFeatureCodeGeneratorBucketAssignmentTest.kt
```diff
@@ -0,0 +1,143 @@
...
+            flavorNameProvider = { appBuildConfig.flavor.name },
+            appVariantProvider = { variantManager.getVariantKey() },
+            forceDefaultVariant = { variantManager.saveVariants(emptyList()) },
+        ).build().create(TestTriggerFeature::class.java)
+    }
+
+    @Test
+    fun `test probability`() {
+        fun isWithinRange(): Boolean {
+            if (testCase.run < 9000) return true
```
CrisBarreiro, line 67: Why do we need this?

aitorvs: I'll add a comment in the test, the reason is that when the sample size is small, we can't ensure the bucket distribution is within the right error bands. So I am basically ignoring sample sizes below 9000

marcosholgado (approved):
LGTM

PR #4175, opened 2024-02-10
Ensure we re-evaluate newly added feature flags

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206560594088721/f

### Description
We have an edge case where if a feature flag is present in the remote config before it's handled in the app, it will not be re-evaluated unless its `hash` changes.
This causes an issue in cases like the following example:
* release app version 1, that doesn't have any `testFlag` feature flag
* release to remote config a feature flag called `testFlag` rolled out to 99%
* app version 1 downloads the config that contains `testFlag` and ignores it as it doesn't know about it
* release app version 2 that contains and handles `testFlag`
* remote config is re-downloaded
* `testFlag` continues to be ignored because nothing has changed in it and its feature `hash` is the same
* to force a re-evaluation we'll need to eg. bump the rollout step

To overcome this, this PR adds a `PrivacyFeaturePlugin#hash()` method. This method should return the `hash` of privacy feature and the hash should change when eg. new sub-features (aka feature flags) are added.
The hash is used in two places:
* in the plugin `signature()`, called from the `persistPrivacyConfig` to decide whether the remote config should be persisted or not
* in the `PrivacyFeaturePlugin#store()` method (auto-generated for the feature flags). Previously the `store` method would only consider the remote config feature `hash`, but now combines both the remote feature `hash` and the own's feature `hash`. This way the feature is re-evaluated when either it changes remotely or locally 

Note: [Incidentally](https://github.com/duckduckgo/Android/pull/4175/commits/1c2f954b48c5aa5acc369acc910faf8b7668f784) this PR also removes the `rollout.steps` sorting that prohibited to reduce the rollout %. By removing it we allow to re reduce the rollout % for new users. Something that is desirable.

### Steps to test this PR
This PR added 2e2 tests that verify the functionality but this is also the equivalent manual test

_Test_
- [x] install from this branch and launch the app
- [x] go to app settings -> developer settings -> override privacy remote config URL
- [x] toggle "use custom URL" on and enter `https://jsonblob.com/api/jsonBlob/1126094951384629248`
- [x] tap on force reloading
- [x] Use AS (or Flipper) to see the contents of the share prefs `com.duckduckgo.feature.toggle.androidBrowserConfig`
- [x] take note of the `hash`
- [x] Apply the following patch and rebuild / re-install / re-launch the app
```kotlin
Index: app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
IDEA additional info:
<+>UTF-8
diff --git a/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt b/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
--- a/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt	(revision 5fe8bd5216708fbbe945832377600bc90233eb24)
+++ b/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt	(date 1707729280159)
@@ -59,4 +59,7 @@
      */
     @Toggle.DefaultValue(false)
     fun optimizeTrackerEvaluation(): Toggle
+
+    fun testFlag(): Toggle
 }
```
- [x] verify the `androidBrowserConfig_testFlag` is now present in the `com.duckduckgo.feature.toggle.androidBrowserConfig` and its `enable` value is most likely `true` (`rolloutThreshold` < `98`)
- [x] Repeating this test in develop should fail the last step

### DIFF
--- a/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
+++ b/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
@@ -307,11 +307,11 @@ internal class ToggleImpl constructor(
             } ?: state
         }
 
-        val sortedRollout = state.rollout.sorted().filter { it in 0.0..100.0 }
-        if (sortedRollout.isEmpty()) return state
+        val scopedRolloutRange = state.rollout.filter { it in 0.0..100.0 }
+        if (scopedRolloutRange.isEmpty()) return state
 
         return state.copy(
-            enable = (state.rolloutThreshold ?: 0.0) <= sortedRollout.last(),
+            enable = (state.rolloutThreshold ?: 0.0) <= scopedRolloutRange.last(),
         )
     }
 }
--- a/feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/codegen/ContributesRemoteFeatureCodeGeneratorTest.kt
+++ b/feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/codegen/ContributesRemoteFeatureCodeGeneratorTest.kt
@@ -48,6 +48,7 @@ class ContributesRemoteFeatureCodeGeneratorTest {
 
     private val context: Context = InstrumentationRegistry.getInstrumentation().targetContext.applicationContext
     private lateinit var testFeature: TestTriggerFeature
+    private lateinit var anotherTestFeature: AnotherTestTriggerFeature
     private val appBuildConfig: AppBuildConfig = mock()
     private lateinit var variantManager: FakeVariantManager
     private lateinit var toggleStore: FakeToggleStore
@@ -65,6 +66,14 @@ class ContributesRemoteFeatureCodeGeneratorTest {
             appVariantProvider = { variantManager.getVariantKey() },
             forceDefaultVariant = { variantManager.saveVariants(emptyList()) },
         ).build().create(TestTriggerFeature::class.java)
+        anotherTestFeature = FeatureToggles.Builder(
+            toggleStore,
+            featureName = "testFeature",
+            appVersionProvider = { appBuildConfig.versionCode },
+            flavorNameProvider = { appBuildConfig.flavor.name },
+            appVariantProvider = { variantManager.getVariantKey() },
+            forceDefaultVariant = { variantManager.saveVariants(emptyList()) },
+        ).build().create(AnotherTestTriggerFeature::class.java)
     }
 
     @Test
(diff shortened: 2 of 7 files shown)


aitorvs:
* `develop`
  * **#4175** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4175?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -307,11 +307,11 @@ internal class ToggleImpl constructor(
             } ?: state
         }
 
-        val sortedRollout = state.rollout.sorted().filter { it in 0.0..100.0 }
-        if (sortedRollout.isEmpty()) return state
+        val scopedRolloutRange = state.rollout.filter { it in 0.0..100.0 }
```
aitorvs, line 310: By not sorting we allow to reduce the rollout % for new users, ie. users that have the FF disabled when it's re-evaluated.

feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/codegen/ContributesRemoteFeatureCodeGeneratorTest.kt
```diff
@@ -115,6 +124,227 @@ class ContributesRemoteFeatureCodeGeneratorTest {
...
+                        "fooFeature": {
+                            "state": "enabled"
+                        }
+                    }
+                }
+                """.trimIndent(),
+            ),
+        )
+
+        assertTrue(anotherTestFeature.newFooFeature().isEnabled())
```
CrisBarreiro, line 345: Probably irrelevant in the context of this specific change, but isn't `anotherTestFeature.newFooFeature().isEnabled()` supposed to be false when `testFeature.self().isEnabled()` is false, regardless of the actual remote config value?

aitorvs: The FF framework doesn't condition sub-feature `state` values to the root feature `state` value. This is left to the developer to decide if both `state` values should be affect each other or not

CrisBarreiro:
@aitorvs just a small check on the last step

_"verify the `androidBrowserConfig_testFlag` is not present in the `com.duckduckgo.feature.toggle.androidBrowserConfig` and its enable value is most likely true (`rolloutThreshold < 98`)"_

On this branch, `androidBrowserConfig_testFlag` _is_ present and its value is indeed true
On develop, `androidBrowserConfig_testFlag` _is not_ present

I assume that's the desired behavior

aitorvs:
> @aitorvs just a small check on the last step
> 
> _"verify the `androidBrowserConfig_testFlag` is not present in the `com.duckduckgo.feature.toggle.androidBrowserConfig` and its enable value is most likely true (`rolloutThreshold < 98`)"_
> 
oh dear, `not` was meant to be a `now`. So it being present is the expected behavior (fixed the typo too)

> On this branch, `androidBrowserConfig_testFlag` _is_ present and its value is indeed true On develop, `androidBrowserConfig_testFlag` _is not_ present
> 
> I assume that's the desired behavior. If that's the case, all looks good to me

Yeah, on develop it should not be present as it's not re-evaluated.

karlenDimla (approved):

PR #4212, opened 2024-02-26
Detect encryption error state that could be recoverable

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206692157281177/f 

### Description
Attempts to detect when user is in a state whereby the encrypted data is not accessible and tests whether they could access encrypted data on a new test file.
- if they can access encrypted data on a new test file, this is indicative that their local encrypted shared preferences is permanently undecryptable, and we should consider if we need a (possibly user-facing) mitigation to let them wipe it and re-initialise
- if they can't access encrytped data on a new test file, it's probable their device just can't support it

Also brings in the latest version of the AndroidX security library, as there are some related changes to the exception that is thrown in this scenario (including changing the thrown exception type and internally fixing a related race condition that could throw the exception)

### Steps to test this PR

- [ ] Install from `develop` first and add some passwords
- [ ] Install this branch. verify passwords are still accessible
- [ ] Ideally if you can, trigger the scenario whereby the device is backed up and restored, and its encrypted shared preferences become inaccessible. Alternatively, I think the next best thing here would be to manually edit the `com.duckduckgo.app.email.settings.xml` shared preferences file so that it was undecryptable. (perhaps useful to pair on this testing step)
- [ ] Launch the app and verify you see `m_unable_to_decrypt_secure_email_data` pixel

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/email/db/EmailEncryptedSharedPreferences.kt.kt
+++ b/app/src/main/java/com/duckduckgo/app/email/db/EmailEncryptedSharedPreferences.kt.kt
@@ -25,6 +25,7 @@ import com.duckduckgo.app.pixels.AppPixelName
 import com.duckduckgo.app.statistics.pixels.Pixel
 import java.io.IOException
 import java.security.GeneralSecurityException
+import javax.crypto.AEADBadTagException
 
 class EmailEncryptedSharedPreferences(
     private val context: Context,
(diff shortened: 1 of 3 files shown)


CDRussell:
* **#4212** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4212?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

PR #4213, opened 2024-02-27
Automatically logout invalid sync devices

not-a-rootkit:
To prevent phantom devices being allowed to sync a user's data, we automatically logout devices that cannot be decrypted, instead of just ignoring them.

Task/Issue URL:  https://app.asana.com/0/0/1206693867033120/f

### Description
It is possible for someone to sync a device to another user's account such that the device data cannot be decrypted. This results in an invalid "phantom" device that can sync data from a user's account without their knowledge (and they are unable to view or remove said device from the UI).

As a result, we introduce a fix that automatically logs out invalid devices to prevent this issue from occurring. In the very rare case that a device accidentally contains invalid encrypted data, the rightful owner of that device should be able to just re-sync it.

### Steps to test this PR
Note: requires two devices and some form of request interception (i.e. BurpSuite). There is an integration test that covers this test case too.
1. Setup sync on device A
2. Sync device B to device A
3. Turn on request interception in BurpSuite
4. Rename device B on device B to any arbitrary string
5. In Burp Interceptor, replace the encrypted device name with `G9rm2IHsRj9cNg==` (random data)
6. Forward the request to the backend
7. On device A, go to settings>sync & backup
8. Ensure only one device is shown
9. On device B, go to settings>sync & backup
10. Check that the device has been logged out

_Feature 1_
- [ ] Automatically logout invalid devices
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
@@ -307,7 +307,9 @@ class AppSyncAccountRepository @Inject constructor(
                             )
                         } catch (throwable: Throwable) {
                             throwable.asErrorResult().alsoFireAccountErrorPixel()
-                            logout(device.deviceId)
+                            if (syncStore.deviceId != device.deviceId) {
+                                logout(device.deviceId)
+                            }
                             null
                         }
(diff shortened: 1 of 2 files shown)


cmonfortep (approved):
LGTM, thanks for also adding a test case :)

Added a comment about ensuring we logout a remote device, not the current one.

PR #4259, opened 2024-03-07
support deeplink navigation

cmonfortep:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206775412000490/f

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


cmonfortep:
* **#4259** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @cmonfortep and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

anvil/anvil-compiler/src/main/java/com/duckduckgo/anvil/compiler/ContributeToActivityStarterCodeGenerator.kt
```diff
@@ -111,6 +122,39 @@ class ContributeToActivityStarterCodeGenerator : CodeGenerator {
...
+                                if (deeplinkActivityParams.jsonArguments.isEmpty()) {
+                                    val instance = tryCreateObjectInstance(%T::class.java)
+                                    if (instance != null) {
+                                        return instance
+                                    }
+                                }
+                                tryCreateActivityParams(%T::class.java, deeplinkActivityParams)
+                            } else {
+                                null
+                            }
```
aitorvs, line 148: When the screen doesn't have a name (ie. empty in `ContributeToActivityStarter`. This code generates something like
```kotlin
val screenName = ...
return if (screenName == "") {...} else {...}
```
The method eventually returns `null` because `val screenName = ...` returns null right in there, but still the following if statement is wrong.
I'd change to something like

```suggestion
                            val screenName = deeplinkActivityParams.screenName.takeUnless { it.isEmpty() } ?: return null
                            val definedScreenName = %S
                            if (definedScreenName.isNullOrBlank() return null
                            
                            return if (screenName == definedScreenName) {
                                if (deeplinkActivityParams.jsonArguments.isEmpty()) {
                                    val instance = tryCreateObjectInstance(%T::class.java)
                                    if (instance != null) {
                                        return instance
                                    }
                                }
                                tryCreateActivityParams(%T::class.java, deeplinkActivityParams)
                            } else {
                                null
                            }
```

anvil/anvil-compiler/src/main/java/com/duckduckgo/anvil/compiler/ContributeToActivityStarterCodeGenerator.kt
```diff
@@ -68,14 +68,20 @@ class ContributeToActivityStarterCodeGenerator : CodeGenerator {
         val content = FileSpec.buildFile(generatedPackage, moduleClassName) {
             for (annotation in mapperAnnotations) {
                 val paramsType = annotation.paramsTypeOrNull()!!
-                addType(createMapperClass(paramsType, vmClass, module)).build()
+                val screenName = annotation.screenNameOrNull().orEmpty()
+                addType(createMapperClass(paramsType, screenName, vmClass, module)).build()
```
aitorvs, line 72: Skip mapper generation if the screenName is null or blank
```suggestion
                if (screenName.isNullOrBlank() == false) {
                    addType(createMapperClass(paramsType, screenName, vmClass, module)).build()
                }
```

anvil/anvil-compiler/src/main/java/com/duckduckgo/anvil/compiler/ContributeToActivityStarterCodeGenerator.kt
```diff
@@ -111,6 +122,39 @@ class ContributeToActivityStarterCodeGenerator : CodeGenerator {
...
+                        "deeplinkActivityParams",
+                        deeplinkActivityParamsFqName.asClassName(module),
+                    )
+                    .returns(
+                        activityParamsFqName.asClassName(module).copy(nullable = true),
+                    )
+                    .addCode(
+                        """
+                            val screenName = deeplinkActivityParams.screenName.takeUnless { it.isEmpty() } ?: return null
+                            return if (screenName == %S) {
```
aitorvs, line 138: When screen name is not defined in `ContributeToActivityStarter`. This code generates something like
```kotlin
val screenName = ...
return if (screenName == "") {...} else {...}
```
The method eventually returns `null` because `val screenName = ...` returns null right in there, but still the following if statement is wrong.

I added a previous comment to skip the mapper generation when screen name is not defined in `ContributeToActivityStarter`, but anyways we should also guard against that here to make sure changes in codegen are independent.

```suggestion
                            val definedScreenName = %S
                            if (definedScreenName.isNullOrBlank()) return null
                            return if (screenName == definedScreenName) {
```

anvil/anvil-compiler/src/main/java/com/duckduckgo/anvil/compiler/ContributeToActivityStarterCodeGenerator.kt
```diff
@@ -111,6 +122,39 @@ class ContributeToActivityStarterCodeGenerator : CodeGenerator {
                     )
                     .build(),
             )
+            .addFunction(
```
aitorvs, line 125: I think we should skip adding this function if the screenName was blank

```kotlin
.apply {
  if (screenName.isNotBlank()) {
    addFunction(...)
  }
}

aitorvs (approved):
LGTM, left couple comments I think we should do, ie
* generate empty mappers when the screeName is not defined in `ContributeToActivityStarter`
* safeguard empty screenName in the mappers when generated

anvil/anvil-compiler/src/main/java/com/duckduckgo/anvil/compiler/ContributeToActivityStarterCodeGenerator.kt
cmonfortep: Oh that's cool indeed! So depending on `screenName` value, the implementation can be different, and use that to just provide an empty mapper if no `screenName` provided. Makes sense.

PR #4341, opened 2024-03-21
Add scoring algorithm for history suggestions (extracted from BSK)

CrisBarreiro:
Task/Issue URL: https://app.asana.com/0/0/1206816228247419/f 

### Description
Copy the scoring algorithm from BSK

### Steps to test this PR
_Bookmarks shouldn't show in top hits (top section, above DDG suggestions)_
- [ ] Add wikipedia.org as a bookmark (not favorite)
- [ ] Type wiki on the search bar
- [ ] Check the newly added bookmark is not shown in top hits, and is instead shown in the bottom section (below DDG suggestions)

_Favorites should show in top hits_
- [ ] Add wikipedia.org as a favorite
- [ ] Type wiki on the search bar
- [ ] Check the newly added favorite is shown in top hits and uses the bookmark+favorite icon

### Steps to test this PR

_Test pixel fired when selecting a favorite_
- [ ] Select a favorite from autocomplete
- [ ] Check m_autocomplete_favorite_selection has been fired

_Test pixel fired when selecting a bookmark_
- [ ] Select a bookmark from autocomplete
- [ ] Checkk m_aut_s_b has been fired

_Website (not SERP)_
- [ ] Visit a page (not SERP)
- [ ] Open the history database using inspector or flipper, and check there's a new entry in `history_entries`. Take note of the ID. Check the title field contains the same title as the tab, the query field is empty and `isSerp` is 0
- [ ] Check there's an entry in `visits_list` with that same `historyEntryId`
- [ ] Reload the page
- [ ] Check the `history_entries` table still has only one entry for that site
- [ ] Check the `visits_list` contains 2 entries for that same `historyEntryId`

_SERP_
- [ ] Perform a search
- [ ] Open the history database using inspector or flipper, and check there's a new entry in `history_entries`. Take note of the ID. Check the title field contains the same title as the tab, the query field contains your query, and `isSerp` is 0
- [ ] Check there's an entry in `visits_list` with that same `historyEntryId`
- [ ] Reload the page
- [ ] Check the `history_entries` table still has only one entry for that site (In some cases, reload doesn't produce the same exact query params,* in case a second entry is added, check URLs are different)
- [ ] Check the `visits_list` contains 2 entries for that same `historyEntryId`

_DuckDuckGo URL (not SERP)_
- [ ] Visit https://duckduckgo.com/settings
- [ ] Open the history database using inspector or flipper, and check the query field is empty and `isSerp` is 0

**Clear app data between tests, fire button doesn't clear history
(yet)**

[...]

PR #4362, opened 2024-03-27
Store history

CrisBarreiro:
Task/Issue URL: https://app.asana.com/0/1205008441501016/1206816228247418/f 

### Steps to test this PR

_Website (not SERP)_
- [ ] Visit a page (not SERP)
- [ ] Open the history database using inspector or flipper, and check there's a new entry in `history_entries`. Take note of the ID. Check the title field contains the same title as the tab, the query field is empty and `isSerp` is 0
- [ ] Check there's an entry in `visits_list` with that same `historyEntryId`
- [ ] Reload the page
- [ ] Check the `history_entries` table still has only one entry for that site
- [ ] Check the `visits_list` contains 2 entries for that same `historyEntryId`

_SERP_
- [ ] Perform a search
- [ ] Open the history database using inspector or flipper, and check there's a new entry in `history_entries`. Take note of the ID. Check the title field contains the same title as the tab, the query field contains your query, and `isSerp` is 0
- [ ] Check there's an entry in `visits_list` with that same `historyEntryId`
- [ ] Reload the page
- [ ] Check the `history_entries` table still has only one entry for that site (In some cases, reload doesn't produce the same exact query params,* in case a second entry is added, check URLs are different)
- [ ] Check the `visits_list` contains 2 entries for that same `historyEntryId`

_DuckDuckGo URL (not SERP)_
- [ ] Visit https://duckduckgo.com/settings
- [ ] Open the history database using inspector or flipper, and check the query field is empty and `isSerp` is 0

**Clear app data between tests, fire button doesn't clear history
(yet)**

_Feature 1_
- [ ] Visit wikipedia.org
- [ ] Type wiki
- [ ] Check a history result is shown for wikipedia.org

_Feature 2_
- [ ] Visit https://en.wikipedia.org/wiki/Cat
- [ ] Type wiki
- [ ] Check there's no top hits result for Cat at wikipedia
- [ ] Visit https://en.wikipedia.org/wiki/Cat 3 more times
- [ ] Check there's a top hits result for Cat at wikipedia

_Feature 3_
- [ ] Add a random site as a bookmark
- [ ] Then go to bookmarks and edit it, Set https://wikipedia.org as the
URL and Wikipedia as the title
- [ ] Type wiki
- [ ] Check wikipedia is shown on the bottom section with a bookmark
icon, not in top hits
- [ ] Visit wikipedia.org
- [ ] Check now wikipedia is shown in top hits with the bookmark icon

_Feature 4_
- [ ] Add a random site as a favorite
- [ ] Then go to bookmarks and edit it, Set https://wikipedia.org as the
URL and Wikipedia as the title
- [ ] Type wiki
- [ ] Check now wikipedia is shown in top hits with the favorite icon

_Feature 5_

- [ ] Type something on the search bar
- [ ] Check history/bookmark/favorite suggestions aren't shown before
search suggestions are loaded

_Feature 6_
- [ ] Type something on the search bar
- [ ] Quickly delete everything
- [ ] Check history/bookmark/favorite suggestions aren't shown before
search suggestions and no "No suggestions found" message is shown either

### UI changes
[No UI changes](https://github.com/duckduckgo/Android/assets/6297834/645d3666-82a0-490a-9bf6-02901d14a117)


CrisBarreiro:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4362?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4362** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4362?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4357** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4357?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4349** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4349?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4341** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4341?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4332** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4332?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

<h2></h2>

 Join @CrisBarreiro and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.duckduckgo.common.utils.CurrentTimeProvider
+import io.reactivex.Single
+
+interface HistoryRepository {
+    fun getHistoryObservable(): Single<List<HistoryEntry>>
+    fun saveToHistory(
+        url: String,
+        title: String?,
+        query: String?,
+        isSerp: Boolean,
```
cmonfortep, line 29: I would say isSerp can be calculated inside the implementation and hidden from the interface.

1) Seems to me we are exposing the requirement of knowing if a visited url is serp in the interface. 2) thinking all the possible clients of this api, not forcing them to calculate it's simpler for them. So a nice api.

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    override fun saveToHistory(
+        url: String,
+        title: String?,
+        query: String?,
+        isSerp: Boolean,
+    ) {
+        historyDao.updateOrInsertVisit(url, title ?: "", query, isSerp, currentTimeProvider.currentTimeMillis())
+        fetchAndCacheHistoryEntries()
```
cmonfortep, line 55: I assume this will change so we don't query all entries on every insert, as it's just part of the MVP. just flagging in case we missed this part.

app/src/main/java/com/duckduckgo/app/history/HistoryEntry.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    data class VisitedPage(
+        override val url: Uri,
+        override val title: String,
+        override val visits: List<Date>,
+    ) : HistoryEntry(url = url, title = title, visits = visits)
+
+    data class VisitedSERP(
+        override val url: Uri,
+        override val title: String,
```
cmonfortep, line 40: does it make sense that a search has title? Not sure if it's adding value here or if we are just interested on query and url.

app/src/main/java/com/duckduckgo/app/history/HistoryEntry.kt
```diff
@@ -0,0 +1,52 @@
...
+        val query: String,
+        val queryTokens: List<String>? = null,
+        override val visits: List<Date>,
+    ) : HistoryEntry(url = url, title, visits = visits)
+}
+
+fun HistoryEntryWithVisits.toHistoryEntry(): HistoryEntry {
+    return when (historyEntry.isSerp) {
+        true -> VisitedSERP(historyEntry.url.toUri(), historyEntry.title, historyEntry.query ?: "", visits = visits.map { Date(it.date) })
```
cmonfortep, line 49: should we filter searches that are empty?

app/src/main/java/com/duckduckgo/app/history/HistoryEntry.kt
```diff
@@ -0,0 +1,52 @@
...
+package com.duckduckgo.app.history
+
+import android.net.Uri
+import androidx.core.net.toUri
+import com.duckduckgo.app.history.HistoryEntry.VisitedPage
+import com.duckduckgo.app.history.HistoryEntry.VisitedSERP
+import com.duckduckgo.app.history.store.HistoryEntryWithVisits
+import java.util.Date
+
+sealed class HistoryEntry(
```
cmonfortep, line 26: I have a similar question again: Visiting a site or performing a search, are both visited pages. So having 2 models is telling me we will return one or the other, and that we see them as different.

My expectations is that I can query visited sites and I receive all visited sites (including SERP urls); and I can query searches and I only receive searches the use has made.

Does it make sense?

app/src/main/java/com/duckduckgo/app/history/SaveToHistory.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import com.duckduckgo.app.browser.DuckDuckGoUrlDetector
+import com.duckduckgo.app.di.AppCoroutineScope
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+interface SaveToHistory {
```
cmonfortep, line 27: I see we have this interface here that is way simpler that the other one, and this is exactly what I was thinkings about ⭐ 

So question is why 2 interfaces, and which one is the one feature like autocomplete will use.

cmonfortep (commented):
Left some comments. Focused mainly on the models as discussed. Happy to follow on them via zoom next week.

I'm gonna propose you to consider creating a module for history. TBH, what I see makes me think we should have a module with a public interface. We have our own db, we are defining what the api should be, etc. 

The rest looks great to me, and I don't have major comments on the approach. Happy to chat more about this next week. I'm really excited with the feature 💯

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
CrisBarreiro: I don't think we should consider HistoryRepository to be the api, but rather SaveToHistory. If we look at it that way, then the API is pretty simple, we just ask clients for an URL and a title, and the implementation does all the calculations needed

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
CrisBarreiro: I want to make sure what I have in memory is the same I have in the database, but I think using an observable is a better way to go here. Alternatively, I could just make updateOrInsertVisit return the inserted item and update the cached list myself.

app/src/main/java/com/duckduckgo/app/history/HistoryEntry.kt
CrisBarreiro: For the autocomplete feature, we don't really need a title, but I'm adding it anyways as other features might need it. The query and the title are slightly different. For example, if you search for "cat", query will be "cat", but title will be "cat at DuckDuckGo". I think it's best to keep both

app/src/main/java/com/duckduckgo/app/history/HistoryEntry.kt
CrisBarreiro: Sure

app/src/main/java/com/duckduckgo/app/history/SaveToHistory.kt
CrisBarreiro: The way I see them, only this one should belong to an api module (if we decide to create a module for this feature) and therefore be the interface autocomplete and other features will use. As to why having 2 interfaces, it's just a way to separate business logic (usecase/interactor) and data access (repository), in the same way AutoComplete class handles all the autocomplete logic and deals with SavedSitesRepository (and HistoryRepository as a a next step) for data access

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
cmonfortep: Yeah, both approaches should work here.

app/src/main/java/com/duckduckgo/app/history/SaveToHistory.kt
cmonfortep: ok then let's propose a new module for history and open a proposal for api. That way  1) we will have a clear separation of what's public/internal, 2) we will have a more focused discussion on what to expose in the interface

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
cmonfortep: "I don't think we should consider HistoryRepository to be the api, but rather SaveToHistory." 

yeah, I saw there was another interface later on the review. All good if this is internal 👍

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
CrisBarreiro: I tested the observable approach better, but unfortunately, every time a new query is submitted, the observable is disposed, so the approach is useless, as we are effectively not caching anything. As for inserting the item and updating the cache myself, even though it might be a more efficient approach, it could lead to inconsistencies easily, as we'd need to handle duplicates twice (for database and cache)

app/src/main/java/com/duckduckgo/app/history/SaveToHistory.kt
CrisBarreiro: As discussed on MM, a new module has been added. See https://github.com/duckduckgo/Android/pull/4416

app/src/main/java/com/duckduckgo/app/history/HistoryRepository.kt
CrisBarreiro: Closing this thread as we've already discussed the approach

app/src/main/java/com/duckduckgo/app/history/store/HistoryEntryEntity.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override fun hashCode(): Int {
+        return url.hashCode()
+    }
+}
+
+@Entity(tableName = "visits_list", primaryKeys = ["date", "historyEntryId"])
+data class VisitEntity(
+    val historyEntryId: Long,
+    val date: Long,
```
CrisBarreiro, line 45: This becomes a timestamp for LocalDateTime in https://github.com/duckduckgo/Android/pull/4416

PR #4382, opened 2024-04-03
Github Issue: [Bug] Download confirmation dialog is cut off in vertical/tablet mode

anikiki:
Task/Issue URL: https://app.asana.com/0/715106103902962/1206905297723362/f

### Description
Expanded the download bottom sheet.

### Steps to test this PR

- [x] Go to a site from where you can download files. E.g https://www.thinkbroadband.com/download
- [x] With the device in landscape download a file.
- [x] Notice the bottomsheet is shown in full height.

### UI changes (landscape). No changes in portrait.
| Before (landscape) | After (landscape) |
| ------ | ----- |
|Download_land_before|Download_land_after_v2|

| Before (portrait) | After (portrait) |
| ------ | ----- |
|Download_portrait_before|Download_portrait_after|

### DIFF
--- a/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DownloadConfirmationFragment.kt
+++ b/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DownloadConfirmationFragment.kt
@@ -84,7 +84,9 @@ class DownloadConfirmationFragment : BottomSheetDialogFragment() {
     }
 
     private fun setupViews(binding: DownloadConfirmationBinding) {
-        (dialog as BottomSheetDialog).behavior.state = BottomSheetBehavior.STATE_EXPANDED
+        if (resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE) {
+            (dialog as BottomSheetDialog).behavior.state = BottomSheetBehavior.STATE_EXPANDED
+        }
         val fileName = file?.name ?: ""
         binding.downloadMessage.text = fileName
         binding.continueDownload.setOnClickListener {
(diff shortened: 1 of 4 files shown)


anikiki:
* **#4382** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @anikiki and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DownloadConfirmationFragment.kt
```diff
@@ -82,6 +84,7 @@ class DownloadConfirmationFragment : BottomSheetDialogFragment() {
     }
 
     private fun setupViews(binding: DownloadConfirmationBinding) {
+        (dialog as BottomSheetDialog).behavior.state = BottomSheetBehavior.STATE_EXPANDED
```
nalcalag, line 87: With this change, corners won't be rounded anymore even in portrait mode. Therefore, I would apply this state only if the orientation is landscape if you agree. I tried to find a quick solution where the corners would remain rounded even in the expanded mode by default, but I couldn't find one so I think this would be enough to fix this issue just now.
```suggestion
        if (resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE) {
            (dialog as BottomSheetDialog).behavior.state = BottomSheetBehavior.STATE_EXPANDED
        }
```

PR #4386, opened 2024-04-04
allow contributing from modules to matching attributes mappers

cmonfortep:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1207001823978864/f 
Api proposal change: https://app.asana.com/0/1202552961248957/1207001823978859/f

### Description
The PR allows modules to contribute with matching attribute mappers to the remote messaging processor.

As part of the new api we are introducing a new matching attribute that makes use of new api.

### Steps to test this PR

_Feature 1_
- [x] checkout branch
- [x] go to RemoteMessagingService and change the url with `https://jsonblob.com/api/jsonBlob/1225382942368194560`
- [x] fresh install
- [x] skip onboarding
- [x] you shouldn't see any RMF in the new tab
- [x] now create a sync account
- [x] Go to https://www.jsonblob.com/1225382942368194560 and bump version +1 and click save
- [x] restart app (using fire button or killing it)
- [x] you should now see a remote message in the new tab

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/JsonRulesMapper.kt
+++ b/remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/JsonRulesMapper.kt
@@ -16,6 +16,8 @@
 
 package com.duckduckgo.remote.messaging.impl.mappers
 
+import com.duckduckgo.remote.messaging.api.JsonMatchingAttribute
+import com.duckduckgo.remote.messaging.api.JsonToMatchingAttributeMapper
 import com.duckduckgo.remote.messaging.api.MatchingAttribute
 import com.duckduckgo.remote.messaging.impl.models.*
 import com.duckduckgo.remote.messaging.impl.models.toIntOrDefault
@@ -27,14 +29,14 @@ import timber.log.Timber
 private val dateFormatter = SimpleDateFormat("yyyy-mm-dd")
 
 @Suppress("UNCHECKED_CAST")
-private val localeMapper: (JsonMatchingAttribute) -> MatchingAttribute<String> = { jsonMatchingAttribute ->
+private val localeMapper: (JsonMatchingAttribute) -> MatchingAttribute = { jsonMatchingAttribute ->
     Locale(
         value = jsonMatchingAttribute.value.toStringList(),
         fallback = jsonMatchingAttribute.fallback,
     )
 }
 
-private val osApiMapper: (JsonMatchingAttribute) -> MatchingAttribute<Int> = { jsonMatchingAttribute ->
+private val osApiMapper: (JsonMatchingAttribute) -> MatchingAttribute = { jsonMatchingAttribute ->
     Api(
         value = jsonMatchingAttribute.value.toIntOrDefault(MATCHING_ATTR_INT_DEFAULT_VALUE),
         min = jsonMatchingAttribute.min.toIntOrDefault(MATCHING_ATTR_INT_DEFAULT_VALUE),
@@ -43,7 +45,7 @@ private val osApiMapper: (JsonMatchingAttribute) -> MatchingAttribute<Int> = { j
     )
 }
 
-private val webViewMapper: (JsonMatchingAttribute) -> MatchingAttribute<String> = { jsonMatchingAttribute ->
+private val webViewMapper: (JsonMatchingAttribute) -> MatchingAttribute = { jsonMatchingAttribute ->
     WebView(
         value = jsonMatchingAttribute.value.toStringOrDefault(MATCHING_ATTR_STRING_DEFAULT_VALUE),
         min = jsonMatchingAttribute.min.toStringOrDefault(MATCHING_ATTR_STRING_DEFAULT_VALUE),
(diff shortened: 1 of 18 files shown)


cmonfortep:
* **#4386** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @cmonfortep and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/JsonRulesMapper.kt
```diff
@@ -223,12 +225,18 @@ private val attributesMappers = mapOf(
...
+fun List<JsonMatchingRule>.mapToMatchingRules(
+    attributeMatchers: Set<JsonToMatchingAttributeMapper>,
+): Map<Int, List<MatchingAttribute<*>>> = this.map {
+    Pair(it.id, it.attributes.map { attrs -> attrs.map(attributeMatchers) })
 }.toMap()
 
-private fun Map.Entry<String, JsonMatchingAttribute>.map(): MatchingAttribute<*> {
+private fun Map.Entry<String, JsonMatchingAttribute>.map(attributeMatchers: Set<JsonToMatchingAttributeMapper>): MatchingAttribute<*> {
     return runCatching {
+        attributeMatchers.forEach {
```
cmonfortep, line 236: New logic: go through injected mappers and return if there's a hit.

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/JsonRulesMapper.kt
```diff
@@ -223,12 +225,18 @@ private val attributesMappers = mapOf(
...
+            val matchingAttribute = it.map(this.key, this.value)
+            if (matchingAttribute != null) return@runCatching matchingAttribute
+        }
         attributesMappers[this.key]?.invoke(this.value) ?: Unknown(this.value.fallback)
```
cmonfortep, line 240: Notice we keep old mappers and logic. So both (injected mappers and RMF mappers) coexists.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/rmf/SyncMatchingAttributes.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.remote.messaging.api.AttributeMatcherPlugin
+import com.duckduckgo.remote.messaging.api.MatchingAttribute
+import com.duckduckgo.sync.store.SyncStore
+import com.squareup.anvil.annotations.ContributesMultibinding
+import javax.inject.Inject
+
+@ContributesMultibinding(AppScope::class)
+class SyncAttributeMatcherPlugin @Inject constructor(
```
cmonfortep, line 27: This is how a feature adds support for matching attributes evaluation. (this interface already existed)

remote-messaging/remote-messaging-api/src/main/java/com/duckduckgo/remote/messaging/api/AttributeMatcherPlugin.kt
```diff
@@ -17,9 +17,19 @@
 package com.duckduckgo.remote.messaging.api
 
 interface AttributeMatcherPlugin {
-    suspend fun evaluate(matchingAttribute: MatchingAttribute<*>): Boolean?
+    suspend fun evaluate(matchingAttribute: MatchingAttribute): Boolean?
 }
 
-interface MatchingAttribute<T> {
-    fun matches(matchingValue: T): Boolean?
+interface MatchingAttribute
```
cmonfortep, line 23: I've removed the method `matches` from the interface and the generic from the interface, they are not adding any value, and it's simpler like this.

nalcalag (approved):
LGTM and works as expected 💯 
**_NIT:_**
I have a question regarding the public interface `JsonToMatchingAttributeMapper`:
- The implementation doesn't require anything from the target module (sync in this case), and we are exposing some RMF classes which, in my opinion, is not needed?
- Shouldn't the implementation of `JsonToMatchingAttributeMapper` be done directly in the RMF module, and only `MatchingAttribute` and `AttributeMatcherPlugin` be exposed as external API methods?

cmonfortep:
> The implementation doesn't require anything from the target module
This is not 100% correct. The implementation requires to return a `MatchingAttribute` like you pointed out. 
If we only expose `MatchingAttribute` in the api but keep the jsonmapping inside `:sync` we will have another problem. The new attribute will be declared in another module (that `sync` should not depend on) so we won't be able to return that type, meaning you can't map it.
We should give full responsability of declaring, mapping and evaluating matching attributes to other modules, and RMF only rely on their return values to process the config. 

Let's chat if that's not clear, but thanks again for your comments, I really appreciate you trying to close every open question you had when doing the review ⭐

PR #4419, opened 2024-04-11
Do not use store if encryption is not supported

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1207043620647316/f

### Description
See task

### Steps to test this PR
See task


aitorvs (approved):

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/RealSubscriptions.kt
karlenDimla, line 1: Don't we need to add the guard first and not inside the map for getEntitlementStatus?

PR #4438, opened 2024-04-18
Maestro: e2e retries

malmstein:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206929980548371/f

### Description
Fixes failing Sync test

### Steps to test this PR
Run Action

### DIFF
--- a/.github/workflows/sync-critical-path.yml
+++ b/.github/workflows/sync-critical-path.yml
@@ -61,7 +61,7 @@ jobs:
         run: find . -name "*.apk"  -exec mv '{}' apk/release.apk \;
 
       - name: Sync Flows
-        uses: mobile-dev-inc/action-maestro-cloud@v1.6.0
+        uses: mobile-dev-inc/action-maestro-cloud@v1.8.1
         with:
           api-key: ${{ secrets.MOBILE_DEV_API_KEY }}
           name: ${{ github.sha }}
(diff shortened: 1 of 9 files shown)


.github/workflows/sync-critical-path.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         run: find . -name "*.apk"  -exec mv '{}' apk/release.apk \;
 
       - name: Sync Flows
-        uses: mobile-dev-inc/action-maestro-cloud@v1.6.0
+        uses: mobile-dev-inc/action-maestro-cloud@v1.8.1
```
aitorvs, line 64: I know it might not be part of these PR, but we have multiple maestro action versions, 1.4.1, 1.6.0 and now also 1.8.1.
Can we (as part of this PR) update all yml files to use the same maestro version?

.maestro/config.yaml
```diff
@@ -2,6 +2,7 @@ appStartup:
   enabled: false
 appSize:
   enabled: false
+disableRetries: true
```
cmonfortep, line 5: 😱
image

🎉

cmonfortep (approved):
LGTM

.github/workflows/sync-critical-path.yml
malmstein: Sounds good @aitorvs , I’ll take care of it

PR #4445, opened 2024-04-22
Update copy when deleting passwords to consider whether sync enabled or not

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206638124888258/f 

### Description
Copy changes around deleting passwords (now showing a different message when sync is enabled vs disabled). 

Translations will come in https://github.com/duckduckgo/Android/pull/4446

# Steps to test this PR

### single saved password
- [ ] Disable sync (or leave it disabled if off already)
- [ ] Visit password management screen (overflow -> passwords)
- [ ] Add a single saved password (so you only have 1 total). 
- [ ] Return to password management screen
- [ ] Then choose overflow -> Delete All Passwords
- [ ] Verify the dialog title is **Are you sure you want to delete this password?**
- [ ] Verify the dialog message says "password" (singular) and "account" (singular)
- [ ] Verify the dialog message says it'll be "deleted from this device"
- [ ] Cancel the prompt. Click on the overflow of the saved password, and choose **Delete**. Ensure the last two checks you just did are still valid for this scenario.

### multiple saved passwords
- [ ] Add a second saved password (so you now have 2 total), and return to password management screen
- [ ] Then choose overflow -> Delete All Passwords
- [ ] Verify the dialog title is **Are you sure you want to delete 2 passwords**
- [ ] Verify the dialog message says "password**s**" and "account**s**"
- [ ] Verify the dialog message says it'll be "deleted from this device"

### multiple saved passwords
- [ ] Enable sync
- [ ] Visit password management screen (overflow -> passwords). Ensure you have > 1 saved password.
- [ ] Then choose overflow -> Delete All Passwords
- [ ] Verify the dialog title is **Are you sure you want to delete 2 passwords**
- [ ] Verify the dialog message says "password**s**" and "account**s**"
- [ ] Verify the dialog message says it'll be "deleted from all synced devices"

### single saved password
- [ ] Enable sync
- [ ] Visit password management screen (overflow -> passwords). Ensure you have exactly 1 saved password.
- [ ] Then choose overflow -> Delete All Passwords
- [ ] Verify the dialog title is **Are you sure you want to delete this password?**
- [ ] Verify the dialog message says "password" (singular) and "account" (singular)
- [ ] Verify the dialog message says it'll be "deleted from all synced devices"

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
@@ -104,6 +104,9 @@ class AutofillManagementCredentialsMode : DuckDuckGoFragment(R.layout.fragment_a
     @Inject
     lateinit var browserNav: BrowserNav
 
+    @Inject
+    lateinit var stringBuilder: AutofillManagementStringBuilder
+
     // we need to revert the toolbar title when this fragment is destroyed, so will track its initial value
     private var initialActionBarTitle: String? = null
 
@@ -176,21 +179,28 @@ class AutofillManagementCredentialsMode : DuckDuckGoFragment(R.layout.fragment_a
 
     private fun launchDeleteLoginConfirmationDialog() {
         this.context?.let {
-            TextAlertDialogBuilder(it)
-                .setTitle(R.string.autofillDeleteLoginDialogTitle)
-                .setMessage(R.string.credentialManagementDeletePasswordConfirmationMessage)
-                .setDestructiveButtons(true)
-                .setPositiveButton(R.string.autofillDeleteLoginDialogDelete)
-                .setNegativeButton(R.string.autofillDeleteLoginDialogCancel)
-                .addEventListener(
-                    object : TextAlertDialogBuilder.EventListener() {
-                        override fun onPositiveButtonClicked() {
-                            viewModel.onDeleteCurrentCredentials()
-                            viewModel.onExitCredentialMode()
-                        }
-                    },
-                )
-                .show()
+            lifecycleScope.launch(dispatchers.io()) {
+                val dialogTitle = stringBuilder.stringForDeletePasswordDialogConfirmationTitle(numberToDelete = 1)
+                val dialogMessage = stringBuilder.stringForDeletePasswordDialogConfirmationMessage(numberToDelete = 1)
+
+                withContext(dispatchers.main()) {
+                    TextAlertDialogBuilder(it)
+                        .setTitle(dialogTitle)
+                        .setMessage(dialogMessage)
+                        .setDestructiveButtons(true)
+                        .setPositiveButton(R.string.autofillDeleteLoginDialogDelete)
+                        .setNegativeButton(R.string.autofillDeleteLoginDialogCancel)
+                        .addEventListener(
+                            object : TextAlertDialogBuilder.EventListener() {
+                                override fun onPositiveButtonClicked() {
+                                    viewModel.onDeleteCurrentCredentials()
+                                    viewModel.onExitCredentialMode()
+                                }
+                            },
+                        )
+                        .show()
+                }
+            }
         }
     }
 
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementListMode.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementListMode.kt
@@ -69,6 +69,7 @@ import com.duckduckgo.di.scopes.FragmentScope
 import com.duckduckgo.mobile.android.R as CommonR
 import javax.inject.Inject
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
 @InjectWith(FragmentScope::class)
 class AutofillManagementListMode : DuckDuckGoFragment(R.layout.fragment_autofill_management_list_mode) {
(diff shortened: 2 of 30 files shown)


autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
```diff
@@ -176,9 +179,12 @@ class AutofillManagementCredentialsMode : DuckDuckGoFragment(R.layout.fragment_a
 
     private fun launchDeleteLoginConfirmationDialog() {
         this.context?.let {
+            val title = stringBuilder.stringForDeleteAllPasswordsDialogConfirmationDialogTitle(numberToDelete = 1)
```
cmonfortep, line 182: Seems weird to read `stringForDeleteAllPasswords` but then passing as argument `1`. should we create another method like we have for `stringForDeleteLoginsConfirmationDialogMessage` instead?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementListMode.kt
```diff
@@ -319,9 +322,12 @@ class AutofillManagementListMode : DuckDuckGoFragment(R.layout.fragment_autofill
 
     private fun launchDeleteLoginConfirmationDialog(loginCredentials: LoginCredentials) {
         this.context?.let {
+            val title = stringBuilder.stringForDeleteAllPasswordsDialogConfirmationDialogTitle(numberToDelete = 1)
```
cmonfortep, line 325: same as above for `tringForDeleteAllPasswords` but passing 1 as argument.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementStringBuilder.kt
```diff
@@ -0,0 +1,63 @@
...
+        return context.resources.getQuantityString(
+            R.plurals.credentialManagementDeleteAllPasswordsDialogConfirmationTitle,
+            numberToDelete,
+            numberToDelete,
+        )
+    }
+
+    override fun stringForDeleteLoginsConfirmationDialogMessage(numberToDelete: Int): String {
+        with(context.resources) {
+            val firstMessage = deleteAllPasswordsWarning(numberToDelete)
```
cmonfortep, line 48: ok, now I'm a bit confused. Do we use the display the same warning when removing all vs when removing only 1?

There's also a discrepancy in the method names not sure if intended: one says Passwords, the other one says Logins. Should be logins in both?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementStringBuilder.kt
```diff
@@ -0,0 +1,63 @@
...
+            val secondMessage = getQuantityString(R.plurals.credentialManagementDeleteAllSecondInstruction, numberToDelete)
+
+            return "$firstMessage $secondMessage"
+        }
+    }
+
+    private fun Resources.deleteAllPasswordsWarning(numberToDelete: Int): String {
+        val stringResId = if (deviceSyncState.isUserSignedInOnDevice()) {
```
cmonfortep, line 56: This should be executed in background. Sorry it was not explicit, I should add a `checkMainThread` there.

cmonfortep (commented):
lgtm, but leaving some comments about naming and asking to move the sync method call to background. I will go through the test cases when the background part is done so it's final. But those are the only 2 things I've seen.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementStringBuilder.kt
CDRussell: _>Do we use the display the same warning when removing all vs when removing only 1?_

**Dialog Title**
We are re-using the same code here because there are a few scenarios:

1. deleting a single password
1. choosing to delete-all when you have a single password saved
1. choosing to delete-all when you have multiple passwords saved

For 1. and 2., we show the same title. For 3, it's a different one.


**Dialog Message**

For messages, we change the string depending on whether:

- sync is enabled or not
- whether deleting one or multiple (pluralisation of some words change)


**Making the code clearer**
As discussed offline, will rename the string builder functions so that: 
- remove the `all` part from the function name (since it's only sometimes dealing with delete-all vs deleting one)
- naming is consistent; both referring to passwords rather than login

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/ui/credential/management/viewing/AutofillManagementCredentialsMode.kt
CDRussell: See https://github.com/duckduckgo/Android/pull/4445#discussion_r1583009435

CDRussell:
@cmonfortep moved sync check to background thread and renamed string builder functions. ready for review. (can ignore the lint error since that'll be solved in the translations branch for this)

CDRussell:
@cmonfortep pinging for re-review

Note, I had to change how we differentiate between singular and plural cases. Before, got caught out with the "implied plural" lint warnings. So now we check in code if number of elements == 1 and use a normal string if so, and else grab the plural form.

cmonfortep (approved):
LGTM

PR #4462, opened 2024-04-24
Update anvil to 2.5 beta to fix every other build needing a clean

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1207159274091706/f 

### Description
Updates anvil to `2.5.0-beta07`

#### Why jump to a beta?
The current version requires a clean build every other build, meaning even small code changes are taking ~2 mins to test. This problem is resolved on 2.5 branch, so even though there isn't a stable release it's worth it to avoid this tedious build issue and get the code deploying again within seconds.

**Typical build times i'm seeing**
| Before | Now
|--------|--------|
| 1m 40s | 25s | 

##### Deprecated priorities 
Anvil changed how you indicate which binding has priority if there are competing ones.

| Old Way | New Way |
|--------|--------|
|`priority = ContributesBinding.Priority.NORMAL` |`rank = ContributesBinding.RANK_NORMAL` |
|`priority = ContributesBinding.Priority.HIGHEST` | `rank = ContributesBinding.RANK_HIGHEST` |
|etc...|etc...

##### Can no longer access `ContributesBinding::class.java` 

One of the changes means that we can't access the java class like the above, which we only did in two unit tests for the code generator. It's still fine to reference `ContributesBinding::class` which is what's needed in the codegen itself, but our test was needing `ContributesBinding::class.java` which we can't do; have ignored these two tests for now.

### Steps to test this PR

- [x] Ensure all CI tests pass
- [x] Smoke test, since Anvil can affect everything app-wide

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/sync/FakeDeviceSyncState.kt
+++ b/app/src/main/java/com/duckduckgo/app/sync/FakeDeviceSyncState.kt
@@ -26,7 +26,7 @@ import javax.inject.*
  */
 @ContributesBinding(
     scope = AppScope::class,
-    priority = ContributesBinding.Priority.NORMAL,
+    rank = ContributesBinding.RANK_NORMAL,
 )
 class FakeDeviceSyncState @Inject constructor() : DeviceSyncState {
     override fun isFeatureEnabled(): Boolean = false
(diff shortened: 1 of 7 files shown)


aitorvs (approved):
lgtm and smoked test as expected

PR #4463, opened 2024-04-24
Make pause vpn during calls available to prod

karlenDimla:
Task/Issue URL: https://app.asana.com/0/1203137811378537/1206289136451167/f

### Description
See attached task description

### Steps to test this PR
https://app.asana.com/0/0/1207157151658716/f

### DIFF
--- a/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/snooze/VpnDisableOnCall.kt
+++ b/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/snooze/VpnDisableOnCall.kt
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2024 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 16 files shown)


network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/snooze/VpnDisableOnCall.kt
```diff
@@ -0,0 +1,110 @@
...
+            @Deprecated("Deprecated in Java")
+            override fun onCallStateChanged(
+                state: Int,
+                phoneNumber: String?,
+            ) {
+                appCoroutineScope.launch(dispatcherProvider.io()) {
+                    logcat { "Call state: $state" }
+                    if (state == TelephonyManager.CALL_STATE_IDLE) {
+                        _callInProgress = false
+                        vpn.start()
```
karlenDimla, line 65: The downside of using VPN here is that IF user has both AppTP and NetP enabled AND this setting is enabled, once the VPN subscription expires, user will not be able to turn off this setting.

aitorvs: Should we add a check `networkProtectionState.isEnabled()` in the `else` branch?

aitorvs (approved):

PR #4478, opened 2024-04-27
Add pproEligible and pproSubscriber matching attrs

aitorvs:
Task/Issue URL: https://app.asana.com/0/488551667048375/1206990668213579/f

### Description
Add PPro eligible/subscriber matching attrs to RMF

### Steps to test this PR
See [Asana task](https://app.asana.com/0/0/1207185787167221/f)

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/rmf/RMFPProSubscriberMatchingAttribute.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/rmf/RMFPProSubscriberMatchingAttribute.kt
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2024 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.subscriptions.impl.rmf
+
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.remote.messaging.api.AttributeMatcherPlugin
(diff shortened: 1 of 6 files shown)


aitorvs:
* **#4478** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/rmf/RMFPProSubscriberMatchingAttribute.kt
```diff
@@ -0,0 +1,73 @@
...
+                jsonMatchingAttribute.value?.let {
+                    ProSubscriberMatchingAttribute(jsonMatchingAttribute.value as Boolean)
+                }
+            }
+            else -> null
+        }
+    }
+
+    private suspend fun Subscriptions.isSubscribed(): Boolean {
+        return getAccessToken() != null
```
marcosholgado, line 63: Don't have all the context of what's needed here but this is not necessarily true. If you want to know if someone is subscribed so you should use `getSubscriptionStatus` and check for AUTO_RENEWABLE, NOT_AUTO_RENEWABLE or GRACE_PERIOD.

aitorvs: Makes sense. Question, in what cases does the user have an access token and it's not subscribed?

marcosholgado: Two cases of the top of my head:

1. Once a subscription expires
2. If the user never finished the purchase because we create the account first (which gets you an access token) to be able to associate it with the subscription.

aitorvs: Then I think is fine as is for now. We want to RMF users that never subscribed. 
For (1) is what we want, we want to treat expired users as "subscribed "
For (2) is an edge case I guess but should be fine for now. 

I think soon we may want to add RMF attr for account state. We have a task for that afair but not yet concluded.

cmonfortep (approved):
LGTM

aitorvs:
### Merge activity

* **Apr 29, 8:46 AM EDT**: @aitorvs started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/duckduckgo/Android/4478).

PR #4601, opened 2024-05-30
Add ability to support importing passwords from Google as CSV

CDRussell:
Task/Issue URL: https://app.asana.com/0/608920331025315/1207415218447811/f

### Description
Adds ability to import passwords exported as a CSV from Google Password Manager (GPM). This functionality is used by the higher up branches in the stack to provide a guided way to import from GPM, but can be tested directly from autofill dev settings and choosing to import a CSV file.

Test CSV files are available in [Example CSV files for testing](https://app.asana.com/0/1208737781360301/1208737781360301/f)

#### Happy path: all unique
- [x] Fresh install `internal` build type, and launch autofill dev settings
- [x] Choose `Import CSV`
- [x] Choose CSV file containing only unique entries and verify all are imported (e.g., choose `example_csv_20.csv` and make sure 20 are imported)

#### Happy path: importing with some duplicates in the list
- [x] Delete all saved credentials
- [x] Choose `Import CSV` again, and this time choose file containing duplicates 
- [x] verify correct amount imported (e.g., choose `dupes_20_total_17_unique.csv` and make sure 17 are imported)

#### Happy path: importing with a duplicate already saved in the DB
- [x] Add a credential that matches what is in the CSV you will export
- [x] Import the CSV
- [x] Verify the correct amount imported and the already-saved one wasn't duplicated

#### Non-CSV files
- [x] Choose `Import CSV`
- [x] Pick a file that isn't a password CSV export
- [x] Make sure it handles this gracefully


CDRussell:
* **#4602** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4602?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4601** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4601?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

malmstein:
The Purge!

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/ExistingPasswordMatchDetector.kt
```diff
@@ -0,0 +1,51 @@
...
+
+@ContributesBinding(AppScope::class)
+class DefaultExistingPasswordMatchDetector @Inject constructor(
+    private val autofillStore: InternalAutofillStore,
+    private val dispatchers: DispatcherProvider,
+) : ExistingPasswordMatchDetector {
+
+    override suspend fun alreadyExists(newCredentials: LoginCredentials): Boolean {
+        return withContext(dispatchers.io()) {
+            val credentials = autofillStore.getAllCredentials().firstOrNull() ?: return@withContext false
```
cmonfortep, line 40: This method is being called once for every entry on the imported file, and we are accessing db and retrieving allCredentials every time. That can be several calls to db for an import, can we improve this part?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/GooglePasswordBlobDecoder.kt
```diff
@@ -0,0 +1,47 @@
...
+
+@ContributesBinding(AppScope::class)
+class GooglePasswordBlobDecoderImpl @Inject constructor(
+    private val dispatchers: DispatcherProvider,
+) : GooglePasswordBlobDecoder {
+
+    override suspend fun decode(data: String): String {
+        return withContext(dispatchers.io()) {
+            kotlin.runCatching {
+                val base64Data = data.split(",")[1]
```
cmonfortep, line 39: had to check the test to understand this part, probably just leave a comment we are skipping the file type, or something that adds context here.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/CsvPasswordConverter.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.duckduckgo.autofill.impl.importing.CsvPasswordConverter.CsvPasswordImportResult
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+import kotlinx.coroutines.withContext
+import kotlinx.parcelize.Parcelize
+
+interface CsvPasswordConverter {
+    suspend fun readCsv(blob: String): CsvPasswordImportResult
```
cmonfortep, line 31: I would make it clear this is expected to receive an `encodedBlob`, as I see in the implementation we will perform decoding.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/ImportedPasswordValidator.kt
```diff
@@ -0,0 +1,43 @@
...
+import javax.inject.Inject
+
+interface ImportedPasswordValidator {
+    fun isValid(loginCredentials: LoginCredentials): Boolean
+}
+
+@ContributesBinding(AppScope::class)
+class DefaultImportedPasswordValidator @Inject constructor() : ImportedPasswordValidator {
+
+    override fun isValid(loginCredentials: LoginCredentials): Boolean {
```
cmonfortep, line 31: Just validating, do we think is correct when importing credentials that if no domain is present we don't import the entry? I'm checking and we allow users to add credentials with not domain, so wondering why we think they are not valid.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/PasswordImporter.kt
```diff
@@ -0,0 +1,122 @@
...
+    @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
+) : PasswordImporter {
+
+    private val _importStatus = MutableSharedFlow<ImportResult>(replay = 1)
+    private val mutex = Mutex()
+
+    override suspend fun importPasswords(importList: List<LoginCredentials>, originalImportListSize: Int): String {
+        val jobId = UUID.randomUUID().toString()
+
+        mutex.withLock {
```
cmonfortep, line 81: I believe the intention here is to control we don't import while we are importing...right? If so the implementation is not correct because `launch` is fire and forget, and will automatically release the lock. That means in practice we execute as many imports as we trigger.

I think the logic was just for something specific on the internal screens, so I'm not sure if it will be still present in following PRs. Just questioning if we still need this kind of logic here in production. What is the scenario we were trying to solve excluding the internal case?

cmonfortep (commented):
Leaving comments from first review while in draft.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/CsvFileReader.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+@ContributesBinding(AppScope::class)
+class ContentResolverFileReader @Inject constructor(
+    private val context: Context,
+) : CsvFileReader {
+
+    override fun readCsvFile(fileUri: Uri): String {
+        return context.contentResolver.openInputStream(fileUri)?.use { inputStream ->
+            BufferedReader(InputStreamReader(inputStream)).use { reader ->
```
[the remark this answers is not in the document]
CDRussell: UTF 8 is the default, but will add it to be explicit 👍

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/DomainNameNormalizer.kt
```diff
@@ -0,0 +1,40 @@
...
+    suspend fun normalizeDomains(unnormalized: List<LoginCredentials>): List<LoginCredentials>
+}
+
+@ContributesBinding(AppScope::class)
+class DefaultDomainNameNormalizer @Inject constructor(
+    private val urlMatcher: AutofillUrlMatcher,
+) : DomainNameNormalizer {
+    override suspend fun normalizeDomains(unnormalized: List<LoginCredentials>): List<LoginCredentials> {
+        return unnormalized.map {
+            val currentDomain = it.domain ?: return@map null
```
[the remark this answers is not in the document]
CDRussell: >urlMatcher.cleanRawUrl(currentDomain) is returning a domain
- **yes**

>it.domain can contain a url
- **yes**
- normally, `LoginCredentials` would be a domain, but since we've brought in the external format it's temporarily a URL, and converting to a domain is what's happening to get the external format into the correct internal format

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/CsvPasswordConverter.kt
```diff
@@ -0,0 +1,94 @@
...
+            is CsvPasswordParser.ParseResult.Success -> {
+                val toImport = deduplicateAndCleanup(parseResult.passwords)
+                CsvPasswordImportResult.Success(parseResult.passwords.size, toImport)
+            }
+            is CsvPasswordParser.ParseResult.Error -> CsvPasswordImportResult.Error
+        }
+    }
+
+    private suspend fun deduplicateAndCleanup(allPasswords: List<LoginCredentials>): List<LoginCredentials> {
+        val dedupedPasswords = allPasswords.distinct()
```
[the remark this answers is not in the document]
CDRussell: Good point on the conflation between credentials and passwords. Part of the reason is that we used to call it `Credentials` at the product level, but are now referring to it as `Passwords` at the product level, while continuing to make the distinction between passwords/credentials in code. 

I'll take a pass at using `credentials` in this new code, limiting the use of `passwords` to when actually talking about the password within a credential.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/ExistingPasswordMatchDetector.kt
```diff
@@ -0,0 +1,51 @@
...
+import javax.inject.Inject
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.withContext
+
+interface ExistingPasswordMatchDetector {
+    suspend fun alreadyExists(newCredentials: LoginCredentials): Boolean
+}
+
+@ContributesBinding(AppScope::class)
+class DefaultExistingPasswordMatchDetector @Inject constructor(
```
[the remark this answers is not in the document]
CDRussell: https://github.com/duckduckgo/Android/pull/4601#discussion_r1830908176

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/ImportedPasswordValidator.kt
CDRussell: Will take to Asana: [What to do if there is no domain in the imported credential?](https://app.asana.com/0/1208707347544984/1208707347544984/f)

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/CsvPasswordConverter.kt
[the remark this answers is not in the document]
CDRussell: Done

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStore.kt
```diff
@@ -270,7 +270,7 @@ class SecureStoreBackedAutofillStore @Inject constructor(
         return matchType
     }
 
-    private fun LoginCredentials.prepareForReinsertion(): WebsiteLoginDetailsWithCredentials {
+    private fun LoginCredentials.prepareForBulkInsertion(): WebsiteLoginDetailsWithCredentials {
```
CDRussell, line 273: reusing this for bulk inserting, so renaming so it's no longer specific to re-insertion

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/PasswordImporter.kt
CDRussell: Changed this to be simpler

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/importing/DomainNameNormalizer.kt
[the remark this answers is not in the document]
cmonfortep: That's great to know. So we are bringing an external set of credentials, which follow a similar structure but differ on the expected format of the values. And we are using our domain object to map them. That can be problematic. I would like to avoid reusing our domain object to map an external object we don't control values. There's risk of introducing bugs difficult to spot and wrong expectations when working with those objects.

I propose we define a data class where we store the external objects first. We can have GoogleCSVLoginCrendential, with url instead of domain as attribute. Then, they will be mapped into our domain object as part of the processing.

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/DefaultDomainNameNormalizerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    @Test
+    fun whenInputDomainIsNullThenNormalizedToNullDomain() = runTest {
+        val input = listOf(creds(domain = null))
+        val output = testee.normalizeDomains(input)
+        assertEquals(1, output.size)
+        assertEquals(null, output.first().domain)
+    }
+
+    @Test
+    fun whenDomainCannotBeNormalizedThenIsIncludedUnmodified() = runTest {
```
cmonfortep, line 50: I'm not sure about this one 😅. If we don't know how to normalize the input, we just return the input. But then we store it as domain attribute...but that can be anything. 🤔

You have more context than me on Autofill, but what are the side effects of not having a domain in the domain attribute? I assume we are not validating the user manually enters something valid either...so at this point, I wonder why do we even normalize, or why do we even filter the "android://..." objects... 

Short term, just to consider if this is the expected behavior, and then we can chat about the class fields without blocking the changes here as it seems a more broad discussion.

cmonfortep (commented):
Works great. Let's discuss sync the models being used here, so I can have more clarity and decide next steps.

cmonfortep (approved):
Approving as I don't feel this is a blocker, but let's revisit together.

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/DefaultDomainNameNormalizerTest.kt
CDRussell: This behavior mirrors what we do when saving credentials. If we can, we normalize the URLs when we encounter them, or when the user enters them manually (a product decision from way back). If the URL is unrecongizable as a URL and therefore we can't normalize it (when saving from autofill or from a user entering manually), we save it unaltered.

CDRussell: >what are the side effects of not having a domain in the domain attribute

if it's unrecognizable, we won't be able to match on it and it won't be offered as an option in the autofill dialogs but will still be stored in the full credential list

PR #4697, opened 2024-06-26
Select the last tab and update selection after deleting the current tab

0nko:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1202148076467684/f

### Description
This PR selects the last tab in the list after the currently-selected tab is deleted. This fixes the behaviour when no tab was selected and when the user went back to the browser, the first tab was shown.

### Steps to test this PR

1. Create at least 3 tabs (with the last tab being selected)
2. Delete the last (selected) tab
3. Notice the tab is removed and the last of the remaining tab is selected
4. Tap on `Undo`
5. Notice the previous tab is shown and selected again
6. Delete the selected tab again
7. Tap on the back button
8. Notice the correct tab is shown

### UI changes
[Screen_recording_20240626_151442.webm](https://github.com/duckduckgo/Android/assets/1522856/2ae2fe33-275b-4eb8-a33c-62b647715d09)

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabGridItemDecorator.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabGridItemDecorator.kt
@@ -86,7 +86,7 @@ class TabGridItemDecorator(
     }
 
     companion object {
-        private val BORDER_RADIUS = 9.toPx().toFloat()
+        private val BORDER_RADIUS = 15.toPx().toFloat()
         private val BORDER_WIDTH = 2.toPx().toFloat()
         private val BORDER_PADDING = 3.toPx().toFloat()
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/tabs/ui/TabGridItemDecorator.kt
```diff
@@ -86,7 +86,7 @@ class TabGridItemDecorator(
     }
 
     companion object {
-        private val BORDER_RADIUS = 9.toPx().toFloat()
+        private val BORDER_RADIUS = 15.toPx().toFloat()
```
0nko, line 89: This is not directly related to the issue, but after working with the decorator I noticed the radius doesn’t match the card view (see the PR description).

anikiki (approved):

PR #4783, opened 2024-07-19
Add general settings

joshliebe:
Task/Issue URL: https://app.asana.com/0/1200204095367872/1207857476165688/f

### Description
Adds a "General" section to settings containing:
- Autocomplete Suggestions
- Recently Visited Sites
- Private Voice Search

### Steps to test this PR

_General setting_
- [ ] Go to “Settings” > “General”
- [ ] Verify that “m_settings_general_open” pixel is sent

_Autocomplete Suggestions_
- [ ] Toggle the setting on
- [ ] Verify that the “m_settings_general_autocomplete_on” pixel is sent
- [ ] Verify that autocomplete suggestions are shown
- [ ] Toggle the setting off
- [ ] Verify that the “m_settings_general_autocomplete_off” pixel is sent
- [ ] Verify that autocomplete suggestions are no longer shown

_Include Recently Visited Sites_
- [ ] Toggle the setting on
- [ ] Verify that the “m_settings_general_recent_sites_on” pixel is sent
- [ ] Verify that recently visited sites are shown
- [ ] Toggle the setting off
- [ ] Verify that the “m_settings_general_recent_sites_off” pixel is sent
- [ ] Verify that recently visited sites are not shown

_On a device that supports voice search_
- [ ] Verify that the “Private Voice Search” setting is shown
- [ ] Toggle the setting on
- [ ] Verify that the “m_settings_general_voice_search_on” pixel is sent
- [ ] Go back and verify that the voice search microphone is visible in the omnibar
- [ ] Toggle the setting off
- [ ] Verify that the “m_settings_general_voice_search_off” pixel is sent
- [ ] Go back and verify that the voice search microphone is not visible in the omnibar

_On a device that does not support voice search (Emulator)_
- [ ] Verify that the “Private Voice Search” setting is not visible
- [ ] Verify that the divider is not visible

### UI changes
| Before  | After | General |
| ------ | ----- |  ----- |
general_before|general_after|general_settings

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
@@ -108,14 +108,14 @@ class GeneralSettingsViewModel @Inject constructor(
     }
 
     fun onVoiceSearchChanged(checked: Boolean) {
-        voiceSearchRepository.setVoiceSearchUserEnabled(checked)
-        if (checked) {
-            voiceSearchRepository.resetVoiceSearchDismissed()
-            pixel.fire(VOICE_SEARCH_GENERAL_SETTINGS_ON)
-        } else {
-            pixel.fire(VOICE_SEARCH_GENERAL_SETTINGS_OFF)
-        }
-        viewModelScope.launch {
+        viewModelScope.launch(dispatcherProvider.io()) {
+            voiceSearchRepository.setVoiceSearchUserEnabled(checked)
+            if (checked) {
+                voiceSearchRepository.resetVoiceSearchDismissed()
+                pixel.fire(VOICE_SEARCH_GENERAL_SETTINGS_ON)
+            } else {
+                pixel.fire(VOICE_SEARCH_GENERAL_SETTINGS_OFF)
+            }
             _viewState.value = _viewState.value?.copy(voiceSearchEnabled = voiceSearchAvailability.isVoiceSearchAvailable)
         }
     }
--- a/app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2024 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.generalsettings
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.duckduckgo.anvil.annotations.ContributesViewModel
+import com.duckduckgo.app.pixels.AppPixelName.AUTOCOMPLETE_GENERAL_SETTINGS_TOGGLED_OFF
+import com.duckduckgo.app.pixels.AppPixelName.AUTOCOMPLETE_GENERAL_SETTINGS_TOGGLED_ON
+import com.duckduckgo.app.pixels.AppPixelName.AUTOCOMPLETE_RECENT_SITES_GENERAL_SETTINGS_TOGGLED_OFF
+import com.duckduckgo.app.pixels.AppPixelName.AUTOCOMPLETE_RECENT_SITES_GENERAL_SETTINGS_TOGGLED_ON
+import com.duckduckgo.app.settings.db.SettingsDataStore
+import com.duckduckgo.app.statistics.pixels.Pixel
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.di.scopes.ActivityScope
+import com.duckduckgo.history.api.NavigationHistory
+import com.duckduckgo.voice.api.VoiceSearchAvailability
+import com.duckduckgo.voice.impl.VoiceSearchPixelNames.VOICE_SEARCH_GENERAL_SETTINGS_OFF
+import com.duckduckgo.voice.impl.VoiceSearchPixelNames.VOICE_SEARCH_GENERAL_SETTINGS_ON
+import com.duckduckgo.voice.store.VoiceSearchRepository
+import javax.inject.Inject
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.onStart
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import timber.log.Timber
+
+@ContributesViewModel(ActivityScope::class)
+class GeneralSettingsViewModel @Inject constructor(
+    private val settingsDataStore: SettingsDataStore,
+    private val pixel: Pixel,
+    private val history: NavigationHistory,
+    private val voiceSearchAvailability: VoiceSearchAvailability,
+    private val voiceSearchRepository: VoiceSearchRepository,
+    private val dispatcherProvider: DispatcherProvider,
+) : ViewModel() {
+
+    data class ViewState(
+        val autoCompleteSuggestionsEnabled: Boolean = true,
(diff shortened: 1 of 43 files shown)


joshliebe:
* **#4783** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4783?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+            showVoiceSearch = voiceSearchAvailability.isVoiceSearchSupported,
+            voiceSearchEnabled = voiceSearchAvailability.isVoiceSearchAvailable,
+        ),
+    )
+
+    fun viewState(): Flow<ViewState> = viewState.onStart {
+        viewModelScope.launch(dispatcherProvider.io()) {
+            val autoCompleteEnabled = settingsDataStore.autoCompleteSuggestionsEnabled
+            if (!autoCompleteEnabled) {
+                history.setHistoryUserEnabled(false)
```
cmonfortep, line 76: QQ: why do we need to do this here?

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+    }
+
+    fun onAutocompleteSettingChanged(enabled: Boolean) {
+        Timber.i("User changed autocomplete setting, is now enabled: $enabled")
+        viewModelScope.launch(dispatcherProvider.io()) {
+            settingsDataStore.autoCompleteSuggestionsEnabled = enabled
+            if (!enabled) {
+                viewModelScope.launch() {
```
cmonfortep, line 97: this one seems not necessary since we are already inside a coroutine and io

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+                pixel.fire(AUTOCOMPLETE_RECENT_SITES_GENERAL_SETTINGS_TOGGLED_OFF)
+            }
+            withContext(dispatcherProvider.main()) {
+                viewState.emit(currentViewState().copy(autoCompleteRecentlyVisitedSitesSuggestionsUserEnabled = enabled))
+            }
+        }
+    }
+
+    fun onVoiceSearchChanged(checked: Boolean) {
+        voiceSearchRepository.setVoiceSearchUserEnabled(checked)
```
cmonfortep, line 133: should you wrap this also in a coroutine?

app/src/main/java/com/duckduckgo/app/privatesearch/PrivateSearchViewModel.kt
```diff
@@ -49,16 +50,22 @@ class PrivateSearchViewModel @Inject constructor(
...
+        val autoCompleteRecentlyVisitedSitesSuggestionsUserEnabled: Boolean,
+        val storeHistoryEnabled: Boolean,
     )
 
     sealed class Command {
         object LaunchCustomizeSearchWebPage : Command()
     }
 
-    private val viewState = MutableStateFlow(ViewState())
+    private val viewState = MutableStateFlow(
```
cmonfortep, line 62: any reason why we are doing this here?
Now on class init, we are querying the different data sources and we can block main thread (see runblocking), producing ANR.

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+    data class ViewState(
+        val autoCompleteSuggestionsEnabled: Boolean,
+        val autoCompleteRecentlyVisitedSitesSuggestionsUserEnabled: Boolean,
+        val storeHistoryEnabled: Boolean,
+        val showVoiceSearch: Boolean,
+        val voiceSearchEnabled: Boolean,
+    )
+
+    private val viewState = MutableStateFlow(
+        ViewState(
```
cmonfortep, line 63: this one worries me a bit, since getting the value for each attr will be done on class init. we should avoid that, and also using runblocking there.

We can have default values, or use `onStart` in the emitter flow.

cmonfortep (commented):
overall looks good! we need to review some of the parts that can affect class initialisation to ensure we don't block threads.

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
joshliebe: This logic is copied from `PrivateSearchViewModel` since it needs to have the same behavior

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
joshliebe: Again, this is copied from `PrivateSearchViewModel`

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
joshliebe: This is copied from `AccessibilitySettingsViewModel`

app/src/main/java/com/duckduckgo/app/privatesearch/PrivateSearchViewModel.kt
joshliebe: Yes, I’m doing this on purpose. If there are default values to start with, it causes two issues:
- It messes up the activity transition
- You see the switch changing state upon entering the screen

The `runBlocking` is on a method that gets a shared pref boolean (Which shouldn’t be a `suspend` function IMO), so it won’t cause ANRs.

The only other option is to show some sort of loading indicator (Which is not a pattern we follow) or hide the screen until it completes (Which won't look good).

I have recordings here of the transition/toggle behavior: https://app.asana.com/0/0/1207888184745508/1207887624571935/f

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
joshliebe: Same reasoning as https://github.com/duckduckgo/Android/pull/4783#discussion_r1692799765

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
joshliebe: Removed it from here and in `PrivateSearchViewModel`

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
joshliebe: I’ve moved everything in `onVoiceSearchChanged` into a coroutine, here and in `AccessibilitySettingsViewModel`

app/src/main/java/com/duckduckgo/app/privatesearch/PrivateSearchViewModel.kt
joshliebe: I’ve moved the initialization of `viewState` into `init` using a coroutine (https://github.com/duckduckgo/Android/pull/4783/commits/0fc4ea802b1e9e3e8b7b733c76322beedd6da3c3)

app/src/main/java/com/duckduckgo/app/generalsettings/GeneralSettingsViewModel.kt
joshliebe: I’ve changed this to use the same pattern as `PrivateSearchViewModel`, see https://github.com/duckduckgo/Android/pull/4783#discussion_r1694358700

joshliebe:
@cmonfortep I’ve addressed the comments, this is ready for re-review

cmonfortep (approved):
LGTM

app/src/main/java/com/duckduckgo/app/privatesearch/PrivateSearchViewModel.kt
cmonfortep: Some notes from our zoom call on Friday:

- If we need to access database or shared prefs to initialize a value as part of creating a class instance, that can produce an ANR since it happens on main thread.
- Not having an "idle" or "empty" state in UI forces us to define a default value in the code. That default value can be defined in the viewmodel (e.g: default viewstate value) or in the layout (e.g: default state of the view we put, including a toggle).
- `init` is just the early point we can execute a background query without blocking main thread, so we can try to query the datasource and get a value before layout is rendered. Reducing the chances of seeing the toggle changing state in the transition.

PR #4861, opened 2024-08-05
Support autocomplete recent item removal

joshliebe:
Task/Issue URL: https://app.asana.com/0/488551667048375/1207977618326502/f

### Description
Adds the ability to delete recent items from the autocomplete list.

### Steps to test this PR

_Remove query_
- [x] Search for something (eg. “dogs”)
- [x] Clear the search and type the first few letters
- [x] Long press the recent item
- [x] Verify that the keyboard is hidden and the dialog is shown
- [x] Press “Remove"
- [x] Verify that the `m_autocomplete_result_deleted` pixel is sent
- [x] Verify that the item is deleted
- [x] Verify that the keyboard is shown

_Remove URL_
- [x] Visit a URL (eg. “dog.com”)
- [x] Clear the search and type the first few letters
- [x] Long press the recent item
- [x] Verify that the keyboard is hidden and the dialog is shown
- [x] Press “Remove"
- [x] Verify that the `m_autocomplete_result_deleted` pixel is sent
- [x] Verify that the item is deleted
- [x] Verify that the keyboard is shown

_Cancel_
- [x] Do the same thing and press “Cancel" or tap outside the dialog
- [x] Verify that the item is not deleted

_Widget_
- [x] Add the widget
- [x] Follow the same steps as _Remove query_, _Remove URL_ and _Cancel_

### UI changes
| Browser  | Widget |
| ------ | ----- |
remove_history|remove_history_widget

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -94,6 +94,7 @@ import com.duckduckgo.app.browser.WebViewErrorResponse.OMITTED
 import com.duckduckgo.app.browser.applinks.AppLinksLauncher
 import com.duckduckgo.app.browser.applinks.AppLinksSnackBarConfigurator
 import com.duckduckgo.app.browser.autocomplete.BrowserAutoCompleteSuggestionsAdapter
+import com.duckduckgo.app.browser.autocomplete.KeyboardVisibilityUtil
 import com.duckduckgo.app.browser.commands.Command
 import com.duckduckgo.app.browser.commands.Command.ShowBackNavigationHistory
 import com.duckduckgo.app.browser.commands.NavigationCommand
@@ -233,6 +234,7 @@ import com.duckduckgo.common.ui.view.hideKeyboard
 import com.duckduckgo.common.ui.view.makeSnackbarWithNoBottomInset
 import com.duckduckgo.common.ui.view.show
 import com.duckduckgo.common.ui.view.showKeyboard
+import com.duckduckgo.common.ui.view.toPx
 import com.duckduckgo.common.ui.viewbinding.viewBinding
 import com.duckduckgo.common.utils.ConflatedJob
 import com.duckduckgo.common.utils.DispatcherProvider
@@ -548,6 +550,9 @@ class BrowserTabFragment :
     private lateinit var webViewContainer: FrameLayout
 
     private var bookmarksBottomSheetDialog: BookmarksBottomSheetDialog.Builder? = null
+
+    private var autocompleteItemOffsetTop: Int = 0
+    private var autocompleteFirstVisibleItemPosition: Int = 0
 
     private val findInPage
         get() = omnibar.findInPage
@@ -1550,6 +1555,7 @@ class BrowserTabFragment :
             is Command.LaunchScreen -> launchScreen(it.screen, it.payload)
             is Command.HideOnboardingDaxDialog -> hideOnboardingDaxDialog(it.onboardingCta)
             is Command.ShowRemoveSearchSuggestionDialog -> showRemoveSearchSuggestionDialog(it.suggestion)
+            is Command.AutocompleteItemRemoved -> autocompleteItemRemoved()
             else -> {
                 // NO OP
             }
@@ -1557,7 +1563,9 @@ class BrowserTabFragment :
     }
 
     private fun showRemoveSearchSuggestionDialog(suggestion: AutoCompleteSuggestion) {
+        storeAutocompletePosition()
         hideKeyboardRetainFocus()
+
         TextAlertDialogBuilder(requireContext())
             .setTitle(R.string.autocompleteRemoveItemTitle)
             .setCancellable(true)
@@ -1567,19 +1575,40 @@ class BrowserTabFragment :
                 object : TextAlertDialogBuilder.EventListener() {
                     override fun onPositiveButtonClicked() {
                         viewModel.onRemoveSearchSuggestionConfirmed(suggestion, omnibar.omnibarTextInput.text.toString())
-                        showKeyboard()
                     }
-
                     override fun onNegativeButtonClicked() {
-                        showKeyboard()
+                        showKeyboardAndRestorePosition(autocompleteFirstVisibleItemPosition, autocompleteItemOffsetTop)
                     }
-
                     override fun onDialogCancelled() {
-                        showKeyboard()
+                        showKeyboardAndRestorePosition(autocompleteFirstVisibleItemPosition, autocompleteItemOffsetTop)
                     }
                 },
             )
             .show()
+    }
+
+    private fun storeAutocompletePosition() {
+        val layoutManager = binding.autoCompleteSuggestionsList.layoutManager as LinearLayoutManager
+        autocompleteFirstVisibleItemPosition = layoutManager.findFirstVisibleItemPosition()
+        autocompleteItemOffsetTop = layoutManager.findViewByPosition(autocompleteFirstVisibleItemPosition)?.top ?: 0
+    }
+
+    private fun autocompleteItemRemoved() {
+        showKeyboardAndRestorePosition(autocompleteFirstVisibleItemPosition, autocompleteItemOffsetTop)
+    }
+
+    private fun showKeyboardAndRestorePosition(position: Int, offset: Int) {
+        val rootView = omnibar.omnibarTextInput.rootView
+        val keyboardVisibilityUtil = KeyboardVisibilityUtil(rootView)
+        keyboardVisibilityUtil.addKeyboardVisibilityListener {
+            scrollToPositionWithOffset(position, offset)
+        }
+        showKeyboard()
+    }
+
+    private fun scrollToPositionWithOffset(position: Int, offset: Int) {
+        val layoutManager = binding.autoCompleteSuggestionsList.layoutManager as LinearLayoutManager
+        layoutManager.scrollToPositionWithOffset(position, offset - 6.toPx())
     }
 
     private fun launchScreen(
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -1608,7 +1608,7 @@ class BrowserTabFragment :
 
     private fun scrollToPositionWithOffset(position: Int, offset: Int) {
         val layoutManager = binding.autoCompleteSuggestionsList.layoutManager as LinearLayoutManager
-        layoutManager.scrollToPositionWithOffset(position, offset - 6.toPx())
+        layoutManager.scrollToPositionWithOffset(position, offset - AUTOCOMPLETE_PADDING_DP.toPx())
     }
 
     private fun launchScreen(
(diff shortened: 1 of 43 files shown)


joshliebe:
* **#4861** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4861?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1548,12 +1554,63 @@ class BrowserTabFragment :
...
+
+        TextAlertDialogBuilder(requireContext())
+            .setTitle(R.string.autocompleteRemoveItemTitle)
+            .setCancellable(true)
+            .setPositiveButton(R.string.autocompleteRemoveItemRemove)
+            .setNegativeButton(R.string.autocompleteRemoveItemCancel)
+            .addEventListener(
+                object : TextAlertDialogBuilder.EventListener() {
+                    override fun onPositiveButtonClicked() {
+                        viewModel.onRemoveSearchSuggestionConfirmed(suggestion, omnibar.omnibarTextInput.text.toString())
```
CrisBarreiro, line 1577: Shouldn't we send back the suggestion instead?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1548,12 +1554,63 @@ class BrowserTabFragment :
...
+                    }
+                    override fun onDialogCancelled() {
+                        showKeyboardAndRestorePosition(autocompleteFirstVisibleItemPosition, autocompleteItemOffsetTop)
+                    }
+                },
+            )
+            .show()
+    }
+
+    private fun storeAutocompletePosition() {
```
CrisBarreiro, line 1590: I'm not a big fan of storing these in the fragment, but not sure whether it makes sense to store them in the viewmodel either...

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1548,12 +1554,63 @@ class BrowserTabFragment :
...
+        val keyboardVisibilityUtil = KeyboardVisibilityUtil(rootView)
+        keyboardVisibilityUtil.addKeyboardVisibilityListener {
+            scrollToPositionWithOffset(position, offset)
+        }
+        showKeyboard()
+    }
+
+    private fun scrollToPositionWithOffset(position: Int, offset: Int) {
+        val layoutManager = binding.autoCompleteSuggestionsList.layoutManager as LinearLayoutManager
+        layoutManager.scrollToPositionWithOffset(position, offset - 6.toPx())
```
CrisBarreiro, line 1611: Why the 6? Perhaps extract to a constant so we can give it some semantics

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -726,6 +727,37 @@ class BrowserTabViewModel @Inject constructor(
         }
     }
 
+    fun userLongPressedAutocomplete(suggestion: AutoCompleteSuggestion) {
+        when (suggestion) {
+            is AutoCompleteHistorySuggestion -> showRemoveSearchSuggestionDialog(suggestion)
+            is AutoCompleteHistorySearchSuggestion -> showRemoveSearchSuggestionDialog(suggestion)
```
CrisBarreiro, line 733: nit: Since the action is the same in both cases, you can do

` is AutoCompleteHistorySuggestion, is AutoCompleteHistorySearchSuggestion -> showRemoveSearchSuggestionDialog(suggestion)`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -726,6 +727,37 @@ class BrowserTabViewModel @Inject constructor(
...
+            else -> return
+        }
+    }
+
+    private fun showRemoveSearchSuggestionDialog(suggestion: AutoCompleteSuggestion) {
+        command.value = ShowRemoveSearchSuggestionDialog(suggestion)
```
CrisBarreiro, line 739: Even if not strictly needed, I believe we should post the command on the main thread for extra safety

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -726,6 +727,37 @@ class BrowserTabViewModel @Inject constructor(
...
+                        history.removeHistoryEntryByUrl(suggestion.url)
+                    }
+                    is AutoCompleteHistorySearchSuggestion -> {
+                        history.removeHistoryEntryByQuery(suggestion.phrase)
+                    }
+                    else -> {}
+                }
+            }
+            autoCompletePublishSubject.accept(omnibarText)
+            command.value = AutocompleteItemRemoved
```
CrisBarreiro, line 757: Launch in `dispatchers.io()`, then switch to `dispatchers.main` instead

app/src/main/java/com/duckduckgo/app/browser/autocomplete/KeyboardVisibilityUtil.kt
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.autocomplete
+
+import android.graphics.Rect
+import android.view.View
+import android.view.ViewTreeObserver
+
+class KeyboardVisibilityUtil(private val rootView: View) {
```
CrisBarreiro, line 23: Consider moving to `common-utils`

app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchActivity.kt
```diff
@@ -104,6 +111,9 @@ class SystemSearchActivity : DuckDuckGoActivity() {
     private lateinit var quickAccessAdapter: FavoritesQuickAccessAdapter
     private lateinit var itemTouchHelper: ItemTouchHelper
 
+    private var nestedScrollViewPosition: Int = 0
```
CrisBarreiro, line 114: Not to be addressed now, but do you think we could extract the autocomplete logic somehow so we don't need to duplicate it for BrowserTabFragment and SystemSearchActivity?

app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
```diff
@@ -102,15 +108,17 @@ class SystemSearchViewModel @Inject constructor(
...
+        data class ShowRemoveSearchSuggestionDialog(val suggestion: AutoCompleteSuggestion) : Command()
+        data object AutocompleteItemRemoved : Command()
     }
 
     val onboardingViewState: MutableLiveData<OnboardingViewState> = MutableLiveData()
     val resultsViewState: MutableLiveData<Suggestions> = MutableLiveData()
     val command: SingleLiveEvent<Command> = SingleLiveEvent()
 
-    private val resultsPublishSubject = PublishRelay.create<String>()
+    @VisibleForTesting
```
CrisBarreiro, line 120: I'd rather avoid this, if possible

history/history-impl/src/main/java/com/duckduckgo/history/impl/HistoryRepository.kt
```diff
@@ -98,6 +102,26 @@ class RealHistoryRepository(
         }
     }
 
+    override suspend fun removeHistoryEntryByUrl(url: String) {
+        withContext(dispatcherProvider.io()) {
+            historyDao.getHistoryEntryByUrl(url)?.let {
```
CrisBarreiro, line 107: We shouldn't need to query first and then delete. I think we should have a delete query specifying `DELETE FROM history_entries WHERE url = :url`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -726,6 +727,37 @@ class BrowserTabViewModel @Inject constructor(
...
+        command.value = ShowRemoveSearchSuggestionDialog(suggestion)
+    }
+
+    fun onRemoveSearchSuggestionConfirmed(suggestion: AutoCompleteSuggestion, omnibarText: String) {
+        appCoroutineScope.launch(dispatchers.main()) {
+            withContext(dispatchers.io()) {
+                pixel.fire(AUTOCOMPLETE_RESULT_DELETED)
+                when (suggestion) {
+                    is AutoCompleteHistorySuggestion -> {
+                        history.removeHistoryEntryByUrl(suggestion.url)
```
CrisBarreiro, line 748: We should be able to pass the entire suggestion and let the implementation decide which parameter to use as a key for deletion. That way, we can also more strictly enforce the 2 types of suggestions we allow to be deleted

CrisBarreiro (approved):

CrisBarreiro:
Left some comments but works as expected

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
joshliebe: I didn’t know you could do that, TIL 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
joshliebe: I thought about doing that but `AutoCompleteSuggestion` lives in the `app` module, so to do that we would need to add that dependency to `history` module.

I also don’t think that the `history` module should know what autocomplete is, so best to keep it agnostic.

app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchViewModel.kt
joshliebe: I don’t love this but it’s the only way I could test the `PublishSubject`.

app/src/main/java/com/duckduckgo/app/systemsearch/SystemSearchActivity.kt
joshliebe: I wanted to but since `SystemSearchActivity` uses a `NestedScrollView` and `BrowserTabFragment` uses a plain `RecyclerView`, I couldn’t easily share the logic.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
joshliebe: Yeah, that was my thinking… It’s view-related, so best to live in the `Fragment`. If we can extract the whole autocomplete component, then we can move it out of here.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
joshliebe: I need to send the omnibar text to refresh the list with whatever the user entered, if I only send the suggestion then it will update the omnibar text with the query they deleted, which is unexpected.

PR #4865, opened 2024-08-06
Tab manager crash fix

0nko:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1207981775109546/f
Possibly also: https://app.asana.com/0/1202552961248957/1207981775109549

### Description

This PR adds bounds checks for the tab getter in the RecyclerView adapter. The first crash is caused by a rapid double tap on the tab close button, which results in an invalid index being returned.

The second crash is possibly a different manifestation of the same problem but due to a race condition it happens elsewhere. It should be resolved by [8bfcfbd](https://github.com/duckduckgo/Android/commit/8bfcfbddca920ce5e8250e6987179f992c8c82f1).

### Steps to test this PR

_Duble tap_
- [x] Go to the tab manager 
- [x] Add a few tabs
- [x] Tap on a tab close button very quickly
- [x] Notice the app doesn’t crash


karlenDimla (approved):
Works as expected

PR #4871, opened 2024-08-07
Android ad click giving 'Web page not available' error

anikiki:
Task/Issue URL: https://app.asana.com/0/414730916066338/1207905708148686/f

### Description
Fixed ad click.

### Steps to test this PR
See task description for detailed steps.

### DIFF
--- a/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/trackingparameters/RealTrackingParameters.kt
+++ b/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/trackingparameters/RealTrackingParameters.kt
@@ -18,6 +18,7 @@ package com.duckduckgo.privacy.config.impl.features.trackingparameters
 
 import android.net.Uri
 import androidx.annotation.VisibleForTesting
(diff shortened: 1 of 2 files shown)


anikiki:
* **#4871** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4871?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @anikiki and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

joshliebe (approved):
Tested, this fixes the issue.

The new code was a bit hard to follow but this is due to how the existing code is structured. 

Once this is merged, I’ll add some unit tests and refactor this class in a follow-up PR.

joshliebe (changes requested):
I was testing this some more and it breaks this case:

`https://example.com?foo.com?tracking_param=value&other=value`

Before it would return:

`https://example.com?foo.com?other=value`

Since `foo.com` fails `isValid` it now returns `null` and we aren’t stripping that parameter anymore.

joshliebe (approved):

Issue #4869, opened 2024-08-07
[Bug] Last word of each line can jerk to the next line during animation in TypeAnimationTextView


bemusementpark:
### Describe the bug

During the onboarding animation 

[Screen_recording_20240807_211447.webm](https://github.com/user-attachments/assets/101df262-faa7-4d7f-be14-23aa43b2f44e)

### How to Reproduce

1. Delete the app or clear app storage
2. Open the app
3. Observe text on onboarding screens will sometimes jerk to the next line if a word becomes too long.

This is in contrast to DDG iOS which 

### Expected behavior

Although it is expected behaviour when you are actually typing, it looks glitchy in an animation, especially a fast one.

I would expect Android to behave as iOS does where each letter is progressively revealed in its right place.

### Environment

```markdown
- DDG App Version: commit ea0bb89a9acc17d7d547997a186725cb8cfb8486 HEAD -> develop, tag: 5.210.2.2-nightly
- Device: API 35
- OS: Android w/Playstore
```

marcosholgado:
This was fixed a while back with https://github.com/duckduckgo/Android/pull/4874 so closing.

PR #4874, opened 2024-08-07
Fix text glitches to new line in TypeAnimationTextView

bemusementpark:
Task: https://app.asana.com/0/1125189844152671/1208000645162922/f

Issue URL: https://github.com/duckduckgo/Android/issues/4869

### Description

There is a typing animation that adds each letter to the screen sequentially. The last word of each line can jerk to the next line if the word is long enough. On iOS the equivalent animation keeps characters in position as they are revealed. On Android this animation is more akin to what you would see when you are actually typing and you type a long word at the end of a line. The word jerks to the next line. This is fine when you are typing. But at speed an in an animation it does not look good. This PR copies iOS style, and applies 100% transparency to all letters that have not yet been "typed" in the animation.

### Steps to test this PR

1. Install a fresh copy of the app
2. Do onboarding

_Feature 1_
- [ Change TypeAnimationTextView to progressively reveal each character rather than progressively adding each character to the output sting ]

### UI changes

Before

[Screen_recording_20240807_211447.webm](https://github.com/user-attachments/assets/8e619a1f-5cc1-46ac-8080-e0d940aa3ed4)

After

[Screen_recording_20240808_021820.webm](https://github.com/user-attachments/assets/f49489f0-404e-41eb-af94-09f0bade0c9a)


marcosholgado (changes requested):
Thanks for your PR. There's a bug with the new changes, upon tapping on the TextView the full animation should finish and show the full text. Trying to that now it just stops the animation at the point where it is and the full text is never shown.

See video below:

https://github.com/user-attachments/assets/e387c311-7ff3-4f5e-bf7a-9268652f87a6

bemusementpark:
PTAL @marcosholgado

bemusementpark:
Changes are in (:

marcosholgado (approved):
Thanks @bemusementpark, LGTM

PR #4878, opened 2024-08-08
Tab manager crash fix

0nko:
Task/Issue URL: https://app.asana.com/0/1207418217763355/1208000508832396/f

### Description

This is the 2nd attempt at fixing the crash. I wasn’t able to reproduce it, but the theory is that the adapter list gets modified while it’s being diffed, which results in an inconsistency and `IndexOutOfBoundsException` being thrown.

The fix for this is to create a local copy of the diffed lists, which should resolve race condition issue. I also added index range checks to be extra careful even though they *should* be redundant. 

### Steps to test this PR

I wasn’t able to reproduce the crash. Smoke-testing the drag & drop feature should be enough.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
@@ -261,13 +261,7 @@ class TabSwitcherAdapter(
         }
     }
 
-    fun updateData(data: List<TabEntity>?) {
-        if (data != null) {
-            submitList(data)
-        }
-    }
-
-    private fun submitList(updatedList: List<TabEntity>) {
+    fun updateData(updatedList: List<TabEntity>) {
         val diffResult = DiffUtil.calculateDiff(TabEntityDiffCallback(list, updatedList))
 
         list.clear()
@@ -292,9 +286,7 @@ class TabSwitcherAdapter(
 
     fun onTabMoved(from: Int, to: Int) {
         val swapped = list.swap(from, to)
-        list.clear()
-        list.addAll(swapped)
-        notifyItemMoved(from, to)
+        updateData(swapped)
     }
 
     @SuppressLint("NotifyDataSetChanged")
--- a/app/src/main/java/com/duckduckgo/app/browser/tabpreview/TabEntityDiffCallback.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/tabpreview/TabEntityDiffCallback.kt
@@ -20,10 +20,11 @@ import android.os.Bundle
 import androidx.recyclerview.widget.DiffUtil
 import com.duckduckgo.app.tabs.model.TabEntity
 
-class TabEntityDiffCallback(
-    private val oldList: List<TabEntity>,
-    private val newList: List<TabEntity>,
-) : DiffUtil.Callback() {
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
```diff
@@ -292,9 +286,7 @@ class TabSwitcherAdapter(
 
     fun onTabMoved(from: Int, to: Int) {
         val swapped = list.swap(from, to)
-        list.clear()
-        list.addAll(swapped)
-        notifyItemMoved(from, to)
+        updateData(swapped)
```
0nko, line 289: We can let the `DiffUtil` to notify the adapter.

joshliebe (approved):
Adding those guards will fix the issue but may introduce unintended behavior (Why are those indexes out of bounds?). I’m fine with it but we should check feedback regarding tab behavior after releasing this.

Hopefully keeping local copies will fix the root cause.

Thanks for looking into this @0nko! 🙏 LGTM

0nko:
@mikescamell Actually, we did use `ListAdapter` and `DiffUtil.ItemCallback` previously and I refactored it to `DiffUtil.Callback` 😄. The reason for this was that the drag & drop behaviour was very choppy and inconsistent (https://app.asana.com/0/0/1207785323430731/1207801408540689/f). There was a [discussion on SO](https://stackoverflow.com/questions/68450742/add-drag-and-drop-on-recyclerview-with-diffutil) about this. I tried to make it work again today because avoiding list management would be ideal, but with no luck. 

If you’re interested, I pushed `ListAdapter` version to [pushed `ListAdapter` version to [ondrej/tab-diffutil-fix-listadapter](https://github.com/duckduckgo/Android/compare/fix/ondrej/tab-diffutil-fix-listadapter?expand=1).

mikescamell:
> @mikescamell Actually, we did use `ListAdapter` and `DiffUtil.ItemCallback` previously and I refactored it to `DiffUtil.Callback` 😄. The reason for this was that the drag & drop behaviour was very choppy and inconsistent (https://app.asana.com/0/0/1207785323430731/1207801408540689/f). There was a [discussion on SO](https://stackoverflow.com/questions/68450742/add-drag-and-drop-on-recyclerview-with-diffutil) about this. I tried to make it work again today because avoiding list management would be ideal, but with no luck.
> 
> If you’re interested, I pushed `ListAdapter` version to [ondrej/tab-diffutil-fix-listadapter](https://github.com/duckduckgo/Android/compare/fix/ondrej/tab-diffutil-fix-listadapter?expand=1).

Sad times 😢

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
bemusementpark: This is the fix. Because `move != swap`. Previous diff is only correct if swapping adjacent items, otherwise it would've needed:

```
val a = min(from, to)
val b = max(from, to)
notifyItemMoved(a, b)
if (b > a + 1) notifyItemMoved(b - 1, a)
```

But I think there's still a possible race condition from scheduling the `onMoveTab` handling:

PTAL https://github.com/duckduckgo/Android/pull/4881

bemusementpark:
I think there's still a possible race condition from scheduling the `onMoveTab` handling:

PTAL https://github.com/duckduckgo/Android/pull/4881

PR #4881, opened 2024-08-09
Fix Tab manager crash

bemusementpark:
Task/Issue URL: https://github.com/duckduckgo/Android/pull/4878

### Description
`onTabMoved` is scheduled to be run, so the `to` and/or `from` may be invalid by the time they are actually run. This PR removes that scheduling, so `onTabMoved` is called immediately in response to the tab being moved, and there is no time between scheduling and running this for the repository to emit and change the data by the time we handle this touch event.

---
  - https://app.asana.com/0/0/1207990892540430


0nko (approved):
Thank you for the PR, @bemusementpark! Your suggestion makes sense. 👍

Issue #3914, opened 2023-11-23
[Bug] Crashes every time on start


2011:
### Describe the bug

Every time I attempt to start DuckDuckGo, it crashes without ever opening.

### How to Reproduce

1. Tap on the DuckDuckGo icon
2. Get a message (in a few seconds) stating "DuckDuckGo has stopped Open app again"

### Expected behavior

I expect the application to run.

### Environment

```markdown
- DDG App Version: 5.177.1
- Device: Teclast T10 tablet
- OS: Android 7.0

The log file generated:


[11-23 10:46:15.428 691:707 I/ActivityManager]
START u0 {act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=com.duckduckgo.mobile.android/com.duckduckgo.app.launch.Launcher bnds=[906,360][1280,596]} from uid 10011 on display 0

[11-23 10:46:15.483 691:7815 I/ActivityManager]
Start proc 5660:com.duckduckgo.mobile.android/u0a90 for activity com.duckduckgo.mobile.android/com.duckduckgo.app.launch.Launcher

[11-23 10:46:16.254 691:7131 I/ActivityManager]
START u0 {cmp=com.duckduckgo.mobile.android/com.duckduckgo.app.onboarding.ui.OnboardingActivity} from uid 10090 on display 0

[11-23 10:46:16.611 5660:5708 I/WM-WorkerWrapper]
Work [ id=4678c155-7c56-4539-8148-f139e797b60d, tags={ DATA_REMOVAL_AD_CLICK_WORKER_TAG, com.duckduckgo.adclick.impl.DataRemovalAdClickWorker } ] was cancelled
java.util.concurrent.CancellationException: Task was cancelled.
        at androidx.work.impl.utils.futures.AbstractFuture.cancellationExceptionWithCause(AbstractFuture.java:1184)
        at androidx.work.impl.utils.futures.AbstractFuture.getDoneValue(AbstractFuture.java:514)
        at androidx.work.impl.utils.futures.AbstractFuture.get(AbstractFuture.java:475)
        at androidx.work.impl.WorkerWrapper$2.run(WorkerWrapper.java:311)
        at androidx.work.impl.utils.SerialExecutor$Task.run(SerialExecutor.java:91)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
[...]

PR #4888, opened 2024-08-11
Fix onboarding background - Canvas: trying to draw too large bitmap

bemusementpark:
Task/Issue URL: https://github.com/duckduckgo/Android/issues/3914

### Description

We were calling `View#setBackground` which is incorrect and inefficient because:

`android:scaleType="centerCrop"` applies to `src` of the `ImageView`, not the background of a `View`, which is set in `WelcomePage`

I could've changed this to `setImageResource`, but we can just set `src` in xml, and add xml resources for `light` and `night`

_This should fix the OOM, because setting the background probably set an upscaled version of this bitmap, so even though the large version of the bitmap is about 10MB, this device perhaps thinks it is an `xxhdpi` device and does an `x3` making this large bitmap require 100MB. That's probably what we see in the logs anyway. This PR will hopefully scale the bitmap properly so Canvas is not overwhelmed._

I've simplified the bitmap resolution by using the resource folders for each bitmap for each combination of `light | dark & < 600dp | > 600dp`. I just deleted the holder drawable files, because the only interesting thing they did was set `clampMode`, but that wasn't (intentionally) being used anyway, because the bitmap was usual too big, and if it was going to be used it probably wouldn't look very good depending on which edge is clamped. And there was a `scaleMode` set on the `ImageView` which was unused because no `src` was applied.

_I also changed `sceneBg` to `ImageView` in one override to match the other xml file, so that the binding will not be "downcast" to a `View`, but will be accessible as `ImageView`, which should help devs find the `ImageView` methods they need, and remove any desire for casting._

_If anyone doesn't know, `<ImageView>` will be inflated as `AppCompatImageView` if inflated in an `AppCompatActivity`._

### Steps to test this PR

Open the app in a low spec tablet.
Observe app does not crash.

### UI changes
| Before  | After |
| ------ | ----- |
!(Screenshot 2024-08-11 at 9 10 01 PM)|(Screenshot 2024-08-11 at 9 14 15 PM)
!(Screenshot 2024-08-11 at 10 02 00 PM)|(Screenshot 2024-08-11 at 9 14 15 PM))

...and the proper center-cropping kinda looks better too 🎉


malmstein:
This PR has been open for too long. We won’t take this into consideration for now.

bemusementpark:
Hey @malmstein this PR is unlikely to have conflicts and is likely to address this issue: https://github.com/duckduckgo/Android/issues/3914 which I assume is caused by opening these specific Bitmaps too large because 1. the tablet is big 2 it has low resolution, and then it crashes because of low specs.

This PR is also a nice little app-size improvement, as extraneous files are removed.

PR #4918, opened 2024-08-19
Tab manager scrolling fix

0nko:
Task/Issue URL: https://app.asana.com/0/488551667048375/1208062551083727/f

### Description

This PR fixes the initial scroll position after a tab manager is opened. The position should now be centered to the current tab.  

The problem was that the call to scroll to the selected tab was in the `render()` function, which would not always happen before the `LayoutManager` was initialized because the `LiveData` observers were called in different/inconsistent order. This was causing the tabs to sometimes show the first tab when initially opening the tab manager.

The PR also improves the scroll position when switching between the list mode and grid mode, which should now be pretty much centered around the same middle item.

### Steps to test this PR

I was able to reproduce this only when I had a lot of tabs open. If you don’t have many, you can add this to the `TabSwitcherViewModel`:

```kotlin
    init {
            if (tabRepository.flowTabs.first().size < 100) {
                repeat(100) {
                    tabRepository.add("cnn.com")
                }
            }
        }
    }
``` 

_Initial position_
- [x] Go to tab manager
- [x] Select a tab somewhere in the middle
- [x] Leave the tab manager
- [x] Go back
- [x] Notice the scroll position is centered on the selected tab

_Configuration change_
- [x] Go to tab manager
- [x] Select the first tab
- [x] Go back to the tab manager
- [x] Scroll away from the selected tab
- [x] Rotate the device
- [x] Notice the scroll position has not changed

_Switching view mode_
- [x] Go to tab manager
- [x] Scroll somewhere where you can identify the item in the middle
- [x] Change the view mode
- [x] Notice the previous item is still roughly in the middle
- [x] Switch the view mode again
- [x] Notice the position is roughly the same again

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
@@ -56,6 +56,7 @@ import com.duckduckgo.app.tabs.ui.TabSwitcherViewModel.Command.CloseAllTabsReque
 import com.duckduckgo.common.ui.DuckDuckGoActivity
 import com.duckduckgo.common.ui.view.dialog.TextAlertDialogBuilder
 import com.duckduckgo.common.ui.view.gone
+import com.duckduckgo.common.ui.view.hide
 import com.duckduckgo.common.ui.view.show
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.ActivityScope
@@ -234,7 +235,10 @@ class TabSwitcherActivity : DuckDuckGoActivity(), TabSwitcherListener, Coroutine
     }
 
     private fun updateLayoutType(layoutType: LayoutType) {
+        tabsRecycler.hide()
+
         val middleVisibleItem = getMiddleVisibleItem()
+
         this.layoutType = layoutType
         when (layoutType) {
             LayoutType.GRID -> {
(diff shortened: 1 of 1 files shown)


joshliebe (changes requested):
1. I tested this on a Pixel Fold, Galaxy Fold and OnePlus 8 Pro. On these devices it shifts the list down every time I switch between the view modes. It may be a problem with the calculation of the offset used in `scrollToPositionWithOffset(index, offset)`. 

I’ve attached a recording:

https://github.com/user-attachments/assets/3e076cad-b211-4931-a5be-51d26ccb4628

2. It works as expected on a Pixel 7 Pro and a Nexus 6P. However, the transition between the two modes is no longer smooth (It jumps before landing on its final state).

I’ve attached a recording:

https://github.com/user-attachments/assets/2047af99-bcd9-47ee-985f-4667b4413f0b

If we can fix those issues then I think this is a good improvement 👍

0nko:
Thanks for the review, @joshliebe! I’ve addressed both issues:

> On these devices it shifts the list down every time I switch between the view modes.

I changed the way the current scroll offset is calculated to be more precise and consistent and this shouldn’t happen anymore.

> It works as expected on a Pixel 7 Pro and a Nexus 6P. However, the transition between the two modes is no longer smooth (It jumps before landing on its final state).

I hide the `RecyclerView` until the layout is in its final state, which should solve this issue.

Ready for another round.

joshliebe (approved):
Tested and it’s working much better now, good job 🚀

PR #4936, opened 2024-08-26
Add onboarding exception for Privacy Pro site

nalcalag:
Task/Issue URL: https://app.asana.com/0/1201807753394693/1208142627859298/f

### Description
Implement change to not show onboarding dialogs in PrivacyPro site

### Steps to test this PR

- [ ] Fresh install
- [ ] Navigate to duckduckgo.com/pro
- [ ] Check onboarding in-context dialog is not shown
- [ ] Perform a search
- [ ] check SERP in-context dialog is shown

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
@@ -235,7 +235,7 @@ class CtaViewModel @Inject constructor(
         val nonNullSite = site ?: return null
 
         val host = nonNullSite.domain
-        if (host == null || userAllowListRepository.isDomainInUserAllowList(host)) {
+        if (host == null || userAllowListRepository.isDomainInUserAllowList(host) || isSiteNotAllowedForOnboarding(nonNullSite.url)) {
             return null
         }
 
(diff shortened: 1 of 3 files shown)


nalcalag:
* **#4936** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4936?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @nalcalag and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -283,6 +283,11 @@ class CtaViewModel @Inject constructor(
         }
     }
 
+    private fun isSiteNotAllowedForOnboarding(url: String?): Boolean {
+        val isPrivacyProSite = url == "https://duckduckgo.com/pro"
```
marcosholgado, line 287: This can still return false when it shouldn't, the URL can have arguments or more paths and then this won't be true. It should really be doing something similar to `shouldLaunchPrivacyProForUrl`, first check the eTLD is duckduckgo.com and then check the first path is pro.

marcosholgado (changes requested):

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
nalcalag: Thanks for the feedback @marcosholgado ! I'll commit the changes

CrisBarreiro: I don't think we should use that same method exactly, we probably want to skip onboarding even though a subscription is not active

nalcalag: Done. Thank you both :)

CrisBarreiro (approved):

PR #4960, opened 2024-08-30
Fix CompositeException

joshliebe:
Task/Issue URL: https://app.asana.com/0/488551667048375/1208160020676247/f

### Description
 - This fixes a `CompositeException` that is caused when an exception is thrown in `onError`.
 - To fix this, instead of throwing an exception, we return an empty observable instead.

### Steps to test this PR

- [x] Type and delete text in the omnibar.
- [x] Verify that the autocomplete results update as expected.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
+++ b/app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
@@ -227,7 +227,8 @@ class AutoCompleteApi @Inject constructor(
             .toObservable()
             .onErrorResumeNext { throwable: Throwable ->
                 if (throwable is InterruptedIOException) {
-                    // When interrupted, return an empty observable to avoid showing the default state
+                    // If the query text is deleted quickly, the request may be cancelled, resulting in an InterruptedIOException.
+                    // Return an empty observable to avoid showing the default state.
                     Observable.empty()
                 } else {
                     Observable.just(emptyList<AutoCompleteSearchSuggestion>())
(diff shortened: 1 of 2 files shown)


joshliebe:
* **#4960** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4960?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
```diff
@@ -224,10 +224,15 @@ class AutoCompleteApi @Inject constructor(
                 AutoCompleteSearchSuggestion(phrase = it.phrase, isUrl = (it.isNav ?: UriString.isWebUrl(it.phrase)))
             }
             .toList()
-            .onErrorReturn {
-                if (it is InterruptedIOException) throw it else emptyList<AutoCompleteSearchSuggestion>()
-            }
             .toObservable()
+            .onErrorResumeNext { throwable: Throwable ->
+                if (throwable is InterruptedIOException) {
+                    // When interrupted, return an empty observable to avoid showing the default state
```
CrisBarreiro, line 230: Can we explain a bit more what "interrupted" means for more clarity? i.e. quickly deleting all text which would cancel the subscription

CrisBarreiro (approved):
Left a small comment on the comment 😅 but other than that, looks good and works as expected

app/src/main/java/com/duckduckgo/app/autocomplete/api/AutoComplete.kt
joshliebe: Updated:
```
// If the query text is deleted quickly, the request may be cancelled, resulting in an InterruptedIOException.
// Return an empty observable to avoid showing the default state.
```

PR #4988, opened 2024-09-05
GHA: Create Release Task

malmstein:
Task/Issue URL: https://app.asana.com/0/1174433894299346/1207724126864064/f

### Description
Github Action that creates the task for Android Release

### Steps to test this PR
Example of working https://github.com/duckduckgo/Android/actions/runs/11015390483

More test steps in the Asana task


malmstein:
* **#4988** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/4988?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @malmstein and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

.github/workflows/release_create_task.yml
```diff
@@ -0,0 +1,54 @@
+name: Create Android App Release Task
+
+on:
+  workflow_dispatch:
+    inputs:
+      app-version:
+        description: 'App Version for Release'
+        required: true
+        default: '5.215.0'
```
CDRussell, line 9: why is this default here?

.github/workflows/release_create_task.yml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/checkout@v4
+
+      - name: Install Release Bridge from Homebrew
+        run: |
+          brew tap cdrussell/aarb
+          brew install aarb
+
+      - name: Create task in Asana
+        run: |
+          AndroidAsanaBridge version=${{ github.event.inputs.app-version }} action=createRelease,tagPendingTasks,addLinksToDescription
```
CDRussell, line 34: why doesn't it also invoke `removePendingTasks` ?

.github/actions/assign-release-task/action.yml
```diff
@@ -0,0 +1,67 @@
...
+  steps:
+    - name: Find task in Asana and assign it to owner
+      shell: bash
+      run: |
+        task_name="${{ inputs.task_name }}"
+        asana_token="${{ inputs.asana_token }}"
+        project_gid="${{ inputs.project_gid }}"
+        username="${{ inputs.username }}"
+
+        # Make the API request to get tasks from the specified project
```
CDRussell, line 27: `aarb` already knows the release task ID so we could probably work out a way so that it returns it directly, and save the extra search on the Asana API based on the task name. 

but could be filed away as a future improvement assuming the searching approach will work fine.

malmstein: Yes, that’s a good improvement. For now, I’d prefer to keep it as is and file an improvement to aarb at a later stage.

.github/workflows/release_create_task.yml
malmstein: workflow inputs need a default value, so I figured we could have a value close to what we have now, so it’s easy to edit. would you have a different default value?

.github/workflows/release_create_task.yml
malmstein: good catch, that needs changing too.

.github/workflows/release_create_task.yml
CDRussell: If I'm understanding correctly, the default value should never be used, right? But just has to have some default sadly. 

I'd be tempted to make it obviously wrong instead of it being close to being right. E.g., "default=PLACEHOLDER", and throw an exception if it encounters the default value?

malmstein: That’s fair, I like that.

CDRussell (approved):

PR #5006, opened 2024-09-13
Fix database access on main thread that caused a crash

CrisBarreiro:
Task/Issue URL: https://app.asana.com/0/1202552961248957/1208294040908744/f

### Steps to test this PR

_Feature 1_
- [x] Open https://www.disneyaulani.com/ 
- [x] Check app doesn't crash and works normally
- [x] Open maps.google.com and click on the location button
- [x] Check app doesn't crash and works normally

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -3167,7 +3167,7 @@ class BrowserTabViewModel @Inject constructor(
         )
     }
 
-    suspend fun processJsCallbackMessage(
+    fun processJsCallbackMessage(
         featureName: String,
         method: String,
         id: String?,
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3224,21 +3226,23 @@ class BrowserTabViewModel @Inject constructor(
         }
     }
 
-    private fun permissionsQuery(
+    private suspend fun permissionsQuery(
```
0nko, line 3229: You’re launching a new coroutine inside this function, so it doesn’t need to be a suspended function. Also, the call to `permissionQuery()` is from within _another_ coroutine launcher (which is also the same dispatcher as here). I think you want to use one or the other. WDYT?

0nko (commented):
The app behaves as expected but I left a comment in the code because I think some coroutine launchers are redundant.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
CrisBarreiro: Fixed, I missed that on my second commit, thanks!

0nko (approved):
Nice work! :shipit:

PR #5033, opened 2024-09-18
Fix keyboard crashes in BrowserFragment

mikescamell:
Task/Issue URL: https://app.asana.com/0/414730916066338/1208339084893367/f 

### Description

We saw some keyboard crashes due to some recent refactoring of the Omnibar.

The Omnibar was previously solely bound in onActivityCreated.

It was never nulled out in `onDestroyView` so the Fragment held on to it, and when the delay concluded it would happily try to call show/hideKeyboard using the Omnibar without issue.

Now, we’re using the FragmentViewBindingDelegate for the Omnibar, which gets nulled out in onDestroy, so when the delay concluded we tried calling show/hideKeyboard on a null object. 

To fix this we’re introducing new extension functions that follow the [recommended way](https://developer.android.com/develop/ui/views/touch-and-input/keyboard-input/visibility#kotlin) to show and hide the keyboard. 

### Steps to test this PR

_Run the following tests locally they should all pass_
- [x] `whenProtectionsAreDisabledRequestAreNotBlocked`
- [x] `whenProtectionsAreDisabledSurrogatesAreNotLoaded`
- [x] `whenProtectionsAreEnabledHttpsUpgradedCorrectly`
- [x] `whenProtectionsAreEnabledRequestBlockedCorrectly`
- [x] `whenProtectionsAreEnabledSurrogatesAreLoaded`
- [x] `whenProtectionsAreFingerprintProtected`

_Omnibar Keyboard_
- [x] Ensure keyboard shows when opening app with New Tab Page
- [x] Generally play around with the Omnibar EditText and ensure keyboard is shown and hidden where appropriate

### UI changes

N/A

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -268,7 +268,9 @@ import com.duckduckgo.common.utils.ConflatedJob
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.common.utils.FragmentViewModelFactory
 import com.duckduckgo.common.utils.KeyboardVisibilityUtil
+import com.duckduckgo.common.utils.extensions.hideKeyboard
 import com.duckduckgo.common.utils.extensions.html
+import com.duckduckgo.common.utils.extensions.showKeyboard
 import com.duckduckgo.common.utils.extensions.toBinaryString
 import com.duckduckgo.common.utils.extensions.websiteFromGeoLocationsApiOrigin
 import com.duckduckgo.common.utils.extractDomain
@@ -3080,7 +3082,7 @@ class BrowserTabFragment :
     private fun hideKeyboard() {
         if (!isHidden) {
             Timber.v("Keyboard now hiding")
-            binding.legacyOmnibar.omnibarTextInput.postDelayed(KEYBOARD_DELAY) { binding.legacyOmnibar.omnibarTextInput?.hideKeyboard() }
+            hideKeyboard(binding.legacyOmnibar.omnibarTextInput)
             binding.focusDummy.requestFocus()
             binding.legacyOmnibar.omniBarContainer.isPressed = false
         }
@@ -3096,8 +3098,10 @@ class BrowserTabFragment :
     private fun showKeyboard() {
         if (!isHidden) {
             Timber.v("Keyboard now showing")
-            binding.legacyOmnibar.omnibarTextInput.postDelayed(KEYBOARD_DELAY) { binding.legacyOmnibar.omnibarTextInput?.showKeyboard() }
-            binding.legacyOmnibar.omniBarContainer.isPressed = true
+            with(binding.legacyOmnibar.omnibarTextInput) {
+                showKeyboard(this)
+                isPressed = true
+            }
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -3098,10 +3098,8 @@ class BrowserTabFragment :
     private fun showKeyboard() {
         if (!isHidden) {
             Timber.v("Keyboard now showing")
-            with(binding.legacyOmnibar.omnibarTextInput) {
-                showKeyboard(this)
-                isPressed = true
-            }
+            showKeyboard(binding.legacyOmnibar.omnibarTextInput)
+            binding.legacyOmnibar.omniBarContainer.isPressed = true
         }
     }
 
--- a/common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
+++ b/common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
@@ -21,4 +21,4 @@ import androidx.fragment.app.Fragment
 
 fun Fragment.showKeyboard(editText: EditText) = requireActivity().showKeyboard(editText)
 
-fun Fragment.hideKeyboard(editText: EditText) = requireActivity().showKeyboard(editText)
+fun Fragment.hideKeyboard(editText: EditText) = requireActivity().hideKeyboard(editText)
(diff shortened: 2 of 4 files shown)


mikescamell:
* **#5033** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5033?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @mikescamell and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package com.duckduckgo.common.utils.extensions
+
+import android.widget.EditText
+import androidx.fragment.app.Fragment
+
+fun Fragment.showKeyboard(editText: EditText) = requireActivity().showKeyboard(editText)
+
+fun Fragment.hideKeyboard(editText: EditText) = requireActivity().showKeyboard(editText)
```
joshliebe, line 24: This should call `hideKeyboard`, not `showKeyboard`

common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.common.utils.extensions
+
+import android.widget.EditText
+import androidx.fragment.app.Fragment
+
+fun Fragment.showKeyboard(editText: EditText) = requireActivity().showKeyboard(editText)
```
joshliebe, line 22: We should use `activity?.` instead of `requireActivity` for keyboard operations.

The `Activity` can be `null` when the `Fragment` has been detached, and for this case we don’t care if the keyboard shows or not, it’s about to be destroyed.

The user could bring up the keyboard at the wrong time during the lifecycle or it could be shown in an async operation after the `Activity` is destroyed, both of these scenarios would result in a crash.

So I think we should do something like this instead:

```suggestion
fun Fragment.showKeyboard(editText: EditText) {
    activity?.showKeyboard(editText) ?: Timber.d("FragmentExtensions", "Activity is null, cannot show keyboard")
}
```

This way we still get info but it won’t crash the app for edge cases.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -3096,8 +3098,10 @@ class BrowserTabFragment :
     private fun showKeyboard() {
         if (!isHidden) {
             Timber.v("Keyboard now showing")
-            binding.legacyOmnibar.omnibarTextInput.postDelayed(KEYBOARD_DELAY) { binding.legacyOmnibar.omnibarTextInput?.showKeyboard() }
-            binding.legacyOmnibar.omniBarContainer.isPressed = true
+            with(binding.legacyOmnibar.omnibarTextInput) {
+                showKeyboard(this)
+                isPressed = true
```
malmstein, line 3103: it’s not the omnibarTextInput that has the isPressed state as true. 

```suggestion
                binding.legacyOmnibar.omniBarContainer.isPressed = true
```

malmstein (changes requested):
Like this it’s removing the outline of the Omnibar.

common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
mikescamell: Fixed! Thanks 🙏

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
mikescamell: Fixed! Thanks 🙌

common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
mikescamell: Personally, I don’t like defensive programming like this. I’d rather fail fast and know this is not working properly, because if this is failing it’s probably a result of someone calling show/hideKeyboard in the wrong part of the Fragment lifecycle.

I’d worry a log statement might be easily missed by someone if they are using this and it fails, is there any way we can send an error to our crash reporting tool?

malmstein (approved):
Looks good! I let you agree with @joshliebe in the outstanding question, but I don’t consider that a blocker.

common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
joshliebe: > is there any way we can send an error to our crash reporting tool?

We won't receive those in the crash tool, only in the logs.

> probably a result of someone calling show/hideKeyboard in the wrong part of the Fragment lifecycle

Not necessarily, the user can initiate `showKeyboard`/`hideKeyboard` while the `Fragment` is being destroyed. So unless we can always insure that this doesn’t happen, I won't be able to approve the change.

Also, we don't care if the keyboard hides/shows when `Activity` = `null`, so it’s not worth crashing the app over. Show/hide keyboard is not a particularly important operation.

joshliebe (changes requested):
I’m gonna have to disagree with @malmstein on this one, blocking on: https://github.com/duckduckgo/Android/pull/5033#discussion_r1766446812

common/common-utils/src/main/java/com/duckduckgo/common/utils/extensions/FragmentExtensions.kt
mikescamell: As we talked about on MM, I disagree with this approach. If a developer is calling showKeyboard without an Activity then something is wrong and they should know about it. The logging statement for me could easily go unnoticed. I appreciate this is more of an obvious situation in that the keyboard should show and they probably would notice, but in general I like to fail fast and early, and know about it.

For the sake of getting this fixed and moving on I’ll make the changes 👍

joshliebe (approved):
The current issues seem to be unrelated to this PR, thanks for fixing this 👍🙂

PR #5139, opened 2024-10-15
Send enrollment pixel when cohort is assigned

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1208490557030192/f 

### Description
This PR sends an enrollment pixel whenever a cohort is assigned

### Steps to test this PR
See task


feature-toggles/feature-toggles-api/build.gradle
```diff
@@ -32,6 +32,7 @@ kotlin {
 
 dependencies {
     api project(":experiments-api")
+    api project(":feature-toggles-internal-api")
```
aitorvs, line 35: I think this should be implementation, else anybody depending on `feature-toggles-api` will see the internal one as transitive dependency. It's only meant to be an API between feature-toggles-api and feature-toggles-impl

aitorvs (approved):
LGTM, tested as per task and all works as specified

PR #5149, opened 2024-10-17
Omnibar: Url alignment not focused mode

malmstein:
Task/Issue URL: 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


malmstein:
* **#5149** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5149?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @malmstein and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

app/src/main/java/com/duckduckgo/app/browser/viewstate/OmnibarViewState.kt
```diff
@@ -20,6 +20,7 @@ data class OmnibarViewState(
     val omnibarText: String = "",
     val isEditing: Boolean = false,
     val shouldMoveCaretToEnd: Boolean = false,
+    val shouldMoveCaretToStart: Boolean = false,
```
mikescamell, line 23: 🔍 These properties feel odd to me as ViewState. I’d imagine a property for the caret position would be more like an index or enum. 

e.g.

```
    enum class CaretPosition {
        START,
        END,
    }
```

Things get iffy with EditTexts I realised, but these read more like events rather than state.

mikescamell (approved):
LGTM. One comment but nothing stopping merging 👍

app/src/main/java/com/duckduckgo/app/browser/viewstate/OmnibarViewState.kt
malmstein: Not a bad idea! I’m going to add them to the `OmnibarLayoutViewModel`, but leave the old `OmnibarViewState` as is.

malmstein: Actually no, this is not one or the other. Both states can be false, so we can’t use an enum for this.

mikescamell: Would it make sense to have another option which is UNSET in that case?

```
    enum class CaretPosition {
        UNSET,
        START,
        END,
    }
```

PR #5182, opened 2024-10-24
Fix the OmnibarLayout crash

0nko:
Task/Issue URL: https://app.asana.com/0/488551667048375/1208618045189439/f

### Description

We need to make sure all attempts to access the `OmnibarLayout`’s ViewModel are checked to make sure the window is attached, otherwise the view may not be injected yet.

### Steps to test this PR

- [ ] Enable _Don’t keep activities_ option in Developer options
- [ ] Clear the app storage and start the app
- [ ] Go through the flow until the omnibar is visible
- [ ] Switch to a different app
- [ ] Come back to the app
- [ ] Notice the app doesn’t crash
- [ ] Go to settings and change the omnibar position
- [ ] Navigate back to the browser
- [ ] Notice the app doesn’t crash


malmstein (approved):
:shipit:

PR #5223, opened 2024-10-31
Add module for JWT validation and parsing

lmac012:
Task/Issue URL: https://app.asana.com/0/1205648422731273/1208669762068505/f

### Description

This PR adds new module that will be responsible for parsing and validating JWT tokens issued by auth API v2.

### Steps to test this PR

There are no usages of the new module yet, so nothing to test.


lmac012:
* **#5223** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5223?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

 Join @lmac012 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

auth-jwt/auth-jwt-impl/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+    anvil project(path: ':anvil-compiler')
+    implementation project(path: ':anvil-annotations')
+    implementation project(path: ':di')
+    implementation project(path: ':common-utils')
+
+    implementation Google.dagger
+
+    implementation "com.squareup.logcat:logcat:_"
+
+    implementation 'io.jsonwebtoken:jjwt-api:0.12.6'
```
karlenDimla, line 38: @lmac012  why are we specifying the version here?

auth-jwt/auth-jwt-api/src/main/java/com/duckduckgo/authjwt/api/AuthJwtValidator.kt
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+package com.duckduckgo.authjwt.api
+
+import java.time.Instant
+
+/**
+ * This class provides methods for checking the validity of JWTs obtained from Auth API V2 and parsing claims included in those tokens.
+ */
+interface AuthJwtValidator {
```
karlenDimla, line 24: @lmac012 curious question, why separate this module from subscription-impl? As much as I see the APIs are very well defined, I don’t see the need of separating it from subscription-impl.

auth-jwt/auth-jwt-impl/build.gradle
lmac012: I forgot to remove the version. Fixed in [ebfa8df](https://github.com/duckduckgo/Android/pull/5223/commits/ebfa8dfde9a1a738f3136afc9d28da743eeb64c3).

auth-jwt/auth-jwt-api/src/main/java/com/duckduckgo/authjwt/api/AuthJwtValidator.kt
lmac012: The main reason is to encapsulate the third-party library used for validating JWTs. This approach ensures that types defined by the third-party library aren’t used throughout the `subscriptions-impl` module. If we ever need to migrate to a different library, it won't require a lot of refactoring. Putting this code in its own module is a suggestion I received in a TD conversation, and while I don’t feel strongly about it, I do see some value in this approach.

karlenDimla (approved):

PR #5265, opened 2024-11-12
Separate disabled state from non-editable state for DaxTextInput

CDRussell:
Task/Issue URL: https://app.asana.com/0/488551667048375/1208740714999313/f 

### Description
Separates non-editable state from disabled state of `DaxTextInput`. An input can be:
- non-editable but enabled (i.e., read-only text that isn't greyed out, with buttons in the view still working)
- disabled (i.e., non-editable and greyed out, and all buttons disabled too)

### Steps to test this PR

**Passwords**
- [x] Go to Passwords and tap ➕ to manually start adding a password
- [x] Verify the input fields behave correctly
- [x] Save the password; verify the fields correctly go to read-only mode (non-editable state), and that the buttons on the fields work (like copy password, reveal password etc...)
- [x] Edit the password again and verify the fields correctly go to editable state

**App launch setting**
- [x] Go to `Settings` -> `General` -> `Show on App Launch`
- [x] Verify the URL is greyed out and non-editable when it is not selected
- [x] Select it, and verify the URL is no longer greyed out, and is now editable
- [x] Unselect it again and make sure that works correctly

**Anything else that might be affected?**

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/generalsettings/showonapplaunch/ShowOnAppLaunchActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/generalsettings/showonapplaunch/ShowOnAppLaunchActivity.kt
@@ -110,7 +110,7 @@ class ShowOnAppLaunchActivity : DuckDuckGoActivity() {
                         uncheckNewTabCheckListItem()
                         with(binding) {
                             specificPageCheckListItem.setChecked(true)
-                            specificPageUrlInput.isEditable = true
+                            specificPageUrlInput.isEnabled = true
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


CDRussell:
* **#5265** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

malmstein (changes requested):
If we want to use the disabled state, it should also work when the value is set in the XML directly. At the moment it only works if the value is set explicitly by code.
Screenshot_20241113_115901

malmstein (approved):
I’ve added the attribute so it can read from inflation. Rest looks good!

PR #5278, opened 2024-11-15
Pixel: Add tab activity stats to a daily pixel

0nko:
Task/Issue URL: https://app.asana.com/0/1207418217763355/1208414748460831/f

### Description

This PR adds the following [parameters](https://github.com/duckduckgo/Android/compare/feature/ondrej/tab-activity-stats?expand=1#diff-2f21d70376d40ee5085604ef2646239f16abe5dea9268b0010443b5ca39dc5c0R2804-R2811) to the `m_tab_manager_clicked_daily` pixel:

- `tab_count`: The number of open tabs
- `tab_active_7d`: The number of tabs access within the last 7 days
- `tab_inactive_1w`: Tabs not accessed for more than a week, but less than 2 weeks
- `tab_inactive_2w`: Tabs not accessed for more than 2 weeks, but less than 3 weeks
- `tab_inactive_3w`: Tabs not accessed for at least 3 weeks

All stats are reported within [buckets](https://github.com/duckduckgo/Android/compare/feature/ondrej/tab-activity-stats?expand=1#diff-ff2961602b32ef21dbf3199564d1e19aa9f84b0a67590a1c2f768d5fd2359242) due to privacy reasons: 

image

A new column was added to the `TabEntity` — the [`lastAccessTime`](https://github.com/duckduckgo/Android/compare/feature/ondrej/tab-activity-stats?expand=1#diff-61f346afb2e82d2efa34c2a444fda1886e4cfefb1434c69b7475ff65cfd12ba2R52), which is updated [every time a tab is activated](https://github.com/duckduckgo/Android/compare/feature/ondrej/tab-activity-stats?expand=1#diff-53f6cb0433b1d43092e96b2b95aa39158ce4fb1536a91a17540da459a5835c0eR275).

### Steps to test this PR

_DB migration_
- [x] Run the app with existing data with a previous DB version
- [x] Verify the migration to version 56 succeeds (no crashes)

_Pixel firing_
- [x] Run the app
- [x] Open the tab switcher
- [x] Verify the `m_tab_manager_clicked_daily` pixel is fired with all of the parameters mentioned above

_Last access time updates_
- [x] Run the app and access a couple of different tabs
- [x] Open the App Inspector in AS
- [x] Select the DDG process
- [x] Select the `app.db` database
- [x] Select the `tabs` table
- [x] Verify the `lastAccessTime` was updated for the access tabs using correct time


app/src/androidTest/java/com/duckduckgo/app/tabs/store/TabStatsBucketingTest.kt
```diff
@@ -0,0 +1,313 @@
...
+
+import com.duckduckgo.app.tabs.model.TabRepository
+import kotlinx.coroutines.runBlocking
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mockito.mock
+import org.mockito.kotlin.whenever
+
+class DefaultTabStatsBucketingTest {
```
karlenDimla, line 26: This doesn’t have to be an Android Test, we can move it to unitTest instead and run it as such.

app/src/androidTest/java/com/duckduckgo/app/tabs/store/TabStatsBucketingTest.kt
```diff
@@ -0,0 +1,313 @@
...
+
+    private lateinit var defaultTabStatsBucketing: DefaultTabStatsBucketing
+
+    @Before
+    fun setup() {
+        defaultTabStatsBucketing = DefaultTabStatsBucketing(tabRepository)
+    }
+
+    @Test
+    fun testGetNumberOfOpenTabsExactly1() = runBlocking {
```
karlenDimla, line 38: Instead of `runBlocking`, `runTest` is better to use in tests.

app/src/main/java/com/duckduckgo/app/tabs/store/TabStatsBucketing.kt
```diff
@@ -0,0 +1,112 @@
...
+            0..0,
+            1..5,
+            6..10,
+            11..20,
+            21..Int.MAX_VALUE,
+        )
+    }
+}
+
+@ContributesBinding(AppScope::class)
```
karlenDimla, line 67: Curious q, does it have to be `AppScope`?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2825,7 +2829,26 @@ class BrowserTabViewModel @Inject constructor(
...
+    }
+
+    private fun fireDailyLaunchPixel() {
+        val tabCount = viewModelScope.async(dispatchers.io()) { tabStatsBucketing.getNumberOfOpenTabs() }
+        val activeTabCount = viewModelScope.async(dispatchers.io()) { tabStatsBucketing.getTabsActiveLastWeek() }
+        val inactive1w = viewModelScope.async(dispatchers.io()) { tabStatsBucketing.getTabsActiveOneWeekAgo() }
+        val inactive2w = viewModelScope.async(dispatchers.io()) { tabStatsBucketing.getTabsActiveTwoWeeksAgo() }
+        val inactive3w = viewModelScope.async(dispatchers.io()) { tabStatsBucketing.getTabsActiveMoreThanThreeWeeksAgo() }
+
+        viewModelScope.launch {
```
karlenDimla, line 2842: Would it make sense to run this on io as well, since we have a bunch of await() calls here?

karlenDimla (commented):
Tested and works as expected. A few comments that would be good to clarify before merging.

app/src/androidTest/java/com/duckduckgo/app/tabs/store/TabStatsBucketingTest.kt
0nko: Good point!

karlenDimla (approved):

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
0nko: Sure.

app/src/main/java/com/duckduckgo/app/tabs/store/TabStatsBucketing.kt
0nko: Not really, it could have been `FragmentScope`, as well, but since the only dependency here is the `TabManager` which is also scoped in `AppScope`, this allows the class to be used from anywhere.

0nko:
@karlenDimla I’ve addressed your comments in the code.

PR #5280, opened 2024-11-15
Add simplified ("toggle") broken site reporting

laghee:
Task/Issue URL: https://app.asana.com/0/1201870266890790/1208769785329643/f

### Description
We're implementing the simplified broken site reporting flow that already exists on other platforms here, prompting users to report site breakage on URLs where they toggle off protections.

### Steps to test this PR

_Toggle Reports_
- [x] Set up your local environment by checking out the [privacy-dashboard branch](https://github.com/duckduckgo/privacy-dashboard/tree/pr-releases/pr-196) and editing `app/build.gradle` to point to its location ([replacing this bit](https://github.com/duckduckgo/Android/blob/4677ba3e51108db4d4f20071977c73baaaa5bb40/app/build.gradle#L44)).
- [x] Navigate to any non-DuckDuckGo site and tap the privacy shield, then toggle off protections
   - [x] Confirm that you're shown the report prompting screen
   - [x] Click on "See what's sent" and confirm the data details expand into view
   - [x] Tap "Send report" and confirm that the `m_protection-toggled-off-breakage-report` pixel is sent with `reportFlow=on_protections_off_dashboard_main`
- [x] Re-enable protections
- [x] Tap the 3-dot menu and then "Disable privacy protection"
   - [x] Confirm that you're again shown the prompt screen
   - [x] Tap "Send report" and confirm that the above pixel is sent with `reportFlow=on_protections_off_menu`
- [x] Re-enable protections
- [x] Toggle off protections again from either entry point and confirm that you are again show the prompt
   - [x] Tap "Don't send"
   - [x] Confirm that you are no longer shown the prompt from either toggle-off point (block should last 48 hours, but that value can be overridden by the remote privacy config -- there is [an existing JSONblob](https://jsonblob.com/api/1306645031426056192) with both intervals set to 2 minutes if you'd like to test more exhaustively, as I did)


privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsFeatureSettingsStore.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.launch
+import timber.log.Timber
+
+@ContributesBinding(AppScope::class)
+@RemoteFeatureStoreNamed(ToggleReportsFeature::class)
+class ToggleReportsFeatureSettingsStore @Inject constructor(
+    private val toggleReportsDataStore: ToggleReportsDataStore,
+    @AppCoroutineScope private val coroutineScope: CoroutineScope,
+    private val dispatcherProvider: DispatcherProvider,
+) : FeatureSettings.Store {
```
lmac012, line 41: This API has been recently deprecated. You can now access feature settings using `Toggle::getSettings()` - can you migrate to the new API? I'm assuming this shouldn't require much effort, but If I'm wrong then it's fine to leave it as is.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ui/PrivacyDashboardRenderer.kt
```diff
@@ -37,11 +37,16 @@ class PrivacyDashboardRenderer(
...
+    private val onSendToggleReport: () -> Unit,
+    private val onRejectToggleReport: () -> Unit,
+    private val onSeeWhatIsSent: () -> Unit,
+    private val onShowNativeFeedback: () -> Unit,
 ) {
 
     private var lastSeenPrivacyDashboardViewState: ViewState? = null
 
-    fun loadDashboard(webView: WebView, initialScreen: InitialScreen) {
+    fun loadDashboard(webView: WebView, initialScreen: InitialScreen, toggleOpener: String) {
```
lmac012, line 49: nit: looks like there is a set of predefined values for `toggleOpener` - it could be an enum instead of a string.

privacy-dashboard/privacy-dashboard-impl/src/test/java/ToggleReportsDataStoreTest.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    fun whenDatabaseIsEmptyThenReturnsDefaultJson() = runTest {
+        assertEquals("{}", testee.getToggleReportsRemoteConfigJson())
+    }
+
+    @Test
+    fun whenJsonIsSetThenReturnsCorrectValue() = runTest {
+        val testJson = """{"test": "value"}"""
+        testee.setToggleReportsRemoteConfigJson(testJson)
```
lmac012, line 65: `setToggleReportsRemoteConfigJson` and `getToggleReportsRemoteConfigJson` methods appear to have no usages outside their unit tests - can we remove them?

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
```diff
@@ -0,0 +1,243 @@
...
+        val TOGGLE_REPORTS_PROMPT_LIMIT_LOGIC_ENABLED = booleanPreferencesKey(name = "TOGGLE_REPORTS_PROMPT_LIMIT_LOGIC_ENABLED")
+        val TOGGLE_REPORTS_DISMISS_LOGIC_ENABLED = booleanPreferencesKey(name = "TOGGLE_REPORTS_DISMISS_LOGIC_ENABLED")
+        val TOGGLE_REPORTS_SENT = stringSetPreferencesKey(name = "TOGGLE_REPORTS_SENT")
+        val TOGGLE_REPORTS_PROMPTS_DISMISSED = stringSetPreferencesKey(name = "TOGGLE_REPORTS_PROMPTS_DISMISSED")
+        val TOGGLE_REPORTS_DISMISS_INTERVAL = intPreferencesKey(name = "TOGGLE_REPORTS_DISMISS_INTERVAL")
+        val TOGGLE_REPORTS_PROMPT_INTERVAL = intPreferencesKey(name = "TOGGLE_REPORTS_PROMPT_INTERVAL")
+        val TOGGLE_REPORTS_MAX_PROMPT_COUNT = intPreferencesKey(name = "TOGGLE_REPORTS_MAX_PROMPT_COUNT")
+    }
+
+    private val toggleReportsRC: StateFlow<String> = store.data
```
lmac012, line 88: Is there a reason we load every property eagerly into a StateFlow (e.g. a performance issue you observed)? Have you considered making `canPrompt()` a `suspend fun` an reading from the data store only when needed?

laghee: Honestly, I was mostly modeling after duck player because it was the most recent module with a datastore, but it is effectively a lot of overhead for probably marginal benefit so your suggestion makes sense to me.

privacy-dashboard/privacy-dashboard-impl/build.gradle
```diff
@@ -27,6 +28,7 @@ android {
 }
 
 dependencies {
+    implementation 'androidx.datastore:datastore-core-android:1.1.1'
```
lmac012, line 31: We don't keep versions of dependencies in module-level `build.gradle` files, but this line can be removed entirely since you've also added `api AndroidX.dataStore.preferences`

privacy-dashboard/privacy-dashboard-impl/build.gradle
```diff
@@ -42,13 +44,20 @@ dependencies {
...
     implementation AndroidX.lifecycle.runtime.ktx
     implementation AndroidX.lifecycle.viewModelKtx
     implementation Google.android.material
 
+    // Room
+    implementation AndroidX.room.runtime
+    implementation AndroidX.room.rxJava2
+    implementation AndroidX.room.ktx
+    ksp AndroidX.room.compiler
+
```
lmac012, line 60: Unless I missed something, you're not using Room database anywhere, so these dependencies are redundant.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    override suspend fun storeDismissLogicEnabled(dismissLogicEnabled: Boolean) {
+        store.edit { prefs ->
+            prefs[TOGGLE_REPORTS_DISMISS_LOGIC_ENABLED] = dismissLogicEnabled
+        }
+    }
+
+    override suspend fun getDismissLogicEnabled(): Boolean {
+        return store.data.first()[TOGGLE_REPORTS_DISMISS_LOGIC_ENABLED] ?: false
```
aitorvs, line 108: This will throw an exception if there's no element, should we use `firstOrNull` instead?

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsFeature.kt
```diff
@@ -0,0 +1,35 @@
...
+    settingsStore = ToggleReportsFeatureSettingsStore::class,
+)
+interface ToggleReportsFeature {
+    @Toggle.DefaultValue(false)
+    fun self(): Toggle
+}
+
+fun ToggleReportsFeature.isEnabled(): Boolean = self().isEnabled()
+
+fun ToggleReportsFeature.getSettings(): String? = self().getSettings()
```
aitorvs, line 35: We should not do this, it's dangerous. If we ever add a sub-feature to `toggleReports` we'll likely miss that calling isEnabled() on the sub-feature actually calls the top-level feature.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
aitorvs, line 1: I think all implementation of suspend methods should already select the `dispatcher.io()`

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
laghee: We don't need to store these anymore, but I updated the two remaining usages.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsFeature.kt
laghee: Removed as part of migration to remove settingsstore

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
laghee: Done in migration

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
```diff
@@ -0,0 +1,212 @@
...
+    val dismissInterval: Int,
+    @field:Json(name = "promptLimitLogicEnabled")
+    val promptLimitLogicEnabled: Boolean,
+    @field:Json(name = "promptInterval")
+    val promptInterval: Int,
+    @field:Json(name = "maxPromptCount")
+    val maxPromptCount: Int,
+)
+
+@RemoteFeatureStoreNamed(ToggleReportsFeature::class)
```
aitorvs, line 77: Remove, otherwise it won't build
```suggestion
```

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
```diff
@@ -0,0 +1,212 @@
...
+package com.duckduckgo.privacy.dashboard.impl
+
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.stringSetPreferencesKey
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.common.utils.formatters.time.DatabaseDateFormatter
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.feature.toggles.api.RemoteFeatureStoreNamed
```
aitorvs, line 26: Remove
```suggestion
```

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
```diff
@@ -155,8 +155,8 @@ class SharedPreferencesToggleReportsDataStore @Inject constructor(
 
     override suspend fun getPromptsDismissed(): List<String> =
         withContext(dispatcherProvider.io()) {
-            return@withContext store.data.first().let { prefs ->
-                prefs[TOGGLE_REPORTS_PROMPTS_DISMISSED]?.toList() ?: emptyList()
+            return@withContext store.data.firstOrNull().let { prefs ->
```
lmac012, line 155: `store.data` never returns an empty flow, so using `firstOrNull()` is not necessary. Docs recommend using `first()` for accessing store snapshot ->
https://developer.android.com/reference/kotlin/androidx/datastore/core/DataStore#data()

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
```diff
@@ -0,0 +1,210 @@
...
+                val sendTime = DatabaseDateFormatter.millisIso8601().toString()
+                currentSet.add(sendTime)
+                prefs[TOGGLE_REPORTS_SENT] = currentSet
+            }
+        }
+    }
+
+    override suspend fun getReportsSent(): List<String> =
+        withContext(dispatcherProvider.io()) {
+            return@withContext store.data.first().let { prefs ->
```
lmac012, line 177: nit: collecting the `Flow` returned by `store.data` doesn't need to be wrapped in `withContext(dispatcherProvider.io()) { }` - DataStore handles switching dispatchers internally.

privacy-dashboard/privacy-dashboard-impl/src/main/java/com/duckduckgo/privacy/dashboard/impl/ToggleReportsDataStore.kt
```diff
@@ -0,0 +1,210 @@
...
+    suspend fun getReportsSent(): List<String>
+
+    suspend fun insertTogglePromptSend()
+
+    suspend fun getPromptsDismissed(): List<String>
+
+    suspend fun canPrompt(): Boolean
+}
+
+@JsonClass(generateAdapter = true)
```
lmac012, line 62: You're not using Moshi codegen to generate adapters, so this annotation can be removed.

lmac012 (approved):
I tested after the bump to dashboard 7.3.0 and everything works as expected 👍

PR #5305, opened 2024-11-21
Add support for auth API v2

lmac012:
Task/Issue URL: https://app.asana.com/0/1205648422731273/1208821656071388/f

### Description

This PR adds support for new version of the authentication API that uses OAuth 2.0 with PKCE extension. Usage of the new API is controlled via a feature flag which is currently disabled.

### Steps to test this PR

See task.


lmac012:
* **#5305** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5305?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5305?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`

<h2></h2>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/auth2/AuthJwtValidator.kt
karlenDimla, line 1: How is the Entitlement here different from Entitlement in AuthRepository?

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/auth2/BackgroundTokenRefresh.kt
```diff
@@ -0,0 +1,101 @@
...
+interface BackgroundTokenRefresh {
+    fun schedule()
+}
+
+@ContributesBinding(AppScope::class)
+class BackgroundTokenRefreshImpl @Inject constructor(
+    val workManager: WorkManager,
+) : BackgroundTokenRefresh {
+    override fun schedule() {
+        val workRequest = PeriodicWorkRequestBuilder<TokenRefreshWorker>(repeatInterval = Duration.ofDays(7))
```
karlenDimla, line 52: Why are we scheduling this work to repeat every 7 days? Access token expires every 4 hours and refresh tokens expire every month?

karlenDimla (approved):
Test cases passed. Also test a few more VPN scenarios and everything seemed to work as expected. Just a few comments and questions - since the PR is actually quite big.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/SubscriptionsManager.kt
```diff
@@ -277,10 +327,16 @@ class RealSubscriptionsManager @Inject constructor(
     }
 
     override suspend fun signOut() {
+        authRepository.getAccessTokenV2()?.run {
+            coroutineScope.launch { authClient.tryLogout(accessTokenV2 = jwt) }
```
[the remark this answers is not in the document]
lmac012: There is no need to specify io dispatcher anywhere in this method since there are no thread-blocking operations here. Calling `/logout` endpoint is launched in a separate coroutine just because this is a side-effect that the caller of `signOut()` shouldn't care for and wait for. This matters e.g., when the user taps "Remove subscription from device" in settings and we want the repository transition to the signed out state ASAP, so that the UI can be updated accordingly.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/auth2/AuthJwtValidator.kt
lmac012: It's not different, just avoiding having a dependency between `AuthRepository` and `AuthJwtValidator`, since the latter has nothing to do with storage.  Now that both of those classes live in the same module, we could extract `Entitlement` from `AuthRepository`/`AuthJwtValidator` and unify - I'll look into it.

PR #5323, opened 2024-11-27
Tab swiping: Feature flag cache

0nko:
Task/Issue URL: https://app.asana.com/0/1207418217763355/1208852760335259/f

### Description

This PR just caches the feature flag value.

I’ve set the scope of the `IsSwipingTabsFeatureEnabled` to `AppScope` because it’s used inside the `TabManager`, which is used across 2 different activities (`BrowserActivity` and `TabSwitcherActivity`).

### Steps to test this PR

Testing optional.


app/src/main/java/com/duckduckgo/app/browser/IsSwipingTabsFeatureEnabled.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import com.duckduckgo.di.scopes.AppScope
+import dagger.SingleInstanceIn
+import javax.inject.Inject
+
+@SingleInstanceIn(AppScope::class)
+class IsSwipingTabsFeatureEnabled @Inject constructor(
+    swipingTabsFeature: SwipingTabsFeature,
+) {
+    private val isEnabled: Boolean = swipingTabsFeature.self().isEnabled()
```
aitorvs, line 27: are you sure this doens't call IO right when the class is instantiated? If it does (I think it does) we should change it

marcosholgado: +1

0nko: > are you sure this doens't call IO right when the class is instantiated? If it does (I think it does) we should change it.

You mean when the `SwipingTabsFeature` is being initialized for the first time? I don’t think so, at least from what I saw when stepping through the code. But I couldn’t peek inside the store so I may be wrong. We can initialize the variable lazily, though, to be safe.

> Also, I would question why do you need this class anyway, you can more easily do the same (without the overhead of a whole new class) by using lazy in the caller (RealTabManager)

I’m checking for the flag in multiple places (not here but in a subsequent PR), so I need to store the value somewhere for consistency.

PR #5338, opened 2024-12-02
Add localStorage exceptions to duckduckgo.com

aitorvs:
Task/Issue URL: https://app.asana.com/0/1198194956794324/1208875163490368/f

### Steps to test this PR

_Test_
requirement for all tests: IP address needs to be in the US (use DDG production app or another VPN like Nord)

- [ ] fresh install from this branch
- [ ] go through onboarding (sigh)
- [ ] open duckduckgo.com
- [ ] verify a PPro promotion banner appears (`Get Privacy Pro!...`)
- [ ] tap on fire button
- [ ] open duckduckgo.com
- [ ] verify a PPro promotion banner appears (`Get Privacy Pro!...`)
- [ ] dismiss the banner (press `X`)
- [ ] open duckduckgo.com in a new tab
- [ ] verify a PPro promotion banner does NOT appear
- [ ] tap on fire button
- [ ] open duckduckgo.com
- [ ] verify a PPro promotion banner does NOT appear

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/WebViewDataManagerTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/WebViewDataManagerTest.kt
@@ -26,7 +26,9 @@ import androidx.test.platform.app.InstrumentationRegistry
 import com.duckduckgo.app.browser.httpauth.WebViewHttpAuthStore
 import com.duckduckgo.app.browser.session.WebViewSessionInMemoryStorage
 import com.duckduckgo.app.global.file.FileDeleter
+import com.duckduckgo.app.pixels.remoteconfig.AndroidBrowserConfigFeature
 import com.duckduckgo.cookies.api.DuckDuckGoCookieManager
+import com.duckduckgo.feature.toggles.api.FakeFeatureToggleFactory
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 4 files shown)


aitorvs:
* **#5338** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5338?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5338?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`

<h2></h2>

cmonfortep (dismissed):
LGTM 

Just a note, do you think this could impact the fire button process perf depending on the local storage size? unsure how those webview methods perform depending on size. I don't want to slow you down here, but should we consider adding a kill switch in case we detect an increase on feedback around that?

Testing: I don't have vpn, so I've tested this by changing the excluded domain to ".amazon.com" (instead of .duckduckgo.com) and inspecting the local storage via dev tools. After using fire button, local storage was preserved, so it works.

cmonfortep (commented):
I'm taking another look here and there's something I'm missing.

I've been visiting a bunch of sites to see how those webview methods behave, and getOrigins only returns me 1 origin...but from devtools I see all sites stored something inside their localStorage.

The other thing during testing, I've detected a crash in `(origin as Origin).origin` cast, which was not visible since we do runcatching.

Let's pair on this tomorrow to close those open questions.

cmonfortep (approved):
LGTM

The crash when casting is fixed, and we also clear all data if there's any exception there.
As mentioned, I'm not sure how effective the webview api methods are, but since we have a killswitch that if enabled removes all content, 👍

PR #5369, opened 2024-12-09
Intercept requests and add skeleton for malicious site detection

CrisBarreiro:
Task/Issue URL: https://app.asana.com/0/1205008441501016/1207151848931035/f 

### Description
Adds logic to intercept requests using shouldOverrideUrlLoading and shouldInterceptRequest, when the request matches the following conditions
* Is for mainframe
* Is for iframe and the host loading it matches the webview host
* The same URL hasn't been already intercepted during the same page load (used to prevent the same URL being intercepted by the 2 webview callbacks)

### Steps to test this PR

_Feature 1_
- [ ] Load a site and check you get `Timber.tag("MaliciousSiteProtection").d("isMalicious $url")` for all mainframe and iframe requests (only for internal builds). 
- [ ] Check you should never get 2 instances of those logs for the exact same URL

### UI changes
| Before  | After |
| ------ | ----- |


CrisBarreiro:
* **#5369** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`

<h2></h2>

cmonfortep (commented):
Just testing the changes (haven't started the code review yet)

Sharing something I've found:
- when clicking any link in this URL https://www.yahoo.com/?guccounter=1, the url changes but I don't see it  in the logs.
- After some normal browsing on a tab, testing back and forward navigation between sites, I don't see some URLs appearing, and that I can repro. 
- I think this one is fine but just fyi, in the logs I see http://domain.com and https://domain.com (just changing schema http -> https)

Friendly reminder about updating the description of the PR to include what's included and giving some context 🙏

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -158,6 +164,20 @@ class BrowserWebViewClient @Inject constructor(
         try {
             Timber.v("shouldOverride webViewUrl: ${webView.url} URL: $url")
             webViewClientListener?.onShouldOverride()
+
+            if (runBlocking {
```
cmonfortep, line 168: proposal: remove the `suspend` for shouldOverrideUrlLoading, and move the runBlocking inside the implementation of that method. That way we can control where exactly we need it and it's an implementation detail of that method.

malicious-site-protection/malicious-site-protection-impl/src/main/kotlin/com/duckduckgo/malicioussiteprotection/impl/RealMaliciousSiteProtection.kt
```diff
@@ -63,4 +67,10 @@ class RealMaliciousSiteProtection @Inject constructor(
             }
         }
     }
+
+    override suspend fun isMalicious(url: Uri, confirmationCallback: (isMalicious: Boolean) -> Unit): IsMaliciousResult {
```
cmonfortep, line 71: Proposal: implementing `MaliciousSiteProtection` on a different class, and leaving this one exclusively for the feature config state (since the feature state is more than just a boolean and has state). That way we can separate responsibilities, and improve cohesion, ensuring that each class has a well-defined purpose.

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
```diff
@@ -48,6 +49,7 @@ interface RequestInterceptor {
         request: WebResourceRequest,
         webView: WebView,
         documentUri: Uri?,
+        maliciousSiteProtectionWebViewIntegration: MaliciousSiteBlockerWebViewIntegration,
```
cmonfortep, line 52: Is there a specific reason for passing this as a method argument rather than injecting it as a class attribute via the constructor? I would prefer the later since this is a dependency rather than a value.

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
```diff
@@ -92,10 +94,20 @@ class WebViewRequestInterceptor(
...
         webViewClientListener: WebViewClientListener?,
     ): WebResourceResponse? {
         val url: Uri? = request.url
 
+        val confirmationCallback: (isMalicious: Boolean) -> Unit = {
```
cmonfortep, line 102: I assume this will have the same implementation as in `app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt` L131.

Have you considered any options to don't have them duplicated?

app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
```diff
@@ -0,0 +1,160 @@
...
+        if (isForMainFrame && decodedUrl.toUri() == webViewUrl) {
+            if (maliciousSiteProtection.isMalicious(decodedUrl.toUri(), confirmationCallback) == MALICIOUS) {
+                return true
+            }
+            processedUrls.add(decodedUrl)
+        }
+        return false
+    }
+
+    private fun isForIframe(request: WebResourceRequest) = request.requestHeaders["Sec-Fetch-Dest"] == "iframe" ||
```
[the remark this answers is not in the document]
CrisBarreiro: > I'm not very familiar with detecting iframes, so just asking, how can we validate this is the right checks / we are not missing anything?

Neither I am, but I asked Thom on my tech design and he confirmed this is sufficient to detect most iframes and he wouldn't go deeper for now. See https://app.asana.com/0/0/1208756960040667/1208804973280898/f and https://app.asana.com/0/0/1208756960040667/1208809154581536/f

> seems we just want to run our logic for any mainframe OR request related to that documentUri.host, does it matter if iframe is true?

Not really, we want to run our logic for any mainframe OR iframe request related to that documentUri.host. Any other request, (namely resources) shouldn't be processed even if the documentUri.host matches. We only care about mainframe and iframes here

@cmonfortep

CrisBarreiro:
> Just testing the changes (haven't started the code review yet)
> 
> Sharing something I've found:
> 
>     * when clicking any link in this URL https://www.yahoo.com/?guccounter=1, the url changes but I don't see it  in the logs.
> 
>     * After some normal browsing on a tab, testing back and forward navigation between sites, I don't see some URLs appearing, and that I can repro. (edit: actually I think this never works, and the urls I see is just some requests the site triggers)
> 
>     * I think this one is fine but just fyi, in the logs I see http://domain.com and https://domain.com (just changing schema http -> https)
> 
> 
> Friendly reminder about updating the description of the PR to include what's included and giving some context 🙏

Discussed over zoom, there are no requests to such pages when navigating to an article. Documented [here](https://app.asana.com/0/1205008441501016/1208973665379898/f)

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
CrisBarreiro: I'm planning to explore it in an upcoming PR, as part of https://app.asana.com/0/72649045549333/1207151848931034

app/src/main/java/com/duckduckgo/app/browser/WebViewRequestInterceptor.kt
cmonfortep: Singleton + state is something I would be really careful about, race conditions are coming. It's not only users moving to another tab, we also have some scenarios where 2 tabs could be loading a site, and I expect this also changing in the project of swipe tabs. 
For transparency, I would add `shouldOverrideUrlLoading` method to `RequestInterceptor`, only with the logic that applies here, that will simplify things. I know you don't like, and you feel we should then move all the logic we have on that method, I don't think that's necessary.

I'm going to leave you my suggestion if you want to keep the singleton, let's remove the state then (processedUrls). That will save us from unexpected scenarios. If we think is necessary, we can revisit.

app/src/main/java/com/duckduckgo/app/browser/urlextraction/UrlExtractingWebViewClient.kt
```diff
@@ -42,6 +43,7 @@ class UrlExtractingWebViewClient(
     private val appCoroutineScope: CoroutineScope,
     private val dispatcherProvider: DispatcherProvider,
     private val urlExtractor: DOMUrlExtractor,
+    private val maliciousSiteProtection: MaliciousSiteBlockerWebViewIntegration,
```
cmonfortep, line 46: this can be removed

cmonfortep (approved):
LGTM 

Suggested to remove the state from the singleton if we can do that.

app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
```diff
@@ -83,4 +84,13 @@ interface AndroidBrowserConfigFeature {
      */
     @Toggle.DefaultValue(false)
     fun featuresRequestHeader(): Toggle
+
+    /**
+     * @return `true` when the remote config has the global "enableMaliciousSiteProtection" androidBrowserConfig
+     * sub-feature flag enabled
+     * If the remote feature is not present defaults to `false`
+     */
+    @InternalAlwaysEnabled
```
cmonfortep, line 93: can we remove `InternalAlwaysEnabled` and keep it disabled to internals until we have some functionality? right now this is running but no user facing yet.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -424,12 +415,11 @@ class BrowserWebViewClient @Inject constructor(
             // See https://app.asana.com/0/0/1206159443951489/f (WebView limitations)
             if (it != ABOUT_BLANK && start == null) {
                 start = currentTimeProvider.elapsedRealtime()
-                maliciousSiteProtectionWebViewIntegration.onPageLoadStarted()
+                requestInterceptor.onPageStarted(url)
```
CrisBarreiro, line 418: I need to check if moving `requestInterceptor.onPageStarted(url)` here has any unintended side-effect. This is because onPageStarted can be called several times when a page is loaded

PR #5381, opened 2024-12-11
Fix: Don't show the find-in-page bar if it's hidden

0nko:
Task/Issue URL: https://app.asana.com/0/1207418217763355/1205772199506591/f

### Description

This fix prevents showing the find-in-page bar when changing the omnibar position.

### Steps to test this PR

- [x] Go to any URL
- [x] Perform a find in page search for any text
- [x] Dismiss the find in page bar
- [x] Go into settings & switch omnibar to the opposite position
- [x] Notice the find in page toolbar is not visible
- [x] Switch the omnibar position again


malmstein (approved):
Works as expected, thanks @0nko !

PR #5391, opened 2024-12-13
add default browser selection overflow menu item

LukasPaczos:
Task/Issue URL: https://app.asana.com/0/72649045549333/1208944504536342/f

### Description
Adds an overflow menu item that advertises setting DuckDuckGo as the default browser.

The introduced item is not yet used anywhere and click listeners/handlers will be added in following PRs.

### Steps to test this PR
As an example, apply below diff and the item will show up whenever you open the overflow menu.
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/browser/viewstate/BrowserViewState.kt b/app/src/main/java/com/duckduckgo/app/browser/viewstate/BrowserViewState.kt
index bfb487f1e..5db2efe5b 100644
--- a/app/src/main/java/com/duckduckgo/app/browser/viewstate/BrowserViewState.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/viewstate/BrowserViewState.kt
@@ -34,7 +34,7 @@ data class BrowserViewState(
     val showTabsButton: Boolean = true,
     val fireButton: HighlightableButton = HighlightableButton.Visible(),
     val showMenuButton: HighlightableButton = HighlightableButton.Visible(),
-    val showSelectDefaultBrowserMenuItem: Boolean = false,
+    val showSelectDefaultBrowserMenuItem: Boolean = true,
     val canSharePage: Boolean = false,
     val canSaveSite: Boolean = false,
     val bookmark: SavedSite.Bookmark? = null,
```

### UI changes
| Light  | Dark |
| ------ | ----- |
Screenshot_20241213_172011|Screenshot_20241213_172029|
Screenshot_20241213_172048|Screenshot_20241213_172149|


app/src/main/res/values/strings.xml
```diff
@@ -35,6 +35,7 @@
     <string name="settingsDefaultBrowserEnabled">Set as Default Browser</string>
     <string name="defaultBrowserMaybeLater">Maybe Later</string>
     <string name="defaultBrowserLetsDoIt">Set as Default Browser</string>
+    <string name="defaultBrowserMenuTitle">Set as Default Browser</string>
```
LukasPaczos, line 38: I assumed that we want to keep each text view's string separate to avoid potential changes impacting other text views unexpectedly. And since we already had the same string translated, I copied that around.

common/common-ui/src/main/res/drawable/ic_default_browser_mobile_color_16.xml
LukasPaczos, line 1: Is `common-ui` the right place for the new icon? Or should it land in `app` @malmstein?

common/common-ui/src/main/res/values/design-system-colors.xml
LukasPaczos, line 1: Is `common-ui` the right place for the new color? Or should it land in `app` @malmstein?

app/src/main/res/values/strings.xml
LukasPaczos: I added a commit that consolidates the strings instead.

malmstein (changes requested):
You don’t need to change the translations, let the service do it for you. Get rid of all the strings changes, and only create new values under donottranslate.xml. Once we are done, we’ll order translations and let the process take its course.

app/src/main/res/drawable/background_default_browser_menu_item.xml
```diff
@@ -0,0 +1,20 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <solid android:color="@color/blue50_18" />
+    <corners android:radius="8dp" />
```
malmstein, line 19: there’s a radius dimension for 8dp

app/src/main/res/drawable/background_default_browser_menu_item.xml
```diff
@@ -0,0 +1,20 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <solid android:color="@color/blue50_18" />
```
malmstein, line 18: do they really want the same color for dark and light? all button have different colors per theme, seems strange that this is not changing.

app/src/main/res/drawable/background_default_browser_menu_item.xml
LukasPaczos: Fixed.

app/src/main/res/drawable/background_default_browser_menu_item.xml
LukasPaczos: Yes, that was confirmed during kick-off. I'll double check again but I think we can merge and I'll adjust again if I hear something different.

LukasPaczos:
> You don’t need to change the translations, let the service do it for you.

I don't want to change translations, I want to reuse the one we already have in the codebase but I'm adjusting the constant's name to better reflect what it does. Is that a problem? Why would we need to run this through the service again if all the information is already there?

If I removed all the strings and added a new one that's not translated, I'd be regressing the functionality that already relies on these translations in other places.

malmstein (approved):
Looks good, thanks @LukasPaczos !

PR #5458, opened 2025-01-11
Ensure LevelDB Iterator is closed

joshliebe:
Task/Issue URL: https://app.asana.com/0/1200204095367872/1209128100809324/f

### Description

- Closes the `DBIterator` after local storage has been cleared.
- Also keeps the DB open and updates the DB initialization to be `Lazy`.

### Steps to test this PR

_Filter logcat by WebLocalStorageManager_
- [x] Visit cnn.com
- [x] Use the fire button
- [x] Verify that cnn.com entries are deleted

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/weblocalstorage/WebLocalStorageManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/weblocalstorage/WebLocalStorageManager.kt
@@ -21,13 +21,13 @@ import com.duckduckgo.app.pixels.remoteconfig.AndroidBrowserConfigFeature
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesBinding
 import com.squareup.anvil.annotations.ContributesTo
+import dagger.Lazy
 import dagger.Module
 import dagger.Provides
 import dagger.SingleInstanceIn
 import java.io.File
 import java.nio.charset.StandardCharsets
 import javax.inject.Inject
-import javax.inject.Provider
 import kotlinx.coroutines.runBlocking
 import org.iq80.leveldb.DB
 import org.iq80.leveldb.Options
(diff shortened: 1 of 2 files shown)


joshliebe:
* **#5458** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5458?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5458?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/weblocalstorage/WebLocalStorageManager.kt
```diff
@@ -59,16 +59,17 @@ class DuckDuckGoWebLocalStorageManager @Inject constructor(
         Timber.d("WebLocalStorageManager: Matching regex: $matchingRegex")
 
         databaseProvider.get().use { db ->
```
aitorvs, line 61: Unrelated but related to this PR.
Not sure we should use `databaseProvider.get().use {}` block on the db. The DB is a `SingleInstance` in the `AppScope`, which basically means it's a singleton.
Even if we use a `Provider` which tries to provide a new instance every time, DI will return the singleton all the time.

This could potentially caused issues, in the case where `clearWebLocalStorage` is executed twice in the same AppScope lifespane.
Probably this doesn't happen because `clearWebLocalStorage` is executed always when we burn and we kill the process (?) but even so we should fix that.

Given that the DB instance is a singletone in the AppScope, is fine not to ever call `close` on it., It'll automatically close once the app process is killed.

My suggestion is to remove the `use` call on the db.

aitorvs (approved):

PR #5481, opened 2025-01-16
Swiping tabs: Fix duplicate animations & edit mode detection

0nko:
Task/Issue URL: https://app.asana.com/0/1207418217763355/1209152058231741/f

### Description

This PR removes the onboarding restriction that prevents tab swiping until onboarding is complete. It also fixes a bug that shows duplicate omnibar highlight animations. The bug is caused due to RecyclerView calling the `View.onAttachedToWindow()` every time the current item is changed, resulting in duplicate `PulseAnimation` instances.

The fix involves removing `PulseAnimation` initialization from the `onAttachedToWindow()` callback, so that it’s initialized only once. 

I also moved the injection to the constructor to avoid repeated injection calls and refactored the ViewModel flows collection to only happen during active lifecycle states.

While working on this, I discovered that edit mode wasn’t working properly and there’s a simpler way to do it, so I included it here.

### Steps to test this PR

- [x] Make sure you enable the tab swiping feature in the settings and restart the app
- [x] Go through onboarding until you get to the privacy shield step (the shield animation is displayed)
- [x] Add a new tab
- [x] Swipe back to the tab with the shield animation
- [x] Notice there is only one animation
- [x] Go to the tab switcher and come back
- [x] Notice there is still only one animation


app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -126,13 +126,14 @@ class OmnibarLayoutViewModel @Inject constructor(
...
 
-    fun onAttachedToWindow() {
+    init {
         tabRepository.flowTabs
             .onEach { tabs ->
                 _viewState.update {
                     it.copy(
-                        shouldUpdateTabsCount = tabs.count() != it.tabs.count() || tabs.isNotEmpty(),
-                        tabs = tabs,
+                        shouldUpdateTabsCount = tabs.size != it.tabCount && tabs.isNotEmpty(),
```
0nko, line 134: I think the original condition was wrong — we don’t ever want to show 0 (the initial ViewState value), because the actual count will always be at least 1.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -99,7 +98,8 @@ class OmnibarLayoutViewModel @Inject constructor(
         val updateOmnibarText: Boolean = false,
         val shouldMoveCaretToEnd: Boolean = false,
         val shouldMoveCaretToStart: Boolean = false,
-        val tabs: List<TabEntity> = emptyList(),
```
0nko, line 102: This list was only used to count the tabs. Each tab would contain a list of all these entities and with lots of tabs this would unnecessarily take up a lot of memory.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
```diff
@@ -238,24 +246,30 @@ class OmnibarLayout @JvmOverloads constructor(
         )[OmnibarLayoutViewModel::class.java]
     }
 
+    // One OmnibarLayout is always removed from the view hierarchy, we should cancel any flow subscriptions
+    fun onRemoved() {
```
0nko, line 250: Either bottom or the top omnibar is always removed but its ViewModel flow subscriptions would keep collecting & rendering state for no reason.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -81,7 +80,13 @@ class OmnibarLayoutViewModel @Inject constructor(
 ) : ViewModel() {
 
     private val _viewState = MutableStateFlow(ViewState())
-    val viewState = _viewState.asStateFlow()
+    val viewState = combine(_viewState, tabRepository.flowTabs) { state, tabs ->
+        state.copy(
+            shouldUpdateTabsCount = tabs.size != state.tabCount && tabs.isNotEmpty(),
```
0nko, line 85: I think the original condition was wrong — we don’t ever want to show 0 (the initial ViewState value), because the actual count will always be at least 1.

anikiki (approved):
LGTM 👍

PR #5517, opened 2025-01-22
Inject larger autoconsent bundle containing cosmetic filters

dharb:
Task/Issue URL: https://app.asana.com/0/0/1208811874906681/f

### Description
This PR is the first step in expanding CPM coverage on mobile by way of using easylist cookie filterlists. There are no user facing changes in this PR: all that's happening here is we are adding a new autoconsent subfeature called filterlist, and injecting a different autoconsent bundle if that feature flag is enabled via privacy configuration. The corresponding Privacy Configuration PR can be found here: https://github.com/duckduckgo/privacy-configuration/pull/2659.

### Steps to test this PR
1. Checkout this branch, build locally
2. Point app at this config: https://jsonblob.com/api/1331752608628662272. 
The only change here is the added `filterlist` autoconsent subfeature:
``` JSON
"filterlist": {
    "state": "enabled"
}
```
3. Ensure that CPM is on, then visit a website that we have an existing CPM rule for (eg ecosia.org). Confirm that the cookie popup is hidden.
4. Visit a website that we don't have an existing CPM rule for, but there is an easylist rule for (eg rijksmuseum.nl). Confirm that the cookie popup is not hidden.
To fully confirm that the larger bundle is being injected when the feature flag is enabled, you may do so by enabling the feature as follows:
1. On https://github.com/duckduckgo/Android/blob/dharb/autoconsent-filterlist-test/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/InitMessageHandlerPlugin.kt#L76, add `enableFilterList = true` to list of params
2. On https://github.com/duckduckgo/Android/blob/dharb/autoconsent-filterlist-test/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/InitMessageHandlerPlugin.kt#L115, add `val enableFilterList: Boolean` to the list of params
3. Rebuild app
4. Double check that config is still pointed at the jsonblob url above
5. Ensure that CPM is on, then visit a website that we have an existing CPM rule for (eg ecosia.org). Confirm that the cookie popup is hidden.
6. Visit a website that we don't have an existing CPM rule for, but there is an easylist rule for (eg rijksmuseum.nl). Confirm that the cookie popup is hidden.
7. As a final check, use the fire button to clear all local data, then point privacy config back at the production endpoint where this feature isn't enabled. Then visit ecosia.org and rijksmuseum.nl again and confirm that the cookie pop is hidden on ecosia.org, and is not hidden on rijksmuseum.nl


autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/RealAutoconsent.kt
```diff
@@ -47,6 +52,7 @@ class RealAutoconsent @Inject constructor(
     private val settingsRepository: AutoconsentSettingsRepository,
     private val autoconsentExceptionsRepository: AutoconsentExceptionsRepository,
     private val autoconsent: AutoconsentFeature,
+    private val pixel: Pixel,
```
dharb, line 55: Suspect I'm not doing this correctly because it's causing unit tests to fail. How should I do this differently?

dharb:
Closing as we decided not to go this route for the time being.

PR #5476, opened 2025-01-15
Additional "default browser" prompts: variant 2

LukasPaczos:
Task/Issue URL: https://app.asana.com/0/1208671518894266/1208944504536347/f

### Description
Introduces variant 2 of the experiment that proposes new CTAs around setting DDG as a default browser. The PR also adds all of the experiment foundations and abstractions, so the remaining two variants will only be a small formality.

I'm going to follow up on pixels in a separate PR.

### Class diagram
image

### Flow diagram
image

### Steps to test this PR

This PR relies on the new A/B/N framework for rollout, so to test you'll need a custom privacy config which includes the feature definition, for example:
```json
"defaultBrowserPrompts": {
  "state": "enabled",
  "exceptions": [],
  "features": {
    "additionalPrompts": {
      "state": "enabled",
      "rollout": {
        "steps": [
          {
            "percent": 100
          }
        ]
      },
      "settings": {
        "activeDaysUntilStage1": 1,
        "activeDaysUntilStage2": 3,
        "activeDaysUntilStop": 5
      },
      "cohorts": [
        {
          "name": "control",
          "weight": 0
        },
        {
          "name": "variant_2",
          "weight": 1
        }
      ]
    }
  },
}
```

or you can apply this JSON Blob:
```diff
diff --git a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
index ca0870311..5a904f76b 100644
--- a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
+++ b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
@@ -27,4 +27,5 @@ enum class PrivacyFeatureName(val value: String) {
     TrackingParametersFeatureName("trackingParameters"),
 }
 
-const val PRIVACY_REMOTE_CONFIG_URL = "https://staticcdn.duckduckgo.com/trackerblocking/config/v4/android-config.json"
+// const val PRIVACY_REMOTE_CONFIG_URL = "https://staticcdn.duckduckgo.com/trackerblocking/config/v4/android-config.json"
+const val PRIVACY_REMOTE_CONFIG_URL = "https://www.jsonblob.com/api/1329040942036082688"
```

[...]

PR #5523, opened 2025-01-23
Additional "default browser" prompts: enroll only onboarded users

LukasPaczos:
Task/Issue URL: https://app.asana.com/0/1208671518894266/1209220388656415/f

### Description
We only want to enroll users that are already onboarded and have seen the default browser prompt that's part of onboarding.

### Steps to test this PR

You can follow instructions from https://github.com/duckduckgo/Android/pull/5476 but when you're in the onboarding experience, minimize the app, change the date forward, and resume the app. Previously the experiment's default browser dialog would show, with the proposed changes it won't, because we're only enrolling users after onboarding is done.


app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
```diff
@@ -154,9 +165,10 @@ class DefaultBrowserPromptsExperimentImpl @Inject constructor(
     }
 
     private suspend fun evaluate() {
+        val isOnboardingComplete = userStageStore.getUserAppStage() == AppStage.ESTABLISHED
```
malmstein, line 168: Has @nalcalag looked at this? In the past we’ve had issues understanding if onboarding was complete in some edge cases, so just want to double check.

malmstein (approved):
I assume you’ve checked regarding the Onboarding state. Approving this pre-emptively.

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
nalcalag: Thanks for the heads-up @malmstein . This check is correct, when onboarding is completed, the `UserAppStage` should be set to `ESTABLISHED`. This means all onboarding dialogs, both in-context and on the new tab page, will have already been shown.

There are other methods to determine if parts of the onboarding process have been completed, such as checking for dialogs on the new tab page, etc. I’m not sure about the specific requirements for this feature, but feel free to reach out to me if needed @LukasPaczos .

PR #5535, opened 2025-01-24
Store full enrollment date on a/b/n and truncate before sending pixels

marcosholgado:
Task/Issue URL: https://app.asana.com/0/1125189844152671/1209213126411295/f 

### Description
This PR stores the full enrollment date for better comparison but still truncates it when sending pixels.

### Steps to test this PR

- [ ] Tests should pass
- [ ] Load https://www.jsonblob.com/api/1304478638945460224 to enroll in an experiment
- [ ] Filter the logcat by "pixel sent"
- [ ] You should see an experiment_enroll pixel sent with the enrollment date truncated
- [ ] Make a search
- [ ] You should see an experiment_metric... pixel sent with the enrollment date truncated
- [ ] Check shared prefs `com.duckduckgo.feature.toggle.blocklist`. The enrollment date for `blockList_tdsNextExperimentBaselineBackup` should have the full date rather than be truncated to days.

### DIFF
--- a/feature-toggles/feature-toggles-impl/src/main/java/com/duckduckgo/feature/toggles/impl/metrics/RetentionMetricsAtbLifecyclePlugin.kt
+++ b/feature-toggles/feature-toggles-impl/src/main/java/com/duckduckgo/feature/toggles/impl/metrics/RetentionMetricsAtbLifecyclePlugin.kt
@@ -20,12 +20,12 @@ import com.duckduckgo.app.di.AppCoroutineScope
 import com.duckduckgo.app.statistics.api.AtbLifecyclePlugin
 import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.di.scopes.AppScope
-import com.duckduckgo.feature.toggles.api.MetricsPixel
 import com.duckduckgo.feature.toggles.api.PixelDefinition
 import com.duckduckgo.feature.toggles.impl.MetricsPixelStore
 import com.duckduckgo.feature.toggles.impl.RetentionMetric.APP_USE
 import com.duckduckgo.feature.toggles.impl.RetentionMetric.SEARCH
 import com.squareup.anvil.annotations.ContributesMultibinding
+import java.time.LocalDate
 import java.time.ZoneId
 import java.time.ZonedDateTime
 import java.time.temporal.ChronoUnit
@@ -46,7 +46,7 @@ class RetentionMetricsAtbLifecyclePlugin @Inject constructor(
         appCoroutineScope.launch {
             searchMetricPixelsPlugin.getMetrics().forEach { metric ->
                 metric.getPixelDefinitions().forEach { definition ->
-                    if (isInConversionWindow(definition, metric)) {
+                    if (isInConversionWindow(definition)) {
                         store.getMetricForPixelDefinition(definition, SEARCH).takeIf { it < metric.value.toInt() }?.let {
                             store.increaseMetricForPixelDefinition(definition, SEARCH).takeIf { it == metric.value.toInt() }?.apply {
                                 pixel.fire(definition.pixelName, definition.params)
@@ -62,7 +62,7 @@ class RetentionMetricsAtbLifecyclePlugin @Inject constructor(
         appCoroutineScope.launch {
             appUseMetricPixelsPlugin.getMetrics().forEach { metric ->
                 metric.getPixelDefinitions().forEach { definition ->
-                    if (isInConversionWindow(definition, metric)) {
+                    if (isInConversionWindow(definition)) {
                         store.getMetricForPixelDefinition(definition, APP_USE).takeIf { it < metric.value.toInt() }?.let {
                             store.increaseMetricForPixelDefinition(definition, APP_USE).takeIf { it == metric.value.toInt() }?.apply {
                                 pixel.fire(definition.pixelName, definition.params)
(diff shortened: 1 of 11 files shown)


marcosholgado:
* **#5535** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5535?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5535?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

feature-toggles/feature-toggles-impl/src/main/java/com/duckduckgo/feature/toggles/impl/metrics/RetentionMetricsAtbLifecyclePlugin.kt
```diff
@@ -85,8 +85,7 @@ class RetentionMetricsAtbLifecyclePlugin @Inject constructor(
 
     private fun daysBetweenTodayAnd(date: String): Long {
         val today = ZonedDateTime.now(ZoneId.of("America/New_York"))
-        val localDate = LocalDate.parse(date)
-        val zoneDateTime: ZonedDateTime = localDate.atStartOfDay(ZoneId.of("America/New_York"))
-        return ChronoUnit.DAYS.between(zoneDateTime, today)
+        val localDate = ZonedDateTime.parse(date)
+        return ChronoUnit.DAYS.between(localDate, today)
```
LukasPaczos, line 89: This now returns different values than it used to.

Here's an example:
```
    @Test
    fun test() {
        val enrollmentTruncatedZDT = "2025-01-30"
        val enrollmentNotTruncatedZDT = "2025-01-30T18:05:28.548199-05:00[America/New_York]"

        // check on the next day when MORE than 24 hours since enrollment elapsed, both methods return 1 day diff
        var currentZDT = "2025-01-31T18:06:07.467047-05:00[America/New_York]"
        val old1 = oldDaysBetweenTodayAnd(enrollmentDate = enrollmentTruncatedZDT, currentZDTString = currentZDT)
        val new1 = daysBetweenTodayAnd(enrollmentDate = enrollmentNotTruncatedZDT, currentZDTString = currentZDT)
        assertEquals(old1, new1) // passes

        // check on the next day when LESS than 24 hours since enrollment elapsed, old method returns 1 while the new returns 0
        currentZDT = "2025-01-31T18:04:07.467047-05:00[America/New_York]"
        val old2 = oldDaysBetweenTodayAnd(enrollmentDate = enrollmentTruncatedZDT, currentZDTString = currentZDT)
        val new2 = daysBetweenTodayAnd(enrollmentDate = enrollmentNotTruncatedZDT, currentZDTString = currentZDT)
        assertEquals(old2, new2) // fails
    }

    private fun oldDaysBetweenTodayAnd(enrollmentDate: String, currentZDTString: String): Long {
        val today = ZonedDateTime.parse(currentZDTString)
        val localDate = LocalDate.parse(enrollmentDate)
        val zoneDateTime: ZonedDateTime = localDate.atStartOfDay(ZoneId.of("America/New_York"))
        return ChronoUnit.DAYS.between(zoneDateTime, today)
    }

    private fun daysBetweenTodayAnd(enrollmentDate: String, currentZDTString: String): Long {
        val today = ZonedDateTime.parse(currentZDTString)
        val localDate = ZonedDateTime.parse(enrollmentDate)
        return ChronoUnit.DAYS.between(localDate, today)
    }
```

If for the aim of the framework was only to check for calendar day changes without accounting for hours, then the new iteration would regress by being too precise.

feature-toggles/feature-toggles-impl/src/main/java/com/duckduckgo/feature/toggles/impl/metrics/RetentionMetricsAtbLifecyclePlugin.kt
```diff
@@ -85,8 +85,7 @@ class RetentionMetricsAtbLifecyclePlugin @Inject constructor(
 
     private fun daysBetweenTodayAnd(date: String): Long {
```
LukasPaczos, line 86: There's one more copy of this function in [MetricsPixelInterceptor](https://github.com/duckduckgo/Android/blob/e8267daaf97981563608a355eb2d5b4c60ab5262/feature-toggles/feature-toggles-impl/src/main/java/com/duckduckgo/feature/toggles/impl/MetricPixelInterceptor.kt#L105) that we should also adjust, shouldn't we?

LukasPaczos: This one looks fine to stay because it takes truncated days as parameter :+1: It would be great to add a small comment to the function parameter so that someone stumbling on this in the future will immediately understand the difference between the way we parse and use the various `enrollmentDate` instances.

feature-toggles/feature-toggles-impl/src/main/java/com/duckduckgo/feature/toggles/impl/metrics/RetentionMetricsAtbLifecyclePlugin.kt
marcosholgado: We really need to be more precise, should be 24h diff rather than calendar days so the new one seems to work as expected compared to the old.

LukasPaczos: :+1: In this case, https://github.com/duckduckgo/Android/pull/5535#discussion_r1930659661 would need to get updated in the end.

marcosholgado: Rolling back on this thought, it shouldn't be 24h so I'm changing this bit and leaving the other one as it is.

PR #5543, opened 2025-01-27
Additional "default browser" prompts: metrics

LukasPaczos:
Task/Issue URL: https://app.asana.com/0/1201621853593513/1208697958169120/f

Marking as blocked until all pixels are confirmed with the DS team. This is generally ready for a first round of reviews but some additional changes may be still introduced because of that.

### Steps to test this PR

You can follow instructions from https://github.com/duckduckgo/Android/pull/5476 to go through the experiment.

Test for below pixels:

| Action  | Pixel |
| ------ | ----- |
| Enrolled | `experiment_enroll_defaultBrowserAdditionalPrompts202501_variant_2 with params: {enrollmentDate=2025-01-27}` |
| Reached Stage 1 | `experiment_metrics_defaultBrowserAdditionalPrompts202501_variant_2 with params: {metric=stageImpression, value=stage_1, enrollmentDate=2025-01-27, conversionWindowDays=1-X}` (`X` is 20, 40, 60) |
| Reached Stage 2 | `experiment_metrics_defaultBrowserAdditionalPrompts202501_variant_2 with params: {metric=stageImpression, value=stage_2, enrollmentDate=2025-01-27, conversionWindowDays=1-X}` (`X` is 20, 40, 60) |
| Message dialog dismissed or "not now" clicked | `m_set-as-default_prompt_dismissed with params: {expVar=variant_2, expStage=stage_1}`|
| Message dialog confirmation clicked | `m_set-as-default_prompt_click with params: {expVar=variant_2, expStage=stage_1}`|
| Menu item clicked | `m_set-as-default_in-menu_click with params: {expVar=variant_2, expStage=stage_2}`|
| Browser set as default (from any source) | `experiment_metrics_defaultBrowserAdditionalPrompts202501_variant_2 with params: {metric=defaultSet, value=stage_2, enrollmentDate=2025-01-27, conversionWindowDays=1-X}` (`X` is 20, 40, 60) |
| Browser set as default when preceding action is dialog or menu. This is sent in addition to the above metric. | `experiment_metrics_defaultBrowserAdditionalPrompts202501_variant_2 with params: {metric=defaultSetViaCta, value=menu, enrollmentDate=2025-01-27, conversionWindowDays=1-X}` (`X` is 20, 40, 60) |

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
@@ -24,12 +24,13 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.duckduckgo.anvil.annotations.ContributesRemoteFeature
 import com.duckduckgo.anvil.annotations.ContributesViewModel
-import com.duckduckgo.app.browser.BrowserViewModel.Command.HideSetAsDefaultBrowserDialog
+import com.duckduckgo.app.browser.BrowserViewModel.Command.DismissSetAsDefaultBrowserDialog
 import com.duckduckgo.app.browser.defaultbrowsing.DefaultBrowserDetector
 import com.duckduckgo.app.browser.defaultbrowsing.prompts.DefaultBrowserPromptsExperiment
 import com.duckduckgo.app.browser.defaultbrowsing.prompts.DefaultBrowserPromptsExperiment.Command.OpenMessageDialog
 import com.duckduckgo.app.browser.defaultbrowsing.prompts.DefaultBrowserPromptsExperiment.Command.OpenSystemDefaultAppsActivity
 import com.duckduckgo.app.browser.defaultbrowsing.prompts.DefaultBrowserPromptsExperiment.Command.OpenSystemDefaultBrowserDialog
+import com.duckduckgo.app.browser.defaultbrowsing.prompts.DefaultBrowserPromptsExperiment.SetAsDefaultActionTrigger
 import com.duckduckgo.app.browser.omnibar.OmnibarEntryConverter
 import com.duckduckgo.app.fire.DataClearer
 import com.duckduckgo.app.generalsettings.showonapplaunch.ShowOnAppLaunchFeature
@@ -115,7 +116,7 @@ class BrowserViewModel @Inject constructor(
         data class OpenInNewTab(val url: String) : Command()
         data class OpenSavedSite(val url: String) : Command()
         data object ShowSetAsDefaultBrowserDialog : Command()
-        data object HideSetAsDefaultBrowserDialog : Command()
+        data object DismissSetAsDefaultBrowserDialog : Command()
         data class ShowSystemDefaultBrowserDialog(val intent: Intent) : Command()
         data class ShowSystemDefaultAppsActivity(val intent: Intent) : Command()
     }
@@ -173,6 +174,9 @@ class BrowserViewModel @Inject constructor(
         }
     }
 
+    private var lastSystemDefaultAppsTrigger: SetAsDefaultActionTrigger = SetAsDefaultActionTrigger.UNKNOWN
+    private var lastSystemDefaultBrowserDialogTrigger: SetAsDefaultActionTrigger = SetAsDefaultActionTrigger.UNKNOWN
+
     init {
         appEnjoymentPromptEmitter.promptType.observeForever(appEnjoymentObserver)
         viewModelScope.launch {
@@ -183,10 +187,12 @@ class BrowserViewModel @Inject constructor(
                     }
 
                     is OpenSystemDefaultAppsActivity -> {
+                        lastSystemDefaultAppsTrigger = it.trigger
                         command.value = Command.ShowSystemDefaultAppsActivity(it.intent)
                     }
 
                     is OpenSystemDefaultBrowserDialog -> {
+                        lastSystemDefaultBrowserDialogTrigger = it.trigger
                         command.value = Command.ShowSystemDefaultBrowserDialog(it.intent)
                     }
                 }
@@ -381,17 +387,17 @@ class BrowserViewModel @Inject constructor(
         defaultBrowserPromptsExperiment.onMessageDialogShown()
     }
 
-    fun onSetDefaultBrowserDismissed() {
-        defaultBrowserPromptsExperiment.onMessageDialogDismissed()
+    fun onSetDefaultBrowserDialogCanceled() {
+        defaultBrowserPromptsExperiment.onMessageDialogCanceled()
     }
 
     fun onSetDefaultBrowserConfirmationButtonClicked() {
-        command.value = HideSetAsDefaultBrowserDialog
+        command.value = DismissSetAsDefaultBrowserDialog
         defaultBrowserPromptsExperiment.onMessageDialogConfirmationButtonClicked()
     }
 
     fun onSetDefaultBrowserNotNowButtonClicked() {
-        command.value = HideSetAsDefaultBrowserDialog
+        command.value = DismissSetAsDefaultBrowserDialog
         defaultBrowserPromptsExperiment.onMessageDialogNotNowButtonClicked()
     }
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentMetrics.kt
```diff
@@ -0,0 +1,126 @@
...
+@ContributesMultibinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
+class DefaultBrowserPromptsExperimentMetrics @Inject constructor(
+    private val defaultBrowserPromptsFeatureToggles: DefaultBrowserPromptsFeatureToggles,
+) : MetricsPixelPlugin {
+    override suspend fun getMetrics(): List<MetricsPixel> {
+        return listOf(
+            MetricsPixel(
+                metric = METRIC_DEFAULT_SET,
+                value = METRIC_VALUE_ENROLLED,
```
marcosholgado, line 44: What's this for? Enrollment pixels are sent automatically when a cohort is assigned so you don't need to do this.

LukasPaczos: This one will be removed, refs Thttps://app.asana.com/0/0/1208774988133227/1209263941236041/f. It was supposed to be a conversion pixel for day 0, before user has seen any prompts, but we decided that we don't need it.

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
```diff
@@ -361,19 +367,21 @@ class BrowserViewModel @Inject constructor(
     }
 
     fun onSystemDefaultBrowserDialogSuccess() {
-        defaultBrowserPromptsExperiment.onSystemDefaultBrowserDialogSuccess()
+        defaultBrowserPromptsExperiment.onSystemDefaultBrowserDialogSuccess(
+            lastSystemDefaultBrowserDialogTrigger ?: SetAsDefaultActionTrigger.UNKNOWN,
```
LukasPaczos, line 371: In practice, we shouldn't need to fall back to `UNKNOWN`. I added it as a safeguard because after briefly looking at the problem I couldn't find a non-nullable way to round-trip a parameter through the "activity for result" stack.

Since we can't launch the default browser dialog or system apps and expect them to return a specific parameter back, we have to store it locally. This makes the parameter nullable, which is why we need the `UNKNOWN` fallback.

LukasPaczos:
This is ready for review.

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
malmstein: Why not make a case of the enum for not being handled and set it as default instead? You are checking for a null parameter, why not checking for the non-handled case?

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
```diff
@@ -265,20 +293,25 @@ class DefaultBrowserPromptsExperimentImpl @Inject constructor(
...
     }
 
     override fun onMessageDialogConfirmationButtonClicked() {
-        launchBestSelectionWindow()
+        fireInteractionPixel(AppPixelName.SET_AS_DEFAULT_PROMPT_CLICK)
+        launchBestSelectionWindow(trigger = DIALOG)
     }
 
     override fun onMessageDialogNotNowButtonClicked() {
+        fireInteractionPixel(AppPixelName.SET_AS_DEFAULT_PROMPT_DISMISSED)
```
malmstein, line 314: don’t we want to differentiate between dismissing a dialog and actively saying no?

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
```diff
@@ -265,20 +293,25 @@ class DefaultBrowserPromptsExperimentImpl @Inject constructor(
...
+        launchBestSelectionWindow(trigger = MENU)
     }
 
     override fun onMessageDialogShown() {
+        fireInteractionPixel(AppPixelName.SET_AS_DEFAULT_PROMPT_IMPRESSION)
     }
 
-    override fun onMessageDialogDismissed() {
+    override fun onMessageDialogCanceled() {
+        fireInteractionPixel(AppPixelName.SET_AS_DEFAULT_PROMPT_DISMISSED)
```
malmstein, line 305: This is still a dismissed pixel, but it’s cancelled instead.

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/ui/DefaultBrowserBottomSheetDialog.kt
```diff
@@ -47,8 +47,8 @@ class DefaultBrowserBottomSheetDialog(private val context: Context) : BottomShee
             setRoundCorners(dialogInterface)
             eventListener?.onShown()
         }
-        setOnDismissListener {
-            eventListener?.onDismissed()
+        setOnCancelListener {
```
malmstein, line 50: what’s the reasoning for this?

app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt
```diff
@@ -383,4 +383,9 @@ enum class AppPixelName(override val pixelName: String) : Pixel.PixelName {
     DEDICATED_WEBVIEW_URL_EXTRACTION_FAILED("m_dedicated_webview_url_extraction_failed"),
 
     BLOCKLIST_TDS_FAILURE("blocklist_experiment_tds_download_failure"),
+
+    SET_AS_DEFAULT_PROMPT_IMPRESSION("m_set-as-default_prompt_impression"),
```
malmstein, line 387: either use “-“ or “_” but don’t use a combination of both.

malmstein (approved):
Nothing blocking, I’ve got a couple questions on the implementation.

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
LukasPaczos: The pixel is defined as "triggered when user clicks 'Not now' or uses a gesture to close the prompt". In the end, regardless of whether user collapses the dialog or clicks not now, the intention is the same.

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
LukasPaczos: "Dismissed" and "canceled" mean different things in context of an Android dialog, which doesn't translate to what we mean by "dismissed" in context of the pixel. The pixel wants to know when user didn't take the action and closed the dialog either by a gesture or by a "not now" button.

In the Android impl, dialog cancellation represents one of ways to dismiss a dialog, which includes clicking away and collapsing it.
On the other hand, dialog dismissal happens at any time a dialog is removed, including when a gesture is used (so when the dialog is canceled), but also when we're programmatically dismissing a dialog after either button is clicked by the user.

In our case, clicking the "set as default" button on the dialog sends it's own "click" pixel, and we don't want to send a "dismissed" pixel at the same time when we programmatically close the dialog after the user interaction.

So each of the buttons sends its own respective pixel and dismisses the dialog, and we also listen for when dialog is canceled (no button was pressed) and send the respective pixel as well.

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/ui/DefaultBrowserBottomSheetDialog.kt
LukasPaczos: Explained in https://github.com/duckduckgo/Android/pull/5543#discussion_r1942559065.

app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt
LukasPaczos: I'm using this to treat the new pixels like a tree:
- `m`
  - `set-as-default`
    - `prompt`
      - `impression`
      - `click`
      - `dismissed`
    - `in-menu`
      - `click`

I see various examples of that in the codebase, but please let me know if that should cause any issues.

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
LukasPaczos: I have a case in `DefaultBrowserPromptsExperimentImpl` to handle the `UNKNOWN` type, if that's what you mean. I can also just make the cache field default to `UNKNOWN` to get rid of the elvis operator, it won't change anything functionally in the end.

PR #5604, opened 2025-02-06
Add Malicious Site Protection algorithm pixels

CrisBarreiro:
Task/Issue URL: https://app.asana.com/0/488551667048375/1209304069660515

### Description
*  Add Malicious Site Protection algorithm pixels

### Steps to test this PR

_Feature 1_
- [x] Load https://privacy-test-pages.site/security/badware/phishing-legit-iframe-loader.html
- [x] Check m_malicious-site-protection_iframe-loaded with params: {category=phishing}  is fired
- [x] Check m_malicious-site-protection_error-page-shown is fired

_Feature 2_
- [x] Set timeout to 10ms in override suspend fun matches(hashPrefix: String): List<Match>
- [x] Load https://privacy-test-pages.site/security/badware/phishing-legit-iframe-loader.html
- [x] Check no error page is shown and m_malicious-site-protection_client-timeout is fired

_Feature 3_
- [x] Load https://privacy-test-pages.site/security/badware/
- [x] Click on "Phishing iFrame Loader"
- [x] Check m_malicious-site-protection_iframe-loaded and m_malicious-site-protection_error-page-shown are fired only once

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -3219,7 +3219,7 @@ class BrowserTabViewModel @Inject constructor(
 
         if (!exempted) {
             if (currentBrowserViewState().maliciousSiteDetected && previousSite?.url == url.toString()) return
-            Timber.tag("Cris").d("Received MaliciousSiteWarning for $url, feed: $feed, exempted: false, clientSideHit: $clientSideHit")
+            Timber.d("Received MaliciousSiteWarning for $url, feed: $feed, exempted: false, clientSideHit: $clientSideHit")
             val params = mapOf(CATEGORY_KEY to feed.name.lowercase(), CLIENT_SIDE_HIT_KEY to clientSideHit.toString())
             pixel.fire(AppPixelName.MALICIOUS_SITE_PROTECTION_ERROR_SHOWN, params)
             loadingViewState.postValue(
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -3210,12 +3210,16 @@ class BrowserTabViewModel @Inject constructor(
     }
 
     override fun onReceivedMaliciousSiteWarning(url: Uri, feed: Feed, exempted: Boolean, clientSideHit: Boolean) {
+        val previousSite = site
         site = siteFactory.buildSite(url = url.toString(), tabId = tabId)
         site?.maliciousSiteStatus = when (feed) {
             MALWARE -> MaliciousSiteStatus.MALWARE
             PHISHING -> MaliciousSiteStatus.PHISHING
         }
+
         if (!exempted) {
+            if (currentBrowserViewState().maliciousSiteDetected && previousSite?.url == url.toString()) return
+            Timber.tag("Cris").d("Received MaliciousSiteWarning for $url, feed: $feed, exempted: false, clientSideHit: $clientSideHit")
             val params = mapOf(CATEGORY_KEY to feed.name.lowercase(), CLIENT_SIDE_HIT_KEY to clientSideHit.toString())
             pixel.fire(AppPixelName.MALICIOUS_SITE_PROTECTION_ERROR_SHOWN, params)
             loadingViewState.postValue(
--- a/app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
@@ -24,8 +24,10 @@ import com.duckduckgo.app.browser.webview.ExemptedUrlsHolder.ExemptedUrl
 import com.duckduckgo.app.browser.webview.RealMaliciousSiteBlockerWebViewIntegration.IsMaliciousViewData
 import com.duckduckgo.app.di.AppCoroutineScope
 import com.duckduckgo.app.di.IsMainProcess
+import com.duckduckgo.app.pixels.AppPixelName
 import com.duckduckgo.app.pixels.remoteconfig.AndroidBrowserConfigFeature
 import com.duckduckgo.app.settings.db.SettingsDataStore
+import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.malicioussiteprotection.api.MaliciousSiteProtection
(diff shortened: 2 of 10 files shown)


cmonfortep (commented):
Code looks good, left a note in Asana about testing.

app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
```diff
@@ -243,7 +270,7 @@ class RealMaliciousSiteBlockerWebViewIntegration @Inject constructor(
             } else {
                 Safe
             }
-            processedUrls.clear()
+            processedUrls[url] = isMalicious
```
cmonfortep, line 273: We have a previous check  `if (checkId == currentCheckId.get())` that validates this is related to the latest site trying to load. If we detect another site has started loading, we return safe to avoid showing the phishing page in the wrong context. However, we are assigning Safe to possible phishing pages here. I believe the right value here is `processedUrls[url] = it`.

app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt
```diff
@@ -397,4 +397,6 @@ enum class AppPixelName(override val pixelName: String) : Pixel.PixelName {
     SET_AS_DEFAULT_PROMPT_CLICK("m_set-as-default_prompt_click"),
     SET_AS_DEFAULT_PROMPT_DISMISSED("m_set-as-default_prompt_dismissed"),
     SET_AS_DEFAULT_IN_MENU_CLICK("m_set-as-default_in-menu_click"),
+
+    MALICIOUS_SITE_DETECTED_IN_IFRAME("m_malicious-site-protection_iframe-loaded"),
```
cmonfortep, line 401: why is this one here if we have `MaliciousSitePixelName`? this should be inside malicious module.

malicious-site-protection/malicious-site-protection-impl/src/main/kotlin/com/duckduckgo/malicioussiteprotection/impl/data/MaliciousSiteRepository.kt
```diff
@@ -79,18 +84,23 @@ class RealMaliciousSiteRepository @Inject constructor(
...
-                val feed = when (it.feed.uppercase()) {
-                    PHISHING.name -> PHISHING
-                    MALWARE.name -> MALWARE
-                    else -> null
-                }
-                if (feed != null) {
-                    Match(it.hostname, it.url, it.regex, it.hash, feed)
-                } else {
-                    null
+            withTimeout(1000) {
```
cmonfortep, line 87: I haven't found why 1000 is the timeout here in the pixel definition, or in Asana. Do we have it somewhere? would be good to move it into a const, and leave a comment or link with context.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3210,12 +3210,16 @@ class BrowserTabViewModel @Inject constructor(
...
+        val previousSite = site
         site = siteFactory.buildSite(url = url.toString(), tabId = tabId)
         site?.maliciousSiteStatus = when (feed) {
             MALWARE -> MaliciousSiteStatus.MALWARE
             PHISHING -> MaliciousSiteStatus.PHISHING
         }
+
         if (!exempted) {
+            if (currentBrowserViewState().maliciousSiteDetected && previousSite?.url == url.toString()) return
+            Timber.tag("Cris").d("Received MaliciousSiteWarning for $url, feed: $feed, exempted: false, clientSideHit: $clientSideHit")
```
cmonfortep, line 3222: if you want to keep this log, we should use a the right tag.

app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
```diff
@@ -137,6 +140,9 @@ class RealMaliciousSiteBlockerWebViewIntegration @Inject constructor(
         if (!isEnabled()) {
             return IsMaliciousViewData.Safe
         }
+
+        Timber.d("shouldIntercept ${request.url}")
```
cmonfortep, line 144: 👀

app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
```diff
@@ -147,35 +153,49 @@ class RealMaliciousSiteBlockerWebViewIntegration @Inject constructor(
...
-            processedUrls.remove(decodedUrl)
-            Timber.tag("PhishingAndMalwareDetector").d("Already intercepted, skipping $decodedUrl")
-            return IsMaliciousViewData.Safe
-        }
-
         val exemptedUrl = exemptedUrlsHolder.exemptedMaliciousUrls.firstOrNull { it.url.toString() == decodedUrl }
 
         if (exemptedUrl != null) {
-            Timber.tag("MaliciousSiteDetector").d("Previously exempted, skipping $decodedUrl as ${exemptedUrl.feed}")
+            Timber.d("Previously exempted, skipping $decodedUrl as ${exemptedUrl.feed}")
```
cmonfortep, line 159: I see we've removed tag, not sure if intentional

cmonfortep (approved):
LGTM

Few comments, and I see also a bunch of logs that not sure if we want to keep, but we should use the right tag for consistency. (I've commented in some of them, but there are a more).

app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
CrisBarreiro: That makes a lot of sense, thanks!

app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt
CrisBarreiro: Because `MaliciousSitePixelName` belongs to the MSP implementation module, and the MSP module knows nothing about iframes, that happens on the browser level

malicious-site-protection/malicious-site-protection-impl/src/main/kotlin/com/duckduckgo/malicioussiteprotection/impl/data/MaliciousSiteRepository.kt
CrisBarreiro: @cmonfortep, I believe this was discussed on a meeting. We recently created a [task](https://app.asana.com/0/72649045549333/1209504644920163) to document timeouts and it looks like all other platforms are using a different value, so I'll probably need to update this one. We also want to update the timeout for dataset downloads, and I've created a [task](https://app.asana.com/0/72649045549333/1209533874979757) for it, so for now I'm just extracting it to a constant. I've asked whether we want to update this timeout in particular to 5s, so in case the answer is yes, I'll use the task for dataset downloads timeout to also update this one

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
CrisBarreiro: I removed the tag, since we don't seem to be setting any specific one for this class

app/src/main/java/com/duckduckgo/app/browser/webview/MaliciousSiteBlockerWebViewIntegration.kt
CrisBarreiro: Yes, similar to https://github.com/duckduckgo/Android/pull/5604#discussion_r1977087683, we don't use tag for other logs in this class

PR #5618, opened 2025-02-07
Allow users to select password inside DDG from other apps

cmonfortep:
Task/Issue URL: https://app.asana.com/0/1200156640058969/1209114881668819 

### Description
Introduces:
- Asking the user for authentication, if required, after interacting with any suggestion.
- Allowing user to manually select a credential inside our app from a simplified list.

### Steps to test this PR

(heuristics have not changed, so we are good just testing the autofilling part of a few scenarios).

Add some credentials for apps you will test from this task https://app.asana.com/0/1149059203486286/1209279712661812

_Feature 1_
- [ ] install the app
- [ ] Select DuckDuckGo as System Level Autofill
- [ ] Test anything from this task https://app.asana.com/0/1149059203486286/1209279712661812
- [ ] When selecting the credential suggestion, it should request you to authenticate.
- [ ] After authentication, it should autofill

_Feature 1_
- [ ] Test another from this task https://app.asana.com/0/1149059203486286/1209279712661812
- [ ] Instead click on "Search Password"
- [ ] It should open automatically without authentication since you recently did previous step
- [ ] try searching to validate it works
- [ ] now select a credential from the list
- [ ] ensure it autofills correctly

### UI changes
| Before  | After |
| ------ | ----- |


cmonfortep:
* **#5618** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/service/AutofillSimpleCredentialsListFragment.kt
```diff
@@ -0,0 +1,266 @@
...
+
+        viewModel.onViewCreated()
+    }
+
+    private suspend fun credentialsListUpdated(
+        credentials: List<LoginCredentials>,
+        credentialSearchQuery: String,
+    ) {
+        if (credentials.isEmpty() && credentialSearchQuery.isEmpty()) {
+            showEmptyCredentialsPlaceholders()
```
CDRussell, line 194: When there are no saved passwords, this UX is weird. Can’t see explicitly in Asana or Figma what the expectations are for empty state, but I think if there’s no saved passwords, we should prevent the user getting to this screen in the first place.

There’s no value in bringing them here just to tell them we’ve no saved passwords, and the buttons like “Import Passwords From Google” are just undesirable to see in this context.

Screenshot_20250211_103414

CDRussell:
>When selecting the credential suggestion, it should request you to authenticate.

@cmonfortep i’m not seeing the authentication challenge

CDRussell:
This is maybe a limitation of Autofill APIs, but if you choose to search and pick a credential you then lose the ability to search again in the current autofill context. Instead of search, it is replaced with the credential you chose.

Maybe known already, and not a blocker, but flagging in case it’s easy to fix.

Screen Recording 2025-02-11 at 10 48 49

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/service/AutofillProviderFillSuggestionActivity.kt
```diff
@@ -0,0 +1,200 @@
...
+            if (firstDetectedField != null) {
+                return@firstNotNullOfOrNull Pair(node, firstDetectedField)
+            }
+            return@firstNotNullOfOrNull null
+        }
+
+        if (detectedNode == null) {
+            return
+        }
+
```
CDRussell, line 157: nit: we have this identical logic in two activities. it’s non-trivial, and can’t be easily tested in the activity. consider extracting it out to its own class?

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/service/AutofillProviderFillSuggestionActivity.kt
```diff
@@ -0,0 +1,200 @@
...
+        val dataset = buildDataset(fields, credential)
+
+        val resultIntent = Intent().apply {
+            putExtra(AutofillManager.EXTRA_AUTHENTICATION_RESULT, dataset)
+        }
+        this.setResult(Activity.RESULT_OK, resultIntent)
+        this.finish()
+    }
+
+    private fun buildDataset(
```
CDRussell, line 169: Same non-blocking suggestion here:

- the whole of `buildDataset()` is duplicated across two places, non trivial and not easy to test 
- extract as a class ?

autofill/autofill-impl/src/main/res/values/donottranslate.xml
```diff
@@ -16,4 +16,5 @@
 
 <resources>
     <string name="autofill_service_suggestion_search_passwords">Search Passwords</string>
+    <string name="autofill_service_choose_password_activity">Choose a saved password</string>
```
CDRussell, line 19: nit: change to `Change a Saved Password`

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/service/AutofillProviderDeviceAuth.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    override suspend fun recordSuccessfulAuthorization() {
+        store.edit {
+            it[lastAuthTimeKey] = timeProvider.currentTimeMillis()
+        }
+    }
+
+    companion object {
+        private const val AUTH_GRACE_EXTENDED_PERIOD_MS = 180_000
```
CDRussell, line 65: how did we land on 180s for this? not saying it’s bad (i’m not sure what my intuition would be if i was picking it) but curious why this value…was it similar to other providers for instance?

cmonfortep: I don't know if there's a standard here. Some apps allow users to decide, others will go for a default value but I haven't checked.

I just landed in the same value we use for import passwords from google as a way to avoid asking on every field in the same app. I will leave it documented in https://app.asana.com/0/1200156640058969/1209180548400497, to indicate there's no reasoning further than that, and we can change if we are not happy with it.

autofill/autofill-impl/src/main/res/values/donottranslate.xml
cmonfortep: using "Select a Saved Password" after copy review

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/service/AutofillSimpleCredentialsListFragment.kt
CDRussell: Thanks for tidying up the empty state.

One last nit here: we can skip the auth prompt when passwords are empty like we do in the main password management screen. I already think it’s a questionable UX to let users visit this screen when we have no passwords to offer, but it’s a double insult to make them authenticate first to then tell them there’s nothing here for them 😄 

Feel free to add in later, though. 
Screen Recording 2025-02-12 at 15 22 46

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/service/AutofillServiceExtensions.kt
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.autofill.impl.service
+
+import com.duckduckgo.autofill.impl.service.AutofillFieldType.UNKNOWN
+
+fun List<AutofillRootNode>.findBestFillableNode(): AutofillRootNode? {
```
CDRussell, line 21: nice, thanks for pulling this out!

CDRussell (approved):

PR #5627, opened 2025-02-10
Preparations for UI Experiments

malmstein:
Task/Issue URL: https://app.asana.com/0/488551667048375/1209289914504437

### Steps to test this PR

_Enable Experimental UI_
- [ ] Install app in Internal flavour
- [ ] Open Appearance screen
- [ ] Verify “Enable Experimental UI” is visible
- [ ] Toggle it on and restart the app
- [ ] Verify has a pink tint
- [ ] Open Appearance screen and toggle the experimental feature offf
- [ ] Verify app is back to the original colors after restart

_Enable Experimental UI_
- [ ] Install app in Play flavour
- [ ] Open Appearance screen
- [ ] Verify “Enable Experimental UI” is not visible


malmstein:
* **#5627** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5627?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

common/common-ui/src/main/java/com/duckduckgo/common/ui/Theming.kt
```diff
@@ -89,6 +74,8 @@ fun AppCompatActivity.getThemeId(theme: DuckDuckGoTheme): Int {
     return when (theme) {
         SYSTEM_DEFAULT -> getSystemDefaultTheme()
```
LukasPaczos, line 75: `getSystemDefaultTheme` also needs to be updated to return either the regular or experiment themes.

app/src/main/java/com/duckduckgo/app/appearance/AppearanceActivity.kt
```diff
@@ -82,6 +87,28 @@ class AppearanceActivity : DuckDuckGoActivity() {
...
+            .setNegativeButton(R.string.appearanceNightModeDialogSecondaryCTA)
+            .addEventListener(
+                object : TextAlertDialogBuilder.EventListener() {
+                    override fun onPositiveButtonClicked() {
+                        FireActivity.triggerRestart(baseContext, false)
+                    }
+
+                    override fun onNegativeButtonClicked() {
+                        // no-op
+                    }
```
LukasPaczos, line 106: Why would we ever not want to relaunch the app? Looks like the effects of continuing on without restarting are very mixed.

What if we added to the toggle's description that "changing this setting will relaunch the app" and skip the dialog, to make it more straightforward?

common/common-ui-experiments/src/main/java/com/duckduckgo/common/ui/experiments/BrowserThemingFeature.kt
```diff
@@ -0,0 +1,48 @@
...
+@ContributesRemoteFeature(
+    scope = AppScope::class,
+    featureName = "experimentalBrowserTheming",
+)
+interface BrowserThemingFeature {
+    @Toggle.DefaultValue(false)
+    fun self(): Toggle
+
+    @Toggle.DefaultValue(false)
+    fun experimentalUI(): Toggle
```
LukasPaczos, line 32: If this is the feature flag that we'll use for the experiment, I'd recommend being more explicit with its name, because only this name will be part of the metric pixels. Maybe something like `OAJan2025ExperimentalUI`.

app/src/main/java/com/duckduckgo/app/appearance/AppearanceActivity.kt
LukasPaczos, line 1: In this activity we also need to update the `launchThemeSelector` function:
https://github.com/duckduckgo/Android/blob/3acaa58e2929cbf882df96c19c8ad902354aa269/app/src/main/java/com/duckduckgo/app/appearance/AppearanceActivity.kt#L183-L194

The 4 and 5 indices (experiments) should just point to regular values.

common/common-ui/src/main/java/com/duckduckgo/common/ui/store/ThemingSharedPreferences.kt
```diff
@@ -56,16 +57,27 @@ class ThemingSharedPreferences @Inject constructor(
             when (theme) {
                 DuckDuckGoTheme.SYSTEM_DEFAULT -> THEME_SYSTEM_DEFAULT
                 DuckDuckGoTheme.LIGHT -> THEME_LIGHT
-                DuckDuckGoTheme.DARK -> THEME_DARK
+                else -> THEME_DARK
```
LukasPaczos, line 60: In theory we should never receive the experiment values here, but having a catch-all else clause might be a source of future bugs. I'd rather us handle the experiment cases explicitly and map them to regular light/dark values.

LukasPaczos (commented):
We also need to handle the `DARK_EXPERIMENT` value in
https://github.com/duckduckgo/Android/blob/3acaa58e2929cbf882df96c19c8ad902354aa269/common/common-ui/src/main/java/com/duckduckgo/common/ui/DuckDuckGoActivity.kt#L90-L102

app/src/main/java/com/duckduckgo/app/appearance/AppearanceActivity.kt
malmstein: Yeah why not, for this specific scenario we can keep it simple

common/common-ui-experiments/src/main/java/com/duckduckgo/common/ui/experiments/BrowserThemingFeature.kt
malmstein: It’s just a feature flag we’ll use internally. We are far from any actual experiment. When the time comes, we’ll create proper variants, etc...

common/common-ui/src/main/java/com/duckduckgo/common/ui/Theming.kt
malmstein: During internal testing we only want to change Light / Dark. Users on system theming won’t be able to see the experimental colors.

LukasPaczos: This sounds somewhat counter-intuitive for me. The new flag governs whether the experimental light/dark theme is applied, we shouldn't require from internal users any additional implicit steps. The default setting is the "system theme", and in the end it applies one of the light or dark themes, so in my opinion it should respect the experiment flag as well.

common/common-ui-experiments/src/main/java/com/duckduckgo/common/ui/experiments/BrowserThemingFeature.kt
LukasPaczos: No problem, just an fyi.

common/common-ui/src/main/java/com/duckduckgo/common/ui/store/ThemingSharedPreferences.kt
LukasPaczos: I understand that, but if anything changes in the code that calls this function which would make it suddenly pass the experimental value, it would go unnoticed.

common/common-ui/src/main/java/com/duckduckgo/common/ui/Theming.kt
malmstein: "we shouldn't require from internal users any additional implicit steps”. this is still to be decided, for now we will keep it simple and iterate in the project.

app/src/main/java/com/duckduckgo/app/appearance/AppearanceActivity.kt
LukasPaczos, line 1: One more small change needed:
https://github.com/duckduckgo/Android/blob/3acaa58e2929cbf882df96c19c8ad902354aa269/app/src/main/java/com/duckduckgo/app/appearance/AppearanceActivity.kt#L170-L180

`currentTheme` should be mapped back to light/dark when experimental is provided. Otherwise, nothing is pre-selected when the experiment is enabled and radio menu opens.

LukasPaczos (approved):
Thanks for addressing the comments, let's merge and continue but please cut follow up tasks for:
- https://github.com/duckduckgo/Android/pull/5627#discussion_r1952368697
- https://github.com/duckduckgo/Android/pull/5627#pullrequestreview-2612172945

if not addressed here.

LukasPaczos:
Also double checking test instructions:
```
1. Install app in release flavour
2. Open Appearance screen
3. Verify “Enable Experimental UI” is not visible
```

I think you meant "Install app in **external (play/fdroid) flavour**", because `internalRelease` still has the flag present, but I assume that's expected. Just flagging in case that's not the case.

malmstein:
Thanks for the thorough review @LukasPaczos

PR #5631, opened 2025-02-11
Fix for the backwards typing bug

0nko:
Task/Issue URL: https://app.asana.com/0/488551667048375/1209104073191021

### Description

This PR fixes the backwards typing bug (typing from right-to-left) by replacing the `shouldMoveCaretToStart` ViewState parameter with a command, so that it's executed only once when the omnibar loses focus. 

I've also removed `shouldMoveCaretToEnd` because it was not used (enabled) anywhere in the code. 

### Steps to test this PR

_The fix_
- [x] Start with a single NTP
- [x] Type something and verify it's left-to-right
- [x] Delete the text
- [x] Pull down the notification bar and change the dark mode setting
- [x] Pull up the notification bar
- [x] Type something in the address bar
- [x] Notice the text is correctly written left-to-right

_Long URL shown from the start_
- [x] Navigate to some website where the URL is longer than the address bar
- [x] Tap on the webview so that the address bar loses focus
- [x] Verify that the URL is shown from the beginning, not just its end

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -2311,13 +2311,6 @@ class BrowserTabViewModelTest {
     }
 
     @Test
-    fun whenUserSubmitsQueryThenCaretDoesNotMoveToTheEnd() {
-        whenever(mockOmnibarConverter.convertQueryToUrl("foo", null)).thenReturn("foo.com")
-        testee.onUserSubmittedQuery("foo")
-        assertFalse(omnibarViewState().shouldMoveCaretToEnd)
-    }
-
-    @Test
     fun whenUserRequestedToOpenNewTabThenGenerateWebViewPreviewImage() {
         testee.userRequestedOpeningNewTab()
         assertCommandIssued<Command.GenerateWebViewPreviewImage>()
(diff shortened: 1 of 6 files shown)


joshliebe:
@0nko I tested this quite a lot and the issue is fixed 🎉 But it has introduced some new behavior.

If you follow the "Long URL shown from the start” test steps and then focus the address bar (So the caret moves to the end), then change the theme, the caret doesn’t go back to the beginning. 

This isn’t a huge issue for this specific use-case but it may introduce unwanted behavior elsewhere.

0nko:
@joshliebe I've checked the behavior and just want to make sure I understand it correctly -- after you perform those steps and click on the address bar you see the URL selected and it's not showing the beginning? If that's so, then that's expected and it's how it worked before.

joshliebe:
> @joshliebe I've checked the behavior and just want to make sure I understand it correctly -- after you perform those steps and click on the address bar you see the URL selected and it's not showing the beginning? If that's so, then that's expected and it's how it worked before.

@0nko I’ve added before and after videos. Before, the URL is reset to the beginning after changing the theme. After, it no longer resets:

https://github.com/user-attachments/assets/5cf06450-f6fe-44ef-af51-023c8158d77b

https://github.com/user-attachments/assets/7eb1e60b-9672-4e40-b896-bce5ba0f74a8

0nko:
I think I fixed the issue and it works fine on my device now. Ready for another round!

joshliebe (approved):
It’s working as expected now, thanks for fixing it 🙂

PR #5645, opened 2025-02-13
Autofill system setting, internal quick access

cmonfortep:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209397886769939 

### Description
Introduce a quick access to system settings from our internal Autofill Setting

### Steps to test this PR

_Feature 1_
- [ ] install app
- [ ] go to autofill internal settings
- [ ] Click on "Autofill System Settings" list item at the top
- [ ] ensure you Access Autofill System settings

### UI changes
| Before  | After |
| ------ | ----- |


CDRussell (approved):
Nit: it does nothing when i’m already the provider. so it’s great to shortcut into the screen when we’re not already the provider, but it’s weird when you are as it does nothing. i actually didn’t know if i was or not at that point, and kinda wanted to double check. this seems fixable?

Approving since it’s an improvement though.

PR #5697, opened 2025-02-26
Make autofill username comparison case insensitive

CDRussell:
Task/Issue URL: https://app.asana.com/0/608920331025315/1209488709393051 

### Description
Changes how we compare usernames to make them case-insensitive. 

Autofill subfeature `ignoreCaseOnUsernameComparisons` guards the new behaviour.
- new behaviour is enabled by default 
- disable the flag to revert to old behaviour

#### Where this change applies

- If multiple credentials saved which are identical apart from casing on username, these will be considered duplicates from the credential picker view (meaning they’ll be de-duped)
- If you already have a credential saved and are trying to save a new one with same username, the comparison between usernames will be case insensitive. So if `user` is saved and `USER` comes from the JS, these will be treated as equivalent
- The password management view will always show all saved credentials (no de-duping in this view)

#### De-duping credential picker
- [x] fresh `internal` install
- [x] Visit `Autofill Dev Settings` (from `Settings -> Autofill Dev Settings`)
- [x] Tap on `Add mixed case username duplicates`
- [x] Visit https://autofill.me/form/login-simple, verify you are prompted to autofill with **only 1 option offered**
- [x] Agree to autofill, and confirm it autofills successfully

#### Across subdomains
- [x] Visit password management screen, and choose one of the credentials to edit
- [x] Change its URL to have a subdomain (e.g. `example.autofill.me`) but keep the username
- [x] Add a new credential (`username=john`, `password=password`, `url=example.autofill.me`) and save
- [x] Return to https://autofill.me/form/login-simple (and refresh)
- [x] Verify you see 1 option `From This Website`, and 1 option `From example.autofill.me` (this verifies that it de-duped the username where case was different but credential was otherwise identical, and still included the credential for a subdomain with a different username) 

#### Prompted to update password (not save a new one) when only difference is casing on username
- [x] Visit https://autofill.me/form/registration-username
- [x] For username, enter `USERname` (match the case exactly)
- [x] For password, enter something different from what’s stored (e.g., `new-password` or autogenerate it)
- [x] Tap `Register` button. Verify you are prompted to `Update password` for the username
- [ ] Agree. Now visit password management page and verify the password has been updated for all 3 entries (by design, it updates multiple matches)

#### Reverting to old behaviour
- [x] Tap on `Settings -> Feature Flag Inventory`, and disable `ignoreCaseOnUsernameComparisons`
- [x] Verify the production behaviour happens. e.g., 
    - you get offered all saved passwords to autofill, even the “duplicates”
    - if you register using a username containing case differences, you are prompted to save rather than update

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStore.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStore.kt
@@ -31,6 +31,7 @@ import com.duckduckgo.autofill.impl.securestorage.WebsiteLoginDetailsWithCredent
 import com.duckduckgo.autofill.impl.store.InternalAutofillStore
 import com.duckduckgo.autofill.impl.ui.credential.saving.declines.AutofillDeclineStore
 import com.duckduckgo.autofill.impl.urlmatcher.AutofillUrlMatcher
+import com.duckduckgo.autofill.impl.username.AutofillUsernameComparer
 import com.duckduckgo.autofill.store.AutofillPrefsStore
 import com.duckduckgo.autofill.store.LastUpdatedTimeProvider
 import com.duckduckgo.autofill.sync.SyncCredentialsListener
@@ -58,6 +59,7 @@ class SecureStoreBackedAutofillStore @Inject constructor(
     private val autofillUrlMatcher: AutofillUrlMatcher,
     private val syncCredentialsListener: SyncCredentialsListener,
     private val autofillFeature: AutofillFeature,
+    private val usernameComparer: AutofillUsernameComparer,
     passwordStoreEventListenersPlugins: PluginPoint<PasswordStoreEventListener>,
 ) : InternalAutofillStore, AutofillDeclineStore {
 
@@ -171,7 +173,7 @@ class SecureStoreBackedAutofillStore @Inject constructor(
 
         val matchingCredentials = secureStorage.websiteLoginDetailsWithCredentialsForDomain(url)
             .firstOrNull()
-            ?.filter(filter)
+            ?.filter { filter(it) }
 
         if (matchingCredentials.isNullOrEmpty()) {
             Timber.w("Cannot update credentials as no credentials were found for %s", url)
@@ -182,9 +184,19 @@ class SecureStoreBackedAutofillStore @Inject constructor(
 
         var updatedCredentials: WebsiteLoginDetailsWithCredentials? = null
 
-        matchingCredentials.forEach {
-            val modifiedDetails = it.details.copy(username = credentials.username, lastUpdatedMillis = lastUpdatedTimeProvider.getInMillis())
-            val modified = it.copy(password = credentials.password, details = modifiedDetails)
+        matchingCredentials.forEach { existingCredential ->
+            val modifiedDetails = existingCredential.details.copy(
+                // only update username if that was the update type
+                username = if (updateType == Username) credentials.username else existingCredential.details.username,
+                lastUpdatedMillis = lastUpdatedTimeProvider.getInMillis(),
+            )
+
+            val modified = existingCredential.copy(
+                // only update password if that was the update type
+                password = if (updateType == Password) credentials.password else existingCredential.password,
+                details = modifiedDetails,
+            )
+
             updatedCredentials = secureStorage.updateWebsiteLoginDetailsWithCredentials(modified)?.also {
                 syncCredentialsListener.onCredentialUpdated(it.details.id!!)
             }
@@ -193,10 +205,13 @@ class SecureStoreBackedAutofillStore @Inject constructor(
         return updatedCredentials?.toLoginCredentials()
     }
 
-    private fun filterMatchingUsername(credentials: LoginCredentials): (WebsiteLoginDetailsWithCredentials) -> Boolean =
-        { it.details.username == credentials.username }
+    private fun filterMatchingUsername(credentials: LoginCredentials): suspend (WebsiteLoginDetailsWithCredentials) -> Boolean = {
+        // we only update password when usernames are equal
+        usernameComparer.isEqual(it.details.username, credentials.username)
+    }
 
-    private fun filterMatchingPassword(credentials: LoginCredentials): (WebsiteLoginDetailsWithCredentials) -> Boolean = {
+    private fun filterMatchingPassword(credentials: LoginCredentials): suspend (WebsiteLoginDetailsWithCredentials) -> Boolean = {
+        // we only update username if stored username is null or empty, and only when password matches
         it.password == credentials.password && it.details.username.isNullOrEmpty()
     }
 
@@ -340,11 +355,12 @@ class SecureStoreBackedAutofillStore @Inject constructor(
         }
     }
 
-    private fun usernameMatch(
+    private suspend fun usernameMatch(
         credentials: WebsiteLoginDetailsWithCredentials,
         username: String?,
     ): Boolean {
-        return credentials.details.username != null && credentials.details.username == username
+        // special case where we don't want to consider two null usernames as equals
+        return credentials.details.username != null && usernameComparer.isEqual(credentials.details.username, username)
     }
 
     private fun usernameMissing(
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/deduper/AutofillDeduplicationUsernameAndPasswordMatcher.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/deduper/AutofillDeduplicationUsernameAndPasswordMatcher.kt
@@ -17,19 +17,22 @@
 package com.duckduckgo.autofill.impl.deduper
 
 import com.duckduckgo.autofill.api.domain.app.LoginCredentials
+import com.duckduckgo.autofill.impl.username.AutofillUsernameComparer
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesBinding
 import javax.inject.Inject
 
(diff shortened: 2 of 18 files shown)


CDRussell:
* **#5697** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5697?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/5697?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStore.kt
```diff
@@ -171,7 +173,7 @@ class SecureStoreBackedAutofillStore @Inject constructor(
 
         val matchingCredentials = secureStorage.websiteLoginDetailsWithCredentialsForDomain(url)
             .firstOrNull()
-            ?.filter(filter)
+            ?.filter { filter(it) }
```
CDRussell, line 176: tweak to support the filter’s need to call a suspend function

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStore.kt
```diff
@@ -193,10 +205,13 @@ class SecureStoreBackedAutofillStore @Inject constructor(
         return updatedCredentials?.toLoginCredentials()
     }
 
-    private fun filterMatchingUsername(credentials: LoginCredentials): (WebsiteLoginDetailsWithCredentials) -> Boolean =
-        { it.details.username == credentials.username }
+    private fun filterMatchingUsername(credentials: LoginCredentials): suspend (WebsiteLoginDetailsWithCredentials) -> Boolean = {
+        // we only update password when usernames are equal
+        usernameComparer.isEqual(it.details.username, credentials.username)
```
CDRussell, line 210: use centralised `usernameComparer` instead of a direct string comparison

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStore.kt
```diff
@@ -340,11 +355,12 @@ class SecureStoreBackedAutofillStore @Inject constructor(
...
     }
 
-    private fun usernameMatch(
+    private suspend fun usernameMatch(
         credentials: WebsiteLoginDetailsWithCredentials,
         username: String?,
     ): Boolean {
-        return credentials.details.username != null && credentials.details.username == username
+        // special case where we don't want to consider two null usernames as equals
+        return credentials.details.username != null && usernameComparer.isEqual(credentials.details.username, username)
```
CDRussell, line 363: use `usernameComparer` instead of a direct string comparison

note, here we also need to preserve the previous code’s special case where we don’t want to consider two null usernames as equals here

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/deduper/AutofillDeduplicationUsernameAndPasswordMatcher.kt
```diff
@@ -17,19 +17,22 @@
...
 @ContributesBinding(AppScope::class)
-class RealAutofillDeduplicationUsernameAndPasswordMatcher @Inject constructor() : AutofillDeduplicationUsernameAndPasswordMatcher {
+class RealAutofillDeduplicationUsernameAndPasswordMatcher @Inject constructor(
+    private val usernameComparer: AutofillUsernameComparer,
+) : AutofillDeduplicationUsernameAndPasswordMatcher {
 
-    override fun groupDuplicateCredentials(logins: List<LoginCredentials>): Map<Pair<String?, String?>, List<LoginCredentials>> {
-        return logins.groupBy { it.username to it.password }
+    override suspend fun groupDuplicateCredentials(logins: List<LoginCredentials>): Map<Pair<String?, String?>, List<LoginCredentials>> {
+        return usernameComparer.groupUsernamesAndPasswords(logins)
```
CDRussell, line 36: use `usernameComparer` instead of its own grouping logic where it would not consider case sensitivity of username

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/username/AutofillUsernameComparer.kt
```diff
@@ -0,0 +1,58 @@
...
+    private val dispatchers: DispatcherProvider,
+) : AutofillUsernameComparer {
+
+    override suspend fun isEqual(username1: String?, username2: String?): Boolean {
+        return withContext(dispatchers.io()) {
+            if (username1 == null && username2 == null) return@withContext true
+            if (username1 == null) return@withContext false
+            if (username2 == null) return@withContext false
+
+            username1.equals(username2, ignoreCase = autofillFeature.ignoreCaseOnUsernameComparisons().isEnabled())
```
CDRussell, line 48: case insensitive only if `autofillFeature.ignoreCaseOnUsernameComparisons` is enabled

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/FakeSecureStorage.kt
```diff
@@ -29,7 +29,7 @@ class FakeSecureStore(
     private val urlMatcher: AutofillUrlMatcher,
 ) : SecureStorage {
 
-    private val credentials = mutableListOf<WebsiteLoginDetailsWithCredentials>()
+    private val credentials = mutableMapOf<Long, WebsiteLoginDetailsWithCredentials>()
```
CDRussell, line 32: making this “fake storage” more useful for testing

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/deduper/AutofillLoginDeduplicator.kt
```diff
@@ -23,7 +23,7 @@ import javax.inject.Inject
 
 interface AutofillLoginDeduplicator {
 
-    fun deduplicate(
+    suspend fun deduplicate(
```
cmonfortep, line 26: Ditto: not ideal that we are forced to make this suspend.

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStoreTest.kt
```diff
@@ -412,6 +415,67 @@ class SecureStoreBackedAutofillStoreTest {
         }
     }
 
+    @Test
+    fun whenUpdatingPasswordWithUsernameInDifferentCasingThenOriginalUsernameUntouched() = runTest {
```
cmonfortep, line 419: I think this test just validate the opposite to the 4th scenario in the test cases:

- Visit https://autofill.me/form/registration-username
- For username, enter USERname (match the case exactly)
- For password, enter something different from what’s stored (e.g., new-password or autogenerate it)
- Tap Register button. Verify you are prompted to Update password for the username
- Agree. Now visit password management page and verify the password has been updated for all 3 entries (by design, it updates multiple matches)

Seems we expect to update all entries, and in the unit test is expected to keep the original.

cmonfortep (commented):
Overall LGTM 

I seems on test scenario conflicts with the expectation, so we should validate what's the right one.
Leaving for consideration (optional) removing the `suspend` from the username comparer.

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/SecureStoreBackedAutofillStoreTest.kt
CDRussell: >I think this test just validate the opposite to the 4th scenario in the test cases

No, this is different. 
- the scenario from the PR description is about checking the **password** is updated.
- this unit test scenario is about ensuring the **username** is not updated (only the password is updated)

cmonfortep: Discussed offline and password is updated. The username is not, which is the intended behavior.

cmonfortep (approved):
LGTM thanks!

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/username/AutofillUsernameComparer.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.duckduckgo.autofill.api.AutofillFeature
+import com.duckduckgo.autofill.api.domain.app.LoginCredentials
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import javax.inject.Inject
+import kotlinx.coroutines.withContext
+
+interface AutofillUsernameComparer {
+    suspend fun isEqual(
```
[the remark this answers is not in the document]
CDRussell: it’s a fair question. especially since later if we remove the FF check it doesn’t need to be `suspend` any more.

>consider just checking you are not in main (checkMainThread()), and crash otherwise

I don’t want to rely on runtime crashes for this to be found though, especially since the flow is already called from `suspend` functions. and if anything else later decides to call this `isEqual` later it should be aware it’s potentially doing disk IO rather than being surprised that it does it under the hood. so all things considered, my preference here is doing what makes the code safest against unexpected crashes.

PR #5702, opened 2025-02-27
Additional "default browser" prompts: refactored remote config parsing code to match expected schema

LukasPaczos:
Task/Issue URL: https://app.asana.com/0/1208671518894266/1209523191422601

### Description

Adjust the implementation to be able to use config entry that matches the schema.

### Steps to test this PR
Test config:
```json
    "defaultBrowserPrompts": {
      "state": "enabled",
      "exceptions": [],
      "features": {
        "defaultBrowserAdditionalPrompts202501": {
          "state": "enabled",
          "settings": {
            "activeDaysUntilStage1": "1",
            "activeDaysUntilStage2": "3",
            "activeDaysUntilStop": "5"
          },
          "cohorts": [
            {
              "name": "control",
              "weight": 0
            },
            {
              "name": "variant1",
              "weight": 0
            },
            {
              "name": "variant2",
              "weight": 1
            },
            {
              "name": "variant3",
              "weight": 0
            }
          ]
        }
      }
    }
```

1. Load this config: https://www.jsonblob.com/api/1344630015471575040
2. Move the device's calendar date by one day.
3. Restart/resume the app - a CTA should be shown.
4. Move the device's calendar date by one day and restart/resume the app, **twice** - a new menu item should be added.
5. Move the device's calendar date by one day and restart/resume the app, **twice** - menu should be removed.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
@@ -123,14 +123,23 @@ class DefaultBrowserPromptsExperimentImpl @Inject constructor(
         initialValue = false,
     )
 
+    /**
+     * Model used to parse remote config setting. All values are integer strings, for example "1" or "20".
+     */
     @VisibleForTesting
-    data class FeatureSettings(
+    data class FeatureSettingsConfigModel(
+        val activeDaysUntilStage1: String,
+        val activeDaysUntilStage2: String,
+        val activeDaysUntilStop: String,
+    )
+
+    private data class FeatureSettings(
         val activeDaysUntilStage1: Int,
         val activeDaysUntilStage2: Int,
         val activeDaysUntilStop: Int,
     )
 
-    private val featureSettingsJsonAdapter = moshi.adapter(FeatureSettings::class.java)
+    private val featureSettingsJsonAdapter = moshi.adapter(FeatureSettingsConfigModel::class.java)
 
     /**
      * Caches deserialized [Toggle.getSettings] for [DefaultBrowserPromptsFeatureToggles.defaultBrowserAdditionalPrompts202501].
@@ -372,7 +381,7 @@ class DefaultBrowserPromptsExperimentImpl @Inject constructor(
     private suspend fun DefaultBrowserPromptsFeatureToggles.parseFeatureSettings(): FeatureSettings? = withContext(dispatchers.io()) {
         defaultBrowserAdditionalPrompts202501().getSettings()?.let { settings ->
             try {
-                featureSettingsJsonAdapter.fromJson(settings)
+                featureSettingsJsonAdapter.fromJson(settings)?.toFeatureSettings()
             } catch (e: Exception) {
                 Timber.e(e)
                 null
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
```diff
@@ -123,14 +123,23 @@ class DefaultBrowserPromptsExperimentImpl @Inject constructor(
...
 
+    /**
+     * Model used to parse remote config setting. All values are integer strings, for example "1" or "20".
+     */
     @VisibleForTesting
-    data class FeatureSettings(
+    data class FeatureSettingsConfigModel(
+        val activeDaysUntilStage1: String,
+        val activeDaysUntilStage2: String,
+        val activeDaysUntilStop: String,
```
aitorvs, line 133: Why do we use strings when this are `Int` types in reality?

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
```diff
@@ -421,6 +430,13 @@ class DefaultBrowserPromptsExperimentImpl @Inject constructor(
         pixel.fire(it.pixelName, it.params)
     }
 
+    @Throws(NumberFormatException::class)
+    private fun FeatureSettingsConfigModel.toFeatureSettings() = FeatureSettings(
+        activeDaysUntilStage1 = activeDaysUntilStage1.toInt(),
+        activeDaysUntilStage2 = activeDaysUntilStage2.toInt(),
+        activeDaysUntilStop = activeDaysUntilStop.toInt(),
+    )
+
```
aitorvs, line 439: Follow up to the comment above, why do we want to do this when we can just have an `Int` in the config and parse it directly to the appropriate type?

app/src/main/java/com/duckduckgo/app/browser/defaultbrowsing/prompts/DefaultBrowserPromptsExperimentImpl.kt
aitorvs: I am not sure it's true, look at `ampLinks.settings`, it has `deepExtractionTimeout` which is an Int and other params that are eg. boolean values.

I am not sure if "just an experiment" is a good argument for not doing things the right way tbf.

LukasPaczos: `ampLinks.settings` is top-level feature `settings` object, the problem we hit here is with the nested feature `settings` object - the schema unit tests were asking for strictly string values.

Looks like we'd need to tweak the way we define the `SubFeature` to allow the `SettingsType` to be at least `Record<string, string | number>` (if not just `any`), right now this seems to [default to `Record<string, string>`](https://github.com/duckduckgo/privacy-configuration/blob/10cec00018890d882b42877b65a76d187d888234/schema/feature.ts#L20) which would explain the required `string` value by unit tests. I can try a couple of things to make it work.

aitorvs: I think it's fine, leave it as is. Thanks for looking into it

PR #5825, opened 2025-03-27
Update dismiss limiter logic for breakage reporting prompts

laghee:
Task/Issue URL: https://app.asana.com/0/0/1209712075249436/f

### Description
We'd like to alter the logic for refresh breakage prompts to match iOS in the hopes that this will increase volume similarly.

### Steps to test this PR

Notes:
Apply patch: https://app.asana.com/app/asana/-/get_asset?asset_id=1209824280082698
Testing values are 7 seconds (between prompts — normally 7 days) and 30 seconds after 3 dismissals (instead of 30 days), so the 3 dismissals in a row need to happen within those 30 seconds.

_Between-Prompt Cooldown + Max Dismiss Cooldown Logic_
- [x] Load a site
- [x] Refresh 3 times
- [x] Accept the prompt
- [x] Refresh another 3 times immediately
- [x] Verify no prompt (still in cooldown after acceptance)
- [x] Refresh 3 more times after 7-second cooldown has passed
- [x] Verify prompt shows
- [x] Dismiss the prompt
- [x] Refresh another 3 times immediately
- [x] Verify no prompt (still in cooldown after dismissal)
- [x] Refresh 3 more times after 7-second cooldown has passed
- [x] Verify prompt shows
- [x] Dismiss the prompt
- [x] Refresh another 3 times after 7-second cooldown has passed
- [x] Verify prompt shows
- [x] Dismiss the prompt
- [x] After less than 30 seconds refresh another 3 times
- [x] Verify no prompt (still in cooldown)
- [x] Wait out the rest of the cooldown and refresh 3 more times
- [x] Verify prompt shows again

### DIFF
--- a/broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSitePomptDataStore.kt
+++ b/broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSitePomptDataStore.kt
@@ -22,10 +22,13 @@ import androidx.datastore.preferences.core.edit
 import androidx.datastore.preferences.core.intPreferencesKey
 import androidx.datastore.preferences.core.longPreferencesKey
 import androidx.datastore.preferences.core.stringPreferencesKey
+import androidx.datastore.preferences.core.stringSetPreferencesKey
 import com.duckduckgo.app.di.AppCoroutineScope
 import com.duckduckgo.brokensite.impl.SharedPreferencesDuckPlayerDataStore.Keys.COOL_DOWN_DAYS
+import com.duckduckgo.brokensite.impl.SharedPreferencesDuckPlayerDataStore.Keys.DISMISS_EVENTS
 import com.duckduckgo.brokensite.impl.SharedPreferencesDuckPlayerDataStore.Keys.DISMISS_STREAK
 import com.duckduckgo.brokensite.impl.SharedPreferencesDuckPlayerDataStore.Keys.DISMISS_STREAK_RESET_DAYS
+import com.duckduckgo.brokensite.impl.SharedPreferencesDuckPlayerDataStore.Keys.LAST_SHOWN_DATE
 import com.duckduckgo.brokensite.impl.SharedPreferencesDuckPlayerDataStore.Keys.MAX_DISMISS_STREAK
 import com.duckduckgo.brokensite.impl.SharedPreferencesDuckPlayerDataStore.Keys.NEXT_SHOWN_DATE
 import com.duckduckgo.brokensite.impl.di.BrokenSitePrompt
@@ -40,6 +43,8 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
+import java.time.Instant
+import java.time.ZoneOffset
 
 interface BrokenSitePomptDataStore {
     suspend fun setMaxDismissStreak(maxDismissStreak: Int)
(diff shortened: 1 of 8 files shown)


broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/RealBrokenSitePrompt.kt
```diff
@@ -43,16 +43,24 @@ class RealBrokenSitePrompt @Inject constructor(
 
     override suspend fun userDismissedPrompt() {
         if (!_featureEnabled) return
-        if (brokenSiteReportRepository.getDismissStreak() >= brokenSiteReportRepository.getMaxDismissStreak() - 1) {
```
CrisBarreiro, line 46: Since `brokenSiteReportRepository.getDismissStreak()` is not used anymore, we should remove it along with shared preferences supporting it

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSitePomptDataStore.kt
```diff
@@ -138,6 +154,57 @@ class SharedPreferencesDuckPlayerDataStore @Inject constructor(
...
+                val eventDateTime = LocalDateTime.parse(dateString, formatter)
+                eventDateTime.isAfter(t1) && eventDateTime.isBefore(t2)
+            } catch (e: Exception) {
+                false
+            }
+        }
+    }
+
+    override suspend fun deleteAllExpiredDismissals(expiryDate: String) {
+        val normalizedExpiry = LocalDateTime.ofInstant(Instant.parse(expiryDate), ZoneOffset.UTC)
```
CrisBarreiro, line 182: I think it's best to have zoneID passed as a parameter as well, for clarity and extra safety

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/RealBrokenSitePrompt.kt
```diff
@@ -43,16 +43,24 @@ class RealBrokenSitePrompt @Inject constructor(
...
+
+        brokenSiteReportRepository.addDismissal(currentTimestamp)
+
+        val dismissStreakCount = brokenSiteReportRepository.getDismissalCountBetween(
+            currentTimestamp.minusDays(brokenSiteReportRepository.getDismissStreakResetDays().toLong()),
+            currentTimestamp,
+        )
+
+        if (dismissStreakCount >= brokenSiteReportRepository.getMaxDismissStreak()) {
+            val newNextShownDate = currentTimestamp
```
CrisBarreiro, line 57: We don't need to update the next shown date with `getDismissStreakResetDays()` anymore, we just need to store the dismissal at this point, and then use `getDismissalCountBetween` when deciding whether we should show the prompt

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/RealBrokenSitePrompt.kt
```diff
@@ -88,15 +96,48 @@ class RealBrokenSitePrompt @Inject constructor(
...
+
+        val dismissStreakResetDays = brokenSiteReportRepository.getDismissStreakResetDays().toLong()
+        when (val lastShownDate = brokenSiteReportRepository.getLastShownDate()) {
+            null -> {
+                // First time showing the prompt
+                return true
+            }
+
+            else -> {
+                if (currentTimestamp.isAfter(lastShownDate.plusDays(dismissStreakResetDays))) {
```
CrisBarreiro, line 119: We shouldn't check the current timestamp against `dismissStreakResetDays` here, but rather `getDismissalCountBetween`. If the number is lower than `getMaxDismissStreak`, at this point we should return true (since we've already checked `getNextShownDate` above)

CrisBarreiro (changes requested):

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/RealBrokenSitePrompt.kt
```diff
@@ -88,15 +81,33 @@ class RealBrokenSitePrompt @Inject constructor(
...
+        return dismissalCount < brokenSiteReportRepository.getMaxDismissStreak()
     }
 
     override suspend fun ctaShown() {
+        val currentTimestamp = currentTimeProvider.localDateTimeNow()
         val nextShownDate = brokenSiteReportRepository.getNextShownDate()
-        val newNextShownDate = currentTimeProvider.localDateTimeNow().plusDays(brokenSiteReportRepository.getCoolDownDays())
+        val newNextShownDate = currentTimestamp.plusDays(brokenSiteReportRepository.getCoolDownDays())
+
         if (nextShownDate == null || newNextShownDate.isAfter(nextShownDate)) {
```
CrisBarreiro, line 111: nit: I think we can remove this check now and always store `newNextShownDate`, as with the current changes, `newNextShownDate` will always be after `nextShownDate`. Then, we can also remove `val nextShownDate = brokenSiteReportRepository.getNextShownDate()`

PR #5844, opened 2025-04-01
Tab swiping: Add a subfeature flag

0nko:
Task/Issue URL: https://app.asana.com/0/414730916066338/1209846575622295/f

### Description

This PR adds a subfeature flag `onForInternalUsers` and `onForExternalUsers` (the feature flag for prod users to enable staged rollout).

The flag is used inside a `SwipingTabsFeatureProvider` wrapper so nothing else should be affected. The original flag feature flag will be used to keep the feature enabled for all internal users.

After this, the updated privacy config will look like this:
```
"swipingTabs": {
    "state": "enabled",
    "minSupportedVersion": 52271000,
    "features": {
        "onForInternalUsers": {
            "state": "internal"
        },
        "onForExternalUsers": {
            "state": "enabled",
            "rollout": {
                "steps": [
                    {
                        "percent": 20
                    }
                ]
            }
        }
    }
}
```

### Steps to test this PR

_Internal test_
- [x] Replace the privacy config URL to: `https://jsonblob.com/api/1356597958694592512`
- [x] Run the app using an internal build
- [x] Verify the tab swiping is working

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/SwipingTabsFeature.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/SwipingTabsFeature.kt
@@ -25,11 +25,15 @@ import com.duckduckgo.feature.toggles.api.Toggle
     featureName = "swipingTabs",
 )
 interface SwipingTabsFeature {
-    // The flag used to enable the feature for the internal users
+    // The main kill switch for the feature
     @Toggle.DefaultValue(false)
     fun self(): Toggle
 
-    // The toggle used to enable the feature for prod users (during the rollout phase)
+    // The toggle used to enable the feature for internal users
     @Toggle.DefaultValue(false)
-    fun onForExistingUsers(): Toggle
+    fun onForInternalUsers(): Toggle
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/browser/SwipingTabsFeature.kt
```diff
@@ -25,6 +25,15 @@ import com.duckduckgo.feature.toggles.api.Toggle
     featureName = "swipingTabs",
 )
 interface SwipingTabsFeature {
+    // The main kill switch for the feature
     @Toggle.DefaultValue(false)
     fun self(): Toggle
+
+    // The toggle used to enable the feature for internal users
+    @Toggle.DefaultValue(false)
```
CrisBarreiro, line 33: As mentioned on MM, you can just use `@InternalAlwaysEnabled`, so you don't need to have 2 separate subfeatures. You can have a single one with a rollout, so external users will get the feature enabled depending on rollout values, but internal ones will always get it

CrisBarreiro (approved):

PR #5848, opened 2025-04-02
always complete omnibar/minibar transitions with an animation

LukasPaczos:
Task/Issue URL: https://app.asana.com/0/0/1209754552084246

### Description
Adds logic to always complete the omnibar/minibar transition.

The behavior is as follows:
- If user scrolls down, we always immediately start transitioning to minibar.
- If user scrolls up, we only start transitioning back to omnibar after a scroll threshold since the start of last gesture is reached. This is done so that as users scroll back-and-forth around the page we don't switch between omnibar/minibar all the time - only significant scroll gestures and flings, hitting top of the page, or clicking the minibar, will reveal the omnibar back.
- If gesture has any velocity, we complete the transition in the direction of the gesture.

Known issues:
- If we're near the bottom of the page, and **omnibar** is visible, scrolling very slowly until hitting the bottom of the page might cause a little bit of a bounce while we're force-transitioning to the minibar.

### Steps to test this PR

1. Toggle the feature option on: "Experimental UI Setting" -> "Enable all changes".
2. Play around with the transitions in both top and bottom omnibar positions.

### UI changes

_Before_

https://github.com/user-attachments/assets/401aafbb-2d62-4361-873b-838aacbef755

_After_

https://github.com/user-attachments/assets/7c0d2bc0-29db-4ba0-b2d2-299b3de87312

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2751,11 +2751,12 @@ class BrowserTabFragment :
                 }
             }
 
-            it.setOnTouchListener { _, _ ->
+            it.setOnTouchListener { webView, event ->
                 if (omnibar.omnibarTextInput.isFocused) {
                     binding.focusDummy.requestFocus()
                 }
                 dismissAppLinkSnackBar()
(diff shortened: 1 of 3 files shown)


malmstein (changes requested):
When the voice search icon is enabled we are not transitioning properly.

LukasPaczos:
Ready for another round @malmstein @0nko.

PR #5862, opened 2025-04-04
Update 3x pixel-firing logic to more closely match broken site prompt

laghee:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209712075249435/f

### Description

We decided in https://app.asana.com/0/1205142657210376/1209698956036733 that we would prefer the separate 3x pixel to fire for distinct 3-refresh groupings in the chosen time period as opposed to 3 refreshes that fall into a sliding window of time to more closely match the breakage prompt trigger.

### Steps to test this PR

NB: Logic is currently patched to delay 7 seconds between accepted prompts & stop prompting with 3+ dismissals within the past 30 min.

_Feature 1_
- [x] Go to a site
- [x] Refresh 3 times
- [x] Verify that RELOAD-2X and RELOAD-3X pixels fire
- [x] Dismiss prompt
- [x] Refresh 3 times, pause for at least 20 seconds, then refresh 3 times again
- [x] Verify that RELOAD-2X and RELOAD-3X fire twice


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3946,16 +3946,26 @@ class BrowserTabViewModel @Inject constructor(
...
+    private fun fireReloadPixelIfRefreshCountMet() {
+        if (brokenSitePrompt.getUserRefreshesCount(reset = false) >= REFRESH_COUNT_LIMIT) {
+            Timber.d("KateTest--> Sending 3x reload pixel from BTVM fireReloadPixelIfRefreshCountMet")
+            pixel.fire(AppPixelName.RELOAD_THREE_TIMES_WITHIN_20_SECONDS)
+        }
+    }
+
     fun handleMenuRefreshAction() {
         refreshPixelSender.sendMenuRefreshPixels()
+        fireReloadPixelIfRefreshCountMet()
```
marcosholgado, line 3958: Can we move this logic to the RefreshPixelSender? We already have that class to do this so it seems like we are duplicating some logic. For example, the 3x refresh should also send the experimental pixels for the blocklist but it's not.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2759,11 +2763,13 @@ class BrowserTabViewModel @Inject constructor(
             val isBrowserShowing = currentBrowserViewState().browserShowing
             val isErrorShowing = currentBrowserViewState().maliciousSiteBlocked
             if (hasCtaBeenShownForCurrentPage.get() && isBrowserShowing) return null
+            val showBrokenSitePrompt = shouldPromptForBrokenSiteReport()
```
CrisBarreiro, line 2766: nit: should**Show**PromptForBrokenSiteReport

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3946,6 +3952,31 @@ class BrowserTabViewModel @Inject constructor(
...
+                    Timber.d("KateTest--> send pixel only bc 2 refreshes")
+                    false
+                }
+                else -> {
+                    blockListPixelsPlugin.get3XRefresh()?.getPixelDefinitions()?.forEach {
+                        pixel.fire(it.pixelName, it.params)
+                    }
+                    pixel.fire(AppPixelName.RELOAD_THREE_TIMES_WITHIN_20_SECONDS)
+                    Timber.d("KateTest--> should show bc 3 refreshes")
+                    site?.let { brokenSitePrompt.shouldShowBrokenSitePrompt(it.url) } ?: false
```
CrisBarreiro, line 3974: I guess null site here is an edge case here, so probably fine if pixel and prompt fall out of sync

app/src/main/java/com/duckduckgo/app/global/db/AppDatabase.kt
```diff
@@ -693,6 +688,15 @@ class MigrationsProvider(val context: Context, val settingsDataStore: SettingsDa
         }
     }
 
+    private val MIGRATION_57_TO_58: Migration = object : Migration(57, 58) {
+        override fun migrate(database: SupportSQLiteDatabase) {
+            database.execSQL(
+                "DROP TABLE IF NOT EXISTS `refreshes` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, " +
```
CrisBarreiro, line 694: DROP TABLE IF EXISTS, otherwise, there's nothing to drop

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/RealBrokenSitePrompt.kt
```diff
@@ -90,22 +97,25 @@ class RealBrokenSitePrompt @Inject constructor(
...
+                Timber.d("KateTest--> should NOT show bc cooldown: NextDate= $nextDate")
                 return false
             }
+            Timber.d("KateTest--> Passed cooldown test: NextDate= $nextDate")
         }
 
         val dismissStreakResetDays = brokenSiteReportRepository.getDismissStreakResetDays().toLong()
         val dismissalCount = brokenSiteReportRepository.getDismissalCountBetween(
-            currentTimestamp.minusDays(dismissStreakResetDays),
+            currentTimestamp.minusMinutes(dismissStreakResetDays),
```
CrisBarreiro, line 108: Careful to revert this later

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/RealBrokenSitePrompt.kt
```diff
@@ -90,22 +97,25 @@ class RealBrokenSitePrompt @Inject constructor(
...
         )
+        Timber.d("KateTest--> SHOULDshow()?  dismissCount= $dismissalCount, resetDays= $dismissStreakResetDays")
 
         return dismissalCount < brokenSiteReportRepository.getMaxDismissStreak()
     }
 
     override suspend fun ctaShown() {
         val currentTimestamp = currentTimeProvider.localDateTimeNow()
-        val newNextShownDate = currentTimestamp.plusDays(brokenSiteReportRepository.getCoolDownDays())
+        val newNextShownDate = currentTimestamp.plusSeconds(brokenSiteReportRepository.getCoolDownDays())
```
CrisBarreiro, line 118: Same

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3946,6 +3952,31 @@ class BrowserTabViewModel @Inject constructor(
         newTabPixels.get().fireNewTabDisplayed()
     }
 
+    private suspend fun shouldPromptForBrokenSiteReport(): Boolean {
+        return withContext(dispatchers.io()) {
+            val refreshCount = brokenSitePrompt.getUserRefreshesCount()
+            when (refreshCount) {
+                0, 1 -> false
+                2 -> {
+                    blockListPixelsPlugin.get2XRefresh()?.getPixelDefinitions()?.forEach {
```
CrisBarreiro, line 3961: wondering whether we should just send the number of refreshes to RefreshPixelSender and keep firing the pixels there

marcosholgado: +1, I would generally remove as much logic from here as possible. If the refresh count is specific to broken sites, I'd be tempted to move everything there too.

laghee: Makes sense to me, will do.

app/src/main/java/com/duckduckgo/app/global/db/AppDatabase.kt
laghee: 🤦‍♀️

app/src/main/java/com/duckduckgo/app/browser/refreshpixels/RefreshPixelSender.kt
```diff
@@ -19,82 +19,72 @@ package com.duckduckgo.app.browser.refreshpixels
...
         pixel.fire(AppPixelName.REFRESH_ACTION_DAILY_PIXEL, type = Daily())
     }
 
     override fun sendCustomTabRefreshPixel() {
-        sendTimeBasedPixels()
         pixel.fire(CustomTabPixelNames.CUSTOM_TABS_MENU_REFRESH)
     }
 
-    private fun sendTimeBasedPixels() {
+    override fun sendBreakageRefreshPixels(patternsDetected: Set<Int>) {
```
marcosholgado, line 66: I would call this method something like `onRefreshPatternDetected`, the fact that internally we send a pixel or do something else is irrelevant to the caller so I wouldn't be that specific.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2759,11 +2761,13 @@ class BrowserTabViewModel @Inject constructor(
             val isBrowserShowing = currentBrowserViewState().browserShowing
             val isErrorShowing = currentBrowserViewState().maliciousSiteBlocked
             if (hasCtaBeenShownForCurrentPage.get() && isBrowserShowing) return null
+            val showBrokenSitePrompt = shouldShowPromptForBrokenSiteReport()
             val cta = withContext(dispatchers.io()) {
                 ctaViewModel.refreshCta(
                     dispatchers.io(),
                     isBrowserShowing && !isErrorShowing,
                     siteLiveData.value,
+                    showBrokenSitePrompt,
```
marcosholgado, line 2770: Overall I'm struggling with this. I don't fully get why we have pull this out of the `CtaViewModel` to pass a boolean back to it when within the `CtaViewModel` we still have access to the `BrokenSitePrompt`. It seems an unnecessary journey for a value that we can directly get when needed in the `CtaViewModel`

app/src/test/java/com/duckduckgo/app/browser/refreshpixels/RefreshPixelSenderTest.kt
```diff
@@ -143,181 +115,28 @@ class RefreshPixelSenderTest {
...
-    @Test
-    fun whenRefreshedTwiceAndAssignedToExperimentThen2XRefreshPixelsFired() = runTest {
+    fun whenRefreshedTwiceOrThriceAndAssignedToExperimentThen2XRefreshPixelsFired() = kotlinx.coroutines.test.runTest {
         assignToExperiment()
-        testee.sendMenuRefreshPixels()
-        testee.sendMenuRefreshPixels()
+        testee.sendBreakageRefreshPixels(setOf(2, 3))
 
-        blockListPixelsPlugin.get2XRefresh()!!.getPixelDefinitions().forEach {
+        blockListPixelsPlugin.get2XRefresh()?.getPixelDefinitions()?.forEach {
```
marcosholgado, line 122: Go back to `!!` I want this test to fail if `get2Refresh` returns null and this would bypass it.

app/src/test/java/com/duckduckgo/app/browser/refreshpixels/RefreshPixelSenderTest.kt
```diff
@@ -143,181 +115,28 @@ class RefreshPixelSenderTest {
...
-
-    @Test
-    fun whenRefreshedThreeTimesAndAssignedToExperimentThen3XRefreshPixelsFired() = runTest {
-        assignToExperiment()
-        testee.sendMenuRefreshPixels()
-        testee.sendMenuRefreshPixels()
-        testee.sendMenuRefreshPixels()
 
-        blockListPixelsPlugin.get3XRefresh()!!.getPixelDefinitions().forEach {
+        blockListPixelsPlugin.get3XRefresh()?.getPixelDefinitions()?.forEach {
```
marcosholgado, line 126: Same as above

app/src/test/java/com/duckduckgo/app/browser/refreshpixels/RefreshPixelSenderTest.kt
```diff
@@ -143,181 +115,28 @@ class RefreshPixelSenderTest {
     }
 
     @Test
-    fun whenSendMenuRefreshPixelsThenTimeBasedPixelsFired() = runTest {
```
marcosholgado, line 146: Overall a lot of tests have been deleted, are the others covering for the new logic, pixels, etc?

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSiteReportRepository.kt
```diff
@@ -59,15 +59,15 @@ interface BrokenSiteReportRepository {
 
     fun resetRefreshCount()
     fun addRefresh(url: Uri, localDateTime: LocalDateTime)
-    fun getAndUpdateUserRefreshesBetween(t1: LocalDateTime, t2: LocalDateTime): Int
+    fun checkForRefreshPatterns(currentDateTime: LocalDateTime): Set<Int>
```
marcosholgado, line 62: Would probably rename to `getRefreshPatterns` rather than check. Check doesn't tell me what is the thing that is checking so a bit confusing. Also, this returns the refresh patterns but they are not associated to any URL so how do you know it's relevant for the current site?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3946,6 +3950,16 @@ class BrowserTabViewModel @Inject constructor(
         newTabPixels.get().fireNewTabDisplayed()
     }
 
+    private suspend fun shouldShowPromptForBrokenSiteReport(): Boolean {
```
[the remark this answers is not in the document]
laghee: That's a good point. I originally moved it to avoid losing the context for the user refreshes and resetting the count too often, but now that the logic's been altered, that's probably no longer necessary.

app/src/test/java/com/duckduckgo/app/browser/refreshpixels/RefreshPixelSenderTest.kt
laghee: yes!

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSiteReportRepository.kt
laghee: The [in-memory store](https://github.com/duckduckgo/Android/pull/5862/files#diff-d47e78d418739db283d4d048f24012def22255385c67aea5844edb0852bf0998) is keeping track of these per url, so it's built into the underlying method called by the repository.

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -320,7 +322,13 @@ class CtaViewModel @Inject constructor(
             }
 
             if (areInContextDaxDialogsCompleted()) {
-                return if (brokenSitePrompt.shouldShowBrokenSitePrompt(nonNullSite.url)) {
+                val detectedRefreshPatterns = brokenSitePrompt.getUserRefreshesCount()
```
CrisBarreiro, line 325: I don't think we should do this here. Otherwise, we lose signal on breakage for users that haven't completed onboarding

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSiteRefreshesInMemoryStore.kt
```diff
@@ -0,0 +1,97 @@
...
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import java.time.LocalDateTime
+import javax.inject.Inject
+import timber.log.Timber
+
+interface BrokenSiteRefreshesInMemoryStore {
+    fun resetRefreshCount()
+    fun addRefresh(url: Uri, localDateTime: LocalDateTime)
+    fun getRefreshPatterns(currentDateTime: LocalDateTime): Set<Int>
```
CrisBarreiro, line 30: I'd rather encapsulate the return type on a class with more semantics

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -320,7 +322,13 @@ class CtaViewModel @Inject constructor(
...
+                refreshPixelSender.onRefreshPatternDetected(detectedRefreshPatterns)
+
+                return if (3 in detectedRefreshPatterns &&
```
CrisBarreiro, line 328: If we decide to not query the refreshes in `shouldShowBrokenSitePrompt`, which might make sense because of the pixel, I think we should still send the number of refreshes as an attribute, and keep all the logic within the method

CrisBarreiro:
@laghee, the current implementation behaves like a sliding window. If you refresh 4 times in 20 seconds, 3x pixel is fired twice

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSiteRefreshesInMemoryStore.kt
```diff
@@ -0,0 +1,120 @@
...
+    private fun pruneOldRefreshes(
+        currentTime: LocalDateTime,
+        refreshes: LastRefreshedUrl,
+    ): LastRefreshedUrl {
+        val cutoffTime = currentTime.minusSeconds(THRICE_REFRESH_WINDOW_IN_SECS)
+        val newTimes = refreshes.time.filter { it.isAfter(cutoffTime) }
+        return refreshes.copy(time = newTimes)
+    }
+
+    private fun countPattern(refreshEvents: List<LocalDateTime>, pattern: RefreshPattern, window: Long): Int {
```
CrisBarreiro, line 92: I wonder whether we can just remove the pattern after using it. After all, since we fire the pixel and show the CTA right after refreshing 3 times, we shouldn't need to keep several groups

CrisBarreiro (changes requested):

CrisBarreiro (changes requested):

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2784,6 +2784,7 @@ class BrowserTabViewModel @Inject constructor(
             val isErrorShowing = currentBrowserViewState().maliciousSiteBlocked
             if (hasCtaBeenShownForCurrentPage.get() && isBrowserShowing) return null
             val detectedRefreshPatterns = brokenSitePrompt.getUserRefreshesCount()
+            Timber.d("Katetest--> About to handleBreakageRefreshPatterns: $detectedRefreshPatterns")
```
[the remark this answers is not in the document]
laghee: whoops

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4012,6 +4015,10 @@ class BrowserTabViewModel @Inject constructor(
         refreshPixelSender.sendCustomTabRefreshPixel()
     }
 
+    fun handleBreakageRefreshPatterns(refreshPatterns: Set<RefreshPattern>) {
```
CrisBarreiro, line 4018: We shouldn't expose this function, as it shouldn't be called from outside the VM. Can you make it private and refactor `whenHandleBreakageRefreshPatternsThenSendBreakageRefreshPixels` so it calls `refreshCta instead`?

CrisBarreiro (approved):

PR #5864, opened 2025-04-04
Set up support for AppTP blocklist experiments

GuiltyDolphin:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209746219895489/f

### Description

Adds support for AppTP TDS A/B experiments via privacy configuration.

### Steps to test this PR

Automated tests:
- run the `AppTPBlockListInterceptorApiPluginTest.kt` file

Manual check:
1. Using the remote debugger, add in the `features` for `appTrackerProtection` in `android-override.json`:

```
                "tdsExperiment001": {
                    "state": "enabled",
                    "minSupportedVersion": 52240000,
                    "rollout": {
                        "steps": [
                            {
                                "percent": 100
                            }
                        ]
                    },
                    "settings": {
                        "controlUrl": "experiment/android-tds-a0.json",
                        "treatmentUrl": "experiment/android-tds-a1.json"
                    },
                    "cohorts": [
                        {
                            "name": "control",
                            "weight": 1
                        },
                        {
                            "name": "treatment",
                            "weight": 0
                        }
                    ]
                },
```

2. Reload the config
3. in AppTP dev settings, force load the blocklist then leave the settings page
4. in logcat, filter to `AppTP`, you should see the URL being rewritten (similar to screenshot)
5. change control value to 0 and treatment to 1, then verify you're seeing the URL rewritten as treatment

image


app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/blocklist/Blocklist.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.feature.toggles.api.FeatureTogglesInventory
+import com.duckduckgo.feature.toggles.api.Toggle
+import com.duckduckgo.feature.toggles.api.Toggle.State.CohortName
+
+@ContributesRemoteFeature(
+    scope = AppScope::class,
+    featureName = "appTrackerProtection",
+)
+interface BlockList {
```
aitorvs, line 30: Remove this class and add the new features into `AppTpRemoteFeatures`, already existing class for the `appTrackerProtection`

app-tracking-protection/vpn-impl/src/test/java/com/duckduckgo/mobile/android/vpn/blocklist/AppTPBlockListInterceptorApiPluginTest.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    override suspend fun getAllTogglesForParent(name: String): List<Toggle> = withContext(dispatcherProvider.io()) {
+        return@withContext getAll().filter { it.featureName().parentName == name }
+    }
+
+    override suspend fun getAllActiveExperimentToggles(): List<Toggle> = withContext(dispatcherProvider.io()) {
+        return@withContext getAll().filter { it.getCohort() != null && it.isEnabled() }
+    }
+}
```
aitorvs, line 68: Remove, why do we need this?
I think you should be good just by mocking and whenever something called return whatever list of toggles you want.

app-tracking-protection/vpn-impl/src/test/java/com/duckduckgo/mobile/android/vpn/blocklist/AppTPBlockListInterceptorApiPluginTest.kt
```diff
@@ -0,0 +1,267 @@
...
+    private data class Config(
+        val treatmentUrl: String? = null,
+        val controlUrl: String? = null,
+        val nextUrl: String? = null,
+    )
+    private val configAdapter = moshi.adapter(Config::class.java)
+
+    private val pixel = mock<Pixel>()
+    private val sharedPreferencesProvider = mock<SharedPreferencesProvider>()
+    private val prefs = InMemorySharedPreferences()
```
aitorvs, line 91: Use FakeSharedPreferencesProvider from `data-store-test` instead. You'll need to `testImplementation` import `data-store-test`

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpRemoteFeatures.kt
```diff
@@ -55,6 +57,34 @@ interface AppTpRemoteFeatures {
 
     @DefaultValue(true)
     fun setSearchDomains(): Toggle // kill switch
+
+    // TODO: make it possible to dynamically specify experiments without having to specify up front (2025-04-03)
```
aitorvs, line 61: I think we can remove this TODO. We don't have plans to support this.

app-tracking-protection/vpn-impl/src/test/java/com/duckduckgo/mobile/android/vpn/blocklist/AppTPBlockListInterceptorApiPluginTest.kt
```diff
@@ -0,0 +1,239 @@
...
+        testBlockListFeature.atpTdsNextExperimentAnotherTest().setRawStoredState(makeExperiment())
+        checkEndpointIntercept("controlUrl", expectedResponseCode = 200)
+        verifyNoInteractions(pixel)
+    }
+
+    @Test
+    fun `when experiment request fails, sends failure pixel`() = runTest {
+        testBlockListFeature.atpTdsNextExperimentAnotherTest().setRawStoredState(makeExperiment())
+        checkEndpointIntercept("controlUrl", expectedResponseCode = 400)
+        verify(pixel).fire("experiment_metrics_atpTdsNextExperimentAnotherTest_control", mapOf("code" to "400"))
```
GuiltyDolphin, line 174: Hey @aitorvs

I'm having a bit of trouble getting this test to pass. I've tried adding advanceUntilIdle, runCurrent, and a delay, but the test doesn't seem to be picking up the firing pixel (I can see it in testing on a real device). Any tips?

Another option is to switch this to a regular pixel with the experiment name + group as parameters. Unfortunately, it looks like additional info like appVersion and atb gets added in that case, which we don't necessarily want (in the spirit of minimizing information). Is there a way to turn off those extra params?

GuiltyDolphin: I've removed the test as it isn't mandatory to have this tested, and I can see the pixel firing on the test device. Would be nice to have the test but if it's hard to get working we can skip.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/pixels/DeviceShieldPixels.kt
```diff
@@ -892,6 +934,14 @@ class RealDeviceShieldPixels @Inject constructor(
         firePixel(DeviceShieldPixelNames.APPTP_PPRO_UPSELL_REVOKED_INFO_LINK_CLICKED)
     }
 
+    override fun appTPBlocklistExperimentDownloadFailure(statusCode: Int) {
+        appCoroutineScope.launch(dispatcherProvider.io()) {
+            appTpTDSPixelsPlugin.didFailToDownloadTDS()?.getPixelDefinitions()?.forEach {
+                firePixel(it.pixelName, it.params.plus(mapOf("code" to statusCode.toString())))
```
marcosholgado, line 940: As said over MM, we can't add extra parameters to the existing ones. The way to do this would be:

Define new metrics for each error code you want to measure:

```
MetricsPixel(
                metric = "didFailToDownloadTDS",
                value = "500",
                toggle = activeToggle,
                conversionWindow = (0..5).map { ConversionWindow(lowerWindow = 0, upperWindow = it) },
            ),
 MetricsPixel(
                metric = "didFailToDownloadTDS",
                value = "400",
                toggle = activeToggle,
                conversionWindow = (0..5).map { ConversionWindow(lowerWindow = 0, upperWindow = it) },
            ),
```
Then add new functions to retrieve the correct MetricsPixel based on the value too

```
internal suspend fun AppTpTDSPixelsPlugin.didFailToDownloadTDS500Error(): MetricsPixel? {
    return this.getMetrics().firstOrNull { it.metric == "didFailToDownloadTDS" && it.value = "500"}
}
```

Finally before sending the metrics, check the error code and send the metric you want

```
override fun appTPBlocklistExperimentDownloadFailure(statusCode: Int) {
        appCoroutineScope.launch(dispatcherProvider.io()) {
            if (statusCode==500) {
              appTpTDSPixelsPlugin.didFailToDownloadTDS500Error()?.getPixelDefinitions()?.forEach {
                firePixel(it.pixelName, it.params)
            }
        }
    }
```

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/feature/AppTpRemoteFeatures.kt
```diff
@@ -134,3 +179,80 @@ class AppTpRemoteFeaturesStore @Inject constructor(
         private const val PREFS_FILENAME = "com.duckduckgo.vpn.atp.remote.features.v1"
     }
 }
+
+@ContributesMultibinding(AppScope::class)
+class AppTpTDSPixelsPlugin @Inject constructor(private val inventory: FeatureTogglesInventory) : MetricsPixelPlugin {
+
+    override suspend fun getMetrics(): List<MetricsPixel> {
+        val activeToggle = inventory.activeAppTpTdsFlag() ?: return emptyList()
```
aitorvs, line 187: As this is doing disk IO I'd wrap it in `dispatchers.io()`. Probably you can wrap the entire function
```kotlin
override suspend fun getMetrics(): List<MetricsPixel> = withContext(dispatchers.io() {
  val activeToggle = ...
  return@withContext listOf(...)
}

```

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/pixels/DeviceShieldPixels.kt
GuiltyDolphin: Switched to a standard pixel.

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/cohort/CohortPixelInterceptor.kt
```diff
@@ -77,6 +77,7 @@ class CohortPixelInterceptor @Inject constructor(
             "m_atp_ev_cpu_usage_above_",
             "m_atp_unprotected_apps_bucket_",
             "m_atp_breakage_report",
+            "m_atp_tds_experiment_download_failed",
```
marcosholgado, line 80: I don't think this is the interceptor that you are looking for, I think you want `VpnPixelParamRemovalPlugin`

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/blocklist/AppTPBlockListInterceptorApiPlugin.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+    override fun intercept(chain: Chain): Response {
+        val request = chain.request().newBuilder()
+
+        val tdsRequired = chain.request().tag(Invocation::class.java)
+            ?.method()
+            ?.isAnnotationPresent(AppTPTdsRequired::class.java) == true
+
+        val appTpEnabled = runBlocking {
+            vpnFeaturesRegistry.isFeatureRunning(AppTpVpnFeature.APPTP_VPN)
```
marcosholgado, line 65: This is a private API, I asked Aitor and you should use https://duckduckgo.github.io/Android/app-tracking-api/com.duckduckgo.mobile.android.app.tracking/-app-tracking-protection/index.html#-507964147%2FFunctions%2F1663220899 instead which returns true if AppTP is enabled and also running. If you only want to know if it's enabled, then use isEnabled() but always from the public API :)

marcosholgado (changes requested):
Just a minor change to use a public API rather than a private one.

marcosholgado (approved):
LGTM

PR #5878, opened 2025-04-07
fix web view container to shift content up when bottom navigation bar is visible

LukasPaczos:
Task/Issue URL: https://app.asana.com/0/0/1209897414427802

### Description
Adds to the browser container a behavioral dependency on the navigation bar, so that the content of the browser doesn't overlap with the bar.

### Steps to test this PR

- [ ] Verify that in production there's no unexpected content padding on the bottom of the web view, both in top and bottom omnibar positions.
- [ ] Enable the experimental UI feature flag, open a page like YouTube or any other page that has persistent element at the bottom (which is easiest to test with), and ensure that all page content is always visible.

### UI changes

_before_

https://github.com/user-attachments/assets/d6c73985-b667-4c85-a29d-e7381e2fd44d

_after_

https://github.com/user-attachments/assets/c48f8c3c-3641-4b72-89ab-ca920287d36a


malmstein (approved):
Nice, works as expected!

PR #5902, opened 2025-04-11
cache site errors and assign to pages after they load

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1209947027710700

### Description
Resolves an issue where a web view error delivered while user is reloading the current page would overwrite local state, making it invalid and showing wrong URL and warnings about a lack of protections.

### Steps to test this PR

- Open `amazon.com` (either from SERP or manually)
- Type in **manually** `amazon.com` and load the page multiple times.
- Verify that address bar keeps showing the correct URL (including HTTPS) and privacy shield.
- Opening privacy dashboard should always show that protections are active and that the site is secure.

You can also try with `abc.com`, `cnn.com` (which always should show `edition.cnn.com`).


mikescamell (approved):
LGTM! Nice work! 🙌 

I lack some of the original context but read through what happened, tested this on multiple sites and it worked ✅

PR #5907, opened 2025-04-14
update Custom Tab omnibar and navigation bar behavior/layout

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1209703519253563?focus=true

### Description
Updated the behavior of navigation bar to hide when in Custom Tab. Also re-aligned Custom Tab design with production, only with the updated background colors.

I tried reusing as much from the production implementation as possible, since the designs are nearly identical.

### Steps to test this PR
- [ ] With experimental changes flag disabled, ensure that the omnibar's background color in Custom Tab is white when light theme (same as production).
- [ ] With experimental changes flag enabled ensure that:
  - [ ] Omnibar's background color in Custom Tab is warmer white when light theme.
  - [ ] Overflow menu is anchored to the omnibar and navigation buttons are at the top.
  - [ ] There are no rounded corners around the web view.

### UI changes
1. Experiment before
2. Experiment after
3. Production

| 1 | 2 | 3 |
| ------ | ------ | ------ |
|  Screenshot_20250414_135544 | Screenshot_20250414_164147 | Screenshot_20250414_144119 |

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/experiments/FadeOmnibarLayout.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/experiments/FadeOmnibarLayout.kt
@@ -20,6 +20,7 @@ import android.animation.ValueAnimator
 import android.content.Context
 import android.util.AttributeSet
 import android.view.View
+import android.view.ViewOutlineProvider
 import android.view.animation.DecelerateInterpolator
 import android.widget.ImageView
 import android.widget.LinearLayout
@@ -117,8 +118,6 @@ class FadeOmnibarLayout @JvmOverloads constructor(
                 top = 0,
             )
         }
-
-        outlineProvider = null
     }
 
     override fun onDetachedFromWindow() {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/duckduckgo/app/browser/omnibar/experiments/FadeOmnibarLayout.kt
```diff
@@ -127,12 +126,22 @@ class FadeOmnibarLayout @JvmOverloads constructor(
...
-            showTabsMenu = false,
-        )
-        super.render(experimentalViewState)
+        if (viewState.viewMode is ViewMode.CustomTab) {
+            // adds a drop shadow for the AppBarLayout, in case it was removed at any point
+            outlineProvider = ViewOutlineProvider.BACKGROUND
+            super.render(viewState)
+        } else {
+            // removes the drop shadow from the AppBarLayout to make it appear flat in the view hierarchy
+            outlineProvider = null
```
malmstein, line 135: this shouldn’t be needed every single time the view state is rendered. the omnibar won’t change from custom tab to browser mode.

malmstein (changes requested):
Works as expected but there are some UI misalignments we should fix, I’ve added a video to the task.

app/src/main/java/com/duckduckgo/app/browser/omnibar/experiments/FadeOmnibarLayout.kt
LukasPaczos: > the omnibar won’t change from custom tab to browser mode

While that's true now, I don't want to bake that assumption into the codebase. The layout doesn't know it, and it should just react to what's in the view model.

This setter function is guarded by an equality check under the hood, so there's no perf penalty.

LukasPaczos:
Fixing the content jump issue in https://github.com/duckduckgo/Android/pull/5917. This is ready for a final review @malmstein.

malmstein (approved):
LGTM

PR #5908, opened 2025-04-14
Fix: Undo delete snackbar in the browser

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202552961248957/task/1209975829243855?focus=true

### Description

This PR fixes the mechanism to display the undo delete snackbar in the browser screen after all tabs are deleted in the tab switcher, when the snackbar would not show up correctly. 

The PR also fixes the position of the snackbar so that it's shown above the omnibar if it's set to the bottom position.

### Steps to test this PR

- [x] Have 3 tabs and Omnibar set to bottom of the screen
- [x] Close a tab that is not selected
- [x] Switch to the last tab that is not selected while the snackbar is showing
- [x] Verify no snackbar is displayed in the browser screen
- [x] Go back to the tab switcher
- [x] Close one of the tabs, then immediately close the last tab
- [x] Verify the undo snackbar is displayed for the last tab closed
- [x] Verify the snackbar is displayed above the bottom address bar

### UI changes

<image src="https://github.com/user-attachments/assets/7816d82f-004e-4534-ba01-8c91e386d13b" width="300" />


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -617,6 +617,20 @@ class BrowserTabFragment :
 
     private var webView: DuckDuckGoWebView? = null
 
+    private val tabSwitcherActivityResult = registerForActivityResult(StartActivityForResult()) { result ->
+        if (result.resultCode == RESULT_OK) {
+            // Handle any result data if needed
+            result.data?.let { intent ->
+                intent.extras?.let { extras ->
+                    val deletedTabIds = extras.getStringArrayList(TabSwitcherActivity.EXTRA_KEY_DELETED_TAB_IDS)
+                    if (deletedTabIds != null && deletedTabIds.size > 0) {
```
mikescamell, line 626: 💡Could use `isNotEmpty` instead of `> 0`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -617,6 +617,20 @@ class BrowserTabFragment :
...
+                        (activity as? BrowserActivity?)?.onTabsDeletedInTabSwitcher(deletedTabIds)
+                    }
```
mikescamell, line 628: 🔍 I think I'd prefer something more like this:

`(requireActivity() as BrowserActivity).onTabsDeletedInTabSwitche(deletedTabIds)`

but if you want to be safe fair enough 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -651,20 +643,33 @@ open class BrowserActivity : DuckDuckGoActivity() {
...
-            onAction = { viewModel.undoDeletableTabs(tabIds) },
-            onDismiss = { viewModel.purgeDeletableTabs() },
-        ).show()
+        lifecycleScope.launch {
+            // allow the tab fragment to initialize for the snackbar
+            delay(500)
+
+            val anchorView = when (settingsDataStore.omnibarPosition) {
+                TOP -> binding.fragmentContainer
+                BOTTOM -> currentTab?.omnibar?.newOmnibar ?: binding.fragmentContainer
```
mikescamell, line 652: ❓Why can we use the FragmentContainer for the top omnibar but have to switch for the bottom? 🤔The top is fine with the snackbar, right?

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -651,20 +643,33 @@ open class BrowserActivity : DuckDuckGoActivity() {
...
-            message = resources.getQuantityString(R.plurals.tabSwitcherCloseTabsSnackbar, tabIds.size, tabIds.size),
-            action = getString(R.string.tabClosedUndo),
-            showAction = true,
-            onAction = { viewModel.undoDeletableTabs(tabIds) },
-            onDismiss = { viewModel.purgeDeletableTabs() },
-        ).show()
+        lifecycleScope.launch {
+            // allow the tab fragment to initialize for the snackbar
+            delay(500)
+
```
mikescamell, line 649: 🤔Something feels off that we have to wait, but I guess each tab has it's own omnibar, so adding snackbar handling to each tab could get weird

mikescamell (commented):
Something was not right when testing the second part of the steps. I ended up in a state with one new tab page but it was displaying 2 tabs.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
0nko: `BrowserTabFragment` can also be inside a `CustomTabActivity`. You probably wouldn't get here from there but it's better to be safe IMO.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
0nko: This fixes the snackbar covering the bottom address bar.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
0nko: Without this the snackbar wouldn't be visible in certain cases (closing tabs one by one until all are deleted). The problem is that the snackbar is attached to the wrong fragment (I think). This fixes it.

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
```diff
@@ -609,15 +609,16 @@ class TabSwitcherViewModel @Inject constructor(
...
         val mode: Mode = Normal,
         val layoutType: LayoutType? = null,
     ) {
+        val tabs: List<Tab> = tabSwitcherItems.filterIsInstance<Tab>()
         val numSelectedTabs: Int = (mode as? Selection)?.selectedTabs?.size ?: 0
 
         val dynamicInterface = when (mode) {
             is Normal -> {
-                val isThereOnlyNewTabPage = tabItems.size == 1 && tabItems.mapNotNull { it as? Tab }.any { it.isNewTabPage }
+                val isThereOnlyNewTabPage = tabs.size == 1 && tabs.first().isNewTabPage
```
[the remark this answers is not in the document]
0nko: That wouldn't guarantee there is exactly one tab.

PR #5917, opened 2025-04-15
fix Omnibar's content jump when Custom Tab's title becomes available

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1209987826470158?focus=true

### Description

All the of the text views are anchored to the title, so to fix the content jump I'm making sure that the title view is always present in the layout and hidden instead of removed when title text is not available.

### Steps to test this PR

- [ ] Open a Custom Tab and make sure that once content loads, there's no position shift in text views.

### UI changes

_before_

https://github.com/user-attachments/assets/acafa3cf-6e11-4c8d-bca2-3520562b9e75

_after_

https://github.com/user-attachments/assets/3267bda7-0e83-4853-8cf4-1e9936ae127a


malmstein (approved):
LGTM

PR #5919, opened 2025-04-16
Quick win: Skip the current tab in switch-to-tab suggestions

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1208855331182033?focus=true

### Description

This PR filters out the current tab from the switch-to-tab suggestion list.

### Steps to test this PR

- [x] Clear tabs, so there is only 1 active tab
- [x] Visit a page (i.e. [nike.com](https://nike.com/))
- [x] In that same tab, tap into the address bar and type nike
- [x] Verify that only the history suggestion is displayed for the Nike website, not a switch-to-tab suggestion


anikiki (approved):

PR #5931, opened 2025-04-17
fix WebView offset when navigation bar enabled and keyboard visible

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210010551182684?focus=true

### Description
Fixes an issue where navigation bar, even when gone when keyboard is inflated, would still retain the offset applied to the web view content.

The fix is to notify the coordinator layout that navigation bar changed, when its visibility changes, which doesn't happen by default.

We already have handling of that available in the behavior that adjusts browsers padding:

https://github.com/duckduckgo/Android/blob/c0e22478de53029ceb4f6f80188c97782a74562a/app/src/main/java/com/duckduckgo/app/browser/webview/BrowserContainerLayoutBehavior.kt#L53-L78

### Steps to test this PR

- [x] With experimental UI flag enabled, click on the address bar when on a web page other than SERP. Ensure that there's no visible inset/padding on the bottom of the web view.
- [ ] Try that with both top and bottom omnibar.

### UI changes
1. Before
2. After

| 1  | 2 |
| ------ | ------ |
Screenshot_20250417_163916|Screenshot_20250417_164016|

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/navigation/bar/BrowserNavigationBarViewIntegration.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/navigation/bar/BrowserNavigationBarViewIntegration.kt
@@ -29,6 +29,7 @@ import com.duckduckgo.common.ui.view.show
 import com.duckduckgo.common.utils.keyboardVisibilityFlow
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.launch
 
 /**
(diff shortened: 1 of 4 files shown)


0nko:
As I was testing it with the bottom bar, tapping on the address bar, then web view, address bar again (a few times), I still got inset:

image

LukasPaczos:
Thanks @0nko, great catch!

The issue was caused by `browserLayout` not observing the bottom omnibar directly. Instead, it synced its offset with transitions in `BottomAppBarBehavior`, which only responds to web view scrolls. As a result, changes to the omnibar triggered by other sources were ignored.

The bottom omnibar includes the bottom navigation bar, which we hide when the keyboard appears. This alters the omnibar's size, but since `BottomAppBarBehavior` only reacted to scroll events, it didn’t update the web view's padding/margin accordingly.

In https://github.com/duckduckgo/Android/pull/5931/commits/3fcc5be5e1ab8f6aa6867fbccca42e20b7a6cf94, I removed the hardcoded `browserLayout` adjustments tied to `BottomAppBarBehavior`. Instead, I updated `BrowserContainerLayoutBehavior` to apply correct padding for all bottom omnibar changes, whether triggered by scroll or size adjustments.

This commit also changes the codepath for the production omnibar, so when reviewing please make sure to thoroughly test both the experimental and production bottom omnibars.

0nko (approved):
Looks good, nice work! :shipit:

PR #5936, opened 2025-04-18
fix alignment of fade omnibar's input text

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1142021229838617/task/1210011922338726?focus=true

### Description
Removes the unnecessary input text margins and padding but also ensures that the view which follow the omnibar text is at least 10dp from the right edge (width + margin).

### Steps to test this PR

With experimental UI feature flag enabled open a page which URL overflows the address bar and check alignment.

### UI changes
| Before  | After |
| ------ | ----- |
Screenshot_20250418_145734|Screenshot_20250418_145706|


malmstein (approved):
LGTM

PR #5963, opened 2025-04-25
fix onboarding privacy shield and fire button highlights

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210067401123378

### Description

This PR fixes the position and clipping of the privacy shield highlight (pulsating animation) during onboarding.

The class responsible for the animation (`PulseAnimation`) makes _a lot_ of assumptions about the target view hierarchy into which it injects an animated image view. I made the simplest fix possible to adjust the animated circle's position while fitting these assumptions but the solution comes with some hardcoded values. We should separately consider revisiting how the `PulseAnimation` class is built to allow more flexibility.

Additionally, to fix the clipping I moved the animation outside of the omnibar card. Even though the `PulseAnimation` forcefully disables clipping for all parents within the view hierarchy (another assumption), the `MaterialCardView` ignores these calls and will still clip the views. This is justified in the impl by the fact that cards have elevation, etc. I found some ways to disable clipping for the card:
```
app:cardPreventCornerOverlap="false"
app:cardUseCompatPadding="true"
android:clipChildren="false"
android:clipToPadding="false"
android:outlineProvider="none"
```
but I don't like hacking around the card's functionality and relying on some compat properties to maintain things like shadows.

Also in this PR I'm adding a pulse animation to the fire button found in the bottom navigation bar.

### Steps to test this PR

Start with a fresh installation and go through initial onboarding activity until the browser menu becomes available, and then enable the experimental UI feature flag. Continue through onboarding and verify the shield and fire button highlights are shown and dismissed correctly.

https://github.com/user-attachments/assets/e4904da3-d505-46e1-9691-b2e437993d59

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -4115,6 +4115,7 @@ class BrowserTabFragment :
                 }
 
                 omnibar.renderBrowserViewState(viewState)
+                browserNavigationBarIntegration.configureFireButtonHighlight(highlighted = viewState.fireButton.isHighlighted())
                 if (omnibar.isPulseAnimationPlaying()) {
                     webView?.setBottomMatchingBehaviourEnabled(true) // only execute if animation is playing
                 }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4115,6 +4115,7 @@ class BrowserTabFragment :
                 }
 
                 omnibar.renderBrowserViewState(viewState)
+                browserNavigationBarIntegration.configureFireButtonHighlight(highlighted = viewState.fireButton.isHighlighted())
```
malmstein, line 4118: nit naming: setFireButtonHighlight
```suggestion
browserNavigationBarIntegration.setFireButtonHighlighton(highlighted = viewState.fireButton.isHighlighted())
```

PR #5968, opened 2025-04-25
always expand the omnibar when top of the page is reached

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210004588627880?focus=true

### Description
Fixes an issue with the omnibar not showing up when scrolling up by ensuring we always expand it when top of the page is reached. Details in the task.

### Steps to test this PR

- [x] Play around with the omnibar and ensure that it's always shown when top of the page is reached, even when it's via a fling gesture.
- [x] Try bottom omnibar too.

You can also try scrolling up slowly and check for any jitter - should be none. However, if you creep up very very slowly to the top of the page, without flinging and without allowing the omnibar to expand, it won't automatically expand to prevent it being misinterpreted as a gesture in the opposite direction and causing unintended content shift.

### UI changes
| Before  | After |
| ------ | ----- |
hit_top_bug|hits_top|

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -624,6 +624,7 @@ class BrowserTabFragment :
         get() = activity as? BrowserActivity
 
     private var webView: DuckDuckGoWebView? = null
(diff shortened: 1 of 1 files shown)


0nko (approved):
LGTM 👍

PR #5972, opened 2025-04-28
Store initial Duck.ai address bar setting

joshliebe:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1210096722341711?focus=true

### Description

- If the Duck.ai menu setting was initially disabled and the user updates, then the address bar setting is also disabled.
- However, if the the menu setting is toggled, then both settings toggle because there is no initial stored value for the address bar setting.
- This fixes that use-case.

### Steps to test this PR

- [ ] Checkout 5.230.1
- [ ] Disable the Duck.ai menu setting
- [ ] Update to this branch
- [ ] Verify that both settings are disabled
- [ ] Toggle the menu setting
- [ ] Verify that only the menu setting is toggled


malmstein (approved):
This also fixes the same issue on the updated UI. Thanks @joshliebe !

PR #5989, opened 2025-04-30
Ensure blocklist is updated always

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202279501986195/task/1210112806879953?focus=true

### Description
Remove the eTag checks to ensure blocklist is always stored

### Steps to test this PR

_Test 1_
- [x] build and install internal debug build
- [x] launch the app
- [x] verify `Updating the app tracker blocklist, previous/new eTag:...` appears in logs

_Test 2_
- [x] Change from `ExistingPeriodicWorkPolicy.KEEP` to `ExistingPeriodicWorkPolicy.REPLACE` in  `AppTrackerListUpdateWorkerScheduler`
- [x] build and install internal debug build
- [x] launch the app
- [x] verify `Updating the app tracker blocklist, previous/new eTag:...` appears in logs
- [x] force close and reopen the app
- [x] verify `Updating the app tracker blocklist, previous/new eTag:...` appears in logs


nshuba (approved):
LGTM, though I haven't tested - I believe @laghee has?

aitorvs:
@laghee 
> I don't get a second update if I force close and re-launch in Test 2 with REPLACE instead of KEEP. Am I missing a step?
Can you try again? I get the right behavior all the time

> seems like we're not overwriting the etag, but adding one?
I "fixed" this in the latest commit. Now you should get only one tag. Preview behavior was fine to barring that the db table would increase for ever which probably we don't want to.

laghee (approved):
All testing steps WFM now 👍

PR #5991, opened 2025-04-30
FF remove synchronized from EncryptionHelper

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1210112602367622?focus=true 

### Description
Use mutex instead of synchronized in encrypt/decrypt to prevent thread starvation

### Steps to test this PR

_Feature 1_
- [ ] Smoke test autofill

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/securestorage/encryption/EncryptionHelper.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/securestorage/encryption/EncryptionHelper.kt
@@ -21,6 +21,7 @@ import com.duckduckgo.autofill.api.AutofillFeature
 import com.duckduckgo.autofill.impl.securestorage.SecureStorageException
 import com.duckduckgo.autofill.impl.securestorage.SecureStorageException.InternalSecureStorageException
 import com.duckduckgo.autofill.impl.securestorage.encryption.EncryptionHelper.EncryptedBytes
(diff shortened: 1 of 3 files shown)


autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/securestorage/encryption/EncryptionHelper.kt
```diff
@@ -53,12 +56,44 @@ interface EncryptionHelper {
...
+    override suspend fun encrypt(
+        raw: ByteArray,
+        key: Key,
+    ): EncryptedBytes {
+        return if (autofillFeature.createAsyncPreferences().isEnabled()) {
+            encryptAsync(raw, key)
+        } else {
+            encryptSync(raw, key)
+        }
+    }
```
aitorvs, line 77: Given we're doing diskIO here, do we want to directly use dispatchers.io() here via `withContext` 
```suggestion
    override suspend fun encrypt(
        raw: ByteArray,
        key: Key,
    ): EncryptedBytes = withContext(dispatchers.io()) {
        return@withContext if (autofillFeature.createAsyncPreferences().isEnabled()) {
            encryptAsync(raw, key)
        } else {
            encryptSync(raw, key)
        }
    }
```

aitorvs (approved):
left one comment. lgtm

PR #5992, opened 2025-04-30
add tab count to bottom nav bar

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210044541037881?focus=true

### Description
Adds the tab count to the bottom navigation bar's tabs icon. Also introduces a visual cue when there are any tabs that were opened in the background and not visited yet.

I've updated the tabs icon to match the design, but in parallel double checking if these are the right icons.

### Steps to test this PR

- [ ] Enabled visual experiment.
- [ ] Verify that:
  - [ ] There's a number of tabs visible over the tab icon.
  - [ ] If you have 100 or mote tabs open, the number changes to "" character.
  - [ ] If you open a tab in the background, a blue highlight appears over the tab icon.
  - [ ] If you visit the background tab, the highlight disappears.

Tip: you can use `Developer Settings -> Tabs` utility to create random tabs.

### UI changes
| Before  | After |
| ------ | ----- |
Screenshot_20250430_152135|Screenshot_20250430_164800|
||Screenshot_20250430_164808|


common/common-ui/src/main/res/drawable/ic_tab_24_highlighted.xml
```diff
@@ -0,0 +1,16 @@
+<!--TODO this icon needs to be aligned with all the other new icons-->
```
LukasPaczos, line 1: This will be fixed in https://app.asana.com/1/137249556945/project/715106103902962/task/1210170362960641?focus=true.

PR #6004, opened 2025-05-02
Fix: Ensure the scrollbars are always restored

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1210108991225075?focus=true

### Description

This fixes a bug when the WebView scrollbars would stay hidden if the preview generation threw an exception. This may happen fairly often with tab swiping when you swipe between tabs fast so that the webview doesn't get a chance to get laid out.

### Steps to test this PR

- [ ] Swipe tabs until you notice `Failed to generate WebView preview` in the logcat
- [ ] Swipe back to the last tab
- [ ] Notice the WebView has scrollbars when you scroll it

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/tabpreview/WebViewPreviewGenerator.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/tabpreview/WebViewPreviewGenerator.kt
@@ -33,13 +33,17 @@ interface WebViewPreviewGenerator {
 class FileBasedWebViewPreviewGenerator(private val dispatchers: DispatcherProvider) : WebViewPreviewGenerator {
 
     override suspend fun generatePreview(webView: WebView): Bitmap {
-        disableScrollbars(webView)
-        val fullSizeBitmap = createBitmap(webView)
+        val scaledBitmap: Bitmap
+        try {
+            disableScrollbars(webView)
+            val fullSizeBitmap = createBitmap(webView)
 
-        val scaledHeight = webView.context.resources.getDimension(R.dimen.gridItemPreviewHeight).toPx()
-        val scaledWidth = scaledHeight / fullSizeBitmap.height * fullSizeBitmap.width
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/browser/tabpreview/WebViewPreviewGenerator.kt
```diff
@@ -33,13 +33,17 @@ interface WebViewPreviewGenerator {
 class FileBasedWebViewPreviewGenerator(private val dispatchers: DispatcherProvider) : WebViewPreviewGenerator {
 
     override suspend fun generatePreview(webView: WebView): Bitmap {
-        disableScrollbars(webView)
-        val fullSizeBitmap = createBitmap(webView)
+        val scaledBitmap: Bitmap
```
CDRussell, line 36: nit: don't think you need to have `val scaledBitmap: Bitmap` and `return scaledBitmap` separately; could just make it part of the `return` statement at the end of the `try` block. e.g..

```
try {
    disableScrollbars(webView)
    val fullSizeBitmap = createBitmap(webView)

    val scaledHeight = webView.context.resources.getDimension(R.dimen.gridItemPreviewHeight).toPx()
    val scaledWidth = scaledHeight / fullSizeBitmap.height * fullSizeBitmap.width
    return scaleBitmap(fullSizeBitmap, scaledHeight.roundToInt(), scaledWidth.roundToInt())
} finally {
    enableScrollbars(webView)
}
```

CDRussell (approved):

app/src/main/java/com/duckduckgo/app/browser/tabpreview/WebViewPreviewGenerator.kt
0nko: Good call, thanks!

PR #6005, opened 2025-05-02
Fix: Tab swiping taking over Duck Player

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1210107135506123?focus=true

### Description

This PR disables the tab swiping when the browser is in full-screen mode. This fixes an issue when playing a video in Duck player in full-screen mode and it is not possible to control the player.

### Steps to test this PR

- [x] Open a video in Duck player
- [x] Toggle the full-screen mode
- [x] Verify the tab swiping is completely disabled
- [x] Toggled off the full-screen mode
- [x] Verify the tab swiping on the omnibar works as expected


app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4364,15 +4364,15 @@ class BrowserTabFragment :
         private fun goFullScreen() {
             omnibar.hide()
             binding.webViewFullScreenContainer.show()
-            activity?.toggleFullScreen()
+            (activity as DuckDuckGoActivity).toggleFullScreen()
```
marcosholgado, line 4367: Is this correct? We don't necessary go to full screen only on the DuckDuckGoActivity so I take this cast can fail?. cc/ @CrisBarreiro

0nko: Is there any activity that doesn't implement `DuckDuckGoActivity`? I thought all did. The ones that are important here -- `BrowserActivity` and `CustomTabActivity` do.

CrisBarreiro: Not all of them do. For example, `SelectedTextSearchActivity` or `FireActivity` don't. I don't think they're relevant here, but I think we should at least so a safe cast

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4364,15 +4364,15 @@ class BrowserTabFragment :
...
+            (activity as? DuckDuckGoActivity)?.toggleFullScreen()
             showToast(R.string.fullScreenMessage, Toast.LENGTH_SHORT)
         }
 
         private fun exitFullScreen() {
             omnibar.show()
             binding.webViewFullScreenContainer.removeAllViews()
             binding.webViewFullScreenContainer.gone()
-            activity?.toggleFullScreen()
+            (activity as DuckDuckGoActivity).toggleFullScreen()
```
CrisBarreiro, line 4375:
```suggestion
            (activity as? DuckDuckGoActivity)?.toggleFullScreen()
```

PR #6007, opened 2025-05-04
Fix: NTP creation crash with tab swiping

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1210095537840719?focus=true

### Description

This PR fixes the NTP creation. There was a logical error in the previous implementation when the wrong boolean was being returned in `transformWhile()` when requesting a new tab, which in some situation might result in the flow completion without emitting any value, which resulted in a crash.

I added code that throws an exception when a new tab is not created because it's an invalid state in the DB that I'm not sure how we should recover from. I don't think that's what was happening but in case it does, we'll know where to start looking.

### Steps to test this PR

I wasn't able to reproduce it, however, I added unit tests that should verify the fix works as expected.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/tabs/DefaultTabManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/tabs/DefaultTabManager.kt
@@ -18,20 +18,27 @@ package com.duckduckgo.app.browser.tabs
 
 import com.duckduckgo.app.browser.SkipUrlConversionOnNewTabFeature
 import com.duckduckgo.app.browser.omnibar.OmnibarEntryConverter
+import com.duckduckgo.app.browser.tabs.TabManager.Companion.NEW_TAB_CREATION_TIMEOUT_LIMIT
 import com.duckduckgo.app.tabs.model.TabEntity
 import com.duckduckgo.app.tabs.model.TabRepository
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.ActivityScope
 import com.squareup.anvil.annotations.ContributesBinding
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.flow.catch
 import javax.inject.Inject
-import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.timeout
 import kotlinx.coroutines.flow.transformWhile
 import kotlinx.coroutines.withContext
 import timber.log.Timber
+import kotlin.time.Duration.Companion.seconds
 
 interface TabManager {
     companion object {
         const val MAX_ACTIVE_TABS = 20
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/tabs/DefaultTabManager.kt
```diff
@@ -76,15 +83,27 @@ class DefaultTabManager @Inject constructor(
...
+            .timeout(NEW_TAB_CREATION_TIMEOUT_LIMIT.seconds)
+            .catch { e ->
+                if (e is TimeoutCancellationException) {
+                    // timeout expired and the new tab was not found
+                    throw IllegalStateException("A new tab failed to be created within 1 second")
+                } else {
+                    throw e
+                }
+            }
+            .firstOrNull() ?: throw IllegalStateException("Tabs flow completed before finding the new tab")
```
CDRussell, line 106: I'm a bit confused / worried about some of these changes tbh. We're now throwing exceptions if it doesn't complete in time, is something catching that or is crashing what we want here? What happens if the device is just old/slow/overloaded and happens to exceed 1s?

Why aren't these changes being feature flagged?

CDRussell: > throws an exception when a new tab is not created because it's an invalid state in the DB that I'm not sure how we should recover from

That's the main bit i'm concerned about; i'm not sure it's always telling us about an invalid state. IIRC, there was the possibility of the underlying repo refusing to add another tab with the requested ID but that from a user perspective there would be no obvious problems. I'll dig in a little bit more though as I could be wrong. lmk if you want to chat sync about it.

CDRussell: >Why aren't these changes being feature flagged? 
> It is, this is only used by tab swiping.

Thanks, that wasn't obvious from the diff that this function is currently only called from call sites already gated with the feature flag.

0nko: > are you happy with the scope of what will be disabled? e.g., does all of tab swiping get disabled?

Yes, because this change fixes the original crash that caused the feature to be disabled in the first place. The thrown exceptions are just extra for us so that we know some unexpected circumstance occurred, which I assumed was irrecoverable. 

Just for context -- this function is being called when a ViewPager2 adapter requests a new fragment for a tab (which happens synchronously). The adapter expects a fragment to be created so I'm not sure what should happen if we can't get a tab.

CDRussell: I'm not sure, either tbh. But if it's suitable FF-ed, I'm ok with trying it.

CDRussell: >we could increase the time limit if you feel it's too short

I feel like it might be, but not 100% confident on any value for this. If you were pressing me, I'd throw out `2s` but don't expect me to back that up with justification 😅. I'm ok with trying `1s` as well. Will defer to your judgement.

PR #6029, opened 2025-05-08
Narrow down errors when creating null tunnel

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1210194113621115?focus=true

### Description
Improve crash messaging when creating null tun.
Also add a small delay to give `startForeground` time to "settle".

### Steps to test this PR
Smoke test VPN and AppTP

### DIFF
--- a/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/integration/NgVpnNetworkStack.kt
+++ b/app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/integration/NgVpnNetworkStack.kt
@@ -234,9 +234,15 @@ class NgVpnNetworkStack @Inject constructor(
 
             tunnelThread = Thread {
                 logcat { "Running tunnel in context $jniContext" }
-                vpnNetwork.run(jniContext, tunfd)
-                logcat(LogPriority.WARN) { "Tunnel exited" }
-                tunnelThread = null
+                try {
+                    vpnNetwork.run(jniContext, tunfd)
+                } catch (t: Throwable) {
+                    logcat(LogPriority.ERROR) { "Tunnel thread crashed: ${t.asLog()}" }
+                    deviceShieldPixels.reportTunnelThreadAbnormalCrash()
+                } finally {
+                    tunnelThread = null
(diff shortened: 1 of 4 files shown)


app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/integration/NgVpnNetworkStack.kt
```diff
@@ -234,9 +234,15 @@ class NgVpnNetworkStack @Inject constructor(
...
-                vpnNetwork.run(jniContext, tunfd)
-                logcat(LogPriority.WARN) { "Tunnel exited" }
-                tunnelThread = null
+                try {
+                    vpnNetwork.run(jniContext, tunfd)
+                } catch (t: Throwable) {
+                    logcat(LogPriority.ERROR) { "Tunnel thread crashed: ${t.asLog()}" }
+                    deviceShieldPixels.reportTunnelThreadAbnormalCrash()
+                } finally {
+                    tunnelThread = null
```
karlenDimla, line 243: Could this cause a `nullpointer` in line 246 `it.start()`?

karlenDimla (approved):
Smoke tested AppTP and VPN (on pixel and samsung) and seem to work as expected.
Just one question (see below).

app-tracking-protection/vpn-impl/src/main/java/com/duckduckgo/mobile/android/vpn/integration/NgVpnNetworkStack.kt
aitorvs: The assignment to `tunnelThread` has already happened before the thread ever sets `tunnelThread = null` inside its body. Example

```
val thread = Thread {
    // Runs later
}.also { it.start() } // Starts immediately

// thread is still a valid reference here
```

Does it make sense?

karlenDimla: Ah makes sense 👍

PR #6031, opened 2025-05-09
remove local preference for visual design experiment

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210205897263576?focus=true

### Description
Removes the local preference for the visual design experiment. Now, the toggle in the UI changes the state of the actual feature flag.

Also removes the unsued navigation bar toggle. Navigation bar is always visible when the experiment as a whole is enabled.

### Steps to test this PR

- [ ] Verify that new visual design is visible when toggle in the "Experimental UI Settings" is on.
- [ ] Verify that new visual design is not visible when toggle in the "Experimental UI Settings" is off.


common/common-ui-internal/src/main/java/com/duckduckgo/common/ui/internal/experiments/visual/VisualDesignExperimentViewModel.kt
```diff
@@ -56,22 +54,11 @@ class VisualDesignExperimentViewModel @Inject constructor(
...
-            }
-        }
-
-        viewModelScope.launch {
-            visualDesignExperimentDataStore.navigationBarState.collect { state ->
-                _viewState.update {
-                    it.copy(
-                        isNavigationBarAvailable = state.isAvailable,
-                        isNavigationBarEnabled = state.isEnabled,
+                        isBrowserThemingFeatureAvailable = true,
```
LukasPaczos, line 60: In here's we're defining that the toggle in the "Experimental UI Settings" is visible. If in the future we have more toggles that depend on the overarching one, we can make them visible/invisible here depending on whether the parent is enabled. cc @joshliebe

common/common-ui/src/main/java/com/duckduckgo/common/ui/experiments/visual/store/VisualDesignExperimentDataStoreImpl.kt
```diff
@@ -16,149 +16,48 @@
...
-        targetToggle = experimentalUIThemingFeature.browserNavigationBar(),
-        prefsKey = KEY_NAVIGATION_BAR_ENABLED,
-        prefsDefault = true, // for now, navigation bar always enabled when the whole experiment enabled
-    )
-    override val navigationBarState: StateFlow<FeatureState> = _navigationBarState.state
-
-    init {
-        updateFeatureState()
-    }
+    override val isExperimentEnabled: StateFlow<Boolean> = _experimentFeatureFlagEnabled.asStateFlow()
```
LukasPaczos, line 47: In this flow we will be able to combine other feature flags for competing experiments before computing the enabled state. cc @malmstein

common/common-ui/src/main/java/com/duckduckgo/common/ui/experiments/visual/store/VisualDesignExperimentDataStoreImpl.kt
```diff
@@ -16,149 +16,48 @@
...
-        }.stateIn(
-            coroutineScope,
-            SharingStarted.Eagerly,
-            initialValue = FeatureState(
-                isAvailable = _available.value,
-                // FIXME Synchronously fetching the initial preference value is wrong but it's a temporary workaround
-                //  because the omnibar implementation that relies on this value can't recreate itself dynamically,
-                //  it requires the whole activity to be recreated. To avoid that recreation right after initial launch,
-                //  we need the initial value to actually use the valid persisted state.
-                isEnabled = _available.value && runBlocking { _enabled.firstOrNull() == true },
```
LukasPaczos, line 138: We're removing the synchronous check for the data store, only a single synchronous check for the feature toggle remains. However, by the time the feature toggle is accessed, the config should already be cached, so this should in theory resolve the minor ANR issues we've been seeing.

0nko (approved):
Works as expected 👍

PR #6067, opened 2025-05-14
Update Maestro E2E tests Autofill

cmonfortep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1210251123773109?focus=true 

### Description
Update Maestro Autofill E2E workflows

### Steps to test this PR

Should pass https://github.com/duckduckgo/Android/actions/runs/15025185850

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/.maestro/autofill/2_autofill_add_search_update_delete_creds.yaml
+++ b/.maestro/autofill/2_autofill_add_search_update_delete_creds.yaml
@@ -72,8 +72,6 @@ tags:
 
                 - tapOn: "Navigate up"
 
-                - assertVisible:
-                      text: "Save and autofill passwords"
                 - evalScript: ${output.addLogins.counter++}
 
       - scrollUntilVisible:
--- a/.maestro/autofill/4_autofill_settings_creds.yaml
+++ b/.maestro/autofill/4_autofill_settings_creds.yaml
@@ -30,9 +30,6 @@ tags:
           text: "Ask to Save and Autofill"
 
       - assertVisible:
-          id: importPasswordsOption
-
-      - assertVisible:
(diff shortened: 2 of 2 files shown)


.maestro/autofill/4_autofill_settings_creds.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - scrollUntilVisible:
+          element:
+            text: "fill.dev"
+      - assertVisible:
+          text: "fill.dev"
+
+      - assertNotVisible:
+          text: "fill.dev/example"
+
+      - tapOn: "Navigate up"
```
CDRussell, line 124: I don't understand this addition; why is there a new test that is largely a copy/paste of the existing `2_autofill_add_search_update_delete_creds.yaml` test?

Shouldn't this new test be about testing the autofill settings screen rather than acting on the password list screen still, which is already covered?

.maestro/autofill/2_autofill_add_search_update_delete_creds.yaml
```diff
@@ -72,8 +72,6 @@ tags:
 
                 - tapOn: "Navigate up"
 
-                - assertVisible:
-                      text: "Save and autofill passwords"
```
CDRussell, line 76: 👍

.maestro/autofill/4_autofill_settings_creds.yaml
cmonfortep: The test is validating the following:
- Access settings without passwords: counter is 0, options present
- We add passwords (copied from another flow)
- Go back to settings: counter is 3 (validates we show the right number)

PR #6074, opened 2025-05-15
Fix: IllegalStateException crash

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1210250194815685?focus=true

### Description

This PR fixes a crash caused by `FragmentStateAdapter` attempting to restore fragment state when `FragmentManager` loses its fragments after activity recreation.

### Steps to test this PR

- [x] Enable Don't keep activities
- [x] Set DDG as a default browser
- [x] Enable tab swiping
- [x] Open DDG → Open a few tabs
- [x] Go to tab manager
- [x] Switch to an app that can open links (like Twitter)
- [x] Tap on a link
- [x] Notice DDG opens the link and doesn't crash

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
+++ b/app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
@@ -29,6 +29,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewParent;
 import android.widget.FrameLayout;
+
 import androidx.annotation.CallSuper;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
```diff
@@ -107,7 +108,6 @@ public abstract class FragmentStateAdapter extends RecyclerView.Adapter<Fragment
     @SuppressWarnings("WeakerAccess") // to avoid creation of a synthetic accessor
     final LongSparseArray<Fragment> mFragments = new LongSparseArray<>();
 
-    private final LongSparseArray<Fragment.SavedState> mSavedStates = new LongSparseArray<>();
```
0nko, line 110: We don't use stored fragment state itself, we just need to know what fragments are stale in order to remove them when recreating an adapter, so we can remove this as it's only taking up space in a saved bundle.

karlenDimla (approved):
LGTM and worked as expected

PR #6086, opened 2025-05-16
fix an issue where nav bar would stay visible together with the keyboard when NTP was opened

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210267290350843?focus=true

### Description
When opening the new tab page, we initial don't show the keyboard. This scheduled a delayed task to execute and show the navigation bar. Before the time elapses to run the task, we automatically show the keyboard and focus on the omnibar, which hides the nav bar. However, the initially scheduled task would eventually execute and re-show the navigation bar even though the keyboard was also already visible.

The fix is to cancel the delayed task to show the navigation bar whenever the keyboard state changes.

### Steps to test this PR

- [ ] Enable visual design experiment.
- [ ] Go to any page.
- [ ] Open a new tab page.
- [ ] Verify keyboard is visible and navigation bar isn't.
- [ ] Close the app and reopen it.
- [ ] Verify that when NTP loads, keyboard is visible and navigation bar isn't.
- [ ] Try it multiple times.


malmstein (approved):
LGTM

PR #5829, opened 2025-03-28
Fix custom tab mode:

malmstein:
Task/Issue URL: https://app.asana.com/0/1157893581871903/1209822654738726/f

### Description
The custom tab mode wasn’t rendering properly. This PR fixes it

### Steps to test this PR

_Custom Tab_
- [x] Fresh install app and set it as default browser
- [x] Go to an app that will allow you to open a link (Reddit)
- [x] Tap on a link
- [x] Verify Custom Tab is displayed properly (tabs button is not visible, X is visible)
- [x] Open overflow menu and tap on “Open in DuckDuckGo"
- [x] Verify that the omnibar is displayed with the tabs button and without the X

_Malicious site protection_
- [x] Open https://privacy-test-pages.site/security/badware/
- [x] Navigate to https://privacy-test-pages.site/security/badware/phishing-iframe-loader.html
- [x] Once the error page is shown, open a new tab and load [wikipedia.org](https://wikipedia.org/)
- [x] While on the wikipedia tab, close and kill the app
- [x] Reopen the app, and wait for wikipedia to load
- [x] Switch to the previous tab
- [x] Check globe icon is shown

PR #6091, opened 2025-05-16
fix an issue where Custom Tab URL or icon wouldn't update according to page redirects

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210137390470561?focus=true

### Description
Adds a missing change to https://github.com/duckduckgo/Android/pull/5829 and ensures that the domain and/or duck chat icon are updated when Custom Tab page loads and redirects.

### Steps to test this PR

- [x] Set DDG as the default browser.
- [x] Open a link from Gmail or any other source that would redirect.
- [ ] Ensure that the domain is correct after the page fully loads.


0nko (approved):
The fix works! 🎖️

PR #6099, opened 2025-05-19
use fully opaque ripple mask

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210251609451213?focus=true

### Description
Uses a fully opaque (100% alpha) color for the ripple mask.

The actual color doesn't matter, but devices handle the alpha channel differently:
- On tested devices, a 0% alpha mask makes the ripple effect invisible.
- On tested Pixels, any alpha > 0% applies the ripple color correctly.
- On tested Samsungs, the ripple alpha is affected by both the mask and the ripple color alpha, summing them up and causing unexpected results.

To ensure consistent ripple visibility across devices, always use a fully opaque mask.

### Steps to test this PR

- [ ] Enable visual experiment changes.
- [ ] Verify that ripple is fully visible when clicking on buttons in the omnibar or navigation bar.
  - [ ] Test on a Samsung device.
  - [ ] Test on any other device.

### UI changes
| Before  | After |
| ------ | ----- |
Screenshot_20250519_102345|Screenshot_20250519_102420|

The screenshots come from a Samsung Tab S9+. You can see that on the _before_ screenshot, the ripple is barely visible because the 9% alpha of the color was summed with the 9% alpha of the mask (which was using the same color).

### DIFF
--- a/common/common-ui/src/main/res/drawable/selectable_item_experimental_background.xml
+++ b/common/common-ui/src/main/res/drawable/selectable_item_experimental_background.xml
@@ -17,7 +17,18 @@
     android:color="?daxColorControlFillPrimary">
     <item android:id="@android:id/mask">
         <shape android:shape="rectangle">
-            <solid android:color="?attr/daxColorControlFillPrimary" />
+            <!--
+              Use a fully opaque (100% alpha) color for the mask.
+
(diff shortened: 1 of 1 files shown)


malmstein (approved):
LGTM

PR #6109, opened 2025-05-21
fix an issue where omnibar would be stuck in wrong width after orientation change

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1157893581871903/task/1210317910216891?focus=true

### Description
Fixes an issue where omnibar would be stuck in wrong width after orientation change.

Initially, I thought this is because of the state update which occurs right after `onSizeChanged` and re-locks the dimensions but after refactoring the code to ensure animation is started only when needed, I realized that the issue was due to some race condition within the Android framework while the layout params were being updated directly in the `onSizeChanged` loop.

With the layout inspector, this is what I found out:
1. We are in an unfocused state, the card's content is `match_parent`, as expected:
Screenshot from 2025-05-21 11-36-51
2. We click into the omnibar and grab focus, the card's contents get locked so that they don't move while the card animates:
Screenshot from 2025-05-21 11-37-05
3. We rotate the screen and the dimensions get unlocked (they are `match_parent` now) but they view doesn't adjust to the new size.
Screenshot from 2025-05-21 11-37-15
4. Only after requesting another layout change by dropping focus (or by typing in another letter, or interacting with the omnibar in any other way) the view adjusts to the new width:
Screenshot from 2025-05-21 11-37-26

I found it reliable to wait for the first render loop to finish before unlocking the view, without having to refactor any animation state.

### Steps to test this PR

- [ ] Enable visual design flag.
- [ ] Focus on the omnibar and rotate the screen.
- [ ] Verify that the omnibar adjusts to the new width without needing to drop focus.
- [ ] Open "find in page" and rotate the screen.
- [ ] Verify that the "find in page" layout adjusts to the new width without needing to drop focus.
- [ ] When focus is dropped, the address bar shows without content jumping significantly.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/experiments/FadeOmnibarLayout.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/experiments/FadeOmnibarLayout.kt
@@ -18,7 +18,6 @@ package com.duckduckgo.app.browser.omnibar.experiments
 
 import android.animation.ValueAnimator
 import android.content.Context
-import android.content.res.Configuration
 import android.util.AttributeSet
 import android.view.View
 import android.view.ViewGroup
@@ -27,6 +26,7 @@ import android.view.animation.DecelerateInterpolator
 import android.widget.ImageView
 import android.widget.LinearLayout
 import androidx.core.animation.addListener
+import androidx.core.view.doOnLayout
 import androidx.core.view.isVisible
 import androidx.core.view.marginBottom
 import androidx.core.view.marginEnd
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/omnibar/experiments/FadeOmnibarLayout.kt
```diff
@@ -308,8 +313,8 @@ class FadeOmnibarLayout @JvmOverloads constructor(
     }
 
     private fun onFindInPageShown() {
-        omniBarContentContainer.gone()
-        customTabToolbarContainerWrapper.gone()
+        omniBarContentContainer.hide()
+        customTabToolbarContainerWrapper.hide()
```
LukasPaczos, line 317: This change ensures that when:
1. You open find in page.
2. Rotate the device.
3. Close find in page.

That the address bar is already in the correct position. Otherwise, if the view was `gone` while we're in find in page, it would need another render loop to re-adjust.

malmstein (approved):
Works as expected @LukasPaczos , thanks for the fix!

PR #6128, opened 2025-05-23
Increase RMF update frequency while removing parallel execution

cmonfortep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1149059203486286/task/1210364424198509?focus=true

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


cmonfortep:
* **#6128** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6128?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6128?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/RemoteMessagingConfigDownloadScheduler.kt
```diff
@@ -61,44 +66,33 @@ class RemoteMessagingConfigDownloadWorker(
...
             .addTag(REMOTE_MESSAGING_DOWNLOADER_WORKER_TAG)
             .setBackoffCriteria(BackoffPolicy.LINEAR, 10, TimeUnit.MINUTES)
             .build()
-        workManager.enqueueUniquePeriodicWork(REMOTE_MESSAGING_DOWNLOADER_WORKER_TAG, ExistingPeriodicWorkPolicy.REPLACE, workerRequest)
-    }
-
-    companion object {
-        private const val REMOTE_MESSAGING_DOWNLOADER_WORKER_TAG = "REMOTE_MESSAGING_DOWNLOADER_WORKER_TAG"
+        workManager.enqueueUniquePeriodicWork(REMOTE_MESSAGING_DOWNLOADER_WORKER_TAG, workPolicy, workerRequest)
     }
```
LukasPaczos, line 97: We expect the best time to check for RMF config changes is on app startup, as we've always done. With the new logic, we check for RMF config changes on app startup via scheduling a new worker when privacy config updates, since we assume privacy config also updates at launch. If that logic changes or breaks, RMF could be delayed by up to an hour.

To improve this, we could keep an independent RMF update on startup but add a short delay (like 10 seconds). This slightly decouples it from privacy config updates, ensuring it still runs shortly after launch even if the privacy update fails. In normal cases, where privacy config is updated at startup, it would cancel the delayed RMF update and avoid redundant work.

Something like this:
```kotlin
    private fun scheduleDownload() {
        val refreshInterval = if (remoteMessagingFeatureToggles.scheduleEveryHour().isEnabled()) 1L else 4L
        Timber.v("RMF: Scheduling remote config worker with fresh interval of $refreshInterval hours")
        val workerRequest = PeriodicWorkRequestBuilder<RemoteMessagingConfigDownloadWorker>(refreshInterval, TimeUnit.HOURS)
            .addTag(REMOTE_MESSAGING_DOWNLOADER_WORKER_TAG)
            .setBackoffCriteria(BackoffPolicy.LINEAR, 10, TimeUnit.MINUTES)
            .setInitialDelay(10, TimeUnit.SECONDS)
            .build()
        workManager.enqueueUniquePeriodicWork(REMOTE_MESSAGING_DOWNLOADER_WORKER_TAG, ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE, workerRequest)
    }
```

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/RemoteMessagingPrivacyConfigObserver.kt
```diff
@@ -0,0 +1,78 @@
...
+        appCoroutineScope.launch(context = dispatcherProvider.io()) {
+            val invalidateConfig = remoteMessagingFeatureToggles.invalidateRMFAfterPrivacyConfigDownloaded().isEnabled()
+            val alwaysProcess = remoteMessagingFeatureToggles.alwaysProcessRemoteConfig().isEnabled()
+            // we only need to invalidate, so we process config next run, if alwaysProcess is not enabled
+            val shouldInvalidate = invalidateConfig && alwaysProcess.not()
+
+            if (shouldInvalidate) {
+                Timber.d("RMF: onPrivacyConfigDownloaded, invalidate so next run we process the config")
+                remoteMessagingConfigRepository.invalidate()
+            }
```
LukasPaczos, line 57:
```suggestion

```

Is there any reason to keep `invalidateRMFAfterPrivacyConfigDownloaded` and its logic? `canScheduleOnPrivacyConfigUpdates` functionally achieves the same result (through a slightly different path), so I think we can remove it.

LukasPaczos (approved):
Generally LGTM and works as expected. We can merge as-is so I'm going to accept but I have two small suggestion/discussion points below that we could iterate on, let me know what you think.

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/RemoteMessagingPrivacyConfigObserver.kt
cmonfortep: Yes, but depends on the FF states. This is just to preserve optionality in case we need to react to anything during visual updates experiment. Imagine we disable `alwaysProcessRemoteConfig` and `canScheduleOnPrivacyConfigUpdates`, then we need this invalidate in place to at least know in 1h we will process then config again.

LukasPaczos (approved):
:shipit:

PR #6138, opened 2025-05-26
Consider rollouts before assigning cohorts

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1198194956794324/task/1210380290910042?focus=true

### Description
Before assigning any experiment cohort we need to make sure that the feature flag has entered rollout and targets do match, else cohort is not assigned.

### Steps to test this PR
Code review and check end to end tests added.

Alternatively or in addition, do the e2e tests using remote config (json blob etc)


malmstein (approved):
Thanks for fixing this @aitorvs . I’ve tested the different cases and also manually, works as expected.

PR #6142, opened 2025-05-26
Autofill: move store module as package inside impl

cmonfortep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1149059203486286/task/1210382174098207?focus=true 

### Description
Moves store module inside impl as package

### Steps to test this PR
Smoke testing

### UI changes
| Before  | After |
| ------ | ----- |


CDRussell (approved):
LGTM, smoke tested and all seems ok.

Only query I have is around the generated JSON schema files for Room. I'm not convinced we really need them, but the changes to them in this PR look a little suspect. It looks like new ones have been generated instead of having been moved. I'm not clear on the impact of that, if any, so flagging for your attention.

cmonfortep:
Those have no impact on production. Actually they stopped being generated after moving from kapt to ksp (now fixed).

PR #6158, opened 2025-05-29
remove Privacy Pro special URL test from release tests

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1210408722140193?focus=true

### Description
Remove the Privacy Pro special URL from release tests as we can't assert correctness of this behavior on Maestro Cloud emulators.

### Steps to test this PR

No QA needed.


nalcalag (commented):
@LukasPaczos I commented on Asana, but just a heads-up: if we don't have other tests that open the paywall, I’d suggest keeping this test for non-eligible users. If we ever support eligible accounts on Maestro, we can add the others then.

PR #6190, opened 2025-06-06
Measure concurrent load requests

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1211393669911397?focus=true

### Description

The PR adds a counter that measures all concurrent load requests across all `BrowserWebViewClient` instances. It uses `AtomicInteger` and `ConcurrentHashMap` to track the requests in progress and the count is reported to the `PageLoadedHandler`.

For the counter, there's a 30s time-out (maybe it should be longer?) so that we don't count requests that never finish.

### Steps to test this PR

_DB migration_
- [x] Build the app from `develop` and open it
- [x] Open a couple of tabs using these sites: "bbc.com", "weather.com", "reddit.com" or "ebay.com"
- [x] Wait a bit for all of them to load
- [x] Rebuild the app from this PR branch and run it
- [x] Notice the app doesn't crash and runs normally
- [x] Open the database inspector (app database, `page_loaded_pixel_entity` table) and check that `isTabInForegroundOnFinish` is false, `activeRequestsOnLoadStart` is 0 and `concurrentRequestsOnFinish` is 0 for all requests in the table

_1 request_
- [x] Kill the app
- [x] Start the app
- [x] Open a new tab and load `wikipedia.org`
- [ ] Open the database inspector and check that `isTabInForegroundOnFinish` is false, `activeRequestsOnLoadStart` is 0 and `concurrentRequestsOnFinish` is 0 for the request

_Multiple requests_
- [x] Keeping the same app instance, swipe fast a couple of times between tabs
- [x] Notice that the concurrent tab requests are reported
- [x] Open the database inspector and check that `isTabInForegroundOnFinish` is true for some requests and `activeRequestsOnLoadStart` and `concurrentRequestsOnFinish` is not 0 for some


app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -151,6 +159,39 @@ class BrowserWebViewClient @Inject constructor(
...
+
+    private fun decrementAndUntrackLoad(reason: String, urlForLog: String?) {
+        this.currentLoadOperationId?.let { loadId ->
+            val job = activeRequestTimeoutJobs.remove(loadId)
+
+            // if we successfully removed the job (it means it hadn't timed out yet)
+            if (job != null) {
+                job.cancel()
+                parallelRequestCounter.decrementAndGet()
+                logcat(LogPriority.DEBUG) { "Request $reason (ID: $loadId, URL: $urlForLog). Counter: ${parallelRequestCounter.get()}" }
```
CrisBarreiro: Do we need this log here?

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -524,7 +566,9 @@ class BrowserWebViewClient @Inject constructor(
                             title = navigationList.currentItem?.title,
                             start = safeStart,
                             end = currentTimeProvider.elapsedRealtime(),
-                            isTabInForeground = webViewClientListener?.isTabInForeground() ?: true,
+                            isTabInForegroundOnFinish = webViewClientListener?.isTabInForeground() ?: true,
+                            activeRequestsOnLoadStart = parallelRequestsOnStart,
+                            concurrentRequestsOnFinish = parallelRequestCounter.get() - 1,
```
CrisBarreiro: why not `decrementAndUntrackLoad` here, so that we have the actual number?

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -681,16 +736,19 @@ class BrowserWebViewClient @Inject constructor(
...
-            if (parsedError != OMITTED && request?.isForMainFrame == true) {
-                start = null
-                webViewClientListener?.onReceivedError(parsedError, request.url.toString())
-            }
+        error?.let { webResourceError ->
+            val parsedError = parseErrorResponse(webResourceError)
             if (request?.isForMainFrame == true) {
+                if (parsedError != OMITTED) {
+                    if (this.start != null) {
+                        decrementAndUntrackLoad("errored", request.url.toString())
```
CrisBarreiro: Have you verified we don't get onPageFinished as well in this case? Also, do we want to only call `webViewClientListener?.onReceivedError(parsedError, request.url.toString())` is mainframe? The change seems unrelated to the purpose of this PR

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
0nko: No, I just used it for testing.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
0nko: > Have you verified we don't get onPageFinished as well in this case?

Good call, I haven't. But it shouldn't matter because when decrementing the counter we check if `currentLoadOperationId` isn't null. So if the counter is decreased here, it'll just skip decrementing in `onPageFinished()`. I've modified the `decrementAndUntrackLoad()` function so that it decrements and returns the current count in a single call, thus it will always have the correct value in `onPageFinished()`.

> Also, do we want to only call webViewClientListener?.onReceivedError(parsedError, request.url.toString()) is mainframe? The change seems unrelated to the purpose of this PR

The logic to call the listener didn't change. This was the original code (it's easier to see without the diff):

```
            val parsedError = parseErrorResponse(it)
            if (parsedError != OMITTED && request?.isForMainFrame == true) {
                start = null
                webViewClientListener?.onReceivedError(parsedError, request.url.toString())
            }
            if (request?.isForMainFrame == true) {
                logcat { "recordErrorCode for ${request.url}" }
                webViewClientListener?.recordErrorCode(
                    "${it.errorCode.asStringErrorCode()} - ${it.description}",
                    request.url.toString(),
                )
            }
```

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
0nko: I've updated the decrement function to do this in a single call.

CrisBarreiro:
Approved under the assumption that
> Open the database inspector and check that isTabInForegroundOnFinish is false, activeRequestsOnLoadStart is 0 and concurrentRequestsOnFinish is 0 for the request

Contains a typo and we expect isTabInForegroundOnFinish to be true

PR #6210, opened 2025-06-11
 initialize VisualDesignExperimentDataStore on a worker thread (take 2)

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1210355090527279?focus=true

### Description
Follow-up of #6181 
* LaunchBridgeActivity now extends DaggerActivity directly instead of DuckDuckGoActivity. Since it doesn't include any styleable elements, it doesn't need to load the app theme. This allows the visual design data store - needed later to retrieve the theme instance - to be prepared asynchronously.
* VisualDesignExperimentDataStoreImpl is now wrapped in VisualDesignExperimentDataStoreLazyProvider, which adds an initialize function. This allows LaunchBridgeActivity to pre-warm the data store while the splash screen is visible, before launching other activities.

Together, these changes fully move VisualDesignExperimentDataStore initialization to a worker thread, avoiding any feature flag reads or iterations on the main thread.

On top of that PR: Removes `assert` statement that would crash the app after using the Fire Button due to launching `BrowserActivity` without first launching `LaunchBridgeActivity`

### Steps to test this PR

_Feature 1_
- [ ] Launch the app and verify the production theme is loaded.
- [ ] Change configuration to https://www.jsonblob.com/api/1379819763261431808 and verify that the new visual design is loaded, without any flickers.
- [ ] Use the Fire Button and verify the app restarts normally


LukasPaczos:
Closing here, I'm going to look for another way without bypassing pre-warming.

PR #6254, opened 2025-06-18
update the TabDataRepository to use a single source of truth for tab properties

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210530166473085?focus=true

### Description

Updates the `TabDataRepository` to use a single source of truth for tab properties instead of creating independent database queries and observable objects for both `LiveData` and `Flow` variants.

This resolves the original issue where `adClickManager` was receiving new tabs without source URLs when tabs swiping feature was enabled:
https://github.com/duckduckgo/Android/blob/ef6a87d670c66f6764d8f97bd11b8ccc0cf03076/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt#L1657-L1661

This issue was caused by `TabDataRepository#liveTabs` returning `null` because with tabs swiping feature enabled there was no entity actually observing the `LiveData` object so it was producing no values.

Now, even if only the `Flow` object is observed, it will transitively activate the `LiveData` object and the `adClickManager` will receive the correct tab properties.

This make the `LiveData` object the source of truth. I explored making the `Flow` the source of truth but it would require a more complex changes to the eager initialization of the `LiveData` that seem unnecessary. We should remove the `LiveData` object in the future whenever tab swiping feature flag is removed and it becomes the only code path.

### Steps to test this PR

1. Hardcode a config with tab swiping enabled (https://www.jsonblob.com/api/1384585817120497664).
2. Load https://www.search-company.site/#ad-id-5.
3. Click on the ad.
4. Click on the "Publisher site".
5. Click on the "Green T-shirt". This will open in a new tab.
6. Verify that `convert.ad-company.site` requests are loaded.

https://github.com/user-attachments/assets/7ea2f9eb-a713-47d4-a500-54961ff9c763

### DIFF
--- a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
+++ b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
@@ -27,4 +27,4 @@ enum class PrivacyFeatureName(val value: String) {
     TrackingParametersFeatureName("trackingParameters"),
 }
 
-const val PRIVACY_REMOTE_CONFIG_URL = "https://www.jsonblob.com/api/1384585817120497664"
+const val PRIVACY_REMOTE_CONFIG_URL = "https://staticcdn.duckduckgo.com/trackerblocking/config/v4/android-config.json"


LukasPaczos:
I pushed a commit that enables tab swiping and kicked off a privacy dashboard E2E job [here](https://github.com/duckduckgo/Android/actions/runs/15737207513) to verify the fix.

privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
```diff
@@ -27,4 +27,4 @@ enum class PrivacyFeatureName(val value: String) {
     TrackingParametersFeatureName("trackingParameters"),
 }
 
-const val PRIVACY_REMOTE_CONFIG_URL = "https://staticcdn.duckduckgo.com/trackerblocking/config/v4/android-config.json"
+const val PRIVACY_REMOTE_CONFIG_URL = "https://www.jsonblob.com/api/1384585817120497664"
```
0nko: We probably don't want this merged :)

LukasPaczos: Right, that's why I tagged the commit with `[DO NOT MERGE]` :sweat_smile: just waiting for the test to pass https://github.com/duckduckgo/Android/pull/6254#issuecomment-2984717491.

0nko (approved):
The code looks good and works as expected! 🥇 Great job!

0nko:
All the E2E tests have passed. I've reversed the hardcoded privacy config. Since @LukasPaczos is AFK, I'll go ahead and merge this.

PR #6265, opened 2025-06-20
Additional pixels for Duck.ai discoverability tracking

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/715106103902962/task/1210599747553654?focus=true

### Description
Adds more Duck.ai discovery pixels.

### Steps to test this PR
- [x] Use this config which enables Duck.ai button in all omnibar states - https://www.jsonblob.com/1385582049771053056.
  - [x] Optionally, set your logcat filter to `aichat_`.
- [x] Clear the caches and launch the app.
- [x] Close the keyboard and unfocus the omnibar.
- [x] Click on the Duck.ai icon in omnibar while on new tab page.
- [x] Verify that `aichat_searchbar_button_open` is sent with `was_used_before=0` and `source=ntp`.
- [x] Enable "Experimental UI Settings -> Enable Visual design updates".
- [x] Restart the app.
- [x] Focus on the omnibar.
- [x] Click on the Duck.ai icon in omnibar.
- [x] Verify that `aichat_experiment_searchbar_button_open` is sent with `was_used_before=1` and `source=focused`.
- [x] Search for anything.
- [x] While on SERP and with omnibar unfocused, click on the Duck.ai icon in omnibar.
- [x] Verify that `aichat_experiment_searchbar_button_open` is sent with `was_used_before=1` and `source=serp`.
- [x] Go to a website.
- [x] Click on the Duck.ai icon in omnibar.
- [x] Verify that `aichat_experiment_searchbar_button_open` is sent with `was_used_before=1` and `source=website`.
- [x] Go to the tab switcher and click on the Duck.ai FAB.
- [x] Verify that `aichat_open_tab_switcher_fab` is sent with `was_used_before=1`.

### DIFF
--- a/app/src/test/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModelTest.kt
+++ b/app/src/test/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModelTest.kt
@@ -861,7 +861,6 @@ class TabSwitcherViewModelTest {
 
         testee.onDuckChatFabClicked()
 
-        verify(mockPixel).fire(DuckChatPixelName.DUCK_CHAT_OPEN)
         verify(mockPixel).fire(DuckChatPixelName.DUCK_CHAT_OPEN_TAB_SWITCHER_FAB, mapOf("was_used_before" to "0"))
         verify(duckChatMock).openDuckChat()
     }
@@ -872,7 +871,6 @@ class TabSwitcherViewModelTest {
 
         testee.onDuckChatFabClicked()
 
-        verify(mockPixel).fire(DuckChatPixelName.DUCK_CHAT_OPEN)
         verify(mockPixel).fire(DuckChatPixelName.DUCK_CHAT_OPEN_TAB_SWITCHER_FAB, mapOf("was_used_before" to "1"))
         verify(duckChatMock).openDuckChat()
     }


LukasPaczos:
I'm going to keep this in draft while the pixel changes are in triage, but otherwise, the code is ready for review.

LukasPaczos:
This is ready for a final review.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -131,7 +132,7 @@ class OmnibarLayoutViewModel @Inject constructor(
                 (state.viewMode is NewTab || state.hasFocus && state.omnibarText.isNotBlank() || duckChat.isEnabledInBrowser()),
             showClickCatcher = isDuckAIPoCEnabled,
         )
-    }.flowOn(dispatcherProvider.io()).stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000L), _viewState.value)
```
joshliebe, line 134: Do you know why this was 5000ms before? Is it safe to change?

joshliebe (approved):
LGTM, just the one question/comment: https://github.com/duckduckgo/Android/pull/6265#discussion_r2161166513

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
LukasPaczos: I had a comment drafted for this which I didn't submit :facepalm: Here goes:

I'm making this flow eager because the many direct value checks in this class can behave unpredictably if the flow doesn’t emit values, potentially causing subtle bugs. In practice, the flow is always observed by the omnibar, so this change shouldn't impact behavior but removes that risk and simplifies unit testing.

I spoke with @0nko some time ago and the timeout was added as a resource-saving precaution in case the flow wasn’t consumed. But since in our case it's always expected to be consumed, and the internal logic depends on emitted values, I think it's safer to make that explicit by using an eager flow. @0nko please keep me honest about it though in case there's a reason to keep the flow lazy.

0nko: @LukasPaczos That's right, the original reason still holds, I've considered `SharingStarted.WhileSubscribed(5000L)` sort of the default setting to save resources based on some tutorials when I first started using the flows, unless there was a reason to change it, which it seems, we do have. So all good 👍.

PR #6306, opened 2025-06-27
do not change address bar content if no query was made in the Duck.ai Input Screen

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210657930367705?focus=true

### Description
Revert back to the original omnibar content if Duck.ai Input screen query was not submitted.

### Steps to test this PR

- [ ] Go to any website
- [ ] Open input screen and change content
- [ ] Go back
- [ ] Verify the website URL is restored in the omnibar

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
@@ -802,4 +802,15 @@ class OmnibarLayoutViewModel @Inject constructor(
         //     }
         // }
     }
+
+    fun setDraftTextIfNtp(query: String) {
+        if (_viewState.value.viewMode is NewTab) {
+            _viewState.update {
+                it.copy(
+                    omnibarText = query,
+                    updateOmnibarText = true,
+                )
+            }
+        }
+    }
 }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -873,7 +873,7 @@ class BrowserTabFragment :
             if (result.resultCode == RESULT_OK) {
                 submitQuery(query)
             } else {
-                omnibar.setText(query)
+                omnibar.setDraftTextIfNtp(query)
             }
         }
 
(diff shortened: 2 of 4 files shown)


duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/SearchInterstitialFragment.kt
```diff
@@ -142,8 +140,7 @@ class SearchInterstitialFragment : DuckDuckGoFragment(R.layout.fragment_search_i
             exitInterstitial()
         }
         onDuckChatSent = { query ->
-            val data = Intent().putExtra(SearchInterstitialActivity.QUERY, query)
-            requireActivity().setResult(Activity.RESULT_CANCELED, data)
+            requireActivity().setResult(Activity.RESULT_CANCELED)
```
joshliebe: This one was reported as a bug in https://app.asana.com/1/137249556945/project/1210424118459703/task/1210513464852533 (3)

LukasPaczos: >expected: app should take the user to main search screen and address bar should have "hello"
 actual: app takes the user to main search screen but address bar is empty

Looks like more of an issue was that was empty. I think we should keep aligned with production/iOS and revert back to the initial state.

LukasPaczos:
As discussed internally, we also need to handle the NTP case when text from Input Screen should be transferred to the omnibar. I'll give it a try.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -802,4 +802,15 @@ class OmnibarLayoutViewModel @Inject constructor(
...
+    fun setDraftTextIfNtp(query: String) {
+        if (_viewState.value.viewMode is NewTab) {
+            _viewState.update {
+                it.copy(
+                    omnibarText = query,
+                    updateOmnibarText = true,
+                )
+            }
+        }
+    }
```
LukasPaczos, line 815: I'm not truly happy with this approach, but it's the simplest way I could find that improves the experience for the POC without impacting any production codebase. Let me know what you think or if you have any other ideas @joshliebe!

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -802,4 +802,15 @@ class OmnibarLayoutViewModel @Inject constructor(
         //     }
         // }
     }
+
+    fun setDraftTextIfNtp(query: String) {
```
joshliebe, line 806: nit: why “draft” text?

PR #6308, opened 2025-06-27
move duck.ai input screen return button, improve input string management

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210659642047207?focus=true

### Description
- Moves the return button in the Duck.ai input screen to the side of the submit button.
- Fixes an issue where the submit button was visible when only new lines were in the input field.
- Fixes an issue where return key didn't replace selected text.
- Fixes voice input button's background color.

### Steps to test this PR
- [x] Verify that even if voice input is enabled, return key is next to submit button.
- [x] Open the input screen, clear text, add new lines. Verify the submit button is not visible.
- [x] Input some text, select all or portion of it, click the return key and verify that the whole selection is replaced.


joshliebe (approved):
Nice improvement, LGTM

PR #6354, opened 2025-07-07
Fix ANR/Crash on autoconsent init handler

marcosholgado:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1210704553451145?focus=true 

### Description
This PR tries to fix and ANR and crash in the autoconsent init handler. It does remove the call to trimIndent() which is not very performant for long strings (and doesn't seem to be needed) and it moves all the logic to IO leaving only the webview call to evaluate JS in main.

### Steps to test this PR

- [ ] Test autoconsent

### DIFF
--- a/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/ReplyHandler.kt
+++ b/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/ReplyHandler.kt
@@ -16,45 +16,12 @@
 
 package com.duckduckgo.autoconsent.impl.handlers
 
-import java.io.BufferedReader
-import java.io.StringReader
-import kotlin.sequences.forEach
-
 object ReplyHandler {
     fun constructReply(message: String): String {
         return """
             (function() {
                 window.autoconsentMessageCallback($message, window.origin);
             })();
-        """.trimIndentEfficient()
-    }
-
-    private fun String.trimIndentEfficient(): String {
-        val reader1 = BufferedReader(StringReader(this))
-        var minIndent = Int.MAX_VALUE
-        reader1.useLines { lines ->
-            lines.forEach { line ->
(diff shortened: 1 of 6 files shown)


marcosholgado:
* **#6354** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6354?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6354?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/ReplyHandler.kt
```diff
@@ -22,6 +22,6 @@ object ReplyHandler {
             (function() {
                 window.autoconsentMessageCallback($message, window.origin);
             })();
-        """.trimIndent()
+        """
```
karlenDimla: @marcosholgado this method is used in other places. Are those references fine without the trimIndent()?

marcosholgado: I guess we'll see once @sammacbeth tests this :D. It's only used within autoconsent messages so I assume yes.

marcosholgado: Having a look at this, I think it's fine removing trimIndent(). The only reason is there is because of the way we construct the reply and looking back it seems it was more for tests than for anything else.

karlenDimla (approved):
I’ve tested this so far with a few websites with cookie prompts (on pixel and samsung) and it seems to work as expected.

PR #6357, opened 2025-07-07
Optimize bookmarks filtering and sorting

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1210718637089838?focus=true

### Description
1. Adds a new dev setting to populate saved sites repo with random bookmarks.
2. Optimizes bookmark search by skipping relative position recalculations. When filtering, item order remains unchanged, allowing us to bypass an extra round of equality checks.
3. Moves bookmark update diff computation to a worker thread to avoid UI thread contention and ANRs.
4. Decouples bookmark items to display from the rest of the state to avoid unnecessary computations and UI updates. Previously, updating the search query triggered a view state change (due to the persisted `ViewState#searchQuery`), which caused the `RecyclerView` to recalculate before the actual filtering occurred. As a result, the `RecyclerView` was recalculated and updated twice per query, which was also contributing to resource contention.

### Steps to test this PR

- [x] With bookmarks populated, open "Menu -> Bookmarks".
- [x] Verify that both "manual" and "by name" options work correctly.
- [x] Verify that searching for tabs works correctly. 
- [x] Try with an extreme number of tabs (500+) and verify that there are no ANRs or main thread hangs when sorting/filtering. 
  - [x] (optional) Try the new dev setting under "Developer Settings -> Tabs and Saved Sites"


0nko (approved):
Everything works as expected, great work! 👏

PR #6366, opened 2025-07-09
Use auth v2 in subscription activation flow

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205648422731273/task/1210723642420067?focus=true

### Steps to test this PR

- [ ] Build with staging patch from https://app.asana.com/1/137249556945/project/1209991789468715/task/1210448620621729?focus=true
- [ ] Go to settings -> I have a subscription
- [ ] Click "Add via Email"
- [ ] Verify that the app made request to api/auth/v2/.well-known/jwks.json
- [ ] Proceed with resotoring the subscription until the success screen is shown
- [ ] Verify that there were no requests to api/auth/access-token
- [ ] Verify that there were no more requests to api/auth/v2/.well-known/jwks.json
- [ ] Verify that the app settings show subscription in active state, VPN can be enabled, etc.


lmac012:
* **#6366** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6366?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/auth2/AuthClient.kt
```diff
@@ -183,8 +191,12 @@ class AuthClientImpl @Inject constructor(
         )
     }
 
-    override suspend fun getJwks(): String =
-        authService.jwks().string()
+    override suspend fun getJwks(): String {
```
cmonfortep, line 200: Q: I have not enough context here. About caching this for 1h, checking if there's any risk of doing that, should we worry about concurrency here, should we consider adding an specific FF for this change, etc.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/SubscriptionsManager.kt
```diff
@@ -382,6 +387,21 @@ class RealSubscriptionsManager @Inject constructor(
         }
     }
 
+    override suspend fun signInV2(
+        accessToken: String,
+        refreshToken: String,
+    ) {
+        val tokens = TokenPair(accessToken, refreshToken)
+        val jwks = authClient.getJwks()
```
cmonfortep, line 395: Can this throw an exception since it could hit network? should we try-catch?

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/messaging/SubscriptionMessagingInterface.kt
```diff
@@ -222,6 +223,43 @@ class SubscriptionMessagingInterface @Inject constructor(
...
+            }
+
+            appCoroutineScope.launch(dispatcherProvider.io()) {
+                try {
+                    subscriptionsManager.signInV2(accessToken, refreshToken)
+                    subscriptionsChecker.runChecker()
+                    pixelSender.reportRestoreUsingEmailSuccess()
+                    pixelSender.reportSubscriptionActivated()
+                } catch (e: Exception) {
+                    logcat { "Failed to set auth tokens" }
```
cmonfortep, line 253: is there a pixel we can send here to signal a problem and reason? or the methods we are using will already signal enough info to detect any unexpected failures?

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/RealSubscriptions.kt
```diff
@@ -182,7 +182,7 @@ interface PrivacyProFeature {
      * This flag will be used to select FE subscription messaging mode.
      * The value is added into GetFeatureConfig to allow FE to select the mode.
      */
-    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    @Toggle.DefaultValue(DefaultFeatureValue.INTERNAL)
```
cmonfortep, line 185: What's your plan for this?

I see two options, and we should decide depending risk evaluation here:
a) we release enabled by default. If something goes wrong we disable via remote config
b) we release disabled by default, and from remote config we enable it with rollout.

If you pick B, best to add the annotation AlwaysInternalEnabled to avoid inconsistencies in internal buids.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/auth2/AuthClient.kt
lmac012: I've considered concurrency here and I don't think it is a concern. Yes, there is a (small) chance that we'll make more requests to fetch jwks than necessary when/if this method is called by multiple coroutines in parallel, but given the fact this method will not be called often, the impact on overall efficiency should be negligible. The important thing is that there is no potential correctness/consistency issue here, given the nature of the cached data. Caching the token for 1 hour is actually recommended by the API documentation - I'll provide more details in the asana task. I don't think this is a risky change, but adding a kill-switch won't hurt. I'll append that change to this PR.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/SubscriptionsManager.kt
lmac012: Yes, it can throw an exception, and I'm intentionally not caching it here because `signInV2()` should throw and signal the caller that the operation has failed. The invocation of `signInV2()` is wrapped with try-catch to capture all things that might go wrong, including a network issue.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/messaging/SubscriptionMessagingInterface.kt
lmac012: We do have some pixels that are fired inside SubscriptionManager to indicate a few specific errors that may happen. They don't necessarily cover all the possible failures. From the rollout perspective, the most important thing will be to monitor the `reportRestoreUsingEmailSuccess()` - if we notice a drop correlating with the rollout of this change, then we'll dig deeper, potentially adding more pixels.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/RealSubscriptions.kt
lmac012: The idea here was to enable it internally ASAP, and then once the code is in production (and no issues reported internally), enable it for everyone. I'm not sure how much benefit there is to do a gradual rollout of this, given the very low volume of users going through this flow.

cmonfortep: "very low volume of users going through this flow", ok then no need for rollout in this case. If that's the case, we can enable it internal for now, happy to merge as it is.

PR #6380, opened 2025-07-11
add internal remote config patch tool

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210708742763735

### Description
Adds tooling to locally patch remote config. Read the details in the new README file in the diff.

### Steps to test this PR

1. Create a `privacy-config/privacy-config-internal/local-patch-file/test_patch.json` file with this content:
```json
[
  {
    "op": "replace",
    "path": "/features/aiChat/state",
    "value": "disabled"
  }
]
```
2. Create a `privacy-config/privacy-config-internal/local.properties` file with this content:
```
config_patches=privacy-config/privacy-config-internal/local-config-patches/test_patch.json
``` 
3. Clean install the app.
4. Verify Duck.ai is disabled and that Settings -> Duck.ai is gone.
5. Update the patch file to:
```json
[
  {
    "op": "replace",
    "path": "/version",
    "value": "2752244802866"
  },
  {
    "op": "remove",
    "path": "/features/aiChat/hash"
  }
]
```
6. Relaunch the app (without clean installing it). This will download the production config and force update the feature because of the bumped version number and mismatching hash.
7. Verify Duck.ai is enabled and that Settings -> Duck.ai is available.
8. Update the `test_patch.json` to disable Duck.ai again and bump version:
```json
[
  {
    "op": "replace",
    "path": "/features/aiChat/state",
    "value": "disabled"
  },
  {
    "op": "remove",
    "path": "/features/aiChat/hash"
  },
  {
    "op": "replace",
    "path": "/version",
    "value": "3752244802866"
  }
]
```
9. Relaunch the app (without clean installing it).
10. Verify Duck.ai is disabled and that Settings -> Duck.ai is gone.

### DIFF
--- a/privacy-config/privacy-config-internal/src/main/java/com/duckduckgo/privacy/config/internal/plugins/DevPrivacyConfigPatchApiInterceptor.kt
+++ b/privacy-config/privacy-config-internal/src/main/java/com/duckduckgo/privacy/config/internal/plugins/DevPrivacyConfigPatchApiInterceptor.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2025 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.privacy.config.internal.plugins
+
+import android.content.Context
+import com.duckduckgo.app.global.api.ApiInterceptorPlugin
+import com.duckduckgo.di.scopes.AppScope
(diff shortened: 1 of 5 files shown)


privacy-config/privacy-config-internal/src/main/java/com/duckduckgo/privacy/config/internal/plugins/DevPrivacyConfigPatchApiInterceptor.kt
```diff
@@ -0,0 +1,115 @@
...
+import okhttp3.Interceptor
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.Response
+import okhttp3.ResponseBody.Companion.toResponseBody
+
+@ContributesMultibinding(
+    scope = AppScope::class,
+    boundType = ApiInterceptorPlugin::class,
+)
+class DevPrivacyConfigPatchApiInterceptor @Inject constructor(
```
LukasPaczos, line 39: All interceptors are currently added in default order, sorted by class name. If I assign a low priority to this interceptor without changing the others, it would run first, since it would be the only one with a defined priority. I tried giving all plugins the same priority (to preserve class name sorting as `PriorityKey` docs suggest) and setting this interceptor last, but the build fails due to duplicate priority keys, despite what the docs say.

I don’t want to risk changing any production behavior, so it’s fine if this interceptor doesn’t run last for now. If it ever becomes a problem, we can revisit.

PR #6387, opened 2025-07-14
Update content scope scripts workflow description

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1210785772751175?focus=true

### Description

This PR enhances the automated content scope scripts update workflow by:

1. Clarifying when tests need to run:
   - Tests will only run if the `node_modules/@duckduckgo/content-scope-scripts` folder content has changed
   - If only the package version changed, tests don't need to run

2. Adding clearer PR checklist with two scenarios:
   - When content-scope-scripts folder is updated: all tests and privacy tests must pass
   - When only the package version is updated: all tests must pass but privacy tests can be skipped

3. Updating the Asana task description to include the same clarification about when tests need to run

### Steps to test this PR

_Workflow Changes_
- [x] Confirm PR description includes the correct checklist based on what was updated

### DIFF
--- a/.github/workflows/update-content-scope.yml
+++ b/.github/workflows/update-content-scope.yml
@@ -70,9 +70,20 @@ jobs:
             - Automated content scope scripts dependency update
 
             This PR updates the content scope scripts dependency to the latest available version and copies the necessary files.
+
+            Tests will only run if something has changed in the `node_modules/@duckduckgo/content-scope-scripts` folder.
+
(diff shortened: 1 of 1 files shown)


mikescamell:
The failure was Asana sync, my AI description stripped the task url 🙄 I will manually update

mikescamell:
Bypassed as it's just an update to the workflow file description and our usual checks probably should not apply to workflow files as they didn't in the past.

PR #6397, opened 2025-07-15
Add support for reportMetric message from CSS

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/45878998844068/task/1210462957793120?focus=true 

> **Warning**
Before merging this PR, we should first merge https://github.com/duckduckgo/content-scope-scripts/pull/1766 and update CSS version on the app

### Description
Fire pixel when `reportMetric` message comes from CSS

### Steps to test this PR

_Feature 1_
- [ ] Apply this [patch](https://duckduckgo-my.sharepoint.com/:u:/p/cbarreiro/Ed0uViGgej9Fv5aTsPdezCsBrowApZtiGLsrVbYxf3HY8A?e=rr4nXv) to update CSS version and enable forcing the error on the Duck Player page
- [ ] Load any video on Duck Player, edit the URL appending `?willThrow=true` and load again
- [ ] Check `duckplayer_js-error with params: {message=Simulated Exception, kind=Error, origin=duckPlayerPage} {}` is fired

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/duckplayer/DuckPlayerJSHelper.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/duckplayer/DuckPlayerJSHelper.kt
@@ -27,6 +27,7 @@ import com.duckduckgo.app.browser.commands.Command.SendResponseToJs
 import com.duckduckgo.app.browser.commands.Command.SendSubscriptions
 import com.duckduckgo.app.browser.commands.NavigationCommand.Navigate
 import com.duckduckgo.app.pixels.AppPixelName
+import com.duckduckgo.app.pixels.AppPixelName.DUCK_PLAYER_JS_ERROR
 import com.duckduckgo.app.pixels.AppPixelName.DUCK_PLAYER_SETTING_ALWAYS_DUCK_PLAYER
 import com.duckduckgo.app.pixels.AppPixelName.DUCK_PLAYER_SETTING_ALWAYS_OVERLAY_YOUTUBE
 import com.duckduckgo.app.pixels.AppPixelName.DUCK_PLAYER_SETTING_ALWAYS_SERP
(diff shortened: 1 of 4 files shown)


CrisBarreiro:
* **#6397** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/6397?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/6397?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/duckplayer/DuckPlayerJSHelper.kt
```diff
@@ -291,6 +292,19 @@ class DuckPlayerJSHelper @Inject constructor(
                 pixel.fire(impressionPixelName)
                 pixel.fire(dailyPixelName, emptyMap(), emptyMap(), Daily())
             }
+            "reportMetric" -> {
```
marcosholgado: IMO we are leaking by default a lot of implementation details and using `DuckPlayerPageHandler` as a bypass rather than extracting messages to their own handler. Then all these messages have a meaningless trip around the tab fragment, view model, js helper, etc. 
This is a perfect example of that, you could do something like this instead which will keep all the logic in impl (I would actually do an internal plugin just for duck player to not have a lot of inner classes and sort out DI but this works for now anyways).

```
inner class DuckPlayerReportMetricHandler(private val pixel: Pixel) : JsMessageHandler {
        override fun process(jsMessage: JsMessage, secret: String, jsMessageCallback: JsMessageCallback?) {
            try {
                val params = jsMessage.params
                val message = params.getString("message") ?: return
                val kind = params.getString("kind") ?: return
                pixel.fire(
                    DUCK_PLAYER_JS_ERROR,
                    mapOf("message" to message, "kind" to kind, "origin" to featureName),
                )
            } catch (e: Exception) {
                logcat { "Error reporting metric: $e" }
            }
        }

        override val allowedDomains: List<String> = listOf(
            runBlocking { duckPlayer.getYouTubeEmbedUrl() },
        )
        override val featureName: String = "duckPlayerPage"
        override val methods: List<String> = listOf(
            "reportMetric",
        )
    }
```

marcosholgado (approved):
LGTM

PR #6409, opened 2025-07-16
suppress autocomplete when Input Screen initially opened from a web page

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210716487258255?focus=true

### Description
- Suppresses the initial autocomplete request when user open the Input Screen if the input text is the current web page's URL.
- Updates the autocomplete request logic to not debounce the initial request. This helps speed up autocomplete results, for example, when user comes from SERP or from NTP. In these situations, they don't need to wait additional 300ms for us to request suggestions which visibly speeds up the interaction.

### Steps to test this PR

- [x] Make sure you have some favorites set.
- [x] Open Input screen from NTP and verify favorites are visible.
- [x] Open Input Screen from SERP and verify autocomplete suggestions are immediately shown and there's no blink of favorites icons.
- [x] Clear input text and verify favorites are shown.
- [x] Go to a website and open the Input Screen. Verify there's no suggestions and you can see favorites.
- [x] Start typing and verify you can see suggestions.

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
@@ -16,9 +16,11 @@
 
 package com.duckduckgo.duckchat.impl.inputscreen.ui.viewmodel
 
+import android.util.Patterns
+import androidx.core.net.toUri
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
-import com.duckduckgo.anvil.annotations.ContributesViewModel
 import com.duckduckgo.app.di.AppCoroutineScope
 import com.duckduckgo.browser.api.autocomplete.AutoComplete
 import com.duckduckgo.browser.api.autocomplete.AutoComplete.AutoCompleteResult
@@ -33,7 +35,6 @@ import com.duckduckgo.browser.api.autocomplete.AutoComplete.AutoCompleteSuggesti
 import com.duckduckgo.browser.api.autocomplete.AutoCompleteSettings
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.common.utils.SingleLiveEvent
-import com.duckduckgo.di.scopes.FragmentScope
 import com.duckduckgo.duckchat.impl.inputscreen.store.InputScreenDataStore
 import com.duckduckgo.duckchat.impl.inputscreen.store.InputScreenMode
 import com.duckduckgo.duckchat.impl.inputscreen.ui.command.Command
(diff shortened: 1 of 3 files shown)


duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
```diff
@@ -91,12 +99,73 @@ class InputScreenViewModel @Inject constructor(
...
+     * Tracks whether we should show autocomplete suggestions based on the initial input state.
+     *
+     * This becomes true when either:
+     * 1. The user has modified the input text from its initial state, OR
+     * 2. The initial text was not a URL (e.g., search query from SERP)
+     *
+     * We suppress autocomplete when the user is on a webpage and the input still shows
+     * that page's URL unchanged, since autocomplete suggestions would then obfuscate favorites.
+     */
+    private var hasMovedBeyondInitialUrl = false
```
harrytmthy, line 115: Since this is just a plain `var`, there's a risk of race conditions if the input changes rapidly. Consider using `AtomicBoolean` or `MutableStateFlow` for safer updates.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
```diff
@@ -91,12 +99,73 @@ class InputScreenViewModel @Inject constructor(
...
+
+            val shouldShow = userHasModifiedInput || initialTextWasNotWebUrl
+            if (shouldShow) {
+                hasMovedBeyondInitialUrl = true
+            }
+            shouldShow
+        }
+
+        autoCompleteEnabled &&
+            searchInputTextState.value.isNotEmpty() &&
```
harrytmthy: Directly using `.value` here could lead to inconsistent state, since `searchInputTextState` is mutable. Why not use the `searchInput` from the lambda instead?

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
LukasPaczos: The variable is only set and accessed from a single thread (main), so I don't see the risk of races. If that ever changes, we'll be able to address accordingly.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
LukasPaczos: Thanks, updated!

PR #6413, opened 2025-07-17
PIR: Add support for multiple profile queries in PirScan and PirOptOut

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1210594645151737/task/1210750227652612?focus=true

### Description
Adds support for running the initial PIR scan and PIR opt-out scan on multiple stored profiles instead of the first stored profile. See the task for more details and a full tech spec.

### Steps to test this PR

_1 profile (current behavior)_
- [ ] Enter a new profile and trigger scan - scan runs for 1 profile
- [ ] Go to opt-out scan and trigger it for any broker - scan runs for 1 profile

_Multiple profiles (new behavior)_
- [ ] Clear data and trigger scan without any input - scan runs for 3 hardcoded profiles
- [ ] Go to opt-out scan and trigger it for any broker - scan runs for all 3 profiles (assuming that the profiles were found for that broker)

### UI changes
No UI Changes

### DIFF
--- a/pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/common/BrokerStepsParser.kt
+++ b/pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/common/BrokerStepsParser.kt
@@ -23,6 +23,7 @@ import com.duckduckgo.pir.internal.common.BrokerStepsParser.BrokerStep.OptOutSte
 import com.duckduckgo.pir.internal.common.BrokerStepsParser.BrokerStep.ScanStep
 import com.duckduckgo.pir.internal.scripts.models.BrokerAction
 import com.duckduckgo.pir.internal.scripts.models.ExtractedProfile
+import com.duckduckgo.pir.internal.scripts.models.ProfileQuery
 import com.duckduckgo.pir.internal.store.PirRepository
 import com.squareup.anvil.annotations.ContributesBinding
 import com.squareup.moshi.JsonAdapter
@@ -40,12 +41,14 @@ interface BrokerStepsParser {
      *
      * @param brokerName - name of the broker to which these steps belong to
      * @param stepsJson - string in JSONObject format obtained from the broker's json representing a step (scan / opt-out).
+     * @param profileQuery - profile associated with the step (used for the opt-out step)
      * @return list of broker steps resulting from the passed params. If the step is of type OptOut, it will return a list of
      *  OptOutSteps where an OptOut step is mapped to each of the profile for the broker.
      */
     suspend fun parseStep(
         brokerName: String,
         stepsJson: String,
+        profileQuery: ProfileQuery? = null,
     ): List<BrokerStep>
 
     sealed class BrokerStep(
@@ -101,16 +104,22 @@ class RealBrokerStepsParser @Inject constructor(
     override suspend fun parseStep(
         brokerName: String,
         stepsJson: String,
+        profileQuery: ProfileQuery?,
     ): List<BrokerStep> = withContext(dispatcherProvider.io()) {
         return@withContext runCatching {
             adapter.fromJson(stepsJson)?.run {
                 if (this is OptOutStep) {
-                    repository.getExtractProfileResultForBroker(brokerName)?.extractResults?.map {
-                        this.copy(
-                            brokerName = brokerName,
-                            profileToOptOut = it,
-                        )
-                    }
+                    repository.getExtractProfileResultsForBroker(brokerName)
+                        .filter { it.profileQuery != null && it.profileQuery.id == profileQuery?.id }
+                        .map {
+                            it.extractResults.map { extractedProfile ->
+                                this.copy(
+                                    brokerName = brokerName,
+                                    profileToOptOut = extractedProfile,
+                                )
+                            }
+                        }
+                        .flatten()
                 } else {
                     listOf((this as ScanStep).copy(brokerName = brokerName))
                 }
--- a/pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
+++ b/pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
@@ -101,21 +101,41 @@ class RealPirOptOut @Inject constructor(
     private val dispatcherProvider: DispatcherProvider,
     callbacks: PluginPoint<PirCallbacks>,
 ) : PirOptOut, PirJob(callbacks) {
-    private var profileQuery: ProfileQuery = ProfileQuery(
-        firstName = "William",
-        lastName = "Smith",
-        city = "Chicago",
-        state = "IL",
-        addresses = listOf(
-            Address(
-                city = "Chicago",
-                state = "IL",
+    private var profileQueries: List<ProfileQuery> = listOf(
+        ProfileQuery(
+            id = -1,
+            firstName = "William",
+            lastName = "Smith",
+            city = "Chicago",
+            state = "IL",
+            addresses = listOf(
+                Address(
+                    city = "Chicago",
+                    state = "IL",
+                ),
             ),
+            birthYear = 1993,
+            fullName = "William Smith",
+            age = 32,
+            deprecated = false,
         ),
-        birthYear = 1993,
-        fullName = "William Smith",
-        age = 32,
-        deprecated = false,
+        ProfileQuery(
+            id = -2,
+            firstName = "Jane",
+            lastName = "Doe",
+            city = "New York",
+            state = "NY",
+            addresses = listOf(
+                Address(
+                    city = "New York",
+                    state = "NY",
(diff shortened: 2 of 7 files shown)


pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
```diff
@@ -173,26 +217,27 @@ class RealPirOptOut @Inject constructor(
...
         }
 
+        // Create a map of profiles that have opt-out steps for those brokers
+        val profileBrokerStepsMap = profileQueries.associateWith { profileQuery ->
+            brokerOptOutStepsJsons.map { (broker, stepsJson) ->
+                brokerStepsParser.parseStep(broker, stepsJson, profileQuery)
+            }.flatten()
+        }.filter { it.value.isNotEmpty() }
+
+        val stepsCount = profileBrokerStepsMap.values.maxOf { it.size }
```
karlenDimla: This logic seems incorrect. The thing that we want to do here is to use the total number of profile x broker x extractedProfile instances IF it’s lesser than the max limit. In this case, we are just taking the highest value.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/common/BrokerStepsParser.kt
```diff
@@ -101,16 +104,22 @@ class RealBrokerStepsParser @Inject constructor(
     override suspend fun parseStep(
         brokerName: String,
         stepsJson: String,
+        profileQuery: ProfileQuery?,
```
karlenDimla: NIT: We can just pass the profileQuery.Id here too.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
```diff
@@ -143,21 +180,28 @@ class RealPirOptOut @Inject constructor(
...
-            }
+        // Create a map of profiles that have opt-out steps for those brokers
+        val profileBrokerStepsMap = profileQueries.associateWith { profileQuery ->
+            brokerOptOutStepsJsons.map { (broker, stepsJson) ->
+                brokerStepsParser.parseStep(broker, stepsJson, profileQuery)
+            }.flatten()
+        }.filter { it.value.isNotEmpty() }
+
+        // Start each runner on the profile with the broker steps
+        profileBrokerStepsMap.entries.map { (profileQuery, brokerSteps) ->
```
karlenDimla: Why entries.map and not forEach?

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
```diff
@@ -208,45 +253,50 @@ class RealPirOptOut @Inject constructor(
...
-        // Start each runner on a subset of the broker steps
-        brokerSteps.splitIntoParts(maxWebViewCount)
-            .mapIndexed { index, part ->
-                async {
-                    runners[index].start(profileQuery, part)
-                    runners[index].stop()
-                }
-            }.awaitAll()
+        // Start each runner on the profile with the broker steps
+        profileBrokerStepsMap.entries.map { (profileQuery, brokerSteps) ->
```
karlenDimla: Same issue here, we are not utilizing the maximum limit of webviews. Consider if we have this map:

```
mapOf{
profileA, 4 extracted profiles
profileB, 1 extracted profiles
profileC, 15 extracted profiles
)
```

It will only use 4 webviews initially and then 1 and then 15. Ideally this could all have run 20 webviews all at the same time.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/scan/PirScan.kt
```diff
@@ -149,23 +186,32 @@ class RealPirScan @Inject constructor(
             createCount++
         }
 
-        // Start each runner on a subset of the broker steps.
-        brokers.mapNotNull { broker ->
+        // Split all the broker steps into parts that can be executed in parallel
+        val brokerScanSteps = brokers.mapNotNull { broker ->
```
karlenDimla, line 187: This works but could’ve be improved if we consider how many profile x broker combinations there would be and then divide that according to the limit.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
landomen: As discussed, there is a limitation to the runners that requires a profile query to be passed in along with the steps, which makes it a bit harder to implement an optimized solution. I will take some time (until Monday) to try to implement a more optimized solution that utilizes all available runners.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/common/PirUtils.kt
```diff
@@ -17,10 +17,23 @@
...
+
+    for (i in 0 until parts) {
+        val currentPartSize = partSize + if (i < remainder) 1 else 0
+        val endIndex = startIndex + currentPartSize
+
+        result.add(this.subList(startIndex, endIndex))
+        startIndex = endIndex
+    }
+
+    return result
```
landomen, line 38: I noticed that the previous implementation would not use all 20 runners in some cases. Example:
- list size = 21
- parts (max web views) = 20
- result = 10 lists of 2 elements + 1 list of 1 element = 11 runners

The new implementation results in 1 list of 2 elements + 19 lists of 1 element = 20 runners

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/common/BrokerStepsParser.kt
landomen: Addressed in https://github.com/duckduckgo/Android/pull/6413/commits/2133953d714739b021fc2568a2f169b2cd1c5a69

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
landomen: Address in https://github.com/duckduckgo/Android/pull/6413/commits/a8a7496f124a3dbf4ad14b16023d186775c3a859

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/scan/PirScan.kt
landomen: Improved in https://github.com/duckduckgo/Android/pull/6413/commits/8e407655e3a22c475a67ed97eb98b15165beced7. We now split work between all 20 runners.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
landomen: Improved in https://github.com/duckduckgo/Android/pull/6413/commits/8e407655e3a22c475a67ed97eb98b15165beced7. We now split work between all 20 runners.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
landomen: Updated in https://github.com/duckduckgo/Android/pull/6413/commits/8e407655e3a22c475a67ed97eb98b15165beced7 to use all 20 runners or less if number of extracted profiles is lower.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/optout/PirOptOut.kt
```diff
@@ -173,25 +217,28 @@ class RealPirOptOut @Inject constructor(
...
 
-        maxWebViewCount = if (brokerSteps.size <= MAX_DETACHED_WEBVIEW_COUNT) {
-            brokerSteps.size
-        } else {
-            MAX_DETACHED_WEBVIEW_COUNT
-        }
+        maxWebViewCount = minOf(allSteps.size, MAX_DETACHED_WEBVIEW_COUNT)
+
+        // Assign steps to runners based on the maximum number of WebViews we can use
+        val stepsPerRunner = allSteps.splitIntoParts(maxWebViewCount)
```
landomen, line 241: We now split the work between the maximum number of runners to max utilization.

landomen:
@karlenDimla I've updated the runner logic to run steps in parallel on maximum number of runners. The PR is ready for another review.

karlenDimla (approved):
Tested and works as expected! Great work!

PR #6421, opened 2025-07-18
use colored backdrop for Input Screen alpha transitions on API 33+

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210820490944337?focus=true

### Description
Uses the `backdrop` API for window transitions to supply a backdrop color which is drawn below and beyond bounds of the activity to prevent the activity window fading to black during alpha transitions.

I remove the translation transition from API < 31 because it didn't run on tested devices anyway.

Additionally, I cleaned up the transition files and moved them to `duckchat` module, which is more appropriate as they are targeted specifically at the Input Screen transition behavior.

### Steps to test this PR

- [x] Verify the slide + fade transitions look right in light and dark mode on API >= 33
- [x] Verify the fade transition looks right in light and dark mode on API < 33 (unchaged behavior)

### UI changes (for API 33+)
| Before  | After |
| ------ | ----- |
before|after|


duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/InputScreenAnimationResourceProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+ * Only one animation component needs backdrop color to prevent system background from showing through,
+ * so we're only applying it to browser animations.
+ *
+ * **Limitations:**
+ * - Backdrop color API is only available from API 33+, so black still shows through on lower APIs,
+ *   but it's less dramatic because there's no slide animation.
+ * - Cannot use themeable attributes (`?attr`) as they cause crashes of the whole launcher (on tested devices).
+ *   Instead, we use fixed color values and filter resources by current theme state.
+ */
+object InputScreenAnimationResourceProvider {
```
joshliebe: Needs API proposal

joshliebe (approved):

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -288,6 +287,7 @@ import com.duckduckgo.downloads.api.DownloadsFileActions
 import com.duckduckgo.downloads.api.FileDownloader
 import com.duckduckgo.downloads.api.FileDownloader.PendingFileDownload
 import com.duckduckgo.duckchat.api.DuckChat
+import com.duckduckgo.duckchat.api.InputScreenAnimationResourceProvider
 import com.duckduckgo.duckchat.impl.inputscreen.ui.InputScreenActivity.Companion.QUERY
 import com.duckduckgo.duckchat.impl.inputscreen.ui.InputScreenActivity.Companion.TAB_ID
```
marcosholgado, line 292: We should not be using things that are not in the public API.

LukasPaczos: The proposal is already waiting for review, if you want to check it out https://app.asana.com/1/137249556945/project/1208671518894266/task/1210844234770562!

marcosholgado: I mean `import com.duckduckgo.duckchat.impl.inputscreen`

LukasPaczos: Oh, I see. This is not part the this PR and we have a TODO to clean it up, refs:
https://github.com/duckduckgo/Android/blob/ba302c82b4de1a1f0c0953c42e0300122ac8ec8d/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/InputScreenActivity.kt#L65

I'll cut a separate task, so that this doesn't slip.

LukasPaczos: cc @joshliebe ^^^

LukasPaczos: Tracking in https://app.asana.com/1/137249556945/task/1210846340363977?focus=true.

PR #6423, opened 2025-07-21
Update destination screen for clear data notification

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202552961248957/task/1210122720065734

### Description
Updated the navigation from "Data clearing notification" to open the "Data Clearing" screen and not the "Settings" screen.

### Steps to test this PR

_Notification navigation_
- [ ] Start with fresh install if you've previously already seen this notification
- [ ] Change "Automatically Clear Data" option to "None"
- [ ] Leave the app
- [ ] Wait for the notification to display and click on it

### UI changes
| Before  | After |
| ------ | ----- |
| <video src="https://github.com/user-attachments/assets/81d5f3a8-21da-4b95-a578-b77dc80b2f4e" /> | <video src="https://github.com/user-attachments/assets/e8c6c294-55d5-440d-a6bf-cf186adae63e" />|

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/firebutton/FireButtonActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/firebutton/FireButtonActivity.kt
@@ -17,6 +17,8 @@
 package com.duckduckgo.app.firebutton
 
 import android.app.ActivityOptions
+import android.content.Context
+import android.content.Intent
 import android.os.Bundle
 import androidx.annotation.StringRes
 import androidx.lifecycle.Lifecycle
@@ -70,6 +72,10 @@ class FireButtonActivity : DuckDuckGoActivity() {
 
         configureUiEventHandlers()
         observeViewModel()
+
+        intent?.getStringExtra(LAUNCH_FROM_NOTIFICATION_PIXEL_NAME)?.let {
+            viewModel.onLaunchedFromNotification(it)
+        }
     }
 
     private fun configureUiEventHandlers() {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/firebutton/FireButtonActivity.kt
```diff
@@ -238,4 +244,12 @@ class FireButtonActivity : DuckDuckGoActivity() {
             )
             .show()
     }
+
+    companion object {
+        const val LAUNCH_FROM_NOTIFICATION_PIXEL_NAME = "LAUNCH_FROM_NOTIFICATION_PIXEL_NAME"
```
anikiki, line 249: Nit: Remove `_PIXEL_NAME` from both const name and value.

landomen: This is an existing constant that I moved from [SettingsActivity](https://github.com/duckduckgo/Android/blob/develop/app/src/main/java/com/duckduckgo/app/settings/SettingsActivity.kt#L438) to here. Same name is also used in [BrowserActivity](https://github.com/duckduckgo/Android/blob/develop/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt#L972) and [SurveyActivity](https://github.com/duckduckgo/Android/blob/develop/app/src/main/java/com/duckduckgo/app/survey/ui/SurveyActivity.kt#L165). I think the name is accurate since this is the name of the `Bundle` key that represents the name of the Pixel that should be fired ([mnot_l](https://github.com/duckduckgo/Android/blob/develop/app/src/main/java/com/duckduckgo/app/pixels/AppPixelName.kt#L168)). Removing the `_PIXEL_NAME` could create more of a confusion as it would seem more like a Boolean field. 

What do you think?

anikiki: Make sense, I didn't pay attention 👍

PR #6435, opened 2025-07-22
compute Input Screen tab indicator height to always be vertically centered in the tab switcher

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210767017822401

### Description
Previously, the dimensions picked for the tab switcher and its tab indicators often resulted in the switcher having a height of an odd number of pixels while the indicator had height of even number of pixels. This resulted in the pill not being perfectly vertically aligned within the switcher. It depends on the pixel density of the device, but happened often on the devices we tested on.

This PR adds a bit of logic to subtract a pixel when needed, to allow for that perfect vertical alignment.

### Steps to test this PR

- [ ] Try a commit before this change and verify that the space below the tab indicator is bigger than the space above it.
- [ ] Try this change and verify that the space is now equal.


duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/view/InputModeTabLayout.kt
```diff
@@ -83,8 +85,12 @@ class InputModeTabLayout @JvmOverloads constructor(
             .setAllCorners(CornerFamily.ROUNDED, cornerRadius)
             .build()
 
+        // computing a pill height that will be perfectly centered in the container vertically, with equal top and bottom insets
+        val shadowPillHeight = listOf(heightPx, heightPx - 1).firstOrNull { candidate ->
+            (containerHeightPx - candidate) % 2 == 0
+        } ?: heightPx
```
joshliebe: I tried a similar approach but issue is that the padding on the round part is now smaller than the top and bottom:

Screenshot_20250722_154344

So may be some sort of rounding error?

LukasPaczos: Yes, definitely a rounding error.

On your screenshot, the bottom and top padding is 6px while the left padding is 5px.

On my device, the padding is 5px all around:
| Top | Bottom | Left |
|---------|---------|---------|
| Screenshot from 2025-07-23 12-07-46 | Screenshot from 2025-07-23 12-07-39 | Screenshot from 2025-07-23 12-07-32 |


This means that my device needs a `-1` pixel correction for the pill height, while your device needs a `+1` pixel correction.

I just pushed another commit that allows correction in both positive and negative direction - could you give it a try on your device?

joshliebe: Screenshot_20250723_134220

Looks correct now, nice one 🙂

PR #6447, opened 2025-07-23
Fix: Swiping animation stuck between tabs

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1210859467773737?focus=true

### Description

This PR adds a delay before disabling the swiping to allow the swiping animation to finish and avoid getting stuck in between tabs.

### Steps to test this PR

- [x] Create at least 20 tabs
- [x] Starting at the end, keep swiping to the beginning (so that the tabs get loaded in memory)
- [x] When you're on the first tab, long-tap on the Tab switcher button to create a NTP
- [x] Verify the current tab is animated to the new tab and it looks correct
- [x] After 1s, try swiping on the focused toolbar
- [x] Verify the swiping is disabled


anikiki (approved):

PR #6450, opened 2025-07-23
Fix: Extract a single URL from an external shared text

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202552961248957/task/1210862419441744?focus=true

### Description

This PR adds support for extracting an URL from an externally shared text if it can find one. Typing some text that contains an URL in the search box treats it like a normal query, as before.

### Steps to test this PR

_External text containing a single URL_
- [x] Share some text that contains a _single_ URL (for example using the Google Now page)
- [x] Open it in DDG
- [x] Verify the URL is correctly extracted and loaded

_Search query_
- [x] Type some text in the search box that contains a valid URL
- [x] Verify a search runs correctly and a SERP is displayed


anikiki (approved):

PR #6454, opened 2025-07-24
PIR: Implement production style scheduling

karlenDimla:
Task/Issue URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1210869571758850?focus=true 

### Description
This is the main PR for the feature. Implementation is divided into the stacked PRs.

### Steps to test this PR
Testing must be done on the last stacked PR.


karlenDimla:
* **#6499** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6489** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6489?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6455** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6455?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6454** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6454?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6454?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/models/scheduling/JobRecord.kt
```diff
@@ -0,0 +1,56 @@
...
+        val extractedProfileId: Long, // Identifier for which ExtractedProfile the state is for
+        val status: OptOutJobStatus = OptOutJobStatus.NOT_EXECUTED,
+        val attemptCount: Int = 0,
+        val lastOptOutAttemptDateInMillis: Long? = null,
+        val optOutRequestedDateInMillis: Long = 0L, // Date when the opt-out for the record has been successfully submitted.
+        val optOutRemovedDateInMillis: Long = 0L, // Date when we confirmed that the record has been successfully removed from the broker.
+    ) : JobRecord(brokerName, userProfileId) {
+        enum class OptOutJobStatus {
+            NOT_EXECUTED,
+            REQUESTED, // The opt out job has successfully sent a request for removal.
```
landomen: Thanks for adding comments, they help understanding what each value represents.

suggestion: we could also convert these inline comments into KDoc comments above the enum value/property so they show up in the documentation and you don't need to navigate to the definition

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/models/scheduling/JobRecord.kt
```diff
@@ -0,0 +1,56 @@
...
+) {
+    data class OptOutJobRecord(
+        override val brokerName: String,
+        override val userProfileId: Long,
+        val extractedProfileId: Long, // Identifier for which ExtractedProfile the state is for
+        val status: OptOutJobStatus = OptOutJobStatus.NOT_EXECUTED,
+        val attemptCount: Int = 0,
+        val lastOptOutAttemptDateInMillis: Long? = null,
+        val optOutRequestedDateInMillis: Long = 0L, // Date when the opt-out for the record has been successfully submitted.
+        val optOutRemovedDateInMillis: Long = 0L, // Date when we confirmed that the record has been successfully removed from the broker.
```
landomen: Considering `0L` represents a missing value, what do you think about making these fields optional instead? We can leave it as is, if that complicates any checks/usages

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirDatabase.kt
```diff
@@ -46,20 +49,22 @@ import com.squareup.moshi.Types
...
+        BrokerSchedulingConfigEntity::class,
         UserProfile::class,
         PirEventLog::class,
         PirBrokerScanLog::class,
         ScanCompletedBroker::class,
         OptOutCompletedBroker::class,
         OptOutActionLog::class,
         StoredExtractedProfile::class,
+        ScanJobRecordEntity::class,
+        OptOutJobRecordEntity::class,
```
landomen, line 67: Question for my understanding: is the general rule that we name DB classes/models as `*Entity`? And the rest of the DB classes/models in PIR are not named that because it was a PoC and naming was not that important?

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirDatabase.kt
```diff
@@ -98,6 +105,24 @@ abstract class PirDatabase : RoomDatabase() {
                 db.execSQL("DROP TABLE IF EXISTS `pir_scan_error`")
             }
         }
+
+        private val MIGRATION_4_TO_5: Migration = object : Migration(4, 5) {
```
landomen, line 109: Considering that we're now actively re-factoring the PoC and there will be more changes to the DB schema, would it maybe be faster/easier to configure destructive migrations? And go back to requiring migrations after the release to preserve user data.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirSchedulingRepository.kt
```diff
@@ -0,0 +1,223 @@
...
+    override suspend fun getAllValidScanJobRecords(): List<ScanJobRecord> = withContext(dispatcherProvider.io()) {
+        return@withContext jobSchedulingDao.getAllScanJobRecords().map {
+            ScanJobRecord(
+                brokerName = it.brokerName,
+                userProfileId = it.userProfileId,
+                status = ScanJobStatus.valueOf(it.status),
+                lastScanDateInMillis = it.lastScanDateInMillis,
+            )
+        }.filter {
+            it.status != ScanJobStatus.INVALID
```
landomen, line 94: nit: doing the filtering in SQL would be (slightly) faster and avoid mapping objects that will be filtered out anyway

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirSchedulingRepository.kt
```diff
@@ -0,0 +1,223 @@
...
+
+    override suspend fun getValidScanJobRecord(
+        brokerName: String,
+        userProfileId: Long,
+    ): ScanJobRecord? = withContext(dispatcherProvider.io()) {
+        return@withContext jobSchedulingDao.getScanJobRecord(brokerName, userProfileId)?.run {
+            ScanJobRecord(
+                brokerName = this.brokerName,
+                userProfileId = this.userProfileId,
+                status = ScanJobStatus.valueOf(this.status),
```
landomen: nit: this could theoretically result in a not-found exception if we remove an existing enum value and there are still records with that value stored. 

We could change it to something like: 
`ScanJobStatus.entries.find { it.name == this.status } ?: ScanJobStatus.INVALID` 
or create a dedicated `ScanJobStatusTypeConverter` that handles that.

pir/pir-internal/src/test/kotlin/com/duckduckgo/pir/internal/store/RealPirSchedulingRepositoryTest.kt
```diff
@@ -0,0 +1,417 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+
+class RealPirSchedulingRepositoryTest {
```
landomen, line 37: 👍  for adding tests

landomen (approved):
Great job! 👍 Everything looks good to me, the comments I left are more of questions for my understanding of the code style/approaches we take at DDG.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirDatabase.kt
karlenDimla: There’s not much guidelines here really. But as you can see I am slowly starting to separate data layer models and business logic models since that was not really paid much attention too during the PoC. We can call them any way we want but it has been a common pattern in the repo, but not required to append them with Entity.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirDatabase.kt
karlenDimla: I did this before we decided to finally stop the internal test. I was not sure how these changes will affect those users and I don’t want to risk the DDG app crashing / emitting crash pixels for those users specially with the scheduled scans running.

landomen: That makes total sense. I was just thinking if we could make our lives easier while we're in re-factoring mode :)

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/models/scheduling/JobRecord.kt
karlenDimla: I think it’s better to have Long: 0L for all these values so the only check I need to do is if it’s != 0L. We don’t have to bother checking for nulls and optionals.

PR #6521, opened 2025-08-04
remove nested material cards in tab switcher adapter

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1210935007420667?focus=true

### Description
Removes nested `MaterialCardView` that was used only to apply rounded corners to the image. Now, we're using glide transformation instead.

### Steps to test this PR

Verify that tab switcher previews look exactly the same before and after the change.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
@@ -36,6 +36,7 @@ import com.bumptech.glide.Glide
 import com.bumptech.glide.RequestManager
 import com.bumptech.glide.load.Transformation
 import com.bumptech.glide.load.engine.Resource
+import com.bumptech.glide.load.resource.bitmap.RoundedCorners
 import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherAdapter.kt
```diff
@@ -398,10 +401,18 @@ class TabSwitcherAdapter(
...
-                    .into(tabPreview)
+                try {
+                    glide.load(cachedWebViewPreview)
+                        .transition(DrawableTransitionOptions.withCrossFade())
+                        .transform(
+                            fitAndClipBottom(),
+                            RoundedCorners(tabPreview.context.resources.getDimensionPixelSize(CommonR.dimen.smallShapeCornerRadius)),
+                        )
+                        .into(tabPreview)
+                } catch (e: Exception) {
```
LukasPaczos, line 411: Previously, the clip transformation was added as "optional" which means that it would never throw. As there's no `optionalTransform` overload that would take multiple transformations, I'm adding the catch as a precaution. I don't know if the "no throw" was needed for any particular case, hopefully @0nko can confirm later.

PR #6546, opened 2025-08-07
PIR: Fix bug on storing duplicated extracted profiles

karlenDimla:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1210984555734427?focus=true

### Description
See attached task description

### Steps to test this PR
https://app.asana.com/1/137249556945/project/1203581873609357/task/1210992048211447?focus=true


karlenDimla:
* **#6546** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6546?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6546?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/common/PirRunStateHandler.kt
```diff
@@ -242,19 +242,19 @@ class RealPirRunStateHandler @Inject constructor(
         when (state.pirSuccessResponse) {
             is ExtractedResponse -> state.pirSuccessResponse.response.map {
                 ExtractedProfile(
-                    profileUrl = it.profileUrl,
+                    profileUrl = it.profileUrl ?: "",
```
landomen: tip: could also use `.orEmpty()`

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/scheduling/JobRecordUpdater.kt
```diff
@@ -287,17 +287,17 @@ class RealJobRecordUpdater @Inject constructor(
     private data class ExtractedProfileComparisonKey(
         val profileQueryId: Long,
         val brokerName: String,
-        val name: String?,
+        val name: String,
```
landomen, line 290: What is the main benefit of having this stored as empty string vs null?

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirRepository.kt
```diff
@@ -233,7 +233,7 @@ internal class RealPirRepository(
         return@withContext brokerDao.getAllBrokersNamesWithScanSteps()
     }
 
-    override suspend fun getEtagForFilename(fileName: String): String? = withContext(dispatcherProvider.io()) {
+    override suspend fun getEtagForFilename(fileName: String): String = withContext(dispatcherProvider.io()) {
```
landomen, line 246: This function that appears to be unused. Do we need it or can it be removed?

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/brokers/BrokerJsonUpdater.kt
```diff
@@ -79,6 +80,17 @@ class RealBrokerJsonUpdater @Inject constructor(
         }
     }
 
+    private suspend fun confirmEtagIntegrity() {
+        val storedEtag = pirRepository.getCurrentMainEtag()
+        if (storedEtag != null) {
+            val storedJsons = pirRepository.getAllLocalBrokerJsons()
```
landomen: nit: this could be optimized by having a separate SQL query that checks the number of rows instead of loading the whole list of brokers that is then not used

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/brokers/BrokerJsonUpdater.kt
```diff
@@ -79,6 +80,17 @@ class RealBrokerJsonUpdater @Inject constructor(
         }
     }
 
+    private suspend fun confirmEtagIntegrity() {
```
landomen, line 83: In what scenario can it happen that we would have an Etag stored but not brokers?

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/db/ScanResultsDao.kt
```diff
@@ -48,12 +49,15 @@ interface ScanResultsDao {
...
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
     fun insertExtractedProfile(extractedProfile: StoredExtractedProfile)
 
-    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
     fun insertExtractedProfiles(extractedProfiles: List<StoredExtractedProfile>)
 
+    @Update
+    fun updateExtractedProfiles(profiles: List<StoredExtractedProfile>): Int
+
```
landomen: Considering we have unique indexes now on `StoredExtractedProfile`, do we still need the two step approach or would `REPLACE` work? Just asking since we would save an additional query

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirRepository.kt
```diff
@@ -324,7 +324,11 @@ internal class RealPirRepository(
             extractedProfiles.map {
                 it.toStoredExtractedProfile()
             }.also {
-                scanResultsDao.insertExtractedProfiles(it)
+                val updated = scanResultsDao.updateExtractedProfiles(it)
+                if (updated == 0) {
+                    // If no profiles were updated, insert them
+                    scanResultsDao.insertExtractedProfiles(it)
+                }
```
landomen: It might be good to run this in a single transaction to avoid data loss / weird state

landomen (approved):
Looks good to me, left some questions more for my understanding and some suggestions/comments that are not blocking

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/scheduling/JobRecordUpdater.kt
karlenDimla: @landomen usually, it’s to avoid having to check for both - null or empty. That means, we only have to check for empty every time.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirRepository.kt
karlenDimla: We can remove it later. I am not sure if we will need it for the remove defunct brokers.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/brokers/BrokerJsonUpdater.kt
karlenDimla: Ah yeah. I can add that. 👍

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/brokers/BrokerJsonUpdater.kt
karlenDimla: When the db falls back to destructive migration. I ran into this issue and I am unable to force the download of the brokers anymore. The etag is stored in a DataStore.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirRepository.kt
karlenDimla: what do you mean?

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/scheduling/JobRecordUpdater.kt
karlenDimla: In this case, I had to change the corresponding StoredExtractedProfile attributes to nonnull because indices doesn’t seem to handle well null values. So for consistency sake, i removed all nullable fields.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirRepository.kt
landomen: Sorry for not providing more context in the original comment. What I meant is that there are two queries here: 1. to update existing profiles and 2. to store new profiles.

It could happen that 1. would succeed but 2. would fail, meaning that some profiles would be updated but no new ones added as these are separate queries. 

If this was run in a [Transaction](https://github.com/duckduckgo/Android/blob/develop/app-tracking-protection/vpn-store/src/main/java/com/duckduckgo/mobile/android/vpn/dao/VpnAppTrackerBlockingDao.kt#L103-L109) and 2. would fail, then 1. would be reverted and no profiles would be updated, maintaining data consistency.

But writing this out now, we might even prefer the non-Transactional behavior 😅 Sorry for the confusion

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/scheduling/JobRecordUpdater.kt
landomen: That makes sense, thanks for the explanation

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/db/ScanResultsDao.kt
karlenDimla: That’s what I expected to happen to be honest but it doesn’t work. 
Note: StoredExtractedProfile that we are trying to update doesn’t have the the db id since they are coming the script. But they should be able to match the indices.

The way REPLACE worked, as I’ve noticed, is that:
1. When the indices are matached, they removed the previous row.
2. A new row is inserted with a new ID.

That is a problem for us since we use the id as reference for OptOutJobrecords. Hence the way that I thought we could fix this is to attempt to update first which updates all available. And then attempt insert after if nothing was updated.

Having that said, I just realized that since we are inserting a list there might be an issue here. Let me think about it.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/PirRepository.kt
karlenDimla: Right. I’ve updated this logic now in the most recent commit. We don’t need the 2 steps anymore.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/db/ScanResultsDao.kt
karlenDimla: I’ve updated this logic now. I realized we don’t need to update extractedProfiles when they are already stored in our database. So updating the startegy to IGNORE works.
1. New ExtractedProfiles are stored normally.
2. When we do a new scan for the broker, existing ExtractedProfiles don’t need to be updated. We only need to store new ones OR update the record for removed ones. ExtractedProfile today holds the data only and if it’s deprecated. It’s removal status is in the record.

We probably would want to introduce `deprecateExtractedProfile` later that will update the extracted profile. I didn’t add that yet. Let me know what you think.

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/brokers/BrokerJsonUpdater.kt
```diff
@@ -83,8 +83,8 @@ class RealBrokerJsonUpdater @Inject constructor(
     private suspend fun confirmEtagIntegrity() {
         val storedEtag = pirRepository.getCurrentMainEtag()
         if (storedEtag != null) {
-            val storedJsons = pirRepository.getAllLocalBrokerJsons()
-            if (storedJsons.isEmpty()) {
+            val storedJsons = pirRepository.getStoredBrokersCount()
```
landomen: nit: could be renamed to `storedBrokersCount`

pir/pir-internal/src/main/java/com/duckduckgo/pir/internal/store/db/ScanResultsDao.kt
landomen: That makes sense!

Issue #6516, opened 2025-08-02
[Bug] Can't download ebooks from my local Calibre server


wizduin:
### Describe the bug

I can't download my ebooks from my local Calibre server. When I try to download, the app keeps saying the operation was unsuccessful and that I should check my internet connection. However, the Chrome browser has no problem downloading.

### How to Reproduce

1. Go to your Calibre local server IP (http://LOCAL_ADDRESS:PORT)
2. Select one of your ebooks.
3. Click the download button.

### Expected behavior

Download the ebook into the default download folder.

### Environment

```markdown
- DDG App Version: 5.242.0
- Device: OnePlus Nord CE 2 Lite 5G
- OS: OxygenOS 14.0 (Android 14)
```

CDRussell:
Thanks for reporting! I have created a PR which should address this: https://github.com/duckduckgo/Android/pull/6564

PR #6564, opened 2025-08-12
Omit Cookie downloads header when empty rather than empty string

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211033344698349?focus=true 

### Description
From https://github.com/duckduckgo/Android/issues/6516, investigated and found that empty `Cookie` HTTP headers are tripping up Calibre's web server. 
- Empty `Cookie` headers are legal but there are some sparse reports of web servers failing to parse them
- Rather than an empty `Cookie` header, we can just omit sending it at all

That fixes the problem for Calibre (and i'm sure a few other services). Fixes https://github.com/duckduckgo/Android/issues/6516.

Feature flagged so we can revert to the existing behaviour if we notice problems because of this change.

### Steps to test this PR

- [x] Test some downloads still succeed as normal (e.g., https://file-examples.com/)
- [ ] [Optional]: 
    - [ ] [install Calibre](https://calibre-ebook.com/download)
    - [ ] Tap the `Connect/share` button
    - [ ] `Start content server`
    - [ ] Browse to the URL from the Android device running this fixed branch, and download the sample book

### DIFF
--- a/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DownloadFileService.kt
+++ b/downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DownloadFileService.kt
@@ -35,7 +35,7 @@ interface DownloadFileService {
     @Streaming
     @GET
     fun downloadFile(
-        @Header("Cookie") cookie: String,
+        @Header("Cookie") cookie: String?,
         @Url urlString: String,
     ): Call<ResponseBody>
 }
--- a/downloads/downloads-impl/build.gradle
+++ b/downloads/downloads-impl/build.gradle
@@ -86,6 +86,7 @@ dependencies {
     testImplementation 'app.cash.turbine:turbine:_'
     testImplementation Testing.robolectric
     testImplementation project(path: ':common-test')
+    testImplementation project(':feature-toggles-test')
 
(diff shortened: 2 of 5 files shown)


CDRussell:
* **#6564** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6564?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6564?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

downloads/downloads-impl/src/main/java/com/duckduckgo/downloads/impl/DownloadFileService.kt
```diff
@@ -35,7 +35,7 @@ interface DownloadFileService {
     @Streaming
     @GET
     fun downloadFile(
-        @Header("Cookie") cookie: String,
+        @Header("Cookie") cookie: String?,
```
CDRussell, line 38: If null, header is omitted

karlenDimla (approved):
Did some of the testing in file-examples and worked as expected. Changes make sense to me.

PR #6604, opened 2025-08-18
automatically launch the Input Screen when New Tab Page opened

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1211056807546917?focus=true

### Description
With this PR, whenever a New Tab Page is opened, the Input Screen will be launched automatically.

Implementing this had some challenges given how complex and fragile the keyboard management implementation is.

I couldn't rely on the `BrowserTabFragment#onViewVisible`, which is currently used to auto-focus on the omnibar, because that's getting called whenever the fragment is resumed. So, if the logic to open the Input Screen was part of this function, closing the Input Screen (which is a separate activity) would resume the fragment and open the Input Screen again, resulting in a loop.

Instead, I added the logic to observe whenever the tab changes and dispatch a command to launch the Input Screen when user switches to a New Tab Page (this tab has no URL). This logic lives as part of each `BrowserTabFragment` so that the current active fragment, if it's a new tab page, can launch the Input Screen and then listen for its results.

### Steps to test this PR

- [x] Install a clean build of the app.
- [x] Verify that opening a new tab page focuses on the omnibar automatically.
- [x] Open a web page or SERP and verify that switching to a page doesn't focus on the omnibar automatically.
- [x] Go to Settings -> AI Features and enable the Experimental Address bar.
- [x] Go back to the browser and verify that creating a New Tab opens the Input Screen automatically.
- [x] Back out from the Input Screen and verify you can see the New Tab Page.
- [x] Go to another tab with a web page or SERP and verify that the Input Screen doesn't open automatically.
- [x] Go back to the existing New Tab card in the tab switcher and verify that the Input Screen Opens automatically.
- [x] Close the app.
- [x] Reopen the app.
- [x] Verify that the Input Screen launches automatically (because you're still on a New Tab Page).
- [x] Use the fire button and verify that the Input Screen automatically opens only after the fire animation finishes and the process restarts.

### UI changes

https://github.com/user-attachments/assets/11bf872d-bc09-46b4-9ecf-77a4a843a064

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -375,6 +375,7 @@ import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.distinctUntilChangedBy
 import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.flatMapLatest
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -727,6 +731,34 @@ class BrowserTabViewModel @Inject constructor(
...
+                        .distinctUntilChanged() // only observe when the tab changes and ignore further updates
+                        .filter { selectedTabId ->
+                            // if the tab managed by this view model has just been activated, and it's a new tab (it has no URL), then fire an event
+                            val isActiveTab = selectedTabId == tabId
+                            if (isActiveTab) {
+                                val thisTab = tabRepository.getTab(tabId)
+                                thisTab != null && thisTab.url.isNullOrBlank()
+                            } else {
+                                false
+                            }
```
landomen: nit: could be simplified as `isActiveTab && tabRepository.getTab(tabId)?.url.isNullOrBlank()`

landomen (approved):
Looks good to me. I've run through the QA steps on Samsung S10 (Android 12) and S21 Ultra (Android 15) and worked as expected.

LukasPaczos:
Thanks for a review @landomen! Just a heads up that I found a small optimization in https://github.com/duckduckgo/Android/pull/6604/commits/7b30a257747b7b2f0192725e2a6973dd2f96f110 to avoid unnecessary repository query. I re-run the manual tests and everything works as expected, but let me know if you see anything off.

PR #6610, opened 2025-08-19
PIR: Add edit and remove name and addresses messages

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1211075568938198?focus=true

### Description
Adds supports for remaining name and address editing messages during onboarding.

### Steps to test this PR
https://app.asana.com/1/137249556945/project/1203581873609357/task/1211088339515453?focus=true

### UI changes
No UI changes

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/messaging/handlers/PirWebAddAddressToCurrentUserProfileMessageHandler.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/messaging/handlers/PirWebAddAddressToCurrentUserProfileMessageHandler.kt
@@ -51,8 +51,11 @@ class PirWebAddAddressToCurrentUserProfileMessageHandler @Inject constructor(
         val request =
             jsMessage.toRequestMessage(PirWebMessageRequest.AddAddressToCurrentUserProfileRequest::class)
 
+        val city = request?.city?.trim().orEmpty()
+        val state = request?.state?.trim().orEmpty()
+
         // attempting to add an empty address should return success=false
-        if (request == null || request.city.isEmpty() || request.state.isEmpty()) {
+        if (city.isBlank() || state.isBlank()) {
             logcat { "PIR-WEB: PirWebAddAddressToCurrentUserProfileMessageHandler: missing city and/or state" }
             jsMessaging.sendResponse(
                 jsMessage = jsMessage,
@@ -62,7 +65,7 @@ class PirWebAddAddressToCurrentUserProfileMessageHandler @Inject constructor(
         }
 
         // attempting to add a duplicate address should return success=false
-        if (!pirWebOnboardingStateHolder.addAddress(request.city, request.state)) {
+        if (!pirWebOnboardingStateHolder.addAddress(city, state)) {
             logcat { "PIR-WEB: PirWebAddAddressToCurrentUserProfileMessageHandler: address already exists" }
             jsMessaging.sendResponse(
                 jsMessage = jsMessage,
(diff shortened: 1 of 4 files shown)


landomen:
* **#6628** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6628?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6614** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6614?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6610** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6610?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirWebOnboardingStateHolder.kt
```diff
@@ -17,25 +17,126 @@
...
+
+    fun removeNameAtIndex(index: Int): Boolean {
+        if (index in names.indices) {
+            names.removeAt(index)
+            return true
+        }
+        return false
+    }
+
+    fun toUserProfiles(): List<UserProfile> {
```
karlenDimla, line 127: NIT: I think it is better to set this to `ProfileQuery` and then use those to save in the respository. UserProfile is data layer object so preferably it shouldn’t be used here.

Probably we could do something to avoid these confusion. YOu don’t have to fix this in this PR but we can fix it at some other point /PR.

landomen: Good point. I will refactor this to use `ProfileQuery` in a separate PR. I opened a task for this https://app.asana.com/1/137249556945/project/488551667048375/task/1211102791780024?focus=true

I'll also update the `replaceUserProfile(userProfile: UserProfile)` function in the `PirRepository` that's called from the dev screen.

I'll make sure to double check to use models from `pir.impl.models` and not `pir.impl.store.db` outside the repository. It gets a bit confusing because both are accessible since they're in the same module 😅

karlenDimla: We used to have `-store` modules but that is not the pattern nowadays.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirWebOnboardingStateHolder.kt
```diff
@@ -17,25 +17,126 @@
...
+        val profiles = mutableListOf<UserProfile>()
```
karlenDimla, line 128: We need to make sure we trim them before they are added to the DB.

landomen: Do you mean trim the names and addresses? That's done in each message handler before a name or address is added to the state here ([example](https://github.com/duckduckgo/Android/pull/6610/files/928295c2c71e7cd9874971f1f435e9575eb0f02d#diff-975ad78d78e6c46aef8cdff7ffe08daa244da031255d1b21ee9474b6c2c58c94R54)). So at this point all the data is already trimmed and valid

karlenDimla (approved):
Looks good. We can fix the case issue on a separate PR.

PR #6615, opened 2025-08-20
do not open the Input Screen when user pref toggled on and immediately off

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1211091584703241?focus=true

### Description
Fixes an issue where toggling the experimental address bar on and then immediately off again in "AI Features" settings still caused the Input Screen to open when returning to the browser (if left on a New Tab Page).

#### Root Cause
The Input Screen preferences flow in the `BrowserTabViewModel` (and its nested flow that launches the Input Screen on NTP) was being restarted every time the preference changed. As a result, toggling the setting caused a re-evaluation, which queued a command to launch the Input Screen when returning to the browser.

#### Fix
Instead of observing the Input Screen preference flow directly, we now filter out events when the preference is disabled. This prevents unnecessary re-evaluations on every toggle, while still ensuring that whenever the tab changes, the current preference state is correctly respected.

### Steps to test this PR

- [ ] Install a clean build of the app.
- [ ] Open a new tab and ensure that the address bar gets focused but the Input Screen doesn't open.
- [ ] Go to Settings -> AI Features
- [ ] Enabled Experimental Address Bar
- [ ] Disabled Experimental Address Bar
- [ ] Go back to the browser and verify that Input Screen is not opened
- [ ] (optional) Try some additional test steps from https://github.com/duckduckgo/Android/pull/6604 to ensure that the feature still works as expected otherwise

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -379,11 +379,11 @@ import kotlinx.coroutines.flow.distinctUntilChangedBy
 import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.flatMapLatest
-import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.takeWhile
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import logcat.LogPriority.ERROR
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -733,22 +733,16 @@ class BrowserTabViewModel @Inject constructor(
...
-                        }
-                } else {
-                    flowOf()
-                }
+        // observe when user opens a new tab and launch the input screen, if the feature is enabled
+        tabRepository.flowSelectedTab
+            .distinctUntilChangedBy { selectedTab -> selectedTab?.tabId } // only observe when the tab changes and ignore further updates
+            .filter { selectedTab ->
+                // if the tab managed by this view model has just been activated, and it's a new tab (it has no URL), then fire an event
+                val isActiveTab = ::tabId.isInitialized && selectedTab?.tabId == tabId
```
LukasPaczos, line 740: We don't really need `::tabId.isInitialized` check for the actual production behavior because we're not using the immediate dispatcher, however, when run as part of the instrumented unit tests where all dispatchers are made immediate anyway, we see all tests failing because the value is not initialized when the class is instantiated.

The right solution would be to change the view model factory to require the `tabId` as a constructor parameter, instead of passing it in the `loadData` function right after creation, but I didn't want to make that refactor as part of this PR.

LukasPaczos:
I added one more commit to update one of the E2E tests to also verify that Input Screen opens automatically when new tab is opened.

To run it locally (optional), you can do:
```
./gradlew installInternalRelease
```
and then
```
maestro test .maestro/input_screen/input_screen_preference_management.yaml
```

PR #6621, opened 2025-08-20
Jkt/client hint gate

jonathanKingston:
Task/Issue URL: 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


user-agent/user-agent-impl/src/main/java/com/duckduckgo/user/agent/impl/ClientBrandHintProvider.kt
```diff
@@ -50,14 +50,12 @@ class RealClientBrandHintProvider @Inject constructor(
     private var currentBranding: ClientBrandsHints = DDG
 
     override fun setDefault(settings: WebSettings) {
-        if (WebViewFeature.isFeatureSupported(WebViewFeature.USER_AGENT_METADATA)) {
```
jonathanKingston, line 53: Just moving this into `setUserAgentMetadata`. All the rest of this if statement looks safe to call. IE `clientBrandHintFeature.self().isEnabled()` isn't a cause for concern.

PR #6614, opened 2025-08-20
PIR: Add profile saving validation

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211075568938200?focus=true

### Description
Validates that all profiles are saved on ´saveProfile´ message.

### Steps to test this PR
https://app.asana.com/1/137249556945/task/1211099146322592?focus=true

### UI changes
No UI changes

PR #6628, opened 2025-08-21
PIR: Use ProfileQuery on onboarding

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211102791780024?focus=true

### Description
Uses ´ProfileQuery´ objects instead of the ´UserProfile´ database model for onboarding and dev settings.

### Steps to test this PR
https://app.asana.com/1/137249556945/task/1211102791780027?focus=true

### UI changes
No UI changes

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirWebOnboardingStateHolder.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirWebOnboardingStateHolder.kt
@@ -18,7 +18,6 @@ package com.duckduckgo.pir.impl.dashboard.state
 
 import com.duckduckgo.di.scopes.ActivityScope
 import com.duckduckgo.pir.impl.models.Address
-import com.duckduckgo.pir.impl.models.Name
 import com.duckduckgo.pir.impl.models.ProfileQuery
 import dagger.SingleInstanceIn
 import javax.inject.Inject
@@ -28,10 +27,10 @@ class PirWebOnboardingStateHolder @Inject constructor() {
 
     private val names: MutableList<Name> = mutableListOf()
     private val addresses: MutableList<Address> = mutableListOf()
-    private var birthYear: Int? = null
+    private var birthYear: Int = 0
 
     val isProfileComplete: Boolean
-        get() = names.isNotEmpty() && addresses.isNotEmpty() && (birthYear ?: 0) > 0
+        get() = names.isNotEmpty() && addresses.isNotEmpty() && birthYear > 0
 
     fun addAddress(
         city: String,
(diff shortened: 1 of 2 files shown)


landomen:
* **#6628** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6628?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6628?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6614** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6614?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6610** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirWebOnboardingStateHolder.kt
```diff
@@ -124,15 +124,25 @@ class PirWebOnboardingStateHolder @Inject constructor() {
...
+                        middleName = name.middleName,
+                        city = address.city,
+                        state = address.state,
+                        addresses = listOf(address),
                         birthYear = birthYear ?: 0,
-                        addresses = address,
+                        fullName = name.middleName?.let { middleName ->
+                            "${name.firstName} $middleName ${name.lastName}"
+                        } ?: "${name.firstName} ${name.lastName}",
+                        age = currentYear - (birthYear ?: 0),
```
karlenDimla: When can the birthyear be null?

landomen: Just at the beginning, before the user enters it. When we're mapping to profile queries, it should not be null in any case. 

I can also make it non-null and use 0 to indicate that it's not been set if you think it's cleaner? Or if you have any other ideas.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/models/ProfileQuery.kt
```diff
@@ -38,3 +38,9 @@ data class Address(
     val city: String,
     val state: String,
 )
+
```
karlenDimla: What is this for? Seems like a model for the onboarding state only.

landomen: We didn't have a public model for the name. I placed it here as we're already using the Address model from here. But considering it's only used in onboarding, I can move it there

landomen: I moved it to the onboarding state holder since ProfileQuery doesn't use it

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirWebOnboardingStateHolder.kt
landomen: Made it non-null to avoid null checks

PR #6653, opened 2025-08-25
fix an issue where links opening in new tab would also open the Input Screen

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/715106103902962/task/1211121759515670?focus=true

### Description
Fixes an issue where links that open in a new tab would also open the Input Screen automatically.

### Steps to test this PR

- [x] Go to a page that has links which open in a new tab, for example, go to [w3schools links section](https://www.w3schools.com/html//html_links.asp) and use the "Try It Yourself" button. This should open the content in a new tab but the Input Screen should not be opened.
- [x] Create a new tab manually and verify that Input Screen opens automatically.
- [x] Go to Settings -> Feature Flag Inventory -> Search for "showInputScreenAutomatically" and disable the feature.
- [x] Force close and restart the app.
- [x] Verify that creating a new tab doesn't automatically open the Input Screen.


landomen (approved):
Works as expected 👍

PR #6694, opened 2025-08-29
Jetpack Compose - Add stock dependencies

mikescamell:
Task/Issue URL: 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


versions.properties
```diff
@@ -31,6 +33,11 @@ version.androidx.arch.core=2.2.0
 
 version.androidx.browser=1.8.0
 
+version.androidx.compose=2025.08.00
+
+# Cannot update past 1.5.14 because we need Kotlin 2.x
+version.androidx.compose.compiler=1.5.14
```
landomen, line 39: We could use version `1.5.15` which targets Kotlin `1.9.25` and is the last version before you need Kotlin 2.x. However, that also means updating Kotlin and KSP to `1.9.25`, which I've tried and it works fine, but I understand if we don't want to make those changes on top of adding Compose.

versions.properties
```diff
@@ -31,6 +33,11 @@ version.androidx.arch.core=2.2.0
 
 version.androidx.browser=1.8.0
 
+version.androidx.compose=2025.08.00
```
landomen: `2025.08.01` version appears to be the latest (https://developer.android.com/develop/ui/compose/bom/bom-mapping)

versions.properties
```diff
@@ -113,6 +125,9 @@ version.jakewharton.rxrelay2=2.0.0
 
 version.kotlin=1.9.24
 
+# Cannot update to 0.4.0 because it requires Kotlin 2.1.20
```
landomen, line 128: Seems like we will need to update to Kotlin 2.x sooner rather than later to get the benefits of latest versions of libraries

versions.properties
mikescamell: I'd rather not update Kotlin and KSP as part of this but thanks for pointing it out 🙌

versions.properties
mikescamell: Updated to `2025.09.00`

PR #6733, opened 2025-09-04
Load sqlcipher library asynchronously 

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207908166761516/task/1210918149098144

### Description

Implemented `LibraryLoaderListener` interface in `RealSecureStorageDatabaseFactory` to properly handle SQLCipher library loading success and failure callbacks. This change improves error handling by:

1. Adding explicit success and failure callback methods
2. Wrapping database retrieval in a try-catch block to handle exceptions
3. Properly logging errors during library loading and database access

### Steps to test this PR

_SQLCipher Library Loading_

- [ ] Verify that SQLCipher library loads successfully with proper logging
- [ ] Confirm that database access works correctly when the feature is enabled

### UI changes

| Before | After |
| --- | --- |
| N/A | N/A |

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/securestorage/SecureStorageDatabaseFactory.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/securestorage/SecureStorageDatabaseFactory.kt
@@ -47,7 +47,7 @@ class RealSecureStorageDatabaseFactory @Inject constructor(
     private val context: Context,
     private val keyProvider: SecureStorageKeyProvider,
     private val autofillFeature: AutofillFeature,
-) : SecureStorageDatabaseFactory {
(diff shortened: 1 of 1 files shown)


CDRussell:
@mikescamell I'm not convinced by these changes as is, because it feels risky that it could introduce breakage. The main concerns: 

- if the library loading isn't finished by the time we first attempt to call `getDatabase()` then we'll fail to initialise the databases. It shouldn't crash, but will conclude that autofill can't be used on this device and essentially turn it all off.
- if the above happened, we'd have no immediate way of knowing
- the changes aren't guarded by a feature flag and therefore any issues we encounter would require a new app release / hotfix to address

Will follow up with you some more on the linked task.

CDRussell (changes requested):
Added suggestions to the linked task

CDRussell:
Going to close this in favor of a few other PRs to achieve the same aim

PR #6746, opened 2025-09-05
Streamline importing passwords from Google

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/608920331025315/task/1211197010246726?focus=true 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


CDRussell:
* **#6746** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6746?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6746?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/AutofillJavascriptInterface.kt
```diff
@@ -262,7 +262,23 @@ class AutofillStoredBackJavascriptInterface @Inject constructor(
...
+
+    private suspend fun notifyListenerThatCredentialsAvailableToInject(
+        url: String,
+        finalCredentialList: List<LoginCredentials>,
+        triggerType: LoginTriggerType,
+        request: AutofillDataRequest,
+    ) {
+        when (val currentCallback = callback) {
+            is InternalCallback -> {
+                currentCallback.onCredentialsAvailableToInjectWithReauth(url, finalCredentialList, triggerType, request.subType)
```
CDRussell, line 277: don't want the extra parameters to pollute the public API as they aren't needed from other scenarios so this allows for that

package.json
```diff
@@ -14,8 +14,8 @@
   },
   "dependencies": {
     "@duckduckgo/autoconsent": "^14.19.0",
-    "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#18.3.0",
-    "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#11.19.0",
+    "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#dbajpeyi/log-forced-items",
+    "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#pr-releases/pr-1933",
```
CDRussell: These will get a stable release before we merge

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/configuration/AutofillAvailableInputTypesProvider.kt
```diff
@@ -51,16 +55,26 @@ class RealAutofillAvailableInputTypesProvider @Inject constructor(
     private val dispatchers: DispatcherProvider,
 ) : AutofillAvailableInputTypesProvider {
 
-    override suspend fun getTypes(url: String?): AvailableInputTypes {
+    override suspend fun getTypes(
```
cmonfortep, line 59: We should add some unit tests into `RealAutofillAvailableInputTypesProviderTest` to cover the new logic.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/configuration/AutofillAvailableInputTypesProvider.kt
```diff
@@ -24,14 +24,18 @@ import com.duckduckgo.autofill.impl.importing.InBrowserImportPromo
...
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesBinding
 import javax.inject.Inject
 import kotlinx.coroutines.withContext
 
 interface AutofillAvailableInputTypesProvider {
-    suspend fun getTypes(url: String?): AvailableInputTypes
+    suspend fun getTypes(
+        url: String?,
+        reAuthenticationDetails: ReAuthenticationDetails = ReAuthenticationDetails(),
```
cmonfortep: Mixed feelings having this default value in the interface.
If you think we should keep it, to improve readability I would declare a method:
fun EmptyAuthenticationDetails() = ReAuthenticationDetails() in the same file you have the data class, and use that method instead. That way is more clear: reAuthenticationDetails: ReAuthenticationDetails = EmptyAuthenticationDetails().
Without context, I see ReAuthenticationDetails and I don't know if it will have valid value inside or not by default.
Another option is to remove it, you don't have it in other interfaces (`AutofillRuntimeConfigProvider`)
Thoughts?

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/gpm/webflow/ImportGooglePasswordsWebFlowViewModelTest.kt
```diff
@@ -103,6 +119,239 @@ class ImportGooglePasswordsWebFlowViewModelTest {
...
     }
 
+    @Test
+    fun whenStoredCredentialsAvailableWithReauthAllowedAndPasswordThenInjectCredentialsFromReauth() = runTest {
+        val url = "https://example.com"
+        val password = "test-password"
+        val credentials = listOf(creds())
+        val mockToggle: Toggle = mock()
+        whenever(mockToggle.isEnabled()).thenReturn(true)
+        whenever(autofillFeature.canReAuthenticateGoogleLoginsAutomatically()).thenReturn(mockToggle)
```
cmonfortep: I think you have a method for this declared `configureFeatureFlagEnabled` (same for the method below)

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/gpm/webflow/ImportGooglePasswordsWebFlowViewModelTest.kt
```diff
@@ -103,6 +119,239 @@ class ImportGooglePasswordsWebFlowViewModelTest {
...
+        whenever(reauthenticationHandler.retrieveReauthData(url)).thenReturn(ReAuthenticationDetails(password = null))
+
+        testee.commands.test {
+            testee.onStoredCredentialsAvailable(url, credentials, USER_INITIATED, scenarioAllowsReAuthentication = true)
+            assertTrue(awaitItem() is PromptUserToSelectFromStoredCredentials)
+        }
+    }
+
+    @Test
+    fun whenStoredCredentialsAvailableWithReauthNotAllowedThenPromptUser() = runTest {
```
cmonfortep, line 150: Is `NotAllowed` same as `whenReauthFeatureFlagDisabled` (test below)?

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/gpm/webflow/ImportGooglePasswordsWebFlowViewModelTest.kt
```diff
@@ -103,6 +119,239 @@ class ImportGooglePasswordsWebFlowViewModelTest {
...
+        configureReauthData(url, password) // Even with reauth data, should return NoCredentialsAvailable when flag disabled
+
+        testee.commands.test {
+            testee.onNoStoredCredentialsAvailable(url)
+            assertTrue(awaitItem() is NoCredentialsAvailable)
+        }
+    }
+
+    @Test
+    fun whenReauthFeatureFlagEnabledButReauthDataNullThenPromptUser() = runTest {
```
cmonfortep: this looks similar to `whenReauthFeatureFlagEnabledButReauthDataHasNullPasswordThenPromptUser`. Reauth can't be null, only the password inside, so data refers to Password, which I think is the other test doing (?).

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/gpm/webflow/ImportGooglePasswordsWebFlowViewModelTest.kt
```diff
@@ -103,6 +119,239 @@ class ImportGooglePasswordsWebFlowViewModelTest {
...
+        val expectedReauthData = ReAuthenticationDetails(password = "test-password")
+        configureFeatureFlagEnabled()
+        whenever(reauthenticationHandler.retrieveReauthData(url)).thenReturn(expectedReauthData)
+
+        val result = testee.getReauthData(url)
+
+        assertEquals(expectedReauthData, result)
+    }
+
+    private fun configureFeatureFlagEnabled() {
```
cmonfortep: Can you update the method names so it's clear which featureFlag is enabled? Same for the method below.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/AutofillJavascriptInterface.kt
```diff
@@ -262,7 +262,23 @@ class AutofillStoredBackJavascriptInterface @Inject constructor(
...
+    }
+
+    private suspend fun notifyListenerThatCredentialsAvailableToInject(
+        url: String,
+        finalCredentialList: List<LoginCredentials>,
+        triggerType: LoginTriggerType,
+        request: AutofillDataRequest,
+    ) {
+        when (val currentCallback = callback) {
+            is InternalCallback -> {
```
cmonfortep, line 276: Not sure how you get here since Callback is null unless you give it a value from the caller, but I haven't found where you assign the `InternalCallback`. You need to explain me this one day :D I'm curious what I'm missing.

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/configuration/RealAutofillRuntimeConfigProviderTest.kt
```diff
@@ -103,14 +104,14 @@ class RealAutofillRuntimeConfigProviderTest {
     @Test
     fun whenAutofillNotEnabledThenConfigurationUserPrefsCredentialsIsFalse() = runTest {
         configureAutofillCapabilities(enabled = false)
-        testee.getRuntimeConfiguration("", EXAMPLE_URL)
+        testee.getRuntimeConfiguration("", EXAMPLE_URL, ReAuthenticationDetails())
```
cmonfortep: If you go for my suggestion to declare an EmptyReauthenticationDetail, I would use them here, so it's clear this method (and the ones in this file) you don't have inmemory password.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/configuration/AutofillAvailableInputTypesProvider.kt
CDRussell: I like the idea of the default being more explicitly empty using that function; will do that, thanks

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/gpm/webflow/ImportGooglePasswordsWebFlowViewModelTest.kt
CDRussell: Not quite; whether it can be done or not also depends on the scenario. So this is specifically testing with `scenarioAllowsReAuthentication = false`

autofill/autofill-impl/src/test/java/com/duckduckgo/autofill/impl/importing/gpm/webflow/ImportGooglePasswordsWebFlowViewModelTest.kt
CDRussell: Yeah, think it is a dupe, good catch. Will remove one.

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/impl/AutofillJavascriptInterface.kt
CDRussell: It's set when `InlineBrowserAutofill.addJsInterface()` is called, where the callback is provided as a param. 
- `BrowserTabFragment` provides a regular `Callback` 
- `ImportGooglePasswordsWebFlowFragment` implements `InternalCallback` so that it can have the extra data it needs (that isn't needed in the public API). 

There might be more elegant ways of doing the above, but seems ok for now 🙂

CDRussell:
>Tested on an emulator and found some issues 

We followed up on that elsewhere; issue was that the emulator was set to destroy activities as soon as you leave them. 

I've addressed the feedback, and made a commit setting the default value of the feature flag to be enabled. One last thing I'm waiting for before I merge is a release of `c-s-s` library.

PR #6785, opened 2025-09-11
suppress automatic launch of the Input Screen if a new tab request via an Intent

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1211256901076722?focus=true

### Description

Resolves an issue with Input Screen opening unexpectedly when app was launched from an intent.

If the app was left on NTP and closed, and then re-opened through a browsable intent (either from another app or from a widget), then before the target page was loaded the NTP could trigger a launch of the Input Screen that would cover up the desired content.

### Steps to test this PR

- [x] Set the app as the default browser.
- [x] Go to Settings -> AI Features and enable the experimental address bar.
- [x] Create a new tab and leave it on the New Tab Page.
- [x] Close the app and ensure the process is killed (swipe from history).
- [x] Go to anther app that can launch a browsable intent.
- [x] Click on a link to open in the browser.
- [x] Verify that the app launches, opens a new tab with the desired page, and the Input Screen is not shown.
- [x] Open a new tab and verify that the Input Screen is shown automatically.
- [x] Go to the home screen and add a search/favorites widget.
  - [x] Ensure you have some favorites set.
- [x] Create a new tab and leave it on the New Tab Page.
- [x] Close the app and ensure the process is killed (swipe from history).
- [x] Click on a favorite from the widget.
- [x] Verify that the app launches, opens the favorite page in a new tab, and the Input Screen is not shown.
- [x] Open a new tab and verify that the Input Screen is shown automatically.

The original issue was a race condition, so if you want to ensure the fix works, you can apply this diff:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
index 87b8e1e76..9acd0195d 100644
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -616,7 +616,7 @@ open class BrowserActivity : DuckDuckGoActivity() {
                 lifecycleScope.launch { viewModel.onOpenShortcut(sharedText) }
             } else if (intent.getBooleanExtra(LAUNCH_FROM_FAVORITES_WIDGET, false)) {
                 logcat { "Favorite clicked from widget $sharedText" }
-                lifecycleScope.launch { viewModel.onOpenFavoriteFromWidget(query = sharedText) }
+                lifecycleScope.launch { delay(5000); viewModel.onOpenFavoriteFromWidget(query = sharedText) }
             } else if (intent.getBooleanExtra(OPEN_IN_CURRENT_TAB_EXTRA, false)) {
                 logcat(WARN) { "open in current tab requested" }
                 if (currentTab != null) {
@@ -646,7 +646,7 @@ open class BrowserActivity : DuckDuckGoActivity() {
                     } else {
                         sharedText
                     }
-                    launchNewTab(query = query, sourceTabId = sourceTabId, skipHome = skipHome)
+                    lifecycleScope.launch { delay(5000); launchNewTab(query = query, sourceTabId = sourceTabId, skipHome = skipHome) }
                 } else {
                     lifecycleScope.launch { viewModel.onOpenInNewTabRequested(sourceTabId = sourceTabId, query = sharedText, skipHome = skipHome) }
                 }
```

This will delay the launch of a new tab when opened from a browsable intent or from favorites in a widget. When applied, performing the test steps above should result in the app opening, showing the NTP for a few seconds, then loading the target page. This confirms that even if NTP appears first, it doesn't prematurely trigger the Input Screen.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -583,7 +583,9 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
         if (launchNewSearch(intent)) {
             logcat(WARN) { "new tab requested" }
-            externalIntentProcessingState.onIntentRequestToChangeTab()
+            if (duckAiFeatureState.showInputScreenAutomaticallyOnNewTab.value) {
+                externalIntentProcessingState.onIntentRequestToChangeTab()
+            }
             launchNewTab()
             return
         }
@@ -602,7 +604,9 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
         val existingTabId = intent.getStringExtra(OPEN_EXISTING_TAB_ID_EXTRA)
         if (existingTabId != null) {
-            externalIntentProcessingState.onIntentRequestToChangeTab()
+            if (duckAiFeatureState.showInputScreenAutomaticallyOnNewTab.value) {
+                externalIntentProcessingState.onIntentRequestToChangeTab()
+            }
             openExistingTab(existingTabId)
             return
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -579,6 +583,7 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
         if (launchNewSearch(intent)) {
             logcat(WARN) { "new tab requested" }
+            externalIntentProcessingState.onIntentRequestToChangeTab()
```
malmstein: does it make sense to make this specifically about the new Input Screen? also behind a feature flag

PR #6839, opened 2025-09-24
Fix IllegalStateException when closing Duck.ai

joshliebe:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211438741846582?focus=true

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -840,9 +840,11 @@ open class BrowserActivity : DuckDuckGoActivity() {
         val fragment = duckAiFragment
         if (fragment?.isVisible == true) {
             animateDuckAiFragmentOut {
-                val transaction = supportFragmentManager.beginTransaction()
-                transaction.hide(fragment)
-                transaction.commit()
+                if (!supportFragmentManager.isStateSaved) {
+                    val transaction = supportFragmentManager.beginTransaction()
+                    transaction.hide(fragment)
+                    transaction.commit()
+                }
             }
         }
     }


joshliebe:
* **#6839** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6839?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6839?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -840,9 +840,11 @@ open class BrowserActivity : DuckDuckGoActivity() {
         val fragment = duckAiFragment
         if (fragment?.isVisible == true) {
             animateDuckAiFragmentOut {
-                val transaction = supportFragmentManager.beginTransaction()
-                transaction.hide(fragment)
-                transaction.commit()
+                if (!supportFragmentManager.isStateSaved) {
```
LukasPaczos, line 843: I'm not sure how exactly that scenario would play out, but shouldn't this be checked before we run any logic in the function? Otherwise, if a state is restored with fragment visible, our local `isDuckChatVisible` would be set to `false` already.

joshliebe: It needs to be checked when the animation completes, if we check it before, the state may have changed by then. Not sure how to deal with that inconsistency.

LukasPaczos: Right. Anyway, looks like a quick way to prevent the crash and worst case we'll have a one time issue with keyboard pop-up which will resolve itself either way, so that's not a big deal.

LukasPaczos (approved):

PR #6863, opened 2025-09-30
Apply lifecycle checks and fixes to addDocumentStartJavaScript

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1211506668810055?focus=true

### Description
* Check WebView attached before calling `WebViewCompat` and `postMessage` methods
* Check coroutine active before calling `WebViewCompat` and `postMessage` methods
* Launch WebViewCompat` and `postMessage` methods with `lifecycleScope` or `viewModelScope` whenever possible

### Steps to test this PR

_Feature 1_
- [x] Smoke test blob downloads
- [x] Smoke test app with `useNewWebCompatApis` enabled. Requires fresh install after setting setting `const val PRIVACY_REMOTE_CONFIG_URL = "https://duckduckgo.github.io/privacy-configuration/pr-3816/v4/android-config.json"`
- [x] Run tests in https://privacy-test-pages.site/privacy-protections/gpc/. Make sure `frame JS API ` is true 

_Message replies_ 
- [x] Launch https://w3c.github.io/web-share/demos/share-files.html
- [x] Click `Share`
- [x] Cancel native popup
- [x] Check message `Error sharing: AbortError: Share canceled` is displayed 

### UI changes
n/a

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -898,12 +898,14 @@ class BrowserTabFragment :
             contentScopeScripts.onResponse(it)
         }
 
-    private var currentWebShareReplyCallback: ((JSONObject) -> Unit)? = null
+    private var currentWebShareReplyCallback: (suspend (JSONObject) -> Unit)? = null
 
     private val webViewCompatWebShareLauncher =
         registerForActivityResult(WebViewCompatWebShareChooser()) { result ->
-            currentWebShareReplyCallback?.invoke(result)
-            currentWebShareReplyCallback = null
+            lifecycleScope.launch {
+                currentWebShareReplyCallback?.invoke(result)
+                currentWebShareReplyCallback = null
+            }
         }
 
     // Instantiating a private class that contains an implementation detail of BrowserTabFragment but is separated for tidiness
@@ -1991,19 +1993,29 @@ class BrowserTabFragment :
             is Command.ShowFireproofWebSiteConfirmation -> fireproofWebsiteConfirmation(it.fireproofWebsiteEntity)
             is Command.DeleteFireproofConfirmation -> removeFireproofWebsiteConfirmation(it.fireproofWebsiteEntity)
             is Command.RefreshAndShowPrivacyProtectionEnabledConfirmation -> {
-                webView?.let {
-                    webViewClient.configureWebView(it, null)
+                webView?.let { webView ->
+                    lifecycleScope.launch {
+                        webViewClient.configureWebView(webView, null)
+                        refresh()
+                        privacyProtectionEnabledConfirmation(it.domain)
+                    }
+                } ?: run {
+                    refresh()
+                    privacyProtectionEnabledConfirmation(it.domain)
                 }
-                refresh()
-                privacyProtectionEnabledConfirmation(it.domain)
             }
 
             is Command.RefreshAndShowPrivacyProtectionDisabledConfirmation -> {
-                webView?.let {
-                    webViewClient.configureWebView(it, null)
+                webView?.let { webView ->
+                    lifecycleScope.launch {
+                        webViewClient.configureWebView(webView, null)
+                        refresh()
+                        privacyProtectionDisabledConfirmation(it.domain)
+                    }
+                } ?: run {
+                    refresh()
+                    privacyProtectionDisabledConfirmation(it.domain)
                 }
-                refresh()
-                privacyProtectionDisabledConfirmation(it.domain)
             }
 
             is NavigationCommand.Navigate -> {
@@ -3201,28 +3213,31 @@ class BrowserTabFragment :
                     }
                 },
             )
-            webViewClient.configureWebView(
-                it,
-                object : WebViewCompatMessageCallback {
-                    override fun process(
-                        context: String,
-                        featureName: String,
-                        method: String,
-                        id: String?,
-                        data: JSONObject?,
-                        onResponse: (JSONObject) -> Unit,
-                    ) {
-                        viewModel.webViewCompatProcessJsCallbackMessage(
-                            context = context,
-                            featureName = featureName,
-                            method = method,
-                            id = id,
-                            data = data,
-                            onResponse = onResponse,
-                        )
-                    }
-                },
-            )
+            lifecycleScope.launch {
+                webViewClient.configureWebView(
+                    it,
+                    object : WebViewCompatMessageCallback {
+                        override fun process(
+                            context: String,
+                            featureName: String,
+                            method: String,
+                            id: String?,
+                            data: JSONObject?,
+                            onResponse: suspend (JSONObject) -> Unit,
+                        ) {
+                            viewModel.webViewCompatProcessJsCallbackMessage(
+                                context = context,
+                                featureName = featureName,
+                                method = method,
+                                id = id,
+                                data = data,
+                                onResponse = onResponse,
+                            )
+                        }
+                    },
+                )
+            }
+
             duckPlayerScripts.register(
                 it,
                 object : JsMessageCallback() {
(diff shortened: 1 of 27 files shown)


CrisBarreiro:
* **#6864** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6863** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1989,19 +1991,29 @@ class BrowserTabFragment :
...
             is Command.RefreshAndShowPrivacyProtectionEnabledConfirmation -> {
-                webView?.let {
-                    webViewClient.configureWebView(it, null)
+                webView?.let { safeWebView ->
+                    lifecycleScope.launch {
+                        webViewClient.configureWebView(safeWebView, null)
+                        refresh()
+                        privacyProtectionEnabledConfirmation(it.domain)
+                    }
+                } ?: run {
```
CDRussell: nit: curious if we need this `?: run` block
- this would only apply if the `webView` is null, right?
- if the `WebView` is null, we probably don't want anything else to happen anyway?. do we really need to ensure `refresh` and `privacyProtectionEnabledConfirmation` are called in this state? if not, we can maybe delete the `?: run` block altogether here and elsewhere

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4003,10 +4021,12 @@ class BrowserTabFragment :
...
-            webViewClient.destroy(it)
-            it.destroy()
-        }
-        webView = null
+            lifecycleScope.launch {
+                webViewClient.destroy(it)
+                it.destroy()
+                webView = null
+            }
+        } ?: run { webView = null }
```
CDRussell: this seems odd to me; this block will only run if `webView` is already `null`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4003,10 +4021,12 @@ class BrowserTabFragment :
         if (::webViewContainer.isInitialized) webViewContainer.removeAllViews()
         webView?.let {
             it.removeSystemAutofillCallback()
-            webViewClient.destroy(it)
-            it.destroy()
-        }
-        webView = null
+            lifecycleScope.launch {
```
CDRussell: I don't have a solid suggestion here so just highlighting it for more consideration and to gather my thoughts more on it:

- launching a coroutine when trying to `destroyWebView` seems a little dangerous
- for one thing, `destroyWebView` is called from multiple places and the expectation is that after the method executes then the `webView` is destroyed. 
  - However, this change introduces new behaviour since it launches a coroutine; the `destroyWebView` execution would finish but the coroutine might not have run yet. 
  - We might run into the situation where we're trying to create a new `webView` and trying to destroy it at the same time

If we need to call `suspend` functions from `destroyWebView`, then `destroyWebView` should also be a `suspend` function rather than launching a coroutine, IMO, so callers can be sure it's actually finished after the function is called. Assuming we can't avoid calling a `suspend` function at this time?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3249,14 +3254,14 @@ class BrowserTabViewModel @Inject constructor(
         command.postValue(RequestFileDownload(url, contentDisposition, mimeType, requestUserConfirmation))
     }
 
-    @SuppressLint("RequiresFeature") // it's already checked in isBlobDownloadWebViewFeatureEnabled
-    private fun postMessageToConvertBlobToDataUri(url: String) {
+    @SuppressLint("RequiresFeature", "PostMessageUsage") // it's already checked in isBlobDownloadWebViewFeatureEnabled
+    private fun postMessageToConvertBlobToDataUri(webView: WebView, url: String) {
         appCoroutineScope.launch(dispatchers.main()) {
```
CDRussell: Flagging that we're still using `appCoroutineScope.launch` here which is likely wrong for the same lifecycle-related reasons as your other changes. Not sure if you deliberately wanted to keep this one as it is to limit the scope of your changes or not, but flagging in case it was missed.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1991,19 +1993,29 @@ class BrowserTabFragment :
             is Command.ShowFireproofWebSiteConfirmation -> fireproofWebsiteConfirmation(it.fireproofWebsiteEntity)
             is Command.DeleteFireproofConfirmation -> removeFireproofWebsiteConfirmation(it.fireproofWebsiteEntity)
             is Command.RefreshAndShowPrivacyProtectionEnabledConfirmation -> {
-                webView?.let {
-                    webViewClient.configureWebView(it, null)
+                webView?.let { safeWebView ->
```
CrisBarreiro, line 1996: This is going to be simplified on the next PR

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -3201,28 +3213,31 @@ class BrowserTabFragment :
...
-                            featureName = featureName,
-                            method = method,
-                            id = id,
-                            data = data,
-                            onResponse = onResponse,
-                        )
-                    }
-                },
-            )
+            lifecycleScope.launch {
```
CrisBarreiro, line 3210: On the next PR I'm gonna simplify this and launch the coroutine from the ViewModel instead

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsWebMessagingPlugin.kt
```diff
@@ -89,56 +88,68 @@ class ContentScopeScriptsWebMessagingPlugin @Inject constructor(
...
                         .forEach { handler ->
                             handler.process(jsMessage)?.let { processResult ->
                                 when (processResult) {
                                     is SendToConsumer -> {
-                                        sendToConsumer(jsMessageCallback, jsMessage, replyProxy)
+                                        sendToConsumer(webView, jsMessageCallback, jsMessage, replyProxy)
                                     }
                                     is SendResponse -> {
-                                        onResponse(jsMessage, replyProxy)
+                                        appCoroutineScope.launch {
```
CDRussell: we are trying to get away from `appCoroutineScope.launch` in these flows; is this required?

CDRussell: Suggestion as discussed on zoom: 

```
webView.findViewTreeLifecycleOwner()?.lifecycleScope?.launch { }
```

instead of `appCoroutineScope`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4023,12 +4017,9 @@ class BrowserTabFragment :
         if (::webViewContainer.isInitialized) webViewContainer.removeAllViews()
         webView?.let {
             it.removeSystemAutofillCallback()
-            lifecycleScope.launch {
-                webViewClient.destroy(it)
```
CDRussell: if `webViewClient.destroy(it)` isn't getting called any more, should the function be deleted?

CDRussell (approved):

PR #6866, opened 2025-10-01
Implement custom tab handling for app links in BrowserTabViewModel.

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211450043710855?focus=true

### Description

Added support for better handling app links in custom tabs. When an app link is opened in a custom tab, the app will now directly open the app link without showing the app link prompt, regardless of the user's app link prompt settings.

### Steps to test this PR

_App Links in Custom Tabs_
- [x] See https://app.asana.com/1/137249556945/task/1211450043710855/comment/1211477223443413?focus=true for steps and credentials.
- [x] See attached video to understand the flow and prerequisites (DDG is custom tab and DDG -> Settings -> Permissions -> Open Links is Apps is set to either `Ask Every Time` or `Never`).

_Regular App Links_
- [x] Open the app and set prerequisites: DDG is custom tab and DDG -> Settings -> Permissions -> Open Links is Apps is set to either `Ask Every Time` or `Never`.
- [x] Navigate to a page with an app link. I.e: https://play.google.com/store/apps/details?id=com.duckduckgo.mobile.android&hl=en-US
- [x] Verify that you are prompted to open the link in the Play Store app.

### UI changes
No UI changes, only behavior changes in Custom Tab for app links.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -1029,6 +1029,7 @@ class BrowserTabFragment :
 
         if (savedInstanceState == null) {
             viewModel.onViewReady()
+            viewModel.setIsCustomTab(tabDisplayedInCustomTabScreen)
             messageFromPreviousTab?.let {
                 processMessage(it)
             }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -584,6 +584,7 @@ class BrowserTabViewModel @Inject constructor(
     private var isProcessingTrackingLink = false
     private var isLinkOpenedInNewTab = false
     private var allowlistRefreshTriggerJob: Job? = null
+    private var isCustomTabScreen: Boolean = false
 
     private val fireproofWebsitesObserver =
         Observer<List<FireproofWebsiteEntity>> {
@@ -796,6 +797,10 @@ class BrowserTabViewModel @Inject constructor(
         site = siteLiveData.value
 
         initialUrl?.let { buildSiteFactory(it, stillExternal = isExternal) }
+    }
+
+    fun setIsCustomTab(isCustomTab: Boolean) {
+        this.isCustomTabScreen = isCustomTab
     }
 
     fun onViewReady() {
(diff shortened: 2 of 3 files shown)


anikiki:
* **#6866** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6866?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -1029,6 +1029,7 @@ class BrowserTabFragment :
 
         if (savedInstanceState == null) {
             viewModel.onViewReady()
+            viewModel.setIsCustomTab(tabDisplayedInCustomTabScreen)
```
LukasPaczos, line 1032: Shouldn't this also be called when activity is recreated?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -584,6 +584,7 @@ class BrowserTabViewModel @Inject constructor(
     private var isProcessingTrackingLink = false
     private var isLinkOpenedInNewTab = false
     private var allowlistRefreshTriggerJob: Job? = null
+    private var isCustomTabScreen: Boolean = false
```
LukasPaczos, line 587: There are nearly a dozen other functions in this view model that take `"isCustomTab"` as a parameter. I understand that’s not easily doable in this case since the callback is invoked from the web view rather than the fragment.

That said, could you create a follow-up task so that we clean this up and have those functions rely on the local state instead? That way, the view model has a single source of truth.

anikiki: Will do this, thanks 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
anikiki: The view model should live. Will add it as improvement in the follow-up task 👍

PR #6872, opened 2025-10-02
Fix: ANR in the Tab switcher

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1211510795041274?focus=true

### Description

An attempt to fix an ANR when layout out the tab switcher items. Since the stack trace pointed to a delay when calculating breakpoints, I've added a couple of optimizations to how we process and display the title in the item's TextView:

- updated the title's `layout_width` attribute to fill the item's width
- changed the `breakStrategy` to simple
- added a limit to the title length (could be using long URLs if no title available)

### Steps to test this PR

It's not easy to reproduce, so just smoke-check that items are displayed properly. Otherwise QA-optional.


karlenDimla (approved):
I tried to test the tab switcher as much as possible and it seems to work as expected.

PR #6883, opened 2025-10-03
Revert "[Android] The latest Android 16 update broke the update for the Search and Favorites widget (#6843)"

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211543114832843?focus=true 

### Description
Reverts https://github.com/duckduckgo/Android/pull/6843

### Steps to test this PR

- [ ] Add a favorite site
- [ ] Add the `search and favorites` widget to the home screen
- [ ] Check no crashes across a range of Android OS versions

PR #6843, opened 2025-09-24
Refactor widget services and item factories for favorites and empty states.

anikiki:
Task/Issue URL: 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |
!(Upload before screenshot)|(Upload after screenshot)|

PR #6885, opened 2025-10-03
Revert "Revert "[Android] The latest Android 16 update broke the update for the Search and Favorites widget (#6843)" (#6883)"

anikiki:
Task/Issue URL: 

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |


anikiki:
* **#6885** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/6885?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/6885?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

landomen:
@anikiki  @0nko It's still crashing on my S10 (Android 12) with `24dp` size (84px). Anything over `14dp` (50px, 10kb) is crashing. However, at `14dp` the image quality is pretty poor, but still passable. I'm worried that while 14dp works on my phone, it might not work on an older phone like the S7 which is one of the crash reports.

Should we go with `14dp` and see how if it fixes the issue for all devices? Or invest more time in finding an alternative solution?

One idea that comes to mind is continuing @0nko's workaround of using URIs, but also supporting it for the generated avatars by storing it to disk and then using URI for that as well. Not sure if it would work, but worth a try.

Screenshot_20251003_173559

landomen:
I had some time, so I tried continuing the solution that @0nko started with loading favicons via an URI, but expanding it to also load generated placeholder avatars via URI. It turns out that it works! 🥳 

The benefit with this solution is that it should work on all devices, doesn't require hardcoding image size, and the image quality is unchanged. The drawback is that we have to store the generated avatars to the disk first.

Here is a quick solution that Cursor implemented, but needs to be cleaned-up and improved:
https://github.com/duckduckgo/Android/compare/fix/ana/android_take_2_on_this_issue_the_latest_android16_update_broke_the_update_for_the_search_and_favorites_widget...feature/domen/widget-placeholder-uris?expand=1

Screenshot_20251017_181944

PR #6890, opened 2025-10-05
Fix tab switcher menu button visibility on real devices

### LINKED ISSUE #6702: [Bug] Tab switcher '...' menu is invisible
### Describe the bug

It appears that the tab switcher menu is hidden when the app is downloaded from the app store.

I tried building the app myself, when I install it on a real phone, I can't see the menu. I can only see it when installed on a virtual device using AVD. Is not this a bug? 

Image

### How to Reproduce

Install app from play store. Go to the tab switcher, you should see a '...' menu.

### Expected behavior

In the tab switcher, the '...' menu should be visible.

### Environment

```markdown
- DDG App Version: latest
- Device: Pixel 4a & Pixel 8a
- OS: Android LOS
```

mihf05:
## Description

Fixes #6702 - Tab switcher '...' menu is invisible

The issue was that the popup menu button was only having its `isEnabled` property set, but not its `isVisible` property. On some Android devices and different Android versions/themes, disabled menu items can become invisible or very faint, which caused the menu to not appear on real devices while still being visible on virtual devices.

## Changes Made

- Explicitly set `isVisible = true` for the popup menu toolbar button in `TabSwitcherMenuExt.kt`
- Added comment explaining the fix for future reference
- Ensured the menu button is always visible regardless of enabled state

## Testing

The fix addresses the root cause by ensuring the menu button visibility is explicitly controlled, which should resolve the difference in behavior between virtual and real devices.

## Root Cause Analysis

The issue occurred because:
1. The menu button was only having `isEnabled` set based on `dynamicMenu.isMenuButtonEnabled`
2. Different Android devices/versions handle disabled menu items differently - some make them invisible or very faint
3. Virtual devices may use a different theme or rendering that keeps disabled items visible
4. Real devices were applying system-level theming that hid disabled menu items

This fix ensures consistent behavior across all devices by explicitly controlling visibility.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherMenuExt.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherMenuExt.kt
@@ -141,7 +141,10 @@ fun Menu.createDynamicInterface(
         }
     }
 
-    findItem(R.id.popupMenuToolbarButton).isEnabled = dynamicMenu.isMenuButtonEnabled
+    findItem(R.id.popupMenuToolbarButton).apply {
+        isEnabled = dynamicMenu.isMenuButtonEnabled
+        isVisible = true // Always show the menu button regardless of enabled state
+    }
     findItem(R.id.fireToolbarButton).isVisible = dynamicMenu.isFireButtonVisible
     findItem(R.id.duckAIToolbarButton).isVisible = dynamicMenu.isDuckAIButtonVisible
     findItem(R.id.newTabToolbarButton).isVisible = dynamicMenu.isNewTabButtonVisible


mihf05:
Hi maintainers! 👋

This PR fixes issue #6702 - Tab switcher '...' menu is invisible on real devices.

**Quick Summary:**
- This is a simple visibility fix that addresses device-specific rendering differences
- The root cause was that disabled menu items are handled differently across Android devices/versions  
- Virtual devices kept disabled items visible, but real devices made them invisible/faint
- The fix explicitly sets isVisible = true to ensure consistent behavior

**Ready for Review:**
Could someone please approve the workflow runs and review this fix? It's a straightforward change that should resolve the reported issue.

Thanks! 🙏

Fixes #6702

PR #6895, opened 2025-10-06
PIR: Implement resume emailConfirmation

karlenDimla:
Task/Issue URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1211491791085983?focus=true 

### Description
See attached task description.

### Steps to test this PR
https://app.asana.com/1/137249556945/project/481882893211075/task/1211559446932897?focus=true

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
@@ -74,7 +74,6 @@ class RealPirEmailConfirmation @Inject constructor(
         context: Context,
         runType: RunType,
     ): Result<Unit> = withContext(dispatcherProvider.io()) {
-        // onJobStarted()
         if (jobRecords.isEmpty()) {
             logcat { "PIR-EMAIL-CONFIRMATION: Nothing to scan here." }
             return@withContext Result.success(Unit)
@@ -104,14 +103,14 @@ class RealPirEmailConfirmation @Inject constructor(
 
         jobRecordsParts.mapIndexed { index, partSteps ->
             logcat { "PIR-EMAIL-CONFIRMATION:: Record part [$index] -> ${partSteps.size}" }
-            logcat { "PIR-SCAN: Record part [$index] breakdown -> ${partSteps.map { it.first.id to it.second.brokerName }}" }
+            logcat { "PIR-EMAIL-CONFIRMATION: Record part [$index] breakdown -> ${partSteps.map { it.first.id to it.second.brokerName }}" }
             // We want to run the runners in parallel but wait for everything to complete before we proceed
             async {
                 partSteps.forEach { (profile, step) ->
-                    logcat { "PIR-SCAN: Start scan on runner=$index for profile=$profile with step=$step" }
+                    logcat { "PIR-EMAIL-CONFIRMATION: Resuming opt-out on runner=$index for profile=$profile with step=$step" }
                     runners[index].start(profile, listOf(step))
                     runners[index].stop()
-                    logcat { "PIR-SCAN: Finish scan on runner=$index for profile=$profile with step=$step" }
+                    logcat { "PIR-EMAIL-CONFIRMATION: Finished resuming opt-out on runner=$index for profile=$profile with step=$step" }
                 }
             }
         }.awaitAll()
@@ -175,7 +174,6 @@ class RealPirEmailConfirmation @Inject constructor(
             it.stop()
         }
         runners.clear()
-        // onJobStopped()
     }
 
     private fun cleanPreviousRun() {
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
@@ -122,7 +122,7 @@ interface PirRepository {
         profileQueryId: Long,
     ): List<ExtractedProfile>
 
-    suspend fun getExtractedProfiles(
+    suspend fun getExtractedProfile(
         extractedProfileId: Long,
     ): ExtractedProfile?
 
@@ -415,7 +415,7 @@ internal class RealPirRepository(
                 }
         }
 
-    override suspend fun getExtractedProfiles(
+    override suspend fun getExtractedProfile(
         extractedProfileId: Long,
     ): ExtractedProfile? =
         withContext(dispatcherProvider.io()) {
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/BrokerStepsParser.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/BrokerStepsParser.kt
@@ -154,7 +154,7 @@ class RealBrokerStepsParser @Inject constructor(
         emailConfirmationJob: EmailConfirmationJobRecord,
     ): BrokerStep? = withContext(dispatcherProvider.io()) {
         return@withContext runCatching {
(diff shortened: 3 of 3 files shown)


karlenDimla:
* **#6895** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6895?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6895?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirRunStateHandler.kt
```diff
@@ -185,24 +195,32 @@ class RealPirRunStateHandler @Inject constructor(
...
             }
         }
 
+    private suspend fun handleBrokerRecordEmailConfirmationStarted(pirRunState: BrokerRecordEmailConfirmationStarted) {
+        jobRecordUpdater.recordEmailConfirmationAttempt(pirRunState.emailConfirmationJobRecord)
+    }
+
+    private suspend fun handleBrokerRecordEmailConfirmationCompleted(pirRunState: BrokerRecordEmailConfirmationCompleted) {
+        // If the attempt failed, We don't do anything. If the job has still attempts left it will be retried.
+        // If the attempts are maxed, it will be cleaned up in the next run. (Higher chance of being executed)
```
landomen, line 210: I appreciate the comments, thank you

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
```diff
@@ -122,6 +122,10 @@ interface PirRepository {
         profileQueryId: Long,
     ): List<ExtractedProfile>
 
+    suspend fun getExtractedProfiles(
```
landomen: nit: rename to `getExtractedProfile` since we're only returning a single profile

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
```diff
@@ -69,14 +73,117 @@ class RealPirEmailConfirmation @Inject constructor(
         jobRecords: List<EmailConfirmationJobRecord>,
         context: Context,
         runType: RunType,
-    ): Result<Unit> = Result.success(Unit)
+    ): Result<Unit> = withContext(dispatcherProvider.io()) {
+        // onJobStarted()
```
landomen: nit: should this be removed?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
```diff
@@ -69,14 +73,117 @@ class RealPirEmailConfirmation @Inject constructor(
...
 
     override fun stop() {
-        logcat { "PIR-SCAN: Stopping all runners" }
+        logcat { "PIR-EMAIL-CONFIRMATION: Stopping all runners" }
         runners.forEach {
             it.stop()
         }
         runners.clear()
-        onJobStopped()
+        // onJobStopped()
```
landomen: nit: should this be removed?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
```diff
@@ -69,14 +73,117 @@ class RealPirEmailConfirmation @Inject constructor(
...
+
+        logcat { "PIR-EMAIL-CONFIRMATION: Total parts ${jobRecordsParts.size}" }
+
+        jobRecordsParts.mapIndexed { index, partSteps ->
+            logcat { "PIR-EMAIL-CONFIRMATION:: Record part [$index] -> ${partSteps.size}" }
+            logcat { "PIR-SCAN: Record part [$index] breakdown -> ${partSteps.map { it.first.id to it.second.brokerName }}" }
+            // We want to run the runners in parallel but wait for everything to complete before we proceed
+            async {
+                partSteps.forEach { (profile, step) ->
+                    logcat { "PIR-SCAN: Start scan on runner=$index for profile=$profile with step=$step" }
```
landomen: Q: should the log tag be "PIR-EMAIL-CONFIRMATION" instead of "PIR-SCAN"?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
```diff
@@ -69,14 +73,117 @@ class RealPirEmailConfirmation @Inject constructor(
...
+        val processedJobRecords = processJobRecords(jobRecords)
+        logcat { "PIR-EMAIL-CONFIRMATION: Total processed records ${processedJobRecords.size}" }
+
+        if (processedJobRecords.isEmpty()) {
+            logcat { "PIR-EMAIL-CONFIRMATION: No job records." }
+            return@withContext Result.success(Unit)
+        }
+
+        val script = pirCssScriptLoader.getScript()
+        maxWebViewCount = minOf(processedJobRecords.size, MAX_DETACHED_WEBVIEW_COUNT)
```
landomen, line 92: Q: what is the max detached webview count based on? Can there be an issue if for example this email confirmation is running and also scan or opt-out? In that case, theoretically there would be `MAX_DETACHED_WEBVIEW_COUNT` * 2 or * 3 created

landomen (approved):
Works as expected 👍 Left a few comments and questions, but nothing blocking this from merging

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
karlenDimla: @landomen very good question actually. A bit of context, initially we set it to the total number of cores of the device since we were under the assumption that they will be fully parallel. However, after further investigations, we understood that they are not entirely parallel. Some interactions with the webview still has to happen and wait for the main thread(but async). 

Because of this, we have been increasing and benchmarking the total number of webviews to see how much will be able to function relatively well. So far 20 has been an OK number - we will know once we get more users.

Now to the question of "_Can there be an issue if for example this email confirmation is running and also scan or opt-out_”, I don’t think it will be a big issue:

1. Worst case it will be *2 (as scan/opt-outs don’t happen in parallel together).
2. Realistically, only 4 brokers send links today so the chances of us maximizing the limit for email confirmation is low.
3. The email confirmation webviews will be very short-lived. Majority (except for 1 broker) have simple steps for email confirmation - usually just expectation action after loading the links. Hence if we maxed it, it will just for a short time.


We can consider lessening it - but i think we could stick with the current numbers and see how it plays out in the wild. What do you think?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
karlenDimla: I am actually still contemplating about it. 😅 I will remove for now. My problem here is to handle the scenario when the scan and email confirmation run together. We want to have the the callbacks running as long as one type of job is running. I need to think about that more. Will do in another task.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/email/PirEmailConfirmation.kt
landomen: Sorry for missing this. Thanks for the explanation. I agree, let's stick with the current number

PR #6907, opened 2025-10-08
Pause WebView before updating scripts

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1211578851788393?focus=true 

### Description
Stop page loading and JS execution before removing/adding script. Resume once operation completed

### Steps to test this PR

_Pre-requisites_
- [x] Apply this [patch](https://duckduckgo-my.sharepoint.com/:u:/p/cbarreiro/EfoKD1EMQPxFhWiH_LHI2wgB0RxCvVUtJBiCx7ofwCsTeQ?e=Lhx8IW)
- [x] Do a clean install of the app

_Feature 1_
- [x] Smoke test app with `pauseWebViewBeforeUpdatingScript` set to true

### UI changes
n/a

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
+++ b/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
@@ -21,6 +21,7 @@ import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.feature.toggles.api.Toggle
 import com.duckduckgo.feature.toggles.api.Toggle.DefaultFeatureValue
 import com.duckduckgo.feature.toggles.api.Toggle.DefaultFeatureValue.FALSE
+import com.duckduckgo.feature.toggles.api.Toggle.DefaultFeatureValue.INTERNAL
 import com.duckduckgo.feature.toggles.api.Toggle.DefaultFeatureValue.TRUE
 
 /**
@@ -187,6 +188,6 @@ interface AndroidBrowserConfigFeature {
     @Toggle.DefaultValue(TRUE)
     fun updateScriptOnProtectionsChanged(): Toggle
 
-    @Toggle.DefaultValue(TRUE)
+    @Toggle.DefaultValue(INTERNAL)
     fun pauseWebViewBeforeUpdatingScript(): Toggle
 }


CrisBarreiro:
* **#6907** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6907?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6907?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
```diff
@@ -194,4 +194,7 @@ interface AndroidBrowserConfigFeature {
 
     @Toggle.DefaultValue(TRUE)
     fun updateScriptOnProtectionsChanged(): Toggle
+
+    @Toggle.DefaultValue(TRUE)
```
CrisBarreiro: I'll make this false before merging, just in case

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4261,6 +4261,17 @@ class BrowserTabViewModel @Inject constructor(
             return
         }
 
+        val pauseWebViewBeforeUpdatingScript = withContext(dispatchers.io()) {
+            androidBrowserConfig.pauseWebViewBeforeUpdatingScript().isEnabled()
+        }
+
+        if (pauseWebViewBeforeUpdatingScript) {
+            withContext(dispatchers.main()) {
+                webView.stopLoading()
```
CDRussell: why not `webView.onPause()` too out of curiosity?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4272,6 +4283,10 @@ class BrowserTabViewModel @Inject constructor(
         } else {
             addDocumentStartJavaScript(webView)
         }
+
+        if (pauseWebViewBeforeUpdatingScript) {
+            webView.resumeTimers()
```
CDRussell: only concern i have with trying this is that `webView.pauseTimers()` affects all `WebView` instances and so we really need to ensure that `webView.resumeTimers()` fires. as it is, it's possible (however rarely) that the coroutine is cancelled between pause and resume.

could maybe use `try / finally` to try ensure it does always resume?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
CrisBarreiro: Because `onPause` doesn't pause JavaScript and, it's unclear whether the current site will stop loading as well. This is what the docs say:
> Does a best-effort attempt to pause any processing that can be paused safely, such as animations and geolocation. Note that this call does not pause JavaScript. To pause JavaScript globally, use pauseTimers. To resume WebView, call onResume.

Since we're gonna trigger a reload, we thought stopping the load in case the site is still loading plus pausing JS execution would be safer in terms of preventing race conditions

CDRussell: it was more of a "why not **also** call `onPause`" and not a suggestion to do it instead of anything else. but `stopLoading` is probably sufficient

app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
CDRussell: Why not use `INTERNAL` for default value? 

Whether going for `INTERNAL` or `FALSE` as default, it would be good to make that change now and if `TRUE` is required for testing then move it to the testing patch (to avoid accidentally merging it as true if it's forgotten)

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4263,16 +4263,35 @@ class BrowserTabViewModel @Inject constructor(
...
+                        (plugin.context == "contentScopeScripts")
+                    }.forEach {
+                        it.addDocumentStartJavaScript(webView)
+                    }
+            } else {
+                addDocumentStartJavaScript(webView)
+            }
+        } finally {
+            if (pauseWebViewBeforeUpdatingScript) {
+                appCoroutineScope.launch(dispatchers.main()) {
```
CDRussell: I _think_ we can avoid this.

if we have a `withContext(dispatchers.main())` outside of the `try` block then 

- the `try` block will run on the main thread
- you can use nested `withContext(dispatchers.io())` for checking FF as you do already
- then in the `finally` block it'll be on the `main` thread already and you can safely call `webView.resumeTimers`
- even if the coroutine is cancelled after `pauseTimers()` is called, the `finally` block should run (on main thread) and give you a chance to `resumeTimers`

CrisBarreiro: Tested this. However, even if we can guarantee `resumeTimers` is called, that still doesn't guarantee timers will actually be resumed if the WebView instance is already destroyed, and that affects any other present or future WebView as well. Closing the PR and getting back to the drawing board

PR #6910, opened 2025-10-09
Cache autoconsent settings in singleton

sammacbeth:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1163321984198618/task/1211608075081851?focus=true

### Description
We are currently parsing the autoconsent settings JSON object every time the `init` message is received (every page load). As this is a heavy operation (parsing a >300kb string), and the output changes infrequently (config changes max once per day), we can cache the value when the config is downloaded, then only reparse if the settings value changes.

### Steps to test this PR

 1. Visit https://privacy-test-pages.site/features/autoconsent/index.html.
 2. The reject button should be clicked.

### DIFF
--- a/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/RealAutoconsent.kt
+++ b/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/RealAutoconsent.kt
@@ -65,7 +65,7 @@ class RealAutoconsent @Inject constructor(
 
     init {
         if (isMainProcess) {
-            onPrivacyConfigDownloaded()
+            cacheSettingsFromConfig()
         }
     }
 
@@ -136,6 +136,10 @@ class RealAutoconsent @Inject constructor(
 
     override fun onPrivacyConfigDownloaded() {
         settingsRepository.invalidateCache()
+        cacheSettingsFromConfig()
+    }
+
+    private fun cacheSettingsFromConfig() {
         appCoroutineScope.launch(dispatcherProvider.io()) {
             val settingsJson = autoconsent.self().getSettings()
             if (settingsCache.getHash() != settingsJson.hashCode()) {


autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/RealAutoconsent.kt
```diff
@@ -18,22 +18,31 @@ package com.duckduckgo.autoconsent.impl
...
 import com.duckduckgo.common.utils.plugins.PluginPoint
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.privacy.config.api.PrivacyConfigCallbackPlugin
 import com.duckduckgo.privacy.config.api.UnprotectedTemporary
 import com.squareup.anvil.annotations.ContributesBinding
 import com.squareup.anvil.annotations.ContributesMultibinding
+import com.squareup.moshi.Moshi
 import javax.inject.Inject
 
+object SettingsCache {
```
joshliebe: Instead of `object`, rather use `@SingleInstanceIn(AppScope::class)` and create a separate singleton class. Then inject that into the `RealAutoconsent` `constructor`.

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/InitMessageHandlerPlugin.kt
```diff
@@ -43,11 +43,14 @@ class InitMessageHandlerPlugin @Inject constructor(
     @AppCoroutineScope val appCoroutineScope: CoroutineScope,
     private val dispatcherProvider: DispatcherProvider,
     private val settingsRepository: AutoconsentSettingsRepository,
-    private val autoconsentFeature: AutoconsentFeature,
 ) : MessageHandlerPlugin {
 
     private val moshi = Moshi.Builder().add(JSONObjectAdapter()).build()
 
+    fun getSettings(): AutoconsentSettings? {
+        return SettingsCache.settings
```
joshliebe: Inject the same singleton instance mentioned in https://github.com/duckduckgo/Android/pull/6910/files#r2417921419 into this class and remove this function

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/RealAutoconsent.kt
```diff
@@ -120,5 +136,20 @@ class RealAutoconsent @Inject constructor(
...
+        val settingsJson = autoconsent.self().getSettings()
+        if (SettingsCache.hash != settingsJson.hashCode()) {
+            settingsJson?.let {
+                updateSettingsCache(it)
+            }
+        }
+    }
+
+    companion object {
+        fun updateSettingsCache(settingsJson: String) {
```
joshliebe: Can remove this `companion object` and move the function into this class or the singleton

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/RealAutoconsent.kt
```diff
@@ -120,5 +136,20 @@ class RealAutoconsent @Inject constructor(
 
     override fun onPrivacyConfigDownloaded() {
         settingsRepository.invalidateCache()
+        val settingsJson = autoconsent.self().getSettings()
```
joshliebe: `getSettings` should be called in a coroutine with `dispatchers.io()`

joshliebe:
[autoconsent_cache_changes.patch](https://github.com/user-attachments/files/22805618/autoconsent_cache_changes.patch)

@sammacbeth Here’s a patch with the changes I requested, just double-check it but it should be gtg

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/RealAutoconsent.kt
```diff
@@ -49,10 +55,20 @@ class RealAutoconsent @Inject constructor(
...
+    @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
+    private val dispatcherProvider: DispatcherProvider,
+    @IsMainProcess isMainProcess: Boolean,
 ) : Autoconsent, PrivacyConfigCallbackPlugin {
 
     private lateinit var autoconsentJs: String
 
+    init {
+        if (isMainProcess) {
+            onPrivacyConfigDownloaded()
```
joshliebe: Do we need to call `settingsRepository.invalidateCache()` on init? If not, extract the coroutine into a function and just call that

PR #6914, opened 2025-10-09
PIR: Handle jobs after profile edit

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/481882893211075/task/1210822763676216?focus=true

### Description
Adds scan and opt-out jobs handling after a profile is edited. This includes:

- removing old scan, opt-out and email jobs if a profile query is deleted
- removing old scan jobs for deprecated profile queries for which no extracted profiles were found
- allowing running confirm opt-out scans on deprecated profile queries
- allowing running opt-outs on deprecated profile queries
- marking opt-outs as removed and deprecated once confirm opt-out scan shows no results for a deprecated profile query
- marking scans as deprecated once opt out is completed for a deprecated profile query

Note: check [1st commit](https://github.com/duckduckgo/Android/pull/6914/commits/eec09a6402723b9a856903805beed712596a112d) for code and logic changes, [2nd commit](https://github.com/duckduckgo/Android/pull/6914/commits/d0cedc31d5a0414950889f63630dd22d229e577f) only updates tests (new tests will be added in next PR/task) 

### Steps to test this PR
https://app.asana.com/1/137249556945/project/481882893211075/task/1211596475903593?focus=true

### UI changes
No UI changes

### DIFF
--- a/pir/pir-impl/schemas/com.duckduckgo.pir.impl.store.PirDatabase/14.json
+++ b/pir/pir-impl/schemas/com.duckduckgo.pir.impl.store.PirDatabase/14.json
@@ -1,0 +1,940 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 14,
+    "identityHash": "70b508e46c00ddc98bd16621676997ec",
+    "entities": [
+      {
+        "tableName": "pir_broker_json_etag",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`fileName` TEXT NOT NULL, `etag` TEXT NOT NULL, PRIMARY KEY(`fileName`))",
+        "fields": [
+          {
+            "fieldPath": "fileName",
+            "columnName": "fileName",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "etag",
+            "columnName": "etag",
+            "affinity": "TEXT",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "fileName"
+          ]
+        },
+        "indices": [],
+        "foreignKeys": []
+      },
+      {
+        "tableName": "pir_broker_details",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`name` TEXT NOT NULL, `fileName` TEXT NOT NULL, `url` TEXT NOT NULL, `version` TEXT NOT NULL, `parent` TEXT, `addedDatetime` INTEGER NOT NULL, `removedAt` INTEGER NOT NULL, PRIMARY KEY(`name`))",
+        "fields": [
+          {
+            "fieldPath": "name",
+            "columnName": "name",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "fileName",
+            "columnName": "fileName",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "url",
+            "columnName": "url",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "version",
+            "columnName": "version",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "parent",
+            "columnName": "parent",
+            "affinity": "TEXT",
+            "notNull": false
+          },
+          {
+            "fieldPath": "addedDatetime",
+            "columnName": "addedDatetime",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "removedAt",
+            "columnName": "removedAt",
+            "affinity": "INTEGER",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "name"
+          ]
+        },
+        "indices": [],
+        "foreignKeys": []
+      },
+      {
+        "tableName": "pir_broker_opt_out",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`brokerName` TEXT NOT NULL, `stepsJson` TEXT NOT NULL, `optOutUrl` TEXT, PRIMARY KEY(`brokerName`), FOREIGN KEY(`brokerName`) REFERENCES `pir_broker_details`(`name`) ON UPDATE NO ACTION ON DELETE CASCADE )",
+        "fields": [
+          {
+            "fieldPath": "brokerName",
+            "columnName": "brokerName",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "stepsJson",
+            "columnName": "stepsJson",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "optOutUrl",
+            "columnName": "optOutUrl",
+            "affinity": "TEXT",
+            "notNull": false
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "brokerName"
+          ]
+        },
+        "indices": [],
+        "foreignKeys": [
+          {
+            "table": "pir_broker_details",
+            "onDelete": "CASCADE",
+            "onUpdate": "NO ACTION",
+            "columns": [
+              "brokerName"
+            ],
+            "referencedColumns": [
+              "name"
+            ]
+          }
+        ]
+      },
+      {
+        "tableName": "pir_broker_scan",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`brokerName` TEXT NOT NULL, `stepsJson` TEXT, PRIMARY KEY(`brokerName`), FOREIGN KEY(`brokerName`) REFERENCES `pir_broker_details`(`name`) ON UPDATE NO ACTION ON DELETE CASCADE )",
+        "fields": [
+          {
+            "fieldPath": "brokerName",
+            "columnName": "brokerName",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "stepsJson",
+            "columnName": "stepsJson",
+            "affinity": "TEXT",
+            "notNull": false
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "brokerName"
+          ]
+        },
+        "indices": [],
+        "foreignKeys": [
+          {
+            "table": "pir_broker_details",
+            "onDelete": "CASCADE",
+            "onUpdate": "NO ACTION",
+            "columns": [
+              "brokerName"
+            ],
+            "referencedColumns": [
+              "name"
+            ]
+          }
+        ]
+      },
+      {
+        "tableName": "pir_broker_scheduling_config",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`brokerName` TEXT NOT NULL, `retryError` INTEGER NOT NULL, `confirmOptOutScan` INTEGER NOT NULL, `maintenanceScan` INTEGER NOT NULL, `maxAttempts` INTEGER, PRIMARY KEY(`brokerName`), FOREIGN KEY(`brokerName`) REFERENCES `pir_broker_details`(`name`) ON UPDATE NO ACTION ON DELETE CASCADE )",
+        "fields": [
+          {
+            "fieldPath": "brokerName",
+            "columnName": "brokerName",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "retryError",
+            "columnName": "retryError",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "confirmOptOutScan",
+            "columnName": "confirmOptOutScan",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "maintenanceScan",
+            "columnName": "maintenanceScan",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "maxAttempts",
+            "columnName": "maxAttempts",
+            "affinity": "INTEGER",
+            "notNull": false
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "brokerName"
+          ]
+        },
+        "indices": [],
+        "foreignKeys": [
+          {
+            "table": "pir_broker_details",
+            "onDelete": "CASCADE",
+            "onUpdate": "NO ACTION",
+            "columns": [
+              "brokerName"
+            ],
+            "referencedColumns": [
+              "name"
+            ]
+          }
+        ]
+      },
+      {
+        "tableName": "pir_user_profile",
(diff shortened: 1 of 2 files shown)


landomen:
* **#6925** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6925?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6914** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6914?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6914?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/models/scheduling/JobRecord.kt
```diff
@@ -52,6 +54,7 @@ sealed class JobRecord(
         val lastOptOutAttemptDateInMillis: Long = 0L,
         val optOutRequestedDateInMillis: Long = 0L,
         val optOutRemovedDateInMillis: Long = 0L,
+        val deprecated: Boolean = false,
```
landomen, line 57: Context: the tech spec mentioned setting the status to INVALID once the opt-out is confirmed (extracted profile was removed). My thinking for having a separate flag instead is that it's better to keep the original status of the job record (REMOVED or later REMOVED_BY_USER)

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
```diff
@@ -392,6 +403,18 @@ internal class RealPirRepository(
...
         withContext(dispatcherProvider.io()) {
+            if (extractedProfiles.isEmpty()) {
+                return@withContext
+            }
+
+            val profileQueryId = extractedProfiles.first().profileQueryId
+            val profileQuery = userProfileDao.getUserProfile(profileQueryId)
+            if (profileQuery?.deprecated == true) {
+                // we should not store any new extracted profiles for a deprecated user profile
+                // also don't mark them as deprecated as we still want to show them on the UI
```
landomen, line 456: Context: `ExtractedProfiles` are not marked as deprecated as we still want to show them on the UI which currently filters out deprecated ones. If preferred, we can mark them as deprecated and remove that filtering

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
```diff
@@ -318,6 +385,9 @@ class RealJobRecordUpdater @Inject constructor(
                         .copy(
                             status = OptOutJobStatus.REMOVED,
                             optOutRemovedDateInMillis = removedDateInMillis,
+                            // we've confirmed that the extracted profile for a deprecated profile query has been removed
```
karlenDimla, line 395: Quick question: What is now the difference between `deprecated` and having the status `INVALID`. `INVALID` has been designed (as per the documentation) to signal that: _The job is now invalid and should NOT be executed anymore._  which is similar to what deprecated seems to be for _Whether this job record is deprecated and should not be processed anymore._?

karlenDimla: Ah okay just saw this comment https://github.com/duckduckgo/Android/pull/6914#discussion_r2416958179.

Taking a step back, so when the profile has been removed, we want to continue running the opt-out attempts until it has been `REMOVED`. Is my understanding correct that what we want to avoid is for the `extractedProfile` to be picked up again for `SCAN`(maintenance scan) and OPT-OUT re-request?

If that is the case, i think it will be better if we can be clear with that in the documentation. Either we remove INVALID - if there is no INVALID state after all or be explicit with the expectation on how deprecated profiles are handled. I think having this 2 around with similar documentation is a recipe for confusion.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
```diff
@@ -209,6 +211,22 @@ interface JobRecordUpdater {
...
+     * This method should be called when a [ProfileQuery] is deleted or set to deprecated, to ensure that
+     * no stale job records remain in the system or get picked up. However, we still want to retain job records
+     * for brokers that have associated [ExtractedProfile] instances, as those profiles still need to be managed.
+     *
+     * @param profileQueryId The ID of the [ProfileQuery] whose associated job records should be removed.
+     * @param brokersToExclude List of broker names for which records should not be deleted. Job records associated with these brokers will be retained.
+     */
+    suspend fun removeJobRecordsForProfile(
+        profileQueryId: Long,
+        brokersToExclude: List<String>,
```
karlenDimla: Is the idea here we will keep scan, opt-out and email jobs for extractedProfiles even if their profilequery has been removed/deprecated?

Wouldn’t it work to just remove all scan jobs for the profile and not mark the optout / email jobs as deprecated / whatever we call them?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/models/scheduling/JobRecord.kt
```diff
@@ -81,12 +84,15 @@ sealed class JobRecord(
...
      * scan job has been completed.
+     * @property deprecated Whether this job record is deprecated and should not be processed anymore.
+     * This can happen when it is linked to a deprecated [ProfileQuery]
      */
     data class ScanJobRecord(
         override val brokerName: String,
         override val userProfileId: Long,
         val status: ScanJobStatus = ScanJobStatus.NOT_EXECUTED,
         val lastScanDateInMillis: Long = 0L,
+        val deprecated: Boolean = false,
```
karlenDimla, line 92: Why do we need a deprecated field too for scanjob records?

karlenDimla: From the usages, we already don’t consider INVALID jobs for scan and the other usages of deprecated we also checked for INVALID.
Is it for UI?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
landomen: >Is my understanding correct that what we want to avoid is for the extractedProfile to be picked up again for SCAN(maintenance scan) and OPT-OUT re-request?

Correct, we want to finish the opt-out (request + confirmation scan) and then ignore the jobs. I agree with your point that having both `INVALID` and `deprecated` is confusing and can result in undesired usages in the future. 

I think my preference would be to remove the `INVALID` status and keep `deprecated`, because that way we can preserve the latest status of the job record before it was marked as `deprecated`. That can be `REMOVED` or `REMOVED_BY_USER` in case of opt-out and `MATCHES_FOUND` and `NO_MATCH_FOUND` in case of scan. What do you think?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/models/scheduling/JobRecord.kt
landomen: We mark the ScanJobRecord as deprecated only after the opt-out confirmation scan has confirmed that the extracted profiles for a deprecated ProfileQuery have been removed.

The reason for using a deprecated flag instead of setting the INVALID status is same as in  https://github.com/duckduckgo/Android/pull/6914#discussion_r2419773516, preserving the latest status.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
landomen: >Is the idea here we will keep scan, opt-out and email jobs for extractedProfiles even if their profilequery has been removed/deprecated?

When user deletes a profile and we mark it as deprecated, we also want to delete all job records that don't have an ExtractedProfile associated to it. So if profile scanned broker A and found a result, and scanned broker B and didn't find a result, we will only keep job records for broker A.

>Wouldn’t it work to just remove all scan jobs for the profile and not mark the optout / email jobs as deprecated / whatever we call them?

Hm, yes, you're right, we could simplify this and remove all the scan jobs regardless if there are results, since the confirmation scan is based off the opt-out with status REQUESTED. Two things come to mind for scans that found results:
- If we delete the scan job record that found matches before an opt-out job was created and stored, the opt-out will not run. This could happen if user edits the profile multiple times closely together, I think.
- We lose scan job records and their status (maybe we need them for statistics)

What do you think?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
karlenDimla: Sure. I think removing INVALID is far better than having both. I initially designed INVALID anyway as a way to deprecate jobs. But if we want to keep scanning them and there’s no invalid state, then we should just remove it.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
karlenDimla: > If we delete the scan job record that found matches before an opt-out job was created and stored, the opt-out will not run. This could happen if user edits the profile multiple times closely together, I think.

The flow is we run scans > store `extracted profile` after each scan > once all scan are done, create optout jobs > run optouts.

Now for this scenario, As long as the `extractedProfile` was stored, we will create a `optOutJob` record for it later on. You can check the logic in `RealPirJobsRunner.attemptCreateOptOutJobs`. 

> We lose scan job records and their status (maybe we need them for statistics)

Aren’t you proposing to remove them anyway? If you are worried then just set their status to INVALID. It will be basically the same?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/models/scheduling/JobRecord.kt
karlenDimla: > We mark the ScanJobRecord as deprecated only after the opt-out confirmation scan has confirmed that the extracted profiles for a deprecated ProfileQuery have been removed.

If you look at `RealEligibleScanJobProvider.getAllEligibleScanJobs`, We have `getValidScanJobsFromOptOut`. This means we will run confirmation scans on the broker + profileQuery to confirm if records were remove will run according to schedule. It will be less frequent since we will not do the maintenance scan anymore for the deprecated profile.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
landomen: Alright, I will remove `INVALID` state then and keep the `deprecated` flag. It should (hopefully) not have any side-effects as we're not setting that status anywhere, just reading it.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/models/scheduling/JobRecord.kt
karlenDimla: Thinking more about this, I guess the downside of what i am proposing is that, if the profiles were removed - even if they were not requested (children sites for example), we will not be able to scan them.

Let me think of other ways.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/JobRecordUpdater.kt
landomen: > Now for this scenario, As long as the extractedProfile was stored, we will create a optOutJob record for it later on. You can check the logic in RealPirJobsRunner.attemptCreateOptOutJobs.

You're right, my bad. This is not an issue then.


>Aren’t you proposing to remove them anyway? If you are worried then just set their status to INVALID. It will be basically the same?

There are two scenarios when a user removes a profile:

1. No ExtractedProfiles associated with the Profile -> we delete the profile and any ScanJobRecords since we will not need them anymore (OptOutJobRecords and EmailConfirmationJobRecords don't exist at this point) -> this is fine and safe to delete
2. ExtractedProfiles associated with the Profile -> we mark the profile as deprecated -> we delete any ScanJobRecords belonging to a broker for which no matches were found for that profile -> we keep any ScanJobRecords, OptOutJobRecords and EmailConfirmationJobRecords belonging to a broker that had matches (that's why there's a `brokersToExclude` argument) and we don't mark them as deprecated because we still want to run them -> we run opt-outs normally and once a confirmation scan is created from the opt-out and finds no matches we mark all records as deprecated so they don't run again

This is the current implementation, but my understanding was that you're proposing to always delete the ScanJobRecords even in scenario 2 since they are not strictly needed. I might have misunderstood your proposal though.
But I think we should still keep them as described in 2. because most of the `RealPirJobsRunner.attemptCreateScanJobs` is based on a ScanJobRecord existing from broker and profile combination.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/models/scheduling/JobRecord.kt
landomen: > If you look at RealEligibleScanJobProvider.getAllEligibleScanJobs, We have getValidScanJobsFromOptOut. This means we will run confirmation scans on the broker + profileQuery to confirm if records were remove will run according to schedule. It will be less frequent since we will not do the maintenance scan anymore for the deprecated profile.

Correct, after OptOutJobRecord has state REQUESTED, a confirmation scan will be "created" from that record in `RealEligibleScanJobProvider.isRequestedAndShouldBeConfirmedNow`. "Created" because I think the existing ScanJobRecord get's picked up for that broker + profileQuery combo. At that point the ScanJobRecord is not yet deprecated. Once it runs and finds (or doesn't find) results in `RealPirRunStateHandler.handleBrokerScanActionSucceeded` and we confirm all ExtractedProfiles were removed from broker, the related OptOutJobRecord will be set to REMOVED and marked as deprecated, and the ScanJobRecord will also be marked as deprecated

karlenDimla: Synced on MM and we will just remove the INVALID status.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scheduling/EligibleScanJobProvider.kt
```diff
@@ -99,12 +101,14 @@ class RealEligibleScanJobProvider @Inject constructor(
...
 
-            schedulingConfig != null && (
-                record.isNotYetExecuted() ||
-                    record.isNoMatchAndShouldBeMaintained(schedulingConfig, timeInMillis) ||
-                    record.hasMatchAndShouldBeMaintained(schedulingConfig, timeInMillis) ||
-                    record.isErrorAndShouldBeRetried(schedulingConfig, timeInMillis)
-                )
+            schedulingConfig != null &&
+                // do not pick-up deprecated jobs as they belong to invalid/removed profiles
+                !record.deprecated && (
```
karlenDimla, line 106: Would this mean scans for anything marked as `deprecated` will not be executed anymore? From our discussion, i though we were keeping the ScanJobRecord with `MATCHES_FOUND` because we want to keep scanning them 🤔

landomen: A `ScanJobRecord` is marked as `deprecated` only after the confirmation scan from a `OptOutJobRecord` has confirmed that all records were removed. So this ensures that once the confirmation scan has run and record was marked as deprecated, it won't be picked up again.

PR #6922, opened 2025-10-10
Filter CPM rules in native before sending to JS

sammacbeth:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1163321984198618/task/1211626409836529?focus=true

### Description
Filters out non-relevant site-specific rules before constructing a JSON message to send to the autoconsent script.

- Reduces the JSON message size for the latest CPM ruleset from >700kb to ~50kb (note, we currently using a smaller list on Android, because this is causing crashes).
 - We only partially define the compact rule type, because moshi isn't able to properly parse tuples. This means we have to cast specific values to use them. This should be safe, because the compact rules are typed at build time.

### Steps to test this PR

1. Visit https://privacy-test-pages.site/features/autoconsent/index.html.
2. The reject button should be clicked.

---
  - https://app.asana.com/0/0/1211626409836529

### DIFF
--- a/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/InitMessageHandlerPlugin.kt
+++ b/autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/InitMessageHandlerPlugin.kt
@@ -39,9 +39,6 @@ import kotlinx.coroutines.withContext
 import logcat.logcat
 import javax.inject.Inject
 
-const val MAX_SUPPORTED_RULES_VERSION = 1
-const val MAX_SUPPORTED_STEP_VERSION = 1
-
 @ContributesMultibinding(AppScope::class)
 class InitMessageHandlerPlugin @Inject constructor(
     @AppCoroutineScope val appCoroutineScope: CoroutineScope,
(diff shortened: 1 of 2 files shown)


autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/handlers/InitMessageHandlerPlugin.kt
```diff
@@ -75,7 +84,7 @@ class InitMessageHandlerPlugin @Inject constructor(
                     val detectRetries = 20
                     val disabledCmps = settings.disabledCMPs
                     val config = Config(enabled = true, autoAction, disabledCmps, enablePreHide, detectRetries, enableCosmeticRules = true)
-                    val initResp = InitResp(config = config, rules = AutoconsentRuleset(settings.compactRuleList))
+                    val initResp = InitResp(config = config, rules = filterCompactRules(settings.compactRuleList, url))
```
joshliebe: We should put a kill switch around this just in case

joshliebe:
Looks good, I’ve added a patch for the kill switch
[rule_filtering_kill_switch.patch](https://github.com/user-attachments/files/22881355/rule_filtering_kill_switch.patch)

PR #6927, opened 2025-10-13
Instrument auth v2 token refresh with wide event

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205648422731273/task/1211623402976678?focus=true

### Description

This PR adds instrumentation to auth v2 token refresh operation using a wide event.

### Steps to test this PR

See task

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/SubscriptionsManager.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/SubscriptionsManager.kt
@@ -90,7 +90,6 @@ import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.asStateFlow
-import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onSubscription
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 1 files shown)


lmac012:
* **#6927** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/6927?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/SubscriptionsManager.kt
```diff
@@ -675,54 +677,79 @@ class RealSubscriptionsManager @Inject constructor(
...
+                            throw e
+                        }
+
+                        StoreLoginResult.Failure.Other -> {
+                            tokenRefreshWideEvent.onPlayLoginFailure(signedOut = false, refreshException = e, loginError = "Unknown error")
+                            throw e
+                        }
+                    }
+                } else {
+                    throw e
```
nalcalag, line 748: I think the flow doesn’t get finished in this specific case. Shouldn’t we be calling `tokenRefreshWideEvent.onFailure(e)` here?

lmac012: The whole operation is wrapped with `try/catch`, so we will call `onFailure()` at line 755. It would make sense to call `onFailure()` (or some other method) here if we wanted to capture some extra information related to this specific error scenario, but I don't think we need that now.

PR #6952, opened 2025-10-17
Fix ADS Maestro test

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211669753183553

### Description

The caption is actually there but Maestro doesn't scroll all the way through the list before timing out so I've increased the timeout.

I've added this to other tabs so that when people add more composables we should hopefully not get failures either.

I've also added a new github action that runs the design-system maestro test when anything in the compose package is updated.

### Steps to test this PR

- [ ] Run `maestro test .maestro/ads_preview_flows/1-_design-system-components.yaml`
- [ ] Tests should pass
- [ ] Check new github action workflow that runs when the `compose` package has been updated. 
- [ ] Workflow should post a comment on failure
- [ ] Workflow should run the design-system Maestro test and they should pass

### UI changes
N/A

### DIFF
--- a/design-system/src/main/java/com/duckduckgo/common/ui/compose/text/DaxText.kt
+++ b/design-system/src/main/java/com/duckduckgo/common/ui/compose/text/DaxText.kt
@@ -61,14 +61,6 @@ fun DaxText(
 
 @PreviewLightDark
 @Composable
-private fun DaxTextTitlePreviewTest() {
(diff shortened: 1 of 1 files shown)


mikescamell:
Example of the new workflow passing when I added a Test commit: https://github.com/duckduckgo/Android/actions/runs/18592546456?pr=6952

PR #6983, opened 2025-10-22
Split Omnibar: Omnibar layout refactoring

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1211565479656354

### Description

This PR refactors the `OmnibarLayout` and `SingleOmnibarLayout`.  `OmnibarLayout` was renamed to `LegacyOmnibarLayout` and a new `OmnibarLayout` is a unified, merged version of `OmnibarLayout` and `SingleOmnibarLayout`.

Thew new omnibar is guarded by a `useUnifiedOmnibarLayout` feature flag. To make that work, an `OminbarView` interface needed to be extracted, which is now used within the `Omnibar` class. Depending on the flag value, either the old `LegacyOmnibarLayout ` or the new `OmnibarLayout` is used. The feature flag is enabled by default.

Also, some of the interfaces from `OmnibarLayout` had to be extracted (`StateChange`, `Decoration`) to avoid duplication.

### Steps to test this PR

_FF enabled_
- [x] Verify that onboarding highlighting works
- [x] Verify all buttons work
- [x] Verify tab swiping works
- [x] Verify tracker animations work
- [x] Verify the cookies animations work
- [x] Verify that omnibar hiding on scroll works

_FF disabled_
- [x] Disable the `useUnifiedOmnibarLayout` FF in the internal settings
- [x] Verify that onboarding highlighting works
- [x] Verify all buttons work
- [x] Verify tab swiping works
- [x] Verify tracker animations work
- [x] Verify the cookies animations work
- [x] Verify that omnibar hiding on scroll works

PR #6991, opened 2025-10-23
Split Omnibar: Data type & storage update

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1211565479656353?focus=true

### Description

This PR updates the storage layer and replaces the `OmnibarPosition` -> `OmnibarType`.

New `OmnibarType` enum has the following values:
- `SINGLE_TOP`
- `SINGLE_BOTTOM`
- `SPLIT_TOP`

There's also a new `splitOmnibar` feature flag, although it's unused at the moment.

### Steps to test this PR

_Change position_
- [x] Go to Settings -> Appearance
- [x] Verify the position option shows the correct value
- [x] Tap on the omnibar position option
- [x] Change the position
- [x] Go back to the browser and verify the omnibar position changed

_Omnibar behavior_
- [x] Test that the omnibar hiding works on scroll both when position at the top
- [x] Test thatthe correct browser menu shown when position at the top (navigation buttons are at the top of menu)
- [x] Test that the omnibar hiding works on scroll both when position at the bottom
- [x] Test that the correct browser menu shown when position at the bottom (navigation buttons are at the bottom of menu)

---
  - https://app.asana.com/0/0/1211565479656353

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -149,6 +149,7 @@ import com.duckduckgo.app.browser.omnibar.Omnibar.LogoClickListener
 import com.duckduckgo.app.browser.omnibar.Omnibar.OmnibarTextState
 import com.duckduckgo.app.browser.omnibar.Omnibar.TextListener
 import com.duckduckgo.app.browser.omnibar.Omnibar.ViewMode
+import com.duckduckgo.app.browser.omnibar.OmnibarFeatureRepository
 import com.duckduckgo.app.browser.omnibar.OmnibarItemPressedListener
 import com.duckduckgo.app.browser.omnibar.QueryOrigin
 import com.duckduckgo.app.browser.print.PrintDocumentAdapterFactory
@@ -595,6 +596,9 @@ class BrowserTabFragment :
     @Inject
     lateinit var webViewCompatWrapper: WebViewCompatWrapper
 
+    @Inject
+    lateinit var omnibarFeatureRepository: OmnibarFeatureRepository
+
     /**
      * We use this to monitor whether the user was seeing the in-context Email Protection signup prompt
      * This is needed because the activity stack will be cleared if an external link is opened in our browser
@@ -1018,7 +1022,7 @@ class BrowserTabFragment :
         omnibar = Omnibar(
             omnibarType = settingsDataStore.omnibarType,
             binding = binding,
-            isUnifiedOmnibarEnabled = androidBrowserConfigFeature.useUnifiedOmnibarLayout().isEnabled(),
+            isUnifiedOmnibarEnabled = omnibarFeatureRepository.isUnifiedOmnibarEnabled,
         )
 
         webViewContainer = binding.webViewContainer
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarRepository.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarRepository.kt
@@ -17,39 +17,57 @@
 package com.duckduckgo.app.browser.omnibar
 
 import androidx.lifecycle.LifecycleOwner
+import com.duckduckgo.app.di.AppCoroutineScope
 import com.duckduckgo.app.lifecycle.MainProcessLifecycleObserver
 import com.duckduckgo.app.pixels.remoteconfig.AndroidBrowserConfigFeature
 import com.duckduckgo.app.settings.db.SettingsDataStore
 import com.duckduckgo.browser.ui.omnibar.OmnibarType
+import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesMultibinding
+import dagger.SingleInstanceIn
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 import javax.inject.Inject
 
-/**
- * Observes app lifecycle to ensure that if the Split Omnibar feature is disabled via remote config,
- * and the user had selected the Split Omnibar, we revert to Single Top Omnibar.
- * If the feature is re-enabled, we restore the user's choice.
- */
 @ContributesMultibinding(
     scope = AppScope::class,
     boundType = MainProcessLifecycleObserver::class,
 )
-class SplitOmnibarFallbackObserver @Inject constructor(
+@SingleInstanceIn(AppScope::class)
+class OmnibarFeatureRepository @Inject constructor(
     private val settingsDataStore: SettingsDataStore,
     private val browserFeatures: AndroidBrowserConfigFeature,
+    private val dispatcherProvider: DispatcherProvider,
+    @AppCoroutineScope private val coroutineScope: CoroutineScope,
 ) : MainProcessLifecycleObserver {
-    override fun onStart(owner: LifecycleOwner) {
-        super.onStart(owner)
+    var isUnifiedOmnibarEnabled: Boolean = false
+        private set
 
-        if (settingsDataStore.omnibarType == OmnibarType.SPLIT &&
-            (!browserFeatures.useUnifiedOmnibarLayout().isEnabled() || !browserFeatures.splitOmnibar().isEnabled())
-        ) {
+    var isSplitOmnibarEnabled: Boolean = false
+        private set
+
+    val isSplitOmnibarAvailable: Boolean
+        get() = isUnifiedOmnibarEnabled && isSplitOmnibarEnabled
+
+    override fun onStart(owner: LifecycleOwner) {
+        updateFeatureFlags()
+    }
+
+    fun updateFeatureFlags() {
+        coroutineScope.launch(dispatcherProvider.io()) {
+            isUnifiedOmnibarEnabled = browserFeatures.useUnifiedOmnibarLayout().isEnabled()
+            isSplitOmnibarEnabled = browserFeatures.splitOmnibar().isEnabled()
+
+            resetOmnibarTypeIfNecessary()
+        }
+    }
+
+    private fun resetOmnibarTypeIfNecessary() {
+        if (settingsDataStore.omnibarType == OmnibarType.SPLIT && !isSplitOmnibarAvailable) {
             settingsDataStore.isSplitOmnibarSelected = true
             settingsDataStore.omnibarType = OmnibarType.SINGLE_TOP
-        } else if (settingsDataStore.isSplitOmnibarSelected &&
-            browserFeatures.useUnifiedOmnibarLayout().isEnabled() &&
-            browserFeatures.splitOmnibar().isEnabled()
-        ) {
+        } else if (settingsDataStore.isSplitOmnibarSelected && isSplitOmnibarAvailable) {
             // Restore user's choice if the feature is re-enabled
             settingsDataStore.omnibarType = OmnibarType.SPLIT
             settingsDataStore.isSplitOmnibarSelected = false


0nko:
* **#7083** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7083?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7062** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7062?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7056** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7056?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7045** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7045?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7042** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7042?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7032** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7032?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7002** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7002?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6991** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/6991?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/6991?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6983** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/6983?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/omnibar/LegacyOmnibarLayout.kt
```diff
@@ -654,7 +653,7 @@ open class LegacyOmnibarLayout @JvmOverloads constructor(
         tabsMenu.isVisible = newTransitionState.showTabsMenu
         fireIconMenu.isVisible = newTransitionState.showFireIcon
         browserMenu.isVisible = newTransitionState.showBrowserMenu
-        browserMenuHighlight.isVisible = newTransitionState.showBrowserMenuHighlight
+        browserMenuHighlight.isVisible = viewState.showBrowserMenuHighlight
```
LukasPaczos:
```suggestion
        browserMenuHighlight.isVisible = newTransitionState.showBrowserMenuHighlight
```

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
```diff
@@ -238,17 +242,19 @@ class OmnibarLayout @JvmOverloads constructor(
...
-        val attr = context.theme.obtainStyledAttributes(attrs, R.styleable.OmnibarLayout, defStyle, 0)
-        omnibarPosition = OmnibarPosition.entries[attr.getInt(R.styleable.OmnibarLayout_omnibarPosition, 0)]
-
         AndroidSupportInjection.inject(this)
 
-        renderPosition()
+        omnibarType = settingsDataStore.omnibarType
+
+        val attr = context.theme.obtainStyledAttributes(attrs, R.styleable.OmnibarLayout, defStyle, 0)
+        val isTopPosition = attr.getInt(R.styleable.OmnibarLayout_omnibarPosition, 0) == TOP_POSITION_INDEX
```
LukasPaczos:
```suggestion
        val omnibarType = OmnibarType.entries[attr.getInt(R.styleable.OmnibarLayout_omnibarPosition, 0)]
        val isTopPosition = omnibarType == OmnibarType.SINGLE_TOP || omnibarType == OmnibarType.SPLIT
```

This avoids hardcoding indices.

app/src/main/java/com/duckduckgo/app/browser/omnibar/Omnibar.kt
```diff
@@ -164,13 +164,13 @@ class Omnibar(
         binding.rootView.removeView(binding.omnibarLayoutTop)
         binding.rootView.removeView(binding.omnibarLayoutBottom)
 
-        return when (omnibarPosition) {
-            OmnibarPosition.TOP -> {
+        return when (omnibarType) {
+            OmnibarType.SINGLE_TOP, OmnibarType.SPLIT -> {
```
LukasPaczos: Single omnibar doesn't support `OmnibarType.SPLIT`, should we throw here?

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -347,6 +346,13 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
         super.onCreate(savedInstanceState = newInstanceState, daggerInject = false)
 
+        // safety check: if the omnibar type is SPLIT but the feature is disabled, reset to SINGLE_TOP
+        if (settingsDataStore.omnibarType == OmnibarType.SPLIT &&
+            (!browserFeatures.useUnifiedOmnibarLayout().isEnabled() || !browserFeatures.splitOmnibar().isEnabled())
+        ) {
+            settingsDataStore.omnibarType = OmnibarType.SINGLE_TOP
+        }
```
LukasPaczos: `settingsDataStore` can be mutated at any point in time, from a lot of places, so having this safety check in the `BrowserActivity` only looks fragile.

Maybe consider adding an `OmnibarTypeResolver` or something along these lines to inject into all places that initialize the omnibar, and do the flag checks there?

app/src/main/java/com/duckduckgo/app/browser/omnibar/LegacyOmnibarLayout.kt
0nko: What the.. I already changed this. Thanks!

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
0nko: Thanks, that's a good call.

app/src/main/java/com/duckduckgo/app/browser/omnibar/Omnibar.kt
0nko: That's true, this was my original way to handle a fallback if the feature is reverted.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
0nko: Thanks @LukasPaczos, that's a good point. However, there are a lot of places where this resolver would litter the code. Considering that this is just a temporary measure while the flags exist, I was thinking maybe we could avoid that and move this type reset somewhere higher up.

What do you think of adding a `PrivacyConfigCallbackPlugin` and checking/resetting in `onPrivacyConfigDownloaded()`?

0nko: Maybe using `MainProcessLifecycleObserver` would be better, because `onStart()` would be run every time the app starts, as opposed to just when the config is downloaded. This would take care of internal builds when the feature is toggled off in the dev settings.

LukasPaczos: The downside of this is that if:
1. We release the feature and a number of users enable it.
2. We have to temporarily disable it remotely.
3. We re-enable the feature.
4. All of the the users that previously opted into using the split omnibar would have to go and enable it again.

There's no friction-less way to manage a situation like that, but I _think_ the preference should persist.

What if we added the flag check before returning the value from the `SettingsSharedPreferences`?
```kotlin
override var omnibarType: OmnibarType
    get() {
        val prefValue = OmnibarType.fromString(
            preferences.getString(KEY_OMNIBAR_TYPE, OmnibarType.SINGLE_TOP.typeName) ?: OmnibarType.SINGLE_TOP.typeName,
        )
        return if (prefValue == OmnibarType.SPLIT && browserFeatures.splitOmnibar().isEnabled()) {
            OmnibarType.SPLIT
        } else {
            OmnibarType.SINGLE_TOP
        }
    }
    set(value) = preferences.edit { putString(KEY_OMNIBAR_TYPE, value.typeName) }
```
There could be a concern about synchronous feature flag check. That said, pref access can already go to disk so all responsible callers should already move this off of the main thread 😁 We can also look into caching the flag value.

0nko: I already thought of your suggestion but unfortunately, it creates a dependency cycle. I've [added a flag](https://github.com/duckduckgo/Android/pull/6991/commits/196590066afe6ec41d06da45c6dc9d18e7422535) to remeber if the split omnibar was used when a feature is disabled. It's then used to restore the choice when it's re-enabled.

0nko: @LukasPaczos You can test this fallback behavior in a subsequent PR where you can see the selection in the settings.

app/src/main/java/com/duckduckgo/app/browser/omnibar/SplitOmnibarFallbackObserver.kt
```diff
@@ -0,0 +1,58 @@
...
+        } else if (settingsDataStore.isSplitOmnibarSelected &&
+            browserFeatures.useUnifiedOmnibarLayout().isEnabled() &&
+            browserFeatures.splitOmnibar().isEnabled()
+        ) {
+            // Restore user's choice if the feature is re-enabled
+            settingsDataStore.omnibarType = OmnibarType.SPLIT
+            settingsDataStore.isSplitOmnibarSelected = false
+        }
+    }
+}
```
LukasPaczos: This approach feels a bit backwards to me. As discussed in [this comment](https://github.com/duckduckgo/Android/pull/6991#discussion_r2468646789), whether the split omnibar is used isn’t determined solely by the user preference - it also depends on resolving a feature flag. That means it should be checked every time.

In this case, we effectively need a small "repository" layer in front of the settings data store to handle that logic cleanly. I understand the hesitation to add it since the flag might be short-lived, but I suspect it could stay around for a while during rollout.

My recommendation would still be to introduce this resolver/repository class to keep the code clean, but the current approach also works right now, so I won’t block the PR.

If you decide to move forward with the current version, please just make sure the `isEnabled` feature flag checks don’t run on the main thread.

LukasPaczos (approved):

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
LukasPaczos: Continuing in https://github.com/duckduckgo/Android/pull/6991#discussion_r2478515573.

app/src/main/java/com/duckduckgo/app/browser/omnibar/SplitOmnibarFallbackObserver.kt
0nko: @LukasPaczos I understand the concern. However, the flags are used during omnibar initialization in the `BrowserTabFragment`, so a flag update would not take effect until a new tab is created. In fact, using the live value might create an inconsistent state when one tab might use a split omnibar and the other one would not. So using a "resolver" to get the most up-to-date omnibar type (based on the latest flag value) might create even more issues this omnibar type is used all over the place.

I think the easiest solution is to simply cache the latest known flag values at the start (and change the omnibar type, if needed) and then use them throughout the lifetime of the app. If there's a new config downloaded while the app's running, it will take effect the next time the app restarts.

I added [OmnibarFeatureRepoitory](https://github.com/duckduckgo/Android/pull/6991/commits/cc81f571b8c3b37f609686084e925a563d261348), which loads the flags (on the background thread) when the app starts.

LukasPaczos: > However, the flags are used during omnibar initialization in the BrowserTabFragment, so a flag update would not take effect until a new tab is created. In fact, using the live value might create an inconsistent state when one tab might use a split omnibar and the other one would not

That same issue can occur with the proposed `OmnibarFeatureRepository`. It updates the preference state in `onStart`, so if the user resumes the app after a config change, new tabs will fallback to `top` omnibar while existing tabs will still use the` split` version.

Either way, each fragment already checks for changes to the omnibar position/type in `BrowserTabFragment#onResume`, so they should recover correctly in both cases.

Let's merge and iterate as needed.

PR #7010, opened 2025-10-28
PIR: Add database encryption

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211635200412099?focus=true

### Description
Encrypts the PIR database.

### Steps to test this PR
https://app.asana.com/1/137249556945/project/488551667048375/task/1211766506550756?focus=true

### UI changes
No UI changes

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageKeyStore.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageKeyStore.kt
@@ -16,13 +16,11 @@
 
 package com.duckduckgo.pir.impl.store.secure
 
-import android.content.Context
 import android.content.SharedPreferences
 import androidx.core.content.edit
-import androidx.security.crypto.EncryptedSharedPreferences
-import androidx.security.crypto.MasterKey
 import com.duckduckgo.app.di.AppCoroutineScope
 import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.data.store.api.SharedPreferencesProvider
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesBinding
 import dagger.SingleInstanceIn
@@ -62,9 +60,9 @@ interface PirSecureStorageKeyStore {
     boundType = PirSecureStorageKeyStore::class,
 )
 class RealPirSecureStorageKeyStore @Inject constructor(
-    private val context: Context,
     @AppCoroutineScope private val coroutineScope: CoroutineScope,
     private val dispatcherProvider: DispatcherProvider,
+    private val sharedPreferencesProvider: SharedPreferencesProvider,
 ) : PirSecureStorageKeyStore {
 
     private val mutex: Mutex = Mutex()
@@ -108,14 +106,9 @@ class RealPirSecureStorageKeyStore @Inject constructor(
     private suspend fun createEncryptedPreferences(): SharedPreferences? {
         return try {
             mutex.withLock {
-                EncryptedSharedPreferences.create(
-                    context,
-                    FILENAME,
-                    MasterKey.Builder(context)
-                        .setKeyScheme(MasterKey.KeyScheme.AES256_GCM)
-                        .build(),
-                    EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
-                    EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM,
+                sharedPreferencesProvider.getEncryptedSharedPreferences(
+                    name = FILENAME,
+                    multiprocess = true,
                 )
             }
         } catch (_: Exception) {


landomen:
* **#7011** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/7011?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7010** <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/7010?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/Android/7010?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageDatabaseFactory.kt
```diff
@@ -0,0 +1,98 @@
...
+    private val keyProvider: PirSecureStorageKeyProvider,
+) : PirSecureStorageDatabaseFactory {
+    private var _database: PirDatabase? = null
+
+    private val mutex = Mutex()
+
+    init {
+        logcat { "PIR-DB: Loading the sqlcipher native library" }
+        try {
+            LibraryLoader.loadLibrary(context, "sqlcipher")
```
landomen, line 54: >If this method is called more than once with the same library name, the second and subsequent calls are ignored. ([source](https://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#loadLibrary%28java.lang.String%29))

It should be fine for both Autofill and PIR to call this multiple times

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/di/PirModule.kt
```diff
@@ -47,23 +45,26 @@ import com.duckduckgo.pir.impl.store.db.OptOutResultsDao
...
-        return Room.databaseBuilder(context, PirDatabase::class.java, "pir.db")
-            .enableMultiInstanceInvalidation()
-            .fallbackToDestructiveMigration()
-            .build()
+    fun bindPirDatabase(
+        databaseFactory: PirSecureStorageDatabaseFactory,
+    ): PirDatabase {
+        return runBlocking {
+            databaseFactory.getDatabase()
+        } ?: throw IllegalStateException("Failed to create PIR encrypted database")
```
landomen, line 87: For now, I've added this exception since we need non-nullable instance to provide the DAOs. The pro is that it requires almost no changes in our existing repository setup. The downside is that if database fails to decrypt the app will crash.

Autofill handles this more gracefully, it creates the database inside the repository and uses it there. If database is not available, they just log the error and Autofill is not available. However, they only have a single repository and 2 DAOs, while we have multiple repositories and even more DAOs. Handling the exceptions in the same way in PIR would mean a bit more refactoring and more complexity to add new functions in repositories.

It's probably still worth to do it for the UX, but what do you think @karlenDimla ?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageKeyStore.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    override suspend fun canUseEncryption(): Boolean = withContext(dispatcherProvider.io()) {
+        getEncryptedPreferences() != null
+    }
+
+    private suspend fun getEncryptedPreferences(): SharedPreferences? {
+        return encryptedPreferencesDeferred.await()
+    }
+
+    private suspend fun createEncryptedPreferences(): SharedPreferences? {
```
karlenDimla, line 106: We should use `SharedPreferencesProvider.getEncryptedSharedPreferences` which is an API in the `data-store` module. We most likely need this to be multiprocessor which that api has already support for.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/di/PirModule.kt
karlenDimla: I am not a fan of having logic like this in the dagger module. As you said, it could crash the app / the pir process. 

It is also worth pointing out that not everything we store in the `PirDatabase` today needs to be encrypted. At most we only need: `pir_extracted_profiles`, `pir_user_profile` and `pir_email_confirmation_log` (what we have identified in the triages). Broker jsons, logs/events logging and other job records are fine to stay in a non-encrypted db.

karlenDimla (changes requested):
See comments for more information.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageKeyStore.kt
landomen: @karlenDimla Didn't know about this, will use it instead. I was copying this from Autofill, so do we need to change it in Autofill as well?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/di/PirModule.kt
landomen: >I am not a fan of having logic like this in the dagger module. As you said, it could crash the app / the pir process.

This will be improved in https://app.asana.com/1/137249556945/project/488551667048375/task/1211766964402360?focus=true to handle errors

>Broker jsons, logs/events logging and other job records are fine to stay in a non-encrypted db.

@karlenDimla Are you proposing we have two separate databases? Wouldn't that add additional complexity in fetching data and handling errors (normal database would be available, but encrypted one fails for example)?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageKeyStore.kt
karlenDimla: @landomen I’d rather not touch Autofill for now since that is out of scope of this change. Iirc the Autofill code got added first and then we added the shared one but we just didn’t migrate yet.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/di/PirModule.kt
karlenDimla: We discussed this offline and it seems like we will need to check the availability of the encrypted database in every step of the PIR flow anyway to determine feature availability. Aside from semantics, there’s no clear benefit of having an encrypted database (user specific data) and normal database (broker jsons, jobrecords, events) to store PIR data. We will go ahead with having all data in one db for now and revert in case we notice any issue.

landomen: As discussed on Zoom, we are going to proceed with having a single encrypted database for now. If we notice any performance issues, we can split into two databases.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageKeyStore.kt
landomen: @karlenDimla Updated to use `SharedPreferencesProvider.getEncryptedSharedPreferences`

PR #7019, opened 2025-10-28
Implement caching Toggle.Store and Toggle@enabled() flow API

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1198194956794324/task/1211774033967841?focus=true

### Description
Implement a caching mechanism for the feature toggles and add `enable()` flow API

### Steps to test this PR
Smoke tests feature flags

---

> [!NOTE]
> Introduce a cached Toggle.Store with change listeners and expose a Flow-based enabled() API; wire up dependencies and add tests (including androidTest and perf).
> 
> - **Feature Toggles API**:
>   - **Caching store**: Add `CachedToggleStore` using Guava `LoadingCache` with listener support in `feature-toggles-api/internal/CachedToggleStore.kt`.
>   - **Flow API**: Add `Toggle.enabled(): Flow<Boolean>` and implement via `callbackFlow` listening to `CachedToggleStore` updates in `FeatureToggles.kt`.
>   - **Automatic wrapping**: `FeatureToggles` now wraps provided `Toggle.Store` with `CachedToggleStore` if needed.
> - **Build/Config**:
>   - Add Guava and Kotlin Coroutines to `feature-toggles-api` and globally substitute `com.google.guava:listenablefuture` with `guava:33.4.0-jre` in root `build.gradle`.
>   - Allow Android tests in `feature-toggles-impl`; add androidTest deps, Turbine, and desugaring; add `androidTestsBuild` task.
> - **Tests**:
>   - Extend `FeatureTogglesTest` to use `CachedToggleStore` and verify `enabled()` emissions and multi-observer behavior.
>   - Add `CachedToggleStoreTest` (functional), `CachedToggleStoreListenerTest` (listener lifecycle/robustness), and `CachedToggleStorePerfTest` (instrumented perf).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0e194e4f4be900c30fd347d907614c1ec71ddeb8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/internal/CachedToggleStore.kt
+++ b/feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/internal/CachedToggleStore.kt
@@ -23,13 +23,13 @@ import com.google.common.cache.CacheLoader
 import com.google.common.cache.LoadingCache
 import org.jetbrains.annotations.TestOnly
 import java.util.Optional
+import java.util.concurrent.CopyOnWriteArraySet
 import kotlin.jvm.optionals.getOrNull
 
 class CachedToggleStore constructor(
     private val store: Toggle.Store,
 ) : Toggle.Store {
-    @Volatile
-    private var listener: Listener? = null
+    private val listeners = CopyOnWriteArraySet<Listener>()
 
     private val cache: LoadingCache<String, Optional<State>> =
         CacheBuilder
@@ -52,16 +52,16 @@ class CachedToggleStore constructor(
         // Notify AFTER compute() to avoid deadlocks or re-entrancy into the cache/store.
         // If the store.set() above throws, this never runs (which is what we want).
         // Swallow listener exceptions so they don't break writes.
-        listener?.runCatching { onToggleStored(state) }
+        listeners.forEach { listener -> listener.runCatching { onToggleStored(state) } }
     }
 
     /**
      * Registers a [Listener] to observe changes in toggle states stored by this [CachedToggleStore].
      *
-     * Only a single listener is supported at a time. When a new listener is set, it replaces any
-     * previously registered listener. To avoid memory leaks, callers should always invoke the returned
-     * unsubscribe function when the listener is no longer needed (for example, when the collector
-     * of a [kotlinx.coroutines.flow.callbackFlow] is closed).
+     * Multiple listeners are supported concurrently. Each call adds the provided listener and returns
+     * a function that removes only that listener when invoked. To avoid memory leaks, callers should
+     * always invoke the returned unsubscribe function when the listener is no longer needed (for example,
+     * when the collector of a [kotlinx.coroutines.flow.callbackFlow] is closed).
      *
      * Example usage:
      * ```
@@ -76,14 +76,18 @@ class CachedToggleStore constructor(
      * ```
      *
      * @param listener the [Listener] instance that will receive callbacks for each `set` operation.
+     *        Passing `null` clears all registered listeners.
      * @return a function that removes the listener when invoked. The returned function is safe to call
-     *         multiple times and will only clear the listener if it is the same instance that was
-     *         originally registered.
+     *         multiple times.
      */
     fun setListener(listener: Listener?): () -> Unit {
-        this.listener = listener
+        if (listener == null) {
+            listeners.clear()
+            return {}
+        }
+        listeners.add(listener)
 
-        return { if (this.listener === listener) this.listener = null }
+        return { listeners.remove(listener) }
     }
 
     /**
--- a/feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/api/FeatureTogglesTest.kt
+++ b/feature-toggles/feature-toggles-impl/src/test/java/com/duckduckgo/feature/toggles/api/FeatureTogglesTest.kt
@@ -27,7 +27,9 @@ import com.duckduckgo.feature.toggles.api.Toggle.State
 import com.duckduckgo.feature.toggles.api.Toggle.State.CohortName
 import com.duckduckgo.feature.toggles.api.internal.CachedToggleStore
 import com.duckduckgo.feature.toggles.internal.api.FeatureTogglesCallback
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.test.runTest
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
@@ -140,6 +142,38 @@ class FeatureTogglesTest {
             assertFalse(awaitItem())
             expectNoEvents()
         }
+    }
+
+    @Test
+    fun whenMultipleObserversSubscribeToEnabledThenAllReceiveUpdates() = runTest {
+        val firstObserverReady = CompletableDeferred<Unit>()
+        val secondObserverReady = CompletableDeferred<Unit>()
+        val toggle = feature.enabledByDefault()
+
+        val firstJob = launch {
+            toggle.enabled().test {
+                assertTrue(awaitItem())
+                firstObserverReady.complete(Unit)
+                assertFalse(awaitItem())
+                cancelAndIgnoreRemainingEvents()
+            }
+        }
+
+        val secondJob = launch {
+            toggle.enabled().test {
+                assertTrue(awaitItem())
+                secondObserverReady.complete(Unit)
+                assertFalse(awaitItem())
+                cancelAndIgnoreRemainingEvents()
+            }
+        }
+
+        firstObserverReady.await()
+        secondObserverReady.await()
+        toggle.setRawStoredState(Toggle.State(enable = false))
+
+        firstJob.join()
+        secondJob.join()
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


aitorvs:
* **#7019** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7019?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7019?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -172,6 +177,39 @@ interface Toggle {
...
     suspend fun enroll(): Boolean
 
+    /**
+     * Returns a cold [Flow] of [Boolean] values representing whether this toggle is enabled.
+     *
+     * ### Behavior
+     * - When a collector starts, the current toggle value is emitted immediately.
+     * - Subsequent emissions occur whenever the underlying [store] writes a new [State].
+     * - The flow is cold: a listener is only registered while it is being collected.
+     * - When collection is cancelled or completed, the registered listener is automatically.
```
CrisBarreiro: nit: I think a word is missing here 😅

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -386,6 +424,28 @@ internal class ToggleImpl constructor(
...
+        // emit current value when someone starts collecting
+        trySend(isEnabled())
+
+        val unsubscribe = when (val s = store) {
+            is CachedToggleStore -> {
+                s.setListener(
+                    object : Listener {
+                        override fun onToggleStored(newValue: State) {
+                            // emit value just stored
+                            trySend(isEnabled())
```
CrisBarreiro: Why not just `newValue`?

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -386,6 +424,28 @@ internal class ToggleImpl constructor(
         return enrollInternal()
     }
 
+    override fun enabled(): Flow<Boolean> = callbackFlow {
+        // emit current value when someone starts collecting
+        trySend(isEnabled())
```
CrisBarreiro, line 435: If a coroutine context isn't specified, we'll use the same one that the consumer has specified (which in practice, might be `Main` most of he time. Shoud we do `callbackFlow {...}.flowOn(Dispatchers.IO)`? In practice, if the value is cached (and it should) it might not make much of a difference, but I think it still doesn't hurt to take some load off the main thread regardless

aitorvs: No strong opinions, that also has the issue that -- other than API documentation -- it's hard to make the caller aware of it, it can happen the caller is not and then it tries to refresh the UI in IO, which will crash.
That said, I can add it

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
aitorvs: Because `newValue` is the state value however isEnable() is a computed value that has to check not only the toggle state value but also targets, min version etc

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
aitorvs: ah yeah, done

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
CrisBarreiro: If the caller doesn't specify its own context when calling the coroutine and they use `appCoroutineScope` instead of `lifecycleScope` then yes, it will crash. That being said, I'd rather have the app crashing in that case (which will be very hard to miss during implementation and/or automated testing) than have the app silently run non-UI work in the main thread

The context we set in `flowOn` will only apply to that first `trySend` block, though, as the second one is within a callback, so if we want to also set the context there, we'd need to launch a coroutine

aitorvs: Done

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -386,6 +428,28 @@ internal class ToggleImpl constructor(
...
+
+        val unsubscribe = when (val s = store) {
+            is CachedToggleStore -> {
+                s.setListener(
+                    object : Listener {
+                        override fun onToggleStored(newValue: State) {
+                            // emit value just stored
+                            trySend(isEnabled())
```
CrisBarreiro: we should explicitly set the context here as well for consistency. Since this is called from a callback, it doesn't inherit the one from `flowOn`

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -386,6 +428,28 @@ internal class ToggleImpl constructor(
...
+                        }
+                    },
+                )
+            }
+            else -> { -> Unit }
+        }
+
+        // when flow collection is cancelled/closed, run the unsubscribe to avoid leaking the listener
+        awaitClose { unsubscribe() }
+    }.flowOn(Dispatchers.IO)
```
CrisBarreiro: Can we inject `DispatcherProvider` instead?

CrisBarreiro:
We need to add 

```
configurations.all {
    exclude(group: "com.google.guava", module: "listenablefuture")
}
```

to `privacy-dashboard-impl` as well

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
aitorvs: Done

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
aitorvs: Didn't do it because atm `:feature-toggles-api` is a Java module and `DispatcherProvider` is in `:common-utils` which is Android module. 
I'd rather defer that to later (?)

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
```diff
@@ -386,6 +429,28 @@ internal class ToggleImpl constructor(
         return enrollInternal()
     }
 
+    override fun enabled(): Flow<Boolean> = callbackFlow {
+        // emit current value when someone starts collecting
+        trySend(isEnabled())
```
CrisBarreiro, line 435: Since we are in a coroutine context here, why `trySend` instead of `send`?

CrisBarreiro (approved):

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/FeatureToggles.kt
aitorvs: It's not related to the context. 
I want the flow to drop, not block on emissions -- as callbackFlow is a buffered channel so it can outrun the collector.
Also the listener is called by `CachedToggleStore`, which doesn't know (and should not know) about coroutines and doesn't expect to be blocked or suspended.


In this particular case `send` would work just because we'll likely not have so many emissions that will outrun any collector, but that doesn't make it the best choice vs trySend

CrisBarreiro: No strong opinions since, as you said, in this particular case it's very unlikely we'll have too many emissions

However, since we're launching a new coroutine within the listener, the caller won't know anything about coroutines and/or being blocked or suspended

feature-toggles/feature-toggles-api/src/main/java/com/duckduckgo/feature/toggles/api/internal/CachedToggleStore.kt
```diff
@@ -0,0 +1,110 @@
...
+import com.google.common.cache.LoadingCache
+import org.jetbrains.annotations.TestOnly
+import java.util.Optional
+import kotlin.jvm.optionals.getOrNull
+
+class CachedToggleStore constructor(
+    private val store: Toggle.Store,
+) : Toggle.Store {
+    @Volatile
+    private var listener: Listener? = null
```
CrisBarreiro: Having a single listener dones't allow for multiple observers, even when the API docs say: _"Multiple collectors will each register their own listener instance."_

CrisBarreiro (approved):

PR #7054, opened 2025-11-04
[PIR] Handle DB encryption related errors

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211766964402360?focus=true

### Description
Adds error handling for when PIR encrypted database fails to decrypt and we can't access the data.

### Steps to test this PR
See https://app.asana.com/1/137249556945/task/1211824590968188?focus=true

### UI changes
A new error dialog when opening PIR. See https://app.asana.com/1/137249556945/task/1211824590968188?focus=true

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/pixels/PirPixel.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/pixels/PirPixel.kt
@@ -129,7 +129,7 @@ enum class PirPixel(
 
     PIR_INTERNAL_SECURE_STORAGE_UNAVAILABLE(
         baseName = "pir_internal_secure-storage_unavailable",
-        type = Count,
+        types = setOf(Count, Daily()),
     ), ;
 
     constructor(


pir/pir-api/src/main/java/com/duckduckgo/pir/api/PirFeature.kt
```diff
@@ -24,4 +24,11 @@ interface PirFeature {
      * @return true if the PIR beta is enabled, false otherwise
      */
     suspend fun isPirBetaEnabled(): Boolean
+
+    /**
+     * Runs on the IO thread by default.
+     *
+     * @return true if PIR storage is available, false otherwise
+     */
+    suspend fun isPirStorageAvailable(): Boolean
```
karlenDimla: This is a public API so we need to go through https://app.asana.com/1/137249556945/project/1202561462274611/task/1205060262189411?focus=true.


This new method is exposing too much implementation detail. I don’t think any consuming feature needs to know anything about the storage. Consuming feature just needs to know if the feature is enabled, disabled, unavailable or maybe error.

landomen: Thanks, didn't know it's needed even for features that are not live. Created an API proposal here: https://app.asana.com/1/137249556945/project/1210594645151737/task/1211847357779390?focus=true

karlenDimla: Yeah, because I don’t think once we go live, there will be a point where we would open this proposal since technically they are already in the codebase 😅

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirEventsRepository.kt
```diff
@@ -32,38 +33,47 @@ import com.duckduckgo.pir.impl.store.db.PirEventLog
...
+import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
+import logcat.LogPriority.ERROR
+import logcat.logcat
 import javax.inject.Inject
 
 interface PirEventsRepository {
-    fun getAllEventLogsFlow(): Flow<List<PirEventLog>>
+    suspend fun getAllEventLogsFlow(): Flow<List<PirEventLog>>
```
karlenDimla: Why do we need a suspend when we are already returning a Flow?

landomen: As discussed on MM, replaced with `flow {}` and verified it works as expected.

pir/pir-api/src/main/java/com/duckduckgo/pir/api/PirFeature.kt
landomen: Considering there are already 3 approvals on the API proposal, I've gone ahead and made the changes.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/pixels/PirPixel.kt
```diff
@@ -125,6 +125,11 @@ enum class PirPixel(
     PIR_EMAIL_CONFIRMATION_RUN_COMPLETED(
         baseName = "pir_email-confirmation_completed",
         type = Count,
+    ),
+
+    PIR_INTERNAL_SECURE_STORAGE_UNAVAILABLE(
+        baseName = "pir_internal_secure-storage_unavailable",
+        type = Count,
```
karlenDimla: NIT: We might want a daily pixel to see how many unique users daily experience this issue. I just noticed that we emit this pixel every time we call `prepareDatabase` which can happen a few times a day (in app process or PIR process)

landomen: Added daily pixel in https://github.com/duckduckgo/Android/pull/7054/commits/b79948d1ebf24ea085cb3399f728585eb11ae4c3

PR #7072, opened 2025-11-06
[Android] "What’s New" promo message: Add new message type to show a list view

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211783737656574?focus=true

### Description

Added support for a new "What's New" message type in the remote messaging system. This includes:

1. Added a new `REMOTE_WHATS_NEW_MESSAGE` type to the `MessageCta` class.
2. Added new icon assets for AI features, key import, and radar functionality.
3. Implemented a new `CardsList` content type for displaying feature cards.
4. Added support for `UrlInContext` action type to open URLs in a contextual webview.
5. Created placeholder implementations for handling the new message type and actions.

### Steps to test this PR

- [x] Tests should pass.
- [x] No remote message shown if one containing CardList comes from the server as the feature is not complete.
- [x] A remote message not containing CardList is shown.

See Asana for more details.

### NO UI changes

### DIFF
--- a/remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/newtab/RemoteMessageView.kt
+++ b/remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/newtab/RemoteMessageView.kt
@@ -140,7 +140,7 @@ class RemoteMessageView @JvmOverloads constructor(
             is LaunchScreen -> launchScreen(command.screen, command.payload)
             is SharePromoLinkRMF -> launchSharePromoRMFPageChooser(command.url, command.shareTitle)
             is SubmitUrl -> submitUrl(command.url)
-            is Command.SubmitUrlInContext -> submitUrlInContext(command.url)
+            is Command.SubmitUrlInContext -> submitUrl(command.url)
         }
     }
 
@@ -222,9 +222,6 @@ class RemoteMessageView @JvmOverloads constructor(
 
     private fun submitUrl(url: String) {
         context.startActivity(browserNav.openInCurrentTab(context, url))
-    }
-
-    private fun submitUrlInContext(url: String) {
     }
 }
 


anikiki:
* **#7072** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7072?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7072?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/MessageCta.kt
```diff
@@ -199,5 +201,6 @@ class MessageCta : FrameLayout {
     enum class MessageType {
         REMOTE_MESSAGE,
         REMOTE_PROMO_MESSAGE,
+        REMOTE_LIST_MESSAGE,
```
cmonfortep: Do we want to define this type here? The new message is not supported as MessageCta, I'd expect this class doesn't change.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/MessageCta.kt
```diff
@@ -73,6 +74,7 @@ class MessageCta : FrameLayout {
         when (message.messageType) {
             REMOTE_MESSAGE -> setRemoteMessage(message)
             REMOTE_PROMO_MESSAGE -> setPromoMessage(message)
+            REMOTE_LIST_MESSAGE -> { }
```
cmonfortep: if we drop the enum value, we can remove.

app/src/main/java/com/duckduckgo/app/browser/remotemessage/RemoteMessageMapper.kt
```diff
@@ -71,6 +75,12 @@ fun RemoteMessage.asMessage(isLightModeEnabled: Boolean): Message {
             promoAction = content.actionText,
             messageType = MessageType.REMOTE_PROMO_MESSAGE,
         )
+        is CardsList -> Message(
```
cmonfortep: The mapper here is used to receive a remote message and map it into Message so it's rendered as MessageCTA in the NTP.
However the new type, it's not expected to be a MessageCTA in NTP, so we should not map it here.

Did I miss something and we want to show it somehow in NTP as a normal card?

Let me know if you want to catch up on this. I'd like to know how you are planning to show the modal, that's the place where I would expect we have a similar mapping.

app/src/main/java/com/duckduckgo/app/browser/remotemessage/RemoteMessageMapper.kt
```diff
@@ -89,5 +99,8 @@ private fun Placeholder.drawable(isLightModeEnabled: Boolean): Int {
         } else {
             R.drawable.ic_visual_design_update_artwork_dark
         }
+        IMAGE_AI -> R.drawable.ic_image_ai
```
cmonfortep, line 96: 👍

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/RemoteMessageMapper.kt
```diff
@@ -71,6 +75,12 @@ fun RemoteMessage.asMessage(isLightModeEnabled: Boolean): Message {
             promoAction = content.actionText,
             messageType = MessageType.REMOTE_PROMO_MESSAGE,
         )
+        is CardsList -> Message(
```
cmonfortep: ditto

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/newtab/RemoteMessageView.kt
```diff
@@ -221,6 +222,9 @@ class RemoteMessageView @JvmOverloads constructor(
     private fun submitUrl(url: String) {
         context.startActivity(browserNav.openInCurrentTab(context, url))
     }
+
+    private fun submitUrlInContext(url: String) {
```
cmonfortep: can you leave a TODO if we plan to cover this in another PR?

app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabLegacyPageView.kt
```diff
@@ -217,6 +219,9 @@ class NewTabLegacyPageView @JvmOverloads constructor(
         context.startActivity(browserNav.openInCurrentTab(context, url))
     }
 
+    private fun submitUrlInContext(url: String) {
```
cmonfortep: Can you leave a TODO if we plan to cover this in another PR?

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/JsonActionMappers.kt
```diff
@@ -45,6 +46,19 @@ class UrlActionMapper @Inject constructor() : MessageActionMapperPlugin {
     }
 }
 
+@ContributesMultibinding(
+    AppScope::class,
+)
+class UrlInContextActionMapper @Inject constructor() : MessageActionMapperPlugin {
```
cmonfortep, line 53: Actions are generic types that will be available to all remote messages.

If we add this now, we add support to this action on any remote message, not only cards, however the implementation is not completed (we don't handle the event in the UI).

So we will have few app versions out there, that know how to process this action from the config, but have an empty implementation.

I'd say is preferred if we implement and merge the action when it's complete.
If you prefer not to do it now, but keep the code:
1) remove the DI annotation so you keep the code, but it's not used until you implement it entirely, then inject it
or 2) introduce a FF or some value to make it disabled

for simplicity I would go for option 1.

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/JsonRemoteMessageMapper.kt
cmonfortep, line 1: "Similar" comment to the one I left on the Action mapping.

Right now, we are telling RMF how to process the new type, so after merging, clients will now how to process a config with the new message type. Which means that for this app versions onward, the message is valid.
It's great you already thought about this, and in the UI, we won't show it until is fully supported.
However, it's better to treat it as invalid until it's fully supported.

Quick explanation why, but probably best over zoom:
- app downloads the config
- the process the config and they will show first valid message the user is eligible, stops processing the rest
- when the user interacts or dismisses that message, next time we process the config other messages will be eligible
- if an app version doesn't recognises a message/action/matchingattribute, it skips the message and continues processing other messages

I believe by introducing "how to process" but not "how to render", we are creating an unexpected scenario. We will process message, it will be valid, we won't process other messages but at the same time it won't appear in the UI.

Happy to chat more about this. But overall I think, if we want to start merging this logic, we should FF or similar, so we can enable the logic starting with minAppVersion.

cmonfortep (commented):
Few comments that I think is best if we review them together.

code looks great, it's just to discuss merging strategy around RMF.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/MessageCta.kt
anikiki: Good catch! Removed 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/MessageCta.kt
anikiki: Removed 👍

app/src/main/java/com/duckduckgo/app/browser/remotemessage/RemoteMessageMapper.kt
anikiki: Good catch! Removed 👍

app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabLegacyPageView.kt
anikiki: Not needed for now, will deal with it later. Removed 👍

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/mappers/JsonActionMappers.kt
anikiki: Good catch! Removed the DI annotation 👍

Issue #7097, opened 2025-11-11
[Bug] <handleAppLink ignores shouldHaltWebNavigation for same-domain app links after first trigger>


chrisegener22:
### Describe the bug

In DuckDuckGoAppLinksHandler.handleAppLink, when previousUrl is non-null and isSameOrSubdomain(previousUrl, urlString) is true, the method can launch the app link but always returns false unless the domain is in alwaysTriggerList.

This means:

For the first app link navigation on a domain (when previousUrl is null), handleAppLink returns shouldHaltWebNavigation, so the caller can stop web navigation if requested.

For subsequent navigations on the same (sub)domain, the function can call launchAppLink() but always returns false (unless alwaysTriggerList matches), ignoring the shouldHaltWebNavigation flag completely.

This leads to inconsistent behavior: the first navigation correctly halts the web view according to shouldHaltWebNavigation, but subsequent navigations on the same domain do not, even though the app link is launched again.

Relevant code:

override fun handleAppLink(
    isForMainFrame: Boolean,
    urlString: String,
    appLinksEnabled: Boolean,
    shouldHaltWebNavigation: Boolean,
    launchAppLink: () -> Unit,
): Boolean {
    if (!appLinksEnabled || !isForMainFrame) {
        return false
    }

    previousUrl?.let {
        if (isSameOrSubdomain(it, urlString)) {
            val shouldTrigger = alwaysTriggerList.contains(urlString.extractDomain())
            if (isAUserQuery || !hasTriggeredForDomain || shouldTrigger) {
                previousUrl = urlString
                launchAppLink()
                hasTriggeredForDomain = true
                if (shouldTrigger) return true
            }
            return false        // <- shouldHaltWebNavigation is ignored here
        }
    }

    previousUrl = urlString
    launchAppLink()
    hasTriggeredForDomain = true
    return shouldHaltWebNavigation
}

### How to Reproduce

High-level reproduction using the handler logic:

Ensure:

appLinksEnabled = true

isForMainFrame = true

isAUserQuery = true (or hasTriggeredForDomain = false)

urlString is an app-link-eligible URL on domain example.com, not in alwaysTriggerList.

shouldHaltWebNavigation = true.

First navigation:

previousUrl is null.

Call handleAppLink(..., urlString = "https://example.com/path1", shouldHaltWebNavigation = true, launchAppLink = { /* record call */ }).

handleAppLink:

Skips previousUrl?.let block.

Sets previousUrl = "https://example.com/path1".

[...]

PR #7106, opened 2025-11-12
[Fix Bug] <handleAppLink ignores shouldHaltWebNavigation for same-domain app links after first trigger> #7097

marktech0813:
I found the handler and implemented the fix: same-domain navigations now return shouldHaltWebNavigation when an app link is launched.
- Change made in app/src/main/java/com/duckduckgo/app/browser/applinks/AppLinksHandler.kt
- In the same-domain branch, after launching the app link, the function now returns shouldTrigger || shouldHaltWebNavigation instead of always returning false unless shouldTrigger was true.

Summary
Updated DuckDuckGoAppLinksHandler.handleAppLink to consistently respect shouldHaltWebNavigation on subsequent same-domain app links, while preserving alwaysTriggerList behavior.

Contribution by Gittensor, learn more at https://gittensor.io/


joshliebe:
https://github.com/duckduckgo/Android/issues/7097#issuecomment-3568343552

PR #7144, opened 2025-11-19
RMF: subscriptionFreeTrialActive Matching Attribute

nalcalag:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201807753394693/task/1211809773817997?focus=true

### Description
Adding new `subscriptionFreeTrialActive` matching attribute

### Steps to test this PR

_Pre steps_
- [x] Apply patch from https://app.asana.com/1/137249556945/project/1209991789468715/task/1210448620621729?focus=true
- [x] Set `RemoteMessagingService` API to `https://api.jsonblob.com/019aa2b9-6aaa-7cda-826c-bf0c06ae0db0`

_subscriptionFreeTrialActive_
- [x] Fresh install
- [x] Skip onboarding
- [x] Check Remote Message is not visible
- [x] Purchase a test subscription (Free Trial)
- [x] Go to browser > tap Fire Button > Clear Data
- [x] Wait a few seconds until message is shown
- [x] Wait 3 minutes until Free Trial is renewed to a paid subscription
- [x] Go to browser > tap Fire Button > Clear Data
- [x] Check message has been removed

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/rmf/RMFSubscriptionFreeTrialActiveMatchingAttribute.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/rmf/RMFSubscriptionFreeTrialActiveMatchingAttribute.kt
@@ -21,8 +21,7 @@ import com.duckduckgo.remote.messaging.api.AttributeMatcherPlugin
 import com.duckduckgo.remote.messaging.api.JsonMatchingAttribute
 import com.duckduckgo.remote.messaging.api.JsonToMatchingAttributeMapper
 import com.duckduckgo.remote.messaging.api.MatchingAttribute
-import com.duckduckgo.subscriptions.api.ActiveOfferType
-import com.duckduckgo.subscriptions.impl.SubscriptionsManager
+import com.duckduckgo.subscriptions.impl.repository.AuthRepository
 import com.squareup.anvil.annotations.ContributesMultibinding
 import dagger.SingleInstanceIn
 import javax.inject.Inject
@@ -37,15 +36,12 @@ import javax.inject.Inject
 )
 @SingleInstanceIn(AppScope::class)
 class RMFSubscriptionFreeTrialActiveMatchingAttribute @Inject constructor(
-    private val subscriptionsManager: SubscriptionsManager,
+    private val authRepository: AuthRepository,
 ) : JsonToMatchingAttributeMapper, AttributeMatcherPlugin {
     override suspend fun evaluate(matchingAttribute: MatchingAttribute): Boolean? {
         return when (matchingAttribute) {
             is SubscriptionFreeTrialActiveMatchingAttribute -> {
-                val hasActiveTrial = subscriptionsManager.getSubscription()
-                    ?.activeOffers
-                    ?.any { it == ActiveOfferType.TRIAL } == true
-                hasActiveTrial == matchingAttribute.remoteValue
+                authRepository.isFreeTrialActive() == matchingAttribute.remoteValue
             }
 
             else -> null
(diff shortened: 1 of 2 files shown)


nalcalag:
* **#7144** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7144?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7144?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/rmf/RMFSubscriptionFreeTrialActiveMatchingAttribute.kt
```diff
@@ -0,0 +1,76 @@
...
+    boundType = AttributeMatcherPlugin::class,
+)
+@SingleInstanceIn(AppScope::class)
+class RMFSubscriptionFreeTrialActiveMatchingAttribute @Inject constructor(
+    private val subscriptionsManager: SubscriptionsManager,
+) : JsonToMatchingAttributeMapper, AttributeMatcherPlugin {
+    override suspend fun evaluate(matchingAttribute: MatchingAttribute): Boolean? {
+        return when (matchingAttribute) {
+            is SubscriptionFreeTrialActiveMatchingAttribute -> {
+                val hasActiveTrial = subscriptionsManager.getSubscription()
```
cmonfortep: @nalcalag why not using `suspend fun isFreeTrialActive(): Boolean` from `AuthRepository`. It relies on the value stored in `SubscriptionsDataStore`.

I'm curious because I'm using that same function in other places to check the same.

nalcalag: @cmonfortep I didn’t know that method existed, thanks for pointing it out! I had initially checked `SubscriptionManager`, since I assumed that kind of info would live there rather than in the `AuthRepository`. I was thinking of adding a new method to the Manager to avoid putting business logic outside of it. But now that I see the method is already available, I’ll go ahead and use it. 
That said, I’m still not sure if AuthRepository is the ideal place for it, happy to follow up on that separately if needed. For now, I’ll move forward with the one you suggested 🫡

cmonfortep: @nalcalag I created that one recently because I need to consume that state for a metric.
However, you know better than me subscriptions internals, and even free trials, so I'm happy to follow what you think is more correct here.

I'm not sure what's the mental model between `AuthRepository` vs `SubscriptionManager`. For the `getSubscription()` method, subscriptionManager performs a call to the same method inside `AuthRepository`.
Inspecting `getSubscription()`, they both rely on `subscriptionsDataStore.freeTrialActive`. 

So at the end, everything goes to that `subscriptionsDataStore.freeTrialActive`. Happy to also follow your suggestion or take that offline.

nalcalag: @cmonfortep yes, I totally understand your reasoning as well. `subscriptionsDataStore.freeTrialActive` was originally created to persist offers within the subscription. My thought process was to retrieve the entire subscription from auth, and then handle anything related to it, such as period, IDs, or offers, within the subscription layer. That way, we keep the AuthRepository focused on authentication rather than subscription business logic.

However, since we already persist that specific parameter in the DataStore, it also makes sense to have the method there. So I think your approach is perfectly fine for now. If offers become more complex in the future, we can always move that business logic to the subscription layer later.

**FYI**: I updated the jsonBlob URL in the description, as the previous one was expired

cmonfortep (approved):
LGTM 🤖

PR #7161, opened 2025-11-20
Handle render process crashes in SubscriptionsWebViewActivity

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205648422731273/task/1211524858737620?focus=true

### Description

This PR adds handling of renderer process crashes reported by the `WebView` instance used in `SubscriptionsWebViewActivity`.

### Steps to test this PR

- [x] Build with the patch attached to the task.
- [x] Navigate to duckduckgo.com/pro or launch subscription page from app settings. The app will simulate a `WebView` render process crash 10 seconds after opening the page. 
- [x] Verify that the first crash causes the subscriptions screen to restart and second crash closes it and returns to app settings.
- [ ] Verify that subscription_webview_render_process_crash_* pixels are sent.
- [ ] Go to FF inventory and disable `handleSubscriptionsWebViewRenderProcessCrash`
- [ ] Navigate to the subscription page again.
- [ ] Verify that the app crashes 10 seconds after entering that screen and subscription_webview_render_process_crash_* pixels are not sent.

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/ui/SubscriptionsWebViewActivity.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/ui/SubscriptionsWebViewActivity.kt
@@ -655,8 +655,18 @@ class SubscriptionsWebViewActivity : DuckDuckGoActivity(), DownloadConfirmationD
 
     override fun onDestroy() {
         downloadMessagesJob.cancel()
+        destroyWebView()
         super.onDestroy()
     }
+
+    private fun destroyWebView() {
+        binding.webview.stopLoading()
+        binding.webview.removeJavascriptInterface(subscriptionJsMessaging.context)
+        binding.webview.removeJavascriptInterface(itrJsMessaging.context)
+        binding.root.removeView(binding.webview)
+        binding.webview.destroy()
+    }
+
     companion object {
         private const val DOWNLOAD_CONFIRMATION_TAG = "DOWNLOAD_CONFIRMATION_TAG"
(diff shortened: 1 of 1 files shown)


lmac012:
* **#7161** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/ui/SubscriptionsWebViewActivity.kt
nalcalag, line 1: While reviewing the implementation, I found a bug where the crash pixel (`subscription_webview_render_process_crash_c`) is sent multiple times for a single crash event. This happens when `SubscriptionsWebViewActivity` is opened and closed multiple times without properly cleaning up the WebView instance.

**Root cause:**
- The WebView is not fully destroyed in onDestroy()
- Old WebViewClient callbacks remain active across activity recreations
- When a render process crash occurs, all previously registered callbacks fire and send duplicate pixels

**Fix**
Add this code in `onDestroy()` callback within `SubscriptionsWebViewActivity` so it destroys the internal state of the existing webview and it is not duplicated.

    binding.webview.apply {
        stopLoading()
        destroy()
    }

lmac012: Good catch! I added a commit that addresses this.

PR #7193, opened 2025-11-25
Create databaseProvider

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1212086811371661?focus=true

### Description
* Create DatabaseProvider. See https://github.com/duckduckgo/Android/pull/7194 for an integration example
* See https://app.asana.com/1/137249556945/project/414730916066338/task/1210288567313601 for proposal

### Steps to test this PR
n/a

### UI changes
| Before  | After |
| ------ | ----- |
n/a

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,5 @@
+{
+  "disabledMcpjsonServers": [
+    "doc-bot"
+  ]
+}
--- a/data-store/data-store-impl/src/test/java/com/duckduckgo/data/store/impl/RealDatabaseExecutorProviderTest.kt
+++ b/data-store/data-store-impl/src/test/java/com/duckduckgo/data/store/impl/RealDatabaseExecutorProviderTest.kt
@@ -12,7 +12,6 @@ import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
 import org.junit.Assert.assertSame
 import org.junit.Assert.assertTrue
-import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import java.util.concurrent.Executor
@@ -25,14 +24,9 @@ class RealDatabaseExecutorProviderTest {
     private val databaseProviderFeature: DatabaseProviderFeature = FakeFeatureToggleFactory.create(DatabaseProviderFeature::class.java)
     private lateinit var subject: RealDatabaseExecutorProvider
 
-    @Before
-    fun setUp() {
-        subject = RealDatabaseExecutorProvider(databaseProviderFeature)
-    }
-
     @Test
     fun whenFeatureFlagEnabledAndDefaultExecutorThenSameExecutorInstanceIsReturned() = runTest {
-        databaseProviderFeature.self().setRawStoredState(State(enable = true))
+        prepareSubject(true)
 
         val queryExecutor1 = subject.createQueryExecutor(DatabaseExecutor.Default)
         val queryExecutor2 = subject.createQueryExecutor(DatabaseExecutor.Default)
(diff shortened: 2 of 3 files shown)


CrisBarreiro:
* **#7193** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7193?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 24  dependent PRs ([#7194](https://github.com/duckduckgo/Android/pull/7194) <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7194?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7195](https://github.com/duckduckgo/Android/pull/7195) <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7195?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7197](https://github.com/duckduckgo/Android/pull/7197) <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7197?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 21 others) 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7193?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

data-store/data-store-impl/src/main/java/com/duckduckgo/data/store/impl/DatabaseExecutorProvider.kt
```diff
@@ -0,0 +1,107 @@
...
+        ThreadPoolExecutor(
+            defaultPoolSize,
+            defaultPoolSize,
+            0L,
+            TimeUnit.SECONDS,
+            ArrayBlockingQueue(defaultPoolSize * 2),
+            ThreadPoolExecutor.CallerRunsPolicy(),
+        )
+
+    private val isFeatureFlagEnabled = databaseProviderFeature.self().isEnabled()
```
CrisBarreiro, line 38: Using a feature flag is useful in terms of being able to default to previous values for executors, but it introduces other issues on its own:
1. Evaluating the RC flag adds computational overhead
2. RC flags are also stored in a database, which introduces a circular dependency if said database uses `DatabaseProvider`

karlenDimla: This will be evaluated in the main thread right? Will that be an issue?

data-store/data-store-impl/src/main/java/com/duckduckgo/data/store/impl/DatabaseExecutorProvider.kt
```diff
@@ -0,0 +1,107 @@
...
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+interface DatabaseExecutorProvider {
+    fun createQueryExecutor(executor: DatabaseExecutor): Executor?
+    fun createTransactionExecutor(executor: DatabaseExecutor): Executor?
+}
+
+@SingleInstanceIn(AppScope::class)
+@ContributesBinding(AppScope::class)
```
karlenDimla, line 22: Is this intentionally SingleInstance? I don’t have much context and knowledge on the change but is the intention we share the same `defaultExecutor` for every db when the flag is enabled?

CrisBarreiro: Yes, according to the approved proposal:

> Allow databases to define how many threads they want to use for their executor, but provide a default executor shared amongst all databases as a default parameter if no value specified. This way, we gain control over threading (not using Room's default thread pool), while allowing databases to define their own executors if they need to perform critical operations (as is the case for MaliciousSitesDatabase at the moment)

For context, Room's default is to use a pool of 4 threads shared with all Android libs (and all databases except MSP), so a shared executor with 6 threads only used for our databases is already an improvement over that (and there's of course the option for databases to define their own custom executor if needed)

data-store/data-store-impl/src/main/java/com/duckduckgo/data/store/impl/DatabaseExecutorProvider.kt
CrisBarreiro: Yes. To minimize the impact, I'm only evaluating it once (the first time it's used, as `DatabaseExecutorProvider` is injected using `Lazy`. If we want to keep the feature flag, I don't think there's much else we could do, though

PR #7208, opened 2025-11-25
Fix: IllegalArgumentException - New WebView for popup error

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1210754142557620?focus=true

### Description
When tab swiping s enabled, opening popup windows  sometimes fails with:

```
java.lang.IllegalArgumentException: New WebView for popup window must not have been previously navigated
```

#### Root Cause Hypothesis 

There was a race condition in how popup window messages were delivered to fragments when using ViewPager2:

**Original Flow (Broken):**

  - ViewPager2 creates fragments asynchronously and in unpredictable order
  - The message was stored for "the next fragment to be created" (any fragment)
  - If ViewPager2 created a different tab's fragment first, that fragment would incorrectly consume the message
  - The actual popup tab fragment would be created without the message
  - OR when a message tried to be processed, the wrong fragment's WebView may have been navigated → Chromium error

**Solution**

1. [Associate Messages with Specific Source Tab IDs](https://github.com/duckduckgo/Android/compare/fix/ondrej/popup-webview-error?expand=1#diff-fdd33774c79195cc491b6a1e3b6c07cb121c5b1770d3e7f8fed37b371feaa6eeR36)

  - When a popup tab is created with openNewTab(sourceTabId), it gets sourceTabId stored in the tab entity
  - Message is stored keyed by that same sourceTabId
  - Fragment creation looks up the message using tab.sourceTabId
  - Only the correct popup fragment can retrieve the message

2. [Process Message Before Any Navigation](https://github.com/duckduckgo/Android/compare/fix/ondrej/popup-webview-error?expand=1#diff-0310658ba75e79d5b076e4d56fdb249b9541585565310f7722222a91514ba6b9L1059-L1063)

  - Guarantees Chromium gets a completely untouched WebView for the popup transport
  - Defensive against future changes to onViewReady() that might touch the WebView
  - No dependencies - processMessage() doesn't need anything from onViewReady()

### Steps to test this PR

I don't have steps to reproduce the issue.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/TabPagerAdapter.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/TabPagerAdapter.kt
@@ -20,25 +20,29 @@ import android.annotation.SuppressLint
 import android.os.Bundle
 import android.os.Message
 import androidx.fragment.app.Fragment
+import androidx.lifecycle.lifecycleScope
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListUpdateCallback
 import androidx.recyclerview.widget.RecyclerView
 import com.duckduckgo.app.browser.BrowserActivity
 import com.duckduckgo.app.browser.BrowserTabFragment
 import com.duckduckgo.app.browser.tabs.TabManager.TabModel
-import com.duckduckgo.common.utils.DispatcherProvider
-import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 
 class TabPagerAdapter(
     private val activity: BrowserActivity,
-    private val dispatcherProvider: DispatcherProvider,
 ) : FragmentStateAdapter(activity) {
     private val tabs = mutableListOf<TabModel>()
 
-    // Key is the source tab ID, value is the message for the popup
-    private val messagesForTabs = mutableMapOf<String, Message>()
+    private data class PendingMessage(
+        val message: Message,
+        val cleanupJob: Job,
+    )
+
+    // Key is the source tab ID, value contains the message and its cleanup job
+    private val pendingMessages = mutableMapOf<String, PendingMessage>()
 
     var currentTabIndex = -1
         @SuppressLint("NotifyDataSetChanged")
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/TabPagerAdapter.kt
```diff
@@ -31,7 +31,9 @@ class TabPagerAdapter(
     private val activity: BrowserActivity,
 ) : FragmentStateAdapter(activity) {
     private val tabs = mutableListOf<TabModel>()
-    private var messageForNewFragment: Message? = null
+
+    // Key is the source tab ID, value is the message for the popup
+    private val messagesForTabs = mutableMapOf<String, Message>()
```
anikiki: My worry is that `messagesForTabs` can grow in cases where an entry is added and something fails before line https://github.com/duckduckgo/Android/pull/7208/files#diff-fdd33774c79195cc491b6a1e3b6c07cb121c5b1770d3e7f8fed37b371feaa6eeR66 so the entry is not removed.

While testing (aggressively opening new tabs from source) I managed to end up with an entry that was never removed, even after all tabs were closed.

0nko: Thanks! That's a good observation. I'll think about how to manage this.

0nko: @anikiki I added a [clean up mechanism](https://github.com/duckduckgo/Android/pull/7208/commits/425f9d374a7fe583755728e06dcd3fcb969a5dfd) that will remove unpicked messages after a 10s timeout.

app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/TabPagerAdapter.kt
```diff
@@ -88,9 +92,18 @@ class TabPagerAdapter(
...
+     * The message will be automatically cleared after 10 seconds if not picked up.
      */
     fun setMessageForNewFragment(sourceTabId: String, message: Message) {
         messagesForTabs[sourceTabId] = message
+        scheduleMessageCleanup(sourceTabId)
+    }
+
+    private fun scheduleMessageCleanup(sourceTabId: String) {
+        CoroutineScope(Dispatchers.Main).launch {
+            delay(10_000L)
```
anikiki, line 109: I think we can leave the code as if was before this change.

PR #7227, opened 2025-11-26
Make Shorter URL option the default for all new users

GerardPaligot:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1212071899795122?focus=true

### Description

Make the Shorter URL option the default for all new users

### Steps to test this PR

_Scenario 1: Fresh install - New user gets shorter URL_
- [x] Uninstall DuckDuckGo completely from device
- [x] Install the app with feature enabled
- [x] Complete onboarding
- [x] Navigate to any website
- [x] Verify URL bar shows shorter format
- [x] Go to Settings → Appearance
- [x] Verify toggle is OFF

_Scenario 2: Fresh install - Feature disabled (rollback)_
- [x] Uninstall DuckDuckGo completely from device
- [x] Install the app with feature disabled
- [x] Complete onboarding
- [x] Navigate to any website
- [x] Verify URL bar shows full URL
- [x] Go to Settings → Appearance
- [x] Verify toggle is ON

_Scenario 3: Existing user update - Preserves full URL_
- [x] Install old version of DuckDuckGo (before this feature)
- [x] Use the app normally (preference implicitly set to full URL)
- [x] Update to new version with feature enabled
- [x] Navigate to any website
- [x] Verify URL bar still shows full URL (legacy behavior preserved)
- [x] Go to Settings → Appearance
- [x] Verify toggle is ON

_Scenario 4: Manual toggle - User changes to shorter URL_
- [x] Start with app showing full URL (any scenario above)
- [x] Go to Settings → Appearance
- [x] Turn OFF toggle
- [x] Return to browser
- [x] Navigate to a website
- [x] Verify URL bar shows shorter format
- [x] Restart app
- [x] Verify preference persists (still shows shorter URL)

_Scenario 5: Manual toggle - User changes to full URL_
- [x] Start with app showing shorter URL (Scenario 1)
- [x] Go to Settings → Appearance
- [x] Turn ON toggle
- [x] Return to browser
- [x] Navigate to a website
- [x] Verify URL bar shows full URL
- [x] Restart app
- [x] Verify preference persists (still shows full URL)

_Scenario 6: Rollback protection - Manual preference survives_
- [x] Start with fresh install and feature enabled (Scenario 1)
- [x] Manually change to full URL
- [x] Update to new version with feature disabled (rollback)
- [x] Navigate to a website
- [x] Verify URL bar still shows full URL (manual choice protected)
- [x] Go to Settings → Appearance
- [x] Verify toggle is still ON (unchanged)

_Scenario 7: Rollback effect - Auto-assigned users affected_
- [x] Start with fresh install and feature enabled (Scenario 1)
- [x] Don't manually change the preference
- [x] Navigate to websites, verify shorter URL works
- [x] Update to new version with feature disabled (rollback)
- [x] Navigate to a website
- [x] Verify URL bar now shows full URL (rollback worked)
- [x] Go to Settings → Appearance
- [x] Verify toggle switched to ON (auto-updated by rollback)

_Scenario 8: Old app user migration - Preference Preserved_
- [x] Install old version with full URL manually set by user
- [x] Verify URL shows full format in old version
- [x] Update to new version with feature enabled
- [x] Navigate to a website
- [x] Verify URL bar still shows full URL (old choice preserved)
- [x] Update to new version with feature disabled (rollback)
- [x] Navigate again
- [x] Verify URL bar still shows full URL (migrated to manual, protected from rollback)

_Scenario 9: Old app user migration - Shorter URL user_
- [x] Install old version
- [x] Go to Settings → Appearance and turn OFF
- [x] Verify shorter URL works in old version
- [x] Update to new version with feature enabled
- [x] Navigate to a website
- [x] Verify URL bar still shows shorter format (old choice preserved)
- [x] Update to new version with feature disabled (rollback)
- [x] Verify URL bar still shows shorter format (migrated to manual, protected)

### UI changes
| Before  | After |
| ------ | ----- |
| image | image |

---

> [!NOTE]
> Introduce flow-based UrlDisplayRepository to manage full vs shorter URL display, defaulting new installs to shorter URLs, and wire it through browser/omnibar/settings with migration, remote toggle, and new-install detection.
> 
> - **URL Display Preference (new)**:
>   - Add `UrlDisplayRepository` with flow-based `isFullUrlEnabled` and `setFullUrlEnabled`.
>   - Auto-defaults: new installs → shorter URL; existing users preserved; rollback via `androidBrowserConfig.shorterUrlDefault`.
>   - Persist/migrate via new `SettingsDataStore` flags: `urlPreferenceMigrated`, `urlPreferenceSetByUser`, plus `hasUrlPreferenceSet`.
> - **Build Config**:
>   - Add `AppBuildConfig.isNewInstall()` and implement in `RealAppBuildConfig`.
> - **Wiring/Refactors**:
>   - Replace direct `SettingsDataStore.isFullUrlEnabled` reads with `UrlDisplayRepository` in `BrowserTabViewModel`, `OmnibarLayoutViewModel`, `AppearanceViewModel`.
>   - `BrowserTabViewModel` now observes preference changes and issues `RefreshOmnibar`.
>   - `DailyPixelSender` sources daily pixel from repository.
>   - `UserStateReporter` uses `AppBuildConfig.isNewInstall()`.
> - **Remote Config**:
>   - Add `androidBrowserConfig.shorterUrlDefault` toggle.
> - **Tests**:
>   - Add `RealUrlDisplayRepositoryTest` and update affected unit/instrumentation tests to use flow-based preference and new behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 843a8b59d96315a97b600a2bc67c8bad0a81ce3c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/test/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModelTest.kt
+++ b/app/src/test/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModelTest.kt
@@ -44,12 +44,9 @@ import com.duckduckgo.serp.logos.api.SerpEasterEggLogosToggles
 import com.duckduckgo.serp.logos.api.SerpLogo
 import com.duckduckgo.voice.api.VoiceSearchAvailability
 import com.duckduckgo.voice.api.VoiceSearchAvailabilityPixelLogger
-import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.test.runTest
-import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
@@ -89,7 +86,7 @@ class OmnibarLayoutViewModelTest {
     private val duckAiShowOmnibarShortcutOnNtpAndOnFocusFlow = MutableStateFlow(true)
     private val duckAiShowOmnibarShortcutInAllStatesFlow = MutableStateFlow(true)
     private val duckAiShowInputScreenFlow = MutableStateFlow(false)
-    private val isFullUrlEnabledFlow = Channel<Boolean>()
+    private val isFullUrlEnabledFlow = MutableStateFlow(true)
     private val settingsDataStore: SettingsDataStore = mock()
     private val urlDisplayRepository: UrlDisplayRepository = mock()
     private val mockAddressDisplayFormatter: AddressDisplayFormatter by lazy {
@@ -121,7 +118,7 @@ class OmnibarLayoutViewModelTest {
         whenever(duckPlayer.isDuckPlayerUri(DUCK_PLAYER_URL)).thenReturn(true)
         whenever(duckAiFeatureState.showOmnibarShortcutOnNtpAndOnFocus).thenReturn(duckAiShowOmnibarShortcutOnNtpAndOnFocusFlow)
         whenever(duckAiFeatureState.showOmnibarShortcutInAllStates).thenReturn(duckAiShowOmnibarShortcutInAllStatesFlow)
-        whenever(urlDisplayRepository.isFullUrlEnabled).then { isFullUrlEnabledFlow.consumeAsFlow() }
+        whenever(urlDisplayRepository.isFullUrlEnabled).then { isFullUrlEnabledFlow }
         whenever(duckAiFeatureState.showInputScreen).thenReturn(duckAiShowInputScreenFlow)
         whenever(serpEasterEggLogosToggles.feature()).thenReturn(mock())
         whenever(serpEasterEggLogosToggles.feature().isEnabled()).thenReturn(false)
@@ -129,11 +126,6 @@ class OmnibarLayoutViewModelTest {
         whenever(androidBrowserToggles.handleAboutBlank().isEnabled()).thenReturn(false)
 
         initializeViewModel()
-    }
-
-    @After
-    fun after() {
-        isFullUrlEnabledFlow.close()
     }
 
     @Test
@@ -1416,7 +1408,7 @@ class OmnibarLayoutViewModelTest {
 
     @Test
     fun `when input text click catcher clicked and random URL then input screen launched with full URL`() = runTest {
-        isFullUrlEnabledFlow.send(false)
+        isFullUrlEnabledFlow.emit(false)
         initializeViewModel()
         val omnibarViewState = OmnibarViewState(omnibarText = "test", queryOrFullUrl = "test", isEditing = false)
         testee.onExternalStateChange(StateChange.OmnibarStateChange(omnibarViewState))
@@ -1436,7 +1428,7 @@ class OmnibarLayoutViewModelTest {
     fun whenOmnibarLosesFocusAndFullUrlEnabledThenAddressDisplayDisplaysFullUrl() = runTest {
         val query = "query"
         val url = "https://example.com/test.html"
-        isFullUrlEnabledFlow.send(true)
+        isFullUrlEnabledFlow.emit(true)
         initializeViewModel()
         testee.onOmnibarFocusChanged(hasFocus = true, inputFieldText = query) // Initial focus
         testee.onExternalStateChange(StateChange.LoadingStateChange(LoadingViewState(url = url))) // Set URL
@@ -1456,7 +1448,7 @@ class OmnibarLayoutViewModelTest {
         val query = "query"
         val url = "https://example.com/test.html"
         val formattedUrl = "example.com"
-        isFullUrlEnabledFlow.send(false)
+        isFullUrlEnabledFlow.emit(false)
         testee.onOmnibarFocusChanged(hasFocus = true, inputFieldText = query) // Initial focus
         testee.onExternalStateChange(StateChange.LoadingStateChange(LoadingViewState(url = url))) // Set URL
 
@@ -1475,7 +1467,7 @@ class OmnibarLayoutViewModelTest {
         val url = "https://example.com/test.html"
         val formattedUrl = "https://example.com/test.html"
         val omnibarViewState = OmnibarViewState(omnibarText = url, queryOrFullUrl = url, isEditing = false)
-        isFullUrlEnabledFlow.send(true)
+        isFullUrlEnabledFlow.emit(true)
 
         testee.onExternalStateChange(StateChange.OmnibarStateChange(omnibarViewState))
 
@@ -1491,7 +1483,7 @@ class OmnibarLayoutViewModelTest {
     fun whenOmnibarGainsFocusAndFullUrlDisabledThenOmnibarTextIsFullUrl() = runTest {
         val url = "https://example.com/test.html"
         val shortUrl = "example.com"
-        isFullUrlEnabledFlow.send(false)
+        isFullUrlEnabledFlow.emit(false)
 
         // Set initial state: site loaded, omnibar not focused (showing short url)
         givenSiteLoaded(url)
@@ -1511,7 +1503,7 @@ class OmnibarLayoutViewModelTest {
     fun whenOmnibarGainsFocusAndFullUrlEnabledThenOmnibarTextIsFullUrl() = runTest {
         val url = "https://example.com/test.html"
         val shortUrl = "example.com"
-        isFullUrlEnabledFlow.send(true)
+        isFullUrlEnabledFlow.emit(true)
 
         // Set initial state: site loaded, omnibar not focused (showing short url)
         givenSiteLoaded(url)
@@ -1531,7 +1523,7 @@ class OmnibarLayoutViewModelTest {
     @Test
     fun whenExternalOmnibarStateChangedWithForceRenderAndNotDDGUrlAndFullUrlEnabledThenOmnibarTextIsFullUrl() = runTest {
         val omnibarViewState = OmnibarViewState(omnibarText = RANDOM_URL, queryOrFullUrl = RANDOM_URL)
-        isFullUrlEnabledFlow.send(true)
+        isFullUrlEnabledFlow.emit(true)
 
         testee.onExternalStateChange(StateChange.OmnibarStateChange(omnibarViewState, forceRender = true))
 
@@ -1547,7 +1539,7 @@ class OmnibarLayoutViewModelTest {
     fun whenExternalOmnibarStateChangedWithForceRenderAndNotDDGUrlAndFullUrlDisabledThenOmnibarTextIsShortUrl() = runTest {
         val shortUrl = RANDOM_URL.toUri().baseHost!!
         val omnibarViewState = OmnibarViewState(omnibarText = RANDOM_URL, queryOrFullUrl = RANDOM_URL)
-        isFullUrlEnabledFlow.send(false)
+        isFullUrlEnabledFlow.emit(false)
 
         testee.onExternalStateChange(StateChange.OmnibarStateChange(omnibarViewState, forceRender = true))
 
@@ -1563,7 +1555,7 @@ class OmnibarLayoutViewModelTest {
     fun whenExternalOmnibarStateChangedWithForceRenderAndNotDDGUrlAndFullUrlEndabledThenOmnibarTextIsFullUrl() = runTest {
         val shortUrl = RANDOM_URL.toUri().baseHost!!
         val omnibarViewState = OmnibarViewState(omnibarText = shortUrl, queryOrFullUrl = RANDOM_URL)
-        isFullUrlEnabledFlow.send(true)
+        isFullUrlEnabledFlow.emit(true)
 
         testee.onExternalStateChange(StateChange.OmnibarStateChange(omnibarViewState, forceRender = true))
 
@@ -1578,7 +1570,7 @@ class OmnibarLayoutViewModelTest {
     @Test
     fun whenExternalOmnibarStateChangedWithoutForceRenderThenOmnibarTextIsFromState() = runTest {
         val omnibarViewState = OmnibarViewState(omnibarText = RANDOM_URL, queryOrFullUrl = RANDOM_URL)
-        isFullUrlEnabledFlow.send(false)
+        isFullUrlEnabledFlow.emit(false)
 
         testee.onExternalStateChange(StateChange.OmnibarStateChange(omnibarViewState, forceRender = false))
 
--- a/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
@@ -614,8 +614,7 @@ class BrowserTabViewModelTest {
 
     private val selectedTab = TabEntity("TAB_ID", exampleUrl, position = 0, sourceTabId = "TAB_ID_SOURCE")
     private val flowSelectedTab = MutableStateFlow(selectedTab)
-
-    private val isFullSiteAddressEnabled = Channel<Boolean>()
+    private val isFullSiteAddressEnabledFlow = MutableStateFlow(true)
 
     private val mockWebViewCompatWrapper: WebViewCompatWrapper = mock()
 
@@ -682,7 +681,7 @@ class BrowserTabViewModelTest {
             whenever(mockRemoteMessagingRepository.messageFlow()).thenReturn(remoteMessageFlow.consumeAsFlow())
             whenever(mockSettingsDataStore.automaticFireproofSetting).thenReturn(AutomaticFireproofSetting.ASK_EVERY_TIME)
             whenever(mockSettingsDataStore.omnibarType).thenReturn(OmnibarType.SINGLE_TOP)
-            whenever(mockUrlDisplayRepository.isFullUrlEnabled).then { isFullSiteAddressEnabled.consumeAsFlow() }
+            whenever(mockUrlDisplayRepository.isFullUrlEnabled).then { isFullSiteAddressEnabledFlow }
             whenever(mockSSLCertificatesFeature.allowBypass()).thenReturn(mockEnabledToggle)
             whenever(subscriptions.shouldLaunchPrivacyProForUrl(any())).thenReturn(false)
             whenever(mockDuckDuckGoUrlDetector.isDuckDuckGoUrl(any())).thenReturn(false)
@@ -899,7 +898,6 @@ class BrowserTabViewModelTest {
         bookmarksListFlow.close()
         favoriteListFlow.close()
         remoteMessageFlow.close()
-        isFullSiteAddressEnabled.close()
         testee.onCleared()
         db.close()
         testee.command.removeObserver(mockCommandObserver)
@@ -1442,7 +1440,7 @@ class BrowserTabViewModelTest {
 
     @Test
     fun whenBrowsingAndUrlLoadedAndFullUrlDisabledThenUrlTitleAndOmnibarTextUpdatedToMatch() = runTest {
-        isFullSiteAddressEnabled.send(false)
+        isFullSiteAddressEnabledFlow.emit(false)
         val exampleUrl = "http://example.com/abc"
         val exampleTitle = "Title"
         loadUrl(exampleUrl, title = exampleTitle, isBrowserShowing = true)
@@ -1471,7 +1469,7 @@ class BrowserTabViewModelTest {
 
     @Test
     fun whenBrowsingAndUrlIsUpdatedAndFullUrlDisabledThenUrlAndOmnibarTextUpdatedToMatch() = runTest {
-        isFullSiteAddressEnabled.send(false)
+        isFullSiteAddressEnabledFlow.emit(false)
         val currentUrl = "http://example.com/current"
         loadUrl(exampleUrl, isBrowserShowing = true)
         updateUrl(exampleUrl, currentUrl, true)
@@ -1491,7 +1489,7 @@ class BrowserTabViewModelTest {
 
     @Test
     fun whenNotBrowsingAndUrlIsUpdatedAndFullUrlDisabledThenUrlAndOmnibarTextRemainUnchanged() = runTest {
-        isFullSiteAddressEnabled.send(false)
+        isFullSiteAddressEnabledFlow.emit(false)
         val currentUrl = "http://example.com/current"
         loadUrl(exampleUrl, isBrowserShowing = true)
         updateUrl(exampleUrl, currentUrl, false)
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
```diff
@@ -221,4 +221,13 @@ interface AndroidBrowserConfigFeature {
     @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
     @Toggle.InternalAlwaysEnabled
     fun newCustomTab(): Toggle
+
+    /*
```
mikescamell: 🔍We're missing a "*" here to make it a KDoc

app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
```diff
@@ -221,4 +221,13 @@ interface AndroidBrowserConfigFeature {
...
+     * Controls default URL display for new users only.
+     * @return `true` when the remote config has the global "shorterUrlDefault" androidBrowserConfig
+     * sub-feature flag enabled
+     * If the remote feature is not present defaults to `false`
```
mikescamell: ❓This should say "defaults to 'true'". And just to make sure we definitely want this rolling out to everyone in the next release? In this way we'll be using this more as a kill switch rather than way to rollout the feature, right?

app/src/main/java/com/duckduckgo/app/browser/urlDisplay/RealUrlDisplayPlugin.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val settingsDataStore: SettingsDataStore,
+    private val browserConfigFeature: AndroidBrowserConfigFeature,
+    private val context: Context,
+) : PrivacyConfigCallbackPlugin {
+    override fun onPrivacyConfigDownloaded() {
+        // User has explicitly set their preference - always honor it
+        if (settingsDataStore.hasUrlPreferenceSet()) {
+            return
+        }
+        val packageInfo = context.packageManager.getPackageInfo(context.packageName, 0)
```
mikescamell: ❓Should we account for the `NameNotFoundException` this might throw? Maybe this should be in a try catch? What do you think?

app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
GerardPaligot: Yes, in my first implementation, I set the value to false and forgot to update the documentation when I changed to true.
Because we want to make the shorter URL the default option for any new user in the next release, yes we should have a default value to true and switch back to false if we want to deactivate the feature.

app/src/main/java/com/duckduckgo/app/browser/urlDisplay/RealUrlDisplayPlugin.kt
```diff
@@ -0,0 +1,46 @@
...
+        val isNewInstall = packageInfo.firstInstallTime == packageInfo.lastUpdateTime
+        val defaultValue = !(isNewInstall && browserConfigFeature.shorterUrlDefault().isEnabled())
```
mikescamell: 💡I think we should move the isEnabled() check to the io dispatcher, otherwise we might block the main thread. Checkout OmnibarFeatureRepository where we check feature flags in a PrivacyConfigCallbackPlugin for an example

GerardPaligot: Damn yes, Łukasz already informed me about that. We really need to convert `isEnabled` as a suspend function to avoid this kind of error...

app/src/main/java/com/duckduckgo/app/browser/urlDisplay/RealUrlDisplayPlugin.kt
GerardPaligot: We can but it is probably not necessary because we are using `context.packageName` to get package info for our own application and because we're using this API in so many other places without a try/catch.

app/src/main/java/com/duckduckgo/app/browser/urlDisplay/RealUrlDisplayPlugin.kt
```diff
@@ -0,0 +1,54 @@
...
+class RealUrlDisplayPlugin @Inject constructor(
+    private val settingsDataStore: SettingsDataStore,
+    private val browserConfigFeature: AndroidBrowserConfigFeature,
+    private val context: Context,
+    @AppCoroutineScope private val coroutineScope: CoroutineScope,
+    private val dispatcherProvider: DispatcherProvider,
+) : PrivacyConfigCallbackPlugin {
+    override fun onPrivacyConfigDownloaded() {
+        // User has explicitly set their preference - always honor it
+        if (settingsDataStore.hasUrlPreferenceSet()) {
```
mikescamell: 🔍We should probably move this check inside the launch as well, it's not actually using DataStore but SharedPrefs

GerardPaligot: I hesitated. This function isn't a suspend function so I can use it here. I prefer to avoid to create a new coroutine if I can. What do you think?

app/src/test/java/com/duckduckgo/app/browser/urlDisplay/UrlDisplayPluginTest.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.mockito.kotlin.atLeastOnce
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+
+class UrlDisplayPluginTest {
```
mikescamell: 🔍 It might be worth renaming this to RealUrlDisplayPluginTest as we're testing that implementation

app/src/test/java/com/duckduckgo/app/browser/urlDisplay/RealUrlDisplayPluginTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    fun whenUserHasExplicitPreferenceSet_thenNothingChange() = runTest {
+        // Given: User has explicitly set preference to true
+        whenever(settingsDataStore.hasUrlPreferenceSet()).thenReturn(true)
+
+        // When
+        testee.onPrivacyConfigDownloaded()
+
```
mikescamell: ❓ Do you think is worth adding `advanceUntilIdle()` to all the tests to just to be sure the new coroutine we're launching is complete before verifying?

GerardPaligot: I just checked again `UnconfinedTestDispatcher` and we don't have any guarantees about the tread used. I'll add `advanceUntilIdle` everywhere!

app/src/test/java/com/duckduckgo/app/browser/urlDisplay/RealUrlDisplayPluginTest.kt
```diff
@@ -0,0 +1,135 @@
...
+            settingsDataStore = settingsDataStore,
+            browserConfigFeature = browserConfigFeature,
+            context = context,
+            coroutineScope = coroutineRule.testScope,
+            dispatcherProvider = coroutineRule.testDispatcherProvider,
+        )
+    }
+
+    @Test
+    fun whenUserHasExplicitPreferenceSet_thenNothingChange() = runTest {
```
marcosholgado: Can we follow the same patterns as in other tests and avoid `_`? I would just go with `when user blah blah`

malmstein:
@GerardPaligot  nit, rename the PR to match what you’re doing now

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
```diff
@@ -0,0 +1,125 @@
...
+        .distinctUntilChanged()
+        .shareIn(
+            scope = appCoroutineScope,
+            replay = 1,
+            started = SharingStarted.WhileSubscribed(),
+        )
+
+    override suspend fun isFullUrlEnabled(): Boolean {
+        // User manually set preference - honor it
+        if (settingsDataStore.urlPreferenceManuallySet) {
```
malmstein: I’m confused with his. It seems like you are checking the same condition but using different methods. 

    /**
     * Indicates whether the user has manually set their URL display preference.
     * When true, the user's preference should be preserved even if default behavior changes.
     * When false, the preference may be updated if business rules change (via feature flags).
     */
    var urlPreferenceManuallySet: Boolean

and then you also have 

    /**
     * Check if user has explicitly set their URL display preference.
     */
    fun hasUrlPreferenceSet(): Boolean

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
```diff
@@ -0,0 +1,125 @@
...
+            return settingsDataStore.isFullUrlEnabled
+        }
+
+        // Feature flag disabled - rollback mode
+        if (!browserConfigFeature.shorterUrlDefault().isEnabled()) {
+            return true
+        }
+
+        // Auto-assigned preference exists - use it for consistency
+        if (settingsDataStore.hasUrlPreferenceSet()) {
```
malmstein, line 111: ditto, this seems redundant

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -797,6 +804,12 @@ class BrowserTabViewModel @Inject constructor(
         viewModelScope.launch(dispatchers.io()) {
             handleAboutBlankEnabled = androidBrowserConfig.handleAboutBlank().isEnabled()
         }
+
+        isFullUrlEnabled
```
malmstein, line 810: this one needs a test

app/src/main/java/com/duckduckgo/app/userstate/UserStateReporter.kt
```diff
@@ -36,17 +36,14 @@ import javax.inject.Inject
...
 ) : MainProcessLifecycleObserver {
 
     @UiThread
     override fun onCreate(owner: LifecycleOwner) {
         appCoroutineScope.launch(dispatchers.io()) {
-            val firstInstallTime: Long = context.packageManager.getPackageInfo(context.packageName, 0).firstInstallTime
-            val lastUpdateTime: Long = context.packageManager.getPackageInfo(context.packageName, 0).lastUpdateTime
-
-            repository.setIsUserNew(firstInstallTime == lastUpdateTime)
+            repository.setIsUserNew(appBuildConfig.isNewInstall())
```
malmstein, line 46: @0nko can you clean this up? the user state is only used to send a pixel when the user starts dragging tabs. However, we are continuously setting a value on app start. Instead, we could evaluate the condition in the `TabSwitcherViewModel` directly, using the new API here created.

    fun onTabDraggingStarted() {
            val params = mapOf("userState" to tabRepository.tabSwitcherData.first().userState.name)
            pixel.fire(AppPixelName.TAB_MANAGER_REARRANGE_TABS_DAILY, parameters = params, encodedParameters = emptyMap(), Daily())
        }
    }

app/src/test/java/com/duckduckgo/app/appearance/AppearanceViewModelTest.kt
```diff
@@ -84,6 +88,7 @@ internal class AppearanceViewModelTest {
         whenever(mockThemeSettingsDataStore.theme).thenReturn(DuckDuckGoTheme.SYSTEM_DEFAULT)
         whenever(mockAppSettingsDataStore.selectedFireAnimation).thenReturn(FireAnimation.HeroFire)
         whenever(mockAppSettingsDataStore.omnibarType).thenReturn(OmnibarType.SINGLE_TOP)
+        whenever(mockUrlDisplayRepository.isFullUrlEnabled).thenReturn(flowOf(true))
```
malmstein, line 92: this needs tests for the `ViewState` values

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
GerardPaligot: Yes, it is this one: https://github.com/duckduckgo/Android/pull/7227/files#diff-50eec8c691f8b02e973db882088521e0bc1a4ff69cb6ef85fc733b91e2c6f05eR7129-R7137

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
GerardPaligot: They check different things:

`hasUrlPreferenceSet()` checks if the key exists in SharedPreferences at all (old or new app versions) but return true for auto-assigned scenario (if no manual action has been done by the user)

`urlPreferenceManuallySet()` tracks if the user explicitly toggled the setting in the UI

We need the manually set flag to migrate old app users:

* Old users have a stored preference
* But no manual flag (urlPreferenceManuallySet = false) since it didn't exist yet
* We detect this combo to migrate them and protect them from future rollback

Want me to rename `hasUrlPreferenceSet()` to something clearer?

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
GerardPaligot: It is for a specific scenario:

* Fresh new install for a user, we calculate the value from `appBuildConfig.isNewInstall` value
* Every time we come back in this function, we'll enter in this specific case because our preference is set and migrated (so we aren't in this case https://github.com/duckduckgo/Android/pull/7227/files#diff-2ab6c3945b8e0b82cb38316eb346e222e28ce04b58aec1d17502da912f00ce1fR95) and not in this one if the user didn't manually set the preference (https://github.com/duckduckgo/Android/pull/7227/files#diff-2ab6c3945b8e0b82cb38316eb346e222e28ce04b58aec1d17502da912f00ce1fR90)
* If the user has an app update, the user is no more considered as a new user so `appBuildConfig.isNewInstall` will return `false` and the user will switch to the full url but we won't.

app/src/test/java/com/duckduckgo/app/appearance/AppearanceViewModelTest.kt
GerardPaligot: I'm not sure to know what you really want here. Can you give me more information? 🤔

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
malmstein: @GerardPaligot  yeah I think so. Something like `urlPreferenceForNewInstallers` . It’s confusing like it is now.

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
malmstein: @GerardPaligot  gotcha, maybe you can have this as a comment or extract to a self explanatory method.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
malmstein: Oh I missed it, thanks!

app/src/test/java/com/duckduckgo/app/appearance/AppearanceViewModelTest.kt
malmstein: Sure, can you check that the values are correct in the ViewState class depending on what value the Repository returns?

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
GerardPaligot: This flag isn't used only for new installers, we're migrating old users too.

For existing user who previously manually set the preference (so `hasUrlPreferenceSet` return `true` but we don't have `urlPreferenceManuallySet` yet), we migrated these users in the next condition ([here](https://github.com/duckduckgo/Android/pull/7227/files/ca591cccb0a0531a25a7968770340841f264fd38#diff-2ab6c3945b8e0b82cb38316eb346e222e28ce04b58aec1d17502da912f00ce1fR95-R99)) and yes, all new users will edit this flag too [here](https://github.com/duckduckgo/Android/pull/7227/files/ca591cccb0a0531a25a7968770340841f264fd38#diff-2ab6c3945b8e0b82cb38316eb346e222e28ce04b58aec1d17502da912f00ce1fR120).

That allow us to distinguish auto-assigned preference and manual set (previously set or not) and it is important to have both (`hasUrlPreferenceSet` and `urlPreferenceManuallySet`) because we need to update the preference in `suspend fun isFullUrlEnabled(): Boolean` (to prevent future update liked explained [here](https://github.com/duckduckgo/Android/pull/7227#discussion_r2571119298)).

So `hasUrlPreferenceSet` can be set but it is no longer necessary a manual set if this function return `true`. The combination of both allow us to know if it is a manual set or not (and it is important because we won't rollback this specific case if we change the feature flag).

I tried to find a better naming but it is hard. Yours isn't valid too because it isn't only about new installers and I won't contaminate the `SettingsDataStore` interface about implementation details here too.

What I can suggest is to impact the kdoc. The existing one on `hasUrlPreferenceSet` was bad:

```
    /**
     * Check if user has explicitly set their URL display preference.
     */
    fun hasUrlPreferenceSet(): Boolean
```

to:

```
    /**
     * Check if a value has been set to the URL display preference.
     */
    fun hasUrlPreferenceSet(): Boolean
```

and for the migration one:

```
    /**
     * Indicates whether the user has manually set their URL display preference.
     * When true, the user's preference should be preserved even if default behavior changes.
     * When false, the preference may be updated if business rules change (via feature flags).
     */
    var urlPreferenceManuallySet: Boolean
```

to:

```
    /**
     * Indicates whether the user has manually set their URL display preference.
     * You shouldn't use `hasUrlPreferenceSet` to check the same thing because we can edit
     * automatically (e.g. in `UrlDisplayRepository`) the preference without a manual action
     * from the user.
     *
     * When true, the user's preference should be preserved even if default behavior changes.
     * When false, the preference may be updated if business rules change (via feature flags).
     */
    var urlPreferenceManuallySet: Boolean
```

What do you think?

GerardPaligot: Updated

app/src/main/java/com/duckduckgo/app/browser/urldisplay/UrlDisplayRepository.kt
GerardPaligot: Updated

app/src/test/java/com/duckduckgo/app/appearance/AppearanceViewModelTest.kt
GerardPaligot: New test cases added

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -86,8 +86,31 @@ interface SettingsDataStore {
...
+    /**
+     * Indicates whether the user has manually set their URL display preference.
+     * You shouldn't use `hasUrlPreferenceSet` to check the same thing because we can edit
+     * automatically (e.g. in `UrlDisplayRepository`) the preference without a manual action
+     * from the user.
+     *
+     * When true, the user's preference should be preserved even if default behavior changes.
+     * When false, the preference may be updated if business rules change (via feature flags).
+     */
+    var urlPreferenceManuallySet: Boolean
```
mikescamell: 🔍  On the naming thing, maybe something like `urlPreferenceSetByUser` instead?

app/src/test/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModelTest.kt
```diff
@@ -85,7 +89,9 @@ class OmnibarLayoutViewModelTest {
     private val duckAiShowOmnibarShortcutOnNtpAndOnFocusFlow = MutableStateFlow(true)
     private val duckAiShowOmnibarShortcutInAllStatesFlow = MutableStateFlow(true)
     private val duckAiShowInputScreenFlow = MutableStateFlow(false)
+    private val isFullUrlEnabledFlow = Channel<Boolean>()
```
mikescamell: ❓ Why a Channel instead a of StateFlow, for instance?

GerardPaligot: `UrlDisplayRepository` exposes a `Flow` and not a `StateFlow`, and a `Channel` can be consumed as a `Flow`, not a `StateFlow`.

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
GerardPaligot: Done

app/src/test/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModelTest.kt
mikescamell: Can you clarify what you mean here? StateFlow is a type of Flow so my thought was that you could drop it in instead and not have to add the code to cleanup the Channel in `after`

GerardPaligot: Yes, you are right. It's me and it is fixed. Thanks Mike!

app/src/main/java/com/duckduckgo/app/userstate/UserStateReporter.kt
0nko: @malmstein Since the new API was added here, why not just refactor it as part of this PR? Otherwise, you can create a task and I'll have a look at it when I'm on maintenance rotation.

PR #7229, opened 2025-11-26
Fix: Optimize omnibar initialization

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1211909011128655?focus=true

### Description

This PR optimizes the way omnibars are loaded. Instead of inflating all versions and then removing the ones we don't need, we dynamically add the ones we do, based on the omnibar type.

### Steps to test this PR

- [x] Smoke test changing the omnibar type and make sure everything works as expected

---

> [!NOTE]
> Dynamically creates the omnibar based on type (removing XML/attrs and old accessors) and adds DEBUG timing logs for tab view attachment.
> 
> - **Omnibar**:
>   - Dynamically create and insert `OmnibarLayout` via `createAndAddOmnibarLayout`, setting gravity/elevation and bottom behaviors.
>   - Remove attribute-driven setup; `OmnibarLayout` now receives `omnibarType` in constructor; simplify `renderPosition`.
>   - Hide outline strokes for `SPLIT`; adjust coordinator behavior for `SINGLE_BOTTOM`; minor util import (`toPx`).
> - **Layout**:
>   - Remove static `OmnibarLayout` views (`omnibarLayoutTop`/`omnibarLayoutBottom`) from `fragment_browser_tab.xml`.
>   - Delete `values/attrs-omnibar-view.xml` (no longer needed).
> - **Browser components**:
>   - Replace `getBottomNavigationBar()` with `val navigationBar` in `BrowserTabFragment`; update `BrowserActivity` snackbar anchor to `currentTab?.navigationBar`.
> - **Tabs adapter (ViewPager2)**:
>   - Add DEBUG perf logging to `onViewAttachedToWindow` (timing with `SystemClock`), guarded by `BuildConfig.DEBUG`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d720a6a29e80eacdd69efe10400542f9d7c4d41. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
+++ b/app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
@@ -295,8 +295,8 @@ public abstract class FragmentStateAdapter extends RecyclerView.Adapter<Fragment
             placeFragmentInViewHolder(holder);
             gcFragments();
         }
-        long duration = SystemClock.elapsedRealtime() - startTime;
         if (BuildConfig.DEBUG) {
+            long duration = SystemClock.elapsedRealtime() - startTime;
             LogcatKt.logcat("TabPerf",
                     LogPriority.DEBUG, () -> String.format(Locale.getDefault(),
                             "[TabCreation] onViewAttachedToWindow completed in %dms (itemId: %d)",


app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
```diff
@@ -243,20 +243,16 @@ class OmnibarLayout @JvmOverloads constructor(
 
     private var focusAnimator: ValueAnimator? = null
 
-    override val omnibarType: OmnibarType
+    override val omnibarType: OmnibarType by lazy {
+        settingsDataStore.omnibarType
```
CrisBarreiro: Can we do this in a non-blocking way?

app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
cmonfortep, line 1: RE perf measuring: for local testing / debugging is great, but if we wan to keep this in the codebase, can we clean it / extract it into something that doesn't pollute that code that way? Also, since this is just for logging, we don't need this to be executed in production.

marcosholgado: +1

0nko: > can we clean it / extract it into something that doesn't pollute that code that way

Yeah, there are too many of them. I'll only keep the most important one, which shows the total cost of the tab display -- `onViewAttachedToWindow()`.

> Also, since this is just for logging, we don't need this to be executed in production

While in Kotlin the `logcat` calls are not executed (the lambda of the inline function is only executed if a logger is installed) in the release build, in Java they are. I added a BuildConfig check to prevent that.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
0nko: I've [refactored](https://github.com/duckduckgo/Android/pull/7229/commits/f0d10e197416e023e8600d20b74116887685b374) the way `OmnibarLayout` gets the `OmnibarType` value.

cmonfortep (approved):
LGTM

PR #7247, opened 2025-11-28
Compose: Add DaxTextField

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1210594645151737/task/1211659112661196?focus=true

### Description
Adds a Compose version of the DaxTextField.

### Steps to test this PR

_ADS Preview_
- [ ] Go to ADS Preview screen and click on Text Input tab 
- [ ] Check compose versions 

### UI changes
See ADS Preview screen

---

> [!NOTE]
> Introduces Compose text fields (standard and secure) with trailing icon API and previews, adds theme text-field colors, and enforces trailing icon usage via new lint checks.
> 
> - **Design System (Compose)**:
>   - **New components**: `DaxTextField` and `DaxSecureTextField` with `DaxTextFieldInputMode`, `DaxTextFieldLineLimits`, keyboard options, input transformations, error/supporting text, and `DaxTextFieldTrailingIconScope` API.
>   - **Secure field**: visibility toggle integrated with optional trailing icon.
>   - **Previews**: comprehensive previews for states and configurations.
> - **ADS Preview/Examples**:
>   - Integrates multiple Compose samples in `ComponentTextInputFragment` and `component_text_input_view.xml` (hint, single/multi/form, read-only, disabled, error, URL/IP, focus behaviors, observable text, password variants).
> - **Theme**:
>   - Adds `DuckDuckGoTextFieldColors` (`colors.textField.borders`) and exposes `DuckDuckGoTheme.iconColors`; minor color addition `White78`.
> - **Lint Rules**:
>   - New detectors: `DaxTextFieldTrailingIconDetector` and `DaxSecureTextFieldTrailingIconDetector`; registered in `DuckDuckGoIssueRegistry` with extensive tests to enforce using `DaxTextFieldTrailingIconScope` in `trailingIcon`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 33aa9f12d2ecac08e12727d003e5366c74f59633. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
+++ b/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
@@ -260,7 +260,7 @@ internal fun DaxSecureTextField(
                             DaxText(
                                 text = error,
                                 style = DuckDuckGoTheme.typography.caption,
-                                color = DuckDuckGoTheme.colors.destructive,
+                                color = DuckDuckGoTheme.textColors.destructive,
                             )
                         }
                     } else {
@@ -469,7 +469,7 @@ private fun DaxSecureTextFieldPreviewBox(
     content: @Composable () -> Unit,
 ) {
     DuckDuckGoTheme {
-        PreviewBox(backgroundColor = DuckDuckGoTheme.colors.background) {
+        PreviewBox {
             content()
         }
     }
--- a/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
+++ b/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
@@ -155,7 +155,7 @@ fun DaxTextField(
                         DaxText(
                             text = error,
                             style = DuckDuckGoTheme.typography.caption,
-                            color = DuckDuckGoTheme.colors.destructive,
+                            color = DuckDuckGoTheme.textColors.destructive,
                         )
                     }
                 } else {
@@ -250,10 +250,6 @@ object DaxTextFieldTrailingIconScope {
             )
         }
     }
-
-    @Composable
-    fun SomeComposable() {
-    }
 }
 
 @Stable
@@ -313,32 +309,32 @@ internal fun daxTextFieldColors(): TextFieldColors = OutlinedTextFieldDefaults.c
     unfocusedContainerColor = Transparent,
     disabledContainerColor = Transparent,
     errorContainerColor = Transparent,
-    focusedBorderColor = DuckDuckGoTheme.colors.accentBlue,
+    focusedBorderColor = DuckDuckGoTheme.colors.brand.accentBlue,
     unfocusedBorderColor = DuckDuckGoTheme.colors.textField.borders,
     disabledBorderColor = DuckDuckGoTheme.colors.textField.borders,
-    errorBorderColor = DuckDuckGoTheme.colors.destructive,
-    focusedLabelColor = DuckDuckGoTheme.colors.accentBlue,
+    errorBorderColor = DuckDuckGoTheme.textColors.destructive,
+    focusedLabelColor = DuckDuckGoTheme.colors.brand.accentBlue,
     unfocusedLabelColor = DuckDuckGoTheme.textColors.secondary,
     disabledLabelColor = DuckDuckGoTheme.textColors.secondary,
-    errorLabelColor = DuckDuckGoTheme.colors.destructive,
+    errorLabelColor = DuckDuckGoTheme.textColors.destructive,
     focusedTrailingIconColor = DuckDuckGoTheme.iconColors.primary,
     unfocusedTrailingIconColor = DuckDuckGoTheme.iconColors.primary,
     disabledTrailingIconColor = DuckDuckGoTheme.iconColors.primary,
-    errorTrailingIconColor = DuckDuckGoTheme.colors.destructive,
-    focusedSupportingTextColor = DuckDuckGoTheme.colors.destructive,
-    unfocusedSupportingTextColor = DuckDuckGoTheme.colors.destructive,
-    disabledSupportingTextColor = DuckDuckGoTheme.colors.destructive,
-    errorSupportingTextColor = DuckDuckGoTheme.colors.destructive,
-    cursorColor = DuckDuckGoTheme.colors.accentBlue,
-    errorCursorColor = DuckDuckGoTheme.colors.accentBlue,
+    errorTrailingIconColor = DuckDuckGoTheme.textColors.destructive,
+    focusedSupportingTextColor = DuckDuckGoTheme.textColors.destructive,
+    unfocusedSupportingTextColor = DuckDuckGoTheme.textColors.destructive,
+    disabledSupportingTextColor = DuckDuckGoTheme.textColors.destructive,
+    errorSupportingTextColor = DuckDuckGoTheme.textColors.destructive,
+    cursorColor = DuckDuckGoTheme.colors.brand.accentBlue,
+    errorCursorColor = DuckDuckGoTheme.textColors.primary,
     selectionColors = TextSelectionColors(
-        handleColor = DuckDuckGoTheme.colors.accentBlue,
-        backgroundColor = DuckDuckGoTheme.colors.accentBlue.copy(alpha = DaxTextFieldDefaults.ALPHA_DISABLED),
+        handleColor = DuckDuckGoTheme.colors.brand.accentBlue,
+        backgroundColor = DuckDuckGoTheme.colors.brand.accentBlue.copy(alpha = DaxTextFieldDefaults.ALPHA_DISABLED),
     ),
     focusedLeadingIconColor = DuckDuckGoTheme.iconColors.primary,
     unfocusedLeadingIconColor = DuckDuckGoTheme.iconColors.primary,
     disabledLeadingIconColor = DuckDuckGoTheme.iconColors.primary,
-    errorLeadingIconColor = DuckDuckGoTheme.colors.destructive,
+    errorLeadingIconColor = DuckDuckGoTheme.textColors.destructive,
     focusedPrefixColor = DuckDuckGoTheme.textColors.secondary,
     unfocusedPrefixColor = DuckDuckGoTheme.textColors.secondary,
     disabledPrefixColor = DuckDuckGoTheme.textColors.secondary,
@@ -539,7 +535,7 @@ private fun DaxTextFieldPreviewBox(
     content: @Composable () -> Unit,
 ) {
     DuckDuckGoTheme {
-        PreviewBox(backgroundColor = DuckDuckGoTheme.colors.background) {
+        PreviewBox {
             content()
         }
     }
--- a/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
+++ b/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
@@ -41,7 +41,7 @@ object DuckDuckGoTheme {
         @ReadOnlyComposable
         get() = colors.text
 
-    val iconColors: DuckDuckGoIconColors
+    val iconColors: DuckDuckGoIconsColors
         @Composable
         @ReadOnlyComposable
         get() = colors.icons


android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
```diff
@@ -0,0 +1,388 @@
...
+ * @param trailingIcon Optional [DaxTextFieldTrailingIcon] that will be displayed at the end of the text field.
+ * @param keyboardOptions Software keyboard options that contains configuration such as [KeyboardType] and [ImeAction].
+ * @param onClick Optional callback that is triggered when the text field is [clickable] or when [trailingIcon] is clicked.
+ * @param onFocusChanged Optional callback that is triggered when the focus state of the text field changes.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1212213756433276?focus=true
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?m=auto&node-id=3202-5150
+ */
+@Composable
+fun DaxSecureTextField(
```
landomen: This needs to be a separate composable as it's using a `OutlinedSecureTextField` underneath instead of the regular `OutlinedTextField`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,562 @@
...
+
+    /**
+     * Converts this [DaxTextFieldLineLimits] to a [TextFieldLineLimits] used by the underlying TextField.
+     */
+    fun toLineLimits(): TextFieldLineLimits = when (this) {
+        SingleLine -> TextFieldLineLimits.SingleLine
+        MultiLine -> TextFieldLineLimits.MultiLine(minHeightInLines = 1)
+        Form -> TextFieldLineLimits.MultiLine(minHeightInLines = 3)
+    }
+}
```
landomen, line 282: I initially had three separate composables that called a base composable underneath: `DaxSingleLineTextField`, `DaxMultiLineTextField` and `DaxFormTextField`. But all three were exactly the same, with same arguments, only difference was what they used for `TextFieldLineLimits`, so I thought it was clearer this way. 
It was also a bit hard to make and maintain the same changes across four composables, as any change to the base one needs to be propagated.

If there is a strong preference towards having separate composables, I can easily split this.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,562 @@
...
+ * @param inputTransformation Optional transformation to apply to the input text before it's written to the state. Can be used for input filtering.
+ * @param onClick Optional callback that is triggered when the text field is [clickable] or when [trailingIcon] is clicked.
+ * @param onFocusChanged Optional callback that is triggered when the focus state of the text field changes.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1212213756433276?focus=true
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?m=auto&node-id=2691-3327
+ */
+@Composable
+fun DaxTextField(
+    state: TextFieldState,
```
landomen, line 83: Decided to go with the [state-based text fields](https://developer.android.com/develop/ui/compose/text/user-input?textfield=state-based) as that is the recommended approach. There is a synchronization issue with the old value-based text fields.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
```diff
@@ -81,6 +81,7 @@ fun DuckDuckGoTheme(
         window = colorResource(R.color.white),
         destructive = colorResource(R.color.alertRedOnLightDefault),
         lines = colorResource(R.color.black9),
+        borders = colorResource(R.color.black30),
```
malmstein: `borders` is too generic, so far it only applies to the input field. how do you feel about doing `inputFieldBorders` instead?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
malmstein: I prefer having one instead, good call!

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
landomen: My preference is to keep it generic. This list of colors should be enough to build any component that follows ADS. I don't think we should add one-off colors to this. But I don't feel strongly about this, especially since it doesn't seem any other component has borders and uses `T-Control/Border-Primary` color as it's named in Figma.

One other option is that each component specifies it's own colors. We kinda already do that for the `DaxText`. So for text field it would be something like:
```
textField = DuckDuckGoTextFieldColors(
    text: Color,
    error: Color,
    borders: Color,
    ...
)
```
We could either provide all needed colors or just the ones that are not generic.

What do you think?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,562 @@
...
+    modifier: Modifier = Modifier,
+    hint: String? = null,
```
GerardPaligot: Warning: We need to take care about the naming here. The Compose TextField has two native parameters: `placeholder` and `label`. `placeholder` is displayed when you click on the text field but the user didn't enter anything yet and `label` is displayed at the beginning in the text field and is pushed at the top when the user click on the component.

`hint` seems too vague and according to the Figma specification, it seems that the `placeholder` isn't documented in our component specification. Maybe we should ask to a designer what they want and impact the api contract here according to their feedback?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,562 @@
...
+    lineLimits: DaxTextFieldLineLimits = DaxTextFieldLineLimits.MultiLine,
+    enabled: Boolean = true,
+    editable: Boolean = true,
+    clickable: Boolean = false,
+    error: String? = null,
+    trailingIcon: DaxTextFieldTrailingIcon? = null,
+    keyboardOptions: KeyboardOptions = DaxTextFieldDefaults.TextKeyboardOptions,
+    inputTransformation: InputTransformation? = null,
+    onClick: (() -> Unit)? = null,
+    onFocusChanged: ((Boolean) -> Unit)? = null,
```
GerardPaligot: Suggestion: Instead of declaring a callback, maybe we can expose a `interactionSource: MutableInteractionSource` to observe a focus change on the consumer side (`interactionSource.collectIsFocusedAsState()`)?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,562 @@
...
+    hint: String? = null,
+    lineLimits: DaxTextFieldLineLimits = DaxTextFieldLineLimits.MultiLine,
+    enabled: Boolean = true,
+    editable: Boolean = true,
+    clickable: Boolean = false,
+    error: String? = null,
+    trailingIcon: DaxTextFieldTrailingIcon? = null,
+    keyboardOptions: KeyboardOptions = DaxTextFieldDefaults.TextKeyboardOptions,
+    inputTransformation: InputTransformation? = null,
+    onClick: (() -> Unit)? = null,
```
GerardPaligot: Question: It seems that this parameter is used when you click on the text field (L115) and when you click on the trailing icon (L142), can we distinguish these two sub components?

Suggestion: Or do we really need to add this callback on the  root component and maybe add an api slot for the trailing icon where we'll declare the callback when we need to observe user click on this sub component? This will allow us to propose an api contract much more cleaner on the text field component and a better responsibility definition for every sub components.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,562 @@
...
+ */
+@Composable
+fun DaxTextField(
+    state: TextFieldState,
+    modifier: Modifier = Modifier,
+    hint: String? = null,
+    lineLimits: DaxTextFieldLineLimits = DaxTextFieldLineLimits.MultiLine,
+    enabled: Boolean = true,
+    editable: Boolean = true,
+    clickable: Boolean = false,
```
GerardPaligot: Question: `enabled`, `editable` and `clickable` are very confusing and I didn't find a clear design spec about these 3 states. Are they really necessary? 🤔

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
GerardPaligot: Another approach with Compose is to define one generic component that will handle for us all the UI customisation and then, propose easy to use variant components for dedicated cases. It depends a lot the experience we want to propose to the team.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
GerardPaligot: It depends a lot on the design spec. This `borders` is used only in the text field or every time we need to use a border, they'll use the exact same border color?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: @GerardPaligot I agree, that was my initial approach. But there was no difference between the components apart from the number of lines, so the maintenance overhead of having three separate composables vs one argument didn't seem worth it. However, I'm okay with going with separate composables if that is our approach we want to take.

The password text field is a separate composable as it's has a number of differences, including a different base composable.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
landomen: @GerardPaligot I'm not familiar enough with the ADS to answer that confidently. I didn't see any other references to this color in Figma page for ADS components. For now, the `borders` is only used by the text field. But keep in mind that the text field is only the second Compose component we have 😅

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: @GerardPaligot I'm open to changing the name. I was on the fence between `hint` and `label`, but went with `hint` for two reasons: it's also used in the old `DaxTextInput` component and combines both `label` and `placeholder` since we're not using `placeholder`.

We're not using the `placeholder` parameter, but only rely on the `label` parameter, because it matches how our current View-based `TextInput` component works. The `label` is displayed inside the text field when it's empty and not focused, and animates to the top of the text field when it's focused or has text.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: @GerardPaligot I like the idea, could you expand a bit more what you had in mind? We're using  `Modifier.onFocusChanged` to observe and dispatch changes and to detect clicks on the text field.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: @GerardPaligot Great question! I initially had a composable slot for the trailing icon so consumers could put their own `IconButton` there as that's the Compose way. However, we would then need to set up lint rules to ensure people are not putting other elements there or using wrong styling. It also puts more burden on the consumers as they need to be aware of what to put in the slot. But I'm open to using that or other approaches if we think it would be better. It depends on what we want to do with our Compose ADS.

I went with a single `onClick()` callback based on how the old TextInput works. There are four possible options: 
- there is no icon and text field is not clickable
- icon is clickable
- text field is clickable
- both are clickable
I felt having two separate callbacks could be a bit confusing, but I'm open to changing that.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: This was ported from the existing View-Based `DaxTextInput` to support various different states (like having a non-editable, but enabled and clickable input). Taking a second look I agree it's a bit confusing, especially since `enabled` for example can be overwritten by `editable` or `clickable`. Maybe using an enum could be better? I'm open to suggestions

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
GerardPaligot: Yes, here is an usage example:

With the actual contract, you have this:

```
            var isFocused by remember { mutableStateOf(false) }
            LaunchedEffect(isFocused) {
                if (isFocused) {
                    state.edit {
                        selectAll()
                    }
                }
            }

            DaxTextField(
                state = state,
                hint = "Autoselect text on focus",
                onFocusChanged = {
                    isFocused = it
                },
                lineLimits = DaxTextFieldLineLimits.SingleLine,
            )
```

And with the interaction source, we have this:

```
            val interactionSource = remember { MutableInteractionSource() }
            val isFocused by interactionSource.collectIsFocusedAsState()
            LaunchedEffect(isFocused) {
                if (isFocused) {
                    state.edit {
                        selectAll()
                    }
                }
            }

            DaxTextField(
                state = state,
                hint = "Autoselect text on focus",
                lineLimits = DaxTextFieldLineLimits.SingleLine,
                interactionSource = interactionSource,
            )
```

We avoid to create our own state because this state already exists in `MutableInteractionSource` and we just use it to observe the focus state (but we can also observe other kind of states like drag, hover and pressed states).

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
GerardPaligot: Ok, I understand why the `label` parameter is called `hint`. When you come from XML framework, it is understandable but I'm afraid that this naming will be bad in the future. More we'll use Compose, more we'll forget XML property names and at some point, we'll ask us again why we called this `hint` when the native name for this parameter is `label`. I don't have a strong opinion about the naming of this parameter but we need to choose carefully from my opinion.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
GerardPaligot: The need behind the click on the text field seems to be more linked to the focus than something else, so the need behind this click can be covered by this https://github.com/duckduckgo/Android/pull/7247#discussion_r2603559243? What do you think?

About api slot or not and with a scope or not, it is always hard to take this kind of decision in compose. For example, we can suggest this kind of contract:

```
@Composable
fun DaxTextField(
    state: TextFieldState,
    modifier: Modifier = Modifier,
    hint: String? = null,
    lineLimits: DaxTextFieldLineLimits = DaxTextFieldLineLimits.MultiLine,
    enabled: Boolean = true,
    editable: Boolean = true,
    clickable: Boolean = false,
    error: String? = null,
    trailingIcon: (@Composable TrailingScope.() -> Unit)? = null,
    keyboardOptions: KeyboardOptions = DaxTextFieldDefaults.TextKeyboardOptions,
    inputTransformation: InputTransformation? = null,
    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
    // onClick: (() -> Unit)? = null, With interactionSource and api slot trailingIcon, we don't need this parameter anymore
)

object TrailingScope {
    @Composable
    fun IconButton(
        @DrawableRes iconResId: Int,
        contentDescription: String?,
        modifier: Modifier = Modifier,
        tint: Color = DuckDuckGoTheme.colors.iconPrimary,
        enabled: Boolean = true,
        onClick: (() -> Unit)? = null,
    ) {
      // expected composable here
    }
}
```

Then, you can use the component like this:

```
val state = rememberTextFieldState("This is not editable.")
DaxTextField(
    state = state,
    hint = "Non-editable text with end icon",
    trailingIcon = {
        IconButton(
            iconResId = R.drawable.ic_copy_24,
            contentDescription = "Copy",
            onClick = { toastOnClick(Action.PerformEndAction) }
        )
    },
    editable = false,
    lineLimits = DaxTextFieldLineLimits.SingleLine,
)
```

This is useful if you want to help the component consumer to declare the correct component in the api slot but it isn't required. If for some reasons, you want something else, scoped api slot doesn't force you to use one component present in the scope object.

This is also possible:

```
DaxTextField(
    state = state,
    hint = "Non-editable text with end icon",
    trailingIcon = {
        Text(text = "My end icon")
    },
    editable = false,
    lineLimits = DaxTextFieldLineLimits.SingleLine,
)
```

So yes, if we chose the api slot contract and we won't this kind of thing in the codebase, we'll need to write linter rules. (btw, we'll probably need to write linter rules when we'll start to use compose in the codebase for many other usecases).

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
GerardPaligot: Let me build the state matrix here to be sure I'm understanding everything well:

Usecases | Enabled | Editable | Clickable
----- | -------- | -------- | ----------
I can fully interact with the text field | x | x | x
Read only mode | x | | x
Edited by the business logic, not by user |  | x |  x
**No use case for this scenario** | x | x | 
**No use case for this scenario** | x |  | 
Edited by the business logic, not by user | | x | 
Focusable mode | | | x

Am I right?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
GerardPaligot: Me neither, and I'm agree with you. I was just thinking if we should onboard a designer on this topic or not to build with us property theming or not but I'm probably overthinking it.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Color.kt
```diff
@@ -40,11 +40,13 @@ data class DuckDuckGoColors(
     val window: Color,
     val destructive: Color,
     val lines: Color,
+    val borders: Color,
     val accentContentPrimary: Color,
     val accentBlue: Color,
     val accentYellow: Color,
     val ripple: Color,
     val text: DuckDuckGoTextColors,
+    val iconPrimary: Color,
```
mikescamell: 💡Could we already create a DuckDuckGoIconColors with iconPrimary as our first? Similar to DuckDuckGoTextColors

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
mikescamell: Yeah tough one seeing as the Design System has nothing in this regard except "label". I'm open to having both label and placeholder as we moving away from XML, but I understand why Domen went for this due how it's used. I'm not too concerned about forgetting XML property names, Domen has documented how hint behaves in the kdoc and in my opinion we should build Composables to match our design system components where possible, and not Google's Material APIs. I lean towards label and placeholder I think in this case 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
mikescamell: Yeah I agree with Gerard in this case, and yourself Domen about it being confusing. Maybe these 3 states could become an enum instead? Then we can deal with setting the different combination of states ourselves in the Composable

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
```diff
@@ -0,0 +1,388 @@
...
+                    }
+                }
+                .alpha(
+                    if (enabled || (!editable && clickable)) {
+                        DaxTextFieldDefaults.ALPHA_ENABLED
+                    } else {
+                        DaxTextFieldDefaults.ALPHA_DISABLED
+                    },
+                ),
+            enabled = enabled || editable || clickable,
```
mikescamell: ❓We don't have a readOnly state like DaxTextField, is that intentional?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
mikescamell: My gut says for now keep the color limited to DaxTextField. I don't see it being used elsewhere. If we see this is needed for another Composable in future then we can expose it globally.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
GerardPaligot: My hypothesis is that we can manage with the two native parameters (`readOnly` and `enabled`) but I need to be sure that I understand well the state matrix before.

mikescamell: Maybe something like?

```
 sealed interface DaxTextFieldMode {
      data object Editable : DaxTextFieldMode
      data object ReadOnly : DaxTextFieldMode
      data object Clickable : DaxTextFieldMode
      data object Disabled : DaxTextFieldMode
  }
```

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
mikescamell: Yeah I agree with the enum approach in this case specifically, it's the same InputText with just different line limits. Usually I'd move towards what Gerard suggested but in this case I think it makes more sense. We'll see more of this when I raise the DaxButton PR 👀

mikescamell:
Screenshot_20251210_103909

@landomen I can still click on the disabled version and get focus.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,562 @@
...
+
+/**
+ * Represents a trailing icon for the text field.
+ *
+ * @param iconResId Resource ID of the drawable icon.
+ * @param contentDescription Optional content description for accessibility.
+ */
+@Stable
+data class DaxTextFieldTrailingIcon(
+    @DrawableRes val iconResId: Int,
```
mikescamell: 💡 Could we use a `painter` here instead of a DrawableRes? I don't think we need to limit this to DrawableRes only.

mikescamell:
> Screenshot_20251210_103909
> @landomen I can still click on the disabled version and get focus.

I think we need yo end up in a state of

```
            enabled = false,
            readOnly = true,
```

For it not to be focusable

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
GerardPaligot: Why not an enum, in this case we can define degressive interactions on the component?

* Editable: You can edit, focus and select the content
* Clickable: You can focus and select the content (why not `Focusable`?)
* Readonly: You can't interact with the component at all
* Disable: All capabilities are deactivate

That is why I said we can probable keep only `enabled` and `readOnly`:

Enabled | Read Only | Capability
-------- | ---------- | ---------- 
true | false | Editable
false | true | ReadOnly
true | true | Clickable
false | false | Disabled

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Theme.kt
landomen: Limited the borders color to TextField only for now.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/Color.kt
landomen: Good idea! Introduced `DuckDuckGoIconColors`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: I agree with both of you. Renamed the parameter from `hint` to `label` to follow the standard Compose naming. 

I'm not exposing a `placeholder` parameter at this time, as it conflicts with how our existing ADS text input currently behaves (see video, the placeholder is constantly visible). It should be pretty simple to expose it later if we need it.

Let me know if you still think it's a good idea to expose it, I can add it.

https://github.com/user-attachments/assets/d1a3293c-2c33-4be8-8689-6e575bec3607

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: Updated to use `MutableInteractionSource` instead of the `onFocusChanged` callback.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: Thanks for the suggestion, @GerardPaligot! I went with the scoped api slot. I'll write a lint rule to ensure we're only using the expected composable. We already have a lint rule for DaxText that was written by @mikescamell .

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: Decided to go with an enum for simplicity with 3 values: `Editable`, `ReadOnly` and `Disabled`. I initially also had `Clickable` but it resulted in the same combination as `ReadOnly` (`enabled=true`, `readOnly =true`). Since we'll be using an InteractionSource now to detect focus and clicks, it's not really needed. Consumers can simply set the text field to ReadOnly and apply the interaction source.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
landomen: `SecureTextField` doesn't support `readOnly` property, only `enabled` 😢 I'm trying to find a way around it by either using the BasicSecureTextField or leveraging the regular OutlinedTextField with custom output transformation.

landomen:
> > Screenshot_20251210_103909
> > @landomen I can still click on the disabled version and get focus.
> 
> I think we need yo end up in a state of
> 
> ```
>             enabled = false,
>             readOnly = true,
> ```
> 
> For it not to be focusable

Thanks for catching this! This was a bug because of the wrong combination of arguments after I cleaned up the code 😅. It's fixed now with the new enum.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: Replaced with `painter`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
landomen: Solved by using `BasicSecureTextField`

landomen:
@mikescamell @GerardPaligot Thanks for all your great feedback! I've refactored the composable, please take another look when you get a chance.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
```diff
@@ -58,11 +96,429 @@ class ComponentTextInputFragment : Fragment() {
...
+                )
+            DaxTextField(
+                state = state,
+                label = "Form mode editable text",
+                lineLimits = DaxTextFieldLineLimits.Form,
+            )
+        }
+
+        // Non-editable text full click listener with end icon
+        view.setupThemedComposeView(id = com.duckduckgo.common.ui.internal.R.id.compose_text_input_30, isDarkTheme = isDarkTheme) {
```
GerardPaligot, line 194: We have a difference between the XML and Compose components here.
When I'm clicking on the XML view, the view isn't focused.
The same thing on the Compose component switch the border in blue (focus mode).

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
```diff
@@ -58,11 +96,429 @@ class ComponentTextInputFragment : Fragment() {
...
+                    )
+                },
+                inputMode = DaxTextFieldInputMode.ReadOnly,
+                lineLimits = DaxTextFieldLineLimits.SingleLine,
+                interactionSource = interactionSource,
+            )
+        }
+
+        // Non-editable text full click listener
+        view.setupThemedComposeView(id = com.duckduckgo.common.ui.internal.R.id.compose_text_input_31, isDarkTheme = isDarkTheme) {
```
GerardPaligot, line 225: We have a difference between the XML and Compose components here.
When I'm clicking on the XML view, the view isn't focused.
The same thing on the Compose component switch the border in blue (focus mode).

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
```diff
@@ -0,0 +1,385 @@
...
+ * @param interactionSource Optional interaction source for observing and emitting interaction events.
+ * You can use this to observe focus, pressed, hover and drag events.
+ * @param trailingIcon Optional trailing icon composable to display at the end of the text field.
+ * Use [DaxTextFieldTrailingIconScope.DaxTextFieldTrailingIcon] to create the icon.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1212213756433276?focus=true
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?m=auto&node-id=3202-5150
+ */
+@Composable
+fun DaxSecureTextField(
```
GerardPaligot: In dark mode, the text inside the dax secure text field is black but should be white.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
```diff
@@ -58,11 +96,429 @@ class ComponentTextInputFragment : Fragment() {
...
+                state = state,
+                label = "Non-editable text without end icon",
+                inputMode = DaxTextFieldInputMode.ReadOnly,
+            )
+        }
+
+        // Editable password that fits in one line
+        view.setupThemedComposeView(id = com.duckduckgo.common.ui.internal.R.id.compose_text_input_6, isDarkTheme = isDarkTheme) {
+            val state = rememberTextFieldState("Loremipsumolor")
+            DaxSecureTextField(
```
GerardPaligot, line 324: We have a difference between the XML and Compose components here.
When I'm clicking on the XML view, the content is no more masked.
When I'm clicking on the Compose view, the content stay masked.
From my opinion, we should keep the Compose behavior but because there is a difference, I want to report it here.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
```diff
@@ -58,11 +96,429 @@ class ComponentTextInputFragment : Fragment() {
...
+                            toastOnClick(Action.PerformEndAction)
+                        },
+                    )
+                },
+                inputMode = DaxTextFieldInputMode.ReadOnly,
+            )
+        }
+
+        // Error
+        view.setupThemedComposeView(id = com.duckduckgo.common.ui.internal.R.id.compose_text_input_21, isDarkTheme = isDarkTheme) {
```
GerardPaligot, line 381: We have a difference between the XML and Compose components here.
When I'm clicking on the XML view, the cursor is white/black (according to if you are in dark/light mode).
When I'm clicking on the Compose view, the cursor is in red.
From my opinion, we should keep the Compose behavior but because there is a difference, I want to report it here.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
```diff
@@ -0,0 +1,385 @@
...
+    label: String? = null,
+    inputMode: DaxTextFieldInputMode = DaxTextFieldInputMode.Editable,
+    error: String? = null,
+    keyboardOptions: KeyboardOptions = DaxTextFieldDefaults.PasswordKeyboardOptions,
+    interactionSource: MutableInteractionSource? = null,
+    trailingIcon: (@Composable DaxTextFieldTrailingIconScope.() -> Unit)? = null,
+) {
+    // needed by the OutlinedTextField container
+    val internalInteractionSource = interactionSource ?: remember { MutableInteractionSource() }
+    var isPasswordVisible by remember { mutableStateOf(false) }
```
GerardPaligot: Suggestion: What do you think to put a boolean parameter in the compose contract and let the component consumer control the visibility state of this component? This will allow us to keep a stateless component and give a better control of the component to the consumer. If we want to display the password for any other reasons than a click on the trailing icon, we can't with our actual contract.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
```diff
@@ -0,0 +1,385 @@
...
+            DaxTextFieldTrailingIconScope.DaxTextFieldTrailingIcon(
+                painter = painterResource(
+                    if (isPasswordVisible) {
+                        R.drawable.ic_eye_closed_24
+                    } else {
+                        R.drawable.ic_eye_24
+                    },
+                ),
+                contentDescription = null,
+                onClick = { isPasswordVisible = !isPasswordVisible },
```
GerardPaligot: Suggestion: It is a good idea to declare our eye icon as an implementation details but if we remove `isPasswordVisible` as an internal state, we need to add a callback in the component contract too. Because this icon is mandatory for our secured text field, I think it is okay.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
```diff
@@ -0,0 +1,385 @@
...
+                fontFamily = FontFamily.Default,
+                color = DuckDuckGoTheme.colors.text.secondary,
+            ),
+            bodyLarge = DuckDuckGoTheme.typography.body1.asTextStyle.copy(
+                fontFamily = FontFamily.Default,
+                color = DuckDuckGoTheme.colors.text.secondary,
+            ),
+        ),
+    ) {
+        CompositionLocalProvider(LocalTextSelectionColors provides daxTextFieldColors().textSelectionColors) {
```
GerardPaligot: Suggestion: You are calling multiple times `daxTextFieldColors()`, consider to save it in a dedicated variable with a remember call because the value of colors won't change between two compositions.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
```diff
@@ -0,0 +1,385 @@
...
+                    },
+                ),
+                contentDescription = null,
+                onClick = { isPasswordVisible = !isPasswordVisible },
+                enabled = inputMode == DaxTextFieldInputMode.Editable || inputMode == DaxTextFieldInputMode.ReadOnly,
+            )
+
+            trailingIcon?.let {
+                DaxTextFieldTrailingIconScope.it()
+            }
```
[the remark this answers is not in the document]
landomen: It's the caller's responsibility to disable it

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
landomen: Nice catch, thanks!

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
landomen: Stored it to a local variable. I can't use remember because it's a composable function, but it should be fine without it. Since it doesn't read any parameters, it won't get re-composed.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
landomen: Lifted the state outside so it's now the consumers responsibility. It makes sense to have a stateless component, however I do think this impacts the ease of use by needing more boilerplate state management. 

In all the currently supported use-cases , the only reason to show/hide password is because of the internal icon click. It seems we're predicting potential other usages in the future, incorporating that in the API, and by doing that,  making it harder to use for existing purpose.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
landomen: Lifted the state outside so it's now the consumers responsibility. See https://github.com/duckduckgo/Android/pull/7247#discussion_r2621917191 for more info

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
landomen: Thanks for reporting it! 

Yeah, the XML one first shows unmasked text when focused and masking it requires action.
The Compose one first shows masked text when focused and unmasking it requires action.

I agree the Compose behavior is better, more secure and I think more in line with how a secure text input should behave. So I'm voting for keeping it. @mikescamell Your thoughts?

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
landomen: Good catch! I also vote for the Compose behavior, but making the cursor color match XML should be simple. @mikescamell How do you feel?

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
landomen: Not sure if there is an easy way to prevent that. Would probably have to do something with the `MutableInteractionSource` but that could be tricky. In my opinion, the current behavior is okay, because you interacted with the text field (it's still enabled). 

The XML one sometimes also behaves like that, at least on the "Non-editable text with line truncation" example. And there is a bug on the XML one where the keyboard still shows up and you can enter new text even though its set to non-editable 😅

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
landomen: https://github.com/duckduckgo/Android/pull/7247#discussion_r2621948522

landomen:
@GerardPaligot @mikescamell I pushed two new commits:
* [Lift password state - addresses comments](https://github.com/duckduckgo/Android/pull/7247/commits/9c4d4a6bb480ea0e5853fb0b0f47aa245a0120ad)
* [Add new lint detector - ensures we can only use `DaxTextFieldTrailingIcon` composable in the `trailingIcon` parameter](https://github.com/duckduckgo/Android/pull/7247/commits/a7aa6a289444bfe32a768ac367b56a41fcc9ac6f)

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,540 @@
...
+     */
+    fun toLineLimits(): TextFieldLineLimits = when (this) {
+        SingleLine -> TextFieldLineLimits.SingleLine
+        MultiLine -> TextFieldLineLimits.MultiLine(minHeightInLines = 1)
+        Form -> TextFieldLineLimits.MultiLine(minHeightInLines = 3)
+    }
+}
+
+@Stable
+enum class DaxTextFieldInputMode {
```
mikescamell, line 285: ❤️

lint-rules/src/main/java/com/duckduckgo/lint/ui/DaxTextFieldTrailingIconDetector.kt
```diff
@@ -0,0 +1,111 @@
...
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.SourceCodeScanner
+import com.android.tools.lint.detector.api.TextFormat
+import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.getParameterForArgument
+import java.util.EnumSet
+
+@Suppress("UnstableApiUsage")
+class DaxTextFieldTrailingIconDetector : Detector(), SourceCodeScanner {
```
mikescamell, line 34: ❤️

lint-rules/src/main/java/com/duckduckgo/lint/ui/DaxSecureTextFieldTrailingIconDetector.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.SourceCodeScanner
+import com.android.tools.lint.detector.api.TextFormat
+import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.getParameterForArgument
+import java.util.EnumSet
+
+@Suppress("UnstableApiUsage")
+class DaxSecureTextFieldTrailingIconDetector : Detector(), SourceCodeScanner {
```
mikescamell, line 34: ❤️

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,540 @@
...
+                fontFamily = FontFamily.Default,
+                color = DuckDuckGoTheme.colors.text.secondary,
+            ),
+            bodyLarge = DuckDuckGoTheme.typography.body1.asTextStyle.copy(
+                fontFamily = FontFamily.Default,
+                color = DuckDuckGoTheme.colors.text.secondary,
+            ),
+        ),
+    ) {
+        CompositionLocalProvider(LocalTextSelectionColors provides daxTextFieldColors().textSelectionColors) {
```
mikescamell: ❓We call daxTextFieldColors twice, here and ln 166. Can we just allocate this once at the top of the Composable and use it here and ln 166?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
GerardPaligot: Yes, the actual implementation isn't "bad". Just, we'll probably need to deprecate this variant if we have the use case in the future. I'm okay to keep the actual component if you think it is better because it'll simplify the actual usage in the codebase.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
mikescamell: 💡 Wouldn't this be solved by setting `enabled` to false when in `ReadOnly` state? I'm wondering if there are any ramifications to this.

GerardPaligot: Yes, from my opinion, the current behavior is okay too because if the component is enabled, it should be focusable (and therefore readable by screen readers). I'm okay to keep the compose behavior.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
mikescamell: Yeah I agree, it feels weird to have the show password icon button if clicking into the shows the password implicitly 🤔 

@malmstein any historical context on why this happens?

Clicking on the View version, password shown automatically:

Screenshot_20251216_095657

What we'd expect, clicking doesn't automatically show the password until the show password button is clicked:

Screenshot_20251216_095652

lint-rules/src/main/java/com/duckduckgo/lint/ui/DaxSecureTextFieldTrailingIconDetector.kt
```diff
@@ -0,0 +1,113 @@
...
+                reportInvalidComposableUsage(trailingIconArgument)
+            }
+        }
+
+        private fun isInvalidComposable(argument: org.jetbrains.uast.UExpression): Boolean {
+            val source = argument.sourcePsi?.text ?: return false
+
+            // Only DaxTextFieldTrailingIcon should be used in the trailingIcon parameter
+            // If the source doesn't contain it, then it's invalid
+            return !source.contains("DaxTextFieldTrailingIcon")
```
GerardPaligot: We can probably get member list declared in the scope object class and check if the source contains one member of this list? This will allow us to avoid to edit this linter rule every time we need to edit the scope.

lint-rules/src/main/java/com/duckduckgo/lint/ui/DaxSecureTextFieldTrailingIconDetector.kt
```diff
@@ -0,0 +1,113 @@
...
+            context.report(
+                issue = INVALID_DAX_SECURE_TEXT_FIELD_TRAILING_ICON_USAGE,
+                location = context.getLocation(arg),
+                message = INVALID_DAX_SECURE_TEXT_FIELD_TRAILING_ICON_USAGE.getExplanation(TextFormat.RAW),
+            )
+        }
+    }
+
+    companion object {
+        val INVALID_DAX_SECURE_TEXT_FIELD_TRAILING_ICON_USAGE = Issue
```
GerardPaligot, line 99: If we improve our implementation to be aware about all member in the scope class, we'll need to edit this issue

lint-rules/src/main/java/com/duckduckgo/lint/ui/DaxTextFieldTrailingIconDetector.kt
```diff
@@ -0,0 +1,111 @@
...
+                reportInvalidComposableUsage(trailingIconArgument)
+            }
+        }
+
+        private fun isInvalidComposable(argument: org.jetbrains.uast.UExpression): Boolean {
+            val source = argument.sourcePsi?.text ?: return false
+
+            // Only DaxTextFieldTrailingIcon should be used in the trailingIcon parameter
+            // If the source doesn't contain it, then it's invalid
+            return !source.contains("DaxTextFieldTrailingIcon")
```
GerardPaligot: We can probably get member list declared in the scope object class and check if the source contains one member of this list? This will allow us to avoid to edit this linter rule every time we need to edit the scope.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
mikescamell: In this case my opinion would be to match existing behaviour and keep it white 👍 

Edit: Or even make it blue so it matches the cursor colour we have in the design system, this seems to one of the only places it's black/white?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
mikescamell: Could we consider creating a stateful version of this Composable alongside this one? That way we have one where callers can deal with the state and one where callers don't care and it just shows/hides.

We could even make DaxSecureTextField internal for the time being, and then if there is a need to expose this version in future when we eventually migrate then we can consider it.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
landomen: > Wouldn't this be solved by setting enabled to false when in ReadOnly state?

@mikescamell Setting `enabled=false` and `readOnly=true` does make it not focus like we want. However, there are three side-effects: longer text is not scrollable, you can't copy text and the text field won't be picked up by accessibility services. I think those tradeoffs are not worth it.

These are descriptions of the fields from the documentation. I think Compose behavior is the correct one and the one we should prefer over the XML one. But let me know if you feel different.
- **Enabled:** _"When false, this component will not respond to user input, and it will appear visually disabled and disabled to accessibility services."_

- **ReadOnly:** _"When true, the text field cannot be modified. However, a user can focus it and copy text from it. Read-only text fields are usually used to display pre-filled forms that a user cannot edit."_

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
landomen: I've changed it so it's white/black for now, same as the xml version. You're right @mikescamell, the cursor is blue in all other cases, just for this error case it's white/black. It could be intentional though, the blue cursor on mostly red text field does look a bit off to my untrained eye 😅  

Screenshot 2025-12-18 at 17 01 32

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
landomen: Yes, good catch. Updated!

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,540 @@
...
+    ) {
+        IconButton(
+            onClick = { onClick?.invoke() },
+            enabled = enabled,
+            modifier = modifier,
+        ) {
+            Icon(
+                painter = painter,
+                contentDescription = contentDescription,
+                tint = DuckDuckGoTheme.iconColors.primary,
```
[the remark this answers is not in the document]
landomen: Behavior is same as in the XML version

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
landomen: @malmstein Any idea why the xml input field reveals the password automatically when clicked?

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/textinput/ComponentTextInputFragment.kt
mikescamell: SGTM @landomen, lets go with it 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxSecureTextField.kt
landomen: Good idea, Mike. I made the current stateless version internal for now, and created a public stateful version for ease of use. If there are cases in the future where we need the stateless version, we can make it public then.

lint-rules/src/main/java/com/duckduckgo/lint/ui/DaxSecureTextFieldTrailingIconDetector.kt
landomen: I've updated the detector so it works with any additional new composable we declare in the scope. Let me know if this is what you had in mind

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/textfield/DaxTextField.kt
```diff
@@ -0,0 +1,546 @@
...
+            )
+        }
+    }
+
+    @Composable
+    fun SomeComposable() {
+    }
```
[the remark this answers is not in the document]
landomen: Removed

landomen:
@mikescamell @GerardPaligot I've addressed the open comments, please take another look

GerardPaligot (approved):
LGTM! 🎉 Congrats @landomen for this contribution

PR #7249, opened 2025-11-29
Fix: Wrong custom tab domain

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1212224936885750?focus=true

### Description

This PR fixes a bug when the wrong domain is displayed if a user navigates to a different domain URL within a custom tab. 

See the recording:

https://github.com/user-attachments/assets/b11f036a-cbcb-4dab-a44e-30ec365d501e

Until now, the domain was only updated initially and when the page title was updated, however, the URL used was not the one from the WebView but the one in `Site`, which was not the latest one. That's why the current domain shown was always one step behind the actual one.

I've refactored the custom tab view state updates so that the domain is not set ad-hoc, but rather whenever an actual navigation change occurs. Since the current logic used a domain that may be invalid for determining the Duck Player state, I moved to the ViewState flow, as well.

### Steps to test this PR

- [x] Open an app that opens a custom tab with a website that has links to different domain sites
- [x] Tap on a link to navigate to a different domain
- [x] Notice the domain shown in the toolbar is correct

### DIFF
--- a/app/src/test/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModelTest.kt
+++ b/app/src/test/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModelTest.kt
@@ -2133,6 +2133,65 @@ class OmnibarLayoutViewModelTest {
     }
 
     @Test
+    fun `when custom tab URL changes to simulated YouTube NoCookie and DuckPlayer enabled then showDuckPlayerIcon is true`() = runTest {
+        val youtubeNoCookieUrl = "http://youtube-nocookie.com/videoID=1234"
+        val duckPlayerUrl = "duck://player/1234"
+
+        whenever(duckPlayer.isSimulatedYoutubeNoCookie(any())).thenReturn(true)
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/duckduckgo/app/browser/BrowserTabViewModelTest.kt
```diff
@@ -2301,37 +2300,6 @@ class BrowserTabViewModelTest {
             assertEquals(title, command.savedSiteChangedViewState.savedSite.title)
         }
 
-    @Test
```
CrisBarreiro, line 2304: Can we test these cases in [OmnibarLayoutViewModelTest.kt](https://github.com/duckduckgo/Android/pull/7249/files#diff-84e19dbcb053ce7da73a6d0035829882b8f84d06cce227e0e95253df8eaf0216)?

0nko: Good call! I've [added the tests back](https://github.com/duckduckgo/Android/pull/7249/commits/f3ef66a42de367935f285ff7cb7dbd8ddc204465).

PR #7370, opened 2025-12-17
Create DaxBottomSheetDialog composable components

GerardPaligot:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202857801505092/task/1211659112661228?focus=true

### Description

* Creating `DaxBottomSheetDialog` to customize M3 `ModalBottomSheet` component with DuckDuckGo theme
* Creating variants `DaxPromoBottomSheetDialog` and `DaxPromoLayoutBottomSheetDialog`
* Creating variants `DaxActionBottomSheetDialog`

### Steps to test this PR

_Use bottom sheet developed with Compose_
- [ ] Open the application
- [ ] Go to "Android Design System Preview" item
- [ ] Go to "Dialogs" tab
- [ ] Scroll to "Bottom Sheet Dialogs" section
- [ ] Compare the view and compose components with a tap on buttons

### UI changes

> **Important notes about differences between XML and Compose**
> - Shadow follow the design specification described in Figma
> - Item List component isn't developed yet. The item used in the Compose version is the M3 which is higher than ours.
> - Buttons used in the dialog fragment or in the promo bottom sheet will be replaced by official buttons when they'll be developed

| XML | Compose |
| ------ | ----- |
image | image
image | image
image | image
image | image
image | image

---

> [!NOTE]
> Adds Compose bottom sheet building blocks and integrates them into the internal Dialogs preview for parity with existing View dialogs.
> 
> - New composables: `DaxBottomSheetDialog` (base), `DaxActionBottomSheetDialog`, `DaxPromoBottomSheetDialog` with previews and configurable state/gestures
> - Temporary Compose UI primitives: `Large/SmallPrimaryButton`, `Large/SmallSecondaryButton`, `Large/SmallGhostButton`, and `DaxOneLineListItem`
> - Update `DialogsFragment` and `fragment_components_dialogs.xml` to add Compose counterparts (via `ComposeView`) for action/promo sheets, theme-aware setup, and dismiss handling using `rememberModalBottomSheetState` + coroutine scope
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 657d111ff2190afd78cf9fc9ae8741a0cf34828f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
+++ b/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
@@ -85,7 +85,7 @@ object DaxBottomSheetDefaults {
         get() = DuckDuckGoTheme.colors.text.primary
 
     val scrimColor: Color
-        @Composable get() = Black.copy(alpha = .68f)
+        @Composable get() = Black.copy(alpha = .6f)
 
     val shape = RoundedCornerShape(topStart = 12.dp, topEnd = 12.dp)
 }


android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxActionBottomSheetDialog.kt
```diff
@@ -0,0 +1,299 @@
...
+ * @param sheetGesturesEnabled Controls whether the bottom sheet can be interacted with via
+ * touch gestures.
+ * @param verticalArrangement The vertical arrangement of the items in the LazyColumn.
+ * @param horizontalAlignment The horizontal alignment of the items in the LazyColumn.
+ * @param userScrollEnabled Controls whether the user can scroll the LazyColumn.
+ * @param content The content of the bottom sheet, defined as a [LazyListScope].
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DaxActionBottomSheetDialog(
```
mikescamell, line 69: 🔍We should add links to Figma and Asana task representing the component like DaxText, same for the Promo version 👍

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/dialogs/DialogsFragment.kt
```diff
@@ -351,6 +381,50 @@ class DialogsFragment : Fragment() {
...
+                                onDismissRequest = {
+                                    scope.launch { sheetState.hide() }.invokeOnCompletion {
+                                        if (!sheetState.isVisible) {
+                                            showBottomSheet = false
+                                        }
+                                    }
+                                },
+                                content = {
+                                    item {
+                                        ListItem(
```
mikescamell: 🔍 Might be worth adding TODOs here to come back and update these with our own OneLineListItem when they are created. Maybe we can also mention it in the ListItem Asana task for whoever ends up picking it up as well to come and update it 👍

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/dialogs/DialogsFragment.kt
```diff
@@ -372,6 +446,63 @@ class DialogsFragment : Fragment() {
...
+                                                    painter = painterResource(CommonR.drawable.ic_add_24),
+                                                    contentDescription = null,
+                                                )
+                                            },
+                                            colors = colors,
+                                        )
+                                    }
+                                },
+                            )
+                        }
```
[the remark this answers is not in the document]
mikescamell: @GerardPaligot I think Cursor is right 😅

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
```diff
@@ -0,0 +1,88 @@
...
+ * @param containerColor The background color of the bottom sheet container.
+ * @param contentColor The content color of the bottom sheet.
+ * @param scrimColor The color of the scrim that obscures content outside the bottom sheet.
+ * @param sheetGesturesEnabled Controls whether the bottom sheet can be interacted with via
+ * touch gestures.
+ * @param content The content of the bottom sheet.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DaxBottomSheetDialog(
```
mikescamell, line 53: 🔍I think we should make DaxBottomSheetDialog internal for now. If we find a reason to open it up in future we can decide then to expose it. I'd like us to lock down the generic base versions to encourage people to use what is already there.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
```diff
@@ -0,0 +1,88 @@
...
+    onDismissRequest: () -> Unit,
+    modifier: Modifier = Modifier,
+    sheetState: SheetState = rememberModalBottomSheetState(),
+    shape: Shape = DaxBottomSheetDefaults.shape,
+    containerColor: Color = DaxBottomSheetDefaults.containerColor,
+    contentColor: Color = DaxBottomSheetDefaults.contentColor,
+    scrimColor: Color = DaxBottomSheetDefaults.scrimColor,
```
mikescamell, line 60: 🔍I noticed the scrim colour for the compose versions seem to be lighter than our View equivalent. We should match the View version unless we have a good reason not too 👍

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/dialogs/DialogsFragment.kt
```diff
@@ -395,6 +526,37 @@ class DialogsFragment : Fragment() {
...
+                                onDismissRequest = {
+                                    scope.launch { sheetState.hide() }.invokeOnCompletion {
+                                        if (!sheetState.isVisible) {
+                                            showBottomSheet = false
+                                        }
+                                    }
+                                },
+                                buttons = {
+                                    SmallGhostButton(text = "Button", onClick = { showBottomSheet = false })
+                                    SmallPrimaryButton(text = "Button", onClick = { showBottomSheet = false })
```
mikescamell: 🔍 For all the dismissing via the buttons, I think we should be passing sheetState and calling `sheetState.hide()` instead of setting this property to false, otherwise it just abruptly disappears.

We should do this for all the Promo versions we display on the ADS screen

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/buttons/GhostButton.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import com.duckduckgo.common.ui.compose.text.DaxText
+import com.duckduckgo.common.ui.compose.theme.DuckDuckGoTheme
+
+/**
+ * TODO When buttons will be available in the design system, replace this with the official GhostButton composable.
```
mikescamell, line 29: 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxActionBottomSheetDialog.kt
```diff
@@ -0,0 +1,299 @@
...
+                DaxActionTitleBottomSheetDialog(text = title)
+            }
+            LazyColumn(
+                state = lazyColumnState,
+                verticalArrangement = verticalArrangement,
+                horizontalAlignment = horizontalAlignment,
+                userScrollEnabled = userScrollEnabled,
+                content = content,
+            )
+            Spacer(modifier = Modifier.height(DaxActionBottomSheetDefaults.sheetVerticalPadding))
```
mikescamell: ❓ Out of curiosity, why did you go with a combination of Spacer rather than using the `LazyColumn` to handle display and padding?

E.g. where `contentPadding` would be a new default: `contentPadding: PaddingValues = DaxActionBottomSheetDefaults.sheetContentPadding,`

 
```suggestion
            LazyColumn(
                state = lazyColumnState,
                verticalArrangement = verticalArrangement,
                horizontalAlignment = horizontalAlignment,
                userScrollEnabled = userScrollEnabled,
                contentPadding = contentPadding,
                content = {
                    if (title != null) {
                        item {
                            DaxActionTitleBottomSheetDialog(text = title)
                        }
                    }

                    content()
                },
            )
```

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxPromoBottomSheetDialog.kt
```diff
@@ -0,0 +1,267 @@
...
+ * @param sheetState The state of the bottom sheet.
+ * @param sheetGesturesEnabled Controls whether the bottom sheet can be interacted with via
+ * touch gestures.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DaxPromoBottomSheetDialog(
+    title: String?,
+    description: String,
+    painter: Painter?,
```
mikescamell: 🔍 Maybe we should be more specific with the naming about what this painter used for? In the View world it's called `bottomSheetPromoIcon`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxActionBottomSheetDialog.kt
```diff
@@ -0,0 +1,299 @@
...
+
+@Composable
+private fun DaxActionTitleBottomSheetDialog(
+    text: String,
+    modifier: Modifier = Modifier,
+    contentColor: Color = DaxActionBottomSheetDefaults.contentColor,
+    style: TextStyle = DaxActionBottomSheetDefaults.style,
+) {
+    Box(modifier = modifier.padding(DaxActionBottomSheetDefaults.sheetActionTitleContentPadding)) {
+        Text(
```
mikescamell: 🔍I think we should be using DaxText here.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxPromoBottomSheetDialog.kt
```diff
@@ -0,0 +1,267 @@
...
+            }
+        },
+    )
+}
+
+object DaxPromoBottomSheetDefaults {
+    internal val sheetPadding = 24.dp
+    internal val sheetVerticalGap = 16.dp
+    internal val sheetHorizontalGap = 16.dp
+    internal val imageSize = 50.dp
```
mikescamell: ❓ Why is the imageSize 50dp when the View version is 48dp?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxPromoBottomSheetDialog.kt
```diff
@@ -0,0 +1,267 @@
...
+ * @param onDismissRequest Callback invoked when the user tries to dismiss the bottom sheet.
+ * @param modifier The [Modifier] to be applied to this bottom sheet.
+ * @param sheetState The state of the bottom sheet.
+ * @param sheetGesturesEnabled Controls whether the bottom sheet can be interacted with via
+ * touch gestures.
+ * @param content The content of the promo.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DaxPromoLayoutBottomSheetDialog(
```
mikescamell: ❓What was the reason for creating this Layout Composable rather than doing all the content here in `DaxPromoBottomSheetDialog`? Was it just to try to make it easier to read? Are there instances in the codebase where we need don't use a title or description? 

I'm not sure I see the advantage of keeping this and not just having one composable, but if we do we should make this private, I don't think we want people to be able to use this version.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/dialogs/DialogsFragment.kt
GerardPaligot: Yes, we can but without `ListItem` component here, it is difficult to test our bottom sheet Compose component.
Because we often need to use a list item component in our bottom sheet, maybe we should put this contribution in pause and wait the list item component? 🤔

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxActionBottomSheetDialog.kt
GerardPaligot: It depends if the title should scroll with the bottom sheet content or not, I'll ask to designers to know.

But if it is the case, this will get more complicate to integrate the title in the component because we can't just invoke the content parameter here like a composable because it isn't a component and should be pass to the content parameter of the LazyColumn.

According to the clarification from designers team, I'll suggest a new implementation here.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
GerardPaligot: Yes, I forgot to ask but impossible to find the actual scrim color in the XML version of the component. Do you know where I can find it?
Note my implementation is following the design specification, I don't know if we should update the scrim color of the XML implementation or align the Compose version to the XML one.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxPromoBottomSheetDialog.kt
GerardPaligot: I think you're right. I tend to prefer api slot contract for Compose components because it is much more flexible than a very opinionated contract with primitive types but it is what is expected for our project. I'll remove this component and keep only the component above.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxPromoBottomSheetDialog.kt
GerardPaligot: On Figma, the size is defined to 50. I'll update it to align with the XML value.

Screenshot 2026-01-06 at 11 23 27

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/dialogs/DialogsFragment.kt
GerardPaligot: Another suggestion is to do like the button component: Create a simple version of the list item component in this contribution, which is following our design spec, and use it in this contribution with a comment to explain that it should be replaced by the future real implementation. I think it is probably the best choice to do. What do you think?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxPromoBottomSheetDialog.kt
mikescamell: Yeah I think we lean toward strictness for now and later if we feel the need to, we open up for customization.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/dialogs/DialogsFragment.kt
mikescamell: I'm not against using ListItem as a placeholder for now. I just think it's worth adding a TODO where we've used it and also making a comment on our ListItem Asana task that these should be updated when they are created, that's all 👍 

You did similar here for the buttons: https://github.com/duckduckgo/Android/pull/7370#discussion_r2661336575

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
mikescamell: As we discussed on MM, we're trying to match the View versions for now and we can circle back later after we're complete to evaluate all the Figma design discrepancies 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
GerardPaligot: I found a usecase where this variant is necessary. In my [recent project](https://github.com/duckduckgo/Android/pull/7451), the bottom sheet used isn't an action or promo bottom sheet but a generic one that we need to use to show the new browser contextual menu.

I can convert it in internal but It'll become necessary very quickly when we'll start to use Compose in our development.

GerardPaligot:
* **#7370** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7370?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7370?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/sheets/DaxBottomSheetDialog.kt
mikescamell: Perfect timing 😁 Ok makes sense to me 👌

GerardPaligot:
Big thanks for the review Mike!

PR #6760, opened 2025-09-08
Fix incorrect handling of chrome://... and/or edge://... type of URLs

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1211244094852892?focus=true

### Description
Fix the way that we handle non http URI requests. The first attempts fixed it but also introduced a bug because we should only look at user initiated requests, and not redirects.

This fix should do that

### Steps to test this PR
- [ ] repeat tests in https://github.com/duckduckgo/Android/pull/6687
- [ ] ensure the [side effect issue](https://app.asana.com/1/137249556945/project/1206777341262243/task/1211202801652410) doesn't repro anymore

PR #7382, opened 2025-12-17
Improve intent:// handling when the user clicks on an intent link

catalinradoiu:
Task/Issue URL: 

### Description

This PR improves intent validation logic by considering the package parameter when resolving activities. This is a follow up on the fix implemented in this PR: https://github.com/duckduckgo/Android/pull/7379

When the user is in a custom tab and clicks on a link that has an intent url, if there is no app to handle that intent, we are still considering it as an app link.

This is done because in that case there is a user initiated action and not a redirect. The current logic is not taking that scenarion into consideration. This improves the way this is handled, as instead of showing the error page, it open the google play store for that app. 

This covers the edge case when the user is on a webpage that has an intent embeded in a link and the intent has a package but it cannot be handled by any app on the device.

### Steps to test this PR

_External link handling_
- [ ] Repeat tests in https://github.com/duckduckgo/Android/pull/6760
- [ ] Check the scenario below

1. Go to Developer Settings and open a custom tab with the following url: https://codepen.io/pen/
2. In the HTML box paste the following code:
`<button onclick="window.open('intent://open/#Intent;scheme=dd;package=com.airbnb.android;end', '_syste')">Open other APP in STORE</button>`
3. After the button is rendered click on the button
4. The playstore should be open

### UI changes
| Before  | After |
| ------ | ----- |
|[before.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/a237c38d-f1c9-4ac9-9339-a0209ab72626.mp4)|[after.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/94bbe498-b2f7-45bb-8ede-540a148e94bd.mp4)|

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/SpecialUrlDetector.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/SpecialUrlDetector.kt
@@ -61,6 +61,7 @@ class SpecialUrlDetectorImpl(
     ): UrlType {
         val uriString = uri.toString()
 
+        logcat(tag = "RadoiuC") { "Uri: $uri scheme: ${uri.scheme}" }
         return when (val scheme = uri.scheme) {
             TEL_SCHEME -> buildTelephone(uriString)
             TELPROMPT_SCHEME -> buildTelephonePrompt(uriString)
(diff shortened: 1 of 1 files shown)


catalinradoiu:
* **#7382** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7382?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

PR #7436, opened 2026-01-05
Standardize url bar icon state to match other platforms

laghee:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1201870266890790/task/1210936487262413?focus=true

### Description

Our other browsers handle these states similarly for the most part, so we should adapt Android to match other platforms. As it stands, we get breakage reports from Android users whenever the privacy shield is showing the red warning dot, which means we get a lot of non-actionable feedback for people's router IPs and for sites we're trying to speculatively mitigate breakage on (turning off blocking to see if it reduces reports).

### Steps to test this PR
- [x] Navigate to `noaprints.com` and `marvel.com` (both have protection disabled in the config)
- [x] Confirm for each that you see the normal green privacy shield state for both (though if you click into the dashboard, the fact that protections are disabled is shown & the menu item shows as "enable privacy protection"
- [x] Navigate to your local network address(es) and confirm that you see the globe icon rather than the UNPROTECTED state (privacy shield with a red dot)

### UI changes
| Before  | After |
| ------ | ----- |
marvelOld|marvelNew|
|localOld|localNew|

---

> [!NOTE]
> **Medium Risk**
> Changes how privacy/shield and leading icon states are derived (user-visible and used by reporting), gated by a remote toggle but touching core URL classification and privacy-state logic.
> 
> **Overview**
> Standardizes omnibar leading icon/shield behavior behind a new remote feature toggle `standardizedLeadingIcon`.
> 
> When enabled, `OmnibarLayoutViewModel` shows the **Globe** icon for localhost/private-network/file URLs (instead of the privacy shield state), and `SiteMonitor.privacyProtection()` only returns **UNPROTECTED** for *user-initiated* allowlisting (not remote-config exceptions), reducing misleading red-dot states. Adds a new `Uri.isLocalUrl` helper (IPv4/IPv6 ranges + localhost, no DNS lookups) with extensive tests, and updates constructors/tests to inject the new toggle through `SiteFactoryImpl`/`SiteMonitor`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d5bc5221add6278f786a828467111e5f9cc32b77. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
@@ -478,14 +478,14 @@ class OmnibarLayoutViewModel @Inject constructor(
             NewTab -> Search
             else -> {
                 when {
-                  hasFocus -> Search
-                  serpLogoUrl != null -> EasterEggLogo(logoUrl = serpLogoUrl, serpUrl = url, isFavourite = isFavouriteSerpLogo)
-                  shouldShowDaxIcon(url) -> Dax
-                  shouldShowDuckPlayerIcon(url) -> DuckPlayer
-                  // Show globe icon for localhost and private network addresses
-                  standardizedLeadingIconToggle.self().isEnabled() && isLocalUrl(url) -> Globe
-                  url.isEmpty() -> Search
-                  else -> PrivacyShield
+                    hasFocus -> Search
+                    serpLogoUrl != null -> EasterEggLogo(logoUrl = serpLogoUrl, serpUrl = url, isFavourite = isFavouriteSerpLogo)
+                    shouldShowDaxIcon(url) -> Dax
+                    shouldShowDuckPlayerIcon(url) -> DuckPlayer
+                    // Show globe icon for localhost and private network addresses
+                    standardizedLeadingIconToggle.self().isEnabled() && isLocalUrl(url) -> Globe
+                    url.isEmpty() -> Search
+                    else -> PrivacyShield
                 }
             }
         }
@@ -496,7 +496,7 @@ class OmnibarLayoutViewModel @Inject constructor(
             url.toUri().isLocalUrl
         } catch (e: Exception) {
             false
-          }
+        }
     }
 
     private fun getCurrentSerpLogoUrl(


common/common-utils/src/main/java/com/duckduckgo/common/utils/UriExtension.kt
```diff
@@ -64,6 +64,31 @@ val Uri.hasIpHost: Boolean
...
+        if (ipv4Parts.size != 4) return false
+
+        val octets = ipv4Parts.mapNotNull { it.toIntOrNull() }
+        if (octets.size != 4 || octets.any { it !in 0..255 }) return false
+
+        val (first, second) = octets
+        if (first == 127) return true
+        if (first == 10) return true
+        if (first == 172 && second in 16..31) return true
+        if (first == 192 && second == 168) return true
```
jonathanKingston: Probably simpler to rely on the libraries themselves (will support ipv6 too). Assuming we don't already have a method like this?

```kotlin
val Uri.isLocalUrl: Boolean
    get() {
        if (scheme == "file") return true

        val h = host?.lowercase() ?: return false
        if (h == "localhost") return true

        val addr = runCatching { InetAddresses.parseNumericAddress(h) }.getOrNull() ?: return false
        return addr.isLoopbackAddress || addr.isSiteLocalAddress
    }
```

jonathanKingston: (there's a bunch of others here folk use like `blah.local` and `localhost.localhost`)

jonathanKingston: @aitorvs do you think we could add this as a shared method to https://github.com/duckduckgo/url_predictor as a `is_local`?

karlenDimla: @laghee The current solution you have only considers ipv4. Seem like what @jonathanKingston is suggesting in https://github.com/duckduckgo/Android/pull/7436/files#r2662628187 is a more complete solution so I think we should update to that instead.

app/src/main/java/com/duckduckgo/app/browser/omnibar/StandardizedLeadingIconFeatureToggle.kt
```diff
@@ -0,0 +1,54 @@
...
+    scope = AppScope::class,
+    featureName = "standardizedLeadingIcon",
+)
+interface StandardizedLeadingIconFeatureToggle {
+    /**
+     * Root feature toggle for standardized leading icon behavior.
+     *
+     * @return Toggle that controls whether to use the new icon determination logic
+     */
+    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
```
karlenDimla: @laghee NIT: You can have it enabled by default and then disable it as a kill switch later on if something goes wrong.

laghee:
```suggestion
    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
```

PR #7438, opened 2026-01-06
Fix: Fire dialog null binding access cases

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1212674866476209?focus=true

### Description

This PR fixes 2 potential instances when the binding might be null.

If the dialog's view is destroyed while the coroutine is still running on the IO dispatcher, when the coroutine completes and calls onFireDialogClearAllEvent, _binding is already null, causing the NPE.

### Steps to test this PR

QA-optional since the issue's a race condition that's not easy to reproduce.

---

> [!NOTE]
> Prevents null-binding access in `LegacyFireDialog` during async/animation callbacks to avoid race-condition crashes.
> 
> - Use `_binding?.fireAnimationView?.let { ... }` in the animator update to safely adjust animation speed
> - Gate `onFireDialogClearAllEvent` logic with `_binding != null` before touching views; otherwise proceed to restart
> - Mitigates NPE when the dialog view is destroyed while background work completes
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fad9d4dc353a7e2f3c0fa9aa1af916cbd5ed25c5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/view/LegacyFireDialog.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/view/LegacyFireDialog.kt
@@ -104,7 +104,7 @@ class LegacyFireDialog : BottomSheetDialogFragment(), FireDialog {
 
     private val accelerateAnimatorUpdateListener = object : ValueAnimator.AnimatorUpdateListener {
         override fun onAnimationUpdate(animation: ValueAnimator) {
-            binding.fireAnimationView.let {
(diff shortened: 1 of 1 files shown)


joshliebe (approved):
LGTM

PR #7453, opened 2026-01-08
Fire dialog: New fire dialog shown pixel

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1212705988619991?focus=true

### Description

This PR adds a pixel that measures the fire dialog display.

### Steps to test this PR

_Browser_
- [ ] Start the app and tap on the Fire icon
- [ ] Verify the `m_fire_dialog_shown` pixel is fired
- [ ] Rotate the device
- [ ] Verify the pixel is not fired

_Settings_
- [ ] Go to Settings -> Data clearing
- [ ] Tap on Clear tabs and data
- [ ] Verify the `m_fire_dialog_shown` pixel is fired

_Tab manager_
- [ ] Go to the Tab manager
- [ ] Tap on Fire icon
- [ ] Verify the `m_fire_dialog_shown` pixel is fired

---

> [!NOTE]
> Introduces a new metric for fire dialog visibility and wires it into both implementations.
> 
> - Adds `m_fire_dialog_shown` definition and `AppPixelName.FIRE_DIALOG_SHOWN`; fires in `GranularFireDialogViewModel.onShow()` (guarded to once) and in `LegacyFireDialog` on first creation
> - Updates `PixelParamRemovalInterceptor` to remove `atb` from `FIRE_DIALOG_SHOWN`
> - Hooks `GranularFireDialog` to call `viewModel.onShow()` in `onStart`; preserves fragment result events
> - Adds unit tests verifying the pixel is fired (and only once) and performs minor test cleanups
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5b34af268aa0459eda49b9792b1ad6178858378b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7460, opened 2026-01-09
Fire dialog: Refactoring of the granular dialog and data clearing logic

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1212685654036724?focus=true

### Description

The granular data-clearing dialog will be shelved (at least temporarily), however, the new data clearing logic is still needed. We need to separate the two into different feature flag so that granular automatic data clearing can be used without the granular dialog.

We also want to reuse the existing fire dialog UI but use a new infrastructure (ViewModel), so that the logic can be tested, and eventually extended to include single-tab burning.

### Steps to test this PR

_Legacy dialog_
- [x] Make sure both the `useGranularFireDialog` and `improvedDataClearingOptions` flags are disabled
- [x] Open the app and tap on the Fire button
- [x] Verify the data clearing works as expected

_Improved data clearing (refactored dialog)_
- [x] Go to Settings -> Feature flag settings and enable the `improvedDataClearingOptions` flag
- [x] Go back to the browser and tap on the Fire button
- [x] Verify the data clearing works just like the legacy dialog (data + tabs + chats (optional, if enabled) cleared)

---

> [!NOTE]
> Separates UI and logic and modernizes the Fire flow while keeping legacy behavior available via flags.
> 
> - Introduces `improvedDataClearingOptions` and `granularFireDialog` toggles; removes usages of `moreGranularDataClearingOptions`
> - New `NonGranularFireDialog` + `NonGranularFireDialogViewModel` reuse legacy UI with ViewModel-driven clearing, pixeling, and animation handling
> - `FireDialogProvider` now selects `GranularFireDialog`, `NonGranularFireDialog`, or `LegacyFireDialog` based on flags
> - `AutomaticDataClearer`, `DataClearingWorker`, `BrowserActivity`, and `ClearDataNotification` migrated to check `improvedDataClearingOptions` and use manual/automatic clearing accordingly
> - Enhances `FireDataStore`: default/legacy migration logic, manual/automatic options flows, and inclusion of `DUCKAI_CHATS` when enabled
> - Refactors `GranularFireDialog` control flow (command handling, safe binding, restart logic)
> - Adds/updates tests: Android tests for automatic clearer/notification; unit tests for `FireDataStore` and `NonGranularFireDialogViewModel`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ff46e85370967c8172cc92f7f61480dfb2d248fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/view/FireDialogProvider.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/view/FireDialogProvider.kt
@@ -21,8 +21,7 @@ import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesBinding
 import dagger.SingleInstanceIn
-import kotlinx.coroutines.async
-import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.withContext
 import javax.inject.Inject
 
 /**
@@ -48,18 +47,12 @@ class FireDialogProviderImpl @Inject constructor(
     private val dispatcherProvider: DispatcherProvider,
 ) : FireDialogProvider {
 
-    override suspend fun createFireDialog(): FireDialog = coroutineScope {
-        val useImprovedDataClearing = async(dispatcherProvider.io()) {
-            androidBrowserConfigFeature.improvedDataClearingOptions().isEnabled()
-        }
-
-        val useGranularFireDialog = async(dispatcherProvider.io()) {
-            androidBrowserConfigFeature.granularFireDialog().isEnabled()
-        }
-
+    override suspend fun createFireDialog(): FireDialog = withContext(dispatcherProvider.io()) {
         when {
-            useGranularFireDialog.await() -> GranularFireDialog.newInstance()
-            useImprovedDataClearing.await() -> NonGranularFireDialog.newInstance()
+            androidBrowserConfigFeature.granularFireDialog().isEnabled() ->
+                GranularFireDialog.newInstance()
+            androidBrowserConfigFeature.improvedDataClearingOptions().isEnabled() ->
+                NonGranularFireDialog.newInstance()
             else -> LegacyFireDialog.newInstance()
         }
     }


0nko:
* **#7460** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7453** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7453?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/global/view/FireDialogProvider.kt
```diff
@@ -47,15 +48,19 @@ class FireDialogProviderImpl @Inject constructor(
     private val dispatcherProvider: DispatcherProvider,
 ) : FireDialogProvider {
 
-    override suspend fun createFireDialog(): FireDialog {
-        val isGranularMode = withContext(dispatcherProvider.io()) {
-            androidBrowserConfigFeature.moreGranularDataClearingOptions().isEnabled()
+    override suspend fun createFireDialog(): FireDialog = coroutineScope {
```
CDRussell: curious why the introduction of `coroutineScope` and two `async` / `await` ? 

seems like it all could have been done like before using just a single `withContext`.  Do we really need the feature flag checks to be executed in parallel (seems overkill)?

Not blocking; but maybe you can reply here _why_ you're doing it that way because i'm curious. (if there's no good reason for it, removing them would make the code simpler to read)

CDRussell (approved):

app/src/main/java/com/duckduckgo/app/global/view/FireDialogProvider.kt
0nko: The idea was that since each FF access is potentially blocking, this would speed up the dialog creation by accessing them in parallel. But you're right, I guess it's not really worth making it complicated, I'll change it to a single coroutine.

PR #7525, opened 2026-01-19
GHA: Checkout and assemble flavour

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1157893581871903/task/1212838282644677?focus=true

### Description
This PR adds a composite action to build and assemble flavours

---

> [!NOTE]
> Adds a reusable CI building block and streamlines E2E pipelines.
> 
> - **New composite action** `./.github/actions/checkout-and-assemble` to checkout code, set up JDK/Gradle, decode signing secrets, assemble `play`/`internal` releases (with optional flags), and output `play_apk_path`/`internal_apk_path`.
> - **Refactors E2E and release workflows** to use this action, removing duplicated setup/build steps; pass `flavours`, optional `commit`, and `gradle_flags`; update Maestro steps to use action outputs for `maestro_app_file`.
> - **Full suite workflow** now uploads both Internal and Release binaries first and runs tests via `maestro_app_binary_id` instead of local APK paths.
> - **Adds Maestro config** `.maestro/binary_upload/binary_upload_maestro.yaml` to support the binary upload flow.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a1bb3c942d1ce042ca36929a15cd8bff5a66ea52. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e-nightly-full-suite.yml
+++ b/.github/workflows/e2e-nightly-full-suite.yml
@@ -33,6 +33,40 @@ jobs:
           release_key: ${{ secrets.FAKE_RELEASE_KEY }}
           gradle_encryption_key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
 
+      - name: Upload Internal Binary to Maestro
+        id: maestro-upload-internal-binary
+        uses: ./.github/actions/maestro-cloud-asana-reporter
+        timeout-minutes: 120
+        with:
+          maestro_api_key: ${{ secrets.ROBIN_API_KEY }}
+          maestro_project_id: ${{ vars.ROBIN_ANDROID_PROJECT_ID }}
+          maestro_test_name: binary_internal_upload_${{ github.sha }}
+          maestro_timeout: ${{ vars.ROBIN_TIMEOUT_MINUTES }}
+          maestro_app_file: ${{ steps.assemble.outputs.internal_apk_path }}
+          maestro_api_level: 30
+          maestro_include_tags: binaryUpload
+          asana_pat: ${{ secrets.ASANA_ACCESS_TOKEN }}
+          asana_project_id: ${{ vars.GH_ANDROID_APP_PROJECT_ID }}
+          asana_section_id: ${{ vars.GH_ANDROID_APP_INCOMING_SECTION_ID }}
+          test_suite_name: Internal Binary Upload
+
+      - name: Upload Release Binary to Maestro
+        id: maestro-upload-release-binary
+        uses: ./.github/actions/maestro-cloud-asana-reporter
+        timeout-minutes: 120
+        with:
+          maestro_api_key: ${{ secrets.ROBIN_API_KEY }}
+          maestro_project_id: ${{ vars.ROBIN_ANDROID_PROJECT_ID }}
+          maestro_test_name: binary_release_upload_${{ github.sha }}
+          maestro_timeout: ${{ vars.ROBIN_TIMEOUT_MINUTES }}
+          maestro_app_file: ${{ steps.assemble.outputs.play_apk_path }}
+          maestro_api_level: 30
+          maestro_include_tags: binaryUpload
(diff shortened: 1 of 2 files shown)


malmstein:
* **#7526** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7526?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7525** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7525?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7525?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

CDRussell:
@malmstein i'm liking the look of this. Some questions...

**Customizing which APKs are built each time**
would it make sense to allow workflows to specify which APK they want to build? Right now, looking at the composite script, it's going to build both `internal` and `play` APKs for each workflow that runs but that's more work than we're doing now in these workflows where they often only need one of them to build.

If each workflow could pass to the composite action which type(s) it needs, that would help cut out some wasted builds. 

**If composite action limited to working with fake release keys?**
From the looks of the input parameter names, it seems to suggest it's restricted to the fake cert only (`fake_release_key`, `fake_release_properties` etc...) but each workflow is responsible for what it provides here.
- is it really restricted to fake signing configs only or would it work for production builds if you provided those key, properties etc...? (if so, could rename the input params?)

.github/actions/checkout-and-assemble/action.yml
```diff
@@ -0,0 +1,83 @@
...
+        destination-path: $HOME/jenkins_static/com.duckduckgo.mobile.android/
+
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@v5
+      with:
+        cache-encryption-key: ${{ inputs.gradle_encryption_key }}
+
+    - name: Assemble release APK
+      shell: bash
+      run: ./gradlew assemblePlayRelease -Pforce-default-variant -Pskip-onboarding ${{ inputs.gradle_flags }}
```
CDRussell: In many of the existing workflows we add `-x lint` to this to skip the expensive lint checks (since that has a check elsewhere). I know we could continue to provide that flag manually from each workflow as custom params, but would it make sense to add that to the default params alongside `-Pforce-default-variant -Pskip-onboarding` so that it's always skipped while building these APKs?

Same question for the `assembleInternalRelease`.

malmstein: I think that’s right @CDRussell, there’s no need to have lint in here.

malmstein:
@CDRussell thanks.

Re the input names, it makes sense to change the name to avoid confusion. It’s not a fake release key, we can pass in whatever.

Re the apk customisation. I’d like to be able to build both APKs if needed. This is specifically handy for the full suite of e2e where we can to use two different APKs. That being said, some other workflows only need one APK, so we’d be adding 5 minutes of execution time to those. I’ll think about a way to allow all apks in a build type to be generated, but also allowing for one single flavour.

CDRussell:
> @CDRussell thanks.
> 
> Re the input names, it makes sense to change the name to avoid confusion. It’s not a fake release key, we can pass in whatever.
> 
> Re the apk customisation. I’d like to be able to build both APKs if needed. This is specifically handy for the full suite of e2e where we can to use two different APKs. That being said, some other workflows only need one APK, so we’d be adding 5 minutes of execution time to those. I’ll think about a way to allow all apks in a build type to be generated, but also allowing for one single flavour.

Thinking sideways for a little: if the action only did one build type at a time, would we lose or gain anything  from that? E.g., 
- does it make it easier or harder to get both APKs building in parallel when both are wanted?
- would that add more significantly more overhead calling this action twice with different build types vs calling it once to build both?


(Feel free to ignore questions if not useful, just trying to poke it a little to make sure the setup feels optimal)

PR #7624, opened 2026-01-29
Add learn more translation in ADS module for Page Header component

GerardPaligot:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212015278241917/task/1213020424158489?focus=true 

### Description

Add translation "Learn More" in ADS module for PageHeader component

### Steps to test this PR

N/A

---

> [!NOTE]
> **Low Risk**
> Low risk localization-only change: adds a new string resource across multiple locales with no logic or behavior changes.
> 
> **Overview**
> Adds a new Page Header string resource, `pageHeaderLearnMore`, to `strings-common-ui.xml` in the default `values/` resources and all translated `values-*/` locales so the PageHeader component can display a localized "Learn More" label.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f76f0da5cdca12b80016dff2bf55a50166faf899. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7625, opened 2026-01-29
Create PageHeader Compose component and templates tab in ADS

GerardPaligot:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212015278241917/task/1211670072973939?focus=true

### Description

Create DaxPageHeader component in Compose for settings screens.

### Steps to test this PR

_Check DaxPageHeader_
- [x] Open the application
- [x] Go to the ADS screens from developer settings
- [x] Open the "Templates" tab
- [x] Check PageHeaders component follow our actual XML implementation

### UI changes
image

---

> [!NOTE]
> **Medium Risk**
> Adds new Compose UI components and a new ADS demo tab, plus significantly broadens the custom lint detector logic for `DaxText` color validation, which could introduce new/changed warnings in builds.
> 
> **Overview**
> Adds new Compose UI templates: `DaxPageHeader` (optional icon/subtitle/body/status/learn-more) and a supporting `DaxStatusIndicator`/`Status` enum, with previews.
> 
> Extends the ADS demo by adding a new **Templates** tab (`TemplatesFragment`/`TemplatesPane`) to showcase `DaxPageHeader`, and refactors compose view setup by introducing a `ComposeView.setupThemedComposeView` helper.
> 
> Updates the custom lint rule `InvalidDaxTextColorUsage` to allow `DuckDuckGoTheme.colors.text` (and legacy `textColors`), allow theme static colors, and validate colors referenced via defaults objects; baselines are adjusted accordingly and tests are expanded to cover the new acceptance rules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a918eeaeea85845d2d44ef3c6e2fee2a9ed238d6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
+++ b/android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
@@ -106,7 +106,7 @@ enum class Status {
     Off,
 }
 
-object StatusIndicatorDefaults {
+internal object StatusIndicatorDefaults {
     internal val iconSize = 8.dp
     internal val spacing = 4.dp
     internal val minHeight = 16.dp


GerardPaligot:
* **#7625** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7625?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7624** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7624?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
```diff
@@ -0,0 +1,209 @@
...
+        }
+        if (body != null) {
+            Column(
+                modifier = Modifier.fillMaxWidth(),
+                horizontalAlignment = Alignment.CenterHorizontally,
+            ) {
+                DaxText(
+                    text = body,
+                    style = DaxPageHeaderDefaults.bodyTypography,
+                    color = DaxPageHeaderDefaults.bodyColor,
```
mikescamell, line 120: 🔍Why did you opt to create default textColors for this component when we already have the textColors in the theme? Is this a case of we could more easily swap it out later if we added DaxPageHeaderDefaults as a parameter?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+    active: Boolean,
+    label: String,
+    modifier: Modifier = Modifier,
+) {
+    Row(
+        modifier = modifier
+            .heightIn(min = StatusIndicatorDefaults.minHeight)
+            .semantics(mergeDescendants = true) {
+                this.text = AnnotatedString(label)
+                this.selected = active
```
mikescamell: 🤖CC is telling me it would be preferential this was using stateDescription and I think that makes sense:

```
  The issue is that selected in Compose semantics means the user can select/deselect this item — like a list item in a selectable list, or a tab in a tab bar. TalkBack will announce "selected" or "not selected", implying interactivity.

  But StatusIndicator is read-only — it just displays a status (e.g., "Always on", "Enabled", "Disabled"). The user can't toggle it. Using selected is semantically misleading because it tells accessibility services this element is interactive when it isn't.

  The better fit is stateDescription, which is designed for conveying the current state of non-interactive elements:

  .semantics(mergeDescendants = true) {
      this.text = AnnotatedString(label)
      this.stateDescription = if (active) "Active" else "Inactive"
  }

  With stateDescription, TalkBack would announce something like "Always on, Active" rather than "Always on, selected" — which more accurately reflects a read-only status indicator.
```

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
```diff
@@ -0,0 +1,209 @@
...
+                    textAlign = TextAlign.Center,
+                )
+                if (learnMoreClick != null) {
+                    DaxText(
+                        text = stringResource(R.string.pageHeaderLearnMore),
+                        style = DaxPageHeaderDefaults.bodyTypography,
+                        color = DaxPageHeaderDefaults.learnMoreColor,
+                        textAlign = TextAlign.Center,
+                        modifier = Modifier.clickable(onClick = learnMoreClick),
```
mikescamell: 🤖Should we add `onClickLabel` and `role = Role.Button` to the clickable modifier for a11y?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+                this.text = AnnotatedString(label)
+                this.selected = active
+            },
+        horizontalArrangement = Arrangement.spacedBy(space = StatusIndicatorDefaults.spacing),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        Box(
+            modifier = Modifier
+                .size(StatusIndicatorDefaults.iconSize)
+                .clip(CircleShape)
```
mikescamell, line 71: 🤖I think we can remove this clip modifier if we're providing a shape in the background modifier 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+     */
+    OFF,
+}
+
+object StatusIndicatorDefaults {
+    internal val iconSize = 8.dp
+    internal val spacing = 4.dp
+    internal val minHeight = 16.dp
+
+    internal val disableDotColor: Color
```
mikescamell: 🤖disableDotColor -> disabledDotColor

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
```diff
@@ -0,0 +1,209 @@
...
+                    )
+                }
+            }
+        }
+    }
```
mikescamell, line 137: 🤖"TalkBack announces each child element separately. Non-interactive parts should be merged as a cohesive unit. Add `Modifier.semantics(mergeDescendants = true)` to text-only areas, keep "Learn More" independently focusable" https://developer.android.com/develop/ui/compose/accessibility/merging-clearing#merge-semantics I need to read up on this more 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+ * A simple status indicator with a colored dot and a label.
+ *
+ * @param active Whether the status is active (true) or inactive (false).
+ * @param label The label to display next to the dot.
+ * @param modifier The [Modifier] to be applied to this status indicator.
+ */
+@Composable
+fun StatusIndicator(
+    active: Boolean,
+    label: String,
```
mikescamell: ❓Do we want to be exposing the label? We don't in the View version as there are only 3 possible states. I think we should lock this down until we know otherwise. What do you think?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+            )
+            StatusIndicator(
+                active = true,
+                label = "On",
+            )
+            StatusIndicator(
+                active = false,
+                label = "Off",
+            )
+        }
```
mikescamell: 🔍Should these previews be using the StatusIndicator Composable that takes a Status as a parameter?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+}
+
+object StatusIndicatorDefaults {
+    internal val iconSize = 8.dp
+    internal val spacing = 4.dp
+    internal val minHeight = 16.dp
+
+    internal val disableDotColor: Color
+        @Composable
+        get() = DisabledColor
```
mikescamell: 🔍 Hmm looks like our DisabledColor does not work here for the StatusIndicator. It looks black when it prod it's gray. The View version uses     <item android:color="@color/gray50"  android:state_enabled="false" tools:ignore="InvalidColorAttribute" />

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
```diff
@@ -0,0 +1,209 @@
...
+}
+
+object DaxPageHeaderDefaults {
```
mikescamell: 💡If we're not exposing this then should we just make it private instead of internal for all the properties?

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/ADSComposeTools.kt
```diff
@@ -30,12 +30,20 @@ fun View.setupThemedComposeView(
...
-            DuckDuckGoTheme(isDarkTheme) {
-                content()
-            }
+    findViewById<ComposeView>(id)?.setupThemedComposeView(isDarkTheme, content)
+}
+
+/**
+ * Helper to setup a ComposeView and properly handle its lifecycle.
+ */
+fun ComposeView.setupThemedComposeView(
```
mikescamell, line 39: ❤️

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/templates/TemplatesPane.kt
```diff
@@ -0,0 +1,77 @@
...
+            DaxPageHeader(
+                title = "Private Search",
+                subtitle = "Lorem impsum dolor sit amet",
+                status = Status.ALWAYS_ON,
+                iconHeader = painterResource(CommonR.drawable.ic_privacy_pro_128),
+                body = "DuckDuckGo Private Search is your default search engine, so you can search the web without being tracked.",
+                learnMoreClick = {
+                    Toast.makeText(context, "Learn more clicked", Toast.LENGTH_SHORT).show()
+                },
+            )
```
mikescamell, line 100: 🔍 I think we should all variations of DaxPageHeader so we can see it in all of it's different states, what do you think?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
```diff
@@ -0,0 +1,209 @@
...
+    internal val subtitleColor: Color
+        @Composable
+        get() = DuckDuckGoTheme.colors.text.secondary
+    internal val bodyColor: Color
+        @Composable
+        get() = DuckDuckGoTheme.colors.text.secondary
+    internal val learnMoreColor: Color
+        @Composable
+        get() = DuckDuckGoTheme.colors.brand.accentBlue
+}
```
mikescamell, line 174: 🤖"Add `@ReadOnlyComposable` to every `@Composable get()` in both Defaults objects. Prevents Compose compiler optimization — generates unnecessary restart groups in slot table. Overhead is marginal per composition but is a correctness gap." We already do this for DuckDuckGoTheme and Colors 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+    }
+}
+
+object DaxPageHeaderDefaults {
+    internal val iconHeaderHeight = 96.dp
+    internal val iconHeaderWidth = 128.dp
+    internal val containerVerticalSpacing = 8.dp
+    internal val containerPaddingValues: PaddingValues
```
mikescamell: 🤖"Change from custom getter to stored value: `internal val containerPaddingValues = PaddingValues(all = 24.dp)`. Creates new `PaddingValues` object per recomposition. Trivial GC pressure but inconsistent with other stored `val` declarations."

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Indicates that the status is on.
+     */
+    ON,
+
+    /**
+     * Indicates that the status is off.
+     */
+    OFF,
```
mikescamell: ❓Should we use PascalCase here to match other enums in Compose like `DaxTextFieldLineLimits`?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
```diff
@@ -0,0 +1,175 @@
...
+import com.duckduckgo.mobile.android.R
+
+/**
+ * A status indicator representing one of three states: Always On, On, or Off.
+ *
+ * @param status The [Status] to represent.
+ * @param modifier The [Modifier] to be applied to this status indicator.
+ */
+@Composable
+fun StatusIndicator(
```
mikescamell: ❓I'm wondering if we should name this DaxStatusIndicator to be inline with the other components, or we reserve "Dax" for when we're matching existing Compose components. I lean towards prefixing because when typing you'll get all our components by entering "Dax" but not strong feelings, what do you think?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
GerardPaligot: Should I create a smartling request for that?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
GerardPaligot: `onClickLabel` seems like a bit of overkill, but the `role` is a good idea

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
GerardPaligot: Those are two different things: the shape applied to the background is applied to the color, whereas the shape used by the clip is applied to the component itself. A good practice is to use the shape in both locations to avoid any visual issues

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
GerardPaligot: It's definitely a personal opinion, and I can remove it if you think it's overkill: using a default class allows us to change a component's style more easily, rather than having to understand the implementation and update values in one or multiple places

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
GerardPaligot: I think we shouldn't have a semantic merge at the root level but it could be interesting to apply it on the Column with the body and Learn more texts. I'll update this!

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/template/DaxPageHeader.kt
mikescamell: Got it! Yeah I like it, I think most of the time we probably won't expose it for others, but I guess if there's a need then this makes it easier.

What I want to be conscious of is that we all end up following the same pattern when building Composables, so if we adopt this then we should also ensure other Composables adopt this going forward.

We'd also need to update/suppress the Lint rule for using textColors otherwise we'll always see the warnings 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
GerardPaligot: I agree, but because we are using uppercase in the rest of the app, I wasn't sure what was best. Do you think we should adopt a new format for Enums in Compose components?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
GerardPaligot: Dax is always better!

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
mikescamell: In fact I wonder if we can remove this completely? You can't select a StatusIndicator and you are already covering it with the label, what do you think?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
mikescamell: Got it, Is there any documentation on that as a good practice?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/StatusIndicator.kt
mikescamell: The docs recommend it, I think as we're starting out fresh in Compose lets do it 👍  https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#singletons_constants_sealed-class-and-enum-class-values

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
GerardPaligot: The [official documentation](https://developer.android.com/develop/ui/compose/graphics/draw/shapes) has been recently updated (pretty cool btw), other compose documentation pages are showing shape usages on border and surfaces (e.g. [here](https://developer.android.com/develop/ui/compose/graphics/images/customize)) but this [stackoverflow issue](https://developer.android.com/develop/ui/compose/graphics/images/customize) is describing pretty well the issue. Shape aren't very well applied at every component layer if you are using the shape only in the `Modifier.clip` function.

.editorconfig
```diff
@@ -95,4 +95,7 @@ ktlint_standard_no-wildcard-imports = disabled
 ktlint_standard_filename = disabled
 ktlint_standard_package-name = disabled
 ktlint_standard_annotation = disabled
-ktlint_standard_kdoc-wrapping = disabled
\ No newline at end of file
+ktlint_standard_kdoc-wrapping = disabled
+
+[android-design-system/**/ui/compose/**/*.kt]
+ktlint_standard_enum-entry-name-case = disabled
```
mikescamell, line 101: ❤️

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/**
+ * The possible statuses for the [DaxStatusIndicator].
+ */
+enum class Status {
+    /**
+     * Indicates that the status is always on.
+     */
+    Always_On,
```
mikescamell: 🔍 "Enum entry name 'Always_On' should not contain underscores when camel case is used"

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
```diff
@@ -0,0 +1,140 @@
...
+        get() = AlertGreen
+
+    internal val contentColor: Color
+        @Composable
+        get() = DuckDuckGoTheme.colors.text.secondary
+
+    internal val typography: DuckDuckGoTextStyle
+        @Composable
+        get() = DuckDuckGoTheme.typography.caption
+}
```
mikescamell, line 133: 🔍We should apply the `@ReadOnlyComposable` annotation to these properties

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/DaxStatusIndicator.kt
```diff
@@ -0,0 +1,140 @@
...
+     */
+    On,
+
+    /**
+     * Indicates that the status is off.
+     */
+    Off,
+}
+
+object StatusIndicatorDefaults {
```
mikescamell: 🔍 Should we mark this as internal?

PR #7678, opened 2026-02-05
Add launch pixel to VPN Reminder notification

nalcalag:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213129522980997

### Description
Add launch pixel for vpn reminder notification

### Steps to test this PR
- [ ] Apply patch on https://app.asana.com/1/137249556945/task/1213129522980997
      - Allow subscriptions
      -  It will set the notification to launch 15 second after the free trial starts
- [ ] Set device language to English
- [ ] Fresh install
- [ ] Purchase a Test Subscription (Free Trial)
- [ ] Filter in logcat for `vpn_reminder`
- [ ] Wait 15 seconds until notification is shown
- [ ] Tap in the VPN reminder notification
- [ ] Make sure `Pixel sent: mnot_l_vpn_reminder with params: {} {}` is shown in logcat 

### No UI changes

---

> [!NOTE]
> **Low Risk**
> Small, analytics-only change that adds an extra pixel on notification tap; no auth, payments, or persistent data flow changes.
> 
> **Overview**
> Adds a new navigation params type, `NetworkProtectionManagementScreenWithLaunchPixel`, to allow the VPN management screen to be launched with an associated *launch* pixel.
> 
> Updates the VPN reminder notification to open the management screen using this params object and introduces the `mnot_l_*` pixel, which is fired on activity launch via new plumbing in `NetworkProtectionManagementActivity`/`NetworkProtectionManagementViewModel` and `NetworkProtectionPixels`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 08ce32d5ef4d9d3b6013cd99fa29ca03bf599e82. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementActivity.kt
+++ b/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementActivity.kt
@@ -49,6 +49,7 @@ import com.duckduckgo.navigation.api.getActivityParams
 import com.duckduckgo.networkprotection.api.NetworkProtectionScreens.NetPAppExclusionListNoParams
 import com.duckduckgo.networkprotection.api.NetworkProtectionScreens.NetworkProtectionManagementScreenAndEnable
 import com.duckduckgo.networkprotection.api.NetworkProtectionScreens.NetworkProtectionManagementScreenNoParams
+import com.duckduckgo.networkprotection.api.NetworkProtectionScreens.NetworkProtectionManagementScreenWithLaunchPixel
 import com.duckduckgo.networkprotection.impl.R
 import com.duckduckgo.networkprotection.impl.autoexclude.VpnAutoExcludePromptFragment
 import com.duckduckgo.networkprotection.impl.autoexclude.VpnAutoExcludePromptFragment.Companion.Source.VPN_SCREEN
@@ -79,6 +80,7 @@ import javax.inject.Inject
 @InjectWith(ActivityScope::class)
 @ContributeToActivityStarter(NetworkProtectionManagementScreenNoParams::class, screenName = "vpn.main")
 @ContributeToActivityStarter(NetworkProtectionManagementScreenAndEnable::class, screenName = "vpn.main")
+@ContributeToActivityStarter(NetworkProtectionManagementScreenWithLaunchPixel::class, screenName = "vpn.main")
 class NetworkProtectionManagementActivity : DuckDuckGoActivity() {
 
     @Inject
@@ -122,8 +124,8 @@ class NetworkProtectionManagementActivity : DuckDuckGoActivity() {
             }
         }
 
-        intent.getStringExtra(LAUNCH_FROM_NOTIFICATION_PIXEL_NAME)?.let { pixelName ->
-            viewModel.onLaunchedFromNotification(pixelName)
+        intent.getActivityParams(NetworkProtectionManagementScreenWithLaunchPixel::class.java)?.let { params ->
+            viewModel.onLaunchedFromNotification(params.pixelName)
         }
 
         observeViewModel()
@@ -550,7 +552,6 @@ class NetworkProtectionManagementActivity : DuckDuckGoActivity() {
     }
 
     companion object {
-        private const val LAUNCH_FROM_NOTIFICATION_PIXEL_NAME = "LAUNCH_FROM_NOTIFICATION_PIXEL_NAME"
         private const val REPORT_ISSUES_ANNOTATION = "report_issues_link"
         private const val OPEN_SETTINGS_ANNOTATION = "open_settings_link"
         private const val TAG_PROMOTION_DIALOG = "TAG_PROMO_DIALOG"
--- a/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementViewModel.kt
+++ b/network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementViewModel.kt
@@ -24,7 +24,6 @@ import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.duckduckgo.anvil.annotations.ContributesViewModel
-import com.duckduckgo.app.statistics.pixels.Pixel
 import com.duckduckgo.common.utils.ConflatedJob
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.ActivityScope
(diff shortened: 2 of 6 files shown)


nalcalag:
* **#7678** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementActivity.kt
```diff
@@ -122,6 +125,10 @@ class NetworkProtectionManagementActivity : DuckDuckGoActivity() {
             }
         }
 
+        intent.getStringExtra(LAUNCH_FROM_NOTIFICATION_PIXEL_NAME)?.let { pixelName ->
```
karlenDimla: NIT: This logic should be handled in the ViewModel. It will more correct to do it that way and also easier to add test.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/notification/VpnReminderNotification.kt
```diff
@@ -140,14 +140,17 @@ class VpnReminderNotificationPlugin @Inject constructor(
...
             ?: return null
+        intent.putExtra(LAUNCH_FROM_NOTIFICATION_PIXEL_NAME, pixelName(NOTIFICATION_LAUNCHED_PIXEL))
         return taskStackBuilderFactory.createTaskBuilder().run {
             addNextIntentWithParentStack(intent)
             getPendingIntent(0, PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE)
         }
     }
 
     companion object {
+        const val LAUNCH_FROM_NOTIFICATION_PIXEL_NAME = "LAUNCH_FROM_NOTIFICATION_PIXEL_NAME"
```
aitorvs: This should not be done like this.
You have to consts, in different modules and you're relying on the name being the same.

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementViewModel.kt
```diff
@@ -441,6 +443,10 @@ class NetworkProtectionManagementViewModel @Inject constructor(
         networkProtectionPixels.reportExcludePromptExcludeAppClicked()
     }
 
+    fun onLaunchedFromNotification(pixelName: String) {
+        pixel.fire(pixelName)
```
aitorvs: Why don't we follow the pattern we've always followed, using `NetworkProtectionPixels` as the class that sends pixels?

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/notification/VpnReminderNotification.kt
```diff
@@ -140,14 +140,17 @@ class VpnReminderNotificationPlugin @Inject constructor(
     override fun getLaunchIntent(): PendingIntent? {
         val intent = globalActivityStarter.startIntent(context, NetworkProtectionManagementScreenNoParams)
             ?: return null
+        intent.putExtra(LAUNCH_FROM_NOTIFICATION_PIXEL_NAME, pixelName(NOTIFICATION_LAUNCHED_PIXEL))
```
aitorvs: We're doing this wrong. You're getting the intent for no params and then adding one extra param.
Why not rather add a new `NetworkProtectionManagementScreenXXX` that already contains the pixel?
That way we don't need to define the `LAUNCH_FROM_NOTIFICATION_PIXEL_NAME` here and all is encapsulated

Like adding a new `NetworkProtectionManagementScreenWithLaunchPixel(val pixelName: String)` instead how we do it here. Bypassing all the launch infra we've got?

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementViewModel.kt
karlenDimla: We can probably add this pixel in `NetworkProtectionPixels` so you dont have to pass it in the intent.

karlenDimla: You can just pass a signal to the activity that it came from a notification

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementActivity.kt
aitorvs: This logic should not exist, see comments below.
The pixelName should come from a new class inside `NetworkProtectionScreens` so that we can get all launch params using `intent.getActivityParams()`.
Rather than hardcoding it as it's done today

karlenDimla (changes requested):
Reverting my approval as I agree with Aitor’s comments now that i read them

network-protection/network-protection-impl/src/main/java/com/duckduckgo/networkprotection/impl/management/NetworkProtectionManagementViewModel.kt
aitorvs: I don't think we need to pass any signal to the activity. We just need to create a new screen data class with pixel name to be send on launch. The origin doesn't matter.

karlenDimla: She will still need the Pixel reference if we go with `NetworkProtectionManagementScreenWithLaunchPixel(val pixelName: String)`. Probably `NetworkProtectionManagementScreenFromNotification` is  better without the actual pixel and then let NetworkProtectionPixels know the details about the pixel.

aitorvs: Thinking open-close principle, ie. the API scales without having to touch (open) it.
If we do `NetworkProtectionManagementScreenFromNotification` we need add new API classes for every entry point that need to send a pixel.
Sending the pixel is the business of the VPN, knowing what pixel to send or who's the caller is not its business.

nalcalag:
@aitorvs @karlenDimla I made all the requested changes

karlenDimla (approved):

PR #7714, opened 2026-02-10
Address decryption issues due to preferences migration

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213215582601746

### Description
* When querying preferences, if preferences file is not available, throw an exception instead of falling back to null
* When updating preferences, if preferences file is not available, throw an exception so the consumer doesn't move forward with the operation (fire and forget otherwise)
* Force a migration from legacy→harmony
* Fire pixel if values don't match between harmony and legacy (only if autofill service isn't supported, otherwise it's expected to see discrepancies due to sharedPreferences not supporting multi-process)

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |

### DIFF
--- a/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/store/keys/SecureStorageKeyStore.kt
+++ b/autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/store/keys/SecureStorageKeyStore.kt
@@ -30,6 +30,7 @@ import com.duckduckgo.autofill.impl.pixel.AutofillPixelNames.AUTOFILL_HARMONY_PR
 import com.duckduckgo.autofill.impl.pixel.AutofillPixelNames.AUTOFILL_HARMONY_PREFERENCES_UPDATE_KEY_FAILED
 import com.duckduckgo.autofill.impl.pixel.AutofillPixelNames.AUTOFILL_PREFERENCES_GET_KEY_FAILED
 import com.duckduckgo.autofill.impl.pixel.AutofillPixelNames.AUTOFILL_PREFERENCES_UPDATE_KEY_FAILED
+import com.duckduckgo.autofill.impl.service.AutofillServiceFeature
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.common.utils.sanitizeStackTrace
 import com.duckduckgo.data.store.api.SharedPreferencesProvider
@@ -71,6 +72,7 @@ class RealSecureStorageKeyStore constructor(
     private val autofillFeature: AutofillFeature,
     private val sharedPreferencesProvider: SharedPreferencesProvider,
     private val pixel: Pixel,
+    private val autofillServiceFeature: AutofillServiceFeature,
 ) : SecureStorageKeyStore {
 
     private val mutex: Mutex = Mutex()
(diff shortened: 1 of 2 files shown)


CrisBarreiro:
* **#7714** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7714?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/store/keys/SecureStorageKeyStore.kt
```diff
@@ -106,7 +108,7 @@ class RealSecureStorageKeyStore constructor(
             try {
                 harmonyMutex.withLock {
                     if (autofillFeature.useHarmony().isEnabled()) {
-                        sharedPreferencesProvider.getMigratedEncryptedSharedPreferences(FILENAME).also {
+                        sharedPreferencesProvider.getMigratedEncryptedSharedPreferences(FILENAME, "${FILENAME}2").also {
```
CrisBarreiro, line 111: Giving the harmony preferences a new name, so that we can force the migration from legacy->harmony without changing the internal key used by SharedPreferencesProviderImpl

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/store/keys/SecureStorageKeyStore.kt
```diff
@@ -141,8 +143,24 @@ class RealSecureStorageKeyStore constructor(
...
+                        throw IllegalStateException("Harmony preferences unavailable")
+                    }
+            } else {
+                null
+            }
+
+            val preferences = getEncryptedPreferences()
+                ?: run {
+                    pixel.fire(AUTOFILL_PREFERENCES_UPDATE_KEY_FAILED, mapOf("error" to "preferences_null"), type = Daily())
+                    throw IllegalStateException("Encrypted preferences unavailable")
```
CrisBarreiro, line 159: Retrieving both preferences beforehand, so in case  there as a problem accessing them and they're null, we fail early without making any changes

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/store/keys/SecureStorageKeyStore.kt
```diff
@@ -159,39 +177,45 @@ class RealSecureStorageKeyStore constructor(
...
-                preferences?.getString(keyName, null)?.run {
-                    this.decodeBase64()?.toByteArray()
+                val legacyPrefs = getEncryptedPreferences()
+                    ?: throw IllegalStateException("Encrypted preferences unavailable")
+                val legacyResult = legacyPrefs.getString(keyName, null)?.decodeBase64()?.toByteArray()
+
+                if (useHarmony) {
+                    val harmonyPrefs = getHarmonyEncryptedPreferences()
+                        ?: throw IllegalStateException("Harmony preferences unavailable")
+                    val harmonyResult = harmonyPrefs.getString(keyName, null)?.decodeBase64()?.toByteArray()
```
CrisBarreiro, line 211: Failing early if files aren't accessible. Technically, if harmony is available with useHarmony set to true, and legacy is not, we could proceed, but I think it's safer not to

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/store/keys/SecureStorageKeyStore.kt
```diff
@@ -159,39 +177,45 @@ class RealSecureStorageKeyStore constructor(
...
+
+                    if (!autofillServiceFeature.self().isEnabled() && !harmonyResult.contentEquals(legacyResult)) {
+                        throw IllegalStateException("Mismatching preference values")
```
CrisBarreiro, line 214: Firing a pixel to detect cases where legacy and harmony have different values. Not doing it if autofill service is enabled, as discrepancies are expected there due to legacy not supporting multi-process

autofill/autofill-impl/src/main/java/com/duckduckgo/autofill/store/keys/SecureStorageKeyStore.kt
CDRussell: 👍

should maybe also attempt to remove the old one if it exists as a small cleanup

PR #7749, opened 2026-02-16
Track for page load performance

GerardPaligot:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213167063549551

### Description

* Mesure timing milestones during page load (start, visible, progress escape and finish)
* Capture page load metadata

### Steps to test this PR

_Website tracked_
- [x] Open your logcat and apply filter RealPageLoadWideEvent
- [x] Open one website listed in PageLoadedSites class (e.g. bbc.com)
- [x] See page loading progress from start to finish with milestones (check description)

_Website not tracked_
- [x] Open your logcat and apply filter RealPageLoadWideEvent
- [x] Open a website not listed in PageLoadedSites class
- [x] Nothing is shown in your logcat

_Website loaded in parallel_
- [x] Open your logcat and apply filter RealPageLoadWideEvent
- [x] Prepare two tab to be able to switch quickly between them
- [x] Load two different websites in these two tabs before the end of the first load
- [x] See page loading progress for these two tabs

---

> [!NOTE]
> **Medium Risk**
> Adds new telemetry and hooks into WebView page-load lifecycle (start/visible/progress/finish) including error paths, so mistakes could impact metrics correctness and add overhead during browsing, but it is gated by remote config and limited to an allowlist of sites.
> 
> **Overview**
> Adds a new `wide_page-load` Wide Event pixel definition to measure page-load lifecycle timing (start → visible → escape fixed progress → finish) and attach metadata like WebView version, consent/optimization flags, foreground state, and request concurrency.
> 
> Introduces `RealPageLoadWideEvent` and wires it into `BrowserWebViewClient` and `BrowserTabViewModel` to start/step/finish flows on page start, commit-visible, first progress >50%, completion, and main-frame failures/render-process-gone, gated by the new `AndroidBrowserConfigFeature.sendPageLoadWideEvent()` toggle and restricted to a new `PageLoadedSites.perfSites` domain allowlist.
> 
> Expands unit tests to cover the new event triggers and edge cases (duplicate progress events, about:blank, missing tab id, omitted errors, crashes/kills).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e1338d1d814a216ec9759427357a7763078e64b2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
+++ b/app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
@@ -326,8 +326,8 @@ interface AndroidBrowserConfigFeature {
      * Controls whether page load wide events are sent.
      * @return `true` when the remote config has the global "sendPageLoadWideEvent" androidBrowserConfig
      * sub-feature flag enabled
-     * If the remote feature is not present defaults to `true`
+     * If the remote feature is not present defaults to `internal`
      */
-    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
+    @Toggle.DefaultValue(DefaultFeatureValue.INTERNAL)
     fun sendPageLoadWideEvent(): Toggle
 }


GerardPaligot:
* **#7749** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7749?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7749?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,278 @@
...
+        tabId: String,
+        outcome: String,
+        errorCode: String?,
+        isTabInForegroundOnFinish: Boolean,
+        activeRequestsOnLoadStart: Int,
+        concurrentRequestsOnFinish: Int,
+    ) {
+        if (!isFeatureEnabled()) return
+        val flowId = popActiveFlowId(tabId) ?: return
+
```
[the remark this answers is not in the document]
GerardPaligot: It is probably acceptable. We wont change the feature flag value often (maybe never).

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,303 @@
...
+    private val dispatchers: DispatcherProvider,
+) : PageLoadWideEvent {
+    private val activeFlows = ConcurrentHashMap<String, PageLoadState>()
+
+    override fun isInProgress(tabId: String, url: String): Boolean {
+        val state = activeFlows[tabId] ?: return false
+        val ageMillis = System.currentTimeMillis() - state.createdAt
+        val isStale = ageMillis > CLEANUP_TIMEOUT_WITH_BUFFER.inWholeMilliseconds
+        return state.url == url && !isStale
+    }
```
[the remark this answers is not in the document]
GerardPaligot: It is what is expected. The actual behavior is to reset the progress when the user is redirected to another URL.

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadPerformanceMonitor.kt
```diff
@@ -0,0 +1,171 @@
...
+    override fun onPageLoadSucceeded(
+        tabId: String,
+        url: String,
+        isTabInForegroundOnFinish: Boolean,
+        activeRequestsOnLoadStart: Int,
+        concurrentRequestsOnFinish: Int,
+    ) {
+        if (!pageLoadWideEvent.isInProgress(tabId, url)) return
+        appCoroutineScope.launch(dispatchers.io()) {
+            pageLoadWideEvent.finishPageLoad(
```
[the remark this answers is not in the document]
GerardPaligot: The risk is acceptable because the frequency is very low. It requires page finish and new navigation with a very short milliseconds delay.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,299 @@
...
+        wideEventClient.intervalEnd(
+            wideEventId = flowId,
+            key = KEY_ELAPSED_TIME_TO_ESCAPED_FIXED_PROGRESS,
+        )
+
+        wideEventClient.flowStep(
+            wideEventId = flowId,
+            stepName = STEP_PAGE_ESCAPED_FIXED_PROGRESS,
+            metadata = mapOf(
+                KEY_PROGRESS to actualProgress.toString(),
```
lmac012: Do we need the actual progress value? Isn't it enough to know that we've moved from the fixed value?

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,299 @@
...
+    private val optimizeTrackerEvaluationRCWrapper: OptimizeTrackerEvaluationRCWrapper,
+    private val androidBrowserConfigFeature: Lazy<AndroidBrowserConfigFeature>,
+    private val dispatchers: DispatcherProvider,
+) : PageLoadWideEvent {
+    private val activeFlows = ConcurrentHashMap<String, PageLoadState>()
+
+    override fun isInProgress(tabId: String, url: String): Boolean {
+        val state = activeFlows[tabId] ?: return false
+        val ageMillis = System.currentTimeMillis() - state.createdAt
+        val isStale = ageMillis > CLEANUP_TIMEOUT_WITH_BUFFER.inWholeMilliseconds
```
lmac012: Why do we need to add buffer to the cleanup timeout? What would happen if we just used `CLEANUP_TIMEOUT` here?

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadManager.kt
```diff
@@ -0,0 +1,171 @@
...
+import com.squareup.anvil.annotations.ContributesBinding
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+/**
+ * Manages page load event tracking lifecycle.
+ * Coordinates page load events across the browser, tracking visibility, progress, and completion.
+ */
+interface PageLoadManager {
```
lmac012: nit: I think the naming here is slightly misleading - `PageLoadManager` looks like this can impact page loading logic, while in fact this is only instrumentation. Perhaps `PageLoadPerformanceMonitor` would work better?

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadManager.kt
```diff
@@ -0,0 +1,171 @@
...
+class WideEventPageLoadManager @Inject constructor(
+    private val pageLoadWideEvent: PageLoadWideEvent,
+    @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
+    private val dispatchers: DispatcherProvider,
+) : PageLoadManager {
+
+    override fun onPageStarted(tabId: String, url: String) {
+        if (!shouldTrackUrl(url)) return
+        if (pageLoadWideEvent.isInProgress(tabId, url)) return
+        appCoroutineScope.launch(dispatchers.io()) {
```
lmac012: You should not need to specify dispatcher  explicitly in order to execute suspending (which implies non blocking) method. 

This applies to other places in this class where dispatcher is set explicitly as well.

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadPerformanceMonitor.kt
```diff
@@ -0,0 +1,171 @@
...
+        if (pageLoadWideEvent.isInProgress(tabId, url)) return
+        appCoroutineScope.launch(dispatchers.io()) {
+            pageLoadWideEvent.startPageLoad(tabId, url)
+        }
+    }
+
+    override fun onPageVisible(tabId: String, url: String, progress: Int) {
+        if (!pageLoadWideEvent.isInProgress(tabId, url)) return
+        appCoroutineScope.launch(dispatchers.io()) {
+            pageLoadWideEvent.recordPageVisible(tabId, progress)
```
lmac012: Have you considered the scenario where `recordPageVisible()` is executed after `onPageLoadSucceeded()`? Async coroutines that invoke those methods are launched sequentially, but they are not guaranteed to execute in the same order.

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadPerformanceMonitor.kt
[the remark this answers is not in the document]
lmac012: Yup, if the result is just skipping the next event then I agree this is fine.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -724,6 +744,19 @@ class BrowserWebViewClient @Inject constructor(
...
+                    // Trigger Wide Event failure for main frame errors
+                    request.url?.toString()?.let { url ->
+                        webViewClientListener?.getCurrentTabId()?.let { tabId ->
+                            pageLoadManager.onPageLoadFailed(
+                                tabId = tabId,
+                                url = url,
+                                errorDescription = webResourceError.errorCode.asStringErrorCode(),
+                                isTabInForegroundOnFinish = webViewClientListener?.isTabInForeground() ?: true,
+                                activeRequestsOnLoadStart = parallelRequestsOnStart,
+                                concurrentRequestsOnFinish = decrementLoadCountAndGet(),
```
lmac012: The `onReceivedError` method is now updating `concurrentRequestsOnFinish` more times than before - I don't know what exactly is this counter for, but it looks like this may affect the existing perf monitoring?

app/src/test/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
```diff
@@ -1100,6 +1151,223 @@ class BrowserWebViewClientTest {
...
 
+    @Test
+    fun whenPageStartedThenManagerOnPageStartedCalled() {
+        val mockWebView = getImmediatelyInvokedMockWebView()
+        val tabId = "test-tab-123"
+        whenever(mockWebView.safeCopyBackForwardList()).thenReturn(TestBackForwardList())
+        whenever(mockWebView.settings).thenReturn(mock())
+        whenever(listener.getCurrentTabId()).thenReturn(tabId)
+        testee.onPageStarted(mockWebView, EXAMPLE_URL, null)
+        verify(mockPageLoadWideEvent).onPageStarted(tabId, EXAMPLE_URL)
```
lmac012: nit: `PageLoadWideEvent` is not a direct dependency of the `BrowserWebViewClient` implementation, but an implementation detail of `PageLoadManager`, so we should probably assert invocations of `PageLoadManager` interface here.

PixelDefinitions/pixels/wide_page_load.json5
```diff
@@ -0,0 +1,110 @@
...
+            "widePixelDevMode",
+            {
+                "key": "feature.name",
+                "description": "Feature identifier for this wide event",
+                "enum": ["page-load"]
+            },
+            {
+                "key": "feature.data.ext.elapsed_time_to_finish_ms_bucketed",
+                "description": "Total duration from page_start to page_finish, bucketed. Uses upper end of the bucket.",
+                "enum": ["1000", "5000", "10000", "30000", "60000", "300000", "600000"]
```
lmac012: The bucketing logic in wide event library has been recently updated to use lower end of the bucket, so "0" was introduced as a possible value.

app/src/test/java/com/duckduckgo/app/browser/pageload/PageLoadWideEventTest.kt
```diff
@@ -0,0 +1,402 @@
...
+
+    private lateinit var pageLoadWideEvent: PageLoadWideEvent
+
+    @Before
+    fun setup() = runTest {
+        whenever(webViewVersionProvider.getMajorVersion()).thenReturn("120")
+        whenever(autoconsent.isAutoconsentEnabled()).thenReturn(true)
+        whenever(optimizeTrackerEvaluationRCWrapper.enabled).thenReturn(true)
+
+        val mockToggle = mock<Toggle>()
```
lmac012: nit: We have fake implementation for this which is preferred to using mocks; see `FakeFeatureToggleFactory` for reference.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,299 @@
...
+     * @param actualProgress The progress when escaping fixed zone (typically >= 50)
+     */
+    suspend fun recordExitedFixedProgress(tabId: String, actualProgress: Int)
+
+    /**
+     * Called from BrowserWebViewClient.onPageFinished or onReceivedError when page load completes.
+     * Records page_finish flow step and completes the Wide Event flow.
+     *
+     * @param tabId The unique identifier for the tab
+     * @param outcome "success", "error", or "timeout"
```
lmac012: nit: Looks like the `outcome` parameter should be an enum.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,299 @@
...
+    private val autoconsent: Autoconsent,
+    private val optimizeTrackerEvaluationRCWrapper: OptimizeTrackerEvaluationRCWrapper,
+    private val androidBrowserConfigFeature: Lazy<AndroidBrowserConfigFeature>,
+    private val dispatchers: DispatcherProvider,
+) : PageLoadWideEvent {
+    private val activeFlows = ConcurrentHashMap<String, PageLoadState>()
+
+    override fun isInProgress(tabId: String, url: String): Boolean {
+        val state = activeFlows[tabId] ?: return false
+        val ageMillis = System.currentTimeMillis() - state.createdAt
```
lmac012: nit: From best practices perspective, it's better to use DI to provide time instead of calling static `System.currentTimeMillis()`. We have an interface `CurrentTimeProvider` defined in the common module which serves this purpose and makes the code more testable by letting you easily manage the clock in unit tests.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,299 @@
...
+        result.onSuccess { flowId ->
+            activeFlows[tabId] = PageLoadState(flowId, url)
+            logcat { "Page load flow started: tabId=$tabId, url=$url, flowId=$flowId" }
+
+            wideEventClient.flowStep(
+                wideEventId = flowId,
+                stepName = STEP_PAGE_START,
+            )
+
+            wideEventClient.intervalStart(
```
lmac012: `intervalStart()` uses wall clock to measure time - do you think it matters in this scenario that we don't use `SystemClock.elapsedRealtime()`? Probably not a concern since this is a foreground operation but I wanted to flag this just in case.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
GerardPaligot: I can’t guarantee that the timeouts are identical between the wide event client and mine. The original intent was to prevent desynchronization. However, after further consideration, using the same timeout is acceptable, probably better than forcing a difference. I'll make the change.

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadManager.kt
GerardPaligot: I wasn't satisfied by this name too, I like yours!

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadPerformanceMonitor.kt
GerardPaligot: When `finishPageLoad()` executes, it calls `popActiveFlowId()` which removes the flow from `activeFlows`. If `recordPageVisible()` runs after, `getActiveFlowId()` returns null.

In my opinion, this is the desired behavior. I would prefer to omit this step entirely rather than maintain a state just to trigger it during the success or failure phases. What do you think?

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
GerardPaligot: Yes, here it isn't an issue because it'll be executed in foreground and the time will be short but it could be interesting to know if we want to keep a wall clock in the wide event implementation. This kind of time measurement can be affected by system time but it shouldn't in our application.

app/src/test/java/com/duckduckgo/app/browser/BrowserWebViewClientTest.kt
GerardPaligot: Yes, it is the case but forgot to rename the variable when I created the manager as extra layer between webviewclient and wideevent class.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,305 @@
...
+
+        wideEventClient.intervalEnd(
+            wideEventId = flowId,
+            key = KEY_ELAPSED_TIME_TO_ESCAPED_FIXED_PROGRESS,
+        )
+
+        wideEventClient.flowStep(
+            wideEventId = flowId,
+            stepName = STEP_PAGE_ESCAPED_FIXED_PROGRESS,
+        )
```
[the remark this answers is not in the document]
GerardPaligot: It is no more expected but I'll clean up the progress parameter, it is no more pertinent to keep it.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,304 @@
...
+
+        logcat { "Exited fixed progress: tabId=$tabId, flowId=$flowId" }
+    }
```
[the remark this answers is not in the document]
GerardPaligot: We don't need anymore to have the progress in this function. It is no more expected to log this there.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -0,0 +1,304 @@
...
+            wideEventClient.flowAbort(existingState.flowId)
+        }
+
+        val result = wideEventClient.flowStart(
+            name = PAGE_LOAD_FEATURE_NAME,
+            cleanupPolicy = CleanupPolicy.OnTimeout(CLEANUP_TIMEOUT.toJavaDuration()),
+        )
+
+        result.onSuccess { flowId ->
+            activeFlows[tabId] = PageLoadState(flowId, url)
```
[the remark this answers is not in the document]
GerardPaligot: The risk is acceptable because the user should refresh its page very fast (probably too fast compared to what is really possible).

app/src/main/java/com/duckduckgo/app/browser/pageload/WideEventPageLoadPerformanceMonitor.kt
lmac012: I wasn't suggesting maintaining the state after the event is finished, I was just pointing to the fact that some events will be incomplete (steps will not be recorded) and we need to make sure this doesn't break any monitoring / alerting.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
lmac012: I would keep using wall clock as default in the wide event client implementation, since some events are longer and span across process restarts, etc. We can consider extending the API so that the caller can choose which clock is user but no need for that now IMO.

lmac012 (approved):

PR #7614, opened 2026-01-28
Add support for breakageData parameter

GuiltyDolphin:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212311623899110/task/1213015238129927?focus=true

### Description

Adds support for the `breakageData` parameter which is passed from CSS.

The parameter is returned as part of the Content Scope Scripts breakage reporting feature as a pre-encoded string which should then be passed along as-is to the breakage report request without further encoding.

### Steps to test this PR

Covered by unit tests and reference tests (https://github.com/duckduckgo/privacy-reference-tests/pull/156).

I've manually tested the change and can see the `breakageData` parameter populated correctly.

1. Make sure you're using the latest Content Scope Scripts
2. Set the state of `webDetection`, `webInterferenceDetection`, and `breakageReporting` to `"enabled"` in the remote config
3. Visit https://duckduckgo.com/?q=disable%20adblocker
4. Add logcat filtering for `breakageData`
5. Make a breakage report. Verify that you can see the populated parameter (e.g., `breakageData=%7B%22webDetection%22%3A%5B%7B%22detectorId%22%3A%22adwalls.generic_en%22%2C%22detected%22%3Atrue%7D%5D%2C%22detectorData%22%3A%7B%22botDetection%22%3A%7B%22detected%22%3Afalse%2C%22type%22%3A%22botDetection%22%2C%22results%22%3A%5B%5D%7D%2C%22fraudDetection%22%3A%7B%22detected%22%3Afalse%2C%22type%22%3A%22fraudDetection%22%2C%22results%22%3A%5B%5D%7D%2C%22adwallDetection%22%3A%7B%22detected%22%3Atrue%2C%22type%22%3A%22adwallDetection%22%2C%22results%22%3A%5B%7B%22detected%22%3Atrue%2C%22detectorId%22%3A%22generic%22%7D%5D%7D%2C%22youtubeAds%22%3A%7B%22detected%22%3Afalse%2C%22type%22%3A%22youtubeAds%22%2C%22results%22%3A%5B%5D%7D%7D%7D&test=1`)

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the broken-site reporting payload and encoding path; incorrect handling could break report submission or produce malformed analytics data.
> 
> **Overview**
> Adds end-to-end support for a new `breakageData` field in broken-site reporting, captured from content-scope-scripts and stored in `BrokenSiteContext`.
> 
[...]

PR #7754, opened 2026-02-17
Add support for Web Telemetry counter

GuiltyDolphin:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1212827841082637

### Description

Adds support for adwall (aggregate) counting pixels via web telemetry.

### Steps to test this PR

Testing steps are covered in https://github.com/duckduckgo/ddg-workflow/blob/gd-detector-telemetry/technical-designs/web-detection-framework/eventhub-android-testing-plan.md

---

> [!NOTE]
> **Medium Risk**
> Adds a new backgrounded/foregrounded lifecycle-driven telemetry pipeline that persists state in Room and fires pixels on timers, which could impact analytics correctness and scheduling behavior. The feature is gated by a remote toggle but touches WebView messaging and app lifecycle hooks.
> 
> **Overview**
> Introduces a new **`eventHub` telemetry feature** that ingests `webEvents`/`webEvent` messages from Content Scope Scripts, aggregates counter-based metrics over configurable periods, and fires bucketed pixels with an `attributionPeriod` parameter.
> 
> Adds a new `event-hub-impl` module with remote-config parsing, per-pixel state persistence (Room), deduping per `webViewId`/navigation, scheduling to fire at period end, and plugins/hooks for privacy-config updates and app lifecycle foreground/background transitions. Updates content-scope messaging to inject `nativeData.webViewId` into `webEvent` messages, wires the module into `app/build.gradle`, and adds pixel definitions for daily/weekly adwall detection telemetry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d880c7805c78dec8083917ef02226e237c22a7ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
+++ b/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
@@ -129,7 +129,7 @@ class RealEventHubPixelManager @Inject constructor(
             val nowMillis = timeProvider.currentTimeMillis()
 
             for (pixelState in repository.getAllPixelStates()) {
-                if (nowMillis > pixelState.periodEndMillis) continue
+                if (nowMillis >= pixelState.periodEndMillis) continue
 
                 val updatedParams = pixelState.params.toMutableMap()
                 var changed = false
--- a/event-hub/event-hub-impl/src/test/java/com/duckduckgo/eventhub/impl/pixels/RealEventHubPixelManagerTest.kt
+++ b/event-hub/event-hub-impl/src/test/java/com/duckduckgo/eventhub/impl/pixels/RealEventHubPixelManagerTest.kt
@@ -139,6 +139,30 @@ class RealEventHubPixelManagerTest {
     }
 
     @Test
+    fun `handleWebEvent ignores events exactly at periodEnd`() {
+        val periodEnd = 100_000L
+        timeProvider.time = periodEnd
+        val state = pixelState("webTelemetry_testPixel1", mapOf("count" to 0), periodEnd = periodEnd)
+        stubPixelStates(state)
+
+        manager.handleWebEvent(webEventData("test"), "")
+
+        verify(repository, never()).savePixelState(any())
+    }
+
+    @Test
+    fun `handleWebEvent accepts events one millisecond before periodEnd`() {
+        val periodEnd = 100_000L
+        timeProvider.time = periodEnd - 1
+        val state = pixelState("webTelemetry_testPixel1", mapOf("count" to 0), periodEnd = periodEnd)
+        stubPixelStates(state)
+
+        manager.handleWebEvent(webEventData("test"), "")
+
+        verify(repository).savePixelState(any())
+    }
+
+    @Test
     fun `handleWebEvent can increment multiple times below max bucket`() {
         val state = pixelState("webTelemetry_testPixel1", mapOf("count" to 0))
         stubPixelStates(state)
@@ -698,6 +722,46 @@ class RealEventHubPixelManagerTest {
         val periodStart = 1000L
         val periodEnd = periodStart + TimeUnit.DAYS.toMillis(1)
         timeProvider.time = periodStart + TimeUnit.HOURS.toMillis(12)
+
+        val state = PixelState(
+            pixelName = "webTelemetry_testPixel1",
+            periodStartMillis = periodStart,
+            periodEndMillis = periodEnd,
+            params = mapOf("count" to ParamState(5)),
+            config = dayPixelConfig,
+        )
+        stubPixelStates(state)
+
+        manager.onAppForegrounded()
+
+        verify(pixel, never()).enqueueFire(any<String>(), any(), any(), any())
+    }
+
+    @Test
+    fun `checkPixels fires pixel exactly at periodEnd`() {
+        val periodStart = 1000L
+        val periodEnd = periodStart + TimeUnit.DAYS.toMillis(1)
+        timeProvider.time = periodEnd
+
+        val state = PixelState(
+            pixelName = "webTelemetry_testPixel1",
+            periodStartMillis = periodStart,
+            periodEndMillis = periodEnd,
+            params = mapOf("count" to ParamState(5)),
+            config = dayPixelConfig,
+        )
+        stubPixelStates(state)
+
+        manager.onAppForegrounded()
+
+        verify(pixel).enqueueFire(any<String>(), any(), any(), any())
+    }
+
+    @Test
+    fun `checkPixels does not fire one millisecond before periodEnd`() {
+        val periodStart = 1000L
+        val periodEnd = periodStart + TimeUnit.DAYS.toMillis(1)
+        timeProvider.time = periodEnd - 1
 
         val state = PixelState(
             pixelName = "webTelemetry_testPixel1",


event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubFeatureName.kt
```diff
@@ -0,0 +1,26 @@
...
+package com.duckduckgo.eventhub.impl
+
+enum class EventHubFeatureName(val value: String) {
+    EventHub("eventHub"),
+    WebEvents("webEvents"),
+}
+
+fun eventHubFeatureValueOf(value: String): EventHubFeatureName? {
+    return EventHubFeatureName.entries.find { it.value == value }
+}
```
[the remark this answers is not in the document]
GuiltyDolphin: @cursoragent remove it.

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
```diff
@@ -73,6 +73,11 @@ class ContentScopeScriptsJsMessaging @Inject constructor(
                     webView.url?.toUri()?.host
                 }
             jsMessage?.let {
+                if (jsMessage.featureName == "webEvents" && jsMessage.method == "webEvent") {
+                    val nativeData = org.json.JSONObject()
+                    nativeData.put("tabId", System.identityHashCode(webView).toString())
```
cmonfortep: Rename key to something more representative after moving away from using tabId. Probably same for other parts of the PR if you are still referencing to tabId.

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
```diff
@@ -73,6 +73,11 @@ class ContentScopeScriptsJsMessaging @Inject constructor(
...
+                    jsMessage.params.put("nativeData", nativeData)
+                }
```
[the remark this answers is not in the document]
cmonfortep: @GuiltyDolphin check this comment to ensure is intended you run you logic before secret verification (and other checks if they are necessary)

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        repository.deletePixelState(pixelConfig.name)
+
+        val latestConfig = getParsedConfig()
+        val latestPixelConfig = latestConfig.telemetry.find { it.name == pixelConfig.name }
+        if (latestPixelConfig != null) {
+            startNewPeriod(latestPixelConfig)
+        }
+    }
```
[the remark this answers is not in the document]
GuiltyDolphin: This is fine. De-duping spans navigation boundary not period boundary.

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
```diff
@@ -74,6 +74,11 @@ class ContentScopeScriptsJsMessaging @Inject constructor(
                 }
             jsMessage?.let {
                 if (this.secret == secret && context == jsMessage.context && isUrlAllowed(allowedDomains, domain)) {
+                    if (jsMessage.featureName == "webEvents" && jsMessage.method == "webEvent") {
```
cmonfortep, line 78: should this be inside `isUrlAllowed(allowedDomains, domain)`?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/WebEventsContentScopeJsMessageHandler.kt
```diff
@@ -23,24 +23,38 @@ import com.duckduckgo.js.messaging.api.JsMessageCallback
...
             jsMessageCallback: JsMessageCallback?,
         ) {
-            jsMessageCallback?.process(featureName, jsMessage.method, jsMessage.id, jsMessage.params)
+            val eventType = jsMessage.params.optString("type", "")
+            if (eventType.isEmpty()) {
+                logcat(WARN) { "webEvent message missing 'type' parameter" }
+                return
+            }
+
+            val tabId = jsMessage.params.optJSONObject("nativeData")?.optString("tabId", "") ?: ""
```
cmonfortep: this is not representing TabId anymore, so best to rename.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/WebEventsContentScopeJsMessageHandler.kt
```diff
@@ -23,24 +23,38 @@ import com.duckduckgo.js.messaging.api.JsMessageCallback
...
+import logcat.logcat
 import javax.inject.Inject
 
-const val BROWSER_UI_LOCK_FEATURE_NAME = "browserUiLock"
+private const val WEB_EVENTS_FEATURE_NAME = "webEvents"
 
 @ContributesMultibinding(AppScope::class)
-class UiLockContentScopeJsMessageHandler @Inject constructor() : ContentScopeJsMessageHandlersPlugin {
+class EventHubContentScopeJsMessageHandler @Inject constructor(
+    private val pixelManager: RealEventHubPixelManager,
```
cmonfortep: Please define an interface and inject the interface instead of concrete class.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
```diff
@@ -0,0 +1,333 @@
...
+import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.launch
+import logcat.LogPriority.DEBUG
+import logcat.LogPriority.VERBOSE
+import logcat.logcat
+import org.json.JSONObject
+import java.util.concurrent.ConcurrentHashMap
+import javax.inject.Inject
+
+class RealEventHubPixelManager @Inject constructor(
```
cmonfortep, line 67: This doesn't match file name. Also please define the interface. `Real...` can be used for concrete implementation.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
```diff
@@ -0,0 +1,335 @@
...
+            logcat(VERBOSE) { "EventHub: navigation detected for tab $tabId ($previousUrl -> $url), clearing dedup" }
+            dedupSeen.removeAll { it.endsWith(":$tabId") }
+        }
+    }
+
+    fun handleWebEvent(data: JSONObject, tabId: String) {
+        val eventType = data.optString("type", "")
+        if (eventType.isEmpty()) return
+
+        val config = getParsedConfig()
```
cmonfortep: This method ends up accessing DataStore down the stream, ensure that doesn't happen in main thread as can cause ANRs.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/di/EventHubModule.kt
```diff
@@ -0,0 +1,91 @@
...
+    @Provides
+    fun provideEventHubPixelStateDatabase(context: Context): EventHubPixelStateDatabase {
+        return Room.databaseBuilder(context, EventHubPixelStateDatabase::class.java, "event_hub_pixel_state.db")
+            .enableMultiInstanceInvalidation()
+            .fallbackToDestructiveMigration()
+            .addMigrations(*ALL_MIGRATIONS)
+            .build()
+    }
+
+    @SingleInstanceIn(AppScope::class)
```
cmonfortep: NIT: Dependencies outside datastore and room don't need to be declared here, you can rely on class annotations.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/WebEventsDataStore.kt
```diff
@@ -0,0 +1,93 @@
...
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.di.scopes.AppScope
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+import javax.inject.Qualifier
+
+@Qualifier
```
cmonfortep: NIT: This is better declared inside Module `EventHubModule` as it's part of DI.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubLifecycleObserver.kt
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * On app foreground: fire any elapsed pixels, then reschedule timers for remaining ones.
+ */
+@ContributesMultibinding(
+    scope = AppScope::class,
+    boundType = MainProcessLifecycleObserver::class,
+)
+class EventHubLifecycleObserver @Inject constructor(
+    private val pixelManager: RealEventHubPixelManager,
```
cmonfortep: Should be injected interface.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubLifecycleObserver.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val dispatcherProvider: DispatcherProvider,
+    @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
+) : MainProcessLifecycleObserver {
+
+    @UiThread
+    override fun onStart(owner: LifecycleOwner) {
```
cmonfortep, line 38: Is the intention to trigger this every time the app moves from background to foreground? or what's is this trying to cover?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubNavigationPlugin.kt
```diff
@@ -14,28 +14,31 @@
...
-        } else {
-            false
-        }
+    override fun onPageStarted(
+        webView: WebView,
+        url: String?,
+        isDesktopMode: Boolean?,
+        activeExperiments: List<Toggle>,
+    ) {
+        val tabId = System.identityHashCode(webView).toString()
```
cmonfortep: NIT: not tabId anymore.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubJsInjectorPlugin.kt
```diff
@@ -14,28 +14,31 @@
...
 
-    override fun store(featureName: String, jsonString: String): Boolean {
-        return if (featureName == this.featureName) {
-            return runBlocking {
-                browserUiLockRepository.insertJsonData(jsonString)
-            }
-        } else {
-            false
-        }
+    override fun onPageStarted(
```
cmonfortep, line 35: annotate as UIthread for clarity

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubFeaturePlugin.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.privacy.config.api.PrivacyFeaturePlugin
+import com.squareup.anvil.annotations.ContributesMultibinding
+import logcat.LogPriority.DEBUG
+import logcat.logcat
+import javax.inject.Inject
+
+@ContributesMultibinding(AppScope::class)
+class EventHubFeaturePlugin @Inject constructor(
```
cmonfortep: Why are you using custom implementation if you are storing settings as plain json?
It seems you could rely on `ContributesRemoteFeature` and `Toggle` which gives you:

- enable/disable methods (respecting minSupportedVersion, and other attributes)
- settings store (as json)

```
/**
     * @return a JSON string containing the `settings` of the feature or null if not present in the remote config
*/
fun getSettings(): String?
```

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubFeaturePlugin.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val repository: EventHubRepository,
+    private val pixelManager: RealEventHubPixelManager,
+) : PrivacyFeaturePlugin {
+
+    override fun store(featureName: String, jsonString: String): Boolean {
+        if (featureName == this.featureName) {
+            logcat(DEBUG) { "EventHub: storing eventHub config (${jsonString.length} chars)" }
+            repository.setEventHubConfigJson(jsonString)
+            pixelManager.onConfigChanged()
```
cmonfortep: Here doesn't mean the config has changed for you, but that config has been downloaded and persisted. Depending on your needs this is ok, or extra logic might be needed.

Also if you don't need a custom implementation, you can rely on existing callbacks `onPrivacyConfigDownloaded` or `onPrivacyConfigPersisted`

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubDataStore.kt
```diff
@@ -0,0 +1,93 @@
...
+import dagger.SingleInstanceIn
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+import javax.inject.Qualifier
+
+@Qualifier
+internal annotation class EventHubPrefs
+
+interface EventHubDataStore {
```
cmonfortep: This datastore might be unnecessary. See `Toggle` comment for reference.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/webevents/WebEventsFeaturePlugin.kt
```diff
@@ -0,0 +1,59 @@
...
+            return true
+        }
+        return false
+    }
+
+    override val featureName: String = EventHubFeatureName.EventHub.value
+}
+
+@ContributesMultibinding(AppScope::class)
+class WebEventsFeaturePlugin @Inject constructor(
```
cmonfortep, line 29: ditto

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
cmonfortep: on every event you are reading a plain json and parsing into a domain object, why not introducing something that already gives you the object parsed?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
```diff
@@ -0,0 +1,335 @@
...
+        if (!config.featureEnabled) return
+
+        val nowMillis = timeProvider.currentTimeMillis()
+
+        synchronized(this) {
+            for (state in repository.getAllPixelStates()) {
+                val storedConfig = EventHubConfigParser.parseSinglePixelConfig(state.pixelName, state.configJson) ?: continue
```
cmonfortep: Instead of doing this parsing here, you can consider the repository already gives you the right data as that the class is interacting directly with database.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
```diff
@@ -0,0 +1,335 @@
...
+
+                if (nowMillis > state.periodEndMillis) continue
+
+                val params = parseParamsJson(state.paramsJson)
```
cmonfortep: ditto about repository knowing better the datastore details

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubRepository.kt
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.eventhub.impl
+
+import com.duckduckgo.eventhub.impl.store.EventHubPixelStateDao
+import com.duckduckgo.eventhub.impl.store.EventHubPixelStateEntity
+
+interface EventHubRepository {
```
cmonfortep, line 29: This class has some methods that will access database. For those make sure from the caller you are using this on background thread.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfigParser.kt
```diff
@@ -0,0 +1,158 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.eventhub.impl
+
+import logcat.LogPriority.WARN
+import logcat.logcat
+import org.json.JSONObject
+
+object EventHubConfigParser {
```
cmonfortep, line 56: Pretty much all the code you are doing here you can use Moshi instead of going for JsonObject, unless this is intended.

cmonfortep:
Forgot to comment on this, you are relying on each pixel definition that lives inside settings (including the state). You should also support the enabled state present at the top level, which should serve as kill switch for all of them, and also support minSupportedVersion in case we need to take actions that affect all pixels.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
GuiltyDolphin: Have added a cached copy so we don't parse each time.

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
GuiltyDolphin: Was this for an older commit? We're in the `isUrlAllowed` check now unless I'm missing something

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
GuiltyDolphin: Got Cursor to assess the paths, let me know if you have any concerns and I can try taking a deeper look.

> `getEventHubConfigJson()` reads from a `@Volatile` in-memory cache (`cachedEventHubJson`) — it doesn't touch DataStore on the calling thread. The DataStore is only read asynchronously in the `init` collector (on `dispatcherProvider.io()`), and writes are also dispatched to IO. So no main thread DataStore access or ANR risk here.

> No ANR risk — all paths are off the main thread:

- **`handleWebEvent`** — called from `@JavascriptInterface` (WebCore background thread). Room DAO calls (`getAllPixelStates`, `savePixelState`) happen there.
- **`checkPixels`** — called from `EventHubLifecycleObserver.onStart()`, dispatched to `dispatcherProvider.io()`.
- **`onConfigChanged`** — called from `EventHubFeaturePlugin.store()`, which runs inside `persistPrivacyConfig` (a `suspend fun` called from coroutines in `LocalPrivacyConfigObserver` / `RealPrivacyConfigDownloader`).
- **`scheduleFireTelemetry`** — launches on `dispatcherProvider.io()`.
- **`getEventHubConfigJson()`** — reads a `@Volatile` in-memory field (`cachedEventHubJson`), no DataStore I/O on the calling thread. DataStore is only read asynchronously in the `init` collector on IO.

GuiltyDolphin:
> Forgot to comment on this, you are relying on each pixel definition that lives inside settings (including the state). You should also support the enabled state present at the top level, which should serve as kill switch for all of them, and also support minSupportedVersion in case we need to take actions that affect all pixels.

Right, the global enabled/disabled state of the feature serves as a kill switch. This was manually modeled, but your suggestion of using Toggle simplifies this :)

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubDataStore.kt
GuiltyDolphin: I removed the EventHubDataStore. There's still one for webEvents which is a CSS feature (from what I can see this is how we handle other CSS features too).

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubFeaturePlugin.kt
GuiltyDolphin: Switched to using Toggle

GuiltyDolphin:
Couple more comments to get to, will do in the morning :slightly_smiling_face:

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubRepository.kt
GuiltyDolphin: Had Cursor investigate this one too. Here's the response, but let me know if you have any concerns:

## No main thread DB access — not an issue

Every call path that hits `EventHubRepository` (Room DB) runs on a background thread:

| Method | Caller | Thread |
|---|---|---|
| `handleWebEvent` | `EventHubContentScopeJsMessageHandler.process()` via `ContentScopeScriptsJsMessaging` `@JavascriptInterface` | WebView JS interface thread (background) |
| `checkPixels` | `EventHubLifecycleObserver.onStart()` inside `appCoroutineScope.launch(dispatcherProvider.io())` | IO dispatcher |
| `onConfigChanged` | `EventHubPrivacyConfigCallback.onPrivacyConfigPersisted()` called from `RealPrivacyConfigPersister` (`@WorkerThread`, `suspend fun`) | Worker/IO thread |
| `scheduleFireTelemetry` timer | `appCoroutineScope.launch(dispatcherProvider.io())` | IO dispatcher |
| `fireTelemetry` / `startNewPeriod` | Called transitively from `checkPixels`, `onConfigChanged`, or timer — all above | Background |

`isFeatureEnabled()` uses the Toggle API (cached in memory), and `WebEventsDataStore.getWebEventsConfigJson()` returns a `@Volatile` cached field — neither hits disk synchronously.

No main thread database access, no ANR risk.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubPixelManager.kt
GuiltyDolphin: Moved the parsing into Repository

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubLifecycleObserver.kt
GuiltyDolphin: Yep this is intended to run on every foreground. In the background/app killed timers might not survive reliably, so we:

1. Fire any pixels whose period elapsed whilst backgrounded
2. Reschedule timers for any in-progress pixels
3. Start periods for any pixels that are enabled but have no active state (e.g., timer elapsed whilst in BG--we don't start a new period in BG as we want pixel tied to activity)

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubFeaturePlugin.kt
GuiltyDolphin: I think this is fine now:
- switched to using `onPrivacyConfigPersisted` to trigger this
- when this triggers, it is convergent (we clear a cache but that's very cheap to rebuild, states are already in memory so this is largely no-ops unless something actually changed)

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/RealEventHubPixelManager.kt
```diff
@@ -0,0 +1,348 @@
...
+        } else {
+            logcat(VERBOSE) { "EventHub: skipping pixel ${pixelConfig.name}, no params" }
+        }
+
+        repository.deletePixelState(pixelConfig.name)
+
+        val latestPixelConfig = getTelemetryConfigs().find { it.name == pixelConfig.name }
+        if (latestPixelConfig != null) {
+            startNewPeriod(latestPixelConfig)
+        }
```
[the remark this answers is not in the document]
GuiltyDolphin: This is intentional as de-duping is intended to span navigations, not periods. If a period ends we don't want to attribute events we would have de-duped in the last period to the new period, until a navigation occurs.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/RealEventHubPixelManager.kt
```diff
@@ -0,0 +1,348 @@
...
+    private val foregroundStateProvider: AppForegroundStateProvider,
+    private val eventHubFeature: EventHubFeature,
+) : EventHubPixelManager {
+
+    @Volatile
+    private var cachedTelemetryConfigs: List<TelemetryPixelConfig>? = null
+
+    private val dedupSeen: MutableSet<String> = ConcurrentHashMap.newKeySet()
+    private val webViewCurrentUrl = ConcurrentHashMap<String, String>()
+    private val scheduledTimers = ConcurrentHashMap<String, Job>()
```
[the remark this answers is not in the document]
GuiltyDolphin: From Cursor:

Low risk in practice — `identityHashCode` keys get reused when old WebViews are GC'd and new ones allocate at the same address. When that happens, `onNavigationStarted` overwrites the stale `webViewCurrentUrl` entry and clears the associated `dedupSeen` keys, so hash reuse self-cleans.

The only leak is from destroyed WebViews whose hash is never reused, which is bounded by JVM address space reuse patterns. Android typically has 1-5 active WebViews, so unbounded growth isn't realistic.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubLifecycleObserver.kt
```diff
@@ -0,0 +1,52 @@
...
+        logcat(VERBOSE) { "EventHub: app foregrounded, checking pixels" }
+        appCoroutineScope.launch(dispatcherProvider.io()) {
+            pixelManager.checkPixels()
+        }
+    }
```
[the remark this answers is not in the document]
GuiltyDolphin: From Cursor:

False positive — both `EventHubLifecycleObserver.onStart` and `EventHubAppForegroundStateProvider.onStart` are called sequentially on the main thread by the lifecycle dispatcher. `checkPixels()` is dispatched to IO via `appCoroutineScope.launch(dispatcherProvider.io())`, which involves a thread hop. By the time the IO dispatcher picks up the coroutine, the main thread has already finished calling all `onStart` observers, so `isInForeground` is guaranteed to be `true`.

The race would require the IO dispatcher to execute between two sequential main-thread calls, which can't happen — the main thread isn't preempted mid-dispatch.

GuiltyDolphin:
@cmonfortep that should be all the existing comments addressed! Switched to using Toggle and avoided the re-parsing on each event. Tidied up the module structure.

Had a few rounds of bugbot feedback. Some false positives, caught a couple of valid, if unlikely, race conditions.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubLifecycleObserver.kt
```diff
@@ -0,0 +1,53 @@
...
```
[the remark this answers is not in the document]
GuiltyDolphin: This is the same as https://github.com/duckduckgo/Android/pull/7754#discussion_r2877781158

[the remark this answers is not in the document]
GuiltyDolphin: @cursoragent can you validate if this can actually happen? Looks like a false positive but want to be sure.

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
cmonfortep: I'm asking about the check in the previous condition
`if (this.secret == secret && context == jsMessage.context && isUrlAllowed(allowedDomains, domain))`

you have `isUrlAllowed(allowedDomains, domain)`, is that intended here?

event-hub/event-hub-impl/src/test/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManagerTest.kt
```diff
@@ -0,0 +1,2552 @@
...
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class EventHubPixelManagerTest {
+
+    private val repository: EventHubRepository = mock()
+    private val pixel: Pixel = mock()
+    private val timeProvider = FakeTimeProvider()
+    private val foregroundState = FakeAppForegroundStateProvider()
+    private val eventHubFeature: EventHubFeature = mock()
+    private val selfToggle: Toggle = mock()
```
cmonfortep: Please, check how we use `FakeFeatureToggleFactory` to create a Toggle when unit tests

event-hub/event-hub-impl/src/test/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManagerTest.kt
```diff
@@ -0,0 +1,2552 @@
...
+
+    private fun webEventData(type: String) = JSONObject().put("type", type)
+
+    private lateinit var testScope: TestScope
```
cmonfortep: Should use:
```
    @get:Rule
    var coroutineRule = CoroutineTestRule()
```

from that rule you can get the scope and dispatcher.

event-hub/event-hub-impl/src/test/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManagerTest.kt
```diff
@@ -0,0 +1,2552 @@
...
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import java.util.concurrent.TimeUnit
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class EventHubPixelManagerTest {
```
cmonfortep: Should use the actual class name under test `RealEventHubPixelManagerTest`

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+            for (pixelConfig in getTelemetryConfigs()) {
+                if (repository.getPixelState(pixelConfig.name) == null) {
+                    startNewPeriod(pixelConfig)
+                }
+            }
+        }
+    }
+
+    override fun onConfigChanged() {
+        synchronized(this) {
```
cmonfortep: onConfigChange, the caller method will wait for you to finish. Validate if you can launch a background coroutine instead so you don't block, this is how we do the same in other features. You receive the callback, you trigger your refresh job.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+
+    override fun handleWebEvent(data: JSONObject, webViewId: String) {
+        val eventType = data.optString("type", "")
+        if (eventType.isEmpty()) return
+
+        if (!isFeatureEnabled()) return
+
+        val nowMillis = timeProvider.currentTimeMillis()
+
+        synchronized(this) {
```
cmonfortep: Synchronized is a blocking operation that will block the thread and other threads trying to access the lock (including other methods that use the synchronizedin this class), that is something to avoid.
Our codebase relies on coroutines so this doesn't align with the purpose of nonblocking but suspend. There are other approached possible, but all depends on what it's the intend here, how much complexity you want, and how necessary this is. This should be reviewed.

This is not exclusively for this method (which comes from JS java bridge), but the usage within this class.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+            logcat(DEBUG) { "EventHub: onConfigChanged — feature enabled, ${telemetry.size} telemetry pixel(s) in config" }
+            for (pixelConfig in telemetry) {
+                if (repository.getPixelState(pixelConfig.name) == null) {
+                    startNewPeriod(pixelConfig)
+                }
+            }
+        }
+    }
+
+    fun scheduleFireTelemetry(pixelName: String, delayMillis: Long) {
```
cmonfortep: This is not declared in the interface. Noticed this should be private method, only used within this class.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+        val job = appCoroutineScope.launch(dispatcherProvider.io()) {
+            delay(delayMillis)
+            ensureActive()
+
+            if (!isFeatureEnabled()) {
+                scheduledTimers.remove(pixelName)
+                return@launch
+            }
+
+            synchronized(this@RealEventHubPixelManager) {
```
cmonfortep: review synchronized here.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+                    scheduledTimers.remove(pixelName)
+                    return@launch
+                }
+                fireTelemetry(pixelState)
+            }
+        }
+        scheduledTimers[pixelName] = job
+    }
+
+    fun hasScheduledTimer(pixelName: String): Boolean = scheduledTimers.containsKey(pixelName)
```
cmonfortep: public method not in the interface. Also seems only used in Tests? exposing a public method only for tests is a smell and should be avoided.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+
+    fun cancelScheduledFire(pixelName: String) {
```
cmonfortep: public method not in the interface.
Usage is within this class only and tests. Seems should be private, and test cases to be reviewed.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+            logcat(VERBOSE) { "EventHub: dedup $key (already seen on current page)" }
+            return true
+        }
+        return false
+    }
+
+    /**
+     * Check all pixel states and fire any whose period has elapsed.
+     * After firing, starts a new period and schedules the next fire.
+     * Called on app foreground to catch pixels that elapsed while backgrounded,
```
cmonfortep: Best to document the interface.

Also `Called on app foreground`, you should not document where this is called from, that can change. This should rely on documentation to the external callers.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+     * and to start new periods for enabled configs that have no active state
+     * (e.g., timer fired while backgrounded and no new period was started).
+     */
+    override fun checkPixels() {
+        if (!isFeatureEnabled()) return
+
+        val nowMillis = timeProvider.currentTimeMillis()
+
+        synchronized(this) {
```
cmonfortep: this synchronized needs to be changed. We should not block a thread.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubRepository.kt
```diff
@@ -0,0 +1,107 @@
...
+            periodEndMillis = entity.periodEndMillis,
+            config = config,
+            params = parseParamsJson(entity.paramsJson),
+        )
+    }
+
+    companion object {
+        fun parseParamsJson(json: String): MutableMap<String, ParamState> {
+            return try {
+                val obj = JSONObject(json)
```
cmonfortep: why jsonObject instead of using Moshi?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubRepository.kt
```diff
@@ -0,0 +1,107 @@
...
+                    }
+                }
+                map
+            } catch (e: Exception) {
+                mutableMapOf()
+            }
+        }
+
+        fun serializeParams(params: Map<String, ParamState>): String {
+            val obj = JSONObject()
```
cmonfortep: ditto

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/webevents/WebEventsDataStore.kt
```diff
@@ -0,0 +1,73 @@
...
+                cachedWebEventsJson = prefs[Keys.WEB_EVENTS_RC] ?: EMPTY_JSON
+            }
+        }
+    }
+
+    override fun getWebEventsConfigJson(): String = cachedWebEventsJson
+
+    override fun setWebEventsConfigJson(value: String) {
+        cachedWebEventsJson = value
+        appCoroutineScope.launch(dispatcherProvider.io()) {
```
cmonfortep: you can remove this launch here

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+            logcat(VERBOSE) { "EventHub: timer already scheduled for $pixelName, skipping" }
+            return
+        }
+
+        logcat(VERBOSE) { "EventHub: scheduling fire for $pixelName in ${delayMillis}ms" }
+        val job = appCoroutineScope.launch(dispatcherProvider.io()) {
+            delay(delayMillis)
+            ensureActive()
+
+            if (!isFeatureEnabled()) {
```
cmonfortep: when this is disabled, you remove and return launch, but after the launch you add the job again into the scheduledTimers.

revisit how this method should work when feature disabled. Probably first this you want to do is to check the feature flag, remove and ensure no extra logic.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+
+    fun scheduleFireTelemetry(pixelName: String, delayMillis: Long) {
+        if (scheduledTimers.containsKey(pixelName)) {
+            logcat(VERBOSE) { "EventHub: timer already scheduled for $pixelName, skipping" }
+            return
+        }
+
+        logcat(VERBOSE) { "EventHub: scheduling fire for $pixelName in ${delayMillis}ms" }
+        val job = appCoroutineScope.launch(dispatcherProvider.io()) {
+            delay(delayMillis)
```
cmonfortep, line 246: what this delay is doing here?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+            ensureActive()
```
cmonfortep, line 247: ditto

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,321 @@
...
+
+    private fun isFeatureEnabled(): Boolean = eventHubFeature.self().isEnabled()
+
+    private fun getTelemetryConfigs(): List<TelemetryPixelConfig> {
+        cachedTelemetryConfigs?.let { return it }
+        val settingsJson = eventHubFeature.self().getSettings() ?: return emptyList()
+        return EventHubConfigParser.parseTelemetry(settingsJson).also { cachedTelemetryConfigs = it }
+    }
+
+    override fun onNavigationStarted(webViewId: String, url: String) {
```
cmonfortep, line 112: anything to do here if FF disabled?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/webevents/WebEventsDataStore.kt
GuiltyDolphin: Couldn't figure this one out. Cursor suggested `runBlocking` or propagating `suspend`, but those didn't seem ideal? Another option was to switch to a Room entity, but from Cris' [feedback](https://github.com/duckduckgo/Android/pull/7614#discussion_r2852421146) on another PR, I understand this wouldn't be desirable for a feature where we're just passing state to CSS.

Here's what Cursor had to say, let me know if this sounds correct or if you had something in mind:

> The `launch` is needed here because `webEvents` isn't in the bundled `privacy_config.json` — it only arrives via remote config download (periodic, up to 1 hour). Without the DataStore persistence, `WebEventsContentScopeConfigPlugin.config()` would return `"{}"` after every app restart until the next remote download, silently disabling web events detection.

> The pattern matches `WebDetectionDataStore.setRemoteConfigJson` — update the in-memory cache synchronously (so reads see the new value immediately), persist to DataStore via `launch` (fire-and-forget, since the caller doesn't need to wait for disk).

cmonfortep: You should keep the datastore, not use room for this case.

I commented you can remove the launch because this method is already called from an io dispatcher. Launching a new coroutine for that operation seems not necessary and simplifies your flow. If you want to do fire and forget keep it, but consider the action will happen when the coroutine runs, might not be immediate. You could add `suspend` to force the caller to be on a coroutine. That will force the caller to be on a coroutine or force other methods to put `suspend` as well, if that doesn't change public apis, you can do that. This is a judgement call.

Summary is: this method should be already called from io, it's a quick operation, so launch seems not necessary and will ensure that happens when you expect.

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/messaging/ContentScopeScriptsJsMessaging.kt
GuiltyDolphin: As discussed - shouldn't be an issue for this feature as the allowed domains is empty so `isUrlAllowed` should always pass.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/webevents/WebEventsDataStore.kt
GuiltyDolphin: Okay it looks like the `suspend` would propagate to a public API (`PrivacyFeaturePlugin.store`).

I've pushed the launch up to the web events plugin (out of the web events store), this seems to better match duck player.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
GuiltyDolphin: was this ditto on the delay?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
GuiltyDolphin: discussed on call. If feature flag isn't cached then we'll need to store the state in onConfigChanged

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
GuiltyDolphin: shouldn't be an issue now we're using ConflatedJob

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubLifecycleObserver.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val dispatcherProvider: DispatcherProvider,
+    @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
+) : MainProcessLifecycleObserver {
+
+    @UiThread
+    override fun onStart(owner: LifecycleOwner) {
+        logcat(VERBOSE) { "EventHub: app foregrounded, checking pixels" }
+        appCoroutineScope.launch(dispatcherProvider.io()) {
+            pixelManager.checkPixels()
+        }
```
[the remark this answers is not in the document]
GuiltyDolphin: @cursoragent check if still relevant

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,330 @@
...
+        repository.deletePixelState(pixelState.pixelName)
+
+        val latestPixelConfig = getTelemetryConfigs().find { it.name == pixelState.pixelName }
+        if (latestPixelConfig != null) {
+            startNewPeriod(latestPixelConfig)
+        }
+    }
+
+    private fun startNewPeriod(pixelConfig: TelemetryPixelConfig) {
+        if (!foregroundStateProvider.isInForeground || !isFeatureEnabled() || !pixelConfig.isEnabled) {
```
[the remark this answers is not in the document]
GuiltyDolphin: @cursoragent is this still relevant?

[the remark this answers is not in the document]
GuiltyDolphin: @cursoragent is this relevant anymore?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/EventHubLifecycleObserver.kt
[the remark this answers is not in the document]
GuiltyDolphin: @cursoragent is this relevant anymore?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
GuiltyDolphin: Looks like this is cached via `CachedToggleStore`.

I've added a guard to exit early if the feature is disabled.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,333 @@
...
+
+    override fun isEnabled(): Boolean = eventHubFeature.self().isEnabled()
+
+    override fun onAppForegrounded() {
+        isInForeground = true
+        checkPixels()
+    }
+
+    private fun checkPixels() {
+        if (!isEnabled()) return
```
cmonfortep: Could you move this into the coroutine block as first step?
I'm seeing in the logs disk violation on app launch, probably cache no hit on fresh launch.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfigParser.kt
```diff
@@ -0,0 +1,146 @@
...
+                TelemetryParameterJson(
+                    template = paramConfig.template,
+                    source = paramConfig.source,
+                    buckets = paramConfig.buckets.mapValues { (_, bucket) ->
+                        BucketJson(gte = bucket.gte, lt = bucket.lt)
+                    },
+                )
+            },
+        )
+        return pixelAdapter.toJson(pixelJson)
```
cmonfortep: is it possible to guard against an exception on this method call, or on the caller side?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubRepository.kt
```diff
@@ -0,0 +1,96 @@
...
+            pixelName = entity.pixelName,
+            periodStartMillis = entity.periodStartMillis,
+            periodEndMillis = entity.periodEndMillis,
+            config = config,
+            params = parseParamsJson(entity.paramsJson),
+        )
+    }
+
+    companion object {
+        private val moshi: Moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
```
cmonfortep: Best to convert this to:

```
private val paramsAdapter by lazy {
          val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()                                                      
          val paramsType = Types.newParameterizedType(Map::class.java, String::class.java, ParamState::class.java)               
          moshi.adapter<Map<String, ParamState>>(paramsType).lenient()                                                             
      }
```

Reason: As companion object, this is init when the outer class is injected. That happens on first used, which happens inside `EventHubJsInjectorPlugin`, and those plugins are first used on webview callbacks. So instance creation will happen in main thread (`KotlinJsonAdapterFactory` uses reflection and has higher cost). With by lazy you defer creating the instance to the moment where you actually use it, which happens on a background thread. Will avoid ANR risk.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -0,0 +1,333 @@
...
+                        val paramState = updatedParams[paramName] ?: ParamState(0)
+                        if (paramState.stopCounting) continue
+                        if (isDuplicateEvent(pixelState.pixelName, paramName, eventType, webViewId)) continue
+
+                        changed = true
+                        if (BucketCounter.shouldStopCounting(paramState.value, paramConfig.buckets)) {
+                            updatedParams[paramName] = paramState.copy(stopCounting = true)
+                            logcat(VERBOSE) { "EventHub: ${pixelState.pixelName}.$paramName already at max bucket, stopCounting" }
+                            continue
+                        }
```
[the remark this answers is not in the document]
GuiltyDolphin: This is very confusingly phrased. Even if persisting were to fail we'd just check again next time.

PR #7773, opened 2026-02-19
PIR: Add a pixel to know if initial foreground scan didn't complete

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203581873609357/task/1211563456620837?focus=true

### Description
Adds a pixel to know if the initial foreground scan is not running and there are still not executed scan jobs.

This can happen if either the system or the user force closes the app.

### Steps to test this PR
See https://app.asana.com/1/137249556945/project/1203581873609357/task/1213346151426402?focus=true

### UI changes
No UI changes

---

> [!NOTE]
> **Low Risk**
> Adds a new analytics pixel and lightweight detection logic without changing scan execution or data handling behavior.
> 
> **Overview**
> Adds a new PIR telemetry event `m_dbp_initial-scan_incomplete` and wires it into the dashboard initial-scan status flow to detect interrupted initial foreground scans.
> 
> The dashboard now checks (once per activity launch) whether the foreground scan service is not running *and* there are remaining unexecuted scan jobs, and emits the new pixel when that condition is met; this includes new state-provider logic plus a service-running check, and accompanying unit tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bd4e69a61423199d8fcbd61b43b0b2486f30c070. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/messaging/handlers/PirWebInitialScanStatusMessageHandler.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/messaging/handlers/PirWebInitialScanStatusMessageHandler.kt
@@ -75,7 +75,7 @@ class PirWebInitialScanStatusMessageHandler @Inject constructor(
 
             // Check once per PirDashboardWebViewActivity launch if we need to resume a scan that was interrupted (e.g., by app kill)
             if (!checkedForResumeScan.getAndSet(true)) {
-                resumeInitialScanIfNeeded()
+                checkForIncompleteInitialScan()
             }
 
             jsMessaging.sendResponse(
@@ -100,7 +100,7 @@ class PirWebInitialScanStatusMessageHandler @Inject constructor(
      * Checks if the initial foreground scan should be resumed (was interrupted with remaining brokers).
      * Currently emits a pixel to measure how often this scenario occurs.
      */
-    private suspend fun resumeInitialScanIfNeeded() {
+    private suspend fun checkForIncompleteInitialScan() {
         if (!stateProvider.shouldRestartInitialScan()) {
             logcat { "PIR-WEB: No need to resume scan, it's either not started or already completed" }
             return


landomen:
* **#7773** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7773?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7773?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirDashboardInitialScanStateProvider.kt
```diff
@@ -113,4 +124,25 @@ class RealPirDashboardInitialScanStateProvider @Inject constructor(
...
+        // Check if a scan is already running
+        if (PirForegroundScanService.isServiceRunning(context)) {
+            logcat { "PIR-WEB: Scan is already running, no need to resume initial scan" }
+            return false
+        }
+
+        // Get all scan jobs that haven't been executed yet
+        val notExecutedJobs = pirSchedulingRepository.getAllValidScanJobRecords().filter { record ->
+            record.status == ScanJobStatus.NOT_EXECUTED && record.lastScanDateInMillis == 0L
+        }
```
landomen, line 138: I didn't add a check for pending opt out jobs, since the initial scan seems more important as it impacts the UI. 

However there is an edge case if the app is killed/stopped after scans finish but opt-outs are in progress. They will only get picked up on next background run. But I can easily add opt-outs here as well if we want the entire initial scan to finish in foreground.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scan/PirForegroundScanService.kt
```diff
@@ -137,5 +139,25 @@ class PirForegroundScanService : Service(), CoroutineScope by MainScope() {
...
+                return false
+            }
+
+            for (service in manager.getRunningServices(Int.MAX_VALUE)) {
+                if (PirForegroundScanService::class.java.name == service.service.className) {
+                    return true
+                }
+            }
+            return false
+        }
```
landomen, line 161: Note that this is copied from `TrackerBlockingVpnService` and seems to work reliably even on Android 16.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/state/PirDashboardInitialScanStateProvider.kt
karlenDimla: I think we should only do this for scans. Ideally if any job (scan or optout) are not completed within the foreground run, it should be OK as they will picked up in the next background run. 

I do see why you wanted to do this for scans because as you said that will make the state of the dashboard stuck in in progress.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/dashboard/messaging/handlers/PirWebInitialScanStatusMessageHandler.kt
```diff
@@ -87,6 +96,21 @@ class PirWebInitialScanStatusMessageHandler @Inject constructor(
         return pirRepository.getValidUserProfileQueries().isNotEmpty()
     }
 
+    /**
+     * Checks if the initial foreground scan should be resumed (was interrupted with remaining brokers).
+     * Currently emits a pixel to measure how often this scenario occurs.
+     */
+    private suspend fun resumeInitialScanIfNeeded() {
```
0nko: Do you plan to add the actual scan resuming logic? Because the name right now is a bit misleading.

landomen: Good catch, I'll rename this. I had the resuming login there initially, but we decided to change it for now to just sent a pixel since there could be side effects from restarting the scan.

PR #7818, opened 2026-02-26
Route URLs typed in Duck.ai toggle mode to browser instead of Duck.ai

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211654189969294/task/1213449554902056?focus=true

### Description
When a user has the Duck.ai omnibar toggle enabled and types a URL (e.g. `bbc.com`), the app was sending it to Duck.ai as a prompt instead of navigating to it. This PR detects URL inputs using `QueryUrlPredictor` and routes them to the browser instead.

Changes:
- Moved `QueryUrlPredictor` interface to `browser-api` so it can be consumed by `duckchat-impl`
- `InputScreenViewModel.onChatSubmitted`: check `isNavigate()` before wrapping the query in a Duck.ai URL
- `BrowserTabViewModel.onDuckChatOmnibarButtonClicked`: same fix for the direct omnibar button path
- Fallback to `UriString.isWebUrl` when the native predictor library hasn't initialised yet

### Steps to test this PR

_URL in Duck.ai toggle mode_
- [x] Enable the Duck.ai omnibar toggle in Settings
- [x] Type `bbc.com` in the address bar and submit — should navigate to bbc.com, not open Duck.ai
- [x] Type a plain search query (e.g. `what is AI`) and submit — should open Duck.ai as before
- [x] Type a full URL (`https://example.com`) and submit — should navigate to the URL

### UI changes
| Before  | After |
| ------ | ----- |
|`bbc.com` opens Duck.ai with prompt|`bbc.com` navigates to bbc.com|

---

> [!NOTE]
> **Medium Risk**
> Changes URL-vs-search classification for Duck.ai submission paths and adds a remote-flagged native URL predictor fallback, which could affect navigation behavior and startup edge cases.
> 
> **Overview**
> Fixes Duck.ai toggle submission so **URL-like inputs navigate in the browser instead of being sent as Duck.ai prompts**, in both `BrowserTabViewModel.onDuckChatOmnibarButtonClicked` and DuckChat’s `InputScreenViewModel.onChatSubmitted`.
> 
> Moves the `QueryUrlPredictor` interface into `browser-api` (and exports `url-predictor-android` via an `api` dependency), and extends the app impl with `isUrl()` that is **feature-flag gated** and falls back to `UriString.isWebUrl()` until the native predictor is initialized; tests are updated/added to cover URL routing and fallback behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8041782873c46a010ca08b7c827628f48ebaef17. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -369,7 +369,6 @@ import com.duckduckgo.subscriptions.api.SUBSCRIPTIONS_FEATURE_NAME
 import com.duckduckgo.subscriptions.api.Subscriptions
 import com.duckduckgo.subscriptions.api.SubscriptionsJSHelper
 import com.duckduckgo.sync.api.favicons.FaviconsFetchingPrompt
-import com.duckduckgo.urlpredictor.Decision
 import dagger.Lazy
 import io.reactivex.schedulers.Schedulers
 import kotlinx.coroutines.CoroutineScope
@@ -4702,13 +4701,6 @@ class BrowserTabViewModel @Inject constructor(
         }
     }
 
-    private fun isNavigate(query: String): Boolean =
-        if (queryUrlPredictor.isReady()) {
-            queryUrlPredictor.classify(query) is Decision.Navigate
-        } else {
-            UriString.isWebUrl(query)
-        }
-
     fun onDuckChatOmnibarButtonClicked(
         query: String?,
         hasFocus: Boolean,
@@ -4737,7 +4729,7 @@ class BrowserTabViewModel @Inject constructor(
             duckAiFeatureState.showFullScreenMode.value -> {
                 val url = when {
                     hasFocus && isNtp && query.isNullOrBlank() -> duckChat.getDuckChatUrl(query ?: "", false)
-                    hasFocus && isNavigate(query ?: "") -> query ?: ""
+                    hasFocus && queryUrlPredictor.isUrl(query ?: "") -> query ?: ""
                     hasFocus -> duckChat.getDuckChatUrl(query ?: "", true)
                     else -> duckChat.getDuckChatUrl(query ?: "", false)
                 }
@@ -4747,7 +4739,7 @@ class BrowserTabViewModel @Inject constructor(
             else -> {
                 when {
                     hasFocus && isNtp && query.isNullOrBlank() -> duckChat.openDuckChat()
-                    hasFocus && isNavigate(query ?: "") -> onUserSubmittedQuery(query ?: "")
+                    hasFocus && queryUrlPredictor.isUrl(query ?: "") -> onUserSubmittedQuery(query ?: "")
                     hasFocus -> duckChat.openDuckChatWithAutoPrompt(query ?: "")
                     else -> duckChat.openDuckChat()
                 }
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
@@ -78,7 +78,6 @@ import com.duckduckgo.duckchat.impl.pixel.DuckChatPixelName.DUCK_CHAT_EXPERIMENT
 import com.duckduckgo.duckchat.impl.pixel.DuckChatPixelParameters
 import com.duckduckgo.duckchat.impl.pixel.inputScreenPixelsModeParam
 import com.duckduckgo.history.api.NavigationHistory
-import com.duckduckgo.urlpredictor.Decision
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4699,6 +4702,13 @@ class BrowserTabViewModel @Inject constructor(
         }
     }
 
+    private fun isNavigate(query: String): Boolean =
+        if (queryUrlPredictor.isReady()) {
+            queryUrlPredictor.classify(query) is Decision.Navigate
+        } else {
+            UriString.isWebUrl(query)
+        }
```
malmstein: nit: `QueryUrlConverter` already gates the predictor behind `useUrlPredictor()` — shouldn't we do the same here? if the kill-switch is flipped, the standard omnibar path falls back correctly but this one keeps using the native predictor.

also nit: `isNavigate` can read as a Jetpack Navigation call at first glance — `isNavigableUrl` or `isUrlInput` would make the intent clearer.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
```diff
@@ -470,14 +473,21 @@ class InputScreenViewModel @AssistedInject constructor(
         }
     }
 
+    private fun isNavigate(query: String): Boolean =
+        if (queryUrlPredictor.isReady()) {
+            queryUrlPredictor.classify(query) is Decision.Navigate
+        } else {
+            isWebUrl(query)
+        }
```
malmstein: nit: this is the same function as the one in `BrowserTabViewModel`, can we share it? especially if the feature-flag fix above gets added to both.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
```diff
@@ -470,14 +473,21 @@ class InputScreenViewModel @AssistedInject constructor(
...
+
     fun onChatSubmitted(query: String) {
         viewModelScope.launch {
             when {
+                isNavigate(query) -> command.value = Command.SubmitSearch(query)
```
malmstein: when `isNavigate(query)` is true we exit the `when` early, but the analytics block below still runs — `DUCK_CHAT_EXPERIMENTAL_OMNIBAR_PROMPT_SUBMITTED`, `onPromptSubmitted()`, and `setHasUsedChatMode(true)` all fire even though no chat was submitted. URL navigations will count as chat sessions and could trigger the `BOTH_MODES` pixel.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
aitorvs: Fixed — rather than adding the flag check at call sites, I moved it into QueryUrlPredictorImpl.isUrl() so it's encapsulated in one place. Both ViewModels now call queryUrlPredictor.isUrl() which respects useUrlPredictor() internally and falls back to UriString.isWebUrl when the flag is off or the library isn't ready. This also resolved the duplication.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
aitorvs: Resolved as part of the fix above — isNavigate() is gone from both ViewModels and replaced with queryUrlPredictor.isUrl().

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/viewmodel/InputScreenViewModel.kt
aitorvs: Fixed — onChatSubmitted now returns early via return@launch when isUrl() is true, so the analytics block (pixels, onPromptSubmitted, setHasUsedChatMode, checkAndFireBothModesPixel) only runs when a chat prompt is actually submitted. Added tests to cover this and the non-fullscreen URL routing path.

.claude/rules/architecture.md
```diff
@@ -196,6 +215,22 @@ import logcat.logcat   // correct
 
 ---
 
+## Git Workflow
+
```
malmstein, line 219: @aitorvs  should this be in a different rule specifically for git?

aitorvs: I asked claude and it preferred in architecture.md. But we can move it to separate file if we think is best, no strong opinions

PR #7823, opened 2026-02-27
Create Claude rules for ADS and support for current Cursor rules

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202857801505092/task/1213467414889871?focus=true

### Description
Adds Claude rule for ADS

---

> [!NOTE]
> **Low Risk**
> Low risk documentation/CI-only changes; main impact is enforcing symlink parity in PR checks, which could fail CI if rule files aren’t linked correctly.
> 
> **Overview**
> Moves the source of truth for agent/rules docs to `.cursor/rules/*.mdc` and replaces `.claude/rules/*.md` with symlinks pointing at those Cursor rule files (including adding the new ADS rule and a new `wide-events.mdc`).
> 
> Adds a GitHub Action (`check-rules-symlinks.yml`) to enforce `.cursor/rules` ↔ `.claude/rules` parity on PRs, and updates `AGENTS.md` to point contributors at the detailed rule files. Also removes legacy Copilot/MCP configuration (`.github/copilot-instructions.md`, `.mcp.json`) and tweaks Maestro rules frontmatter.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ba1ac027e36dd20eb28d845b5fecc69cf39f9f55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,17 +68,6 @@ Note: `jvm_tests` and `jvm_checks` resolve to `testPlayDebugUnitTest` in `:app` 
 | store | `internal`, `fdroid`, `play` |
 | Build types | `debug`, `release`, `upload` |
 
-### Key Gradle Tasks
-
-```bash
-./gradlew jvm_tests              # Unit tests
-./gradlew jvm_checks             # Code quality checks
-./gradlew lint_check             # Lint
-./gradlew spotlessCheck          # Code formatting
-./gradlew installPlayRelease     # Install Play release build
-./gradlew installInternalRelease # Install internal release build
-```
-
 ---
 
 ## Module Architecture


malmstein:
* **#7823** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

.claude/rules/android-design-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+---
+
+## Package
+
+All components live under `com.duckduckgo.common.ui.view.*`. Use this package in both Kotlin code and XML layouts.
+
+---
+
+## Buttons
```
mikescamell: I wonder if it's better to have all the components in this rule file, or break it down even further and have rule files for each component underneath a design-system folder in .rules?

.claude/rules/android-design-system.md
```diff
@@ -0,0 +1,422 @@
...
+```
+
+---
+
+## List Items
+
+Three main variants:
+
+### OneLineListItem
+48dp height (56dp with large icon).
```
mikescamell: I'm also wondering about wether there's any advantage to adding things like these specs, or simply pointing Claude to the View itself like ln 145 and leaving it at that. As in, the specs are already defined by how we build the View, we just want Claude to use them. WDTY?

.claude/rules/android-design-system.md
```diff
@@ -0,0 +1,422 @@
...
+| `NoAlertDialogDetector` | `AlertDialog` / `MaterialAlertDialog` |
+| `NoBottomSheetDialogDetector` | Raw `BottomSheetDialog` without ADS style |
+| `SkeletonViewBackgroundDetector` | `background_skeleton` drawable on `SkeletonView` |
+| `WrongStyleDetector` | Styles not prefixed `Widget.DuckDuckGo.` |
+
+---
+
+## Icons
+
+- Source icons from the internal DDG Icons repository
```
mikescamell: I wonder if we can link the github here and if it's smart enough to go fetch whatever icon we want 🤔

.claude/rules/android-design-system.md
malmstein: I think we can keep it as is for now, it’s still small enough. In the future we’ll have a compose rule, and maybe there it makes more sense to break it down.

.claude/rules/android-design-system.md
malmstein: Claude wanted them, and seems easy to read. Don’t feel strongly about them.

.claude/rules/android-design-system.md
malmstein: Claude has a suggestion, I think it’s good

.claude/rules/android-design-system.md
mikescamell: My concern is not about readability, my concern is more about how much of this rule file ends up in the context window, as far as I know it's better to keep rule files as small as possible and I'm not sure Claude knowing the sizes might provide any advantage when it's going to just use the View directly. But if this rule file is less about letting Claude know what components we have and how to use them, but more about specs, then I think that would make sense.

mikescamell: Quote from Claude docs:

> Files over 200 lines consume more context and may reduce adherence.

.cursor/rules/android-design-system.mdc
```diff
@@ -0,0 +1,427 @@
+---
+alwaysApply: true
```
mikescamell: ❓ I don't think we want to add this to every Cursor chat session? Same for Maestro rules

.claude/rules/android-design-system.md
```diff
@@ -0,0 +1 @@
+../../.cursor/rules/android-design-system.mdc
```
mikescamell, line 1: 💡 It might be better that we @ mention these rule files rather than just the file path. It's what the docs suggest.

e.g.

@~/.cursor/rules/wide-events.mdc

.cursor/rules/android-design-system.mdc
malmstein: agreed

.claude/rules/android-design-system.md
malmstein: discussed offline, we will use symlinks

mikescamell:
@malmstein looks good but we're missing CLAUDE.md in the root of the project with a link to the AGENTS.md file i.e. 

`@AGENTS.md`

.cursor/rules/wide-events.mdc
```diff
@@ -0,0 +1,145 @@
+# Wide Events
+
+Wide events measure a user's complete journey through a multi-step flow, sending a single event when the journey concludes. Use them when a flow has multiple steps that can succeed or fail independently and you need to understand drop-off and errors across the whole journey.
```
jonathanKingston, line 3: If we're not using front matter, does it make sense to just have a docs/ dir that's linked form AGENTS.md?

That'd mean no symlinks from .cursor/rules and .claude/rules or workflow needed.

malmstein: there’s one for the architecture rule, we’ll leave it as is and improve if needed

PR #7834, opened 2026-02-28
Fix StrictMode disk reads on main thread at app startup

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1198194956794324/task/1213475367843818?focus=true

### Description

Eliminates ~871ms of blocking disk I/O from the main thread at app startup by deferring three categories of StrictMode violations:

**P0.1 — Subscriptions URL construction (~397ms)**
- `InternalSubscriptionsBaseUrlModule`: anonymous object's `subscriptionsBaseUrl` property changed to `by lazy {}`
- `RealSubscriptionsUrlProvider`: all six URL `val`s changed to `by lazy {}` to prevent eager resolution at DI construction time

**P0.2 — Room executor feature-flag read (~272ms)**
- `RoomDatabaseProviderImpl.applyExecutors`: now passes a `LazyExecutor` proxy to `setQueryExecutor`/`setTransactionExecutor` instead of resolving the executor immediately
- `LazyExecutor` defers `lazyDatabaseExecutorProvider.get()` (and the `SharedPreferences`-backed `DatabaseProviderFeature` read) to the first actual DB query, which Room dispatches on a background thread

**P0.3 — PIR SQLCipher native library loading (~202ms)**
- `PirSecureStorageDatabaseFactory`: removed `init {}` block that called `LibraryLoader.loadLibrary("sqlcipher")` at DI construction time; moved the call into `getInnerDatabase()` which runs under a `Mutex` on a background coroutine

### Steps to test this PR

- [ ] Build an internal build and verify no StrictMode violations for `SharedPreferencesSubscriptionsInternalStore`, `RealPirSecureStorageDatabaseFactory`, or `RealDatabaseExecutorProvider` on app launch
- [ ] Verify autofill, subscriptions, and PIR features work normally

### UI changes
No UI changes.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes when and how database executors and SQLCipher library loading are initialized, which can affect runtime behavior under concurrency and feature-flag paths, though the changes are localized and covered by updated tests.
> 
> **Overview**
> Reduces app-startup StrictMode disk reads by deferring several DI-time initializations until first use.
> 
> Room DB setup now always installs `LazyExecutor` proxies for query/transaction executors, delaying `RealDatabaseExecutorProvider` (and its feature-flag SharedPreferences read) until the first DB operation and falling back to a cached thread pool when the delegate resolves to null. PIR secure DB creation no longer loads the SQLCipher native library in an `init` block; it loads on demand in `getInnerDatabase()` on `dispatchers.io()`, while preserving coroutine cancellation.
> 
> Subscriptions URL generation is made lazy (`by lazy`) in both the internal base URL provider and `RealSubscriptionsUrlProvider`, avoiding store reads/URL construction during DI graph creation; related Room executor tests are updated to assert against the new lazy wrappers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 938b9afa9622ca371c477bd314e875a160dc4499. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageDatabaseFactory.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageDatabaseFactory.kt
@@ -18,13 +18,16 @@ package com.duckduckgo.pir.impl.store.secure
 
 import android.content.Context
 import androidx.room.Room
+import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.library.loader.LibraryLoader
 import com.duckduckgo.pir.impl.store.PirDatabase
 import com.squareup.anvil.annotations.ContributesBinding
 import dagger.SingleInstanceIn
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
 import logcat.LogPriority.ERROR
 import logcat.asLog
 import logcat.logcat
@@ -42,6 +45,7 @@ interface PirSecureStorageDatabaseFactory {
 )
 class RealPirSecureStorageDatabaseFactory @Inject constructor(
     private val context: Context,
+    private val dispatchers: DispatcherProvider,
     private val keyProvider: PirSecureStorageKeyProvider,
 ) : PirSecureStorageDatabaseFactory {
     private var _database: PirDatabase? = null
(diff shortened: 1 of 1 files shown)


pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/secure/PirSecureStorageDatabaseFactory.kt
```diff
@@ -73,6 +62,15 @@ class RealPirSecureStorageDatabaseFactory @Inject constructor(
             return _database
         }
 
+        logcat { "PIR-DB: Loading the sqlcipher native library" }
```
karlenDimla, line 69: @CDRussell  actually has a fix for autofill secure storage that moves the sqlcipher lib loading logic out of the init block.

We also move the logic in the same block but with different way to handle the loading.

I am not 100% sure if just moving it here could cause any issues or not. Can we put this change behind a FF?

aitorvs: Both approaches defer the load to first DB access — neither pre-loads at startup. The autofill complexity stems from using the async `loadLibrary` overload, which is callback-based: you need a `CompletableDeferred` to bridge the callback to a coroutine, and a timeout because the raw `Thread` it spawns is outside coroutine control. Feature flag and pixels sit on top of that.

For PIR we can do this cleanly: sync overload + `withContext(dispatchers.io())` achieves the same guarantee with none of that machinery — no callback, no `CompletableDeferred`, no timeout needed since exceptions from the IO thread propagate normally. I've added that `withContext` wrap in this PR.

I don't think a feature flag is needed here — this is a correctness fix, not a behaviour change.

aitorvs:
* **#7834** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7834?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7834?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

PR #7840, opened 2026-03-02
Onboarding Dev Setting

nalcalag:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1213459503063596?focus=true

### Description
Added onboarding dev setting

### Steps to test this PR
- [x] Go to App Settings
- [x] Check new Onboarding Dev Settings is there
- [x] Tap on it
- [x] Check onboarding state matches your current state (skipped, completed or active)
- [x] Play with onboarding dialogs and check the show/hide dialogs feature

### UI changes
Screenshot 2026-02-27 at 17 25 20

---

> [!NOTE]
> **Medium Risk**
> Touches onboarding/CTA persistence by adding write access (delete/insert) to dismissed CTA records and exposing the required onboarding CTA list, which could affect onboarding completion behavior if misused or if call sites change unexpectedly.
> 
> **Overview**
> Adds a new internal-only **Onboarding Dev Settings** screen that lets developers view and manually edit onboarding state (completed/skipped) and toggle individual onboarding CTA/dialog “dismissed” flags.
> 
> To support this, it registers a new `InternalFeaturePlugin`, adds UI/resources and an internal Activity, extends `DismissedCtaDao` with a `delete` API, and exposes `CtaViewModel.requiredDaxOnboardingCtas()` as a `List` for use by the dev screen (with accompanying internal unit tests).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6f0e09688821900958e00307cdc591f8d9879146. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/internal/java/com/duckduckgo/app/dev/settings/onboarding/OnboardingDevSettingsViewModel.kt
+++ b/app/src/internal/java/com/duckduckgo/app/dev/settings/onboarding/OnboardingDevSettingsViewModel.kt
@@ -89,13 +89,13 @@ class OnboardingDevSettingsViewModel @Inject constructor(
                 if (checked) {
                     userStageStore.moveToStage(AppStage.ESTABLISHED)
                     settingsDataStore.hideTips = false
-                    visibleCtaIds().forEach { ctaId ->
+                    _viewState.value.visibleCtaIds.forEach { ctaId ->
                         dismissedCtaDao.insert(DismissedCta(ctaId))
                     }
                 } else {
                     userStageStore.moveToStage(AppStage.DAX_ONBOARDING)
                     settingsDataStore.hideTips = false
-                    visibleCtaIds().forEach { ctaId ->
+                    _viewState.value.visibleCtaIds.forEach { ctaId ->
                         dismissedCtaDao.delete(ctaId)
                     }
                 }


nalcalag:
* **#7840** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

internal-features/internal-features-api/src/main/java/com/duckduckgo/internal/features/api/InternalFeaturePlugin.kt
```diff
@@ -49,5 +49,6 @@ interface InternalFeaturePlugin {
         const val WEB_VIEW_DEV_SETTINGS_PRIO_KEY = 1_000
         const val SAVED_SITES_SETTINGS_PRIO_KEY = 1_100
         const val ATTRIBUTED_METRICS_SETTINGS_PRIO_KEY = 1_200
+        const val ONBOARDING_DEV_SETTINGS_PRIO_KEY = 1_201
```
mikescamell: ❓Why a 1 increment when others are 10/50/100?

app/src/internal/java/com/duckduckgo/app/dev/settings/onboarding/OnboardingDevSettingsActivity.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private fun configureUiEventHandlers() {
+        binding.onboardingCompletedToggle.setOnCheckedChangeListener(completedToggleListener)
+        binding.onboardingSkippedToggle.setOnCheckedChangeListener(skippedToggleListener)
+    }
+
+    private fun observeViewModel() {
+        viewModel.viewState
+            .flowWithLifecycle(lifecycle, Lifecycle.State.CREATED)
```
mikescamell: ❓Why CREATED over STARTED? We tend to use STARTED elsewhere. Do we need to do anything when the Activity might not be visible?

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -126,6 +126,9 @@ class CtaViewModel @Inject constructor(
         }
     }
 
+    suspend fun getRequiredDaxOnboardingCtasForDev(): List<CtaId> =
```
mikescamell: ❓ Adding this function just for the dev settings feels off to me 🤔 If we really want to expose the `requiredDaxOnboardingCtas`, why not expose requiredDaxOnboardingCtas and add a `@VisibleForTesting`?

app/src/main/java/com/duckduckgo/app/cta/db/DismissedCtaDao.kt
```diff
@@ -35,4 +35,7 @@ abstract class DismissedCtaDao {
 
     @Query("select count(1) > 0 from dismissed_cta where ctaId = :ctaId")
     abstract fun exists(ctaId: CtaId): Boolean
+
+    @Query("DELETE FROM dismissed_cta WHERE ctaId = :ctaId")
+    abstract fun delete(ctaId: CtaId)
```
mikescamell, line 41: ❓ Can we some how make this deletion internal to the dev setting screen? We've only added the functionality for dev settings but we're exposing this function for everyone to use.

Maybe we could use a RawQuery instead to delete what we need?

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
```diff
@@ -37,7 +37,6 @@ import javax.inject.Inject
 interface SettingsDataStore {
     var lastExecutedJobId: String?
 
-    @Deprecated(message = "hideTips variable is deprecated and no longer available in onboarding")
```
mikescamell, line 40: ❓ Is this no longer true?

app/src/internal/java/com/duckduckgo/app/dev/settings/onboarding/OnboardingDevSettingsActivity.kt
```diff
@@ -0,0 +1,167 @@
...
+        CtaId.DAX_END,
+        CtaId.DAX_INTRO_PRIVACY_PRO,
+        CtaId.ADD_WIDGET,
+    )
+
+    private val inContextDialogIds = setOf(
+        CtaId.DAX_DIALOG_SERP,
+        CtaId.DAX_DIALOG_TRACKERS_FOUND,
+        CtaId.DAX_FIRE_BUTTON,
+    )
```
mikescamell: 💡 Move these to the ViewModel and make them part of the ViewState. You could then probably ditch `orderedCtaIds` in the ViewModel

app/src/main/java/com/duckduckgo/app/cta/db/DismissedCtaDao.kt
mikescamell: If we want to keep it I would add a comment that it's only used internally for now 👍

mikescamell:
In testing I notice that if I check a dialog that's later in the flow e.g. DAX_INTRO_VISIT_SITE then the Dialogs before that do not get checked e.g. DAX_INTRO, even though they technically must have been completed to see the next dialog. Maybe we could make it so that if you check a dialog later in the flow, the dialogs preceding it are checked too so it's obvious which steps must have been completed?

Also, in my opinion, I think it might be easier for others to understand if we list the dialogs in onboarding flow order rather than their type. We could still add their type as part of the ListItem. But then I guess it's also possible some in-context get triggered at different times?

These could  future improvements 👍

internal-features/internal-features-api/src/main/java/com/duckduckgo/internal/features/api/InternalFeaturePlugin.kt
nalcalag: You are right, I'll increment it by 100

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
nalcalag: Yeah, I started the implementation and added this method in case I needed to include any extra logic specific to dev settings before passing them along. But in the end, I’m just returning the same list, so I’ll go with what you suggested.

app/src/internal/java/com/duckduckgo/app/dev/settings/onboarding/OnboardingDevSettingsActivity.kt
nalcalag: Nope, I just copied and pasted it from the DevSettingsActivity where it was originally created. I’ll change it!

app/src/main/java/com/duckduckgo/app/cta/db/DismissedCtaDao.kt
nalcalag: Good suggestion! but given the current structure, I think it’s easier to keep the DAO method and just add a comment. Using a RawQuery in the dev settings layer wouldn’t really prevent other parts of the code from deleting anything (they would still use a DAO), so we’d end up with the same exposure but with more complexity. I'll go ahead and add the comment 🫡

app/src/main/java/com/duckduckgo/app/settings/db/SettingsDataStore.kt
nalcalag: Nope. I added this deprecation when the onboarding changed and we removed this option. However, once the skip functionality was added back, this deprecation should've been removed.

app/src/internal/java/com/duckduckgo/app/dev/settings/onboarding/OnboardingDevSettingsActivity.kt
nalcalag: I’ve been thinking about this, and I’m going to remove it. It overcomplicates things just for a UI detail that doesn’t matter for testing and prevents dialogs from being retrieved directly from allRequiredOnboardingDialogs. I won't add any ordering logic for the moment.

nalcalag:
@mikescamell I’ve addressed all your suggestions and I also decided to remove the ordering logic for the CtaIds. This way, all CTAs will appear automatically once they’re added to the requiredCtas list in the ctaViewModel, which should scale better.
I know there are several improvements we could make, but for an MVP, especially when testing scenarios that aren’t the "happy path", adding extra logic (like toggling dialogs that would normally appear earlier in a real onboarding flow) would introduce unnecessary complexity and hardcoded behavior in Dev Settings. 
As you know, there isn’t a strict “real” order in the current onboarding, so even if some dialogs were originally intended to appear before others, it doesn’t really matter for testing purposes. It might look a bit messy and unordered, but it should be functional enough as an MVP (I think 🙂).

mikescamell:
Screenshot_20260305_083605

@nalcalag I noticed when it's just the DAX_FIRE_DIALOG unchecked and i load a website we don't see the in-context fire dialog appear, it only happens if I also uncheck DAX_DIALOG_TRACKERS_FOUND. Not a blocker but if the trackers dialog does in fact need to be unchecked to see the fire dialog, then this is the kind of behaviour I'd like to see in future in the settings 👍

app/src/internal/java/com/duckduckgo/app/dev/settings/onboarding/OnboardingDevSettingsViewModel.kt
```diff
@@ -0,0 +1,149 @@
...
+                if (checked) {
+                    userStageStore.moveToStage(AppStage.ESTABLISHED)
+                    settingsDataStore.hideTips = false
+                    visibleCtaIds().forEach { ctaId ->
+                        dismissedCtaDao.insert(DismissedCta(ctaId))
+                    }
+                } else {
+                    userStageStore.moveToStage(AppStage.DAX_ONBOARDING)
+                    settingsDataStore.hideTips = false
+                    visibleCtaIds().forEach { ctaId ->
```
mikescamell: 💡 We could just use the ViewState `visibleCtas` property instead so we're not having to fetch and filter CTAs everytime

PR #7842, opened 2026-03-02
Add invalid page context diagnostic pixels

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1198194956794324/task/1213492015096541

### Description

The ratio of `m_aichat_contextual_page_context_placeholder_tapped` to `m_aichat_contextual_page_context_manually_attached_native` is ~30%, meaning ~70% of manual attachment attempts fail `isContextValid()`. Without per-reason breakdowns we can't tell whether extraction is failing, validity criteria are too strict, or certain page types systematically fail.

This PR adds three diagnostic pixels that fire when `addPageContext()` finds the page context invalid, one per failure reason:

- `m_aichat_contextual_page_context_invalid_empty_c/d` — context string was empty
- `m_aichat_contextual_page_context_invalid_no_title_c/d` — extracted title was missing or blank
- `m_aichat_contextual_page_context_invalid_no_content_c/d` — extracted content body was missing or blank

The pixels only fire from the user-initiated path (`addPageContext()`), not from auto-attach or `replacePrompt()`. This is done via a `reportInvalidPixels: Boolean = false` parameter on `isContextValid()`.

### Steps to test this PR

_Diagnostic pixels_
- [ ] Open the contextual sheet on a page where context extraction fails (e.g. a blank/loading page) and tap the placeholder — verify `m_aichat_contextual_page_context_invalid_empty` fires
- [ ] Open on a page that produces context without a title — verify `m_aichat_contextual_page_context_invalid_no_title` fires
- [ ] Open on a page that produces context without body content — verify `m_aichat_contextual_page_context_invalid_no_content` fires
- [ ] Open on a valid page and tap the placeholder — verify none of the invalid pixels fire and `m_aichat_contextual_page_context_manually_attached_native` fires as before

### UI changes
No UI changes.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to additional pixel firing and a small, gated tweak to `isContextValid` used during manual context attachment. Main risk is incorrect/over-firing telemetry if the new pixel names or call sites are miswired.
> 
> **Overview**
> Adds three new **diagnostic contextual pixels** to break down why manual page-context attachment fails: empty context, missing title, and missing content.
> 
> Updates `DuckChatContextualViewModel.addPageContext()` to call `isContextValid(..., reportInvalidPixels = true)` so these pixels only fire on the user-initiated attach path, and extends `DuckChatPixels`/`RealDuckChatPixels` (plus pixel name definitions and param-removal mapping) to support firing count+daily variants. Unit tests were expanded to assert the new pixels fire for each invalid case.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b1af41c98f8eae20833cc3f285a33183a17bbdc7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualViewModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualViewModel.kt
@@ -346,7 +346,7 @@ class DuckChatContextualViewModel @Inject constructor(
         logcat { "Duck.ai Contextual: addPageContext" }
         duckChatPixels.reportContextualPlaceholderContextTapped()
         viewModelScope.launch {
-            val isContextValid = isContextValid(updatedPageContext)
+            val isContextValid = isContextValid(updatedPageContext, reportInvalidPixels = true)
             if (isContextValid) {
                 duckChatPixels.reportContextualPageContextManuallyAttachedNative()
                 _viewState.update { current ->
@@ -360,12 +360,18 @@ class DuckChatContextualViewModel @Inject constructor(
         }
     }
 
-    private fun isContextValid(pageContext: String): Boolean {
-        if (pageContext.isEmpty()) return false
+    private fun isContextValid(pageContext: String, reportInvalidPixels: Boolean = false): Boolean {
+        if (pageContext.isEmpty()) {
+            if (reportInvalidPixels) duckChatPixels.reportContextualPageContextInvalidEmpty()
+            return false
+        }
         val json = JSONObject(pageContext)
         val title = json.optString("title").takeIf { it.isNotBlank() }
         val content = json.optString("content").takeIf { it.isNotBlank() }
-
+        if (reportInvalidPixels) {
+            if (title == null) duckChatPixels.reportContextualPageContextInvalidNoTitle()
+            if (content == null) duckChatPixels.reportContextualPageContextInvalidNoContent()
+        }
         return title != null && content != null
     }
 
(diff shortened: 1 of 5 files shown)


duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualViewModel.kt
```diff
@@ -346,7 +346,7 @@ class DuckChatContextualViewModel @Inject constructor(
         logcat { "Duck.ai Contextual: addPageContext" }
         duckChatPixels.reportContextualPlaceholderContextTapped()
         viewModelScope.launch {
-            val isContextValid = isContextValid(updatedPageContext)
+            val isContextValid = isContextValid(updatedPageContext, reportInvalidPixels = true)
```
malmstein, line 349: this should be sent from `onPageContextReceived`, which is when the FE sends us the context. Doing this only in `addPageContext` doesn’t cover all cases, like when the user taps on “Attach Page Context” via FE

aitorvs: The placement is intentional — these diagnostic pixels are meant to correlate with `reportContextualPlaceholderContextTapped` and `reportContextualPageContextManuallyAttachedNative`, which are both fired exclusively from `addPageContext` (the native attachment path). Keeping the invalid-context pixels here gives us a clean valid/invalid ratio for that same path.

The FE-triggered path (`onPageContextReceived`) is a separate flow and doesn't fire those pixels, so adding the diagnostic there would mix two different flows and break the correlation. If we want to track FE-triggered context failures separately, that would be a different pixel scoped to `onPageContextReceived` — but that's a separate concern from what this PR is measuring.

PR #7849, opened 2026-03-02
Single tab burning: Wire up data-clearing with single tab burning

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1213087856649980?focus=true

### Description

Implements the single tab burning flow end-to-end and simplifies the history repository.

#### Single tab burn wiring (DataClearing):
 - `clearSingleTabData()` is no longer a no-op — it now retrieves visited sites for the tab, clears site-specific WebStorage via `WebStorageCompat.deleteBrowsingDataForSite()`, removes history entries for the tab, cleans up the visited-sites store, and deletes the tab itself.
 - Added `clearDataForSpecificDomains()` to `ClearDataAction` for domain-scoped clearing.

#### History repository simplification:
 - Replaced the manual `cachedHistoryEntries` + `fetchAndCacheHistoryEntries()` pattern with a Room Flow (`getHistoryEntriesWithVisitsFlow()`), so history updates are reactive via the database and the cache is no longer needed.
 - Removed redundant `fetchAndCacheHistoryEntries()` calls after every mutation.

#### History deletion behavior change:
 - `deleteHistoryForTab()` now deletes entire history entries that were visited by the tab, including visits from other tabs to those same URLs. Previously it only removed the tab's visits and orphaned entries. This ensures that burning a tab fully removes all traces of the URLs it visited from history (aligns with iOS behavior).

###  Steps to test this PR

_History deletion_
 - [ ] Open a tab and navigate to cnn.com
 - [ ] On the same tab, navigate to msn.com
 - [ ] Open a new tab and start typing cnn.com
 - [ ] Verify the cnn.com history entry is visible
 - [ ] Tap on the cnn.com history entry
 - [ ] Burn that single tab
 - [ ] Open a new tab and start typing cnn.com
 - [ ] Verify the history suggestion doesn't appear for cnn.com

 _Cookie deletion_
 - [ ] Open a tab and navigate to a site where you can login
 - [ ] Authenticate to the website
 - [ ] Open a new tab and navigate to the same domain
 - [ ] Burn the single tab
 - [ ] On the previously authenticated website, try navigating to some link
 - [ ] Notice the user is no longer logged in

 ### UI changes

  No UI changes — this is backend/data layer wiring only.

---

> [!NOTE]
> **High Risk**
> Adds new single-tab data clearing flow that deletes per-domain WebStorage and changes history deletion semantics for burned tabs, which can impact user data retention and correctness. Also refactors history retrieval to a reactive Room `Flow`, affecting history UI update behavior and performance.
> 
> **Overview**
> Implements end-to-end *single-tab burning*: `DataClearing.clearSingleTabData` now gathers visited eTLD+1 domains, clears per-site WebView storage via new `ClearDataAction.clearDataForSpecificDomains`, removes tab history, clears the tab’s visited-sites records, and deletes the tab; callers now receive a `ClearDataResult` (success / feature-not-supported / error).
> 
> Adds WebView capability gating and user feedback for single-tab clear failures (new `FireDialog` events + snackbar strings), and ensures full clears also wipe `TabVisitedSitesRepository`.
> 
> Simplifies history storage by removing manual caching in `RealHistoryRepository` in favor of a Room `Flow`, and **changes `deleteHistoryForTab` to delete entire history entries visited by the tab** (even if other tabs also visited them).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0b827efdf83743585938a014dcab7a2f58d00394. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/view/ClearPersonalDataAction.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/view/ClearPersonalDataAction.kt
@@ -40,12 +40,12 @@ import com.duckduckgo.history.api.NavigationHistory
 import com.duckduckgo.savedsites.api.SavedSitesRepository
 import com.duckduckgo.site.permissions.api.SitePermissionsManager
 import com.duckduckgo.sync.api.DeviceSyncState
-import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.withContext
 import logcat.LogPriority.INFO
 import logcat.LogPriority.WARN
 import logcat.logcat
+import kotlin.coroutines.cancellation.CancellationException
 import kotlin.coroutines.resume
 
 sealed class ClearDataResult {


0nko:
* **#7849** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7849?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7849?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/global/view/ClearPersonalDataAction.kt
```diff
@@ -196,6 +211,26 @@ class ClearPersonalDataAction(
         }
     }
 
+    @SuppressLint("RequiresFeature")
```
CDRussell, line 227: Why suppressing this? Is there a runtime check elsewhere guarding this? otherwise it will crash on older `WebView` versions.

Even if there's a check elsewhere, i would recommend checking ability to call `WebStorageCompat.deleteBrowsingDataForSite` in here close to where it's executed rather than relying on something else having done the right thing first.

0nko: @CDRussell Thanks for the review!

> Why suppressing this? Is there a runtime check elsewhere guarding this? otherwise it will crash on older WebView versions.

There is. The only place where this gets triggered is the `Delete this tab` in the [new single tab burn dialog](https://github.com/duckduckgo/Android/blob/60afbafc64f0f80d8cbe55e52d3c35a8dfb2dc37/app/src/main/java/com/duckduckgo/app/global/view/SingleTabFireDialog.kt#L229), which uses the [runtime feature check](https://github.com/duckduckgo/Android/blob/60afbafc64f0f80d8cbe55e52d3c35a8dfb2dc37/app/src/main/java/com/duckduckgo/app/global/view/SingleTabFireDialogViewModel.kt#L110).

> Even if there's a check elsewhere, i would recommend checking ability to call WebStorageCompat.deleteBrowsingDataForSite in here close to where it's executed rather than relying on something else having done the right thing first.

We could do that but getting here without the check already done would be the result of a logical error.  I feel like silently failing seems worse than crashing. Or perhaps we could bubble up the failure and display some kind of an error message... 🤔

0nko:
@CDRussell I added a feature check and error handling to the `WebStorageCompat.deleteBrowsingDataForSite` call site. Ready for another round.

app/src/main/java/com/duckduckgo/app/global/view/ClearPersonalDataAction.kt
CDRussell: What about bubbling up the @requires annotation on this function instead of suppressing lint?

Callers of this function should be aware of this call and the annotation would signal that the caller needs to take care since this function won't check.

CDRussell: Never mind, we cross-posted (i.e., i didn't see your reply before posting this).

Adding the capability check inside the function is great, thanks for doing it that way

0nko:
@cursor push 932e21482e

PR #7872, opened 2026-03-04
Improve GlobalActivityStarter with launcher overloads and defensive fixes

aitorvs:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1198194956794324/task/1213548840330684?focus=true

### Description

Three fixes to `GlobalActivityStarter`:

**1. `startForResult()` overloads**

The previous pattern for launching an activity for result was `startIntent() + launcher.launch(intent)`. `startIntent()` returns `Intent?` — passing `null` to `launcher.launch()` crashes at runtime with no clear error. The new `startForResult(context, params, launcher)` overloads resolve this and make the intent clearer at the call site.

Before:
```kotlin
val intent = globalActivityStarter.startIntent(this, SomeActivityParams)
someLauncher.launch(intent) // crashes if intent is null
```

After:
```kotlin
globalActivityStarter.startForResult(this, SomeActivityParams, someLauncher)
```

**2. Automatic `FLAG_ACTIVITY_NEW_TASK` for non-Activity contexts**

`startIntent()` and `start()` now automatically add `FLAG_ACTIVITY_NEW_TASK` when `context !is Activity`. Callers in Services, broadcast receivers, and JS message handlers no longer need to add the flag manually after calling `startIntent()`.

Before:
```kotlin
val intent = globalActivityStarter.startIntent(context, DuckChatNativeSettingsNoParams)
intent?.flags = Intent.FLAG_ACTIVITY_NEW_TASK
context.startActivity(intent)
```

After:
```kotlin
globalActivityStarter.start(context, DuckChatNativeSettingsNoParams)
```

**3. Defensive logging**

- `logcat(ERROR)` is emitted before `IllegalArgumentException` when no mapper is found for a params type, making registration bugs visible in logcat before the crash.
- `logcat(WARN)` is emitted when multiple mappers claim the same params type (previously the first match silently won with no diagnostic output).

Updated callers: `RestoreSubscriptionActivity` (migrated to `startForResult`), `OpenNativeSettingsHandler` (removed manual flag), `BookmarksActivity`/`BookmarksViewModel` (removed unused `LaunchSyncSettings` + `syncActivityLauncher` — the result callback only re-ran promotion eligibility, which already happens via other paths).

### Steps to test this PR

- [ ] Launch an activity for result using `startForResult()` — confirm it launches correctly and the result callback fires
- [ ] Restore a subscription via **Settings → Subscription → Restore** — confirm the restore flow launches and completes without a crash
- [ ] Open a Duck.ai chat, trigger the native settings from the SERP settings JS handler — confirm it opens without a crash and without needing `FLAG_ACTIVITY_NEW_TASK` manually
- [ ] Confirm no regression in Bookmarks — open bookmarks, verify sync promotion behaviour is unchanged
- [ ] Run `./gradlew :navigation-impl:testDebugUnitTest` — all tests should pass

---

> [!NOTE]
> **Medium Risk**
> Touches shared navigation infrastructure used across modules; intent flagging and mapper selection changes could affect activity launches if edge cases exist, though covered by new unit tests.
> 
> **Overview**
> Improves `GlobalActivityStarter` to be safer and more ergonomic: adds `startForResult(...)` overloads for `ActivityResultLauncher`, and centralizes intent construction with clearer error logging when no mapper is found.
> 
> `start()`/`startIntent()` now automatically apply `FLAG_ACTIVITY_NEW_TASK` for non-`Activity` contexts, and log a warning when multiple mappers match the same params (first match still wins). Call sites are updated to use the new APIs (e.g. subscriptions restore flow and SERP native-settings handler), and bookmarks removes a now-unused sync-settings launch path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8d5ce61f8d40f71f14aee8efd707812c118c023a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/navigation/navigation-impl/src/main/java/com/duckduckgo/navigation/impl/GlobalActivityStarterImpl.kt
+++ b/navigation/navigation-impl/src/main/java/com/duckduckgo/navigation/impl/GlobalActivityStarterImpl.kt
@@ -85,12 +85,18 @@ class GlobalActivityStarterImpl @Inject constructor(
         context: Context,
         deeplinkActivityParams: DeeplinkActivityParams,
     ): Intent? {
-        val activityParams: ActivityParams? = activityMappers.firstNotNullOfOrNull {
-            it.map(deeplinkActivityParams)
+        return buildIntent(context, deeplinkActivityParams)?.apply {
+            if (context !is Activity) addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
         }
-        return activityParams?.let { startIntent(context, it) }
     }
 
+    private fun buildIntent(context: Context, deeplinkActivityParams: DeeplinkActivityParams): Intent? {
+        val activityParams: ActivityParams? = activityMappers.firstNotNullOfOrNull { it.map(deeplinkActivityParams) }
+        return activityParams?.let { buildIntent(context, it) }
+    }
+
+    // context is used only to construct the Intent — FLAG_ACTIVITY_NEW_TASK is never added,
+    // regardless of context type. Callers should pass the Activity that owns the launcher.
     override fun startForResult(context: Context, params: ActivityParams, launcher: ActivityResultLauncher<Intent>) {
         val intent = buildIntent(context, params)
         if (intent == null) {
@@ -101,8 +107,7 @@ class GlobalActivityStarterImpl @Inject constructor(
     }
 
     override fun startForResult(context: Context, deeplinkActivityParams: DeeplinkActivityParams, launcher: ActivityResultLauncher<Intent>) {
-        val activityParams: ActivityParams? = activityMappers.firstNotNullOfOrNull { it.map(deeplinkActivityParams) }
-        val intent = activityParams?.let { buildIntent(context, it) }
+        val intent = buildIntent(context, deeplinkActivityParams)
         if (intent == null) {
             logcat(ERROR) { "No activity found for params $deeplinkActivityParams" }
             throw IllegalArgumentException("Activity for params $deeplinkActivityParams not found")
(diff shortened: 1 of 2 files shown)


navigation/navigation-impl/src/main/java/com/duckduckgo/navigation/impl/GlobalActivityStarterImpl.kt
```diff
@@ -74,10 +88,26 @@ class GlobalActivityStarterImpl @Inject constructor(
...
+        if (intent == null) {
+            logcat(ERROR) { "No activity found for params $params" }
+            throw IllegalArgumentException("Activity for params $params not found")
+        }
+        launcher.launch(intent)
+    }
 
-        return activityParams?.let {
-            startIntent(context, it)
+    override fun startForResult(context: Context, deeplinkActivityParams: DeeplinkActivityParams, launcher: ActivityResultLauncher<Intent>) {
```
malmstein, line 109: nit: this reimplements what startIntent(context, deeplinkActivityParams) already does, minus the flag. if we ever change the deeplink resolution logic we'd need to update it in two places. consider delegating: call buildIntent(context, activityParams) after resolving the params via a shared private method.

aitorvs: Exactly — extracted a private `buildIntent(context, DeeplinkActivityParams)` overload that both `startIntent(deeplink)` and `startForResult(deeplink)` delegate to. Also added the missing flag tests for the deeplink `startIntent` path while I was at it.

PR #7880, opened 2026-03-05
Create the new RequestBlocklist interface and implementation

catalinradoiu:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211724162604201/task/1213530900303284?focus=true

### Description

This PR introduces a new Request Blocklist interface and implementation  for the privacy configuration system. The implementation provides the ability to determine if specific network requests are in the blocklist based on configurable rules that match request URLs and document domains.

The feature includes:
- A new `RequestBlocklist` API interface with a `containedInBlocklist` method to check if requests should be blocked
- `RealRequestBlocklist` implementation that loads blocklist rules from privacy configuration settings and evaluates them against incoming requests

### Steps to test this PR
Unit tests pass. This PR only introduces the new interface with its implementation and new data models. The integration in the request interceptor will be added in a follow-up PR.

### UI changes
No UI changes. 

---

> [!NOTE]
> **Medium Risk**
> Introduces new parsing and matching logic for remotely-supplied blocklist rules (URL/domain matching), which could cause unintended blocking or performance overhead when enabled, though it is gated by a feature toggle and well-tested.
> 
> **Overview**
> **Adds a new `RequestBlocklist` capability** to the privacy-config module, exposing `containedInBlocklist(documentUrl, requestUrl)` for checking requests against blocklist rules.
> 
> Implements `RealRequestBlocklist`, which listens for privacy-config downloads, loads `requestBlocklist` feature settings into an in-memory map keyed by top private domain, validates/filters rule entries, and matches requests via a simple `*`-wildcard-to-regex conversion plus document-domain constraints (including `<all>`). Adds extensive unit tests covering enablement, parsing/validation, wildcard semantics, domain matching, and reload behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9a9368e3a67660aef3d438f49c3bd14cd67be73d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
+++ b/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
@@ -43,9 +43,13 @@ class RealRequestBlocklist @Inject constructor(
     private val dispatchers: DispatcherProvider,
     @IsMainProcess private val isMainProcess: Boolean,
     @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
+    moshi: Moshi,
 ) : RequestBlocklist, PrivacyConfigCallbackPlugin {
 
     private val blockedRequests = ConcurrentHashMap<String, List<BlocklistRuleEntity>>()
+
+    private val blockListSettingsJsonAdapter: JsonAdapter<RequestBlocklistSettings> =
+        moshi.adapter(RequestBlocklistSettings::class.java)
 
     init {
         if (isMainProcess) {
@@ -71,7 +75,7 @@ class RealRequestBlocklist @Inject constructor(
         val normalizedUrl = httpUrl.toString()
 
         return rules.any { rule ->
-            ruleMatches(normalizedUrl, rule.rule) && domainMatches(documentUrl, rule.domains)
+            rule.rule.containsMatchIn(normalizedUrl) && domainMatches(documentUrl, rule.domains)
         }
     }
 
@@ -81,10 +85,7 @@ class RealRequestBlocklist @Inject constructor(
 
             requestBlocklistFeature.self().getSettings()?.let { settingsJson ->
                 runCatching {
-                    val moshi = Moshi.Builder().build()
-                    val jsonAdapter: JsonAdapter<RequestBlocklistSettings> =
-                        moshi.adapter(RequestBlocklistSettings::class.java)
-                    val settings = jsonAdapter.fromJson(settingsJson)
+                    val settings = blockListSettingsJsonAdapter.fromJson(settingsJson)
 
                     settings?.blockedRequests?.entries?.forEach { entry ->
                         val domain = entry.key
@@ -105,19 +106,6 @@ class RealRequestBlocklist @Inject constructor(
             blockedRequests.putAll(newBlockedRequests)
         }
     }
-
-    private fun ruleMatches(
-        normalizedUrl: String,
-        rule: String,
-    ): Boolean = buildString {
-        for (char in rule) {
-            if (char == '*') {
-                append("[^/]*")
-            } else {
-                append(Regex.escape(char.toString()))
-            }
-        }
-    }.toRegex().containsMatchIn(normalizedUrl)
 
     private fun domainMatches(
         documentUrl: String?,
--- a/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
+++ b/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
@@ -16,16 +16,12 @@
 
 package com.duckduckgo.privacy.config.impl.features.requestblocklist
 
-data class RequestBlocklistSettings(
-    val blockedRequests: Map<String, BlockedRequestEntry>?,
-)
-
 data class BlockedRequestEntry(
     val rules: List<Map<String, @JvmSuppressWildcards Any>>?,
 )
 
 class BlocklistRuleEntity(
-    val rule: String,
+    val rule: Regex,
     val domains: List<String>,
     val reason: String?,
 ) {
@@ -39,9 +35,21 @@ class BlocklistRuleEntity(
         @Suppress("UNCHECKED_CAST")
         fun fromJson(map: Map<String, Any>): BlocklistRuleEntity? {
             if (map.keys.any { it !in KNOWN_PROPERTIES }) return null
-            val rule = map[PROPERTY_RULE] as? String ?: return null
+
+            val ruleString = map[PROPERTY_RULE] as? String ?: return null
             val domains = map[PROPERTY_DOMAINS] as? List<String> ?: return null
             val reason = map[PROPERTY_REASON] as? String
+
+            val rule = buildString {
+                for (char in ruleString) {
+                    if (char == '*') {
+                        append("[^/]*")
+                    } else {
+                        append(Regex.escape(char.toString()))
+                    }
+                }
+            }.toRegex()
+
             return BlocklistRuleEntity(rule = rule, domains = domains, reason = reason)
         }
     }
--- a/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistFeature.kt
+++ b/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistFeature.kt
@@ -29,3 +29,7 @@ interface RequestBlocklistFeature {
     @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
     fun self(): Toggle
 }
+
+data class RequestBlocklistSettings(
+    val blockedRequests: Map<String, BlockedRequestEntry>?,
+)
--- a/privacy-config/privacy-config-impl/src/test/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklistTest.kt
+++ b/privacy-config/privacy-config-impl/src/test/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklistTest.kt
@@ -19,6 +19,7 @@ package com.duckduckgo.privacy.config.impl.features.requestblocklist
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.duckduckgo.common.test.CoroutineTestRule
 import com.duckduckgo.feature.toggles.api.Toggle
+import com.squareup.moshi.Moshi
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Rule
@@ -34,6 +35,8 @@ class RealRequestBlocklistTest {
     var coroutineTestRule = CoroutineTestRule()
 
     private val mockToggle: Toggle = mock()
+
+    private val moshi = Moshi.Builder().build()
     private val mockFeature: RequestBlocklistFeature = mock<RequestBlocklistFeature>().apply {
         whenever(self()).thenReturn(mockToggle)
     }
@@ -50,6 +53,7 @@ class RealRequestBlocklistTest {
             dispatchers = coroutineTestRule.testDispatcherProvider,
             isMainProcess = isMainProcess,
             appCoroutineScope = coroutineTestRule.testScope,
+            moshi = moshi,
         )
     }
 
@@ -242,6 +246,7 @@ class RealRequestBlocklistTest {
             dispatchers = coroutineTestRule.testDispatcherProvider,
             isMainProcess = false,
             appCoroutineScope = coroutineTestRule.testScope,
+            moshi = moshi,
         )
 
         assertFalse(testee.containedInBlocklist("https://example.com", "https://testing.com/image.jpg"))


catalinradoiu:
* **#7888** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7880** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,131 @@
...
+                            }
+                        }
+                    }
+                }.onFailure {
+                    logcat { "RequestBlocklist: Failed to parse settings: ${it.message}" }
+                }
+            }
+
+            blockedRequests.clear()
+            blockedRequests.putAll(newBlockedRequests)
```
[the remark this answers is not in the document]
catalinradoiu: The blocklist is processed at app launch or afer the privacy config is downloaded, this shouldn’t be a real issue.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,131 @@
...
+            println(blockedRequests)
```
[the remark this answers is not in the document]
catalinradoiu: Removed

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+        rule: String,
+    ): Boolean = buildString {
+        for (char in rule) {
+            if (char == '*') {
+                append("[^/]*")
+            } else {
+                append(Regex.escape(char.toString()))
+            }
+        }
+    }.toRegex().containsMatchIn(normalizedUrl)
```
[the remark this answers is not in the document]
catalinradoiu: I am seeing the same in the TdsClient. Could it be a real performance issue?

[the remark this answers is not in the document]
CrisBarreiro: pre-compiling should offer better performance, but I don't think I've measured the actual impact in the past. The downside is that it'll take longer to initialize the dataset, but again, haven't measure to what extent 

we could defer the decision to performance testing

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+        requestUrl: String,
+    ): Boolean {
+        if (!requestBlocklistFeature.self().isEnabled()) return false
+
+        val httpUrl = requestUrl.toHttpUrlOrNull() ?: return false
+        val requestDomain = httpUrl.topPrivateDomain() ?: return false
+
+        val rules = blockedRequests[requestDomain] ?: return false
+
+        val normalizedUrl = httpUrl.toString()
```
CrisBarreiro, line 75: Why do we need this `toString` again? If the URL is malformed, `requestUrl.toHttpUrlOrNull()` should've returned null, which already short-circuits

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    override fun containedInBlocklist(
+        documentUrl: String,
+        requestUrl: String,
+    ): Boolean {
+        if (!requestBlocklistFeature.self().isEnabled()) return false
+
+        val httpUrl = requestUrl.toHttpUrlOrNull() ?: return false
+        val requestDomain = httpUrl.topPrivateDomain() ?: return false
```
CrisBarreiro, line 71: Just thinking out loud, but if in the future we want to better orchestrate all types of request interception, we might want to avoid each implementation performing the same URL manipulations, as a way to improve performance

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    private fun loadToMemory() {
+        appCoroutineScope.launch(dispatchers.io()) {
+            val newBlockedRequests = ConcurrentHashMap<String, List<BlocklistRuleEntity>>()
+
+            requestBlocklistFeature.self().getSettings()?.let { settingsJson ->
+                runCatching {
+                    val moshi = Moshi.Builder().build()
```
CrisBarreiro: nit: Can we inject, or at least avoid creation every time we call `loadToMemory`? The latter also applies to the next line

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+                    val jsonAdapter: JsonAdapter<RequestBlocklistSettings> =
+                        moshi.adapter(RequestBlocklistSettings::class.java)
+                    val settings = jsonAdapter.fromJson(settingsJson)
+
+                    settings?.blockedRequests?.entries?.forEach { entry ->
+                        val domain = entry.key
+                        val topPrivateDomain = "https://$domain".toHttpUrlOrNull()?.topPrivateDomain()
+                        if (topPrivateDomain != null && topPrivateDomain == domain) {
```
CrisBarreiro, line 93: Why do we need this?

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    private fun ruleMatches(
+        normalizedUrl: String,
+        rule: String,
+    ): Boolean = buildString {
+        for (char in rule) {
+            if (char == '*') {
+                append("[^/]*")
```
CrisBarreiro: can we do the manipulation at the rule parsing level instead?

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+                        moshi.adapter(RequestBlocklistSettings::class.java)
+                    val settings = jsonAdapter.fromJson(settingsJson)
+
+                    settings?.blockedRequests?.entries?.forEach { entry ->
+                        val domain = entry.key
+                        val topPrivateDomain = "https://$domain".toHttpUrlOrNull()?.topPrivateDomain()
+                        if (topPrivateDomain != null && topPrivateDomain == domain) {
+                            val validRules = entry.value.rules?.mapNotNull { BlocklistRuleEntity.fromJson(it) } ?: emptyList()
+                            if (validRules.isNotEmpty()) {
+                                newBlockedRequests[domain] = validRules
```
CrisBarreiro, line 96: At this point, we have already validated the domain, so could we use `com.duckduckgo.app.browser.Domain`? That will allow us to use a more performant implementation of `sameOrSubdomain` later on. Otherwise, we would convert this back to Uri, so we can get the domain from it

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        val httpUrl = requestUrl.toHttpUrlOrNull() ?: return false
+        val requestDomain = httpUrl.topPrivateDomain() ?: return false
+
+        val rules = blockedRequests[requestDomain] ?: return false
+
+        val normalizedUrl = httpUrl.toString()
+
+        return rules.any { rule ->
+            ruleMatches(normalizedUrl, rule.rule) && domainMatches(documentUrl, rule.domains)
```
CrisBarreiro: Similar to what I mentioned in https://github.com/duckduckgo/Android/pull/7880/changes#r2895185205, we could create a new implementation for `SameOrSubdomain` that accepts HttpUrl and Domain, so that we avoid converting back and forth

privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/RequestBlocklist.kt
```diff
@@ -14,29 +14,16 @@
...
-     * Unable to classify startup type reliably.
+     * Checks whether a request is contained in the blocklist rules
+     * from the Privacy Configuration.
+     *
+     * @param documentUrl the URL of the page making the request
+     * @param requestUrl the URL being requested
+     * @return true if the request matches a blocklist rule
      */
-    UNKNOWN,
+    fun containedInBlocklist(documentUrl: String, requestUrl: String): Boolean
```
CrisBarreiro, line 28: I know this was accepted on the proposal, but looking at the implementation details, and some notes on performance I added below, I'm wondering whether the function should use `HttpUrl` instead. That way we could get pretty much the same performance benefits as with `Uri`, while still keeping this as a Kotlin module

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.privacy.config.impl.features.requestblocklist
+
+data class RequestBlocklistSettings(
```
CrisBarreiro: nit: I'd move this to the `RequestBlocklistFeature`, for clarity

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
```diff
@@ -0,0 +1,48 @@
...
+    val blockedRequests: Map<String, BlockedRequestEntry>?,
+)
+
+data class BlockedRequestEntry(
+    val rules: List<Map<String, @JvmSuppressWildcards Any>>?,
```
CrisBarreiro, line 20: Shouldn't this be a String (or a Regex)?

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
```diff
@@ -0,0 +1,48 @@
...
+) {
+    companion object {
+        private const val PROPERTY_RULE = "rule"
+        private const val PROPERTY_DOMAINS = "domains"
+        private const val PROPERTY_REASON = "reason"
+
+        private val KNOWN_PROPERTIES = setOf(PROPERTY_RULE, PROPERTY_DOMAINS, PROPERTY_REASON)
+
+        @Suppress("UNCHECKED_CAST")
+        fun fromJson(map: Map<String, Any>): BlocklistRuleEntity? {
```
CrisBarreiro, line 36: What do we need this manual parsing for?

Also, I think failing on unknown properties is too restrictive and will break older versions if we add a new field. If we really need this, I think we should at least fire a pixel to alert on the situation, but I think we can just ignore unknown fields

privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/RequestBlocklist.kt
catalinradoiu: HttpUrl is from the okhttp module, which the API module is not using. We cannot use it unless we add the dependency there. It’s the same issue that was with Uri, which required Android dependencies in a pure Kotlin module.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
catalinradoiu: The requirement was to skip rules with unknown properties or missing required ones. I discussed with Dave as well and this is how he suggested to handle this. I can add a pixel there so that we know about this situation. But the reason for this was to avoid situations where the rule is accepted but behaves differently on older versions (I guess in scenarios where we would add a new property which should be considered as well when checking if the request is in the blocklist).

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
catalinradoiu: I agree with that. But we need to take into account that we will need to add other dependencies to the API module. As in this case we’re using strings for the urls, but we’re converting them here because we cannot pass them as HttpUrl/ Uri becase the API module doesn’t have Android/ OkHttp dependencies.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
catalinradoiu: The `requestUrl.toHttpUrlOrNull()` is used to normalise the request URL. The requirement is that the request URL should be normalised first, so protocol and domain parts are always considered to be lowercase. That function is used to handle that. However, the regex matchins requires a string, that’s why I am calling that again.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
catalinradoiu: Entries must be for eTLD+1 domains, other entries will be ignored. I need to make sure that the entry is in that format, because we don’t support subdomains in the entries. `example.com` is a valid entry, while `some.subdomain.example.com` should be skipped.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
catalinradoiu: Rules have the following format, so we cannot use String or Regex, because the parsing would fail. 

```
"rules": [
                    {
                        "rule": “testing/*.jpg",
                        "domains": ["example.com"],
                        "reason": ["EXAMPLE"]
                    }
                ]
```

This `List<Map<String, @JvmSuppressWildcards Any>>?` is used here because this is a Moshi/JSON deserialization model. It represents the raw JSON structure before it's parsed into typed objects.

Each rule in the JSON is an object with varying keys (rule, domains, reason), so Moshi deserializes each one as a `Map<String, Any>`. The typed conversion happens later in BlocklistRuleEntity.fromJson(map), which validates the keys and casts values to their expected types.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
catalinradoiu: I won’t really help us, becasue the domain in this case if the one from the request. We take it like this:
```
val httpUrl = requestUrl.toHttpUrlOrNull() ?: return false
val requestDomain = httpUrl.topPrivateDomain() ?: return false
```
`sameOrSubdomain` is used to check if the document url is the same or a subdomain of the domans that the rule needs to apply for.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
catalinradoiu: The `HttpUrl` is used for the request url, to extract the top domain so that we can find the match in rules map. The domain match is used for the documentUrl, which is passed as String. The solution would be to pass it as an Uri, but that would require a change in the API module, as we cannot import it there.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
catalinradoiu: Updated. We are now manipulating this when we create the object from the JSON entry.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
catalinradoiu: Updated.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
catalinradoiu: Updated.

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RequestBlocklistModels.kt
```diff
@@ -0,0 +1,56 @@
...
+            if (map.keys.any { it !in KNOWN_PROPERTIES }) return null
+
+            val ruleString = map[PROPERTY_RULE] as? String ?: return null
+            val domains = map[PROPERTY_DOMAINS] as? List<String> ?: return null
```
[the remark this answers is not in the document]
catalinradoiu: This is not a real concern, the same would happen if we would use a data model for parsing instead of using the `fromJson` method.

PR #7892, opened 2026-03-06
Android: Update dependencies 06/03/2026

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1157893581871903/task/1213556662786513

### Description

Runs `refreshVersions` to discover all available dependency updates and applies safe, low-risk updates to stable releases only.

**Libraries updated:**
- `androidx.appcompat` 1.7.0 → 1.7.1
- `androidx.fragment` 1.8.6 → 1.8.9
- `androidx.security-crypto` 1.1.0-alpha06 → 1.1.0
- `androidx.webkit` 1.13.0 → 1.14.0
- `com.google.zxing.core` 3.5.3 → 3.5.4
- `android.tools.desugar_jdk_libs` 2.1.2 → 2.1.5
- `robolectric` 4.14 → 4.16.1
- `org.json` 20240303 → 20241224

**Deferred (require Kotlin 2.x or larger migrations):** Kotlin, AGP, Room, Dagger/Anvil, Coil, Compose compiler, kotlinx.collections.immutable, logcat, harmony/harmony-crypto, turbine

**Deferred (kept at current versions intentionally):** rxjava2.rxjava, rxjava2.rxandroid

### Steps to test this PR

- E2E Nightly Full Suite workflow: https://github.com/duckduckgo/Android/actions/runs/22780893076

### UI changes

None

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes runtime and test dependencies (e.g., `androidx.webkit`, `androidx.security-crypto`, `robolectric`), which can introduce behavioral or build/test differences despite being minor/stable bumps.
> 
> **Overview**
> Updates `versions.properties` with several safe/stable dependency bumps, including `androidx.appcompat`, `androidx.fragment`, `androidx.webkit`, `androidx.security-crypto` (to `1.1.0`), `robolectric`, `org.json`, `zxing`, and `desugar_jdk_libs`.
> 
> Adds a new dependency-update playbook (`.cursor/rules/dependency-updates.mdc`, surfaced via `AGENTS.md` and `.claude/rules/dependency-updates.md`) covering `refreshVersions` usage, preserving the `####` header in `versions.properties`, Kotlin 1.9.x vs 2.x compatibility checks, and PR/testing conventions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47c9b71bc242c30c3f788e1a401e6bc64a60ed4e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.cursor/rules/dependency-updates.mdc
+++ b/.cursor/rules/dependency-updates.mdc
@@ -70,7 +70,7 @@ These libraries look like safe minor/patch bumps but are not — always check Ko
 
 ### Generally safe to update
 - AndroidX libraries (`androidx.*`) — backward compatible, Java/Kotlin mixed
-- Pure Java libraries (`rxjava`, `rxandroid`, `zxing`, `org.json`, `robolectric`, `desugar_jdk_libs`)
+- Pure Java libraries (`zxing`, `org.json`, `robolectric`, `desugar_jdk_libs`)
 
 ### Needs Kotlin version check
 - Any Kotlin-first library (Square, Cash App, JetBrains, etc.)
@@ -84,6 +84,7 @@ These libraries look like safe minor/patch bumps but are not — always check Ko
 - Coil
 - Compose compiler
 - `kotlinx.collections.immutable`
+- RxJava (`rxjava2.rxjava`, `rxjava2.rxandroid`) — kept at current versions intentionally
 
 ---
 


malmstein:
@GerardPaligot now!

marcosholgado:
There are still a lot of random comments, can we also look to update wekbit?

.cursor/rules/dependency-updates.mdc
```diff
@@ -0,0 +1,117 @@
+---
+alwaysApply: true
```
aitorvs: One issue: `alwaysApply: true` means these 117 lines get injected into every AI context window for the project, including feature work, bug fixes, code reviews — any session where dependencies aren't being touched. That's noise most of the time.

The other rules in AGENTS.md are consulted contextually ("read before working in those areas"). This one should follow the same pattern. Suggest either `alwaysApply: false` or `globs: ["versions.properties"]` so it only activates when that file is actually being edited. The AGENTS.md entry is enough to make it discoverable.

malmstein: @aitorvs  you’re right, cleaning that up.

versions.properties
```diff
@@ -170,18 +170,18 @@ version.com.jakewharton.retrofit..retrofit2-kotlin-coroutines-adapter=0.9.2
...
-version.robolectric=4.14
+version.robolectric=4.16.1
 
 version.com.facebook.shimmer..shimmer=0.5.0
 
-version.rxjava2.rxandroid=2.0.2
+version.rxjava2.rxandroid=2.1.1
 
-version.rxjava2.rxjava=2.2.9
+version.rxjava2.rxjava=2.2.21
```
mikescamell: ❓ Our [docs](https://app.asana.com/1/137249556945/project/1202552961248957/task/1199899332680683?focus=true) state we should not update RxJava dependencies, has that changed?

.cursor/rules/dependency-updates.mdc
```diff
@@ -0,0 +1,117 @@
+---
+description: "How to safely update Android library dependencies using refreshVersions"
+---
+# Dependency Update Rules
+
+## Overview
+
+Dependencies are managed via `versions.properties` using the `refreshVersions` Gradle plugin (v0.60.5).
+The only file that should change in a dependency update PR is `versions.properties`.
```
GerardPaligot, line 9: Question: This is only true if there are no breaking changes. How should the LLM behave if it needs to update a dependency that also requires updating the codebase?

.cursor/rules/dependency-updates.mdc
```diff
@@ -0,0 +1,117 @@
...
+
+> "Library X update from A → B requires Kotlin 2.x (currently on 1.9.24).
+> Do you want to:
+> 1. Skip this update for now
+> 2. Include it as a separate task to upgrade Kotlin first"
+
+The engineer may decide to batch multiple blocked libraries into a Kotlin upgrade task,
+or simply defer them. Either way, document the decision in the PR and Asana task.
+
+### Known libraries requiring Kotlin 2.x (as of March 2026)
```
GerardPaligot, line 56: Question: This list will be hard for us to maintain. Should we ask the LLM to update the list whenever a new dependency is identified, or should we create an Asana task and have the LLM update that task when a new library is found?

versions.properties
GerardPaligot: I didn't have the context for this documentation task, but my two cents is that hotfix versions are often related to bug fixes or security issues, so it's better to update them instead of ignoring them.

malmstein: Good catch @mikescamell , I’ll revert that and add to the rule

.cursor/rules/dependency-updates.mdc
malmstein: The LLM should not change any file outside of the versions.properties, which is why this rule is in place. If there are breaking changes, it should be the engineer who decides what to do.

.cursor/rules/dependency-updates.mdc
malmstein: I don’t want to create processes around this, more than what we already have. This rule is to make it easy to start library updates via LLM, but it’s not a place to add team processes, that’s what the Asana task is for. 

It’s the first time we do it this way, so I’m hoping for changes as soon as other engineers start doing it.

PR #7914, opened 2026-03-09
Check if document url is in the exceptions list before blocking the request

catalinradoiu:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211724162604201/task/1213570605076697?focus=true

### Description

Added exception handling to the request blocklist feature to skip blocking when document domains match configured exceptions, allowing fine-grained control over request blocking behavior.

Also optimized WebView request interception by caching URL string conversions to avoid repeated `toString()` calls on the same URI objects. 

### Steps to test this PR
CI passes. The request blocklist change will be covered by reference tests as well. 

### UI changes
No UI changes

---

> [!NOTE]
> **Medium Risk**
> Adjusts tracker/request blocking decisions based on remotely-configured exception domains, which could reduce protection or change site behavior if domain matching/parsing is wrong. Scope is contained to request-blocklist evaluation and adds tests to cover the new exception paths.
> 
> **Overview**
> Adds support for `requestBlocklist` *feature exceptions*: `RealRequestBlocklist.containedInBlocklist` now extracts the document domain and returns false early when it matches an exception (including subdomains), with exceptions loaded/cached from the feature toggle.
> 
> Refactors request-blocklist rule parsing to treat `<all>` as a dedicated flag (`applyToAllDomains`) and store rule domains as `Domain` objects, and updates `WebViewRequestInterceptor` to reuse cached `toString()` values for `documentUri`/`request.url` when evaluating filters and reloads.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fb3c70fae088355d66ab1cb08b4c4705ca291e25. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
+++ b/privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
@@ -49,6 +49,8 @@ class RealRequestBlocklist @Inject constructor(
 ) : RequestBlocklist, PrivacyConfigCallbackPlugin {
 
     private val blockedRequests = ConcurrentHashMap<String, List<BlocklistRuleEntity>>()
+
+    @Volatile
     private var exceptions = listOf<Domain>()
 
     private val blockListSettingsJsonAdapter: JsonAdapter<RequestBlocklistSettings> =


catalinradoiu:
* **#7915** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7914** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7914?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7914?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

privacy-config/privacy-config-impl/src/main/java/com/duckduckgo/privacy/config/impl/features/requestblocklist/RealRequestBlocklist.kt
```diff
@@ -104,6 +106,13 @@ class RealRequestBlocklist @Inject constructor(
 
             blockedRequests.clear()
             blockedRequests.putAll(newBlockedRequests)
+            exceptions = requestBlocklistFeature.self().getExceptions().map { it.domain }
```
CrisBarreiro: can we wrap this as Domain, so that we can use a more performant option of sameOrSubdomain?

catalinradoiu: If we do this then we need to validate that the documentUrl is a valid domain, because we don’t have an option of sameOrSubdomain that takes a String and a Domain as params. 
The documentUrl is a String and represent a full url. We could create a Domain from it, but we need to validate it first, and convert it to Uri so that we can properly extract the domain.
I could do that before checking if it’s an exception. That way if I extract the domain I can update the other usages of sameOrSubdomain in the RealRequestBlocklist and always use Domains, instead of ending up converting the domainUrl to Uri for each sameOrSubdomain call. However, with that change, it would mean that we always convert the domainUrl to Uri, for each URL that we check. This would impact the performance more than using it as it is now, because now the Uri conversion is done only if there are exceptions defined in the exceptions list (we might not have that at the beginning but we could end up with a few of them) or if request matches one of the rules. 
This could not be optimal initially but could be the better solution long term, once we start adding more domains to the exceptions list.

PR #7932, opened 2026-03-10
Improve remote config patches documentation

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1213611390820529?focus=true

### Description

- Rewrites the README for conciseness and readability; adds a Quick Start section
- Adds build variant warning (patches only work in `internal` builds)
- Documents patch ordering, conflict resolution, and failure behaviour
- Adds "Writing a patch file" section with examples for overriding feature state, sub-features, and adding new sub-features
- Clarifies when the full patch (hash removal + version bump) is needed vs. a simple state override
- Adds "Verifying a patch was applied" section covering Gradle build output and logcat
- Restructures usage: local dev via `local.properties` vs. committed patches via `-Pconfig_patches` flag
- Adds Maestro test file layout example and end-to-end build+test workflow
- Adds GitHub Actions workflow example using `checkout-and-assemble` `gradle_flags` input
- Adds PR review workflow: how to share patch files in a PR description so reviewers can verify feature flag states step by step
- Adds patch examples section with full before/after config for common scenarios

### Steps to test this PR

- [ ] Read through the updated README and verify accuracy
- [ ] Verify the Maestro and GitHub Actions examples match existing workflow patterns in the repo

### DIFF
--- a/privacy-config/privacy-config-internal/README.md
+++ b/privacy-config/privacy-config-internal/README.md
@@ -6,7 +6,9 @@ Internal-only module with dev tools for testing the remote privacy config integr
 
 ## Local remote config patches
 
-Patches are JSON files in [JSON Patch format](https://jsonpatch.com/) applied at runtime after the remote config is fetched, overriding values from the server.
+Patches are JSON files in [JSON Patch format](https://jsonpatch.com/) applied at runtime after the remote config is fetched, overriding values from the server. Multiple patches are applied in the order listed — if two patches modify the same path, the last one wins; a failing patch is skipped without aborting the rest.
+
+> **Important:** Each patch file name must be unique, regardless of which directory it lives in.
 
 ### Quick start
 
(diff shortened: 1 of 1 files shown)


privacy-config/privacy-config-internal/README.md
```diff
@@ -36,38 +99,131 @@ Optionally, you can also adjust the remote config version and remove the feature
...
+Copied config patch to build assets: .../disable_something.json -> .../build/generated/assets/configPatches/disable_something.json
+```
+
+If a patch file path is wrong you'll see a warning instead:
+
+```
+Warning: Config patch file not found: /path/to/disable_something.json
+```
+
+**Logcat** — when the app fetches the remote config, the patch interceptor logs each step under the `DevPrivacyConfigPatchApiInterceptor` tag:
```
GerardPaligot: Suggestion: If you want to add the log when a patch can't be applied in local like you did for Maestro:

```
Failed to apply patch disable_something.json: <reason>
```

This can appear often if you don't know you need to use "add" instead of "replace" for a FF which isn't present in the remote file.

```
Failed to apply patch disable_something.json: Missing field "newSubFeature"
```

privacy-config/privacy-config-internal/README.md
```diff
@@ -36,38 +99,131 @@ Optionally, you can also adjust the remote config version and remove the feature
...
-    "value": "999999999999999"
+    "value": "90000000000001"
   }
 ]
 ```
 
-Make sure to provide a version number higher than the one currently used in production, so the app will recognize it as a new version of the remote config.
+Use a version number higher than production (e.g. `90000000000001`) so the app treats it as a newer config and applies it. Without this, a cached config from a previous run may not be overridden.
+
+Each time you modify the patch without clearing the app cache, increment the version number. For example, after changing `state` back to `disabled`, bump the version from `90000000000001` to `90000000000002`:
```
GerardPaligot: Question: What do you think about having a section dedicated to caching that explains how patches and the local privacy configuration file (previously cached on the device) interact? Updating the version isn't always enough when you are loading the FF too soon.

GerardPaligot (commented):
The new versioning value and concrete examples with sub features are really nice improvements!

privacy-config/privacy-config-internal/README.md
LukasPaczos: Good idea!

privacy-config/privacy-config-internal/README.md
LukasPaczos: I'll an additional note about the asynchronous behavior of config but I also don't want to overload config patching documentation with details on how the config itself works - that's a separate topic and keeping these docs here could get quickly out-of-date.

LukasPaczos:
Thanks for taking a look @GerardPaligot! I gave this another pass, incorporated your ideas, and re-did the structure to keep it more concise at the top and add more examples at the bottom.

This is ready for another round.

GerardPaligot (approved):
LGTM!

PR #7945, opened 2026-03-11
Add Duck.ai dev setting

joshliebe:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1213310129933666?focus=true

### Description

- Adds a Duck.ai dev setting to override the Duck.ai URL
- Centralizes all uses of the “duck.ai” domain
- Changes the default `DUCK_CHAT_WEB_LINK` to "https://duck.ai/chat?duckai=5” (the same URL that’s used in the config)

### Steps to test this PR

- [ ] Go to Settings > Developer Settings > Custom Duck.ai URL
- [ ] Enter a custom URL
- [ ] Tap “Save"
- [ ] Verify that the app is restarted
- [ ] Go to Duck.ai
- [ ] Verify that the custom URL is used

### UI changes
| Developer Settings  | Custom Duck.ai URL |
| ------ | ----- |
Screenshot_20260312_212642|Screenshot_20260312_213453

---

> [!NOTE]
> **Medium Risk**
> Touches multiple privacy/cookie-clearing and JS messaging allowlist paths by replacing hardcoded `duck.ai` checks with an injected host provider, so a misconfiguration could break Duck.ai navigation, data clearing, or messaging on that domain.
> 
> **Overview**
> Adds a `DuckAiHostProvider` abstraction and rewires Duck.ai-related logic to use it instead of hardcoded `duck.ai` (cookies/third‑party cookie exceptions, IndexedDB/site data clearing exclusions, site permissions microphone recovery allowlist, and multiple JS messaging `allowedDomains` lists).
> 
> Introduces an *internal/dev-only* `duckchat-internal` module and Developer Settings UI to override the Duck.ai URL, persisting it in a small data store and replacing the default provider via DI; the DuckChat entry URL is also updated to the direct `https://duck.ai/chat?duckai=5` form.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4a3a6247060d3918879f598952bf32aa305fb9b8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsManagerTest.kt
+++ b/site-permissions/site-permissions-impl/src/test/java/com/duckduckgo/site/permissions/impl/SitePermissionsManagerTest.kt
@@ -58,21 +58,24 @@ class SitePermissionsManagerTest {
         MicrophoneSitePermissionsDomainRecoveryFeature::class.java,
     )
 
-    private val testee = SitePermissionsManagerImpl(
-        mockPackageManager,
-        mockLocationManager,
-        mockSitePermissionsRepository,
-        coroutineRule.testDispatcherProvider,
-        mockContext,
-        fakeMicrophoneSitePermissionsDomainRecoveryFeature,
-        mockDuckAiHostProvider,
-    )
+    private val testee by lazy {
+        SitePermissionsManagerImpl(
+            mockPackageManager,
+            mockLocationManager,
+            mockSitePermissionsRepository,
+            coroutineRule.testDispatcherProvider,
+            mockContext,
+            fakeMicrophoneSitePermissionsDomainRecoveryFeature,
+            mockDuckAiHostProvider,
+        )
+    }
 
     private val url = "https://domain.com/whatever"
     private val tabId = "tabId"
 
     @Before
     fun before() {
+        whenever(mockDuckAiHostProvider.getHost()).thenReturn("duck.ai")
         whenever(mockPackageManager.hasSystemFeature(PackageManager.FEATURE_CAMERA_ANY)).thenReturn(true)
         fakeMicrophoneSitePermissionsDomainRecoveryFeature.self().setRawStoredState(Toggle.State(false))
     }
@@ -227,5 +230,24 @@ class SitePermissionsManagerTest {
     @Test
     fun whenAndroidPermissionNotSupportedThenGetPermissionsQueryResponseReturnsDenied() = runTest {
         assertEquals(SitePermissionQueryResponse.Denied, testee.getPermissionsQueryResponse(url, tabId, "unsupported"))
+    }
+
+    @Test
+    fun whenRecoveryEnabledAndDuckAiAudioGrantedAndAndroidPermissionDeniedThenAudioNotAutoAccepted() = runTest {
+        val duckAiUrl = "https://duck.ai/chat"
+        fakeMicrophoneSitePermissionsDomainRecoveryFeature.self().setRawStoredState(Toggle.State(true))
+        val resources = arrayOf(PermissionRequest.RESOURCE_AUDIO_CAPTURE)
+        whenever(mockSitePermissionsRepository.isDomainAllowedToAsk(duckAiUrl, PermissionRequest.RESOURCE_AUDIO_CAPTURE)).thenReturn(true)
+        whenever(mockSitePermissionsRepository.isDomainGranted(duckAiUrl, tabId, PermissionRequest.RESOURCE_AUDIO_CAPTURE)).thenReturn(true)
+        whenever(mockContext.checkPermission(any(), any(), any())).thenReturn(PackageManager.PERMISSION_DENIED)
+
+        val permissionRequest: PermissionRequest = mock()
+        whenever(permissionRequest.origin).thenReturn(duckAiUrl.toUri())
+        whenever(permissionRequest.resources).thenReturn(resources)
+
+        val permissions = testee.getSitePermissions(tabId, permissionRequest)
+        assertEquals(1, permissions.userHandled.size)
+        assertEquals(PermissionRequest.RESOURCE_AUDIO_CAPTURE, permissions.userHandled.first())
+        assertEquals(0, permissions.autoAccept.size)
     }
 }


joshliebe:
* **#7945** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7945?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7945?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/RealDuckChat.kt
```diff
@@ -639,6 +647,13 @@ class RealDuckChat @Inject constructor(
             }
     }
 
+    private fun getDuckChatLink(): String {
+        if (appBuildConfig.isInternalBuild()) {
+            return duckChatLink.toUri().buildUpon().authority(duckAiHostProvider.getHost()).build().toString()
```
aitorvs: This check defeats the purpose of `DuckAiHostProvider`. The provider already encapsulates the build-type decision — the prod impl always returns `duck.ai`, the internal impl returns the override (or falls back to `duck.ai`). `RealDuckChat` should not need to know which build it is in.

This can be simplified to:
```kotlin
private fun getDuckChatLink(): String {
    return duckChatLink.toUri().buildUpon().authority(duckAiHostProvider.getHost()).build().toString()
}
```

On prod, `getHost()` returns `duck.ai`, so this is a no-op. On internal with a custom host set, it substitutes. `AppBuildConfig` can then be removed from the constructor entirely.

duckchat/duckchat-internal/src/main/java/com/duckduckgo/duckchat/internal/InternalDuckAiHostProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val dataStore: DuckAiInternalSettingsDataStore,
+    @AppCoroutineScope private val appScope: CoroutineScope,
+    private val dispatcherProvider: DispatcherProvider,
+) : DuckAiHostProvider {
+
+    @Volatile
+    private var cachedHost: String? = null
+
+    init {
+        appScope.launch(dispatcherProvider.io()) {
```
aitorvs: This async load creates a race: `getHost()` can be called before the coroutine completes, silently returning `duck.ai` instead of the stored custom host. This matters because several consumers (JS messaging `allowedDomains`, clearing domains, etc.) capture the host value at construction time.

Since this is internal-only and the value is a single string, a synchronous read is simpler and safer:
```kotlin
init {
    cachedHost = runBlocking { dataStore.customUrl }?.extractHost()
}
```

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/RealDuckChat.kt
```diff
@@ -497,7 +501,11 @@ class RealDuckChat @Inject constructor(
     }
 
     override fun canHandleOnAiWebView(url: String): Boolean {
-        return runCatching { HOST_DUCK_AI == url.toHttpUrl().topPrivateDomain() || url == REVOKE_URL }.getOrElse { false }
+        return runCatching {
+            val httpUrl = url.toHttpUrl()
+            val duckAiHost = duckAiHostProvider.getHost()
+            httpUrl.topPrivateDomain() == duckAiHost || httpUrl.host == duckAiHost || url == REVOKE_URL
+        }.getOrElse { false }
```
aitorvs: The original code used `topPrivateDomain() == "duck.ai"` to match any subdomain of `duck.ai` (e.g. `chat.duck.ai`). With `DuckAiHostProvider`, `getHost()` now returns the exact intended host — either `duck.ai` in prod or whatever the override is in internal builds.

That means `httpUrl.topPrivateDomain() == duckAiHost` is only true when `duckAiHost` happens to equal the eTLD+1 (i.e. production `duck.ai`). For a custom subdomain host like `staging.duck.ai`, `topPrivateDomain()` returns `duck.ai` ≠ `staging.duck.ai`, so the first condition is always false — effectively dead code for internal overrides. The `httpUrl.host == duckAiHost` check already handles both cases correctly on its own.

Was the intent here to keep the broader "any subdomain of duck.ai" matching for production? If yes, `topPrivateDomain()` should be the only check and `getHost()` shouldn't be used here. If the intent is exact-host matching (which seems right given the new abstraction), then `topPrivateDomain()` can be dropped:
```kotlin
httpUrl.host == duckAiHost || url == REVOKE_URL
```

duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/DuckAiHostProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.duckchat.api
+
+interface DuckAiHostProvider {
+
+    /** Returns either a custom override or `duck.ai`. */
+    fun getHost(): String = DEFAULT_HOST
+
```
aitorvs, line 23: 6 call sites across the codebase manually construct `"https://${duckAiHostProvider.getHost()}"`. Consider adding a `getBaseUrl()` default method here to centralise that:
```kotlin
fun getBaseUrl(): String = "https://${getHost()}"
```
This keeps URL construction in one place and makes callers simpler.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/RealDuckChat.kt
joshliebe: Updated the default URL to the one used in the config, so this isn’t needed anymore

duckchat/duckchat-internal/src/main/java/com/duckduckgo/duckchat/internal/InternalDuckAiHostProvider.kt
joshliebe: Fair enough, I’ve made it synchronous

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/RealDuckChat.kt
joshliebe: Removed the redundant check

duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/DuckAiHostProvider.kt
joshliebe: I’m not going to muddy the interface with that change, I will leave it up to the call sites to use the domain as necessary

PR #7957, opened 2026-03-12
Single tab burn: Refactor the Duck chat sync mechanism

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1213622659790663?focus=true

### Description

This PR refactors the Duck chat sync mechanism to reuse the existing Sync API for the `PATCH` operation, which is used to delete specific Duck chats.

### Steps to test this PR

Testing requires sync to be enabled on 2 separate devices. Follow the steps here on both before proceeding with the testing: https://app.asana.com/1/137249556945/project/72649045549333/task/1212472824864986?focus=true

 - [x] Both `improvedDataClearingOptions` and `singleTabBurnDialog` flags must be enabled manually
 - [x] Device 1: Start a new chat and pin it
 - [x] Device 1: Open a different, non-chat tab
 - [x] Device 2: Open Duck.ai and verify the chat from device 1 was synced and is present
 - [x] Device 2: Tap on the chat to open it
 - [x] Device 2: Tap on the Fire button and burn the single tab
 - [x] Device 2: Verify the chat was successfully deleted and is gone from the list
 - [x] Device 1: Open Duck.ai and verify the chat was successfully deleted and is gone from the list
 

---

> [!NOTE]
> **Medium Risk**
> Refactors core sync engine and API surface by removing the dedicated patchable-deletion pathway and adding a new syncable type with per-type operation support; mistakes could prevent chat deletions (or other sync types) from syncing correctly.
> 
> **Overview**
> Duck AI chat “delete specific chats” sync is refactored to reuse the existing `SyncChangesRequest`/`SyncableDataProvider`/`SyncableDataPersister` PATCH flow instead of the separate `PatchableDataManager` + `SyncPatchRequest/Response` pathway.
> 
> The sync API/engine is updated accordingly: `PatchableDataManager` and patch models are removed, `DeletableDataManager.getType` is renamed to `getDeletableType`, and `SyncableType` gains `DUCK_AI_CHATS` with a new `SyncableTypeConfig` so the engine only performs supported GET/PATCH operations per type. Networking is adjusted to add `PATCH /sync/ai_chats` (`patchChats`, optional `since`) and `SyncApiClient.patch` now routes DUCK_AI_CHATS to that endpoint while other types keep using `PATCH /sync/data`.
> 
> `SyncDateProvider` is centralized into `common-utils` and imports are updated across autofill/settings/duckchat; tests are updated/added to reflect the new sync flow and chat patch response parsing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c3d8728c2339031ec7419f7d9d799b78abc75059. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/test/java/com/duckduckgo/duckchat/impl/sync/DuckChatSyncDataManagerTest.kt
+++ b/duckchat/duckchat-impl/src/test/java/com/duckduckgo/duckchat/impl/sync/DuckChatSyncDataManagerTest.kt
@@ -220,7 +220,7 @@ class DuckChatSyncDataManagerTest {
     @Test
     fun whenOnPatchErrorThenPendingQueueIsKept() = runTest {
         val error = SyncErrorResponse(
-            type = DeletableType.DUCK_AI_CHATS,
+            type = SyncableType.DUCK_AI_CHATS,
             featureSyncError = FeatureSyncError.INVALID_REQUEST,
         )
 


sync/sync-api/src/main/java/com/duckduckgo/sync/api/engine/Models.kt
```diff
@@ -126,10 +104,15 @@ interface SyncFeatureType {
...
-enum class SyncableType(override val field: String) : SyncFeatureType {
+enum class SyncableType(
+    override val field: String,
+    val endpoint: String = "data",
+    val supportsGet: Boolean = true,
+) : SyncFeatureType {
     BOOKMARKS("bookmarks"),
     CREDENTIALS("credentials"),
     SETTINGS("settings"),
+    DUCK_AI_CHATS("ai_chats", "ai_chats", false),
```
CDRussell: nit: Named params would be useful here

```suggestion
    DUCK_AI_CHATS(field = "ai_chats", endpoint = "ai_chats", supportsGet = false),
```

sync/sync-api/src/main/java/com/duckduckgo/sync/api/engine/Models.kt
```diff
@@ -126,10 +104,15 @@ interface SyncFeatureType {
 /**
  * Features that support bidirectional sync (PATCH/GET operations).
  */
-enum class SyncableType(override val field: String) : SyncFeatureType {
+enum class SyncableType(
+    override val field: String,
+    val endpoint: String = "data",
+    val supportsGet: Boolean = true,
```
CDRussell: supportsGet: Boolean is awkward; why only something for GET, what does that imply about POST/DELETE/PATCH ?
should this be more like a list of supported HTTP methods? e.g.
DUCK_AI_CHATS
 - supported methods = [PATCH, DELETE] // or maybe drop DELETE for now but it can be added when we ditch DeletableType in future

BOOKMARKS
- supported methods = [GET, PATCH]

CREDENTIALS
- supported methods = [GET, PATCH]

feels like supportsGet: Boolean is too limited/awkward. And it also feels like which HTTP methods are supported and which endpoints to use should not be part of the public API and are all impl details.

CDRussell: Further on the above, i don't think it's clear what `field` and `endpoint` mean. 
- `endpoint` is only used for patching. that's not clear from the variable name which makes it sound like it would also `GET` from there too
- there's a confusing asymmetry between `GET` and `PATCH`, where with `GET` we don't define the endpoint in the type itself but put that logic instead in `SyncApiClient` to handle. whereas with `PATCH` the endpoint is defined in the type itself using that `endpoint` param

Someone reading `SyncableType` would probably assume endpoint is the canonical URL path for all operations on this type, which it isn't. Any way we we can either move the `GET` endpoint into the type too, or remove endpoint from the type and handle both in `SyncApiClient` consistently? 

e.g., `SyncApiClient.patch` does a `when` on the type just like `SyncApiClient.get` has to do. in the when is where you provide the endpoint to use for `PATCH`.

What about if we keep the types in the public API clean of those details, and have an internal config for each type that contains the supported HTTP methods and endpoints?

```
private enum class SyncableTypeConfig(                                                                                                                                                                                                     
      val type: SyncableType,
      val supportedOperations: Set<SyncOperation>,                                                                                                                                                                                           
  ) {
      BOOKMARKS(SyncableType.BOOKMARKS, setOf(                                                                                                                                                                                               
          SyncOperation(GET, endpoint="data"),
          SyncOperation(PATCH, endpoint="data"),
      )),
      CREDENTIALS(SyncableType.CREDENTIALS, setOf(
          SyncOperation(GET, endpoint="data"),
          SyncOperation(PATCH, endpoint="data"),
      )),
      SETTINGS(SyncableType.SETTINGS, setOf(
          SyncOperation(GET, endpoint="data"),
          SyncOperation(PATCH, endpoint="data"),
      )),
      DUCK_AI_CHATS(SyncableType.DUCK_AI_CHATS, setOf(
          SyncOperation(PATCH, endpoint="ai_chats"),
      )),
      ;

**Comment by 0nko** [comment]
Thanks @CDRussell! I've updated the code and added [the helper config class](https://github.com/duckduckgo/Android/pull/7957/commits/f6f2e4e62f9e4594f9f7fcdd2c850451368bb45b), as we discussed last week.

Ready for another round!

File: duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/sync/SyncDateProvider.kt [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.duckchat.impl.sync
+
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+
+class SyncDateProvider {
```
malmstein, line 23: can this be in common module instead?

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/engine/SyncApiClient.kt
```diff
@@ -102,6 +107,7 @@ class AppSyncApiClient @Inject constructor(
             BOOKMARKS -> syncApi.getBookmarks(token, since)
             CREDENTIALS -> syncApi.getCredentials(token, since)
             SETTINGS -> syncApi.getSettings(token, since)
+            DUCK_AI_CHATS -> error("DUCK_AI_CHATS does not support GET")
```
malmstein: you sure you want to crash here?

duckchat/duckchat-impl/src/test/java/com/duckduckgo/duckchat/impl/sync/DuckChatSyncDataManagerTest.kt
```diff
@@ -150,55 +157,60 @@ class DuckChatSyncDataManagerTest {
...
-        val result = testee.getPatches()
+        val result = testee.getChanges()
 
-        assertNull(result)
+        assertTrue(result.isEmpty())
     }
 
     @Test
-    fun whenGetPatchesAndNoPendingDeletionsThenReturnsNull() = runTest {
+    fun whenGetChangesAndNoPendingDeletionsThenReturnsEmpty() = runTest {
```
malmstein, line 179: The test enables the feature flag but doesn't set up `isAIChatHistoryEnabled`. Add `whenever(duckChatFeatureRepository.isAIChatHistoryEnabled()).thenReturn(true) ` to make the test prove what it claims.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/sync/DuckChatSyncDataManager.kt
```diff
@@ -24,34 +24,67 @@ import com.duckduckgo.common.utils.checkMainThread
...
     private val duckChatFeatureRepository: DuckChatFeatureRepository,
     private val dispatchers: DispatcherProvider,
     private val appBuildConfig: AppBuildConfig,
     private val duckChatFeature: DuckChatFeature,
     @AppCoroutineScope private val appCoroutineScope: CoroutineScope,
-) : DeletableDataManager, PatchableDataManager {
+) : DeletableDataManager, SyncableDataProvider, SyncableDataPersister {
 
-    override fun getType(): DeletableType = DeletableType.DUCK_AI_CHATS
+    override fun getDeletableType(): DeletableType = DeletableType.DUCK_AI_CHATS
```
malmstein, line 67: this is only used for error recording, not sure we can do much now but it would be good to revisit this @CDRussell

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/engine/SyncableTypeConfig.kt
```diff
@@ -0,0 +1,47 @@
...
+
+package com.duckduckgo.sync.impl.engine
+
+import com.duckduckgo.sync.api.engine.SyncableType
+
+enum class SyncHttpMethod {
+    GET,
+    PATCH,
+}
+
```
malmstein, line 25: this is nice!

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/engine/RealSyncEngine.kt
```diff
@@ -150,30 +147,35 @@ class RealSyncEngine @Inject constructor(
...
-            if (changes.isEmpty()) {
-                logcat(INFO) { "Sync-Engine: no changes to sync for $changes, asking for remote changes" }
-                getRemoteChanges(changes, TIMESTAMP)
-            } else {
-                logcat(INFO) { "Sync-Engine: $changes changes to update $changes" }
-                patchLocalChanges(changes, TIMESTAMP)
+        regularSyncChanges
+            .forEach { changes ->
+                if (changes.isEmpty()) {
+                    if (changes.type.supports(SyncHttpMethod.GET)) {
```
malmstein, line 153: 👍

malmstein (approved):
Great work @0nko , thanks for taking in the feedback.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/engine/SyncApiClient.kt
0nko: Yes, this already guarded [here](https://github.com/duckduckgo/Android/pull/7957/changes/BASE..67826335dcb6b77c1d0560931ecc53a0516ffa86#diff-26dd8362632e81435e883ec00423002a422bc8bcd1215eaae757555bc57e7d6aR153). I added it because `when` expects a result. But I guess we could return an error, like Cursor suggested.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/sync/SyncDateProvider.kt
marcosholgado: Why common module if it's coupled to sync? Either the name doesn't make sense or the place doesn't make sense.

marcosholgado: Also there's already a CurrentTimeProvider

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/sync/DuckChatSyncDataManager.kt
```diff
@@ -86,34 +120,33 @@ class DuckChatSyncDataManager @Inject constructor(
...
-        logcat { "DuckChat-Sync: patch successful for ${response.entryIds.size} entries" }
-        appCoroutineScope.launch(dispatchers.io()) {
-            duckChatSyncRepository.removePendingChatDeletions(response.entryIds.toSet())
-        }
-    }
-
-    override fun onPatchError(syncErrorResponse: SyncErrorResponse) {
-        logcat(LogPriority.ERROR) { "DuckChat-Sync: patch failed with ${syncErrorResponse.featureSyncError}" }
+    override fun onError(error: SyncErrorResponse) {
+        logcat(LogPriority.ERROR) { "DuckChat-Sync: patch failed with ${error.featureSyncError}" }
```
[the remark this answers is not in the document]
0nko: @cursor push d7b0c03236

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/sync/SyncDateProvider.kt
0nko: The `CurrentTimeProvider` would need to be injected in every place where the `SyncDateProvider` is used and there would be a lot of "unrelated" changes. But we can do that if that's ok.

0nko: Actually I'll do it in a different PR, it'll be a mess.

duckchat/duckchat-impl/src/test/java/com/duckduckgo/duckchat/impl/sync/DuckChatSyncDataManagerTest.kt
```diff
@@ -210,7 +224,7 @@ class DuckChatSyncDataManagerTest {
             featureSyncError = FeatureSyncError.INVALID_REQUEST,
         )
 
-        testee.onPatchError(error)
+        testee.onError(error)
```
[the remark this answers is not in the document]
0nko: @cursor push ada72117f7

PR #7990, opened 2026-03-16
Subscription CTA: Check for active subscription

nalcalag:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1213690753964717?focus=true

### Description
Don't show subscription dialog for subscribers

### Steps to test this PR

_Pre steps_
- [x] Apply patch on https://app.asana.com/1/137249556945/project/1209991789468715/task/1210448620621729?focus=true

_Returning users who skip onboarding_
- [x] Fresh install
- [x] Skip onboarding as a returning user (Skip onboarding as a returning user (I've been here before > Start Browsing)
- [x] Go to Feature Flag Inventory and enable `privacyProCtaSkippedOnboarding`
- [x] Purchase a test subscription
- [x] Close the app
- [x] Change the date in your device for >=7 days after today
- [x] Open the app and check subscription onboarding dialog doesn't appear in a new tab page

_Regular onboarding_
- [x] Set the device time back to normal
- [x] Fresh install
- [x] Don't skip onboarding and go to browser
- [x] Purchase a test subscription
- [x] Go through onboarding and check Subscription onboarding dialog doesn't appear after 'end onboarding dialog' in new tab page.

### No UI changes

---

> [!NOTE]
> **Medium Risk**
> Changes onboarding CTA eligibility logic to depend on subscription status, which can affect when onboarding is considered complete and whether promo dialogs appear. Risk is moderate due to potential edge cases if `SubscriptionStatus` is incorrect or delayed.
> 
> **Overview**
> Prevents the Privacy Pro onboarding CTA/dialog from showing to users who already have an active subscription by additionally gating CTA availability on `subscriptions.getSubscriptionStatus()` (only show when status is `UNKNOWN`).
> 
> Updates onboarding completion/required CTA logic and adjusts/adds unit tests to cover subscribed vs unsubscribed scenarios and to stub the new subscription-status dependency.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9e0b7dd283770ba0d850ca8aef9b7dbdebb2f2d9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
@@ -51,9 +51,6 @@ import com.duckduckgo.duckplayer.api.DuckPlayer
 import com.duckduckgo.duckplayer.api.DuckPlayer.DuckPlayerState
 import com.duckduckgo.duckplayer.api.PrivatePlayerMode.AlwaysAsk
 import com.duckduckgo.subscriptions.api.SubscriptionStatus
-import com.duckduckgo.subscriptions.api.SubscriptionStatus.AUTO_RENEWABLE
-import com.duckduckgo.subscriptions.api.SubscriptionStatus.GRACE_PERIOD
-import com.duckduckgo.subscriptions.api.SubscriptionStatus.NOT_AUTO_RENEWABLE
 import com.duckduckgo.subscriptions.api.Subscriptions
 import dagger.SingleInstanceIn
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -107,8 +104,7 @@ class CtaViewModel @Inject constructor(
             }
 
     private suspend fun isPrivacyProCtaAvailable(): Boolean =
-        subscriptions.isEligible() && !subscriptions.getSubscriptionStatus().isActive() && extendedOnboardingFeatureToggles.privacyProCta()
-            .isEnabled()
+        subscriptions.isEligible() && hasNoSubscription() && extendedOnboardingFeatureToggles.privacyProCta().isEnabled()
 
     // Exposed for onboarding dev settings and tests. Used internally for completion checks
     @VisibleForTesting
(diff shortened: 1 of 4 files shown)


nalcalag:
* **#7990** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7990?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/7990?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -103,7 +107,8 @@ class CtaViewModel @Inject constructor(
             }
 
     private suspend fun isPrivacyProCtaAvailable(): Boolean =
-        subscriptions.isEligible() && extendedOnboardingFeatureToggles.privacyProCta().isEnabled()
+        subscriptions.isEligible() && !subscriptions.getSubscriptionStatus().isActive() && extendedOnboardingFeatureToggles.privacyProCta()
```
lmac012: This is an edge case, but I think we could just check if `status == UNKNOWN`. I guess if someone had a subscription and it recently expired, then they are aware that the app offers subscription and the CTA won't be useful. It the subscription expired more than 20 days ago then the status will change to UNKNOWN anyway and the user becomes eligible for the CTA again.

nalcalag: @lmac012 would this be an edge case for all `isActive` checks across the code? If so, I think we should fix that edge case in a different task in the AoR for all the use cases

lmac012: No, `isActive()` works as expected, I just think in this specific scenario we don't just want to check if the subscription is active - we should rather check if there is any subscription at all.

nalcalag: Sorry, I misunderstood the message the first time, I thought UNKNOWN was some kind of edge-case state that I needed to handle.
I double-checked with the product stakeholders last night as I wasn't sure of the expected behaviour here, and they agreed on not showing this if they have an expired subscription so I’ll implement what you suggested 👌

PR #7993, opened 2026-03-16
Enable Develocity Predictive Test Selection

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1213614217074373?focus=true

### Description
Enables Develocity PTS for JVM unit tests for local builds and PR checks. Keeps it disabled for post-merge checks and nightly flows since we want to run the full suite.

This runs with the [Standard profile](https://docs.gradle.com/develocity/current/using-develocity/predictive-test-selection/#selection-profiles) that balances speed and selecting relevant tests. Alternatively, we can go with Conservative to select more tests but reduces time savings.

### Steps to test this PR

_Run tests locally_
- [x] Run `./gradlew :pir-impl:testDebugUnitTest -Dpts.enabled=false` (replace `pir-impl` with any module you are familiar with the unit tests) to get a baseline time of how long it takes. It should run all tests in that module
- [x] Run `./gradlew :pir-impl:testDebugUnitTest` twice in a row. The second time it should finish in less than a second and run no tests.
- [ ] Now change some code in that module in a way that would break the test. For example in `pir-impl`, edit `PirAuthInterceptor:65` and change `bearer` to `Bearer`. This change will break one of the tests that checks for correct header value.
- [ ] Run `./gradlew :pir-impl:testDebugUnitTest` again. You should see something like `Predictive Test Selection: 4 of 74 test classes selected with profile 'Standard' (saving 54.371s serial time)` and the tests that were run should fail.

### UI changes
No UI change

---

> [!NOTE]
> **Medium Risk**
> Changes the JVM unit-test execution model in CI by enabling Develocity Predictive Test Selection by default, which can unintentionally skip relevant tests if misconfigured. Adds JUnit Platform/Vintage dependencies across modules, which may change how tests are discovered and run.
> 
> **Overview**
> **Enables Develocity Predictive Test Selection (PTS) for JVM unit tests** by configuring all Gradle `Test` tasks to `useJUnitPlatform()` and wiring `develocity.predictiveTestSelection.enabled` to a `-Dpts.enabled` system property (defaulting to `true`).
> 
> CI workflows now **disable PTS for post-merge/nightly and external reference test runs** by passing `-Dpts.enabled=false`, while PR checks keep PTS enabled by default. This also adds `org.junit.vintage:junit-vintage-engine` (and its version pin) to test dependencies so existing JUnit 4 tests continue to run under the JUnit Platform.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e0dfd3874b5f5ee6ba4b19830cdb3edb89b09444. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -582,7 +582,7 @@ dependencies {
 
     testImplementation project(':vpn-api-test')
     testImplementation "org.mockito.kotlin:mockito-kotlin:_"
-    testImplementation 'org.junit.vintage:junit-vintage-engine:5.10.2'
+    testImplementation 'org.junit.vintage:junit-vintage-engine:_'
     testImplementation Testing.junit4
     testImplementation AndroidX.archCore.testing
     testImplementation AndroidX.core
--- a/gradle/android-library.gradle
+++ b/gradle/android-library.gradle
@@ -54,7 +54,7 @@ android {
 }
 dependencies {
     lintChecks project(':lint-rules')
-    testImplementation 'org.junit.vintage:junit-vintage-engine:5.10.2'
+    testImplementation 'org.junit.vintage:junit-vintage-engine:_'
 }
 
 tasks.register('checks') {
--- a/lint-rules/build.gradle
+++ b/lint-rules/build.gradle
@@ -22,7 +22,7 @@ dependencies {
     compileOnly "com.android.tools.lint:lint-checks:$lint_version"
 
     testImplementation Testing.junit4
-    testImplementation 'org.junit.vintage:junit-vintage-engine:5.10.2'
+    testImplementation 'org.junit.vintage:junit-vintage-engine:_'
     testImplementation "com.android.tools.lint:lint-api:$lint_version"
     testImplementation "com.android.tools.lint:lint-tests:$lint_version"
 }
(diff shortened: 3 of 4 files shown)


landomen:
cc @ribafish

CDRussell:
@landomen is it supposed to work like this, where even if i supply `-Dpts.enabled=false` it is still skipping tests? Would have expected with the flag supplied set to false then it would run everything

Screenshot 2026-03-18 at 10 15 41


**UPDATE**
OK never mind. this is _other_ caching that's doing this and is the same on `develop` already

app/build.gradle
```diff
@@ -582,6 +582,7 @@ dependencies {
 
     testImplementation project(':vpn-api-test')
     testImplementation "org.mockito.kotlin:mockito-kotlin:_"
+    testImplementation 'org.junit.vintage:junit-vintage-engine:5.10.2'
```
CDRussell: we should centralise this version unless there's a specific reason it was done this way?

landomen: No specific reason. Centralized the version now, thanks for the suggestion

PR #8026, opened 2026-03-19
Extract tracker counter logic

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1213529826859023?focus=true

### Description
- Extracts blocked trackers counter logic from TrackerDetector to WebViewRequestInterceptor. Since TrackerDetector will be used in PIR and we don't want to contribute to counter logic.
- Enables multi process invalidation for the AppDatabase since it will be also accessed from :pir process.

### Steps to test this PR

- [x] Smoke test web page loading and tracker blocking
- [x] Load a few pages, check that the blocked trackers counter on new tab page increases correctly

### UI changes
No UI changes

---

> [!NOTE]
> **Medium Risk**
> Moves persistence of blocked-tracker records to a different call site and enables Room multi-process invalidation, which could affect tracker counts and database behavior across processes.
> 
> **Overview**
> **Blocked-tracker recording is moved out of `TrackerDetectorImpl` and into `WebViewRequestInterceptor`.** The interceptor now inserts `WebTrackerBlocked` rows via `WebTrackersBlockedDao` whenever a request (including cloaked CNAME cases and service worker interceptions) is blocked.
> 
> `WebViewRequestInterceptor` and its DI wiring are updated to accept `WebTrackersBlockedDao`, and tests/reference tests are updated accordingly (including new assertions that uncloaked CNAME URLs and entity display names are recorded).
> 
> **Database config change:** `AppDatabase` now enables Room `enableMultiInstanceInvalidation()` to support multi-process access.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c74f660ac701cd68e64861ff08ae6ad4e816334. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8014, opened 2026-03-18
Move TrackerDetector interface to a new module

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1210594645151737/task/1213435544907075?focus=true

### Description
Moves `TrackerDetector` interface to new `tracker-detection-api` module so it can later be used in PIR.

### Steps to test this PR
Smoke test tracker blocking (load a few pages, check trackers are blocked, check blocked count is increased.

### UI changes
No UI changes

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes DI bindings and Gradle module boundaries for tracker detection, which could break injection/build wiring or tracker blocking if misconfigured, though runtime logic is largely unchanged.
> 
> **Overview**
> Moves the `TrackerDetector` interface out of `app` into a new `:tracker-detection-api` module (`com.duckduckgo.tracker.detection.api`) and updates `app`/tests to import it from the new package.
> 
> Splits responsibilities by introducing `TrackerDetectorClientProvider` for `addClient` wiring, updating `TrackerDataLoader` and reference tests to use that provider while `TrackerDetectorImpl` now binds to both the new API interface and the client-provider interface via DI.
> 
> Updates Gradle configuration to include `:tracker-detection-api` as an app dependency and relaxes the top-level API-module dependency rule to allow API modules to depend on `:browser-api`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ae8c939f0f502dc05e5ea67c7a5f2235657ba493. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8033, opened 2026-03-20
PIR: Add tracker blocking

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1213529826859025?focus=true

### Description
Integrated tracker blocking with PIR

### Steps to test this PR
See https://app.asana.com/1/137249556945/task/1213740105102149?focus=true

### UI changes
No UI steps

---

> [!NOTE]
> **Medium Risk**
> Adds request interception/blocking in PIR’s WebView and loads tracker data in the PIR process, which can affect page behavior and introduce subtle WebView/threading issues if misconfigured.
> 
> **Overview**
> Integrates tracker blocking into PIR by introducing a `PirRequestInterceptor` (`PirWebRequestInterceptor`) that uses `TrackerDetector` to evaluate subresource requests and returns an empty `WebResourceResponse` to block those classified as `TrackerStatus.BLOCKED`, gated behind a new remote feature toggle `pir.trackerBlocking` (default *internal*).
> 
> Hooks this interceptor into PIR’s detached WebView via `shouldInterceptRequest`, and ensures tracker datasets are available in the PIR secondary process by having `TrackerDataLoader` also run on `onPirProcessCreated` (plus extra PIR process lifecycle logging). Also adds `tracker-detection-api` as a PIR implementation dependency and includes unit tests covering the interceptor’s gating and blocking behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 455572576cb227e3a358df0ee346bd67106b449d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/DuckDuckGoApplication.kt
@@ -126,8 +126,8 @@ open class DuckDuckGoApplication : HasDaggerInjector, MultiProcessApplication() 
                     // is good enough.
                     // See https://developer.android.com/reference/android/arch/lifecycle/ProcessLifecycleOwner#get
                     ProcessLifecycleOwner.get().lifecycle.apply {
+                        logcat { "PIR-LIFECYCLE: New PIR process created with pid=${android.os.Process.myPid()}" }
                         pirLifecycleObserverPluginPoint.getPlugins().forEach {
-                            logcat { "PIR-LIFECYCLE: New PIR process created with pid=${android.os.Process.myPid()}" }
                             it.onPirProcessCreated()
                         }
                     }


landomen:
* **#8045** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8045?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8033** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8033?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8033?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8026** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8026?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8014** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8014?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/PirRemoteFeatures.kt
```diff
@@ -45,6 +45,9 @@ interface PirRemoteFeatures {
 
     @DefaultValue(DefaultFeatureValue.TRUE)
     fun useBundledBrokerJsons(): Toggle
+
+    @DefaultValue(DefaultFeatureValue.INTERNAL)
```
karlenDimla, line 49: Just note that once we update this in privacy config, it will automatically be replaced with the value from privacy config.

app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDataLoader.kt
```diff
@@ -61,12 +66,16 @@ class TrackerDataLoader @Inject constructor(
...
     private val optimizeTrackerEvaluationRCWrapper: OptimizeTrackerEvaluationRCWrapper,
-) : MainProcessLifecycleObserver {
+) : MainProcessLifecycleObserver, PirProcessLifecycleObserver {
 
     override fun onCreate(owner: LifecycleOwner) {
         appCoroutineScope.launch(dispatcherProvider.io()) { loadData() }
     }
 
+    override fun onPirProcessCreated() {
+        appCoroutineScope.launch(dispatcherProvider.io()) { loadData() }
```
karlenDimla, line 76: NIT: Ideally if we have _trackerBlocking_ disabled, we don’t need to do this. It does not harm but we are doing extra work that will not be used at the beginning of the process.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/PirRemoteFeatures.kt
landomen: I thought about using `@InternalAlwaysEnabled` but decided against it in case we want to turn it off for everyone. I can use that if you think it's better in this case?

app/src/main/java/com/duckduckgo/app/trackerdetection/TrackerDataLoader.kt
landomen: @karlenDimla I was thinking the same and I think we should do it so we can disable it in case of issues. 

But that would mean exposing `isTrackerBlockingEnabled` in PIR public API since `PirRemoteFeatures` is in the `pir-impl` module and `:app` should't import from `-impl` modules. However, I think we shouldn't really expose that since it's an implementation detail of PIR.

Alternatively, we import `PirRemoteFeatures` directly from `-impl` module and ignore the lint warning. What do you think?

karlenDimla: `isTrackerBlockingEnabled` is an implementation detail of PIR so I don’t think we should expose it in a public API.

I don’t think we should import `PirRemoteFeatures` either. We need to rethink more the approach. Don’t have to be blocking for now.

PR #8044, opened 2026-03-20
Move RemoteMessagingRepository from API to impl module

catalinradoiu:
Task/Issue URL:https://app.asana.com/1/137249556945/project/1211724162604201/task/1213744279081578?focus=true

### Description

Refactored remote messaging architecture by consolidating functionality into `RemoteMessageModel`. The `RemoteMessagingRepository` interface was moved from the API module to the implementation module, and two duplicate classes (`RemoteMessagingModel` and `RemoteMessagingViewModel`) were removed from the app module. The `NewAddressBarOptionManager` now uses `RemoteMessageModel` instead of `RemoteMessagingRepository` and calls the new `isMessageDismissed()` method instead of checking dismissed messages directly. Added `isMessageDismissed()` and `dismissMessage()` methods to the `RemoteMessageModel` interface to provide cleaner access to message state.

### Steps to test this PR
CI passes

Also check https://app.asana.com/1/137249556945/project/1211724162604201/task/1213748927794483?focus=true

### UI changes
No UI changes - internal refactoring only

---

> [!NOTE]
> **Medium Risk**
> Moderate refactor of remote messaging plumbing (interfaces, DI wiring, and share/dismiss side effects) that could affect message display/dismissal and related pixel firing, though behavior is largely preserved.
> 
> **Overview**
> **Refactors remote messaging to reduce API surface and centralize state in `RemoteMessageModel`.** The `RemoteMessagingRepository` interface is removed from the API module and re-homed in the impl module, with call sites updated accordingly.
> 
> **Updates RMF consumption points** to use the new model APIs: `getActiveMessages` is renamed to `observeActiveMessages`, and `RemoteMessageModel` gains `isMessageDismissed` (used by `NewAddressBarOptionManager` instead of reading `dismissedMessages` directly).
> 
> **Cleans up app-level duplicates and share plumbing.** Removes app-module `RemoteMessagingModel`/`RemoteMessagingViewModel` and the app `SharePromoLinkRMFBroadCastReceiver` + manifest entry; share-chooser callbacks now use an injected `SharePromoLinkIntentFactory` (implemented in remote-messaging impl) to generate the `PendingIntent` for the impl `SharePromoLinkBroadCastReceiver`.
> 
> **Small threading adjustments**: `AppRemoteMessagingRepository.dismissMessage` no longer wraps work in `withContext`, and modal/cards-list dismiss pixel helpers are explicitly dispatched on `dispatchers.io()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4047abaea2433858156b95ff68edaa21db6a307d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


catalinradoiu:
* **#8044** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8044?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8044?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/remotemessage/SharePromoLinkRMFBroadCastReceiver.kt
```diff
@@ -39,7 +39,7 @@ class SharePromoLinkRMFBroadCastReceiver : BroadcastReceiver() {
     lateinit var pixel: Pixel
 
     @Inject
-    lateinit var remoteMessagingRepository: RemoteMessagingRepository
+    lateinit var remoteMessageModel: RemoteMessageModel
```
anikiki: What's the difference between this class and the one in `remote-messaging` module -> `SharePromoLinkBroadCastReceiver`? Can they be consolidated into one that lives in the `remote-messaging` module?

remote-messaging/remote-messaging-api/src/main/java/com/duckduckgo/remote/messaging/api/RemoteMessageModel.kt
anikiki, line 21: Now it might be a good time to make some changes to this interface:

1/ We have "dismissMessage" and "onMessageDismissed". This introduces uncertainty for the future: which one to use and when? If both are needed, make the names clearer.

2/ Rename `getActiveMessages()` to `observeActiveMessage()`

3/ `getRemoteMessageImageFile` and `clearMessageImage` don't really belong here as they're leaking implementation details.

remote-messaging/remote-messaging-api/src/main/java/com/duckduckgo/remote/messaging/api/RemoteMessageModel.kt
```diff
@@ -37,4 +37,8 @@ interface RemoteMessageModel {
     suspend fun getRemoteMessageImageFile(surface: Surface): String?
 
     suspend fun clearMessageImage(surface: Surface)
+
+    fun isMessageDismissed(id: String): Boolean
```
anikiki: This should be suspending.

remote-messaging/remote-messaging-impl/src/main/java/com/duckduckgo/remote/messaging/impl/AppRemoteMessagingRepository.kt
anikiki, line 114: This needs to be done on IO.

catalinradoiu: I have removed the dispatchers from the repository, as it’s the caller the one who should ensure that the method is being called on the IO dispatcher and not the repository itself. This method it being called from IO dispatchers.

remote-messaging/remote-messaging-api/src/main/java/com/duckduckgo/remote/messaging/api/RemoteMessageModel.kt
catalinradoiu: Fixed

app/src/main/java/com/duckduckgo/app/browser/remotemessage/SharePromoLinkRMFBroadCastReceiver.kt
catalinradoiu: I have remove the duplicated receiver and we are now using only the one inside the `remote-messaging` module. I have added another API class thar provides the pending intent from the receiver from the module.

remote-messaging/remote-messaging-api/src/main/java/com/duckduckgo/remote/messaging/api/RemoteMessageModel.kt
catalinradoiu: 1. I haved remove the other method, as we no longer need it.
2. Done
3. I agree with that. However this would require checking each place where the image is being cleared and moving some of the method calls around, so I was thinking if we should do this in a separate PR.

app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabPageView.kt
```diff
@@ -48,7 +47,6 @@ import com.duckduckgo.app.browser.newtab.NewTabPageViewModel.Command.SubmitUrl
 import com.duckduckgo.app.browser.newtab.NewTabPageViewModel.NewTabPageViewModelFactory
 import com.duckduckgo.app.browser.newtab.NewTabPageViewModel.NewTabPageViewModelProviderFactory
 import com.duckduckgo.app.browser.newtab.NewTabPageViewModel.ViewState
-import com.duckduckgo.app.browser.remotemessage.SharePromoLinkRMFBroadCastReceiver
 import com.duckduckgo.app.browser.remotemessage.asMessage
```
anikiki, line 50: We should use `com.duckduckgo.remote.messaging.impl.mappers.asMessage`, and delete the duplicated `RemoteMessageMapper` class from the `app` module.

catalinradoiu: We will need to revise the public API in thi case. `asMessage` is in the remote messaging implementation module. If we delete the one from the `app` module then we need to move the mapper to the API module, as we cannot use implementation dependencies in the `app` module. However the `impl` module depends on the design system module, as the `Message` comes from that module. So far I haven’t seen design system dependencies in the API modules. Do we want to do that? 
I also have a follow-up task for this to revise the public API of the remote messaging module so I can defer this work for that task if you think this can complicate the current PR. I can also add the design system dependency to the API module but I am wondering if that’s ok to do.

anikiki: Let's revise this in the next iteration.

PR #8154, opened 2026-03-30
Enforce allowed domains check for Duck Player

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1176047645786601/task/1213813172746773?focus=true

### Description
Enforce allowed domains check for Duck Player

### Steps to test this PR
QA-optional, @not-a-rootkit already tested it 

### UI changes
n/a

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes runtime gating of JS-to-native messaging and could block legitimate messages if domain configuration/URL parsing is wrong, but scope is limited to Duck Player messaging.
> 
> **Overview**
> Duck Player’s JS bridge now enforces *allowed-domain validation* before processing incoming `JsMessage`s.
> 
> `process` gates handling on `isUrlAllowed(allowedDomains, url)` and additionally requires each matched `JsMessageHandler` to pass its own `isUrlAllowed(it.allowedDomains, url)` check; `isUrlAllowed` was updated to take an explicit domain list parameter.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aacb64458b7208ce50db53deda6f99a978498548. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/duckplayer/duckplayer-impl/src/main/java/com/duckduckgo/duckplayer/impl/DuckPlayerScriptsJsMessaging.kt
+++ b/duckplayer/duckplayer-impl/src/main/java/com/duckduckgo/duckplayer/impl/DuckPlayerScriptsJsMessaging.kt
@@ -64,7 +64,7 @@ class DuckPlayerScriptsJsMessaging @Inject constructor(
             }
             jsMessage?.let {
                 logcat { jsMessage.toString() }
-                if (this.secret == secret && context == jsMessage.context) {
+                if (this.secret == secret && context == jsMessage.context && isUrlAllowed(allowedDomains, url)) {
                     handlers.firstOrNull {
                         it.methods.contains(jsMessage.method) && it.featureName == jsMessage.featureName && isUrlAllowed(it.allowedDomains, url)
                     }?.process(jsMessage, this, jsMessageCallback)
@@ -107,7 +107,7 @@ class DuckPlayerScriptsJsMessaging @Inject constructor(
     override val context: String = "specialPages"
     override val callbackName: String = "messageCallback"
     override val secret: String = "duckduckgo-android-messaging-secret"
-    override val allowedDomains: List<String> = handlers.flatMap { it.allowedDomains }
+    override val allowedDomains: List<String> = listOf()
 
     private fun isUrlAllowed(allowedDomains: List<String>, url: String?): Boolean {
         if (allowedDomains.isEmpty()) return true


CrisBarreiro:
* **#8154** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8154?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8154?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

duckplayer/duckplayer-impl/src/main/java/com/duckduckgo/duckplayer/impl/DuckPlayerScriptsJsMessaging.kt
```diff
@@ -64,9 +64,9 @@ class DuckPlayerScriptsJsMessaging @Inject constructor(
             }
             jsMessage?.let {
                 logcat { jsMessage.toString() }
-                if (this.secret == secret && context == jsMessage.context && isUrlAllowed(url)) {
```
marcosholgado, line 67: Not sure about this, more below.

duckplayer/duckplayer-impl/src/main/java/com/duckduckgo/duckplayer/impl/DuckPlayerScriptsJsMessaging.kt
```diff
@@ -107,9 +107,9 @@ class DuckPlayerScriptsJsMessaging @Inject constructor(
     override val context: String = "specialPages"
     override val callbackName: String = "messageCallback"
     override val secret: String = "duckduckgo-android-messaging-secret"
-    override val allowedDomains: List<String> = listOf()
+    override val allowedDomains: List<String> = handlers.flatMap { it.allowedDomains }
```
marcosholgado: I don't think this is correct. Allowed domains at the JsMessaging level are to allow or not any messages regardless of the handlers to return earlier. If this was implemented correctly we would be bypassing handlers when we shouldn't but we are not because this is not used. I think we should do the same that we do in `ContentScopeScriptsJsMessaging`, which is...

`if (this.secret == secret && context == jsMessage.context && isUrlAllowed(allowedDomains, domain)) {`
...
`it.methods.contains(jsMessage.method) && it.featureName == jsMessage.featureName && isUrlAllowed(it.allowedDomains, domain)`
... 
```
 private fun isUrlAllowed(
        allowedDomains: List<String>,
        url: String?,
    ): Boolean {
        if (allowedDomains.isEmpty()) return true
        val host = url ?: return false
        val eTld = host.toTldPlusOne()
        return allowedDomains.contains(host) || (eTld != null && allowedDomains.contains(eTld))
    }
```

note: This is why I always thought the JsMessaging implementations should be codegen or similar.

CrisBarreiro: I had that at first, but then decided to remove the first check `if (this.secret == secret && context == jsMessage.context && isUrlAllowed(allowedDomains, domain)) {` because that seemed redundant with the current implementation `override val allowedDomains: List<String> = handlers.flatMap { it.allowedDomains }`. 

So, do we just leave `override val allowedDomains: List<String> = listOf()`, and then keep the content at the handler level?

marcosholgado: The first check is for JsMessaging and the 2nd is for each handler. Right now, JsMessaging won't ever check anything because the checks are only at the handler level so this part `override val allowedDomains: List<String> = handlers.flatMap { it.allowedDomains }` is not doing anything.

If all the messages should only be executed in DuckPlayer then imo we should check at the JsMessaging and leave each handler with an empty list.

PR #8180, opened 2026-04-01
Restore the tab pager adapter state after populating the list of tabs and avoid navigating to input screen on fragment restoration

catalinradoiu:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1213836390603173?focus=true

### Description

Fixed 2 issues:

1. FragmentStateAdapter GC'd restored fragments: In BrowserActivity.initializeTabs(), tabPagerAdapter.restore() was called before the adapter's tabs list was populated. The gcFragments() call inside restoreState() checks containsItem() for each restored fragment, but with tabs empty, every fragment failed this check and was removed. The adapter then created a brand-new fragment via createFragment() with null savedInstanceState.
2. ViewModel flow re-emitted on recreation: The flowSelectedTab subscription in the ViewModel's init block re-evaluated on every new ViewModel instance. Since the tab still had no URL, the filter passed and LaunchInputScreen was emitted again, causing the loop

Fix:

1. Populate adapter tabs before restoring state — ensures containsItem() returns true during GC, so restored fragments survive with their savedInstanceState.
2. Extract the flow subscription to observeSelectedTab(isRestored) — the fragment calls this method, passing whether it was restored from saved state. The first emission is skipped when isRestored is true, preventing the re-launch. The flag resets after the first emission so subsequent tab switches work normally.

### Steps to test this PR

_Tab Restoration Behavior_

- [ ] Enable "Don't keep activities" in Developer Options
- [ ] Enable “Search&[Duck.ai](http://Duck.ai)” omnibar
- [ ] Go to new tab page
- [ ] Click on the omnibar
- [ ] Click on the menu icon
- [ ] The menu should be opened

### UI changes

| Before | After |
| --- | --- |
|  |  |

---

> [!NOTE]
> **Medium Risk**
> Touches tab restoration and lifecycle-driven flow collection in core browsing UI; ordering/remote-config gating changes could affect tab switching or state restore on low-memory/rotation scenarios.
> 
> **Overview**
> Fixes swiping-tabs state restoration by adding a remote-configed `tabStateRestorationFix` that restores `TabPagerAdapter` state **after** `onTabsUpdated` populates the tab list (instead of during `initializeTabs`), avoiding restored fragments being GC’d; state is removed from the bundle after applying.
> 
> Prevents an Input Screen relaunch loop on fragment/viewmodel recreation by moving the selected-tab flow logic out of `BrowserTabViewModel` init into `observeSelectedTab(isRestored)`, called from `BrowserTabFragment`, and skipping the first emission on restore when the fix is enabled. Also ensures shown fragments’ root views are explicitly made `VISIBLE` in `FragmentStateAdapter.showFragment`, and updates/adds unit tests for the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25f60d2771ddcda083db9839a14f7cab0c555183. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -1340,14 +1340,14 @@ open class BrowserActivity : DuckDuckGoActivity() {
                         tabPagerAdapter.restore(it)
                     }
                 }
+
+                // LiveData observers are restarted on each showWebContent() call; we want to subscribe to
+                // flows only once, so a separate initialization is necessary
+                configureFlowCollectors()
             }
 
             binding.fragmentContainer.isVisible = !swipingTabsFeature.isEnabled
             tabPager.isVisible = swipingTabsFeature.isEnabled
-
-            // LiveData observers are restarted on each showWebContent() call; we want to subscribe to
-            // flows only once, so a separate initialization is necessary
-            configureFlowCollectors()
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -1340,14 +1340,16 @@ open class BrowserActivity : DuckDuckGoActivity() {
                         tabPagerAdapter.restore(it)
                     }
                 }
-
-                // LiveData observers are restarted on each showWebContent() call; we want to subscribe to
-                // flows only once, so a separate initialization is necessary
-                configureFlowCollectors()
             }
 
             binding.fragmentContainer.isVisible = !swipingTabsFeature.isEnabled
             tabPager.isVisible = swipingTabsFeature.isEnabled
+
+            if (swipingTabsFeature.isEnabled) {
+                // LiveData observers are restarted on each showWebContent() call; we want to subscribe to
+                // flows only once, so a separate initialization is necessary
+                configureFlowCollectors()
+            }
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -1332,9 +1332,11 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
             configureViewPagerForSystemAutofill(tabPager)
 
-            // savedInstanceState?.getBundle(KEY_TAB_PAGER_STATE)?.let {
-            //     tabPagerAdapter.restore(it)
-            // }
+            if (!androidBrowserConfigFeature.tabStateRestorationFix().isEnabled()) {
+                savedInstanceState?.getBundle(KEY_TAB_PAGER_STATE)?.let {
+                    tabPagerAdapter.restore(it)
+                }
+            }
         }
 
         binding.fragmentContainer.isVisible = !swipingTabsFeature.isEnabled
@@ -1487,10 +1489,13 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
     private fun onTabsUpdated(updatedTabIds: List<TabModel>, savedInstanceState: Bundle?) {
         tabPagerAdapter.onTabsUpdated(updatedTabIds)
-        savedInstanceState?.getBundle(KEY_TAB_PAGER_STATE)?.let {
-            tabPagerAdapter.restore(it)
+
+        if (androidBrowserConfigFeature.tabStateRestorationFix().isEnabled()) {
+            savedInstanceState?.getBundle(KEY_TAB_PAGER_STATE)?.let {
+                tabPagerAdapter.restore(it)
+            }
+            savedInstanceState?.remove(KEY_TAB_PAGER_STATE)
         }
-        savedInstanceState?.remove(KEY_TAB_PAGER_STATE)
     }
 
     fun launchNewTab(
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -358,9 +358,6 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
         initializeTabs(savedInstanceState)
 
-        // LiveData observers are restarted on each showWebContent() call; we want to subscribe to
-        // flows only once, so a separate initialization is necessary
-        configureFlowCollectors()
         setupFireDialogListener()
 
         viewModel.viewState
@@ -401,27 +398,23 @@ open class BrowserActivity : DuckDuckGoActivity() {
         }
     }
 
-    private fun configureFlowCollectors() {
-        if (swipingTabsFeature.isEnabled) {
-            lifecycleScope.launch {
-                repeatOnLifecycle(STARTED) {
-                    launch {
-                        viewModel.tabsFlow.collectLatest {
-                            tabManager.onTabsChanged(it)
-                        }
-                    }
+    private suspend fun configureFlowCollectors() {
+        repeatOnLifecycle(STARTED) {
+            launch {
+                viewModel.tabsFlow.collectLatest {
+                    tabManager.onTabsChanged(it)
+                }
+            }
 
-                    launch {
-                        viewModel.selectedTabFlow.collectLatest {
-                            tabManager.onSelectedTabChanged(it)
-                        }
-                    }
+            launch {
+                viewModel.selectedTabFlow.collectLatest {
+                    tabManager.onSelectedTabChanged(it)
+                }
+            }
 
-                    launch {
-                        viewModel.selectedTabIndex.collectLatest {
-                            onMoveToTabRequested(it)
-                        }
-                    }
+            launch {
+                viewModel.selectedTabIndex.collectLatest {
+                    onMoveToTabRequested(it)
                 }
             }
         }
@@ -1319,28 +1312,43 @@ open class BrowserActivity : DuckDuckGoActivity() {
     }
 
     private fun initializeTabs(savedInstanceState: Bundle?) {
-        if (swipingTabsFeature.isEnabled) {
-            tabManager.registerCallbacks(
-                onTabsUpdated = {
-                    onTabsUpdated(it, savedInstanceState)
-                },
-            )
+        lifecycleScope.launch {
+            val tabRestorationFixEnabled = withContext(dispatcherProvider.io()) {
+                androidBrowserConfigFeature.tabStateRestorationFix().isEnabled()
+            }
+            if (swipingTabsFeature.isEnabled) {
+                tabManager.registerCallbacks(
+                    onTabsUpdated = {
+                        onTabsUpdated(it, savedInstanceState, tabRestorationFixEnabled)
+                    },
+                )
 
-            tabPager.adapter = tabPagerAdapter
-            tabPager.registerOnPageChangeCallback(onTabPageChangeListener)
-            tabPager.setPageTransformer(MarginPageTransformer(resources.getDimension(com.duckduckgo.mobile.android.R.dimen.keyline_1).toPx().toInt()))
+                tabPager.adapter = tabPagerAdapter
+                tabPager.registerOnPageChangeCallback(onTabPageChangeListener)
+                tabPager.setPageTransformer(
+                    MarginPageTransformer(
+                        resources.getDimension(com.duckduckgo.mobile.android.R.dimen.keyline_1)
+                            .toPx()
+                            .toInt(),
+                    ),
+                )
 
-            configureViewPagerForSystemAutofill(tabPager)
+                configureViewPagerForSystemAutofill(tabPager)
 
-            if (!androidBrowserConfigFeature.tabStateRestorationFix().isEnabled()) {
-                savedInstanceState?.getBundle(KEY_TAB_PAGER_STATE)?.let {
-                    tabPagerAdapter.restore(it)
+                if (!tabRestorationFixEnabled) {
+                    savedInstanceState?.getBundle(KEY_TAB_PAGER_STATE)?.let {
+                        tabPagerAdapter.restore(it)
+                    }
                 }
             }
-        }
 
-        binding.fragmentContainer.isVisible = !swipingTabsFeature.isEnabled
-        tabPager.isVisible = swipingTabsFeature.isEnabled
+            binding.fragmentContainer.isVisible = !swipingTabsFeature.isEnabled
+            tabPager.isVisible = swipingTabsFeature.isEnabled
+
+            // LiveData observers are restarted on each showWebContent() call; we want to subscribe to
+            // flows only once, so a separate initialization is necessary
+            configureFlowCollectors()
+        }
     }
 
     private fun configureViewPagerForSystemAutofill(viewPager: ViewPager2) {
@@ -1487,10 +1495,14 @@ open class BrowserActivity : DuckDuckGoActivity() {
         }
     }
 
-    private fun onTabsUpdated(updatedTabIds: List<TabModel>, savedInstanceState: Bundle?) {
+    private fun onTabsUpdated(
+        updatedTabIds: List<TabModel>,
+        savedInstanceState: Bundle?,
+        tabRestorationFixEnabled: Boolean,
+    ) {
         tabPagerAdapter.onTabsUpdated(updatedTabIds)
 
-        if (androidBrowserConfigFeature.tabStateRestorationFix().isEnabled()) {
+        if (tabRestorationFixEnabled) {
             savedInstanceState?.getBundle(KEY_TAB_PAGER_STATE)?.let {
                 tabPagerAdapter.restore(it)
             }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -1321,7 +1321,9 @@ open class BrowserActivity : DuckDuckGoActivity() {
     private fun initializeTabs(savedInstanceState: Bundle?) {
         if (swipingTabsFeature.isEnabled) {
             tabManager.registerCallbacks(
-                onTabsUpdated = ::onTabsUpdated,
+                onTabsUpdated = {
+                    onTabsUpdated(it, savedInstanceState)
+                },
             )
 
             tabPager.adapter = tabPagerAdapter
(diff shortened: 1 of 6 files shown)


catalinradoiu:
* **#8180** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8180?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8180?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -924,6 +898,38 @@ class BrowserTabViewModel @Inject constructor(
...
+                val isActiveTab = ::tabId.isInitialized && selectedTab?.tabId == tabId
+                val isOpenedFromAnotherTab = selectedTab?.sourceTabId != null
+                showInputScreenAutomatically && isActiveTab && selectedTab?.url.isNullOrBlank() && !isOpenedFromAnotherTab
+            }.flowOn(dispatchers.main()) // don't use the immediate dispatcher so that the tabId field has a chance to initialize
+            .onEach {
+                val hasPendingTabLaunch = externalIntentProcessingState.hasPendingTabLaunch
+                val hasPendingDuckAiOpen = externalIntentProcessingState.hasPendingDuckAiOpen
+                val hasPendingSnackbar = externalIntentProcessingState.hasPendingSnackbar
+
+                if (!hasPendingTabLaunch && !hasPendingDuckAiOpen && !hasPendingSnackbar && !isTabRestored) {
```
LukasPaczos:
```suggestion
                if (!hasPendingTabLaunch && !hasPendingDuckAiOpen && !hasPendingSnackbar && !isRestored) {
```
Do we need the class field at all? Wouldn't the local value be enough?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -924,6 +898,38 @@ class BrowserTabViewModel @Inject constructor(
         observeAccessibilitySettings()
     }
 
+    fun observeSelectedTab(isRestored: Boolean) {
+        isTabRestored = isRestored
+        // auto-launch input screen for new, empty tabs (New Tab Page)
+        tabRepository.flowSelectedTab
```
LukasPaczos: Since this flow is now started whenever `#observeSelectedTab` is called, we should store the job reference and cancel it to avoid potentially having multiple flows running at the same time.

LukasPaczos (commented):
Generally, the change looks good to me and seem to get the job done well! Good find and I think we can continue in this direction. Let's make sure to feature flag it though.

One issue I was able to reproduce:
1. Enable "Don't Keep Activities".
2. Create a bunch of tabs (more than 3).
3. Open Tab Manager, and select one of edge tabs.
4. I can reproduce nearly every time that if start swiping tabs, when I go beyond what would be typically cached, the tabs are broken (completely blank).

https://github.com/user-attachments/assets/8a30d754-cb5d-4a4b-90ca-402a3c54fd60

I'm not certain it's related to changes from this PR but I can't reproduce on main (with Input Screen disable of course, impossible otherwise due to the original issue). cc @0nko in case you have any ideas

catalinradoiu:
@LukasPaczos I have further investigated this issue and found out why this happens. We have our custom `FragmentStateAdapter`, which uses `hide()`/`show()` to keep fragments alive instead of removing and recreating them (the default library behavior). Fragments are added without a container ID (tag-only `add()`), so `Fragment.mContainer` is `null`.

The issue involves three pieces of the Fragment framework interacting:

1. `hideFragment` sets `mHidden = true`
When `onViewRecycled` calls `hideFragment`, it runs: `mFragmentManager.beginTransaction().hide(fragment).commitAllowingStateLoss();`

`FragmentManager.hideFragment()` sets `fragment.mHidden = true`. During normal swiping, the view stays `VISIBLE`, the fragment is just hidden at the FragmentManager level.

2. Tab switcher triggers a lifecycle transition that sets the view to `GONE`
When the tab switcher activity opens the `BrowserTabFragment` is destroeyd (due to “Don’t keep activities being enabled). When we go back and select one of the tabs at the edge, which has been hiden before due to holder being recycled (and  `FragmentManager.hideFragment()` called), we go through the `onCreateView` stage (as the fragment view is being createaed), inside `FragmentStateManager.createView()` which runs:

```
if (mFragment.mHidden) {
    mFragment.mView.setVisibility(View.GONE);
}
```

Since `mHidden` is `true`, the view is set to `GONE`. This doesn't happen during normal swiping because the fragment is just hidden and the view is not being destroyed.

3. `show()` can't undo the `GONE` visibility
When returning from the tab switcher, `showFragment` calls `show(fragment).commitNow()`. This sets `mHidden = false`, but the `SpecialEffectsController`, which is responsible for calling `view.setVisibility(View.VISIBLE)` is skipped because `mFragment.mContainer` is `null` (fragments are added with a tag only, no container ID):

```
// FragmentStateManager.java
if (mFragment.mHiddenChanged) {
    if (mFragment.mView != null && mFragment.mContainer != null) {  // skipped — mContainer is null
        SpecialEffectsController controller = ...
        controller.enqueueShow(this);  // never reached — this is what sets VISIBLE
    }
}
```

Result: `mHidden = false` but the view is still `GONE`.

The solution is to explicitly set the view to `VISIBLE` in `showFragment`, since `FragmentManager` cannot do it when `mContainer` is `null`:

```
private void showFragment(long itemId) {
    Fragment fragment = mFragments.get(itemId);
    if (fragment == null) return;

    if (fragment.isAdded() && fragment.isHidden()) {
        mFragmentManager.beginTransaction().show(fragment).commitNow();
    }
    if (fragment.getView() != null && fragment.getView().getVisibility() != View.VISIBLE) {
        fragment.getView().setVisibility(View.VISIBLE);
    }
}
```


This was not an issue before because `gcFragments` would remove the framents (as the state was not being handled correctly (the `tabs` in `TabPagerAdapter` were not initialised so `containsItem(itemId)` was false, therefore the fragment was removed). 

I will push a commit with the fix.
CC: @0nko

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
catalinradoiu: We are setting it false later, after the flow emits a new value, that’s why I used the class field.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
catalinradoiu: Added

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
LukasPaczos: I see, we're doing this so that swiping back to the new tab automatically opens the input screen.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -924,6 +899,39 @@ class BrowserTabViewModel @Inject constructor(
         observeAccessibilitySettings()
     }
 
+    fun observeSelectedTab(isRestored: Boolean) {
+        isTabRestored = isRestored
+        selectedTabObserver?.cancel()
+        // auto-launch input screen for new, empty tabs (New Tab Page)
+        selectedTabObserver = tabRepository.flowSelectedTab
+            .distinctUntilChangedBy { selectedTab -> selectedTab?.tabId } // only observe when the tab changes and ignore further updates
+            .filter { selectedTab ->
```
LukasPaczos:
```suggestion
    fun observeSelectedTab(isRestored: Boolean) {
        selectedTabObserver?.cancel()
        // auto-launch input screen for new, empty tabs (New Tab Page)
        selectedTabObserver = tabRepository.flowSelectedTab
            .distinctUntilChangedBy { selectedTab -> selectedTab?.tabId } // only observe when the tab changes and ignore further updates
            .let {
                if (isRestored) {
                    // Skip the first emission on restore to avoid a loop
                    // where closing InputScreenActivity would return to a restored NTP,
                    // which would restart this flow and reopen InputScreenActivity again.
                    // This happens when memory pressure destroys BrowserActivity while
                    // InputScreenActivity is in the foreground.
                    it.drop(1)
                } else {
                    it
                }
            }
            .filter { selectedTab ->
```

Wouldn't this do the same job without having to add a class field? I'm just worried that the field will be accessible across the class while being only usable within this specific function body, so it becomes noise for everything else.

LukasPaczos (commented):
Works as expected! Adding one suggestion below and let's also include the feature flag.

Also tagging @0nko to double-check the fragment management changes.

app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
```diff
@@ -579,6 +579,9 @@ private void showFragment(long itemId) {
         if (fragment.isAdded() && fragment.isHidden()) {
             mFragmentManager.beginTransaction().show(fragment).commitNow();
         }
+        if (fragment.getView() != null && fragment.getView().getVisibility() != View.VISIBLE) {
```
catalinradoiu: @LukasPaczos do you think it makes sense to feature flag this as well?

LukasPaczos: This looks relatively safe to me.

I'd only change it to work on a local ref to avoid any race conditions that could lead to NPE. This all in theory runs on the main thread, but some defensive programming shouldn't hurt.

```java
        View fragmentView = fragment.getView();
        if (fragmentView != null && fragmentView.getVisibility() != View.VISIBLE) {
            fragmentView.setVisibility(View.VISIBLE);
        }
```

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
catalinradoiu: Done, looks much better then using the local field 👍

app/src/main/java/com/duckduckgo/app/browser/tabs/adapter/FragmentStateAdapter.java
catalinradoiu: Done

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -924,6 +897,50 @@ class BrowserTabViewModel @Inject constructor(
...
+                    // Skip the first matching emission on restore to avoid a loop
+                    // where closing InputScreenActivity would return to a restored NTP,
+                    // which would restart this flow and reopen InputScreenActivity again.
+                    // This happens when memory pressure destroys BrowserActivity while
+                    // InputScreenActivity is in the foreground.
+                    it.drop(1)
+                } else {
+                    it
+                }
+            }
```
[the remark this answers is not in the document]
LukasPaczos: @catalinradoiu this is correct, that's why in my suggestion I had the `drop` before filtering.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -924,6 +897,50 @@ class BrowserTabViewModel @Inject constructor(
         observeAccessibilitySettings()
     }
 
+    fun observeSelectedTab(isRestored: Boolean) {
+        val isTabRestorationFixEnabled = androidBrowserConfig.tabStateRestorationFix().isEnabled()
```
LukasPaczos: This should be done on a worker thread.

0nko:
@catalinradoiu Really great investigation! 👏 Must not have been easy to pinpoint the culprit behind the behavior..

I've looked over the changes and don't see any problems, aside from the [BugBot comment](https://github.com/duckduckgo/Android/pull/8180/#discussion_r3051427344), which I think is legitimate.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
catalinradoiu: I think I will need some kind of synchronization mechanism in this case. I can move that to the coroutine but I just figured out that I need to do the same for the BrowserActivity. However I will need to get the value of the FF first before initialising the tabs (if using a coroutine), otherwise we could end up restoring the state after the fragments are initialised in the adapter, which will crash the app:

```
public final void restoreState(@NonNull Parcelable savedState) {
        if (!mFragments.isEmpty()) {
            throw new IllegalStateException(
                    "Expected the adapter to be 'fresh' while restoring state.");
        }

        Bundle bundle = (Bundle) savedState;
        if (bundle.getClassLoader() == null) {
            /* TODO(b/133752041): pass the class loader from {@link ViewPager2.SavedState } */
            bundle.setClassLoader(getClass().getClassLoader());
        }
```

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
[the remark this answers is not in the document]
catalinradoiu: I fixed it and will need to push the commit. Initially I had a failing unit test when using it before filtering but then I found out that the test was not set up correctly.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
catalinradoiu: I had to do a change inside `BrowserActivity`, since `initializeTabs()` was calling `androidBrowserConfigFeature.tabStateRestorationFix().isEnabled()` directly on the main thread. 

Additionally, `configureFlowCollectors()` was called separately from `initializeTabs()`. The flow collectors invoke `tabManager.onTabsChanged()`, which relies on the `onTabsUpdated` callback — but that callback is only registered inside `initializeTabs()` via `tabManager.registerCallbacks()`. If the flow emitted before `registerCallbacks()` was called, it would crash on an uninitialized callback.

The toggle check is now performed off the main thread `(withContext(dispatcherProvider.io()))`. The result must be resolved before proceeding because both `onTabsUpdated` (which receives it as a parameter) and the tab state restoration logic in `initializeTabs` depend on this flag.  

@0nko can you please take another look as well here?

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -1319,24 +1312,43 @@ open class BrowserActivity : DuckDuckGoActivity() {
...
+                // flows only once, so a separate initialization is necessary
+                configureFlowCollectors()
             }
-        }
 
-        binding.fragmentContainer.isVisible = !swipingTabsFeature.isEnabled
-        tabPager.isVisible = swipingTabsFeature.isEnabled
+            binding.fragmentContainer.isVisible = !swipingTabsFeature.isEnabled
+            tabPager.isVisible = swipingTabsFeature.isEnabled
+        }
```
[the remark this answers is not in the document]
catalinradoiu: I moved this before `configureFlowCollectors()`. Didn’t see an issue while testing (maybe the visibility change was executed before the flow was launched) but the comment makes sense.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -924,6 +897,54 @@ class BrowserTabViewModel @Inject constructor(
...
+                            // unless an onboarding promo message is displayed
+                            val hasPendingOnboardingPromo = ctaViewModel.isPromoOnboardingDialogShowing()
+                            if (!hasPendingOnboardingPromo) {
+                                command.value = LaunchInputScreen
+                            }
+                        }
+                    }
+                }.launchIn(viewModelScope)
+        }
+    }
```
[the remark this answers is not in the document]
catalinradoiu: I don’t think this can actually happen as the method is being called from `onCreate` inside `BrowserTabFragment` so we cannot really end up with multiple calls.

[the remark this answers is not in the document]
LukasPaczos: @catalinradoiu the fix is to move the `selectedTabObserver?.cancel()` right before new `selectedTabObserver` is launched, inside of the `viewModelScope.launch`:

```kotlin
    fun observeSelectedTab(isRestored: Boolean) {
        viewModelScope.launch {
            val isTabRestorationFixEnabled = withContext(dispatchers.io()) {
                androidBrowserConfig.tabStateRestorationFix().isEnabled()
            }
            selectedTabObserver?.cancel()
            // auto-launch input screen for new, empty tabs (New Tab Page)
            selectedTabObserver = tabRepository.flowSelectedTab
            // ...
```

app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
```diff
@@ -356,4 +356,16 @@ interface AndroidBrowserConfigFeature {
...
+     * Controls whether the tab state restoration fix is enabled for swiping tabs.
+     * When enabled, the adapter populates tabs before restoring state to prevent
+     * garbage collection.
+     * @return `true` when the remote config has the global "tabStateRestorationFix" androidBrowserConfig
+     * sub-feature flag enabled
+     * If the remote feature is not present defaults to `false`
+     */
+    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    @Toggle.InternalAlwaysEnabled
+    fun tabStateRestorationFix(): Toggle
```
LukasPaczos, line 369:
```suggestion
     * @return `true` when the remote config has the global "tabStateRestorationFix" androidBrowserConfig
     * sub-feature flag enabled
     * If the remote feature is not present defaults to `true`
     */
    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
    fun tabStateRestorationFix(): Toggle
```

This should serve as a kill-switch, so enabled by default with an option to remotely disable.

0nko:
@catalinradoiu The changes look good to me. I've tested it and didn't find any issues.

0nko (approved):
LGTM

app/src/main/java/com/duckduckgo/app/pixels/remoteconfig/AndroidBrowserConfigFeature.kt
catalinradoiu: Commited

catalinradoiu:
@0nko @LukasPaczos I have pushed 2 more commits to make the feature flag enabled by default and move the `selectedTabObserver?.cancel()` call that was flaged in the comment above. I will merge the PR once the CI passes.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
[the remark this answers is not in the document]
catalinradoiu: Updated

PR #8195, opened 2026-04-03
Added pixels for users that haven't seen subscription paywall

nalcalag:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1213904259888438?focus=true

### Steps to test this PR
_Pre steps_
- [x] Apply patch on https://app.asana.com/1/137249556945/project/1209991789468715/task/1210448620621729
- [x] In line 123 of _PaywallNotSeenWorker.kt_, change the setInitialDelay to for example `.setInitialDelay(mapOf("d0" to 30L, "d3" to 40L, "d7" to 50L, "d14" to 60L, "d30" to 70L)[dayBucket] ?: 30L, TimeUnit.SECONDS)` so pixels are fired at 30 seconds, 40 seconds, etc.. instead waiting for days.

_Paywall Not Seen Pixels fired_
- [x] Uninstall any previous versions of staging app and do a fresh install
- [x] Verify in logcat that pixels are fired after the scheduled time of each work:
            1. `Pixel sent: 1 m_subscription_paywall_not_seen_d0_u with params: {appVersion=5.274.2, returning_user=true, privacy_dashboard_opened=false} {}`
            2. `Pixel sent: 2 m_subscription_paywall_not_seen_d3_u with params: {appVersion=5.274.2, returning_user=true, privacy_dashboard_opened=false} {}`
            3. `Pixel sent: 3 m_subscription_paywall_not_seen_d7_u with params: {appVersion=5.274.2, returning_user=true, privacy_dashboard_opened=false} {}`
            4. `Pixel sent: 4 m_subscription_paywall_not_seen_d14_u with params: {appVersion=5.274.2, returning_user=true, privacy_dashboard_opened=false} {}`
            5. `Pixel sent: 5 m_subscription_paywall_not_seen_d30_u with params: {appVersion=5.274.2, returning_user=true, privacy_dashboard_opened=false} {}`

_Privacy Dashboard Param (Optional)_
- [x] Fresh install again
- [x] Before all pixels are sent, visit the privacy dashboard
- [x] Check `privacy_dashboard_opened` pixel param is set to `true` after that in the remaining pixels

_Paywall Not Seen Pixels stop after visiting paywall_
- [x] Uninstall any previous versions of staging app and do a fresh install
- [x] Let a couple of pixels be fired
- [x] Visit the paywall from any of the entry points (Settings > Subscription)
- [x] Verify pixels for the remaining milestones are not sent

_Subscription not eligible_
- [x] Fresh install of a debug app (without patch)
- [x] Check pixels are not scheduled or sent

### No UI changes

---

> [!NOTE]
> **Medium Risk**
> Adds new WorkManager scheduling and persistent state to fire analytics pixels based on install-age milestones, which could impact background work behavior and pixel volume if misconfigured. Gated by a new remote feature toggle and eligibility/subscription checks, reducing blast radius.
> 
> **Overview**
> Adds new subscription paywall visibility metrics: a **first-time paywall impression** unique pixel with a `days_since_install` cohort bucket, plus **"paywall not seen"** unique pixels scheduled for day 1/4/8/15/31 after install.
> 
> Introduces `PaywallMetricsManager` + `PaywallMetricsDataStore` to persist whether the paywall was ever seen, which milestone pixels have already fired, and whether the Privacy Dashboard was ever opened (via `PrivacyDashboardOpenedPlugin`). A new `PaywallNotSeenScheduler` schedules one-shot WorkManager jobs, and `PaywallNotSeenWorker` fires the appropriate pixel only when the feature flag is enabled, the user is eligible, and not already subscribed.
> 
> Updates pixel definitions/params dictionary, extends `SubscriptionPixel`/`SubscriptionPixelSender` to send the new pixels, and adds `privacy-dashboard-api` dependency plus unit tests covering bucketing, scheduling, and worker firing guards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 660a57158d7e8b8646d5ae79a3cde8ca67eb8771. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenWorker.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenWorker.kt
@@ -121,10 +121,11 @@ class PaywallNotSeenScheduler @Inject constructor(
     private fun scheduleMilestones() {
         MILESTONES.forEach { (dayBucket, checkAfterDays) ->
             if (paywallMetricsManager.isNotSeenDayFired(dayBucket)) return@forEach
+            val delay = paywallMetricsManager.delayUntilMilestone(checkAfterDays) ?: return@forEach
 
             val request = OneTimeWorkRequestBuilder<PaywallNotSeenWorker>()
                 .setInputData(workDataOf(PaywallNotSeenWorker.KEY_DAY_BUCKET to dayBucket))
-                .setInitialDelay(paywallMetricsManager.delayUntilMilestone(checkAfterDays), TimeUnit.MILLISECONDS)
+                .setInitialDelay(delay, TimeUnit.MILLISECONDS)
                 .addTag(workTag(dayBucket))
                 .build()
 
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallMetricsManager.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallMetricsManager.kt
@@ -56,12 +56,13 @@ class PaywallMetricsManager @Inject constructor(
     fun markNotSeenDayFired(dayBucket: String) = paywallMetricsStore.markNotSeenDayFired(dayBucket)
 
     /**
-     * Returns the delay in milliseconds until [checkAfterDays] have elapsed since install.
-     * Returns 0 if that point is already in the past.
+     * Returns the delay in milliseconds until [checkAfterDays] have elapsed since install,
+     * or null if that point is already in the past (milestone should not be scheduled).
      */
-    fun delayUntilMilestone(checkAfterDays: Long): Long {
+    fun delayUntilMilestone(checkAfterDays: Long): Long? {
         val checkAtMillis = paywallMetricsStore.firstInstallTimestamp + TimeUnit.DAYS.toMillis(checkAfterDays)
-        return maxOf(0L, checkAtMillis - System.currentTimeMillis())
+        val delay = checkAtMillis - System.currentTimeMillis()
+        return if (delay < 0) null else delay
     }
 
     private fun getDayBucket(firstInstallTimestamp: Long): String {
--- a/subscriptions/subscriptions-impl/src/test/java/com/duckduckgo/subscriptions/impl/pixels/PaywallMetricsManagerTest.kt
+++ b/subscriptions/subscriptions-impl/src/test/java/com/duckduckgo/subscriptions/impl/pixels/PaywallMetricsManagerTest.kt
@@ -131,17 +131,17 @@ class PaywallMetricsManagerTest {
     }
 
     @Test
-    fun `delayUntilMilestone returns 0 when milestone is in the past`() {
+    fun `delayUntilMilestone returns null when milestone is in the past`() {
         whenever(store.firstInstallTimestamp).thenReturn(System.currentTimeMillis() - TimeUnit.DAYS.toMillis(10))
 
-        assertEquals(0L, manager.delayUntilMilestone(checkAfterDays = 4))
+        assertNull(manager.delayUntilMilestone(checkAfterDays = 4))
     }
 
     @Test
     fun `delayUntilMilestone returns positive delay when milestone is in the future`() {
         whenever(store.firstInstallTimestamp).thenReturn(System.currentTimeMillis())
 
-        assertTrue(manager.delayUntilMilestone(checkAfterDays = 4) > 0)
+        assertTrue(manager.delayUntilMilestone(checkAfterDays = 4)!! > 0)
     }
 
     @Test
--- a/subscriptions/subscriptions-impl/src/test/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenSchedulerTest.kt
+++ b/subscriptions/subscriptions-impl/src/test/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenSchedulerTest.kt
@@ -106,6 +106,41 @@ class PaywallNotSeenSchedulerTest {
     }
 
     @Test
+    fun `when all milestones are in the past then no workers are scheduled`() = runTest {
+        whenever(paywallMetricsManager.paywallEverSeen).thenReturn(false)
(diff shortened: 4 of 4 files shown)


nalcalag:
* **#8195** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8195?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8195?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/SubscriptionPixelSender.kt
```diff
@@ -143,8 +152,12 @@ class SubscriptionPixelSenderImpl @Inject constructor(
             ),
         )
 
-    override fun reportOfferScreenShown() =
+    override fun reportOfferScreenShown() {
+        paywallMetricsManager.recordFirstPaywallSeen()?.let { dayBucket ->
```
catalinradoiu, line 156: I’m wonderig if this should be recorded here, as the class name suggests that it’s job is to send pixels. I am not sure if it should also call the manager to update the shared prefs data.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenWorker.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    @Inject
+    lateinit var subscriptions: Subscriptions
+
+    override suspend fun doWork(): Result {
+        val dayBucket = inputData.getString(KEY_DAY_BUCKET) ?: return Result.failure()
+
+        if (!subscriptions.isEligible()) return Result.success()
+
+        if (paywallMetricsManager.paywallEverSeen || paywallMetricsManager.isNotSeenDayFired(dayBucket)) {
```
catalinradoiu, line 70: Shouldn’t we check if the user is already subscribed? Also what happens if the user has seen the paywall before we enable the feature flag? Wouldn’t we end up sending a pixel that the paywall was not seen 30+ days after install? (in case it’s an user that had the app for a while)

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenWorker.kt
```diff
@@ -0,0 +1,141 @@
...
+        if (paywallMetricsManager.paywallEverSeen) return
+
+        appCoroutineScope.launch(dispatcherProvider.io()) {
+            scheduleMilestones()
+        }
+    }
+
+    private fun scheduleMilestones() {
+        MILESTONES.forEach { (dayBucket, checkAfterDays) ->
+            if (paywallMetricsManager.isNotSeenDayFired(dayBucket)) return@forEach
```
catalinradoiu, line 123: Shouldn’t we check if the user is still eligible for that pixel? What I’m thinking about is that we might have users who had the app for a while and once we enable the feature flag for these pixel we will schedule some extra workers for the d0, d3 etc pixels to be sent, while the user is not really “eligible” for sending these pixels because the app was installed a while ago. Does this make sense?

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/SubscriptionPixelSender.kt
nalcalag: Good thinking @catalinradoiu. Initially, I was calling the manager from the ViewModel and passing `dayBucket` as an argument to `reportOfferScreenShown()`. However, since the pixels aren’t supposed to be permanent, I opted for a different approach: passing the bucket as a lambda so the persistence logic stays in the manager. This way, I avoided injecting the Paywall manager as a dependency into more classes than necessary, which felt like a better temporary solution.

That said, I agree with your point, so I’ll add a reminder in the project, and if we decide to keep the pixels permanently, I’ll refactor it accordingly

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenWorker.kt
nalcalag: @catalinradoiu good catch! I didn’t check whether the user is subscribed. This is an edge case, I assumed that if a user subscribed, they must have visited the paywall. However, some users might restore an existing subscription and never visit the paywall. I’ll add that check.

> Wouldn’t we end up sending a pixel that the paywall was not seen 30+ days after install?

Yes, we discussed this offline. There will be some initial ramp-up, so we’ll review metrics day by day rather than relying on longer timeframes since install at the beginning

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/pixels/PaywallNotSeenWorker.kt
nalcalag: @catalinradoiu eligibility checks are also added in `doWork()`, so by the time the pixel is fired, it should meet the criteria. Does that answer your question, or am I missing something?

catalinradoiu: I was rather thinking about the scenario where we enable the feature flag and the user had the app installed for a while, but we end up scheduling the workers for each milestone, when we actually need it only for one of them.

nalcalag: Schedulers are based on install time, so if a user has had the app for more than 30 days, all workers would run immediately. The eligibility check would prevent the pixel from firing, so it would still behave correctly.
That said, I see your point and I think we shouldn't even schedule and run those workers, so I’ve added logic to avoid scheduling and running those workers at all if the date is in the past 🫡

PR #8310, opened 2026-04-17
Treat "Always" idle option as 0 seconds (immediate)

YoussefKeyrouz:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212810093780571/task/1214086381926895?focus=true

### Description
The idle-return "Always" option was mapped to 1L second. With a 1s timeout, a background+foreground cycle shorter than one second wouldn't trigger the return-to-NTP behaviour, which doesn't match what the label 'Always' implies.
This PR updates it to be 0 seconds instead (as immediate as possible)

Important notes ⚠️:
- Existing testers who had the "Always" option selected will see "0 minutes" upon update. They should reselect "Always" for the setting to apply. A proper migration was deemed unnecessary because it complicates the code and since this is not released yet, it's only internal. This only impacts people who downloaded the older version of the build and selected "Always" as their setting.
- There is a very short duration (~200ms) after the app is put in the background where android delays running the onStop lifecycle callback. If the user re-foregrounds within that window, we never observe a background event. So if you are REALLY fast in reopening the app, android did not have time to inform the app of the background status. In this case the app will reopen as if it was never put in the background. Nothing we can do about that. 

### Steps to test this PR
- Enable the showNTPAfterIdleReturn remote feature flag
- Select "Always" in Settings -> General -> AFter Inactivity -> Choose an inactivity timer
- Load any website
- Put the app in the background and quickly open it again. Note the timing is tricky here because you should still wait around 200ms for android to put the app in the background. 
- The NTP should show up

---

> [!NOTE]
> **Medium Risk**
> Changes the semantics of the "Always" after-inactivity timeout from 1s to 0s, which affects when the app returns to the NTP on background/foreground and could alter behavior for existing saved preferences and related analytics.
> 
> **Overview**
> Treats the after-inactivity "Always" setting as **0 seconds (immediate)** instead of `1L` across settings defaults, UI labeling, and pixel mapping.
> 
> Updates associated unit tests to expect `0L` for the "Always" option and verifies the corresponding timeout-selection pixel pair is fired for `0L`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 051d890569e11e7e0ac103094ac86191e552619d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/generalsettings/showonapplaunch/FirstScreenHandler.kt
+++ b/app/src/main/java/com/duckduckgo/app/generalsettings/showonapplaunch/FirstScreenHandler.kt
@@ -98,7 +98,7 @@ class FirstScreenHandlerImpl @Inject constructor(
 
     companion object {
         const val DEFAULT_IDLE_THRESHOLD_SECONDS = 300L
-        val DEFAULT_IDLE_THRESHOLD_OPTIONS = listOf(1L, 60L, 300L, 600L, 1800L, 3600L, 43200L, 86400L)
+        val DEFAULT_IDLE_THRESHOLD_OPTIONS = listOf(0L, 60L, 300L, 600L, 1800L, 3600L, 43200L, 86400L)
 
         fun parseDefaultIdleThresholdSeconds(settingsJson: String?): Long? {
             if (settingsJson == null) return null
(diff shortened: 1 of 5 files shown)


YoussefKeyrouz:
* **#8310** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8310?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8310?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8177** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

aitorvs (approved):

PR #8315, opened 2026-04-17
Fix favicons not showing on the favorites widget

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1214072492090532?focus=true

### Description
We introduced a bug when fixing the favorites widget for Android 16, causing the generated favicon to be used in most cases instead of the actual favicon. This was due to using the domain instead of base host as the cache key.

Generated placeholder favicons are now stored in a separate folder to avoid overwriting the main browser one. The files are removed when widget is deleted.

### Steps to test this PR

- [ ] Add a few favorites
- [ ] Add favorites and search widget to home screen
- [ ] Favicons should be visible instead of the generated placeholders with letters

### UI changes
| Before  | After |
| ------ | ----- |
Screenshot_20260417_114914| Screenshot_20260417_114614|

---

> [!NOTE]
> **Medium Risk**
> Changes widget favicon resolution and introduces new on-disk caching/cleanup logic; risk is mainly around file persistence/heuristic deletion potentially impacting cached favicons on upgrade.
> 
> **Overview**
> Fixes favorites widget favicon selection by centralizing lookup/generation in a new `WidgetFaviconProvider`, prioritizing persisted real favicons, then cached widget placeholders, and only then generating/storing a new placeholder.
> 
> Widget placeholders are now stored in a dedicated cache directory (`FAVICON_WIDGET_PLACEHOLDERS_DIR`) with a `FileProvider` path entry, cleaned up when the widget is disabled, and the widget now derives the cache key using `baseHost` (instead of `domain`) to avoid mismatches.
> 
> Adds an instrumentation test covering the provider’s lookup/store behavior and includes a heuristic to delete stale widget-sized placeholders that were previously saved into the main `favicons` directory.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c26fd1411223c1600f780c65e9f7b1202ee543cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/widget/FavoritesWidgetItemFactory.kt
+++ b/app/src/main/java/com/duckduckgo/widget/FavoritesWidgetItemFactory.kt
@@ -20,23 +20,16 @@ import android.annotation.SuppressLint
 import android.appwidget.AppWidgetManager
 import android.content.Context
 import android.content.Intent
-import android.graphics.BitmapFactory
 import android.net.Uri
 import android.os.Bundle
 import android.view.View
 import android.widget.RemoteViews
 import android.widget.RemoteViewsService
 import androidx.core.content.FileProvider
-import androidx.core.graphics.drawable.toBitmap
 import androidx.core.net.toUri
 import com.duckduckgo.app.browser.BrowserActivity
 import com.duckduckgo.app.browser.R
-import com.duckduckgo.app.browser.favicon.FaviconPersister
-import com.duckduckgo.app.browser.favicon.FileBasedFaviconPersister.Companion.FAVICON_PERSISTED_DIR
-import com.duckduckgo.app.browser.favicon.FileBasedFaviconPersister.Companion.FAVICON_WIDGET_PLACEHOLDERS_DIR
-import com.duckduckgo.app.browser.favicon.FileBasedFaviconPersister.Companion.NO_SUBFOLDER
 import com.duckduckgo.app.global.DuckDuckGoApplication
-import com.duckduckgo.app.global.view.generateDefaultDrawable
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/duckduckgo/widget/FavoritesWidgetItemFactory.kt
```diff
@@ -125,38 +127,47 @@ class FavoritesWidgetItemFactory(
...
-            directory = FAVICON_PERSISTED_DIR,
-            subFolder = NO_SUBFOLDER,
-            domain = domain,
-        )
-        var uri: Uri? = null
-
-        if (existingFile != null) {
-            // found existing file on disk (favicon or placeholder) - use it without network call
-            uri = existingFile.getContentUri()
+        val persistedFile = faviconPersister.faviconFile(FAVICON_PERSISTED_DIR, NO_SUBFOLDER, domain)
```
marcosholgado: There's a lot going on here that should likely be in the favicon manager, like you are effectively deleting favicons, etc.

anikiki (approved):

app/src/main/java/com/duckduckgo/widget/FavoritesWidgetItemFactory.kt
landomen: I extracted the logic to a new provider class. I don't think it belongs in the FaviconManager as it's a bit too widget-specific. If you feel strongly it should be in the FaviconManager, I can follow up on this in a separate PR and open an API proposal.

PR #8391, opened 2026-04-30
Create DaxMessage composable with components for RMF messaging, Dax banner message and notify view

catalinradoiu:
Task/Issue URL:https://app.asana.com/1/137249556945/project/1211724162604201/task/1211659112661253?focus=true

### Description

Adds Compose-based implementations of the remote message card components (SmallMessage, MediumMessage, BigSingleActionMessage, BigTwoActionsMessage and PromoSingleActionMessage) to the design system, plus the underlying DaxMessage primitive and the DaxBannerMessage and NotifyMe variants.

DaxMessage is the shared primitive used by all message variants (except PromoSingleActionMessage) — title, body, optional top illustration slot, optional dismiss icon (`isDismissable` + `onDismiss`), and a typed `DaxMessageButtonRowScope` slot that restricts the bottom row to the approved button layouts (`RightAlignButtons`, `CenterAlignedButtons`, `FullWidthSingleButton`, `SmallSingleButton`).

The new composables support static image illustrations via Painter, remote images via Coil's async painter, and animated Lottie illustrations. DaxCard now accepts optional shape and elevation parameters (typed as `DaxCardElevation` to keep Material3 types out of the public API).

The lottie-compose dependency has been added to support Lottie animations in Compose contexts. The design system internal showcase has been updated to display all new Compose message variants alongside the existing View-based ones, including an example using a Lottie animation and one using a server-hosted image.

### Steps to test this PR

_Compose Remote Message Components_

- [ ] Open the design system internal app and navigate to the messaging component section
- [ ] Verify that all Compose message variants are displayed: Small, Medium, Big Single Action (static illustration), Big Single Action (Lottie animation), Big Two Actions (static illustration), Big Two Actions (app update illustration), Big Two Actions (server image) and Promo Single Action
- [ ] Verify that the server image loads correctly in the Big Two Actions server image variant, and that the fallback illustration is shown if the image fails to load
- [ ] Verify all variants render correctly in both light and dark themes
- [ ] Check the previews for DaxBannerMessage and NotifyMe

### UI changes

No UI changes. This PR adds the compose variants for the DaxMessag, which can be visualised in the ADS Preview Screen. 

---

> [!NOTE]
> **Medium Risk**
> Introduces new public Compose UI components and adjusts `DaxCard`/`DaxSurface` APIs (including a default shape change for `DaxSurface`), which could impact downstream consumers’ visuals and binary compatibility.
> 
> **Overview**
> Adds a new `DaxMessage` Compose primitive (typed action-row variants, optional top content, and optional dismiss control) plus new message components built on it: `DaxBannerMessage`, `DaxNotifyMe`, and Compose-based Remote Message card variants (`DaxSmallMessage`, `DaxMediumMessage`, `DaxBigSingleActionMessage`, `DaxBigTwoActionsMessage`, `DaxPromoSingleActionMessage`).
> 
> Updates `DaxCard` to accept configurable `shape` and a simplified `DaxCardElevation`, and updates `DaxSurface` to accept `shape`/`shadowElevation` while changing its default shape to `RectangleShape` (callers now pass explicit shapes where needed, e.g. `DaxAlertDialog`).
> 
> Enhances the internal design-system showcase to render the new Compose message variants alongside existing view-based ones, including examples using a bundled Lottie animation (`anim_password_keys.json`) and a server-loaded illustration via Coil; adds `lottie-compose` dependency and related version pin, plus new dimension resources for remote/promo message sizing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14ac3da9954b329b60f8d5e0d207d72e8b83ebe0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


catalinradoiu:
* **#8391** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8391?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8391?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/RemoteMessage.kt
```diff
@@ -0,0 +1,210 @@
...
+                topContent()
+                DaxText(
+                    text = title,
+                    style = DuckDuckGoTheme.typography.h3,
+                    textAlign = TextAlign.Center,
+                    modifier = Modifier
+                        .padding(
+                            top = dimensionResource(R.dimen.keyline_2),
+                            start = 40.dp,
+                            end = 40.dp,
```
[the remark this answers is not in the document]
catalinradoiu: Fixed

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/PromoSingleActionMessage.kt
```diff
@@ -0,0 +1,164 @@
...
+                    .size(dimensionResource(R.dimen.messageCtaCloseButtonSize))
+                    .clip(CircleShape)
+                    .clickable(
+                        onClick = onDismissed,
+                        interactionSource = remember { MutableInteractionSource() },
+                        indication = ripple(bounded = true),
+                    )
+                    .padding(
+                        horizontal = dimensionResource(R.dimen.keyline_2),
+                    ),
```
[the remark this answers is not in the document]
catalinradoiu: Fixed

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/RemoteMessage.kt
```diff
@@ -0,0 +1,152 @@
...
+                    painter = topIllustration,
+                    contentDescription = null,
+                    contentScale = ContentScale.Fit,
+                    modifier = Modifier
+                        .padding(
+                            top = dimensionResource(R.dimen.keyline_3),
+                            bottom = dimensionResource(R.dimen.keyline_2),
+                        )
+                        .align(Alignment.Center)
+                        .size(48.dp),
```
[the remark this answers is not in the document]
catalinradoiu: fixed

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+                    modifier = Modifier
+                        .fillMaxWidth(),
+                    horizontalAlignment = alignment,
+                ) {
+                    DaxText(
+                        text = title,
+                        style = DuckDuckGoTheme.typography.h3,
+                        textAlign = textAlign,
+                        modifier = Modifier
+                            .padding(top = dimensionResource(R.dimen.keyline_2))
```
mikescamell: 🔍 Recommend to add `Spacer` instead of padding for lists in general. Sometimes padding make sense but in general this is the "old" View way of doing things https://www.jetpackcompose.app/articles/space-hoisting-should-I-use-the-padding-modifier-or-the-spacer-composable

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+                            .fillMaxWidth()
+                            .align(alignment),
+                    )
+                    DaxText(
+                        text = body,
+                        style = DuckDuckGoTheme.typography.body2,
+                        textAlign = textAlign,
+                        modifier = Modifier
+                            .padding(top = dimensionResource(R.dimen.keyline_1))
```
mikescamell: 🔍 Recommend to add Spacer instead of padding for lists in general. Sometimes padding make sense but in general this is the "old" View way of doing things https://www.jetpackcompose.app/articles/space-hoisting-should-I-use-the-padding-modifier-or-the-spacer-composable

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+                            .align(alignment),
+                    )
+                    if (buttonRow != null) {
+                        Box(
+                            modifier = Modifier
+                                .fillMaxWidth()
+                                .padding(top = dimensionResource(R.dimen.keyline_4)),
```
mikescamell: 🔍 Recommend to add Spacer instead of padding for lists in general. Sometimes padding make sense but in general this is the "old" View way of doing things https://www.jetpackcompose.app/articles/space-hoisting-should-I-use-the-padding-modifier-or-the-spacer-composable

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+                                ContentAlignment.Center -> Alignment.Center
+                            },
+                        ) {
+                            DaxMessageButtonRowScopeImpl.buttonRow()
+                        }
+                    }
+                }
+            }
+            if (isDismissable) {
+                Icon(
```
mikescamell: ⚠️Can we use DaxIconButton here instead?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+                            indication = ripple(bounded = true),
+                        )
+                        .padding(dimensionResource(R.dimen.keyline_2)),
+                )
+            }
+        }
+    }
+}
+
+enum class ContentAlignment {
```
mikescamell: 🔍Maybe we can be more specific like `DaxMessageContentAlignment`? WDYT?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/cards/DaxSurface.kt
```diff
@@ -47,14 +47,16 @@ import com.duckduckgo.common.ui.compose.tools.PreviewBox
 fun DaxSurface(
     modifier: Modifier = Modifier,
     border: BorderStroke? = null,
+    shape: Shape = DaxSurfaceDefaults.shape,
```
mikescamell, line 51: 🔍 I think we should default to RectangleShape like the Material3 impl, unless I'm missing a reason why we need this rounded. I appreciate this is not part of your PR so ok to leave it for now 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/RemoteMessage.kt
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @param title The message title.
+ * @param body The message body text.
+ * @param topIllustration Composable slot rendered above the title.
+ * @param onDismissClicked Called when the user taps the dismiss button.
+ * @param modifier Modifier for this message card.
+ * @param bottomContent Composable slot rendered below the body — typically action buttons.
+ */
+@Composable
+internal fun RemoteMessageWithIllustration(
```
mikescamell: 🔍Can we generate previews for these even if we have the messages that sue them? That way people have some idea of how these look

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/PromoSingleActionMessage.kt
```diff
@@ -0,0 +1,162 @@
...
+                    )
+                    .padding(dimensionResource(R.dimen.keyline_2)),
+            )
+            DaxText(
+                text = title,
+                style = DuckDuckGoTheme.typography.h2,
+                textAlign = TextAlign.Center,
+                modifier = Modifier
+                    .padding(
+                        top = dimensionResource(R.dimen.keyline_2),
```
mikescamell: 🔍 Recommend to add Spacer instead of padding for lists in general. Sometimes padding make sense but in general this is the "old" View way of doing things https://www.jetpackcompose.app/articles/space-hoisting-should-I-use-the-padding-modifier-or-the-spacer-composable

I won't comment on all paddings in this composable, but we should check through 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+fun DaxMessage(
+    title: String,
+    body: String,
+    modifier: Modifier = Modifier,
+    shape: Shape = RoundedCornerShape(0.dp),
+    elevation: Dp = 0.dp,
+    contentAlignment: ContentAlignment = ContentAlignment.Center,
+    topContent: (@Composable () -> Unit)? = null,
+    buttonRow: (@Composable DaxMessageButtonRowScope.() -> Unit)? = null,
+    isDismissable: Boolean = false,
```
mikescamell: 💡The more I see this the more i think we should just go with your idea of a nullable onDismiss callback, if someone registers it, they get the X, if they don't then we do not show it 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+import com.duckduckgo.common.ui.compose.button.DaxPrimaryButton
+import com.duckduckgo.common.ui.compose.cards.DaxCard
+import com.duckduckgo.common.ui.compose.cards.DaxCardElevation
+import com.duckduckgo.common.ui.compose.text.DaxText
+import com.duckduckgo.common.ui.compose.theme.DuckDuckGoTheme
+import com.duckduckgo.common.ui.compose.tools.PreviewBox
+import com.duckduckgo.mobile.android.R
+
+@Composable
+fun DaxMessage(
```
mikescamell: 🔍 Are we missing kdoc for this?

mikescamell: 🔍Also should we make this internal for now? Does anything else actually use it in the View world in general or is it always built on top of?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+    title: String,
+    body: String,
+    modifier: Modifier = Modifier,
+    shape: Shape = RoundedCornerShape(0.dp),
+    elevation: Dp = 0.dp,
+    contentAlignment: ContentAlignment = ContentAlignment.Center,
+    topContent: (@Composable () -> Unit)? = null,
+    buttonRow: (@Composable DaxMessageButtonRowScope.() -> Unit)? = null,
```
mikescamell: 💡Could we provide a typed Parameter for this, a bit like the button enums?

E.g.

```kotlin
sealed interface DaxMessageActions {
    data class RightAlign(val primary: DaxAction, val secondary: DaxAction) : DaxMessageActions
    data class CenterAligned(val primary: DaxAction, val secondary: DaxAction) : DaxMessageActions
    data class FullWidthSingle(val primary: DaxAction) : DaxMessageActions
    data class SmallSingle(val primary: DaxAction) : DaxMessageActions
}

@Composable
internal fun DaxMessage(
    title: String,
    body: String,
    modifier: Modifier = Modifier,
    actions: DaxMessageActions? = null,   // null = no actions
    // ...
)
```

And then in future if people really need to add something different in the actions row, they can ask for it to be opened up for anything adn we add a scope? We could then kill the lint decection, the custom scope etc. WDYT?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/BigSingleActionMessage.kt
```diff
@@ -0,0 +1,130 @@
...
+ * @param body The message body text.
+ * @param topIllustration The illustration shown above the title. Use [painterResource]
+ *   for a local drawable, or `coil3.compose.rememberAsyncImagePainter` for a remote URL.
+ * @param actionText The label of the primary action button.
+ * @param onActionClick Called when the user taps the primary action.
+ * @param onDismissed Called when the user taps the dismiss button.
+ * @param modifier Modifier for this message card.
+ */
+@Composable
+fun BigSingleActionMessage(
```
mikescamell: 🔍I think **all** the new message composables you've added should be prefixed with "Dax" to be consistent with every other design system composable

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+
+@Composable
+fun DaxMessage(
+    title: String,
+    body: String,
+    modifier: Modifier = Modifier,
+    shape: Shape = RoundedCornerShape(0.dp),
+    elevation: Dp = 0.dp,
+    contentAlignment: ContentAlignment = ContentAlignment.Center,
+    topContent: (@Composable () -> Unit)? = null,
```
mikescamell: 🔍I think we should add ColumnScope for consumers e.g. `@Composable ColumnScope.() -> Unit`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/RemoteMessage.kt
```diff
@@ -0,0 +1,152 @@
...
+ * @param topIllustration Composable slot rendered above the title — typically an illustration.
+ * @param bottomActions Composable slot rendered below the body — typically action buttons.
+ */
+@Composable
+internal fun RemoteMessage(
+    title: String,
+    body: String,
+    onDismissClicked: () -> Unit,
+    modifier: Modifier = Modifier,
+    topIllustration: (@Composable () -> Unit)? = null,
```
mikescamell: 🔍I think we should add ColumnScope for consumers e.g. `@Composable ColumnScope.() -> Unit`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/BigSingleActionMessage.kt
```diff
@@ -0,0 +1,130 @@
...
+    title: String,
+    body: String,
+    topIllustration: @Composable () -> Unit,
```
mikescamell: 🔍I think we should add ColumnScope for consumers e.g. `@Composable ColumnScope.() -> Unit`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: Yes, I think it is a better approach. initially I wanted to use slots for that but I think that isn’t really a requirements and as you said, we can always open the API in case it’s needed.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,341 @@
...
+    buttonRow: (@Composable DaxMessageButtonRowScope.() -> Unit)? = null,
+    isDismissable: Boolean = false,
+    onDismiss: () -> Unit = {},
```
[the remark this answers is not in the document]
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/BigSingleActionMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/BigSingleActionMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: done

catalinradoiu: added

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: I changed and we’re now using the nullable callback

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/PromoSingleActionMessage.kt
catalinradoiu: updated

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/RemoteMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/cards/DaxSurface.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
catalinradoiu: done

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/NotifyMe.kt
```diff
@@ -0,0 +1,79 @@
...
+ * primary/secondary action pair, and a rounded, elevated card surface.
+ *
+ * @param title The message title.
+ * @param body The message body text.
+ * @param primaryAction The primary call-to-action (e.g. "Notify me").
+ * @param secondaryAction The secondary ghost action (e.g. "Not now").
+ * @param modifier Modifier for this message card.
+ */
+@Composable
+fun NotifyMe(
```
mikescamell: 🔍 This one wasn't renamed. The `done` reply to https://github.com/duckduckgo/Android/pull/8391#discussion_r3207857016 covered the six `remote/` variants but `NotifyMe` stayed as is. To match the "all new message composables should be Dax-prefixed" rule, this should be `DaxNotifyMe` (and the filename `NotifyMe.kt` rename to `DaxNotifyMe.kt`).

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/NotifyMe.kt
```diff
@@ -0,0 +1,79 @@
...
+        actions = DaxMessageActions.RightAligned(
+            primary = primaryAction,
+            secondary = secondaryAction,
+        ),
+    )
+}
+
+@PreviewLightDark
+@Composable
+private fun DaxBannerMessagePreview() {
```
mikescamell: 🔍 The preview function is named `DaxBannerMessagePreview` but it previews `NotifyMe`. Looks like a copy-paste from `DaxBannerMessage.kt`. Should be `NotifyMePreview` (or `DaxNotifyMePreview` if you rename the composable per the other comment).

mikescamell (commented):
Re-review follow-ups. The structural changes are great (typed `DaxMessageActions`, `internal` `DaxMessage`, lint rule deletion). Two items from the renaming pass that look like they slipped through:

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/NotifyMe.kt
```diff
@@ -0,0 +1,79 @@
...
+    secondaryAction: DaxAction,
+    modifier: Modifier = Modifier,
+) {
+    DaxMessage(
+        title = title,
+        body = body,
+        elevation = dimensionResource(R.dimen.keyline_1),
+        shape = DuckDuckGoTheme.shapes.large,
+        contentAlignment = DaxMessageContentAlignment.Start,
+        modifier = modifier.padding(dimensionResource(R.dimen.keyline_4)),
```
mikescamell: 💡 `modifier.padding(...)` silently appends padding to the caller's modifier. Per the design system guidelines (Section 3, modifier rules), the caller's modifier should be applied as is on the root layout, with any internal padding owned by the composable itself. Consider wrapping the underlying `DaxMessage` in a `Box` that owns the keyline_4 padding, or moving the padding inside `DaxMessage`. Same pattern in `DaxRemoteMessage.kt:150` and `DaxPromoSingleActionMessage.kt:73`.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxRemoteMessage.kt
```diff
@@ -0,0 +1,179 @@
...
+    topIllustration: (@Composable ColumnScope.() -> Unit)? = null,
+    actions: DaxMessageActions? = null,
+) {
+    DaxMessage(
+        title = title,
+        body = body,
+        elevation = dimensionResource(R.dimen.keyline_1),
+        topContent = topIllustration?.let { { it() } },
+        onDismiss = onDismissClicked,
+        modifier = modifier.padding(dimensionResource(R.dimen.keyline_4)),
```
mikescamell: 💡 Same `modifier.padding` mutation antipattern as flagged on `NotifyMe.kt:54`. The caller's modifier should arrive on the root layout untouched.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxPromoSingleActionMessage.kt
```diff
@@ -0,0 +1,142 @@
...
+    title: String,
+    body: String,
+    illustration: Painter,
+    actionText: String,
+    onActionClick: () -> Unit,
+    onDismissed: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    DaxCard(
+        modifier = modifier.padding(
```
mikescamell: 💡 Same `modifier.padding` mutation antipattern as flagged on `NotifyMe.kt:54`. The caller's modifier should arrive on the root `DaxCard` untouched.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxRemoteMessage.kt
```diff
@@ -0,0 +1,179 @@
...
+    onDismissClicked: () -> Unit,
+    modifier: Modifier = Modifier,
+    topIllustration: (@Composable ColumnScope.() -> Unit)? = null,
+    actions: DaxMessageActions? = null,
+) {
+    DaxMessage(
+        title = title,
+        body = body,
+        elevation = dimensionResource(R.dimen.keyline_1),
+        topContent = topIllustration?.let { { it() } },
```
mikescamell: 🔍 `topIllustration?.let { { it() } }` is a redundant lambda re-wrap. Both sides are `@Composable ColumnScope.() -> Unit`, so this can just be `topContent = topIllustration`.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxBigTwoActionsMessage.kt
```diff
@@ -0,0 +1,92 @@
...
+    )
+}
+
+@PreviewLightDark
+@Composable
+private fun DaxBigTwoActionsMessagePreview() {
+    PreviewBox {
+        DaxBigTwoActionsMessage(
+            title = "Big Two Actions Message",
+            body = "Body text goes here. This component has two buttons and showcases and app update",
```
mikescamell: 🔍 Preview body typo: `showcases and app update` should be `showcases an app update`. Same typo in `DaxPromoSingleActionMessage.kt:135`.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,343 @@
...
+
+    /** A single primary button that fills the available width. */
+    data class FullWidthSingle(val primary: DaxAction) : DaxMessageActions
+
+    /** A single small primary button. */
+    data class SmallSingle(val primary: DaxAction) : DaxMessageActions
+}
+
+/** A single button entry used by [DaxMessageActions]. */
+data class DaxAction(
```
mikescamell, line 187: 🔍 Small perf hint: consider annotating `DaxAction` with `@Immutable` and `DaxMessageContentAlignment` with `@Stable`. The Compose compiler can't infer stability for a class containing a function reference (`onClick`), so without `@Immutable`, `DaxAction` is treated as unstable, which can trigger unnecessary recompositions of `DaxMessage` when the parent recomposes.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxPromoSingleActionMessage.kt
```diff
@@ -0,0 +1,142 @@
...
+                text = title,
+                style = DuckDuckGoTheme.typography.h2,
+                textAlign = TextAlign.Center,
+                modifier = Modifier
+                    .padding(horizontal = dimensionResource(R.dimen.keyline_4))
+                    .fillMaxWidth(),
+            )
+            Image(
+                painter = illustration,
+                contentDescription = null,
```
mikescamell: 🔍 `contentDescription` is hardcoded to `null`, which makes the image decorative for accessibility purposes. If the promo illustration carries meaning (most marketing art does), this should be exposed as a parameter so the caller can provide an accessible description.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxPromoSingleActionMessage.kt
```diff
@@ -0,0 +1,142 @@
...
+            Spacer(modifier = Modifier.height(dimensionResource(R.dimen.keyline_2)))
+            DaxText(
+                text = title,
+                style = DuckDuckGoTheme.typography.h2,
+                textAlign = TextAlign.Center,
+                modifier = Modifier
+                    .padding(horizontal = dimensionResource(R.dimen.keyline_4))
+                    .fillMaxWidth(),
+            )
+            Image(
```
mikescamell, line 118: 🔍 This `Image` has no size modifier, so it renders at the painter's intrinsic size. For local drawables that's fine, but the KDoc explicitly mentions `coil3.compose.rememberAsyncImagePainter` for remote URLs, where the intrinsic size is whatever the server returns. A remote image at unexpected dimensions could break the card layout. Consider a `Modifier.height(...)` or `.size(...)` constraint, or document a recommended aspect ratio.

mikescamell (commented):
Medium-severity follow-ups from the re-review. None of these block merge, but worth a sweep before landing.

mikescamell (commented):
Two more small follow-ups.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/DaxMessage.kt
```diff
@@ -0,0 +1,343 @@
...
+    val text: String,
+    val onClick: () -> Unit,
+)
+
+private object DaxMessageDefaults {
+    val shape: Shape = RoundedCornerShape(0.dp)
```
mikescamell: 🔍 `RoundedCornerShape(0.dp)` could just be `RectangleShape` 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxBigSingleActionMessage.kt
```diff
@@ -0,0 +1,126 @@
...
+ * @param onActionClick Called when the user taps the primary action.
+ * @param onDismissed Called when the user taps the dismiss button.
+ * @param modifier Modifier for this message card.
+ */
+@Composable
+fun DaxBigSingleActionMessage(
+    title: String,
+    body: String,
+    topIllustration: @Composable ColumnScope.() -> Unit,
+    actionText: String,
```
mikescamell: ❓ Quick question: was it intentional to expose flat `actionText`/`onActionClick` here (and the four-action equivalent in `DaxBigTwoActionsMessage`) instead of accepting `DaxAction`? The rest of the family uses `DaxAction` internally,  but maybe we wanted to intentionally leave this as is?

catalinradoiu: Yes, I wanted to keep that internally only and externally use a similar API to what we have for the button for example. Let me know if you think we should switch to using the DaxAction directly. I can make that class internal oterwise.

catalinradoiu: I looked over it again and I am going to use `DaxAction` instead. I like it’s better, especially when we have 2 actions and they are wrapped inside a single object instead of being 2 different parameters.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxPromoSingleActionMessage.kt
catalinradoiu: I will add `Modifier.sizeIn` constraint to it.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/NotifyMe.kt
catalinradoiu: I will put the padding inside the caller. I saw this initially being on the style applied on the view version of the component but this pattern doesn’t make sense in compose. I will let the caller decide what spacing the component needs to use.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/cards/DaxSurface.kt
```diff
@@ -99,11 +104,10 @@ fun DaxSurface(
...
 private object DaxSurfaceDefaults {
     /**
-     * The default shape for Dax surfaces is the medium shape from the DuckDuckGo theme, which provides a balanced appearance that works well for card-like components.
+     * The default shape for Dax surfaces is rectangular, matching the Material3 [Surface] default.
+     * Callers that want a rounded surface should pass an explicit [Shape].
      */
-    val shape: Shape
-        @Composable
-        get() = DuckDuckGoTheme.shapes.medium
+    val shape: Shape = RectangleShape
```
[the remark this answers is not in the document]
catalinradoiu: Fixed

catalinradoiu:
@mikescamell I have fixed the comnments in this PR.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxRemoteMessage.kt
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * @param title The message title.
+ * @param body The message body text.
+ * @param topIllustration Composable slot rendered above the title.
+ * @param onDismissClicked Called when the user taps the dismiss button.
+ * @param modifier Modifier for this message card.
+ * @param actions Optional action row.
+ */
+@Composable
+internal fun DaxRemoteMessageWithIllustration(
```
mikescamell: 🔍 This overload looks like a pure pass-through to `DaxRemoteMessage` on line 136. Unlike the sibling `Painter` overload on line 87, it doesn't add any illustration-specific wrapping. The `topIllustration = {}` default also bumps into the nullable-slot rule from the guidance (Section 6) since `DaxRemoteMessage` already exposes `topIllustration: ... = null`. WDYT about deleting this overload and routing `DaxBigSingleActionMessage`'s slot variant directly to `DaxRemoteMessage`?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxRemoteMessage.kt
```diff
@@ -0,0 +1,181 @@
...
+    modifier: Modifier = Modifier,
+    actions: DaxMessageActions? = null,
+) {
+    DaxRemoteMessage(
+        title = title,
+        body = body,
+        onDismissClicked = onDismissClicked,
+        modifier = modifier,
+        topIllustration = {
+            Box(modifier = Modifier.fillMaxWidth()) {
```
mikescamell: 🔍 Small layout cleanup: this `Box(fillMaxWidth)` only exists to enable `Alignment.Center` on the `Image`. Since the parent receiver is `ColumnScope`, you can drop the `Box` and apply `Modifier.align(Alignment.CenterHorizontally)` directly on the `Image` (combined with the existing modifier chain). One less node in the layout tree, same visual result.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxBigSingleActionMessage.kt
```diff
@@ -0,0 +1,117 @@
...
+ * @param title The message title.
+ * @param body The message body text.
+ * @param topIllustration Composable slot rendered above the title. Receives a [ColumnScope]
+ *   so the content can use `Modifier.align(...)` if needed.
+ * @param action The primary action button.
+ * @param onDismissed Called when the user taps the dismiss button.
+ * @param modifier Modifier for this message card.
+ */
+@Composable
+fun DaxBigSingleActionMessage(
```
mikescamell, line 52: 📚 Heads up: We should try to include Asana references in KDocs. `DaxBannerMessage` and `DaxNotifyMe` have them, but the entire `remote/` package is missing them. Could you add links to:

- `DaxBigSingleActionMessage` (both overloads, here)
- `DaxBigTwoActionsMessage`
- `DaxMediumMessage`
- `DaxPromoSingleActionMessage`
- `DaxSmallMessage`

See DaxButton in Asana for how I added each variant to it's own subtask

mikescamell (commented):
Got a fresh pair of eyes on the branch. A couple of items they spotted that I missed in my earlier passes.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/message/remote/DaxMediumMessage.kt
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * DuckDuckGo medium remote message card.
+ *
+ * Variant with a top illustration, title, body, and a dismiss button. No actions.
+ *
+ * @param title The message title.
+ * @param body The message body text.
+ * @param topIllustration The illustration shown above the title. Use [painterResource]
+ *   for a local drawable, or `coil3.compose.rememberAsyncImagePainter` for a remote URL.
```
mikescamell, line 37: 📚 Small KDoc polish: the hint to use `rememberAsyncImagePainter` is helpful, but it stops short of showing callers that for a remote URL they almost certainly want to pass `error` and `fallback` painters too. The showcase at `ComponentViewHolder.kt:338-342` already does this:

```kotlin
topIllustration = rememberAsyncImagePainter(
    model = "https://...",
    error = painterResource(...),
    fallback = painterResource(...),
)
```

Worth mirroring (or linking to) in the KDoc here and in the other Coil-aware variants (`DaxBigSingleActionMessage`, `DaxBigTwoActionsMessage`, `DaxPromoSingleActionMessage`). Without it, callers will hit broken images the first time a CDN burps.

mikescamell (commented):
One more KDoc polish that came out of the fresh review.

mikescamell (approved):
Nice work! LGTM 🚢

PR #8439, opened 2026-05-05
Onboarding Brand Design Update: Add in-context SERP dialog

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207908166761516/task/1212699268790179?focus=true

### Description

Migrates the SERP contextual onboarding dialog ("DuckDuckGo Search!") to the new brand-design layout, gated behind the `onboardingBrandDesignUpdate` feature flag. Lays the foundation (base class, include layouts, dismiss button styling, banner-animation pipeline, rotation handling) that the remaining contextual CTAs will plug into in follow-up PRs.

**With the flag on:**
- SERP CTA renders in the new card chrome with title and body as distinct blocks (legacy was a single inline-bolded sentence).
- SERP background banner slides up from behind the card on appear and slides out on dismiss / content swap.
- New circular bordered dismiss button plus a two-layer drop shadow at the bottom edge so the dialog reads as a layer above the browser content.

**Out of scope** (still legacy / stub-only with the flag on): fire-button, no-trackers, main-network, trackers-blocked, site-suggestions, end. `Dax*BrandDesignUpdateContextualCta` classes exist as scaffolding for follow-ups.

#### Why all the rotation-handling code

`BrowserActivity` declares `android:configChanges="keyboardHidden|orientation|screenSize|smallestScreenSize|screenLayout|navigation|keyboard"` in `AndroidManifest.xml`, so Android does **not** recreate the activity on orientation changes — `onConfigurationChanged` fires and the existing view tree stays in place. The system never auto-swaps the `layout-land` variant in for us, so we have to do it manually. That's why this PR introduces:

- `Command.ReinflateBrandDesignContextualDialog` emitted from the VM on orientation change (only when the brand-design flag is enabled).
- `BrowserTabFragmentRenderer.reinflateContextualBrandDesignDialog()` which inflates the new orientation's layout, transplants its children into the existing root, and re-shows the CTA via a new `instantShow` snap path so it lands fully populated (no replay of the typing animation).
- A landscape variant of `include_onboarding_in_context_dax_dialog_brand_design_update.xml` plus orientation-keyed `bools.xml` for the options-content height cap.
- Coordination with the animation pipeline: running animators are cancelled before re-inflation, a visibility guard prevents resurrecting a dismissed CTA, and the `onCtaShown` pixel is gated so it doesn't double-fire on rotation re-inflate.

#### Other notes

- Dev-settings fix so the bulk "Onboarding Completed" toggle clears `DAX_DIALOG_NETWORK` / `DAX_DIALOG_OTHER` flags — without it, iterative testing of the contextual stack silently fails to re-show the in-context site-suggestions dialog after reset.
- `cancelRunningAnimations()` added to both `BrandDesignUpdateBubbleCta` (pre-existing, same gap on develop) and the new contextual base class — running fade-in/fade-out `AnimatorSet`s and the container's `ViewPropertyAnimator` now get cancelled on dismiss so their listeners can't fire on a `.gone()` view.
- Telemetry parity: pixel names, `ctaPixelParam`, and `ctaId` match the legacy `DaxSerpCta` exactly — no new pixels, no renamed params.

### Steps to test this PR

[Designs](https://www.figma.com/design/YPE94Xkcrk2uqiF2l4VmSv/Onboarding--2026-?node-id=16515-30762&m=dev)

Please run all [testing steps ](https://app.asana.com/1/137249556945/project/1207908166761516/task/1214549026899402?focus=true)for in-context dialog changes from the contextual-end branch/PR to ease the testing burden

To see these changes patch (Linear onboarding flag included just for continuity)

```
Index: app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
IDEA additional info:
<+>UTF-8
--- a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt	(revision 6fd565c8894daba16281ae9bcf3452d038ae8d6d)
+++ b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt	(date 1777967047929)
@@ -34,13 +34,13 @@
      * Main toggle for the onboarding brand design update feature.
      * Default value: false (disabled).
      */
     fun self(): Toggle
 
     /**
      * Toggle for the brand design update variant.
      * Default value: false (disabled).
      */
     fun brandDesignUpdate(): Toggle
 }
Index: app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
IDEA additional info:
<+>UTF-8
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt b/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt	(revision 6fd565c8894daba16281ae9bcf3452d038ae8d6d)
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt	(date 1777657893440)
@@ -45,7 +45,7 @@
     val viewState = _viewState.asStateFlow()
 
     fun initializePages() {
-        pageLayoutManager.buildPageBlueprints()
+        pageLayoutManager.buildBrandDesignUpdatePageBlueprints()
     }
 
     fun pageCount(): Int {
@@ -69,8 +69,8 @@
         }
     }
 
-    fun initializeOnboardingSkipper() {
-        if (!appBuildConfig.canSkipOnboarding) return
+    fun initializeOnboardingSkipper() { 
+        return
 
         // delay showing skip button until privacy config downloaded
         viewModelScope.launch {

```

### UI changes

See [here](https://app.asana.com/1/137249556945/project/1207908166761516/task/1214495557648329?focus=true)

---

> [!NOTE]
> **Medium Risk**
> Changes onboarding CTA rendering and lifecycle in `BrowserTabFragment`/`CtaViewModel`, including new animation/cancelation and orientation-handling paths that could cause regressions in CTA visibility, pixel firing, or stuck animations.
> 
> **Overview**
> Adds a new *brand-design* in-context onboarding dialog system (`BrandDesignContextualDaxDialogCta`) with a dedicated layout (portrait + `layout-land`), banner/shadow/dismiss-button styling, and a full animation pipeline (fade/typing/slide banner) with explicit animation cancellation and tap-to-skip snapping.
> 
> Migrates the SERP contextual CTA to this new UI when the `onboardingBrandDesignUpdate` toggle is enabled (`DaxSerpBrandDesignUpdateContextualCta`), adds brand-design title/body strings across locales, and wires the fragment/viewmodel to render/hide brand-design CTAs alongside legacy ones.
> 
> Introduces orientation-change handling for the new contextual dialog: `BrowserTabViewModel` detects orientation changes and emits `Command.ReinflateBrandDesignContextualDialog`, and `BrowserTabFragmentRenderer.reinflateContextualBrandDesignDialog()` reinflates/transplants the updated layout and re-shows the CTA via an `instantShow` path to avoid replaying animations/pixels.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad9f7767ab23858cf72ccb5c0df4ea7dcc738721. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mikescamell:
* **#8482** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8482?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8481** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8481?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8480** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8480?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8479** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8479?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8483** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8483?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8439** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8439?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4821,7 +4874,8 @@ class BrowserTabViewModel @Inject constructor(
             pixel.fire(DuckChatPixelName.DUCK_CHAT_FIRE_BUTTON_TAPPED)
         }
         val cta = currentCtaViewState().cta
-        if (cta is OnboardingDaxDialogCta.DaxFireButtonCta) {
+        if (cta is OnboardingDaxDialogCta.DaxFireButtonCta || cta is DaxFireButtonBrandDesignUpdateContextualCta) {
+            // TODO: replace in stage 2 (DaxFireButtonCta brand-design migration).
```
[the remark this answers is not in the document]
mikescamell: This is safe today — both arms of the `||` extend `OnboardingDaxDialogCta`, and `CtaViewModel.getFireDialogCta()` still returns the legacy `DaxFireButtonCta` even with the brand-design flag on (see the `// TODO: replace in stage 2` markers in `CtaViewModel.kt`), so the brand-design arm is unreachable in this PR.

app/src/main/java/com/duckduckgo/app/cta/ui/DaxFireButtonBrandDesignUpdateContextualCta.kt
```diff
@@ -0,0 +1,54 @@
...
+    shownPixel = AppPixelName.ONBOARDING_DAX_CTA_SHOWN,
+    okPixel = AppPixelName.ONBOARDING_DAX_CTA_OK_BUTTON,
+    cancelPixel = null,
+    closePixel = AppPixelName.ONBOARDING_DAX_CTA_DISMISS_BUTTON,
+    ctaPixelParam = Pixel.PixelValues.DAX_FIRE_DIALOG_CTA,
+    onboardingStore = onboardingStore,
+    appInstallStore = appInstallStore,
+    isLightTheme = isLightTheme,
+) {
+    override val activeIncludeId: Int = 0 // TODO: replace in stage 2.
```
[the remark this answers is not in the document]
mikescamell: These are scaffolding-only. `CtaViewModel` returns the legacy `Dax{Fire,NoTrackers,TrackersBlocked,MainNetwork,SiteSuggestions,End}Cta` classes for all non-SERP paths even with the brand-design flag on (see the `// TODO: replace in stage 2` markers), so none of these stubs are instantiated in this PR.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5836,30 +5887,62 @@ class BrowserTabFragment :
...
-                onSuggestedOptionsSelected,
-                {
-                    viewModel.onUserClickCtaDismissButton(configuration)
-                },
-            )
+            if (configuration is OnboardingDaxDialogCta.BrandDesignContextualDaxDialogCta) {
+                binding.daxDialogOnboardingCtaContent.layoutTransition.apply {
+                    disableTransitionType(LayoutTransition.APPEARING)
+                    disableTransitionType(LayoutTransition.DISAPPEARING)
+                }
```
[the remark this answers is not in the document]
mikescamell: `layoutTransition` is non-null at both call sites. `configureWebView()` (called from `onViewCreated` at L1137, before any CTA can be observed) initializes it at L3858: `binding.daxDialogOnboardingCtaContent.layoutTransition = LayoutTransition()`. So even though the FrameLayout doesn't set `android:animateLayoutChanges="true"` in XML, `getLayoutTransition()` is non-null by the time `showOnboardingDialogCta` runs. No NPE in either branch.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4818,9 +4842,43 @@ class BrowserTabViewModel @Inject constructor(
...
+                        command.value = HideOnboardingDaxDialog(onboardingCta)
+                    }
+                }
+                null
+            }
+
             is OnboardingDaxDialogCta.DaxMainNetworkCta,
+            is DaxMainNetworkBrandDesignUpdateContextualCta,
             -> {
+                // TODO: replace in stage 2 (DaxMainNetworkCta brand-design migration).
```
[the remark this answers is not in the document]
mikescamell: Intentional, stubs for the next dialogs

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5754,19 +5769,58 @@ class BrowserTabFragment :
...
+            existingRoot.setOnClickListener(null)
+            existingRoot.removeAllViews()
+            while (fresh.childCount > 0) {
+                val child = fresh.getChildAt(0)
+                fresh.removeViewAt(0)
+                existingRoot.addView(child)
+            }
+
+            applyContextualOptionsHeight(cta)
+            showCta(cta, instantShow = true)
```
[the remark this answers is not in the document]
mikescamell: The `if (!instantShow)` guard at the end of `showOnboardingDialogCta` already covers this — `reinflateContextualBrandDesignDialog` calls `showCta(cta, instantShow = true)`, so `viewModel.onCtaShown()` is not invoked on the rotation path. The guard was introduced in the same commit that added the reinflate code (commit `55f90eb26a`), which predates the SHA this review ran on.

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -539,6 +550,493 @@ sealed class OnboardingDaxDialogCta(
...
+                    onDismissCtaClicked = onDismissCtaClicked,
+                    onTypingAnimationFinished = onTypingAnimationFinished,
+                )
+                return
+            }
+
+            animationsSettled = false
+
+            val titleView = container.findViewById<DaxTypeAnimationTextView>(R.id.contextualBrandDesignTitle)
+            val hiddenTitle = container.findViewById<DaxTextView>(R.id.contextualBrandDesignHiddenTitle)
```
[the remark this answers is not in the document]
mikescamell: Good catch — folded into the scaffolding commit.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4914,7 +4977,8 @@ class BrowserTabViewModel @Inject constructor(
             pixel.fire(DuckChatPixelName.DUCK_CHAT_FIRE_BUTTON_TAPPED)
         }
         val cta = currentCtaViewState().cta
-        if (cta is OnboardingDaxDialogCta.DaxFireButtonCta) {
+        if (cta is OnboardingDaxDialogCta.DaxFireButtonCta || cta is DaxFireButtonBrandDesignUpdateContextualCta) {
+            // TODO: replace in stage 2 (DaxFireButtonCta brand-design migration).
```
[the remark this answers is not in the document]
mikescamell: The cast is safe — `DaxFireButtonBrandDesignUpdateContextualCta` extends `BrandDesignContextualDaxDialogCta` which extends `OnboardingDaxDialogCta`, so this is a type-system formality, not a runtime risk.

On the race comparison: `onUserDismissedCta(cta)` already wraps the persistence call in `viewModelScope.launch { ctaViewModel.onUserDismissedCta(it) }` (see `BrowserTabViewModel.kt:3477`), so emitting `HideOnboardingDaxDialog` right after it is exactly how the legacy `DaxFireButtonCta` arm has worked for years. The brand-design check just joins the existing `||` — no new ordering behaviour.

On reachability: agreed, this is scaffolding. `CtaViewModel.getFireDialogCta()` still returns the legacy class with the brand-design flag on, and the `// TODO: replace in stage 2` one line up flags exactly this. Stage 2 will be the first commit where this arm is reachable.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5795,19 +5811,58 @@ class BrowserTabFragment :
...
+        }
+
+        private fun applyContextualOptionsHeight(cta: OnboardingDaxDialogCta.BrandDesignContextualDaxDialogCta) {
```
LukasPaczos: Looks like this needs to be called for every code path that shows the CTA - why not move it directly into `BrandDesignContextualDaxDialogCta` and call somewhere from `applyContent()`? Otherwise, if the code paths in `BrowserTabFragment` change, this can be lost.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5892,32 +5947,66 @@ class BrowserTabFragment :
...
+                binding.daxDialogOnboardingCtaContent.layoutTransition.apply {
+                    disableTransitionType(LayoutTransition.APPEARING)
+                    disableTransitionType(LayoutTransition.DISAPPEARING)
+                }
+                // Both layouts live in the tree; only one should be visible at a time.
+                daxDialogInContext.root.gone()
+                applyContextualOptionsHeight(configuration)
+                configuration.showOnboardingCta(
+                    binding,
+                    { viewModel.onUserClickCtaOkButton(configuration) },
```
LukasPaczos: nit: add names params

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5892,32 +5947,66 @@ class BrowserTabFragment :
         }
 
         @SuppressLint("ClickableViewAccessibility")
-        private fun showOnboardingDialogCta(configuration: OnboardingDaxDialogCta) {
+        private fun showOnboardingDialogCta(configuration: OnboardingDaxDialogCta, instantShow: Boolean = false) {
```
LukasPaczos: nit: Instead of adding an if-check here, maybe split the function to:
- `private fun showOnboardingDialogCta(configuration: OnboardingDaxDialogCta)` - left as-is, no diff
- `private fun showOnboardingDialogCta(configuration: BrandDesignContextualDaxDialogCta, instantShow: Boolean = false)`

Less diff and potentially easier to clean up later.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -6058,6 +6147,12 @@ class BrowserTabFragment :
         private fun hideDaxCta() {
             daxDialogInContext.dialogTextCta.cancelAnimation()
             daxDialogInContext.daxCtaContainer.gone()
+            (lastSeenCtaViewState?.cta as? OnboardingDaxDialogCta.BrandDesignContextualDaxDialogCta)
+                ?.hideOnboardingCta(binding)
+            daxDialogInContextBrandDesign.root
+                .findViewById<DaxTypeAnimationTextView>(R.id.contextualBrandDesignTitle)
+                ?.cancelAnimation()
+            daxDialogInContextBrandDesign.root.gone()
```
LukasPaczos: If we ever add another animation, we'd need to remember to cancel it here and in `BrandDesignContextualDaxDialogCta.hideOnboardingCta` above, which is already doing the same `cancelAnimation()` + `gone()` on the brand design view. Can we make sure there's only one and shared place where cancellation happens?

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -539,6 +550,492 @@ sealed class OnboardingDaxDialogCta(
...
+        ctaPixelParam = ctaPixelParam,
+        onboardingStore = onboardingStore,
+        appInstallStore = appInstallStore,
+    ) {
+
+        protected var ctaView: View? = null
+
+        private var runningFadeIn: AnimatorSet? = null
+        private var runningFadeOut: AnimatorSet? = null
+        private var animationsSettled: Boolean = false
```
LukasPaczos: Why not use the same pattern we used in all the other bubbles?

```suggestion
        private var cardContainer: TouchInterceptingLinearLayout? = null

        private var isAnimating: Boolean = false
            set(value) {
                field = value
                cardContainer?.interceptChildTouches = value
            }
```

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -539,6 +550,492 @@ sealed class OnboardingDaxDialogCta(
...
+         */
+        protected open fun onTypingAnimationSettled(onTypingAnimationFinished: () -> Unit) {
+            // No-op by default — see kdoc for the override contract.
+        }
+
+        override fun hideOnboardingCta(binding: FragmentBrowserTabBinding) {
+            cancelRunningAnimations()
+            binding.includeOnboardingInContextDaxDialogBrandDesign.root
+                .findViewById<DaxTypeAnimationTextView>(R.id.contextualBrandDesignTitle)
+                ?.cancelAnimation()
```
LukasPaczos: Why is this not in `cancelRunningAnimations`?

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -539,6 +550,492 @@ sealed class OnboardingDaxDialogCta(
...
+                            .setDuration(DIALOG_CONTENT_FADE_IN_DURATION),
+                        ObjectAnimator.ofFloat(activeInclude, View.ALPHA, 1f)
+                            .setDuration(DIALOG_CONTENT_FADE_IN_DURATION),
+                    )
+                    // Dismiss button is persistent; on content transition it already sits at alpha=1
+                    // so we omit the fade-in animator (otherwise it would briefly snap then re-show).
+                    if (!isContentTransition) {
+                        animators += ObjectAnimator.ofFloat(dismissButton, View.ALPHA, 1f)
+                            .setDuration(DIALOG_CONTENT_FADE_IN_DURATION)
+                    }
```
LukasPaczos: This assumes the dismissButton was fully shown by the previous CTA but in case of cancellations or CTA change mid-flight, this can be brittle. Wouldn't it be more robust to check the button's alpha here and decide whether to animate based on that?

I think this comment applies to other places in here as well that assume the previous state instead of verifying it - so if we ever needed to skip or re-order steps, they might break. Since we're rewriting the steps from scratch, could you take a look at what it would take to make them more self-reliant?

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -539,6 +550,492 @@ sealed class OnboardingDaxDialogCta(
...
+                onSettled()
+            }
+        }
+
+        /**
+         * Collapse the title slot when the CTA has no title so the FrameLayout (and its
+         * marginBottom) is excluded from the LinearLayout flow, leaving the description sitting
+         * at the card's top padding. Runs synchronously right after [configureContentViews] so
+         * the layout is correct from the first frame.
+         */
```
LukasPaczos: This comment is misplaced.

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -539,6 +550,492 @@ sealed class OnboardingDaxDialogCta(
...
+                .setDuration(BACKGROUND_SLIDE_DURATION)
+                .setInterpolator(AccelerateDecelerateInterpolator())
+                .start()
+        }
+
+        /** Y translation that takes [view] just below its parent's bottom edge. */
+        private fun offScreenY(view: View): Float {
+            val parent = view.parent as? View
+            return if (parent != null) (parent.height - view.top).toFloat() else view.height.toFloat()
+        }
```
LukasPaczos: This could be extracted out to. For example:
```kotlin
  private class BackgroundBanner(private val view: ImageView, @DrawableRes private val res: Int) {
      fun show() { ... }                     // setImageResource + visible + offscreen
      fun slideIn(): Animator { ... }
      fun slideOut(): Animator? { ... }      // null if not currently showing a banner
      val isShowing: Boolean get() = view.visibility == View.VISIBLE
  }
```

  Wins:
  - Base class shrinks meaningfully.
  - The banner is testable in isolation (it doesn't need a full CTA scaffold to exercise slide-in/out).
  - Subclasses without a banner (end-of-journey, fire-button) just construct BackgroundBanner(view, res = 0) once or skip the construction
  entirely.

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
mikescamell: I've switched the dismiss button fade-in based on`dismissButton.alpha < 1f` which will cover first-show and the mid-flight cancel case.

I'm going to leave the rest as it's working as expected, I don't want to delay these changes any longer with the impending deadline 👍 When we get to use Compose this would be a good time to reconsider.

app/src/main/res/layout/include_onboarding_in_context_dax_dialog_brand_design_update.xml
```diff
@@ -0,0 +1,156 @@
...
+        app:layout_constraintWidth_max="@dimen/getDesktopBrowserMaxItemWidth"
+        app:showArrow="false">
+
+        <com.duckduckgo.app.onboarding.ui.view.TouchInterceptingLinearLayout
+            android:id="@+id/contextualBrandDesignCardContainer"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:orientation="vertical"
+            android:paddingTop="@dimen/keyline_6"
+            android:paddingBottom="20dp">
```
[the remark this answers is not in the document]
mikescamell: Keyline scale jumps `keyline_4=16dp` → `keyline_5=24dp` with no 20dp option. Keeping the value as the design-intentional spacing rather than rounding to a different visual gap.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4992,7 +5055,8 @@ class BrowserTabViewModel @Inject constructor(
             return
         }
 
-        if (cta is OnboardingDaxDialogCta.DaxFireButtonCta) {
+        if (cta is OnboardingDaxDialogCta.DaxFireButtonCta || cta is DaxFireButtonBrandDesignUpdateContextualCta) {
+            // TODO: replace in stage 2 (DaxFireButtonCta brand-design migration).
```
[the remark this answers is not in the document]
mikescamell: Same as the previously-resolved threads on this line — the stage-2 stub classes are unreachable today because `CtaViewModel.getFireDialogCta()` still returns the legacy `DaxFireButtonCta` (the `// TODO: replace in stage 2` is still there). Stage 2 will be the first commit where the brand-design fire CTA arm is exercised.

app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
LukasPaczos: Makes sense, thanks for looking into it.

LukasPaczos (approved):
🚢

PR #8469, opened 2026-05-06
Add Duck.ai-focused onboarding experiment

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205648422731273/task/1214573630528382

### Description

Introduces a new extended onboarding path that enrols a portion of new users into a Duck.ai-focused experience. After the standard pre-onboarding dialogs, eligible users are taken straight to Duck.ai with an initial prompt and guided through a shortened onboarding flow:
- A duck.ai page with the prompt's response and a fire-button CTA once the response arrives.
- A duck.ai-specific end CTA on the input screen.
- The subscription CTA on the home page (post-end), reusing the existing subscription onboarding CTA.

#### Duck.ai onboarding path
- [x] Update `DuckAiOnboardingExperimentManager.enroll()` to return one of the treatment variants. Remove `.debug` application id suffix to make the app eligible for purchasing subscriptions (a condition for showing subscription onboarding CTA).
- [x] Going through the initial onboarding pages, select "search & duck.ai"
- [x] Verify that the new screen is shown with the input mode toggle. 
- [x] Submit a **chat prompt** (either using the edit field or one of the suggestions)
- [x] Verify that 
  - [x] duck.ai is opened and generates response to submitted prompt,
  - [x] omnibar is not interactive (except the fire button if highlighted)
  - [x] fire button CTA shows after the response is generated or 2 seconds after duck.ai page finishes loading (whichever comes first)
  - [x] fire button is highlighted while the fire CTA is shown
- [x] use the fire button to perform single-tab data clear
- [x] verify that after the fire animation input screen is visible and "You've got this..." CTA is shown
- [x] Dismiss the CTA
- [x] Close input screen (navigate back)
- [x] Verify that subscription CTA is shown on NTP.
- [x] Repeat this scenario while choosing different omnibar placements during pre-onboarding (top/bottom/split)

#### Regular, search onboarding path
- [x] Update `DuckAiOnboardingExperimentManager.enroll()` to return one of the treatment variants. Remove `.debug` application id suffix to make the app eligible for purchasing subscriptions (a condition for showing subscription onboarding CTA).
- [x] Going through the initial onboarding pages, select "search & duck.ai"
- [x] Verify that the new screen is shown with the input mode toggle. 
- [x] Submit a **search query** (either using the edit field or one of the suggestions)
- [x] Verify that the browser opens with SERP page
- [x] Verify that the rest of the onboarding path is unchanged compared to production.

#### Control
- [x] Update `DuckAiOnboardingExperimentManager.enroll()` to return `CONTROL`. Remove `.debug` application id suffix to make the app eligible for purchasing subscriptions (a condition for showing subscription onboarding CTA).
- [x] Go through onboarding and verify there are no changes.

### UI changes
| Before  | After |
| ------ | ----- |

---

> [!NOTE]
> **Medium Risk**
> Introduces a new Duck.ai-specific onboarding path that alters browser interaction states (omnibar/nav lock, fire dialog handling, swipe-to-refresh) and adds new CTA/metrics plumbing, which could affect navigation and onboarding completion if mis-gated.
> 
> **Overview**
> Adds a Duck.ai-focused onboarding experiment cohort (gated by a new `onboardingDuckAiExperimentMay26` toggle and device/prerequisite checks) plus experiment metrics reporting for option selection, screen impressions, and CTA presses.
> 
> Implements a Duck.ai onboarding flow that deep-links into Duck.ai with a `flow=mobile-app-onboarding` param, **locks omnibar + navigation interactions** during onboarding (with a new `isOmnibarLockedForOnboarding` state and enabled/disabled UI states), delays showing the fire-button onboarding CTA until a Duck.ai response arrives (or a short fallback timer), and disables swipe-to-refresh while locked.
> 
> Extends the input screen to optionally show a Duck.ai end CTA overlay and return its interaction result to the browser; wires this into new Duck.ai-specific CTAs (`DAX_DUCK_AI_FIRE_BUTTON`, `DAX_DUCK_AI_END`), updated CTA ordering (including when subscription CTA can appear), and fire-dialog handling to cleanly dismiss onboarding UI and exit full-screen chat after single-tab clear.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f7f388ca4cac7c813f7e79da85fbc66ce26a781. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/TextExtensions.kt
```diff
@@ -168,3 +168,40 @@ fun Context.prependIconToText(textResId: Int, iconResId: Int): SpannableStringBu
...
+                canvas.save()
+
+                val fontMetricsInt = paint.fontMetricsInt
+                val translationY = ((y + fontMetricsInt.descent + y + fontMetricsInt.ascent) / 2) - (icon.bounds.height() / 2)
+                canvas.translate(x, translationY.toFloat())
+                icon.draw(canvas)
+
+                canvas.restore()
+            }
+        }
```
LukasPaczos: This portion is a copy-paste from `prependIconToText`, maybe let's add a shared helper?

app/src/main/java/com/duckduckgo/app/browser/navigation/bar/view/BrowserNavigationBarView.kt
```diff
@@ -208,6 +214,29 @@ class BrowserNavigationBarView @JvmOverloads constructor(
...
+        binding.tabsButton.isEnabled = !locked
+        binding.tabsButton.alpha = alpha
+        binding.menuButton.isEnabled = !locked
+        binding.menuButton.alpha = alpha
+
+        // Fire button is exempt from the lock
+        val fireButtonLocked = locked && !viewState.fireButtonHighlighted
+        val fireButtonAlpha = if (fireButtonLocked) 0.4f else 1.0f
+        binding.fireButton.isEnabled = !fireButtonLocked
+        binding.fireButton.alpha = fireButtonAlpha
```
LukasPaczos: nit: This logic would be better suited for the view model and only state (enabled/disabled/alpha) be delivered to the view to apply. This way it would also be testable.

app/src/main/java/com/duckduckgo/app/browser/navigation/bar/view/BrowserNavigationBarView.kt
```diff
@@ -120,6 +120,12 @@ class BrowserNavigationBarView @JvmOverloads constructor(
         }
     }
 
+    fun setLockedForOnboarding(locked: Boolean) {
+        doOnAttach {
+            viewModel.setLockedForOnboarding(locked)
+        }
+    }
```
LukasPaczos:
```suggestion
    /** Disables and makes semi-transparent all buttons but the Fire Button */
    fun disableNonFireButtons(disable: Boolean) {
        doOnAttach {
            viewModel.disableNonFireButtons(disable)
        }
    }
```

nit: The navigation bar shouldn't need to know anything about onboarding.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
```diff
@@ -1470,6 +1476,43 @@ class OmnibarLayout @JvmOverloads constructor(
...
+        duckAISidebar.isEnabled = !locked
+        duckAISidebar.alpha = lockedAlpha
+        duckAIHeader.isEnabled = !locked
+        duckAIHeader.alpha = lockedAlpha
+        shieldIcon.isEnabled = !locked
+        shieldIcon.alpha = lockedAlpha
+        omnibarTextInput.alpha = lockedAlpha
+
+        // Fire button is exempt from the lock when highlighted
+        val fireButtonLocked = locked && !viewState.highlightFireButton.isHighlighted()
```
LukasPaczos: nit: Same as above, ideally, logic like this should not happen in the view.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -620,9 +628,11 @@ class BrowserTabViewModel @Inject constructor(
     @ExperimentalCoroutinesApi
     @FlowPreview
     private val showPulseAnimation: LiveData<Boolean> =
-        ctaViewModel.showFireButtonPulseAnimation.asLiveData(
-            context = viewModelScope.coroutineContext,
-        )
+        combine(
+            ctaViewState.asFlow().map { it.cta is OnboardingDaxDialogCta.DaxDuckAiFireButtonCta }.distinctUntilChanged(),
+            ctaViewModel.showFireButtonPulseAnimation,
+        ) { isShowingDuckAiFireButtonCta, showPulseAnimation -> isShowingDuckAiFireButtonCta || showPulseAnimation }
```
LukasPaczos, line 635: Why do we need this logic here, couldn't it be resolved within `ctaViewModel.showFireButtonPulseAnimation`?

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -210,15 +214,43 @@ class CtaViewModel @Inject constructor(
...
+
+    suspend fun onDuckAiEndCtaInteraction(okClicked: Boolean) {
+        withContext(dispatchers.io()) {
+            val params = mapOf(Pixel.PixelParameter.CTA_SHOWN to DUCK_AI_END_CTA_PIXEL_PARAM)
+            if (okClicked) {
+                pixel.fire(AppPixelName.ONBOARDING_DAX_CTA_OK_BUTTON, params)
+            } else {
+                pixel.fire(AppPixelName.ONBOARDING_DAX_CTA_DISMISS_BUTTON, params)
+            }
+            completeStageIfDaxOnboardingCompleted()
```
LukasPaczos: It's impossible to reach this stage without already being ESTABLISHED, so calling this is always no-op. Why did we need it here?

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -210,15 +214,43 @@ class CtaViewModel @Inject constructor(
...
+    suspend fun prepareAndMarkDuckAiEndCtaForInputScreen(): Boolean {
+        return withContext(dispatchers.io()) {
+            val shouldShow = canShowDuckAiEndCta() && !extendedOnboardingFeatureToggles.noBrowserCtas().isEnabled()
+            if (shouldShow) {
+                dismissedCtaDao.insert(DismissedCta(CtaId.DAX_DUCK_AI_END))
+                completeStageIfDaxOnboardingCompleted()
+                if (canSendShownPixel(onboardingStore, DUCK_AI_END_CTA_PIXEL_PARAM)) {
+                    val journey = addCtaToHistory(onboardingStore, appInstallStore, DUCK_AI_END_CTA_PIXEL_PARAM)
+                    pixel.fire(AppPixelName.ONBOARDING_DAX_CTA_SHOWN, mapOf(Pixel.PixelParameter.CTA_SHOWN to journey))
+                }
```
LukasPaczos, line 234: Why are we doing this manually instead of calling `CtaViewModel#onCtaShown`?

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingActivity.kt
```diff
@@ -86,14 +88,14 @@ class OnboardingActivity : DuckDuckGoActivity() {
...
+        viewModel.onOnboardingDone(extendedOnboardingFlow = DEFAULT_WITHOUT_INTRO_CTA)
         startActivity(BrowserActivity.intent(this@OnboardingActivity, queryExtra = query))
         finish()
     }
 
     fun finishAndSubmitChatPrompt(prompt: String) {
-        viewModel.onOnboardingDone()
-        val duckChatUrl = duckChat.getDuckChatUrl(prompt, autoPrompt = true)
+        viewModel.onOnboardingDone(extendedOnboardingFlow = DUCK_AI_FOCUSED)
+        val duckChatUrl = duckChat.getDuckChatUrl(prompt, autoPrompt = true) + "&flow=mobile-app-onboarding"
```
LukasPaczos:
```suggestion
        val duckChatUrl = duckChat.getDuckChatUrl(prompt, autoPrompt = true, onboarding = true)
```

What do you think about moving this over to the duck-chat module to own instead of having it spread across multiple classes?

Then we can do `duckChat.isDuckChatOnboardingUrl()` in the browser tab instead of hardcoding the param name again.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -891,9 +901,23 @@ class BrowserTabViewModel @Inject constructor(
...
+            // Detect the duck.ai onboarding flow as early as possible so the omnibar is locked
+            enterDuckAiOnboardingFlowIfApplicable(it.toUri())
         }
     }
 
+    private fun enterDuckAiOnboardingFlowIfApplicable(uri: Uri?) {
+        if (uri == null) return
+        if (currentBrowserViewState().isOmnibarLockedForOnboarding) return
+        if (!duckChat.isDuckChatUrl(uri)) return
+        if (uri.getQueryParameter("flow") != "mobile-app-onboarding") return
```
LukasPaczos, line 916:
```suggestion
        if (!duckChat.isDuckChatOnboardingUrl(uri)) return
```
Per suggestions above.

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
```diff
@@ -367,6 +422,16 @@ class CtaViewModel @Inject constructor(
                 return null
             }
 
+            // Duck.ai-focused onboarding CTAs
+            if (duckChat.isDuckChatUrl(it.url.toUri())) {
```
LukasPaczos, line 457:
```suggestion
            if (duckChat.isDuckChatOnboardingUrl(it.url.toUri())) {
```
Related to the other comment, if that's something you'd like to do.

app/src/test/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModelTest.kt
LukasPaczos, line 1: Missing tests for dismissing CTAs.

duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/inputscreen/InputScreenActivityParams.kt
```diff
@@ -40,6 +40,7 @@ data class InputScreenActivityParams(
     val launchOnChat: Boolean = false,
     val showReturnHatch: Boolean = false,
     val isNewTab: Boolean = false,
+    val showDuckAiEndCta: Boolean = false,
```
LukasPaczos: Needs public api proposal.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/InputScreenFragment.kt
```diff
@@ -915,6 +934,55 @@ class InputScreenFragment : DuckDuckGoFragment(R.layout.fragment_input_screen) {
         binding.chatSuggestionsBottomFadeContainer.getChildAt(0)?.invalidate()
     }
 
+    private fun showDuckAiEndCta() {
+        duckAiEndCtaVisible = true
+        binding.ddgLogoContainer.isVisible = false
+        binding.viewPager.isVisible = false
+        binding.newTabContainerScrollView.isVisible = false
```
LukasPaczos, line 951: `newTabContainerScrollView` visibility change is not reverted after dismissing the CTA.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/InputScreenFragment.kt
```diff
@@ -915,6 +934,55 @@ class InputScreenFragment : DuckDuckGoFragment(R.layout.fragment_input_screen) {
         binding.chatSuggestionsBottomFadeContainer.getChildAt(0)?.invalidate()
     }
 
+    private fun showDuckAiEndCta() {
```
LukasPaczos, line 941: Would be great if this is driven by a state from the view model (and tested) instead of being executed directly from `onViewCreated`.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/InputScreenFragment.kt
```diff
@@ -728,6 +745,8 @@ class InputScreenFragment : DuckDuckGoFragment(R.layout.fragment_input_screen) {
     }
 
     private fun updateLogoVisibility(state: InputScreenVisibilityState) {
+        if (duckAiEndCtaVisible) return
```
LukasPaczos, line 748: Related to the other comment about moving this state to the VM.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/InputScreenFragment.kt
```diff
@@ -915,6 +934,55 @@ class InputScreenFragment : DuckDuckGoFragment(R.layout.fragment_input_screen) {
...
+            .alpha(1f)
+            .setDuration(OVERLAY_ANIMATION_DURATION)
+            .setStartDelay(400)
+            .start()
+
+        binding.duckAiEndCta.duckAiEndCtaOkButton.setOnClickListener {
+            duckAiEndCtaOkClicked = true
+            hideDuckAiEndCta()
+        }
+    }
```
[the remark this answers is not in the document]
LukasPaczos: Valid and should be addressed.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/InputScreenFragment.kt
```diff
@@ -929,6 +997,8 @@ class InputScreenFragment : DuckDuckGoFragment(R.layout.fragment_input_screen) {
         hasContent: Boolean,
         state: InputScreenVisibilityState,
     ) {
+        if (duckAiEndCtaVisible) return
```
LukasPaczos, line 1012: Same as above.

LukasPaczos (commented):
I tested the PR and it mostly works as expected. Two issues I run into:
1. If device has no connectivity when trying to load the Duck.ai page during onboarding, the only way to recover is a pull-to-refresh once connectivity is regained. Given that the whole UI is locked at that point, I'm not sure how intuitive that's going to be. I don't think it's a blocker now, since recovery is possible, but maybe worth discussing before the experiment ships?
2. After the Subscription CTA, the tab is left in a state with open keyboard but unfocused omnibar. Video below:

https://github.com/user-attachments/assets/840914bd-5f31-4068-92d6-57b0ce13eb2b

LukasPaczos:
One more issue I noticed is that the UI gets unlocked if the app is restarted mid-fow:

https://github.com/user-attachments/assets/7b899c37-979c-4b70-b493-7b070136b44b

Likely not a blocker for this project, but might become one for the follow ups since escaping the flow would mean users never get to revisit a portion of the onboarding. Is there a low-hanging fix for this that might be worth doing now?

app/src/main/java/com/duckduckgo/app/browser/navigation/bar/view/BrowserNavigationBarView.kt
lmac012: The assumption here was that it doesn't make sense to have highlighted view that is also disabled, so the view would not render an incorrect state. But I don't mind moving this to the viewmodel, so I'll refactor this.

app/src/main/java/com/duckduckgo/app/browser/navigation/bar/view/BrowserNavigationBarView.kt
lmac012: I think the suggested name is confusing, since fire button may be disabled as well (it it's not highlighted), but I can refactor this to avoid putting "onboarding" in the naming.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
lmac012: `CtaViewModel` does not currently know if this particular tab is showing duck ai fire CTA. Since `CtaViewModel` is a global singleton, it's not the best place to manage tab-specific state.

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
lmac012: Right, this was needed before but is not anymore - I'll remove it.

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
lmac012: `CtaId.DAX_DUCK_AI_END` Is a bit of a special case right now, because it is shown in `InputScreenActivity` where the existing onboarding infrastructure is not available (for example, there is no `Cta` instance that we could pass to `onCtaShown()`. We will be able to clean this up once we remove InputScreenActivity and move this CTA back to browser tab.

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingActivity.kt
lmac012: When possible, I prefer to avoid updating public API for something that is just an experiment. I think this change is worth considering once the experiment is done.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/inputscreen/ui/InputScreenFragment.kt
lmac012: I don't disagree, but since this is most likely going away in a few weeks (due to migration to the native input widget), I'd prefer to not do any refactoring inside `InputScreenFragment`, especially since the current implementation has already been tested by multiple people.

lmac012:
@LukasPaczos Issues 1. and 3. are known and not blockers; I fixed 2. in https://github.com/duckduckgo/Android/pull/8469/commits/92b278309d5e95f392275d5f180b258743ed3850

app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
LukasPaczos: Why can't `onCtaShown` be called right here, in the `CtaViewModel` (or in `BrowserTabFragment`) right before dispatching the InputScreen launch?

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingActivity.kt
LukasPaczos: We are building one more experiment, and one more permanent feature on top of this change, so I don't think it's going anywhere. Let's at least create a follow up task so that this doesn't slip.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3448,6 +3504,15 @@ class BrowserTabViewModel @Inject constructor(
...
                 command.value = HideOnboardingDaxBubbleCta(cta)
+                // For users with the input screen enabled, the omnibar text input is hidden behind a
+                // click catcher and is not focusable, so showing the keyboard from
+                // hideOnboardingDaxBubbleCta would land on a non-focused field. Launch the input
+                // screen instead — same UX they'd get from tapping the omnibar. Mirrors the
+                // behaviour of onPrivacyProSkippedOnboardingDismissed for SubscriptionPromoModalCta.
+                val isSubscriptionCta = cta is DaxBubbleCta.DaxSubscriptionCta || cta is DaxSubscriptionBrandDesignUpdateBubbleCta
+                if (isSubscriptionCta && duckAiFeatureState.showInputScreenAutomaticallyOnNewTab.value) {
+                    command.value = LaunchInputScreen()
+                }
```
LukasPaczos: This changes the original behavior - in production, we do not auto-focus on the omnibar after subscription CTA is dismissed. Flagging in case you'd like to double-check with the product team.

LukasPaczos (commented):
I run into one more issue: when the final CTA is shown, and you switch to Duck.ai tab and type anything, the primary button is not clickable.

https://github.com/user-attachments/assets/fc0684af-64ad-4d69-aa9e-a987201091a9

lmac012:
* **#8512** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8512?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8469** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8469?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8469?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
lmac012: The original behavior is inconsistent - the code suggests we intend to open the keyboard and focus on address bar after dismissing this CTA, but it doesn't always work. Now that I test this more, it doesn't work most of the time, so I'll adjust the logic here to match.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4064,7 +4064,13 @@ class BrowserTabFragment :
     private fun hideOnboardingDaxBubbleCta() {
         hideDaxBubbleCta()
         renderer.showNewTab()
-        showKeyboard()
+        // When the input-screen feature is on, the omnibar's click catcher disables the text
+        // input, so requestFocus() can't succeed. Showing the IME there leaves the keyboard up
+        // with no focused field. Skip the keyboard call in that case and leave the user on the
+        // NTP — tapping the omnibar will launch the input screen as usual.
+        if (omnibar.omnibarTextInput.isFocusable) {
```
LukasPaczos, line 4071: Is there a specific reason why this check is around `showKeyboard` and guarding only this specific path instead being directly baked into the function?

LukasPaczos (approved):
Works as expected 👍 One comment below that might be worth aligning on though.

LukasPaczos (changes requested):
I was testing the feature E2E with https://github.com/duckduckgo/Android/pull/8512 and https://github.com/duckduckgo/privacy-configuration/pull/5132, and run into a conflict that I think we should address before merging:
Screenshot_20260511_135410

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
lmac012: Just being defensive, since I can't test all the other paths going through `showKeyboard()` before this release.

LukasPaczos (approved):

PR #8478, opened 2026-05-07
Fire Mode: Add BrowserMode with plumbing

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1214579513091724?focus=true

### Description

This PR adds the new `BrowserMode` and the related interfaces and their concrete implementations:

- `BrowserModeStateHolder` that exposes enum class `BrowserMode { REGULAR, FIRE }`, which is the single source of truth for current mode. Theming overlay, tab switcher VM, mode-toggle pill, sync entry points, repository providers — all observe this Flow.
- Adds `FireModeAvailability` that checks the feature flag and the WebView feature support
- Adds `BrowserMode` plumbing to the main entry points

Related API proposal: https://app.asana.com/1/137249556945/project/1207418217763355/task/1214579944546013?focus=true

### Steps to test this PR

Smoke testing that the app is launched correctly and opening an external link works as expected is sufficient.

---

> [!NOTE]
> **Medium Risk**
> Touches core `BrowserActivity` intent processing and tab creation, adding mode-based recreation and deferred-intent handling which could affect external link/widget/shortcut launches. New mode plumbing is guarded by availability checks but has broad surface area across entry points.
> 
> **Overview**
> Introduces a new cross-module `BrowserMode` concept (`REGULAR`/`FIRE`) with a `BrowserModeStateHolder` source-of-truth and an app implementation (`RealBrowserModeStateHolder`).
> 
> Updates `BrowserActivity` and tab creation to be mode-aware (adapter/`BrowserTabFragment` arguments, DuckChat fragment args), and recreates the activity on mode changes; external entry-point intents can now be stamped as `LAUNCH_REQUIRES_REGULAR_MODE` and, if received while in `FIRE`, are deferred across a FIRE→REGULAR switch before processing.
> 
> Adds a new `fire-mode-api`/`fire-mode-impl` with `FireModeAvailability` (feature flag + WebView MultiProfile capability), wires it into `BrowserViewModel.switchToMode`, and updates numerous launch points (widgets, shortcuts, onboarding/settings/internal screens, tests) to pass a `BrowserLaunchSource` and correctly stamp intents.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f0669e6c13e423ab3f065aa51ed2e1cedfcfcdc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


0nko:
* **#8606** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8606?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8591** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8575** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8575?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8574** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8574?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8478** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

duckchat/duckchat-impl/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     implementation project(':data-store-api')
     implementation project(':cookies-api')
     implementation project(':duckchat-store')
+    implementation project(':fire-mode-api')
```
CDRussell: is this needed?

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -260,6 +265,13 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
     private var lastIntent: Intent? = null
 
+    /**
+     * External [Intent] received while in FIRE mode, deferred until the REGULAR-bound activity
+     * comes up. Survives the mode-switch [recreate] via [onSaveInstanceState]; consumed by
+     * [BrowserStateRenderer.showWebContent].
+     */
+    private var pendingExternalIntent: Intent? = null
```
CDRussell: is this only used for an `Intent` received while in fire mode like the comment says? If so, maybe add that to the name of it to make it clearer.

this new thing is confusing because we also already have a similar concept for `lastIntent` where it similarly has a deferred processing step inside of `showWebContent` like the one you've added for `pendingExternalIntent`

it's not clear what relationship these two transient intents share with each other, nor the order in which they should be processed if both exist. I don't have a concrete suggestion yet for improving, but flagging as a point of confusion in case you can come up with a good way of combining these and/or making it clearer how each relate to the other.

app/src/main/java/com/duckduckgo/app/browser/BrowserViewModel.kt
```diff
@@ -111,8 +115,25 @@ class BrowserViewModel @Inject constructor(
...
+        browserModeStateHolder.switchTo(mode)
+        return true
+    }
+
+    /**
+     * Returns true when an incoming external intent must trigger a switch to REGULAR before being
+     * processed.
+     */
+    fun shouldSwitchToRegularModeBeforeProcessingIntent(isExternal: Boolean): Boolean =
+        isExternal && currentMode.value == BrowserMode.FIRE
```
CDRussell: rather than checking if current mode is `FIRE`, would rather we checked if current mode `!= REGULAR`; that's what you really care about. i.e., switch required if:
- request to be in `REGULAR` mode
- it checks and discover you aren't in `REGULAR` mode

guarding against a future footgun if we were to add a third mode

browser-api/src/main/java/com/duckduckgo/browser/api/mode/BrowserModeStateHolder.kt
```diff
@@ -0,0 +1,36 @@
...
+ * mutator. Consumers either collect [currentMode] for reactive updates, or read
+ * [StateFlow.value] for a snapshot.
+ *
+ * Tab-level code should not consume this directly — capture the active mode at tab
+ * construction time instead, so a mid-session switch cannot race with tab-bound writes.
+ */
+interface BrowserModeStateHolder {
+    val currentMode: StateFlow<BrowserMode>
+
+    fun switchTo(mode: BrowserMode)
```
CDRussell, line 35: the question around how/when to use `switchTo` safely from the API proposal doc didn't really get answered, and since it's still here besides the `getter` it would benefit from some javadocs

and nit: there's quite a bit of implementation detail in the existing javadocs that i don't think belongs on the API itself (some of those implementation details may well change over time without actually affecting the API)

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -655,6 +671,16 @@ open class BrowserActivity : DuckDuckGoActivity() {
             return
         }
 
+        // External intents always open in REGULAR mode, so we must stash the intent and carry it
+        // to the recreated REGULAR-bound instance, where showWebContent() consumes it.
+        val isExternal = intent.getBooleanExtra(LAUNCH_FROM_EXTERNAL_EXTRA, false)
```
CDRussell: how you determine whether something `isExternal` is too narrow for determining whether `FIRE` or `REGULAR` modes, I think. there are other intents that could come this way where we'd want to enforce they open in `REGULAR` mode. 

for a concrete example, if you add our favorites widget and tap on a favorite, this is not `LAUNCH_FROM_EXTERNAL_EXTRA` and therefore will bypass your check around whether a mode switch is necessary. (it does have `LAUNCH_FROM_FAVORITES_WIDGET=true`). The impact is that favourites will launch into whatever the current mode is; unclear that's the product-level intent.

This is only one example where this can happen. 

maybe `viewModel.shouldSwitchToRegularModeBeforeProcessingIntent` should accept the `Intent` itself, or a dedicated utility class does first. but the main thing is that we might need a broader range of conditions for which we end up with `isExternal == true` and enforcing `REGULAR` mode. 

i reckon you need to audit the various ways our app can launch to make sure each launches in the correct mode. I can help with this, but if you can start with a first pass at identifying them that would be good.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -636,6 +641,16 @@ open class BrowserActivity : DuckDuckGoActivity() {
...
         }
 
+        // External intents always open in REGULAR mode. The recreate observer picks up the mode
+        // change and the new REGULAR-bound activity re-processes this intent via getIntent().
+        val isExternal = intent.getBooleanExtra(LAUNCH_FROM_EXTERNAL_EXTRA, false)
+        if (viewModel.shouldSwitchToRegularModeBeforeProcessingIntent(isExternal)) {
+            logcat(INFO) { "External intent received in FIRE — switching to REGULAR before processing" }
+            setIntent(intent)
+            lifecycleScope.launch { viewModel.switchToMode(BrowserMode.REGULAR) }
+            return
```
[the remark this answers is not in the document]
CDRussell: add a comment why this feedback was considered and dismissed, please

CDRussell (changes requested):
@0nko there are two main things that need done here:

1. consider all the ways the app can launch with an `Intent` and map carefully which will result in `REGULAR` mode being enforced. more details: https://github.com/duckduckgo/Android/pull/8478#discussion_r3225242469
2. feature flag around all of this! this is a majorly critical codepath and any mistakes here could be painful to pin down (e.g, intermittent feedback that something is weird) and we should be able to disable a single feature flag to get back to existing behavior if there are any problems.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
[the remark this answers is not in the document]
0nko: This was resolved by the bugbot itself after changing the code to address the comment.

duckchat/duckchat-impl/build.gradle
0nko: No, it's a leftover from an earlier iteration when I had things in the `fire-mode` module.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
0nko: > is this only used for an Intent received while in fire mode like the comment says? If so, maybe add that to the name of it to make it clearer.

Yeah, `pendingExternalIntent` is only written in `launchNewSearchOrQuery` behind `shouldSwitchToRegularModeBeforeProcessingIntent()`, which gates on `currentMode == FIRE`. I can rename it to make the trigger explicit, something like `pendingFireToRegularIntent`.

> this new thing is confusing because we also already have a similar concept for lastIntent where it similarly has a deferred processing step inside of showWebContent like the one you've added for pendingExternalIntent

`lastIntent` defers until automatic data clearing is finished and uses a different lifecycle (it has to survive the clearer Job in the same Activity instance, whereas `pendingExternalIntent` has to survive  the Activity recreation and is saved). 

I can rename both flags to make their purpose explicit or I can try to think about a bigger refactoring.. Let me know what you think.

browser-api/src/main/java/com/duckduckgo/browser/api/mode/BrowserModeStateHolder.kt
0nko: You're right, I missed the split suggestion, sorry about that. Was the main motivation behind the split to have the "mutator" interface in the app module and "getter" interface in the API module?

I'll clean up the documentation 👍.

CDRussell: >Was the main motivation behind the split to have the "mutator" interface in the app module and "getter" interface in the API module?

That suggestion was in trying to better understand the original questions raised on the API, and suggesting a path forwards. For me personally, I'm ok with what you have here once better documented.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
CDRussell: >I can rename both flags to make their purpose explicit or I can try to think about a bigger refactoring

I think the main thing is to understand the relationship between the two, if there is one. e.g., are they mutually exclusive? if not, what happens if both are set, does one need to be processed before the other etc...?

I don't know the answers to these because it's not clear in code yet. If you think you can make it clearer with renaming and comments, that sounds ok to me, thanks.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
0nko: I've analyzed different possible entry points for the external links. I've been trying to make this logic testable and also easy to glance over in a single place. That's why I created [BrowserLaunchSource](https://github.com/duckduckgo/Android/pull/8478/commits/27e4fc41c845711823795a477c6a4ba175dc32b7). 

There are different types for different sources, where each is explicitly defined whether it requires a `REGULAR` mode or not. `BrowserLaunchSource` is now required when using the `BrowserActivity` intent factory, so that even if we add new possible browser launch points in the feature, we will be forced to decide how it works with the Fire mode. 

I think I got all the possible launch types but I'd be grateful for a double-check.

0nko:
Thanks @CDRussell! This took longer than I expected because there are browser launch points all over the place.

1. Please see my comment [here](https://github.com/duckduckgo/Android/pull/8478#discussion_r3236160686).
2. This was already handled implicitly: when the FF is off, the current mode is always Regular, so the external link intent wouldn't cause the Activity restart. But I added explicit checks anyway, just to be safe.

Ready for another round.

fire-mode/fire-mode-impl/src/main/java/com/duckduckgo/firemode/impl/RealFireModeAvailability.kt
```diff
@@ -0,0 +1,45 @@
...
+    private val dispatchers: DispatcherProvider,
+) : FireModeAvailability {
+
+    private var multiProfileSupported: Boolean? = null
+
+    override suspend fun isAvailable(): Boolean = withContext(dispatchers.io()) {
+        if (!fireModeFeature.fireTabs().isEnabled()) return@withContext false
+        multiProfileSupported ?: webViewCapabilityChecker.isSupported(MultiProfile).also {
+            multiProfileSupported = it
+        }
```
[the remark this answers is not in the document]
0nko: Fixed.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -1027,15 +1070,17 @@ open class BrowserActivity : DuckDuckGoActivity() {
...
-                        Bundle().apply {
+                arguments =
+                    Bundle().apply {
+                        duckChatUrl?.let {
                             putString(KEY_DUCK_AI_URL, duckChatUrl)
                             putInt(KEY_DUCK_AI_TABS, tabs)
                         }
-                }
+                        putString(KEY_DUCK_AI_BROWSER_MODE, mode.name)
+                    }
```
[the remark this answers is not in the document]
0nko: Fixed.

fire-mode/fire-mode-impl/src/main/java/com/duckduckgo/firemode/impl/RealFireModeAvailability.kt
```diff
@@ -0,0 +1,45 @@
...
+) : FireModeAvailability {
+    @Volatile
+    private var multiProfileSupported: Boolean? = null
+
+    override suspend fun isAvailable(): Boolean = withContext(dispatchers.io()) {
+        if (!fireModeFeature.fireTabs().isEnabled()) return@withContext false
+        multiProfileSupported ?: webViewCapabilityChecker.isSupported(MultiProfile).also {
+            multiProfileSupported = it
+        }
+    }
```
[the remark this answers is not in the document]
0nko: If this actually ever happened, the user can wait for a cold app start to get the Fire mode.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -354,6 +375,9 @@ open class BrowserActivity : DuckDuckGoActivity() {
         renderer = BrowserStateRenderer()
         val newInstanceState = if (dataClearer.isFreshAppLaunch) null else savedInstanceState
         instanceStateBundles = CombinedInstanceState(originalInstanceState = savedInstanceState, newInstanceState = newInstanceState)
+        pendingFireToRegularIntent = savedInstanceState?.let {
+            BundleCompat.getParcelable(it, KEY_PENDING_FIRE_TO_REGULAR_INTENT, Intent::class.java)
+        }
```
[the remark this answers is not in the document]
0nko: Fixed.

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:SuppressLint("NoImplImportsInAppModule")
```
CDRussell, line 17: why is this needed?

0nko: Supresses the lint error for the `com.duckduckgo.duckchat.impl.ui.DuckChatWebViewFragment.Companion.*` keys that are being used for arguments. This was an pre-existing issue that should eventually get fixed but it's unrelated to this change, aside from adding another key.

CDRussell (approved):

PR #8490, opened 2026-05-07
Persist NTP after-idle hatch across background

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1214582550691757?focus=true

### Description

Fixes the bug reported in the linked task: with the escape hatch enabled, the hatch UI disappears after the second background+resume cycle while the user stays on the auto-NTP.

`NtpAfterIdleManagerImpl` was clearing `_isAfterIdleReturn` in both `onClose()` and `onOpen()`. The state is normally restored by `onNtpShown()` (fired by `BrowserViewModel.flowSelectedTab`), but that flow uses `distinctUntilChanged` on `(tabId, isUrlBlank)` and won't re-emit when the same NTP tab is still selected on resume — so the state stayed stuck at `false`. `FirstScreenHandlerImpl.onOpen()` already detects this case and synchronously sets `pendingAfterIdle = true`, but with no `onNtpShown()` to consume it the pending flag stayed unread.

This PR stops clearing `_isAfterIdleReturn` on the lifecycle callbacks. `onNtpShown()` remains the sole mutator: it correctly classifies new NTP showings (auto vs user-initiated) and the state now survives background+resume on the same auto-NTP. `pendingAfterIdle` is still cleared in `onClose()` so stale pending flags don't leak across sessions.

Aligns with the iOS behaviour confirmed in the task thread: if the NTP was automatically shown due to inactivity, backgrounding the app keeps the hatch on the next startup.

### Steps to test this PR

_Escape hatch persistence_
- [x] Set the escape hatch setting to **Always** (Settings → General → Show on app launch → After inactivity).
- [x] Visit a site (e.g. duckduckgo.com).
- [x] Switch apps and come back — confirm the hatch shows below the address bar.
- [x] Switch apps and come back again without doing anything else — **the hatch should still be visible** (this was the bug).
- [x] Repeat the background+resume a few more times — the hatch should remain visible.
- [x] Tap the hatch — confirm it switches to the previous tab and the hatch hides.

_User-initiated NTP still does not show the hatch_
- [x] Open a new tab manually (tap the new-tab button) — confirm no hatch appears.
- [x] Background and resume — the hatch should still not appear on this NTP.

_New auto-NTP after activity_
- [x] On a webpage, background long enough to trigger the after-idle return — confirm a fresh auto-NTP with the hatch.

### UI changes
No visual changes. Only the persistence behaviour of the existing hatch is corrected.

---

> [!NOTE]
> **Medium Risk**
> Changes lifecycle/state management for the after-idle NTP hatch, which can affect when the hatch appears/disappears across app background/resume and fresh launches. Risk is moderate due to timing/order sensitivity around lifecycle callbacks and tab selection events.
> 
> **Overview**
> Fixes a regression where the *after-idle escape hatch* disappeared after repeated background+resume while staying on the same auto-shown NTP.
> 
> `NtpAfterIdleManagerImpl` now **preserves** `_isAfterIdleReturn` across `onClose()` and non-fresh `onOpen()` (background+resume), while still resetting it on **fresh launches** to avoid stale state; `pendingAfterIdle` continues to be cleared on `onClose()`.
> 
> App-launch routing is adjusted to avoid leaking a stale pending flag: `FirstScreenHandlerImpl` only triggers `onIdleReturnTriggered()` synchronously when `isFreshLaunch` and the current tab is already an NTP, and `ShowOnAppLaunchOptionHandlerImpl` only triggers `onIdleReturnTriggered()` when it will actually add/select a new NTP tab (not when already on an NTP). Tests are updated/expanded to cover these scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 935ec3cb36df4ddd0fe925d1fce6bec9eec767c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/generalsettings/showonapplaunch/FirstScreenHandler.kt
+++ b/app/src/main/java/com/duckduckgo/app/generalsettings/showonapplaunch/FirstScreenHandler.kt
@@ -62,12 +62,18 @@ class FirstScreenHandlerImpl @Inject constructor(
 ) : BrowserLifecycleObserver {
 
     override fun onOpen(isFreshLaunch: Boolean) {
-        // Notify the NtpAfterIdleManager synchronously when the currently selected tab is already
-        // an NTP: BrowserViewModel's flowSelectedTab subscription can fire onNtpShown immediately
-        // on activity recreation, and the async handler path below doesn't run in time to classify
-        // it. Gated on "already on NTP" so LastOpenedTab/SpecificPage users on a URL tab don't
-        // leave a stale pendingAfterIdle flag behind for a later user-initiated NTP.
-        if (androidBrowserConfigFeature.showNTPAfterIdleReturn().isEnabled() &&
+        // Notify the NtpAfterIdleManager synchronously on a fresh launch when the currently
+        // selected tab is already an NTP: BrowserViewModel's flowSelectedTab subscription will
+        // fire onNtpShown immediately on activity recreation, and the async handler path below
+        // doesn't run in time to classify it.
+        //
+        // Restricted to isFreshLaunch=true: on plain background+resume, NtpAfterIdleManager
+        // preserves the prior session's classification, so the existing _isAfterIdleReturn value
+        // is correct without a fresh trigger. Setting pendingAfterIdle here would leak — no
+        // onNtpShown fires (same NTP tab), and the next user action that DOES show an NTP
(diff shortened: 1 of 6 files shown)


YoussefKeyrouz (approved):
LGTM!

PR #8520, opened 2026-05-11
Fix WSJ Custom Tab redirect loop

kzar:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202552961248957/task/1211781703927439

### Description
If the user sets the DuckDuckGo browser as the default, and then taps to open an article link inside the WSJ app, they can get trapped in an endless redirect loop, where the WSJ app opens the DDG browser, which in turn opens the WSJ app again, etc.

Let's fix that by mirroring Chromium's behaviour[1]:

1. Block HTTP App Link hand-off unless the user initiated the navigation (e.g. by clicking on a link, or entering a URL in the URL bar). This is what prevents the WSJ redirection loop, since the navigations aren't user initiated, and so the article is loaded inside the DDG Custom Tab instead.
2. Add a "trusted caller" carve-out, to avoid breaking some OAuth flows. If an app opens a Custom Tab, App Links that point back to the same app should be allowed without user interaction.

Notes:
 - The "trusted caller" carve-out doesn't apply for the WSJ app, since it does not take care to propagate its session token into its launch intent (skips `CustomTabsIntent.Builder(session)`). That is the only thing preventing the redirect loop also happening in Chromium.
 - The `isUserQuery()` check allows navigations that were initiated by the user via the address bar to then be opened in an app.
 - Non-HTTP App Links are handled via `handleNonHttpAppLink` which isn't changed here. Chromium seems to force a user prompt for non-user-initiated non-HTTP navigations, we could consider doing the same in the future.

### Steps to test this PR

1. DDG as default browser, Settings → Permissions → "Open Links in Apps" = "Always"

#### Check that the WSJ redirection loop is fixed

1. Install the WSJ Android app, and sign in to a paid account.
2. In the WSJ app, open articles until you find one that opens in the custom tab (with DDG browser).
3. Ensure you can then read the article OK, instead of being redirected in an infinite loop.

#### Check that OAuth flows still work

1. Install this [Android OAuth demo app](https://testpages.kzar.co.uk/oauth/android-demo.apk).
2. Launch the app, and ensure "Require user gesture for auth" is disabled, and "Use browser" is set to "DuckDuckGo (Custom tab)"
3. Press "START AUTHORIZATION".
4. Ensure you are taken back to the "Authorization granted - REFRESH TOKEN / SIGN OUT" screen of the demo app (_not_ left on the ["Authorisation Successful!" webpage](https://testpages.kzar.co.uk/oauth/callback/?code=example%3D%3D&state=example) inside the Custom Tab).

1 - https://source.chromium.org/chromium/chromium/src/+/09aafe21bc1b6c78e92fa4197d4e3c8401caaee9:components/external_intents/android/java/src/org/chromium/components/external_intents/ExternalNavigationHandler.java;l=1230-1235

---

> [!NOTE]
> **Medium Risk**
> Changes core navigation and external app launch behavior in Custom Tabs and normal browsing, gated by a remote toggle but default-on; incorrect gesture or client-package handling could break OAuth or legitimate deep links.
> 
> **Overview**
> Fixes **Custom Tab redirect loops** (e.g. WSJ ↔ browser) by tightening when HTTP App Links hand off to native apps.
> 
> **App link policy:** When remote flag `customTabEndlessLoopFix` is on (default), **HTTP app links no longer launch without a user gesture**, unless the navigation target matches the **verified Custom Tab client package** (trusted-caller OAuth carve-out). `BrowserWebViewClient` forwards `WebResourceRequest.hasGesture()` through `WebViewClientListener` into `AppLinksHandler`, which now takes the full `AppLink` plus optional `clientPackage`.
> 
> **Custom Tab plumbing:** `BrowserTabViewModel` replaces a boolean custom-tab flag with `CustomTabContext(clientPackage)`. The calling package is resolved at dispatch time via a new **`CustomTabsSessionRegistry`** (sessions recorded in `DuckDuckGoCustomTabService.newSession`, cleared on `cleanUpSession`) and passed through `IntentDispatcherActivity` → `CustomTabActivity` → fragment args. **`CustomTabDetector` updates from `IntentDispatcherViewModel` are removed.**
> 
> **Module:** `custom-tabs-api` becomes an Android library exposing the registry API; impl adds `RealCustomTabsSessionRegistry` and DI on the Custom Tabs service.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51f83f13b32e301fc998972ac74d4cc83ae36b0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


custom-tabs/custom-tabs-api/src/main/java/com/duckduckgo/customtabs/api/CustomTabsSessionRegistry.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Tracks verified Custom Tabs sessions (token -> calling package), populated from
+ * DuckDuckGoCustomTabService.newSession() and read by IntentDispatcherViewModel
+ * to support handleAppLink's trusted-caller carve-out.
+ */
+interface CustomTabsSessionRegistry {
+
+    /** Records a verified (session token -> calling package) binding. Called from newSession(). */
+    fun recordSession(sessionToken: CustomTabsSessionToken, packageName: String)
```
joshliebe: As discussed, this and `clearSession` should be removed from the API

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3550,13 +3492,21 @@ class BrowserTabViewModel @Inject constructor(
...
-            !appSettingsPreferencesStore.showAppLinksPrompt,
-        ) { appLinkClicked(appLink) }
+        hasGesture: Boolean,
+    ): Boolean {
+        // HTTP navigations shouldn't launch apps unless started with a user gesture. That is unless
+        // the "trusted-caller" carve-out applies - if an app opens a Custom Tab, App Links that
+        // point back to that same app should be allowed even without user interaction.
+        val targetPackage = appLink.appIntent?.component?.packageName ?: appLink.appIntent?.`package`
+        val isTrustedCaller = targetPackage != null && customTab?.clientPackage == targetPackage
+        return (hasGesture || appLinksHandler.isUserQuery() || isTrustedCaller) &&
```
joshliebe: This logic should be moved into the `appLinksHandler` (it’s checking `appLinksHandler.isUserQuery()` in order to call `appLinksHandler.handleAppLink()`). 

Pass in `hasGesture`, `appLink` (instead of `appLink.uriString`) and `clientPackage`.

joshliebe: Probably a good idea to add a kill switch as well just in case

custom-tabs/custom-tabs-api/src/main/java/com/duckduckgo/customtabs/api/CustomTabsSessionRegistry.kt
anikiki: Updated 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
anikiki: Updated 👍

anikiki: Added 👍

joshliebe (approved):
Changes LGTM, thanks for addressing the comments 🙂

PR #8514, opened 2026-05-11
Onboarding Brand Design Update: Add Default Browser page

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207908166761516/task/1212699268790172?focus=true

### Description

Adds a new brand-design **Default Browser** page to the onboarding flow, gated behind the `onboardingBrandDesignUpdate` feature flag.

The page is only inserted into the flow when `shouldShowDefaultBrowserPage()` returns true — i.e. on devices where the system role-manager / default-browser chooser isn't available and the user has to navigate to system settings manually. This mirrors the gating of the legacy `DefaultBrowserPage`; it's just the surface that's new.

**With the flag on:**
- New `BrandDesignUpdateDefaultBrowserPage` fragment with the 2026 brand-design layout — new header illustration, restyled title / subtitle, primary "Set as default browser" + secondary "Continue" buttons.
- Portrait and landscape layouts wired up; header illustration adapts to available space (caps at design width on tablets, shrinks to a minimum height when content doesn't fit).
- New illustration drawables shipped as WebP at every density.
- `BrandDesignUpdateDefaultBrowserPageBlueprint` is appended to `buildBrandDesignUpdatePageBlueprints()` whenever `shouldShowDefaultBrowserPage()` is true.

**With the flag off:**
- No change to the legacy path — `buildPageBlueprints()` still emits the existing `DefaultBrowserBlueprint`, and the original `DefaultBrowserPage` renders exactly as before.

### Steps to test this PR

[Designs](https://www.figma.com/design/YPE94Xkcrk2uqiF2l4VmSv/Onboarding--2026-?node-id=20832-106044&m=dev)

> ⚠️ This page only renders on devices where the system role-manager isn't used (API ≤ 28). Use a Pie (API 28) or older emulator / device to exercise it.

To see these changes, patch:

```
Index: app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
[...]

PR #8526, opened 2026-05-12
Fix nativeInput remaining focused after performing an onboarding search action

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1214583146087473?focus=true

### Description

On query submit while native input is enabled, hides both the onboarding intro options bubble ("Try a Search" / "Visit Site") and the native input field UI.

### Steps to test this PR

Applying the patch emulates the future state where the new nativeInput is enabled by default after running onboarding.

To turn this off for testing flag off state, remove changes to `DuckChatDataStore`

```
Index: duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/store/DuckChatDataStore.kt
IDEA additional info:
<+>UTF-8
diff --git a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/store/DuckChatDataStore.kt b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/store/DuckChatDataStore.kt
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/store/DuckChatDataStore.kt	(revision 0120753cfc874900c092861cd767a423c06b88d4)
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/store/DuckChatDataStore.kt	(date 1778573552758)
@@ -253,9 +253,9 @@
 
     private val nativeInputFieldUserSettingEnabled: StateFlow<Boolean> =
         store.data
-            .map { prefs -> prefs[DUCK_AI_NATIVE_INPUT_FIELD_SETTING] ?: false }
+            .map { prefs -> prefs[DUCK_AI_NATIVE_INPUT_FIELD_SETTING] ?: true }
             .distinctUntilChanged()
-            .stateIn(appCoroutineScope, SharingStarted.Eagerly, false)
+            .stateIn(appCoroutineScope, SharingStarted.Eagerly, true)
 
     private val duckChatShowInBrowserMenu: StateFlow<Boolean> =
         store.data
@@ -454,7 +454,7 @@
         store.data.firstOrNull()?.let { it[DUCK_AI_NATIVE_INPUT_FIELD_SETTING] } ?: false
 
     override suspend fun isNativeInputFieldUserSettingEnabled() =
-        store.data.firstOrNull()?.let { it[DUCK_AI_NATIVE_INPUT_FIELD_SETTING] } ?: false
+        store.data.firstOrNull()?.let { it[DUCK_AI_NATIVE_INPUT_FIELD_SETTING] } ?: true
 
     override suspend fun setChatSuggestionsUserSetting(enabled: Boolean) {
         store.edit { prefs -> prefs[DUCK_AI_CHAT_SUGGESTIONS_USER_SETTING] = enabled }
```

_Submitting via option tap_
- [ ] Fresh install (patch applied), walk through onboarding until the "Try a Search" bubble appears on the NTP
- [ ] Ensure the native input field is open and visible at the top of the screen
- [ ] Tap one of the suggested-search chips in the bubble
- [ ] Confirm the native input field hides on tap — it should NOT remain visible on top of the SERP that loads

_Submitting a custom query typed into the native input_
- [ ] Fresh install (patch applied), walk through onboarding until the "Try a Search" bubble appears on the NTP
- [ ] Tap the native input field, type a custom query, and submit
- [ ] Confirm the native input field hides on submit — it should NOT remain visible on top of the SERP

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2864,6 +2864,7 @@ class BrowserTabFragment :
             is Command.HideOnboardingDaxDialog -> hideOnboardingDaxDialog(it.onboardingCta)
             is Command.HideBrokenSitePromptCta -> hideBrokenSitePromptCta(it.brokenSitePromptDialogCta)
             is Command.HideOnboardingDaxBubbleCta -> hideOnboardingDaxBubbleCta(it.daxBubbleCta)
+            is Command.HideDaxBubbleCtaOnSubmit -> hideDaxBubbleCta(it.daxBubbleCta)
             is Command.ShowRemoveSearchSuggestionDialog -> showRemoveSearchSuggestionDialog(it.suggestion)
             is Command.AutocompleteItemRemoved -> autocompleteItemRemoved()
             is Command.OpenDuckPlayerSettings -> globalActivityStarter.start(binding.root.context, DuckPlayerSettingsNoParams)
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2865,6 +2865,7 @@ class BrowserTabFragment :
             is Command.HideBrokenSitePromptCta -> hideBrokenSitePromptCta(it.brokenSitePromptDialogCta)
             is Command.HideOnboardingDaxBubbleCta -> hideOnboardingDaxBubbleCta(it.daxBubbleCta)
             is Command.HideDaxBubbleCtaOnSubmit -> hideDaxBubbleCta(it.daxBubbleCta)
+            is Command.HideNativeInputOnOnboardingSubmit -> nativeInputManager.hideNativeInput(animate = false, isNavigation = true)
             is Command.ShowRemoveSearchSuggestionDialog -> showRemoveSearchSuggestionDialog(it.suggestion)
             is Command.AutocompleteItemRemoved -> autocompleteItemRemoved()
             is Command.OpenDuckPlayerSettings -> globalActivityStarter.start(binding.root.context, DuckPlayerSettingsNoParams)
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2695,6 +2695,9 @@ class BrowserTabFragment :
             }
 
             is Command.HideKeyboard -> {
+                if (nativeInputManager.isNativeInputEnabled()) {
+                    nativeInputManager.hideNativeInput(animate = false, isNavigation = true)
+                }
                 hideKeyboard()
             }
 
@@ -2865,7 +2868,6 @@ class BrowserTabFragment :
             is Command.HideBrokenSitePromptCta -> hideBrokenSitePromptCta(it.brokenSitePromptDialogCta)
             is Command.HideOnboardingDaxBubbleCta -> hideOnboardingDaxBubbleCta(it.daxBubbleCta)
             is Command.HideDaxBubbleCtaOnSubmit -> hideDaxBubbleCta(it.daxBubbleCta)
-            is Command.HideNativeInputOnOnboardingSubmit -> nativeInputManager.hideNativeInput(animate = false, isNavigation = true)
             is Command.ShowRemoveSearchSuggestionDialog -> showRemoveSearchSuggestionDialog(it.suggestion)
             is Command.AutocompleteItemRemoved -> autocompleteItemRemoved()
             is Command.OpenDuckPlayerSettings -> globalActivityStarter.start(binding.root.context, DuckPlayerSettingsNoParams)
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -2867,7 +2867,6 @@ class BrowserTabFragment :
             is Command.HideOnboardingDaxDialog -> hideOnboardingDaxDialog(it.onboardingCta)
             is Command.HideBrokenSitePromptCta -> hideBrokenSitePromptCta(it.brokenSitePromptDialogCta)
             is Command.HideOnboardingDaxBubbleCta -> hideOnboardingDaxBubbleCta(it.daxBubbleCta)
-            is Command.HideDaxBubbleCtaOnSubmit -> hideDaxBubbleCta(it.daxBubbleCta)
             is Command.ShowRemoveSearchSuggestionDialog -> showRemoveSearchSuggestionDialog(it.suggestion)
             is Command.AutocompleteItemRemoved -> autocompleteItemRemoved()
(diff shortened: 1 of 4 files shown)


mikescamell:
* **#8526** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8526?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8526?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8515** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8515?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8514** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8514?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8484** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8484?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1382,10 +1389,18 @@ class BrowserTabViewModel @Inject constructor(
...
                 }
+                dismissDaxBubbleCtaOnSubmit(cta as DaxBubbleCta)
             }
 
             is DaxBubbleCta.DaxIntroVisitSiteOptionsCta,
             is DaxVisitSiteOptionsBrandDesignUpdateBubbleCta,
+            -> {
+                if (!ctaViewModel.isSuggestedSiteOption(query)) {
+                    pixel.fire(ONBOARDING_VISIT_SITE_CUSTOM, type = Unique())
+                }
```
LukasPaczos: Is this a new pixel or something we unintentionally removed before?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -5008,6 +5023,13 @@ class BrowserTabViewModel @Inject constructor(
         command.value = HideOnboardingDaxDialog(cta)
     }
 
+    private fun dismissDaxBubbleCtaOnSubmit(cta: DaxBubbleCta) {
+        if (!nativeInputFieldEnabled) return
+        ctaViewState.value = currentCtaViewState().copy(cta = null)
+        command.value = HideDaxBubbleCtaOnSubmit(cta)
+        command.value = HideNativeInputOnOnboardingSubmit
```
LukasPaczos: I think this is the wrong fix.

`onUserSubmittedQuery` already calls `command.value = HideKeyboard`, which should hide the keyboard and drop focus from the omnibar:

https://github.com/duckduckgo/Android/blob/118eeba0e62c1b853fcd814e8dacdcd98a8ba7da/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt#L1404

The real problem is that this isn't wired up to the native input implementation. Adding an onboarding-specific workaround papers over a faulty base flow rather than fixing it.

We should have a single flow for dropping omnibar/input-field focus, and that flow should fork internally based on whether the native input is enabled - not a separate, independent path introduced everywhere we need to hide the keyboard.

Also noting that the original issue of native input retaining focus when onboarding options are clicked is also true for the production onboarding, not only brand design variant, so it should be fixed at the source.

cc @malmstein for visibility

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -5008,6 +5023,13 @@ class BrowserTabViewModel @Inject constructor(
         command.value = HideOnboardingDaxDialog(cta)
     }
 
+    private fun dismissDaxBubbleCtaOnSubmit(cta: DaxBubbleCta) {
+        if (!nativeInputFieldEnabled) return
+        ctaViewState.value = currentCtaViewState().copy(cta = null)
+        command.value = HideDaxBubbleCtaOnSubmit(cta)
```
LukasPaczos: We already have:

https://github.com/duckduckgo/Android/blob/118eeba0e62c1b853fcd814e8dacdcd98a8ba7da/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt#L1374-L1376

Why do we need to manually hide the CTA?

mikescamell: I flagged this [here](https://app.asana.com/1/137249556945/task/1214310944615468/comment/1214456458032274?focus=true)

When opening the native input field later on in the process and closing, the first dialog show in the browser was briefly seen. So we need to hide the dialog.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
mikescamell: It's the same pixel that is fired in ln 1408, but as we need to hide the CTA for the visit sites option then we needed to move it here as well.

LukasPaczos: I see now, thanks 👍

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
LukasPaczos: @mikescamell similarly to https://github.com/duckduckgo/Android/pull/8526#discussion_r3273693932, I also don't think that peeking CTA is something that should be worked around specifically for this case. It can be reproduced effectively on any CTA right now and the fix is tracked in [here](https://app.asana.com/1/137249556945/project/1214157224317277/task/1214882981544707?focus=true) as far as I can tell.

Example:

https://github.com/user-attachments/assets/3dd4db08-1f0a-4b0d-8293-7001e9f3a8a2

mikescamell: Dropped those changes

LukasPaczos:
Note: renamed the PR and changed description to match current scope, which is to only fix the native input persisting after search action.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -2695,6 +2695,9 @@ class BrowserTabFragment :
             }
 
             is Command.HideKeyboard -> {
+                if (nativeInputManager.isNativeInputEnabled()) {
+                    nativeInputManager.hideNativeInput(animate = false, isNavigation = true)
```
YoussefKeyrouz, line 2699: Hide native input does more than hiding the keyboard. It dismisses the whole widget and animates the input and the browser tab up (changing their size/frames), and might have side effects. This does not cause a problem right now but it might in the future add confusion/regression if someone is firing the “HideKeyboard” command, expecting only the keyboard to hide.

I would rather have a separate command to hide the native input and call it in the locations where we do want to intentionally dismiss it (e.g. during onboarding submission), rather than on any keyboard hide.

To give you more context: we had similar implementation in the past (not in this exact location, in the onKeyboardHidden callback), but we removed it. The reason was that the native input was dismissing in places we didn’t want it to be dismissed, for example when scrolling a list we want to hide the keyboard only but not animate the whole view up and dismiss the input.

LukasPaczos:
Closing in favor of https://github.com/duckduckgo/Android/pull/8692.

PR #8540, opened 2026-05-13
Compose: Add DaxAlertDialog to ADS

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202857801505092/task/1211659112661189?focus=true

### Description

Adds the Compose `DaxAlertDialog` family to the Android Design System, replacing the legacy XML `*AlertDialogBuilder` patterns. Slot-based API: `DaxAlertDialog` is the canonical entry point, with small helpers for the common content/buttons shapes. The Compose set is intentionally smaller than the View set — slots collapse what previously needed dedicated dialog classes.

**New public composables:**
- `DaxAlertDialog` — slot-based base (title + headerImage + message + content + buttons)
- `TextAlertDialog` — convenience wrapper for the common title + message + primary/ghost buttons shape (and optional checkbox)
- `DaxRadioOptions` — radio button list helper, designed to live in the `content` slot
- `DaxStackedButtons` — vertically-stacked button helper, designed to live in the `buttons` slot

**Migration mapping (View → Compose):**
- `TextAlertDialogBuilder` → `TextAlertDialog` (or `DaxAlertDialog` directly for non-default button styles)
- `StackedAlertDialogBuilder` → `DaxAlertDialog` + `DaxStackedButtons`
- `RadioListAlertDialogBuilder` → `DaxAlertDialog` + `DaxRadioOptions`
- `CustomAlertDialogBuilder` → `DaxAlertDialog` with arbitrary composables in the `content` slot

**Design alignment** (see [the Document any issues between View/Figma task](https://app.asana.com/1/137249556945/project/1202857801505092/task/1214523588264769?focus=true) for the full delta list)

### Steps to test this PR

_Internal showcase fragment_
- [ ] Open the design system internal demo → Dialogs section
- [ ] Verify each Compose dialog showcase renders alongside its View counterpart in both light and dark themes
- [ ] Verify the Compose variant matches the View variant visually for: text + image, text, text destructive, text single button, text cancellable, text one button, text with checkbox, stacked with image, stacked 4 buttons, stacked destructive, radio, radio destructive

### UI changes
| Before  | After |
| ------ | ----- |
| (View-system XML dialogs — `TextAlertDialogBuilder`, `StackedAlertDialogBuilder`, `RadioListAlertDialogBuilder`) | (Compose `DaxAlertDialog` + helpers; the internal showcase pairs each Compose variant alongside its View counterpart) |

---

> [!NOTE]
> **Medium Risk**
> Adds new public Compose dialog APIs and new lint errors banning raw Material3 `AlertDialog`/`Surface`, which can break builds and affect UI consistency if behavior/styling differs from legacy dialogs.
> 
> **Overview**
> **Adds a new Compose alert dialog family to the Android Design System.** Introduces slot-based `DaxAlertDialog` plus helpers `DaxTextAlertDialog` (common title/message/buttons shape with optional checkbox), `DaxRadioOptions` (scrolling single-choice list), and `Dax*StackedButtons` (primary/destructive stacked actions), along with a small button text alignment tweak.
> 
> **Updates the internal showcase** to render Compose dialog variants alongside existing View-based dialogs (including radio, destructive, stacked, and checkbox cases) and expands the demo layout with ComposeView placeholders and platform labels.
> 
> **Enforces adoption with lint.** Registers new lint rules that error on direct usage of Material3 `AlertDialog`/`BasicAlertDialog` and `Surface` outside the design-system modules, with accompanying unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c722d415db47fce7ba29080b97b99f5465e5fea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
```diff
@@ -0,0 +1,200 @@
...
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1214735717450949
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=685-956&t=DvV3Fi7Mi45nLle2-4
+ */
+@Composable
+fun DaxAlertDialog(
+    onDismissRequest: () -> Unit,
+    title: String,
+    modifier: Modifier = Modifier,
+    message: (@Composable () -> Unit)? = null,
+    @DrawableRes headerImage: Int? = null,
```
catalinradoiu: Do we want to enforce using a drawable? Can we pass a painter instead in case we want to use an image from the server for example?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1214735717450949
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=685-956&t=DvV3Fi7Mi45nLle2-4
+ */
+@Composable
+fun DaxAlertDialog(
+    onDismissRequest: () -> Unit,
+    title: String,
+    modifier: Modifier = Modifier,
+    message: (@Composable () -> Unit)? = null,
```
catalinradoiu, line 82: Should we make the message a text instead of using it as a slot? Looks like we’re always showing a text and not something else there. If needed we can change this API later. What do you think?

Edit: I see we have `TextAlertDialog` for that, so no need to change this.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
```diff
@@ -0,0 +1,200 @@
...
+@Composable
+internal fun DaxAlertDialogContent(
+    title: String,
+    modifier: Modifier = Modifier,
+    message: (@Composable () -> Unit)? = null,
+    @DrawableRes headerImage: Int? = null,
+    content: (@Composable () -> Unit)? = null,
+    buttons: @Composable () -> Unit,
+) {
+    Surface(
```
catalinradoiu: I think we should use `DaxSurface` instead of `Surface`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
```diff
@@ -0,0 +1,200 @@
...
+        modifier = modifier,
+        shape = DuckDuckGoTheme.shapes.medium,
+        color = DuckDuckGoTheme.colors.backgrounds.surface,
+    ) {
+        Column(
+            modifier = Modifier.padding(24.dp),
```
catalinradoiu: Should we use dimension resources for these? Applies to other changes in this PR as well

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
```diff
@@ -0,0 +1,200 @@
...
+        ) {
+            if (headerImage != null) {
+                Image(
+                    painter = painterResource(id = headerImage),
+                    contentDescription = null,
+                    modifier = Modifier
+                        .align(Alignment.CenterHorizontally)
+                        .size(24.dp),
```
catalinradoiu, line 133: I see the view version has a different size (Figma is the right one so the view version is not correct). Can you add this in the task for documenting the discrepancies between the view and the compose version?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxStackedButtons.kt
```diff
@@ -0,0 +1,135 @@
...
+ * - All buttons render as ghost.
+ * - When [destructiveButtonIndex] is non-null, that button uses the destructive ghost style and
+ *   the rest use the muted alt-ghost style.
+ * - End-aligned.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1214735768823555
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=685-956&t=DvV3Fi7Mi45nLle2-4
+ */
+@Composable
+fun DaxStackedButtons(
```
catalinradoiu: In Figma I see we have a variatn that uses `DaxPrimaryButton`. Is that no longer needed?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/TextAlertDialog.kt
```diff
@@ -0,0 +1,239 @@
...
+ */
+@Composable
+fun TextAlertDialog(
+    onDismissRequest: () -> Unit,
+    title: String,
+    positiveButtonText: String,
+    onPositiveClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    message: String? = null,
+    @DrawableRes headerImage: Int? = null,
```
catalinradoiu: Do we want to use a painter or do you want to enforce using a drawable resource?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/TextAlertDialog.kt
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * For anything richer — destructive buttons, custom button styles, radio lists, stacked buttons,
+ * clickable message spans, arbitrary content — use [DaxAlertDialog] directly with the concrete
+ * `Dax*Button` composables from `com.duckduckgo.common.ui.compose.button`.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1214735774211999
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=685-956&t=DvV3Fi7Mi45nLle2-4
+ */
+@Composable
+fun TextAlertDialog(
```
catalinradoiu: We should name it `DaxTextAlertDialog` to keep the naming consistent

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
```diff
@@ -0,0 +1,200 @@
...
+                        .padding(top = 8.dp),
+                ) {
+                    content()
+                }
+            }
+
+            FlowRow(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(top = 16.dp),
```
catalinradoiu: Should we use a `Spacer` instead?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxStackedButtons.kt
```diff
@@ -0,0 +1,135 @@
...
+    buttonTitles: ImmutableList<String>,
+    onButtonClick: (index: Int) -> Unit,
+    modifier: Modifier = Modifier,
+    destructiveButtonIndex: Int? = null,
```
catalinradoiu: Not sure I like this approach. I see in Figma the the destructive button is always the first one. Is there a reason why we want to move this to another position? 
Also we’re missing the `DaxPrimaryButton` variant. I think it would be a better approach to split this into 2 composables (`DaxDestructiveButtonStacked` and `DaxPrimaryButtonStacked` maybe?)

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
0nko: Yeah, that's a good call.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
0nko: I missed that, yes. I will also add a lint rule to prevent this from happening.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
0nko: I thought in Compose it's standard to use constants for paddings/margins but we can use them.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
0nko: I thought I added it already but maybe I missed it. I'll put it there 👍.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
mikescamell: I think we might need to revisit the whole dimensions/keyline thing later on, but for now lets just continue to use them while we have the View world

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/button/DaxButton.kt
```diff
@@ -200,6 +201,7 @@ private fun DaxButtonText(
         text = text,
         style = DuckDuckGoTheme.typography.button,
         color = color,
+        textAlign = TextAlign.Center,
```
0nko, line 204: Fixes the alignment for multi-line titles.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxStackedButtons.kt
0nko: Yeah, that's a good improvement. I tried to keep the original approach but this is cleaner + adds the missing variant.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxStackedButtons.kt
0nko: No, that's my mistake. I'll fix it.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/TextAlertDialog.kt
0nko: Yup!

0nko:
@catalinradoiu Thank you for the thorough review! I believe I've addressed all comments. Ready for another round.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxAlertDialog.kt
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            if (content != null) {
+                Column(modifier = Modifier.fillMaxWidth()) {
+                    content()
+                }
+                Spacer(modifier = Modifier.height(dimensionResource(R.dimen.keyline_2)))
+            }
+
+            Spacer(modifier = Modifier.height(dimensionResource(R.dimen.keyline_2)))
```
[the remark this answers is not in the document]
0nko: Buttons always have 8dp above them, this is expected.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxRadioOptions.kt
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * Selection is hoisted — the caller owns [selectedIndex] so it can drive button enablement
+ * or any other UI state. Scrolls independently when the list exceeds 176dp.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1214735717504168
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=685-956&t=DvV3Fi7Mi45nLle2-4
+ */
+@Composable
+fun DaxRadioOptions(
+    optionTitles: ImmutableList<String>,
```
catalinradoiu: I was thinking now if we should pass the tiles name and the listener in a single data class, instead of using the index. I think this logic is fragile and we could end up in a situation where we might do a different action for an option.

How I see it is something like:
`data class DaxOption(tileText: String, onOptionSelected: () -> Unit)`

What do you think?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxStackedButtons.kt
```diff
@@ -0,0 +1,148 @@
...
+ * - Index 0 renders as [DaxDestructiveGhostButton] (the destructive action).
+ * - All other indices render as [DaxDestructiveGhostAltButton] (muted siblings).
+ * - End-aligned.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1214735768823555
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=685-956&t=DvV3Fi7Mi45nLle2-4
+ */
+@Composable
+fun DaxDestructiveStackedButtons(
+    buttonTitles: ImmutableList<String>,
```
catalinradoiu: Same as above, would it be better to wrap the button text and the click listener in a single data class?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/dialog/DaxRadioOptions.kt
0nko: Yeah, that's a good idea!

catalinradoiu (approved):
Looks good to me! 🎉

PR #8549, opened 2026-05-13
Add per-tab NativeInputStateProvider/Publisher

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1174433894299346/task/1214788735007211?focus=true

### Description

First PR in a series moving native input state from `NativeInputHost.getInputState()` (pull-based, single in-memory snapshot in the widget) to a per-tab, push-based store that any consumer can observe.

This PR is **pure plumbing** — no behavior change. No widget or plugin writes or reads through the new interfaces yet.

What lands here:

- **`duck-chat-api`** gets two new interfaces under `com.duckduckgo.duckchat.api.nativeinput`:
  - `NativeInputStateProvider` — read side, exposes `stateForTab(tabId: String): StateFlow<NativeInputState>`.
  - `NativeInputStatePublisher` — write side, with `publish(tabId, state)`, `update(tabId, transform)`, and `clearTab(tabId)`.
  Only the native input widget should depend on the publisher; everything else observes via the provider.
- **`NativeInputState`** (and its nested `InputMode` / `InputContext` / `InputPosition` / `ToggleSelection`) moves from `duckchat-impl` to `duck-chat-api`. A new `NativeInputState.zero(tabId)` factory documents the placeholder default; a TODO marks that the widget will overwrite it once wired to publish on configure (next PR).
- **`RealNativeInputStateStore`** in `duckchat-impl` binds both interfaces at `AppScope`, backed by a `ConcurrentHashMap<String, MutableStateFlow<NativeInputState>>`.
- **`TabDataRepository`** injects the publisher and calls `clearTab(tabId)` at the three delete sites (`delete`, `deleteTabs`, `deleteTabAndSelectSource`) so per-tab state does not leak after a tab closes.

Follow-up PRs:
1. Wire the widget's ViewModel to publish through `NativeInputStatePublisher` and read its own state from the provider; tighten `tabId` to non-null.
2. Migrate the 5 `NativeInputPlugin` ViewModels to observe `stateForTab(tabId)`; route any plugin contributions through new typed `NativeInputHost` methods that the widget translates into publisher updates.
3. Remove `NativeInputHost.getInputState()` and `NativeInputPlugin.getPromptContribution()`.

### Steps to test this PR

_Plumbing only — no user-visible change._
- [ ] CI green: `:duckchat-api:compileDebugKotlin`, `:duckchat-impl:testDebugUnitTest`, `:app:testPlayDebugUnitTest`
[...]

PR #8550, opened 2026-05-13
Widget publishes native input state per tab

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1157893581871903/task/1214790107756091?focus=true

Stacked on top of #8549.

### Description

PR 2 of the native input state refactor — see the parent task for the full plan.

The native input widget is now the single publisher of `NativeInputState` and reads its own state from the per-tab store (`NativeInputStateProvider`) instead of keeping it as a widget-local field. Plugins still won't see this change until PR 3.

Concretely:

- `NativeInputModeWidgetViewModel` injects `NativeInputStatePublisher`. A new `init { … }` block collects `combine(state, activeTabId)` and pushes each value through `publisher.publish(tabId, state.copy(tabId = tabId))` whenever a tabId is set.
- `NativeInputModeWidget` injects `NativeInputStateProvider`. `observeNativeInputState()` now subscribes to `provider.stateForTab(tabId)` instead of `viewModel.state`. `getInputState()` reads `provider.stateForTab(activeTabId).value`.
- `configure()` and `configureContextual()` take a `tabId: String` parameter and start the observation once the tabId is known.
- `tabId` is threaded:
  - `BrowserTabFragment.tabId` → `NativeInputManager.showNativeInput(tabId, …)` → `attachWidget(widgetView, isBottom, tabId)` → `widget.configure(tabId, …)`.
  - `DuckChatContextualFragment` reads its `KEY_DUCK_AI_CONTEXTUAL_TAB_ID` argument once and passes it to `ContextualNativeInputManager.init(tabId, …)` → `widget.configureContextual(tabId)`.
- `viewModel.state` stays as a `SharedFlow<NativeInputState>` so existing ViewModel-level tests can keep asserting on its values; the widget no longer consumes it.

What this PR intentionally does **not** do:

- `NativeInputState.tabId` stays nullable. Tightening to non-null will follow once the publish-on-configure invariant is verified in practice (and we can drop the `zero("__init__")`-style placeholder).
- `NativeInputHost.getInputState()` is still exposed — its implementation now reads from the provider, but the surface remains until plugins are migrated in PR 3 and it's removed in PR 4.
- Plugins still receive the widget as their `NativeInputHost`; nothing about their behaviour changes here.

### Steps to test this PR

- [x] CI green: `:duckchat-impl:testDebugUnitTest`, `:app:testPlayDebugUnitTest`, `spotlessCheck`.
- [x] Existing `NativeInputModeWidgetViewModelTest` cases still pass after the `configure`/`configureContextual` signature changes and the added publisher mock.
- [x] Smoke-test on device:
  - Open Duck.ai input from the NTP, type, switch to the chat tab, submit — behaviour matches develop.
  - Open the omnibar on a browser tab, switch between Search and Duck.ai modes — toggle/position/context behave as before.
  - Open Duck.ai contextual chat from a browser tab — keyboard, attachments, and submit all still work.
  - Background a tab and reopen it — input state for that tab is restored correctly.

### UI changes
None.

---

> [!NOTE]
> **Medium Risk**
> Changes how the native input widget publishes/consumes state by keying everything on `tabId`, which could cause cross-tab state leaks or missing updates if tab IDs aren’t consistently threaded through all entry points.
> 
> **Overview**
> The native input widget now **publishes `NativeInputState` per tab** and **reads state from `NativeInputStateProvider`** instead of relying on widget-local/ViewModel flow state.
> 
> This threads `tabId` through `BrowserTabFragment` → `NativeInputManager.showNativeInput`/`attachWidget` → `NativeInputModeWidget.configure`, and through contextual chat via `DuckChatContextualFragment`/`ContextualNativeInputManager.init` → `configureContextual`. The ViewModel now injects `NativeInputStatePublisher` to push updates for the active tab, and tests are updated for the new `tabId`-required `configure` APIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9dc03e88539734c8b38f69c431450364f355bb2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualFragment.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualFragment.kt
@@ -424,7 +424,9 @@ class DuckChatContextualFragment :
 
         configureBottomSheet(view)
         setupBackPressHandling()
+        val tabId = requireArguments().getString(KEY_DUCK_AI_CONTEXTUAL_TAB_ID)
         contextualNativeInputManager.init(
+            tabId = tabId.orEmpty(),
             card = binding.contextualNativeInputCard,
             widget = binding.contextualNativeInputWidget,
             jsMessaging = contentScopeScripts,
@@ -442,8 +444,8 @@ class DuckChatContextualFragment :
         )
         observeViewModel()
 
-        requireArguments().getString(KEY_DUCK_AI_CONTEXTUAL_TAB_ID)?.let { tabId ->
-            viewModel.onSheetOpened(tabId)
+        tabId?.let {
+            viewModel.onSheetOpened(it)
             setupKeyboardVisibilityListener()
         }
     }
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualFragment.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualFragment.kt
@@ -424,9 +424,11 @@ class DuckChatContextualFragment :
 
         configureBottomSheet(view)
         setupBackPressHandling()
-        val tabId = requireArguments().getString(KEY_DUCK_AI_CONTEXTUAL_TAB_ID)
+        val tabId = requireNotNull(requireArguments().getString(KEY_DUCK_AI_CONTEXTUAL_TAB_ID)) {
+            "DuckChatContextualFragment requires $KEY_DUCK_AI_CONTEXTUAL_TAB_ID argument"
+        }
         contextualNativeInputManager.init(
-            tabId = tabId.orEmpty(),
+            tabId = tabId,
             card = binding.contextualNativeInputCard,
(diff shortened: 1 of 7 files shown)


GerardPaligot (approved):
Work as expected. I tried to reproduce what is reported by Bugbot but I didn't find a way to reproduce. Good for me.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualFragment.kt
```diff
@@ -424,7 +424,9 @@ class DuckChatContextualFragment :
 
         configureBottomSheet(view)
         setupBackPressHandling()
+        val tabId = requireArguments().getString(KEY_DUCK_AI_CONTEXTUAL_TAB_ID)
         contextualNativeInputManager.init(
+            tabId = tabId.orEmpty(),
```
YoussefKeyrouz: Not sure if this could happen, but if the argument is missing, then we will be storing the data under the “” key. So any contextual fragment that might hit this path will share the same stored map entry. Should be have any null enforcement here?

PR #8556, opened 2026-05-14
Make NativeInputState.tabId non-null

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1214802223978853?focus=true

Stacked on #8550. Until that merges, the diff here shows PR 2's commits too; review the top commit (`7d57719`) for the PR 2.5 change in isolation.

### Description

Small follow-up to PR 2. Now that the widget publishes on configure, every `NativeInputState` in the per-tab store is provably associated with a `tabId`, so we tighten the type.

- Drop the `String? = null` default on `NativeInputState.tabId` and `NativeInputState.zero()`.
- ViewModel `state` flow combines `activeTabId.filterNotNull()` so the emitted state carries the active tabId. The publish init block simplifies to a direct collect-and-publish — the separate combine that re-attached the tabId is gone.
- Widget local cache becomes nullable (`NativeInputState?`). Read sites use null-safe access with the same defaults the previous field initializer carried.
- `getInputState()` throws if called pre-configure instead of returning a stale placeholder; the publish-on-configure invariant ensures plugins only reach it after configure.
- Tests: `@Before` calls `configure()` so the state flow has a tabId; the back-buttons test passes an explicit tabId.

### Steps to test this PR

_Native input widget regressions_
- [x] Open the app on a NTP, type a search query, submit — search works as before
- [x] Open a website, focus the omnibar, switch between Search and Duck.ai tabs — toggle works, no flicker
- [x] Open a Duck.ai contextual session from a long-press menu — contextual UI renders correctly
- [x] Toggle "Address bar position" between top and bottom in Settings — widget reshapes correctly on next omnibar focus
- [x] Delete a tab, then the fire button — no leaked state, next tab renders fresh

### UI changes

No UI changes — refactor only.

---

> [!NOTE]
> **Medium Risk**
> Tightens a core UI/state model contract and changes initialization/placeholder behavior; mis-ordered widget configuration could now throw or suppress UI updates until `tabId` is set.
> 
> **Overview**
> Makes `NativeInputState.tabId` mandatory by removing the nullable/default value and updating `zero(tabId)` accordingly.
> 
> Updates `NativeInputModeWidgetViewModel` to only emit `state` once `activeTabId` is non-null (via `activeTabId.filterNotNull()`), embedding the `tabId` in every emitted `NativeInputState` and simplifying publishing to a direct collect-and-publish.
> 
> Adjusts `NativeInputModeWidget` to treat its cached `NativeInputState` as nullable (null-safe reads with safe defaults) and makes `getInputState()` fail fast if called before `configure()`. Tests are updated to pass explicit `tabId`/call `configure()` up front.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96d5ee16626003cd2a116b5c64b03cbbf8550c8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/NativeInputModeWidget.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/NativeInputModeWidget.kt
@@ -204,10 +204,7 @@ class NativeInputModeWidget @JvmOverloads constructor(
     private var attachmentLimitExceeded: Boolean = false
     private var hasAttachments: Boolean = false
     private var supportsUpload: Boolean = true
-    private var nativeInputState: NativeInputState = NativeInputState(
-        inputMode = NativeInputState.InputMode.SEARCH_ONLY,
-        inputContext = NativeInputState.InputContext.BROWSER,
-    )
+    private var nativeInputState: NativeInputState? = null
     private var chatSuggestionsBinding: NativeInputChatSuggestionsBinder.Binding? = null
     private var onShowSuggestions: ((RecyclerView.Adapter<*>) -> Unit)? = null
     private var onClearSuggestions: ((Boolean) -> Unit)? = null
@@ -430,7 +427,7 @@ class NativeInputModeWidget @JvmOverloads constructor(
 
     private fun updateToggleVisibilityForState() {
         if (isDuckAiPageContext()) return
-        applyToggleVisibility(nativeInputState.toggleVisible)
+        applyToggleVisibility(nativeInputState?.toggleVisible ?: false)
     }
 
     private fun applyToggleVisibility(visible: Boolean) {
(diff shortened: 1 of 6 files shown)


duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/NativeInputModeWidget.kt
```diff
@@ -430,7 +427,7 @@ class NativeInputModeWidget @JvmOverloads constructor(
 
     private fun updateToggleVisibilityForState() {
         if (isDuckAiPageContext()) return
-        applyToggleVisibility(nativeInputState.toggleVisible)
+        applyToggleVisibility(nativeInputState?.toggleVisible ?: false)
```
YoussefKeyrouz: We are scattering the default fallbacks inline whenever we read a state variable. Would be harder to maintain if we add more call sites or if we do future modifications. I think a cleaner approach would be to have one default state. Something like `val DEFAULT_STATE = NativeInputState(SEARCH_ONLY, BROWSER, TOP, tabId = “")` (or similar) and use its values as fallbacks wherever needed. 

Not blocking for this PR but would be good to have while it’s fresh.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/NativeInputModeWidget.kt
```diff
@@ -985,7 +985,8 @@ class NativeInputModeWidget @JvmOverloads constructor(
     }
 
     override fun getInputState(): NativeInputState =
-        activeTabId?.let { nativeInputStateProvider.stateForTab(it).value } ?: nativeInputState
+        activeTabId?.let { nativeInputStateProvider.stateForTab(it).value }
+            ?: error("getInputState called before widget was configured")
```
YoussefKeyrouz, line 992: If we’re throwing in the case where the state is not configured, do we need to have any nullable state? it could be non nullable and that will save us a lot of `:?` checks.  right?

YoussefKeyrouz (approved):
Tested, I didn’t see any regression. Left a few comments, none of them are blocking but good to have.

PR #8566, opened 2026-05-14
Add getEntitlements API method

YoussefKeyrouz:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214808964012146?focus=true

### Description

Adds `Subscriptions.getEntitlements(): Flow<Set<Entitlement>>` to the public API so consumers can read the live entitlement set (each carrying both `name` and `product`).

Changes:
- New `getEntitlements()` method on the `Subscriptions` interface, backed by a new `entitlementSet: Flow<Set<Entitlement>>` on `SubscriptionsManager`. Emits alongside the existing `entitlements: Flow<List<Product>>` from the same trigger points (`emitEntitlementsValues()`).
- `Entitlement` data class moved from `subscriptions-impl/model/` to `subscriptions-api/model/` so it's part of the public surface. All impl-side imports repointed.
- `getEntitlementStatus()` annotated `@Deprecated`.
- Unit tests added in `RealSubscriptionsManagerTest` covering the active-subscription and inactive-subscription emissions.

Design and API proposal:
- [Tech Design](https://app.asana.com/1/137249556945/project/481882893211075/task/1214784714729690)
- [API Proposal](https://app.asana.com/1/137249556945/project/1202552961248957/task/1214804968952054?focus=true)

### Steps to test this PR
- [ ] Smoke: Sign in with a Plus or Pro account, confirm subscription status and features renders as before (existing `getEntitlementStatus()` callers unchanged).
- [ ] Sign out, confirm no regressions on the subscription settings screen.

### UI changes
No UI changes - public API only.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it expands the public `Subscriptions` API (interface method addition + deprecation) and changes entitlement emission plumbing, which could affect downstream implementers/consumers and flow behavior if mismatched.
> 
> **Overview**
> Adds a new public `Subscriptions.getEntitlements(): Flow<Set<Entitlement>>` API that exposes the live *raw* entitlement set (including both `name` and `product`) and deprecates `getEntitlementStatus()`.
> 
> Moves the `Entitlement` model into `subscriptions-api` and wires a new `SubscriptionsManager.entitlementSet` flow through `RealSubscriptions`/`SubscriptionsDummy`, emitting alongside the existing `entitlements` list and resetting to empty on sign-out; updates imports across impl/services/auth and adds unit tests covering entitlement-set emission scenarios.
> 
[...]

PR #8568, opened 2026-05-14
Fix Duck.ai model picker for Pro subscribers

YoussefKeyrouz:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214157224317277/task/1214577731172288?focus=true 

### Description

Fixes the Duck.ai model picker for Pro subscribers, who were incorrectly seeing only Plus/Free models. Migrates `DuckAiModelManager` from the old `getEntitlementStatus()` + `getAvailableProducts()` combination to the new `Subscriptions.getEntitlements()` flow added in the previous PR. The new API exposes the per-entitlement `name` (`"plus"` / `"pro"`), which is what distinguishes the two paid tiers.

Changes:
- `DuckAiModelManager.init` now collects `subscriptions.getEntitlements()` for the "entitlements changed → refresh models" trigger (same semantics as before, just the new flow).
- `resolveUserTier()` reads the current entitlement snapshot via `.first()`, finds the entitlement whose `product == Product.DuckAiPlus.value`, and resolves its `name` to a `UserTier` via `UserTier.from(raw)`.
- Tests updated

Depends on the previous PR being merged: https://github.com/duckduckgo/Android/pull/8566

### Steps to test this PR

Pro subscriber path (the bug being fixed)
- [x] Sign into a Pro account that has Duck.ai Pro entitlement.
- [x] Open the Duck.ai model picker.
- [x] Confirm Pro-tier models are listed and selectable (previously they were not).
- [x] Confirm a Plus-tier model is still selectable (Plus access should not regress).
- [x]  Select Pro model (Opus 4.6 and start chat.

Plus and Free subscriber paths (regression checks)_
- [x] Sign in with a Plus account, confirm Plus models are shown and Pro-only models are gated.
- [x] Sign out, confirm the picker shows only Free-tier models.

Reactivity
- [x] While Duck.ai is open, sign out from settings and return to Duck.ai. 
- [x] Start a new chat. 
- [x] The picker should refresh to free-tier-only without restarting the app.

### UI changes
| Before  | After |
| ------ | ----- |
|rn_image_picker_lib_temp_e529b4d9-6b7d-4970-8d21-49786c0a958a|rn_image_picker_lib_temp_e34e70af-f0e7-41d1-acc8-fe6682118a24|

---

> [!NOTE]
> **Medium Risk**
> Updates tier resolution and refresh triggers to use the new `getEntitlements()` flow; mistakes here could mis-gate paid models/limits for subscribers but scope is limited to Duck.ai model selection logic.
> 
> **Overview**
> Fixes Duck.ai tier detection so **Pro subscribers** are correctly treated as `UserTier.PRO` (and thus see Pro-tier models).
> 
> `RealDuckAiModelManager` now listens to `subscriptions.getEntitlements()` for reactive model refresh, and `resolveUserTier()` derives the tier from the Duck.ai entitlement’s `name` (e.g., `plus`/`pro`) instead of inferring from available products.
> 
> Adds `UserTier.from(raw)` and updates tests to the new entitlements API, including new coverage for Pro, unknown tier names, empty/multiple entitlements, and the updated entitlement-change trigger.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16175cffdd3df09c6b8609a9cba5fea866147dd7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/DuckAiModelManager.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/DuckAiModelManager.kt
@@ -31,6 +31,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
@@ -90,7 +91,7 @@ class RealDuckAiModelManager @Inject constructor(
             } catch (e: Exception) {
                 logcat { "Duck.ai Model Manager: failed to restore cached selection: ${e.message}" }
             }
-            subscriptions.getEntitlementStatus()
+            subscriptions.getEntitlements()
                 .distinctUntilChanged()
                 .collect {
                     logcat { "Duck.ai Model Manager: entitlements changed, re-fetching models" }
@@ -217,14 +218,10 @@ class RealDuckAiModelManager @Inject constructor(
 
     private suspend fun resolveUserTier(): UserTier {
         return try {
-            val status = subscriptions.getSubscriptionStatus()
-            if (!status.isActiveOrWaiting()) return UserTier.FREE
-
-            val products = subscriptions.getAvailableProducts()
-            when {
-                products.contains(Product.DuckAiPlus) -> UserTier.PLUS
-                else -> UserTier.FREE
-            }
+            if (!subscriptions.getSubscriptionStatus().isActiveOrWaiting()) return UserTier.FREE
+            val entitlements = subscriptions.getEntitlements().firstOrNull().orEmpty()
+            val duckAiTier = entitlements.firstOrNull { it.product == Product.DuckAiPlus.value }?.name
+            UserTier.from(duckAiTier) ?: UserTier.FREE
         } catch (e: Exception) {
             logcat { "Duck.ai Model Manager: failed to resolve user tier, defaulting to FREE: ${e.message}" }
             UserTier.FREE
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
@@ -115,6 +115,11 @@ enum class UserTier(val rawValue: String) {
     FREE("free"),
     PLUS("plus"),
     PRO("pro"),
+    ;
+
+    companion object {
+        fun from(raw: String?): UserTier? = entries.firstOrNull { it.rawValue == raw }
+    }
 }
 
 enum class ModelProvider {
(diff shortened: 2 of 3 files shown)


YoussefKeyrouz:
* **#8568** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8568?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8568?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8566** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/DuckAiModelManager.kt
```diff
@@ -217,14 +218,10 @@ class RealDuckAiModelManager @Inject constructor(
...
-            val status = subscriptions.getSubscriptionStatus()
-            if (!status.isActiveOrWaiting()) return UserTier.FREE
-
-            val products = subscriptions.getAvailableProducts()
-            when {
-                products.contains(Product.DuckAiPlus) -> UserTier.PLUS
-                else -> UserTier.FREE
-            }
+            if (!subscriptions.getSubscriptionStatus().isActiveOrWaiting()) return UserTier.FREE
+            val entitlements = subscriptions.getEntitlements().firstOrNull().orEmpty()
```
malmstein, line 222: In resolveUserTier(), getEntitlements() is called via .firstOrNull() to snapshot the current value:

`val entitlements = subscriptions.getEntitlements().firstOrNull().orEmpty()`

But getEntitlements() is also collected in init for the reactive trigger. That's two independent subscriptions to the same flow. If getEntitlements() is backed by a hot StateFlow from the previous PR ([#8566](https://github.com/duckduckgo/Android/issues/8566)), this is fine — firstOrNull() just takes the current value. But if it's a cold flow that hits the network, every fetchModels() call creates a redundant upstream fetch. Worth confirming the flow is hot/cached before landing.

malmstein (approved):
Works as expected, just the one question about the entitlements flow

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/DuckAiModelManager.kt
YoussefKeyrouz: Confirmed hot. getEntitlements() is backed by MutableSharedFlow(replay = 1) in SubscriptionsManager (from PR #8566), and the onSubscription { emitEntitlementsValues() } block reads from the local store not the network. So each firstOrNull() returns the cached replay value immediately and triggers a couple of redundant disk reads on the IO dispatcher in the background, but never an upstream fetch. I considered caching the entitlement set locally in RealDuckAiModelManager to skip the disk-read churn, but it introduces a cold-start race (cache empty until the init-block collector processes its first emission, which would briefly mis-resolve a Pro user as Free if the picker is opened very early) and adds a second source of truth that has to stay in sync with the flow. 

Given the work being saved is a second IO-dispatcher disk reads per tap, I think it’s not worth over engineering this. Happy to do it if you disagree.

PR #8592, opened 2026-05-18
Notify Mattermost when Play Store build starts; fix existing messages

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211760946270935/task/1214893969211117?focus=true

### Description
Add an additional message posting during Play Store release process when app building is starting. Also fixes
- missing app version in the other messages due to incorrect reference
- incorrect title (referencing Internal track)

### Steps to test this PR
- qa optional

---

> [!NOTE]
> **Low Risk**
> Low risk workflow-only change that adds an extra notification step and corrects variables used in Mattermost messages; main risk is incorrect messaging or minor workflow YAML issues, not app behavior.
> 
> **Overview**
> Adds a new Mattermost notification at the start of the Play Store release workflow to announce the build has begun and will be uploaded when complete.
> 
> Updates existing Mattermost messages in `release_upload_play_store.yml` to consistently use `inputs.ref` (instead of `github.event.inputs.ref`) and adjusts the job name wording to reflect the “for review” Play Store upload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1cc153fc76c72cc09762e8d62ea242aff9ae0ded. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release_upload_play_store.yml
+++ b/.github/workflows/release_upload_play_store.yml
@@ -24,7 +24,7 @@ env:
 jobs:
   release-production:
     runs-on: ubuntu-latest
-    name: Publish Bundle to Play Store Internal track
+    name: Publish Bundle to Play Store for review
     env:
       DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
 
@@ -80,6 +80,15 @@ jobs:
           fileName: api.json
           destination-path: $HOME/jenkins_static/com.duckduckgo.mobile.android/
 
+      - name: Notify Mattermost that Play Store build is starting
+        uses: duckduckgo/native-github-asana-sync@v2.0
+        with:
+          mattermost-token: ${{ secrets.MM_AUTH_TOKEN }}
+          mattermost-team-id: ${{ secrets.MM_TEAM_ID }}
+          mattermost-channel-name: ${{ vars.MM_RELEASE_NOTIFY_CHANNEL }}
+          mattermost-message: ${{ env.emoji_info }} Building release ${{ inputs.ref }}. This will be uploaded to Play Store when complete.
+          action: 'send-mattermost-message'
+
       - name: Assemble the bundle
         env:
           MALICIOUS_SITE_PROTECTION_AUTH_TOKEN: ${{ secrets.MALICIOUS_SITE_PROTECTION_AUTH_TOKEN }}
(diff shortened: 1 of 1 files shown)


CDRussell:
* **#8592** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8592?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8592?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

.github/workflows/release_upload_play_store.yml
```diff
@@ -80,6 +80,15 @@ jobs:
           fileName: api.json
           destination-path: $HOME/jenkins_static/com.duckduckgo.mobile.android/
 
+      - name: Notify Mattermost that Play Store build is starting
```
CrisBarreiro, line 83: Do we need this message? We already notify when we cut the release, so I don't think this message brings much value, and creates some additional noise instead.

.github/workflows/release_upload_play_store.yml
```diff
@@ -137,7 +146,7 @@ jobs:
           mattermost-token: ${{ secrets.MM_AUTH_TOKEN }}
           mattermost-team-id: ${{ secrets.MM_TEAM_ID }}
           mattermost-channel-name: ${{ vars.MM_RELEASE_NOTIFY_CHANNEL }}
-          mattermost-message: ${{env.emoji_end}} Release ${{ github.event.inputs.ref }} completed successfully and is now in review.
+          mattermost-message: ${{env.emoji_end}} Release ${{ inputs.ref }} completed successfully and is now in review.
```
CrisBarreiro, line 149: Similar to the above, can we just colapse all these messages into a single one with the summary of the actions performed?

CrisBarreiro (approved):

.github/workflows/release_upload_play_store.yml
CDRussell: I think it's good confirmation that the next steps is automatic and already underway; removes any uncertainty around whether that part still requires manual intervention. Even cutting the release today I did a double-take whether things were progressing past the point of tag creation and is what made me think of this.

After the above, there is a post on MM that says the tag has been created and an Asana checklist which has nothing in it about building the app. it's just transparently happening somewhere in another GHA.

i.e., it's a gap where it created doubt for me so want to remove that for future-me and others.

.github/workflows/release_upload_play_store.yml
CDRussell: Probably, but going to treat as out of scope; just fixing the obvious bug here where the version name isn't included but was supposed to be.

PR #8607, opened 2026-05-18
Upsell ai models subscriptions

YoussefKeyrouz:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212810093780571/task/1214681379562041?focus=true 

### Description

Adds subscription upsell flows to the Duck.ai native model picker. Tapping a gated model now routes the user into the native subscription Purchase flow (Free → Plus/Pro) or Upgrade flow (Plus → Pro), instead of being disabled/silent. Models are grouped into Free / Plus / Pro sections regardless of the viewer's tier.

### High-level changes

**Model + tier knowledge:**
- New `AIChatModel.requiredTier: UserTier?` derived from `accessTier`. Returns the lowest public tier that grants access, or `null` if the model has only non-public access tier strings (e.g. `"internal"`) that the app can't route an upsell to.
- New `PickerSurface` enum (top-level in `ModelPickerViewModel.kt`) carrying the per-surface analytics funnel origin (`funnel_nativeinput_androidapp__modelpicker` / `funnel_duckai_androidapp__modelpicker`).
- `RealDuckAiModelManager.fetchModels()` filters out "ghost" models (empty `accessTier` with `entityHasAccess = false`)

**Picker UI (`ModelPickerView` / `ModelPickerViewModel`):**
- `buildSections()` rewritten to always produce Free / Plus / Pro sections for every viewer tier. Models with `requiredTier == null` are intentionally hidden.
- Section headers visually updated (See screenshot below)
- The `setDisabled()` styling for inaccessible items is gone; every item is fully tappable.
- `ModelPickerViewModel` runs the decision: accessible → `selectModel`; gated → emits `Command.LaunchPurchase` / `Command.LaunchUpgrade`
- `NativeInputHost` plumbed into the view so `currentSurface()` can read `inputContext` at tap time.
- Existing `NativeInputManager.launchUpgrade()` renamed to `launchPurchase()` to properly reflect the behavior it launches

### Steps to test this PR

_Free user_
- [ ] Sign out / confirm Free tier.
- [ ] Open the Duck.ai native input model picker.
- [ ] Verify three sections render: free models (no header), then **plus** header + Plus models, then **pro** header + Pro models.
- [ ] Tap a Plus model → native Purchase flow opens.
- [ ] Tap a Pro model → native Purchase flow opens.
- [ ] Tap a Free model → model selects.

_Plus user_ (I tested this by temporary hardcoding a pro user to be PLUS in `RealDuckAiModelManager.resolveUserTier`)
- [ ] Sign in with a Plus account.
- [ ] Open the model picker.
- [ ] Tap a Plus model → model selects normally.
- [ ] Tap a Pro model → native **Upgrade** flow opens (not Purchase).

_Pro user_
- [ ] Sign in with a Pro account
- [ ] Open the picker.
- [ ] Tap any tier of model → model selects normally. No subscription flow opens.

_Sign out_
- [ ] Sign out while picker is open → picker refreshes; Free model selected. Pro/Plus models trigger purchase flow

### UI changes

**Purchase Flow**

https://github.com/user-attachments/assets/a0d70af7-4fe8-4ed6-9027-f032e68396a8 

**Upgrade Flow**

https://github.com/user-attachments/assets/21ada09c-00fe-4168-a0d7-73da378699c4

**Screenshot**
Screenshot 2026-05-18 at 7 16 55 PM

---

> [!NOTE]
> **Medium Risk**
> Adds new subscription purchase/upgrade routing from the model picker and changes how models are filtered/sectioned by tier, which could affect who can see/select models and trigger billing UI flows.
> 
> **Overview**
> Tapping an inaccessible Duck.ai model in the native model picker now triggers a **native subscription upsell** (Free → `SubscriptionPurchase`, Plus → `SubscriptionUpgrade`) instead of showing a disabled item; the picker emits commands from `ModelPickerViewModel` and `ModelPickerView` executes the navigation with per-surface `origin` values.
> 
> Model tier handling is updated by deriving `AIChatModel.requiredTier` from `accessTier`, grouping the picker into **Free/Plus/Pro** sections (hiding models with only non-public tiers), and filtering out “ghost” models in `RealDuckAiModelManager.fetchModels()` where `accessTier` is empty and the user has no access. Minor tweaks include expanding Mistral provider detection, adding `DUCK_AI_FEATURE_PAGE` constant, and updating section header styling/strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28d09668237541106392123e8d6cd09c19696deb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
@@ -110,6 +110,16 @@ data class AIChatModel(
 
     fun supportsTool(tool: Tool): Boolean = supportedTools.contains(tool)
 
+    /** The lowest [UserTier] required to access this model, or `null` if no public tier applies. */
+    val requiredTier: UserTier?
+        get() = when {
+            accessTier.isEmpty() -> UserTier.FREE
+            accessTier.contains(UserTier.FREE.rawValue) -> UserTier.FREE
+            accessTier.contains(UserTier.PLUS.rawValue) -> UserTier.PLUS
+            accessTier.contains(UserTier.PRO.rawValue) -> UserTier.PRO
+            else -> null
+        }
+
     companion object {
         val NATIVE_SUPPORTED_IMAGE_FORMATS = listOf("png", "jpeg", "webp")
     }
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
@@ -149,7 +149,7 @@ enum class ModelProvider {
         fun from(id: String, providerString: String?): ModelProvider {
             return when {
                 id.startsWith("meta-llama/") || id.startsWith("meta-llama_") || providerString == "azure" -> META
-                id.startsWith("mistralai/") || id.startsWith("mistralai_") -> MISTRAL
+                id.startsWith("mistralai/") || id.startsWith("mistralai_") || providerString == "mistral" -> MISTRAL
                 id.contains("gpt-oss") -> OSS
                 providerString == "anthropic" -> ANTHROPIC
                 providerString == "openai" -> OPENAI
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
@@ -54,7 +54,6 @@ import dagger.android.support.AndroidSupportInjection
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
-import logcat.logcat
 import javax.inject.Inject
 
 interface ModelPicker {
@@ -87,7 +86,7 @@ class ModelPickerView @JvmOverloads constructor(
     private var commandJob: Job? = null
     private var popupWindow: PopupWindow? = null
     private var lastObservedModelId: String? = null
-    private var host: NativeInputHost? = null
+    private lateinit var host: NativeInputHost
     override var onMenuShown: (() -> Unit)? = null
     override var onMenuDismissed: (() -> Unit)? = null
     override var onModelSelected: (() -> Unit)? = null
@@ -165,16 +164,11 @@ class ModelPickerView @JvmOverloads constructor(
         }
     }
 
-    private fun currentSurface(): PickerSurface {
-        val inputContext = host?.getInputState()?.inputContext
-        if (inputContext == null) {
-            logcat { "Duck.ai picker: host not set, defaulting upsell surface to ADDRESS_BAR. Funnel attribution may be wrong." }
-        }
-        return when (inputContext) {
+    private fun currentSurface(): PickerSurface =
+        when (host.getInputState().inputContext) {
             InputContext.DUCK_AI, InputContext.DUCK_AI_CONTEXTUAL -> PickerSurface.DUCK_AI_TAB
-            InputContext.BROWSER, null -> PickerSurface.ADDRESS_BAR
+            InputContext.BROWSER -> PickerSurface.ADDRESS_BAR
         }
-    }
 
     private fun showMenu() {
         val state = viewModel.state.value
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
@@ -40,14 +40,21 @@ import com.duckduckgo.common.ui.view.divider.HorizontalDivider
 import com.duckduckgo.common.ui.view.text.DaxTextView
 import com.duckduckgo.common.utils.ViewViewModelFactory
 import com.duckduckgo.di.scopes.ViewScope
+import com.duckduckgo.duckchat.api.nativeinput.NativeInputState.InputContext
+import com.duckduckgo.duckchat.impl.DuckChatConstants.DUCK_AI_FEATURE_PAGE
 import com.duckduckgo.duckchat.impl.R
 import com.duckduckgo.duckchat.impl.models.AIChatModel
 import com.duckduckgo.duckchat.impl.models.ModelState
+import com.duckduckgo.duckchat.impl.nativeinput.NativeInputHost
+import com.duckduckgo.navigation.api.GlobalActivityStarter
+import com.duckduckgo.subscriptions.api.SubscriptionScreens.SubscriptionPurchase
+import com.duckduckgo.subscriptions.api.SubscriptionScreens.SubscriptionUpgrade
 import com.google.android.material.chip.Chip
 import dagger.android.support.AndroidSupportInjection
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
+import logcat.logcat
 import javax.inject.Inject
 
 interface ModelPicker {
@@ -58,6 +65,7 @@ interface ModelPicker {
     fun isImageGenerationSupported(): Boolean
     fun isWebSearchSupported(): Boolean
     fun setPickerEnabled(enabled: Boolean)
+    fun setHost(host: NativeInputHost)
 }
 
 @InjectWith(ViewScope::class)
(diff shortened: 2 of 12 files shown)


YoussefKeyrouz:
* **#8607** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

malmstein (commented):
a few suggestions to discuss — happy to land any subset.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
```diff
@@ -223,13 +266,9 @@ class ModelPickerView @JvmOverloads constructor(
             setLeadingIcon(viewModel.getIconResForModel(model))
```
malmstein, line 260: nit (UI): Mistral Small 4 renders without an icon. its model id probably doesn't match the `mistralai/` or `mistralai_` prefix in `ModelProvider.from`, so it falls through to UNKNOWN. unrelated to this PR but more visible now that paid models are interleaved.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerViewModel.kt
```diff
@@ -58,36 +63,48 @@ class ModelPickerViewModel @Inject constructor(
...
-    }
+    fun buildSections(state: ModelState): List<ModelSection> {
+        // Models with a null requiredTier (non-public access tiers only, e.g. "internal") have no
+        // section to land in and are intentionally hidden from the picker.
+        state.models.filter { it.requiredTier == null }.forEach {
+            logcat { "Duck.ai picker: hiding model with no public required tier (id=${it.id}, accessTier=${it.accessTier})." }
+        }
+        val freeModels = state.models.filter { it.requiredTier == UserTier.FREE }
+        val plusModels = state.models.filter { it.requiredTier == UserTier.PLUS }
+        val proModels = state.models.filter { it.requiredTier == UserTier.PRO }
```
malmstein: we're filtering the same list four times (once for the warning, three for the sections). could we `groupBy { it.requiredTier }` once and pull each bucket out? reads as one partition rather than independent filters.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
```diff
@@ -133,6 +149,31 @@ class ModelPickerView @JvmOverloads constructor(
...
+            is ModelPickerViewModel.Command.LaunchPurchase ->
+                globalActivityStarter.start(context, SubscriptionPurchase(origin = command.origin, featurePage = DUCK_AI_FEATURE_PAGE))
+            is ModelPickerViewModel.Command.LaunchUpgrade ->
+                globalActivityStarter.start(context, SubscriptionUpgrade(origin = command.origin))
+        }
+    }
+
+    private fun currentSurface(): PickerSurface {
+        val inputContext = host?.getInputState()?.inputContext
+        if (inputContext == null) {
```
malmstein: the plugin sets host before `setPickerEnabled(true)`, so this null branch should be unreachable in practice. since wrong attribution silently degrades the funnel, would we rather make host non-null (set it in `createView` before the picker is returned, or pass it through the constructor) and let this throw if the contract ever breaks?

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerViewModel.kt
```diff
@@ -98,7 +115,16 @@ class ModelPickerViewModel @Inject constructor(
...
+    sealed class Command {
+        data class LaunchPurchase(val origin: String) : Command()
+        data class LaunchUpgrade(val origin: String) : Command()
     }
+
+    private fun List<AIChatModel>.toSectionOrNull(@StringRes headerRes: Int?): ModelSection? =
+        takeIf { it.isNotEmpty() }?.let { ModelSection(headerRes, it) }
+}
+
+enum class PickerSurface(val origin: String) {
```
malmstein: nit: `PickerSurface` is only used by this ViewModel and the view that drives it — would it fit better nested inside `ModelPickerViewModel` alongside `Command`?

duckchat/duckchat-impl/src/test/kotlin/com/duckduckgo/duckchat/impl/ui/ModelPickerViewModelTest.kt
```diff
@@ -153,65 +201,162 @@ class ModelPickerViewModelTest {
...
-            userTier = UserTier.PLUS,
+    fun whenFreeUserTapsPlusModelFromAddressBarThenLaunchPurchaseCommandEmittedWithAddressBarOrigin() = runTest {
+        stateFlow.value = ModelState(userTier = UserTier.FREE)
+        val model = plusModel("p")
+
+        testee.commands.test {
+            testee.onModelTapped(model, PickerSurface.ADDRESS_BAR)
+
+            assertEquals(
+                ModelPickerViewModel.Command.LaunchPurchase("funnel_nativeinput_androidapp__modelpicker"),
```
malmstein: nit: asserting the literal couples the test to the exact funnel name. `PickerSurface.ADDRESS_BAR.origin` (and `DUCK_AI_TAB.origin` below) would still validate the wiring and survive any analytics rename. applies to the other three origin asserts in this file as well.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
```diff
@@ -110,6 +110,16 @@ data class AIChatModel(
 
     fun supportsTool(tool: Tool): Boolean = supportedTools.contains(tool)
 
+    /** The lowest [UserTier] required to access this model, or `null` if no public tier applies. */
+    val requiredTier: UserTier?
```
malmstein, line 114: nit: worth a direct unit test on `requiredTier` — the priority order (FREE > PLUS > PRO) and the empty / non-public branches are load-bearing for both the picker and the upsell logic, and they're currently only exercised transitively via `buildSections` / `onModelTapped`.

malmstein (approved):
Works as expected, left a few nits but nothing blocking. Check the Mistral icon in case we can fix it

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/AIChatModel.kt
YoussefKeyrouz: Added AICHatModelTest file.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
YoussefKeyrouz: I logged the model payload and found that the check was missing a `providerString` check. Icon showing now.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerViewModel.kt
YoussefKeyrouz: Good catch, updated to be a single group by. Also removed the log warning, I don’t think it’s needed. It was mostly for my testing.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerView.kt
YoussefKeyrouz: Good callout. I rather the view crashes over being miss attributed. So moved the host to a lateinit var, it is initialized before it returns so we’re fine. There is a potential human error if we inflate the view in the future through xml and don’t set the host it will crash if you select a model. So theoretically a developer can use this view and not notice the crash if they never tapped on a model or accessed the currentSurface. 

The alternative is to pass it in the constructor. However constructor injection of host would force us to drop `@JvmOverloads` and the standard (Context, AttributeSet?, Int) constructors, which means the view could no longer be inflated from XML. It’s not used from XML now so I can do this change if you prefer but it changes the signature.

duckchat/duckchat-impl/src/test/kotlin/com/duckduckgo/duckchat/impl/ui/ModelPickerViewModelTest.kt
YoussefKeyrouz: Updated. Thanks

PR #8631, opened 2026-05-20
Add chatIdOrNull to duckchat api module and drop chatID string duplication

GerardPaligot:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1214962354449266 

### Description

Promotes the `"chatID"` query parameter handling out of three impl-private helpers into one function on the public `DuckChat` interface: `chatIdOrNull(uri: Uri): String?`. As a side effect, `DuckChatContextualViewModel.hasChatId` now correctly gates on `isDuckChatUrl` — previously a crafted URL like `https://evil.com?chatID=abc` was treated as a Duck.ai chat.

### Steps to test this PR

_Chat-history clears (DuckChatDataClearingPlugin + DuckAiTabsCleanupPlugin)_

- [ ] Open the Duck.ai chat-history screen → long-press to enter select mode → select 2 chats → tap fire → tap Delete all → confirm the chats clear and any open Duck.ai tabs pointing at those chats close.
- [ ] Open the same chat in two browser tabs → from chat-history, delete that chat via the row overflow → confirm both browser tabs close.
- [ ] In chat-history with at least one pinned chat → tap fire (default mode, no selection) → tap Delete all → confirm every chat, pinned included, clears.

_Chat resume from input suggestions_

- [ ] On the input screen, tap a recent chat suggestion → confirm the chat resumes in Duck.ai with the correct `chatID` in the URL.
- [ ] If your build surfaces the NTP widget input mode, tap a recent chat suggestion there → same result.

### UI changes

n/a

---

> [!NOTE]
> **Medium Risk**
> Changes affect chat deletion/tab matching and contextual fullscreen behavior; the stricter chatID gating fixes mis-detection on non-Duck.ai URLs but alters edge-case URL handling.
> 
> **Overview**
> Introduces **`Uri.toChatIdOrNull(duckChat)`** in `duckchat-api` and replaces several impl-local `chatID` parsers (native input, data clearing, tab cleanup, contextual sheet) with that helper, removing the shared **`CHAT_ID_PARAM`** constant.
> 
> **`DuckChatContextualViewModel`** now treats a URL as having a chat only when it passes **`isDuckChatUrl`**, not merely when a `chatID` query param is present on any host.
> 
> **`buildChatUrl`** is the single place for resume links: it adds **`native-input`** when that setting is on, and the input screen / native input widget call it instead of hand-building URIs. Tests cover **`buildChatUrl`** with native input on/off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 825f6047aa39cf60b00ad5c601e53e1324818110. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/DuckChat.kt
+++ b/duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/DuckChat.kt
@@ -63,6 +63,12 @@ interface DuckChat {
     fun isDuckChatUrl(uri: Uri): Boolean
 
     /**
+     * Returns the chatID query parameter from [uri] when it is a Duck.ai chat URL with a
+     * non-blank chatID; null otherwise (including for non-Duck.ai URLs).
+     */
+    fun chatIdOrNull(uri: Uri): String?
+
+    /**
      * Returns `true` if Duck Chat was ever opened before.
      */
     suspend fun wasOpenedBefore(): Boolean
--- a/duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/DuckChat.kt
+++ b/duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/DuckChat.kt
@@ -63,12 +63,6 @@ interface DuckChat {
     fun isDuckChatUrl(uri: Uri): Boolean
 
     /**
-     * Returns the chatID query parameter from [uri] when it is a Duck.ai chat URL with a
-     * non-blank chatID; null otherwise (including for non-Duck.ai URLs).
-     */
-    fun chatIdOrNull(uri: Uri): String?
-
-    /**
      * Returns `true` if Duck Chat was ever opened before.
      */
     suspend fun wasOpenedBefore(): Boolean
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualViewModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualViewModel.kt
@@ -24,6 +24,7 @@ import com.duckduckgo.anvil.annotations.ContributesViewModel
 import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.FragmentScope
 import com.duckduckgo.duckchat.api.DuckChat
+import com.duckduckgo.duckchat.api.toChatIdOrNull
 import com.duckduckgo.duckchat.impl.DuckChatInternal
 import com.duckduckgo.duckchat.impl.feature.DuckChatFeature
 import com.duckduckgo.duckchat.impl.helper.DuckChatJSHelper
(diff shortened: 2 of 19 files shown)


GerardPaligot:
* **#8631** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8631?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8631?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

duckchat/duckchat-api/src/main/java/com/duckduckgo/duckchat/api/DuckChat.kt
```diff
@@ -62,6 +62,12 @@ interface DuckChat {
      */
     fun isDuckChatUrl(uri: Uri): Boolean
 
+    /**
+     * Returns the chatID query parameter from [uri] when it is a Duck.ai chat URL with a
+     * non-blank chatID; null otherwise (including for non-Duck.ai URLs).
+     */
+    fun chatIdOrNull(uri: Uri): String?
```
aitorvs: As commented on Asana, I'd make this an extension function on the Uri type, father than a public fun in this class.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualViewModel.kt
```diff
@@ -666,8 +666,10 @@ class DuckChatContextualViewModel @Inject constructor(
 
     private fun hasChatId(url: String?): Boolean = !extractChatId(url).isNullOrBlank()
 
-    private fun extractChatId(url: String?): String? =
-        url?.toUri()?.getQueryParameter(CHAT_ID_PARAM)?.takeIf { it.isNotBlank() }
+    private fun extractChatId(url: String?): String? {
+        val uri = url?.toUri() ?: return null
```
malmstein, line 670: nit: this gates on `isDuckChatUrl` via `toChatIdOrNull` where it didn't before — the test fake had to be relaxed to recognise `duck.ai`/`duckduckgo.com` hosts. probably fine since the sheet always loads duck.ai urls by construction, but worth flagging in case there's an edge case (redirects, malformed urls) where we'd want the old leniency.

malmstein (approved):
did not install or smoke-test the APK. nice cleanup — folding the four `chatID` extractors into one `toChatIdOrNull` is the right move, and the learnings log from #8666 explicitly flagged the multi-extractor sprawl as a coordination smell, so it's good to see it resolved. centralising chat-url construction through `buildChatUrl` is also the right direction. just one minor question on the stricter `isDuckChatUrl` gating in `DuckChatContextualViewModel.extractChatId`, nothing blocking.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/contextual/DuckChatContextualViewModel.kt
GerardPaligot: Yes, that's true, but even though there is no check before in this function, the URL is initialized by the DuckChat.getDuckChatUrl function, so we won't break anything. We are just being more restrictive.

PR #8632, opened 2026-05-20
[Android] Crash android.os.BadParcelableException in com.duckduckgo.app.notification.NotificationLaunchModalReporter.onActivityCreated

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1214965912649712?focus=true

### Description

Wrap the getBooleanExtra read in NotificationLaunchModalReporter.onActivityCreated with a try/catch and add an early return when savedInstanceState != null. Also call intent?.sanitize() at the top of IntentDispatcherActivity.onCreate (before super.onCreate) for defense in depth.

The reporter is registered as a process-wide Application.ActivityLifecycleCallbacks, so it fires for every activity — including IntentDispatcherActivity, which receives intents from external apps. Reading any extra triggers full Bundle unparcelling, which throws BadParcelableException when the foreign intent carries a Parcelable whose class is absent from our classpath (e.g. an obfuscated zxh class from the sender). That crashed activity creation and brought the app down on launch.

### Steps to test this PR

- [x] Check out the branch and install
- [x] Run the new unit tests:

./gradlew :app:testPlayDebugUnitTest --tests "com.duckduckgo.app.notification.NotificationLaunchModalReporterTest"

All 6 cases should pass, including the two that simulate BadParcelableException / RuntimeException on extras read.

- [x] (Optional) Reproduce the original crash on develop first to confirm the fix:

adb shell am start -n com.duckduckgo.mobile.android.debug/com.duckduckgo.app.dispatchers.IntentDispatcherActivity -a android.intent.action.VIEW -d "https://duckduckgo.com" --es "junk_key" "junk" --ez  com.duckduckgo.notification.launched_from_notification" false

- [x] Confirm you don't see the crash on the branch using the same command as above

### NO UI changes

---

> [!NOTE]
> **Medium Risk**
> Touches process-wide activity lifecycle callbacks and intent handling; while behavior is guarded and covered by tests, it changes how intents with malformed/foreign parcelables are handled at app launch.
> 
> **Overview**
> Prevents app crashes caused by unparcelling foreign `Intent` extras in `NotificationLaunchModalReporter` by skipping work on recreation (`savedInstanceState != null`), wrapping the `getBooleanExtra` read in `try/catch`, and logging failures instead of crashing.
> 
> Adds defense-in-depth by calling `intent?.sanitize()` at the start of `IntentDispatcherActivity.onCreate` (before `super.onCreate`) to clear problematic extras early. Includes new unit tests covering normal notification launches and exception scenarios (e.g., `BadParcelableException`/`RuntimeException`) to ensure the reporter never crashes or misreports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b5097d15b15a2fb5e1c3625129d9620aedf8330. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/dispatchers/IntentDispatcherActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/dispatchers/IntentDispatcherActivity.kt
@@ -27,6 +27,7 @@ import com.duckduckgo.app.browser.customtabs.CustomTabActivity
 import com.duckduckgo.app.browser.mode.ExternalUrl
 import com.duckduckgo.app.browser.mode.InAppNavigation
 import com.duckduckgo.app.dispatchers.IntentDispatcherViewModel.ViewState
+import com.duckduckgo.app.global.sanitize
 import com.duckduckgo.common.ui.DuckDuckGoActivity
 import com.duckduckgo.di.scopes.ActivityScope
 import com.duckduckgo.navigation.api.GlobalActivityStarter
@@ -44,6 +45,9 @@ class IntentDispatcherActivity : DuckDuckGoActivity() {
     lateinit var globalActivityStarter: GlobalActivityStarter
 
     override fun onCreate(savedInstanceState: Bundle?) {
+        // Sanitize before super.onCreate so lifecycle callbacks dispatched from there don't trip over
+        // Parcelable extras whose classes are absent from our classpath.
+        intent?.sanitize()
         super.onCreate(savedInstanceState)
 
         logcat { "onCreate called with intent $intent" }
--- a/app/src/main/java/com/duckduckgo/app/notification/NotificationLaunchModalReporter.kt
+++ b/app/src/main/java/com/duckduckgo/app/notification/NotificationLaunchModalReporter.kt
@@ -23,6 +23,9 @@ import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.modalcoordinator.api.ModalShownReporter
 import com.squareup.anvil.annotations.ContributesMultibinding
 import dagger.SingleInstanceIn
+import logcat.LogPriority.ERROR
+import logcat.asLog
+import logcat.logcat
 import javax.inject.Inject
 
 @ContributesMultibinding(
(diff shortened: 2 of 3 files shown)


anikiki:
* **#8632** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8632?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

catalinradoiu (approved):
Tested and works as expected 🎉 

The bug cannot be reproduced via adb shell am start alone. The am command's extras flags (--es, --ez, --ei, …) only inject primitives and a couple of framework types (Uri, ComponentName) — there is no flag to inject an arbitrary Parcelable. The crash needs a bundle entry whose class is not on DDG's classpath, which means the sender has to be a different APK defining its own Parcelable class. Also the crash happens only on lower API versions, below API 33. 

Android 13 reworked BaseBundle to use lazy parcel reading. On API ≥ 33, unparcel() walks only the bundle's keys; each value's bytes stay parceled until the corresponding get(key) call,  which then resolves only that one value. Since NotificationLaunchModalReporter reads EXTRA_LAUNCHED_FROM_NOTIFICATION — a different key from the bad Parcelable's — only that boolean is decoded, the unknown class is never resolved, and no exception fires. On API ≤ 32 the older eager unparcel() walks the whole array map in one shot, calling readValue → readParcelable → Class.forName for every entry, so a single getBooleanExtra of any key trips on the unknown-class entry and crashes. That's why the crashes in Play Console concentrate on older Android versions, and why the same evil intent against fix-branch DDG on an API 36 emulator wouldn't have crashed even without sanitize() — the lazy bundle hid the bug. The fix matters for the population that still runs API ≤ 32.

PR #8557, opened 2026-05-14
Implement scriptlet download and storage

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1214486375443736?focus=true

### Steps to test this PR

> [!TIP]
> Set logcat filter `package:mine tag~:"RealScriptletUpdater|ScriptletDownloadWorker|AdBlockingExtensionProvider|RealPrivacyConfigDownloader|ScriptletSignatureValidator|RealAdBlockingExtensionConfigProvider" `

_Feature 1_ - RC flag disabled
- [x] Install the app
- [x] Check logs for `onPrivacyConfigDownloaded`
- [x] Check no logs for files actually downloaded
- [x] Open database inspector and check `ad_blocking_extension.db` doesn't exist

_Feature 2_ - RC flag disabled with settings
- [x] Apply Patch 1 - update privacy-config
- [x] Fresh install the app
- [x] Check logs for `onPrivacyConfigDownloaded`
- [x] Check no logs for files actually downloaded
- [x] Open database inspector and check `ad_blocking_extension.db` doesn't exist

Feature 3 - Kill-switch on, contingency mode on (feature is discoverable but dormant)
- [x] From the previous scenario, enable contingecyModeEnabled from the feature flags inventory. Also enable adBlockingExtension (order is important)
- [x] Kill and reopen the app
- [x] Check logs for onPrivacyConfigDownloaded
- [x] Check no logs for files actually downloaded
- [x] Open database inspector and check ad_blocking_extension.db doesn't exist

Feature 4 - Kill-switch on, contingency mode off (full operational path)
- [x] From the previous scenario, enable adBlockingExtension from the feature flags inventory and check contingecyModeEnabled is disabled
- [x] Check logs for onPrivacyConfigDownloaded, Validating scriptlet succeeded (x4), and Storing scriptlets
- [x] Open database inspector and check ad_blocking_extension.db exists, and has the following contents:
    - [x] name: scriptlets/isolated/ublock-filters.js, version: 2026.5.12, content: {some bytes}
    - [x] name: scriptlets/main/ublock-filters.js, version: 2026.5.12, content: {some bytes}
- [x] Kill and reopen the app
- [x] Check logs for Version matches stored. Skipping

_Feature 5_ - RC flag disabled with settings
- [x] Apply Patch 1 - update privacy-config
- [x] Fresh install the app
- [x] Enable adBlockingExtension from the feature flags inventory
- [x] Kill and restart the app
[...]

PR #8660, opened 2026-05-21
Add settings infra for ad blocking

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1214486375443738?focus=true

### Description
Add settings infra for ad blocking:
* Display either "Duck Player" or "YouTube Ad Blocking" settings in "Other Settings" section
* When using YouTube Ad Blocking" settings, add "Duck Player" as s sub-setting
* Take user settings into account for scriptlet injection
* Allow default user settings to be configured remotely 

### Steps to test this PR

_Feature 1_ YouTube ad blocking disabled
- [x] Apply _Patch 1: Update RC URL_
- [x] Fresh install
- [x] Open settings
- [x] Check there's an entry for "Duck Player" and none for "YouTube ad blocking"

_Feature 2_ YouTube ad blocking enabled
- [x] Apply _Patch 2: Feature enabled with settings_
- [x] Fresh install
- [x] Open settings
- [x] Check there's an entry for  "YouTube ad blocking" and none for "Duck Player"
- [x] Open YouTube ad blocking
- [x] Tap on Duck Player description
- [x] Check Duck Player settings are open

_Feature 3_ User setting disabled doesn't inject scripts
- [x] From the previous scenario, open "YouTube ad blocking" settings
- [x] Check feature is off
- [x] Open youtube.com
- [x] Check logs for `User disabled ad blocking` and `Status checker rejected injection, skipping`

_Feature 4_ User setting enabled injects scripts
- [x] From the previous scenario, open "YouTube ad blocking" settings
- [x] Enable the feature
- [x] Reload youtube.com.  Note: Automatic page reload will be addressed as a separate task
- [x] Check no logs for `User disabled ad blocking` and `Status checker rejected injection, skipping`
- [x] Check logs for `Injecting script`

_Feature 5_ Enabled by default on
- [x] Apply _Patch 3: Feature enabled with settings and enabledByDefault_
- [x] Fresh install the app
- [x] Open "YouTube ad blocking" settings
- [x] Check feature is on
- [x] Open youtube.com
- [x] Play a few videos and check ads are not shown
- [x] Check no logs for `User disabled ad blocking` and `Status checker rejected injection, skipping`
- [x] Go back to "YouTube ad blocking" settings
- [x] Disable the feature
- [x] Reload youtube.com. Note: Automatic page reload will be addressed as a separate task
- [x] Check logs for `User disabled ad blocking` and `Status checker rejected injection, skipping`
- [x] Play a few videos and check ads are shown

### Patches

_Patch 1: Update RC URL_
```
Subject: [PATCH] Update privacy-config URL
---
Index: privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
IDEA additional info:
<+>UTF-8
diff --git a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
--- a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt	(revision ca28c5a412606fefb8bcee8532f901d9fe30b55c)
+++ b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt	(date 1779203719053)
@@ -29,4 +29,4 @@
     TrackingParametersFeatureName("trackingParameters"),
 }
 
-const val PRIVACY_REMOTE_CONFIG_URL = "https://staticcdn.duckduckgo.com/trackerblocking/config/v5/android-config.json"
+const val PRIVACY_REMOTE_CONFIG_URL = "https://duckduckgo.github.io/privacy-configuration/pr-5175/v4/android-config.json"
```

_Patch 2: Feature enabled with settings_
```
Subject: [PATCH] Update privacy-config URL - Feature enabled with settings
---
Index: privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
IDEA additional info:
<+>UTF-8
diff --git a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
--- a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt	(revision 61e308ef7401f17000e20c67f6a150f7d424badc)
+++ b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt	(date 1779712348322)
@@ -29,4 +29,4 @@
     TrackingParametersFeatureName("trackingParameters"),
 }
 
-const val PRIVACY_REMOTE_CONFIG_URL = "https://staticcdn.duckduckgo.com/trackerblocking/config/v5/android-config.json"
+const val PRIVACY_REMOTE_CONFIG_URL = "https://gist.githubusercontent.com/CrisBarreiro/bf9e0dd845db24899fbac8d4450f1d93/raw/3f7244595442e3d6fd3d27631ebe826458318985/android-config.json"
```

_Patch 3: Feature enabled with settings and enabledByDefault_
```
Subject: [PATCH] Update privacy-config URL - Feature enabled with settings and enabledByDefault
---
Index: privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
IDEA additional info:
<+>UTF-8
diff --git a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt
--- a/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt	(revision d931e4a72dc1165517cb7c04a92c4cae354281d2)
+++ b/privacy-config/privacy-config-api/src/main/java/com/duckduckgo/privacy/config/api/PrivacyFeatureName.kt	(date 1779722673717)
@@ -29,4 +29,4 @@
     TrackingParametersFeatureName("trackingParameters"),
 }
 
-const val PRIVACY_REMOTE_CONFIG_URL = "https://staticcdn.duckduckgo.com/trackerblocking/config/v5/android-config.json"
+const val PRIVACY_REMOTE_CONFIG_URL = "https://gist.githubusercontent.com/CrisBarreiro/7f27eaa4178f0d232e0282ce7f765213/raw/6642049d9080d3657c641dbe44ff0d9d69c201ce/android-config%2520for%2520testing%2520ad%2520blocking%2520feature%2520(enabled)"
```

### UI changes
| Before  | After |
| ------ | ----- |
|settings_before|settings_after|
|n/a|ad_blocking_settings_after|

---

> [!NOTE]
> **Medium Risk**
> Changes browser behavior on YouTube via injection gating and settings/plugin visibility; scope is feature-flagged but touches core browsing and settings UX.
> 
> **Overview**
> Adds **YouTube Ad Blocking** as a first-class settings experience in **Other Settings**, wired through a new `AdBlockingSettingsPlugin` and dedicated screen (toggle, learn more, link into Duck Player).
> 
> **Settings visibility is mutually exclusive with the old Duck Player row:** when the ad-blocking feature is on in remote config, users see **YouTube Ad Blocking** only; the standalone **Duck Player** list item is hidden and reachable from inside the ad-blocking screen instead. `SettingsViewModel` now derives `isDuckPlayerEnabled` / `isAdBlockingEnabled` from each plugin’s `isShownInSettings()` rather than Duck Player state alone.
> 
> **User preference gates script injection:** a DataStore-backed on/off is persisted, combined with remote **`enabledByDefault`** when the user has never chosen, and `canInject()` returns false when the user has turned blocking off.
> 
> E2E builds apply a config patch to enable the feature, and Maestro navigates via **YouTube Ad Blocking → Duck Player**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e3318c1ef4cf1e06093b90c8b1e5ff18069bdc3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CrisBarreiro:
* **#8660** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8660?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8660?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8622** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8622?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8557** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

settings/settings-api/src/main/java/com/duckduckgo/settings/api/ProSettingsPlugin.kt
```diff
@@ -38,6 +38,21 @@ interface ProSettingsPlugin : SettingsPlugin
 /**
  * This is the plugin for Duck Player settings
  */
-interface DuckPlayerSettingsPlugin : SettingsPlugin
+interface DuckPlayerSettingsPlugin : SettingsPlugin {
+    /**
+     * Whether the settings section should be shown
+     */
+    fun isShownInSettings(): Boolean
```
CrisBarreiro, line 45: Some other settings expose this through the feature API module instead, but since we only need it for settings (and we have no public API for ad blocking), I think it makes more sense here. Another alternative would be to add a custom view with a ViewModel for the settings, and then only inject that. I felt that was a bit overkill for phase 1, but happy to discuss more. Depending on what we land on, I'll open an API proposal if needed

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/ui/AdBlockingSettingsViewModel.kt
```diff
@@ -0,0 +1,83 @@
...
+        data object OpenDuckPlayerSettings : Command()
+    }
+
+    val viewState: StateFlow<ViewState> = statusChecker.isUserEnabledFlow()
+        .map { ViewState(isEnabled = it) }
+        .stateIn(
+            viewModelScope,
+            started = SharingStarted.Companion.WhileSubscribed(),
+            initialValue = ViewState(),
+        )
```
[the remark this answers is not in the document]
CrisBarreiro: Haven't been able to reproduce this, but this is the usual pattern we follow for settings and StateFlow

settings/settings-api/src/main/java/com/duckduckgo/settings/api/ProSettingsPlugin.kt
cmonfortep: Something I don't like is we moved DuckPlayerSettings.kt into adBlocking. That feels counter intuitive if we have a Duckplayer module
Also, why when youtube is show as option it appears at the end of the list? In figma appears in the same position where DuckPlayer currently is.
I think for milestone 1, where we show one of the settings in the same position, mutually exclusive, this should be easy to solve using priority annotation and only showing first item:

keep one interface here
implement from each module, and give more priority to adblocking setting
inside SettingsActivity, in the section for duckplayer, pick the first plugin

No changes in this api.
If that doesn't work, I would explore exposing the value in adblocking api, and explaining this is part of feature rollout and temporary. But we should discuss in api proposal.
Note: we should know in milestone 2 or during ship review feedback, we will show the setting in different section. So if you  think now is the moment to simplify changes, then we can go for it.

Or (unsure of this one), if `ContributesActivePlugin` is something to consider

app/src/main/java/com/duckduckgo/app/settings/SettingsViewModel.kt
```diff
@@ -238,7 +240,8 @@ class SettingsViewModel @Inject constructor(
                     showSyncSetting = deviceSyncState.isFeatureEnabled(),
                     isAutoconsentEnabled = autoconsent.isSettingEnabled(),
                     isSubscriptionEnabled = subscriptions.isEligible(),
-                    isDuckPlayerEnabled = duckPlayer.getDuckPlayerState().let { it == ENABLED || it == DISABLED_WIH_HELP_LINK },
+                    isDuckPlayerEnabled = duckPlayerSettingsPlugins.getPlugins().any { it.isShownInSettings() },
```
cmonfortep, line 243: I leave this pending discussion the approach to how we display settings.

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/ui/DuckPlayerSettings.kt
```diff
@@ -14,19 +14,23 @@
  * limitations under the License.
  */
 
-package com.duckduckgo.duckplayer.impl
+package com.duckduckgo.adblocking.impl.ui
```
cmonfortep, line 17: I don't think this class belongs to this module. If we have duckplayer module, imho they should own this view.

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/store/AdBlockingUserPreferences.kt
```diff
@@ -0,0 +1,56 @@
...
+    suspend fun setEnabled(enabled: Boolean)
+}
+
+@SingleInstanceIn(AppScope::class)
+@ContributesBinding(AppScope::class)
+class RealAdBlockingUserPreferences @Inject constructor(
+    @AdBlockingPreferences private val dataStore: DataStore<Preferences>,
+) : AdBlockingUserPreferences {
+
+    override fun isEnabledFlow(): Flow<Boolean?> = dataStore.data.map { prefs -> prefs[KEY_ENABLED] }
```
cmonfortep, line 45: Got this review comment from claude, thoughts?

> No .catch on dataStore.data. DataStore's flow throws IOException on disk errors and the convention in this repo (and the DataStore docs) is
  .catch { if (it is IOException) emit(emptyPreferences()) else throw it }. Without it, a corrupted file kills the StateFlow in
  AdBlockingStatusChecker (collection terminates, value stays at false forever, canInject() permanently returns false). Worth adding

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/domain/AdBlockingStatusChecker.kt
```diff
@@ -16,23 +16,40 @@
...
 @SingleInstanceIn(AppScope::class)
 @ContributesBinding(AppScope::class)
 class RealAdBlockingStatusChecker @Inject constructor(
     private val feature: AdBlockingExtensionFeature,
+    settingsRepository: AdBlockingSettingsRepository,
+    @AppCoroutineScope appScope: CoroutineScope,
 ) : AdBlockingStatusChecker {
 
+    private val userEnabled: StateFlow<Boolean> = settingsRepository.isEnabledFlow()
+        .map { it ?: feature.enabledByDefault().isEnabled() }
```
cmonfortep, line 50: This is more to confirm expectations here around FF:
a) (this implementation) while the FF is enabled, users that didn't interact with the feature, youtube blocking enabled. But if we ever turn off the flag, youtube adblocking is also false again.
b) while FF is enabled, we opt-in users, setting the choice to true. If FF disabled, opted-in users keep the feature enabled, new users stay as false.

Just to check what's the behavior with other platforms.

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/AdBlockingPixelNames.kt
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.adblocking.impl
+
+object AdBlockingPixelNames {
+    const val DUCK_PLAYER_SETTINGS_PRESSED = "duckplayer_setting_pressed"
```
cmonfortep, line 20: To revisit after we discuss removing DuckPlayerSettings from this module. Having to duplicate this pixel is probably an smell of something that needs fixing.

cmonfortep (commented):
I'd like to revisit the architecture and api modifications for the settings.

settings/settings-api/src/main/java/com/duckduckgo/settings/api/ProSettingsPlugin.kt
CrisBarreiro: > Something I don't like is we moved DuckPlayerSettings.kt into adBlocking. That feels counter intuitive if we have a Duckplayer module

I could expose the ad blocking RC flag status in its API module so I could keep Duck Player in a separate module. However, since we're moving Duck Player settings into Ad Blocking ones as well, and we might change Duck Player's behavior based on ad blocking, both modules might eventually converge into a single one

> Also, why when youtube is show as option it appears at the end of the list? In figma appears in the same position where DuckPlayer currently is.

I left a comment on Figma about this, but TL;DR: settings in the "Other Settings" section are ordered alphabetically on Android (After General and AI features)

> I think for milestone 1, where we show one of the settings in the same position, mutually exclusive, this should be easy to solve using priority annotation and only showing first item:

`PriorityKey` annotation needs to be assigned in compile-time, so I can't use it to dynamically prioritize one or the other. I could technically add a priority field that is populated in runtime, but that would be a hidden dependency between both that I'd rather avoid 

> Inside SettingsActivity, in the section for duckplayer, pick the first plugin

I think this is also coupling settings to individual features more that we probably should. To be honest, the entire "Other Settings" section is already somewhat coupled, because we inject individual plugin points for each type of setting, when the goal should be to have `OtherSettingsPlugin`, and as I said on a prior comment, inject views that can change dynamically based on a `ViewModel`. Even if we're not moving towards that ideal now, I think adding more logic in `SettingsActivity` puts us even further from it

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/AdBlockingPixelNames.kt
CrisBarreiro: It should be replacing instead of duplicating. I think the fact that there's an implicit dependency between both features is already somewhat of a smell (not a code one at this point). I could expose the feature status in the ad-blocking-module API, but my reasoning around that was:
* If only settings needs to know that Duck Player or Ad Blocking can be hidden, why expose it to everybody else through an ad-blocking-api module?
* Ad blocking is the feature that owns this logic (if ad blocking enabled, don't show Duck Player). If I keep `DuckPlayerSettings` in its own module, then Duck Player owns this logic, not Ad blocking. The counterpart is that of course, now ad blocking owns some logic for Duck Player, but there's a product dependency between both features, and ad blocking is the one that orchestrates it

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/store/AdBlockingUserPreferences.kt
CrisBarreiro: I have doubts on what the desired behavior would be here. 

Currently yes, we would default to false. However, if we default to `emptyPreferences()`, that's implicitly defaulting to `isEnabledByDefault`, which means that if a user had explicitly opted out of the feature after we enabled it, we would still inject the scriptlets for them and, even if they went to settings and tried to disable it, they wouldn't be able to. From an optics perspective, I think the fallback to disabling is safer. Alternatively, in case of corruption, we could use `corruptionHandler`, so that we replace the file (they could at least change settings later), but this only works for `CorruptionException`, not other types of corruption. 

My proposal would be to default to the default value and re-create the file for `CorruptionException`, but default to disabling otherwise

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/domain/AdBlockingStatusChecker.kt
CrisBarreiro: Double checking in https://app.asana.com/1/137249556945/task/1215172256527557/comment/1215172256527559?focus=true

PR #8659, opened 2026-05-21
Add reasoningMode and getChatById to DuckAiChat Store

YoussefKeyrouz:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212810093780571/task/1215002056518125?focus=true 

### Description

**Per-chat reasoning persistence groundwork (1 of 3).**

The FE has been storing a `reasoningMode` field per chat in the JSON blob it pushes to native storage, but Android wasn't reading it. This PR starts extracting that field on the read side so it's available to consumers, and adds a single-chat lookup API to the store. No behavioural change yet - the new field isn't consumed by any feature in this PR.

The follow-up PR will use this data to populate the reasoning picker correctly when the user opens an existing chat (showing the chat's stored mode rather than the global default).

**What changed:**

- **`DuckAiChat`** Added a nullable `reasoningMode: String?` field.
- **New API: `getChatById** on `DuckAiChatStore`. Wraps the existing `chatsDao.getById(chatId)`.

**No migration required.** Room only stores `(chatId, data)` where `data` is the raw FE blob, we just parse a new field at read time.

### Steps to test this PR

- [ ] Submit a duck.ai chat after selecting a model and reasoning effort (e.g. gpt 5 mini - fast)
- [ ] Go to Settings -> Duck.ai in developer settings -> Enable Native Storage Debug
- [ ] Validate that the chat is now stored with the selected “reasoningMode"

### UI changes

_N/A — data-layer only._

---

> [!NOTE]
> **Low Risk**
> Low risk data-layer change that only extends the parsed chat metadata and adds a read-only lookup API; main risk is minor integration/consumer expectations around null/empty `reasoningMode` handling.
> 
> **Overview**
> Adds per-chat `reasoningMode: String?` to `DuckAiChat` and updates JSON parsing to populate it while treating missing, empty, or explicit JSON `null` as `null`.
> 
> Extends `DuckAiChatStore` with `getChatById(chatId)` (implemented in `RealDuckAiChatStore`) to fetch and parse a single chat, and adds unit tests covering `reasoningMode` parsing and the new lookup behavior (not found / malformed JSON).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a380f284e86282097ba36bd7d69dec8864dc8459. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #8664, opened 2026-05-21
Publish active chatId to NativeInputState (per-tab)

YoussefKeyrouz:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212810093780571/task/1215004327619963?focus=true 

### Description

Tech Design: https://app.asana.com/1/137249556945/project/1212810093780571/task/1214935104654912?focus=true (Do not merge until approved)

**Per-chat reasoning persistence groundwork (2 of 3).**

https://github.com/duckduckgo/Android/pull/8659 added the `reasoningMode` field on `DuckAiChat` and a `getChatById` lookup on the store. This PR adds the runtime signal: per-tab `NativeInputState` now carries the chat id of the active Duck.ai chat (if any), produced from the active tab's URL. Nothing consumes the field yet. The data-layer publish is verified in isolation. The next PR (3/3) will wire the consumers (reasoning picker derives modes from the chat's model).

**What changed:**

- **`NativeInputState`** adds a `chatId: String?` (default `null`) in `duckchat-api`.
- **`NativeInputModeWidget`** interface adds `bindChatIdSource`
- **`NativeInputModeWidgetViewModel`** publishes the chat id into per-tab state.
- `extractDuckAiChatId(rawUrl)` in **`NativeInputManager`**

**Out of scope (deferred to PR 3/3):**

- Reasoning picker reading the chat id to derive available modes from the chat's stored model.
- Submission code reading chat id to send the chat's stored model and effort.

### Steps to test this PR

- [ ] Filter logcat by `"Duck.ai native input: active chatId"`:
  - [ ] Open Duck.ai with no existing chat → expect a log line with `chatId=null`.
  - [ ] Tap an existing chat from history → expect a log line with the chat's id.
  - [ ] Switch to another chat in the same tab→ expect `chatId=` changes to the new chat.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes per-tab native input state publication and adds a new `chatId` field that downstream consumers may start relying on; includes a temporary log that could be noisy if left in production.
> 
> **Overview**
> Publishes the active Duck.ai `chatId` into per-tab `NativeInputState` so native-input consumers can resolve per-chat state.
> 
> `RealNativeInputManager` now extracts `chatID` from the current tab URL (via new `extractDuckAiChatId`) and binds it into the widget; the widget forwards this flow to the `NativeInputModeWidgetViewModel`, which buffers and writes `chatId` to the per-tab state store (handling the attach/configure race).
> 
> Adds unit coverage for `extractDuckAiChatId` parsing and for `setActiveChatId` state publishing/clearing behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9437246f9c81ec31da5469a763c6f9cca6443b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/NativeInputModeWidgetViewModel.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/NativeInputModeWidgetViewModel.kt
@@ -75,6 +75,7 @@ import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import logcat.logcat
 import javax.inject.Inject
 
 data class ChatTabSuggestions(
@@ -120,6 +121,8 @@ class NativeInputModeWidgetViewModel @Inject constructor(
 
     private val _modelPickerEnabled = MutableStateFlow(true)
     val modelPickerEnabled: StateFlow<Boolean> = _modelPickerEnabled.asStateFlow()
+
+    private val _chatId = MutableStateFlow<String?>(null)
 
     private val commandChannel = Channel<Command>(capacity = 1, onBufferOverflow = DROP_OLDEST)
     val commands = commandChannel.receiveAsFlow()
@@ -207,6 +210,18 @@ class NativeInputModeWidgetViewModel @Inject constructor(
                     }
                 }
         }
+
+        // Publish chatId to per-tab state. setActiveChatId can fire during widget attach before
+        // configure() sets activeTabId — _chatId holds the value until then.
+        viewModelScope.launch {
+            combine(_chatId, activeTabId.filterNotNull()) { chatId, tabId -> tabId to chatId }
+                .collect { (tabId, chatId) ->
+                    nativeInputStatePublisher.update(tabId) { it.copy(chatId = chatId) }
+                    // TODO: logs the chatId observers will see for this tab. Added for testing.
+                    //  Remove once consumer logic is implemented.
+                    logcat { "Duck.ai native input: active chatId for tab=$tabId is now $chatId" }
+                }
+        }
     }
 
     val chatState: Flow<ChatState> = duckChatInternal.chatState
@@ -245,6 +260,10 @@ class NativeInputModeWidgetViewModel @Inject constructor(
     fun setSelectedTool(tool: String?) {
         val tabId = activeTabId.value ?: return
         nativeInputStatePublisher.update(tabId) { it.copy(selectedTool = tool) }
+    }
+
+    fun setActiveChatId(chatId: String?) {
+        _chatId.value = chatId
     }
 
     fun configure(tabId: String, isDuckAiMode: Boolean, isBottom: Boolean) {
--- a/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
@@ -493,16 +493,18 @@ class RealNativeInputManager @Inject constructor(
             if (!isBottom) {
                 setFloatingSubmitContainer(createFloatingSubmitContainer())
             }
+            // Per-tab chatId (null on new chats) published into NativeInputState for
+            // consumers (reasoning picker, submission) to resolve per-chat state.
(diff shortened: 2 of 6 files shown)


YoussefKeyrouz:
* **#8674** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8666** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8666?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8664** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8664?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/NativeInputModeWidgetViewModel.kt
```diff
@@ -207,6 +210,18 @@ class NativeInputModeWidgetViewModel @Inject constructor(
                     }
                 }
         }
+
+        // Publish chatId to per-tab state. setActiveChatId can fire during widget attach before
+        // configure() sets activeTabId — _chatId holds the value until then.
+        viewModelScope.launch {
+            combine(_chatId, activeTabId.filterNotNull()) { chatId, tabId -> tabId to chatId }
```
malmstein, line 217: the `_chatId` MutableStateFlow + `combine(_chatId, activeTabId.filterNotNull())` pattern has a race that the existing per-tab field `selectedTool` already side-stepped. when the user switches tabs, `activeTabId` becomes the new tab id but `_chatId` still holds the previous tab's value — combine fires with `(oldChatId, newTabId)` and the new tab's state briefly carries the wrong chatId until the new `chatIdFlow` emits and corrects it. the `whenSetActiveChatIdBeforeConfigureThenChatIdIsBufferedAndPublishedOnConfigure` test passes because the artificial ordering puts setActiveChatId first, but in production `observeChatIdSource` only runs in `onAttachedToWindow` so the buffer isn't needed — and meanwhile cross-tab carryover on the long-lived VM bites.

can we mirror `setSelectedTool` at L260 — drop `_chatId` and the second init block, make `setActiveChatId` synchronous:

```
fun setActiveChatId(chatId: String?) {
    val tabId = activeTabId.value ?: return
    nativeInputStatePublisher.update(tabId) { it.copy(chatId = chatId) }
}
```

that removes the race entirely, halves the moving parts, and keeps the per-tab String fields (selectedTool, chatId) on the same pattern.

malmstein (approved):
smoke-tested the APK and the chatId log lines fire as expected for each transition (null → existing chat → switching chats). nice that the test suite includes the buffered-before-configure regression — useful defensive coverage. the main concern is the stale-chatId race across tab switches: the existing `setSelectedTool` pattern in the same file already solves this for an equivalent per-tab String field, and switching to it removes a real edge from PR 3/3 before it lands. otherwise the refactor of `isExistingDuckAiChat` → `extractDuckAiChatId` is clean and the new test coverage is thorough.

(separate note for context: this PR adds a field to `NativeInputState` in `duckchat-api`. my own ongoing refactor (#8556 / #8570 / planned PR 3b) is co-evolving the same data class. no conflict in this PR's diff, but worth syncing on the field ownership pattern as more contributors add per-tab properties.)

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/NativeInputModeWidgetViewModel.kt
YoussefKeyrouz: Not an actual bug. `activeTabId` doesn't change for a given widget view model. The VM is fragment-scoped (obtained via `findViewTreeViewModelStoreOwner()` on a view inside the BrowserTabFragment) and each fragment has a fixed tabId read from its arguments. For a given BrowserTabFragment the tabId never changes. So the "combine fires with (oldChatId, newTabId)" race doesn't happen.
That said, the suggestion is still worth considering as a way to match how setSelectedTool works and drop some moving parts. Will see if it's worth doing in the PR.

PR #8669, opened 2026-05-22
Improve autocomplete behaviour 

aibrahim-:
Task/Issue URL:  https://app.asana.com/1/137249556945/project/1214157224317277/task/1214964760144656?focus=true

### Description
Three related fixes around the native-input widget's autocomplete:

1. Avoid spurious "loading from scratch" on widget re-open with same text. showNativeInput dropped the onClearAutocomplete call that emptied searchResults and re-triggered a fetch; restores the suggestions list visibility from the adapter when the cache matches the prefill.

2. Reveal the browser behind the closing widget earlier. hideNtp now runs at the start of hideNativeInput so the webpage is visible during the exit animation, instead of after the 200ms exit + 150ms fade leaves the NTP background filling the screen.

3. Always keep an autocomplete cache for the omnibar's text so in-widget typing never permanently displaces the omnibar's results. Adds an always-on parallel pipeline in BrowserTabViewModel (omnibarAutocompleteCache) fed by onOmnibarTextChanged; the widget restores from it on close (non-navigation) and on open if the cache matches the prefill.

### Steps to test this PR
- enable native input widget
- start a search by typing on the keyboard
- observe autocomplete 
- going back should immediately display the web page.
- lauching the keyboard again should immediately display the autocomplete list.

### UI changes
| Before  | After |
| ------ | ----- |
before|after|

---

> [!NOTE]
> **Medium Risk**
> Adds a second autocomplete pipeline and changes native-input open/close timing; behavior is gated on the native-input setting and covered by new ViewModel tests, but autocomplete and tab lifecycle edge cases warrant manual QA.
> 
> **Overview**
> Improves native-input omnibar autocomplete so reopening the keyboard can show prior suggestions without refetching, and closing the widget feels more like returning to the page you were on.
> 
> **Browser tab / ViewModel:** Adds an `omnibarAutocompleteCache` pipeline (debounced, IO-backed) driven by `onOmnibarTextRendered` when omnibar text is rendered—not user typing. It runs only when the native-input setting is on, skips full URLs and Duck Player URIs, and mirrors autocomplete settings. `restoreOmnibarAutocomplete` reapplies cached results into view state; the fragment also refreshes the suggestions adapter when the cache matches.
> 
> **Native input manager:** Wires `restoreOmnibarAutocomplete` on widget close (non-navigation) and before prefill on open; removes `onClearAutocomplete` on open and manually shows the list when cache restore succeeds. Calls `hideNtp()` at the start of hide in browser mode so the exit animation plays over the live page. Guards `onEnterComplete` with `isAttachedToWindow` to avoid raising the IME after detach.
> 
> **Widget enter animation:** `beginEnterAnimationPreview(isBottom)` applies per-tab `NativeInputState` (or a position-only fallback) before measure so toggle/layout does not snap mid-animation.
> 
> **Tests:** Coverage for restore/no-match, cache gating (setting off, autocomplete off, blank query), and telemetry state after restore.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3cffed1957c14c17c5c1885f2eb2b9e944e07a0e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -661,7 +661,13 @@ class BrowserTabViewModel @Inject constructor(
                         .debounce(300)
                         .distinctUntilChanged()
                         .flatMapLatest { query ->
-                            if (query.isBlank() || !autoCompleteSettings.autoCompleteSuggestionsEnabled) {
+                            // Skip page URLs — omnibarViewState emits on every navigation, and
+                            // running the fetch against full URLs would be both wasteful and
+                            // wrong (the autocomplete API expects user-typed queries).
+                            val isQueryLike = query.isNotBlank() &&
+                                !UriString.isWebUrl(query) &&
+                                !duckPlayer.isDuckPlayerUri(query)
+                            if (!isQueryLike || !autoCompleteSettings.autoCompleteSuggestionsEnabled) {
                                 flowOf(AutoCompleteResult(query, emptyList()))
                             } else {
                                 // Catch inside flatMapLatest so a failed fetch only drops this query's
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -649,6 +649,10 @@ class BrowserTabViewModel @Inject constructor(
     @OptIn(FlowPreview::class, ExperimentalCoroutinesApi::class)
     internal val omnibarAutocompleteCache: StateFlow<AutoCompleteResult> =
         duckChat.observeNativeInputFieldUserSettingEnabled()
+            .catch { t ->
+                logcat(WARN) { "Native input setting flow failed: ${t.asLog()}" }
+                emit(false)
+            }
             .flatMapLatest { isNativeInputEnabled ->
                 if (!isNativeInputEnabled) {
                     flowOf(AutoCompleteResult("", emptyList()))
@@ -660,13 +664,20 @@ class BrowserTabViewModel @Inject constructor(
                             if (query.isBlank() || !autoCompleteSettings.autoCompleteSuggestionsEnabled) {
                                 flowOf(AutoCompleteResult(query, emptyList()))
                             } else {
+                                // Catch inside flatMapLatest so a failed fetch only drops this query's
+                                // result — without it, .catch + stateIn(Eagerly) would terminate the
+                                // pipeline for the ViewModel's lifetime (no equivalent of the existing
+                                // pipeline's ConflatedJob restart).
                                 autoComplete.autoComplete(query)
+                                    .catch { t ->
+                                        logcat(WARN) { "Failed to get omnibar autocomplete: ${t.asLog()}" }
+                                        emit(AutoCompleteResult(query, emptyList()))
+                                    }
                             }
                         }
                 }
             }
             .flowOn(dispatchers.io())
-            .catch { t: Throwable? -> logcat(WARN) { "Failed to get omnibar autocomplete: ${t?.asLog()}" } }
             .stateIn(viewModelScope, SharingStarted.Eagerly, AutoCompleteResult("", emptyList()))
 
     private val fireproofWebsiteState: LiveData<List<FireproofWebsiteEntity>> = fireproofWebsiteRepository.getFireproofWebsites()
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -634,6 +636,56 @@ class BrowserTabViewModel @Inject constructor(
...
+            .catch { t ->
+                logcat(WARN) { "Native input setting flow failed: ${t.asLog()}" }
+                emit(false)
+            }
+            .flatMapLatest { isNativeInputEnabled ->
+                if (!isNativeInputEnabled) {
+                    flowOf(AutoCompleteResult("", emptyList()))
+                } else {
+                    omnibarTextStateFlow
+                        .debounce(300)
```
YoussefKeyrouz, line 661: Not Blocking: I think we could gain more speed on cold start by immediately firing the first emission before debouncing.  We already have a [utility function](https://github.com/duckduckgo/Android/blob/4a05a34b51f08145e05591b479b3b5d3d43d2e86/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/internal/Flows.kt#L27) for that that we are using in duckchat but will probably need to be move out or replicated. Will leave it to you to decide if you want to do that, but I don’t see a reason to delay the first fetch.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2785,6 +2837,39 @@ class BrowserTabViewModel @Inject constructor(
...
+     * Returns the cached result when it matched and was applied (null otherwise); callers
+     * should also refresh any adapter holding the suggestions since the renderer's
+     * `renderIfChanged` may skip a re-render for an equal view state.
+     *
+     * `showSuggestions` and `showFocusedView` are forced to false — the caller decides
+     * whether to surface the list.
+     */
+    fun restoreOmnibarAutocomplete(forQuery: String): AutoCompleteResult? {
+        val cached = omnibarAutocompleteCache.value
+        if (cached.query != forQuery || cached.suggestions.isEmpty()) return null
```
YoussefKeyrouz, line 2861: What happens if the suggestions are actually empty(e.g. searching for random thing like zxcsdcz)? Shouldn't we still update autoCompleteViewState to reflect the empty result? Right now it would still hold the previous query data because we return early and never set it. Unless I’m missing something.

aibrahim-: Suggestions are never empty, at the very least you get 2 options one for duck.ai and one for executing the search on whatever keywoard you entered anyway.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
aibrahim-: This is only the cache, not the actual user query execution, so the effect of it is very minimal. For the user to feel it, they have to try to hit the cache within 300ms of starting a new search.

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
```diff
@@ -342,11 +356,25 @@ class RealNativeInputManager @Inject constructor(
...
                 selectAllText()
             }
+            // hideNativeInput hid the list directly without touching autoCompleteViewState,
+            // so renderIfChanged skips re-showing on reopen. Surface it manually when the
+            // cache is for the current prefill; otherwise leave it hidden so a different
+            // query's stale items don't flash.
+            if (cacheRestored) {
+                rootView.findViewById<RecyclerView?>(R.id.autoCompleteSuggestionsList)?.let { list ->
+                    if ((list.adapter?.itemCount ?: 0) > 0) {
+                        list.show()
```
YoussefKeyrouz, line 374: nit: the renderAutocomplete function manipulates the visibility of the list every time  autoCompleteViewState changes. Calling show here would add another code path that controls visibility. It’s nit because I didn’t find a scenario currently where the 2 code paths disagree and might cause one to override the decision of the other, but it’s good to keep in mind if we can have a quick change to have only one owner control the visibility instead of 2 (maybe add a flag?). Not blocker just something to think about

aibrahim-: You are right, it's cleaner to have one owner.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
YoussefKeyrouz: Ah right. We can do a follow up for the actual query execution to speed up the first tap. but not part of this PR.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
YoussefKeyrouz: I’m just worried that some settings might turn off those (we have too many settings to keep up) so I usually implement it with that in mind. But for this, it’s fine I don’t think we’ll get to that scenario.

YoussefKeyrouz (approved):
Looks good. Thanks!. Tested few scenarios and they worked as expected.

Left few nits that we can address later as follow ups.

PR #8683, opened 2026-05-26
Sync Exchange v2 algorithm encryption foundation work

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1203822806345703/task/1215115673129510?focus=true

### Description
Foundational encryption work for supporting duck chat syncing from 3rd party browsers.

The main goal of this PR is to add the building blocks we'll need for supporting the functionality in a way that:
- can be tested in internal sync dev tooling
- does not affect production code

#### Ability to create protected keys and 3p access credentials

> [!TIP]
> Logcat filter: `message~:"Sync-Scoped"`

> [!TIP]
> There is a screen recording of the required setup provided as a reference in [Video recording of setting up Android to sync with 3rd party browser using dev tools](https://app.asana.com/1/137249556945/task/1215121243098289?focus=true)

**Testing refusal when FF disabled**
- [x] Fresh install `internal` variant
- [x] Launch Sync Dev Settings (SDS) screen
- [x] Scroll all the way down and tap on `Create account`; wait a few seconds and make sure account settings populate
- [x] Back up in the `V2 connect flow flags` section, tap on `Fetch Access Credentials` and verify you see `id=ddg,scope=(unrestricted)`  as you don't currently have any 3rd party access credentials
- [x] Tap on `Fetch Keys` and verify you see `No keys found` as you don't have any created yet
- [x] Tap on `Create Protected Key` (leave purpose set to default `ai_chats`) and verify this fails because the flag is disabled. 
- [x] Tap on `Create 3party Credential` and verify it fails because the flag is disabled

**Creating protected key**
- [x] Enable flag using the `canUseV2ConnectFlow (master kill switch)`
- [x] Tap on `Create Protected Key` (leave purpose set to default `ai_chats`) and verify it succeeds, verifying that under `Fetch Keys` you see something with `purpose=ai_chats, encrypted_with=ddg`

**Creating 3party credential**
- [x] Tap on `Refresh 3party Credential from Server` and verify there are none
- [x] Tap on `Create 3party Credential` and verify it's successful

> [!IMPORTANT]
> This builds on the above scenario so make sure that's been done

**Android**
- [x] Tap on `Open Duck.ai Dev Screen` -> `URL Settings` and enter duck.ai chat URL from [Test Server details](https://app.asana.com/1/137249556945/task/1214777401961082?focus=true). Save and it'll restart some things.
- [x] Return to sync dev screen and tap on `Open Duck.ai Setup URL`. Authenticate with SSO. When prompted, choose `DEVELOPMENT` server.

**3rd party browser (e.g., Chrome desktop)**
- [x] Follow instructions for `3rd Party Setup` in [Test Server details](https://app.asana.com/1/137249556945/task/1214777401961082?focus=true), again choosing `DEVELOPMENT` server when prompted

**Android**
- [x] Return to sync dev screen, and tap on `Copy 3party Recovery Code`

**3rd party browser (e.g., Chrome desktop)**
- [x] Down the bottom left, choose `Settings & More` -> `Settings` -> `Turn On Sync & Backup`
- [x] Choose `Sync With Another Device`
- [x] Paste the 3party recovery code you got from Android. Verify you see the "New Device Added" success dialog.
- [x] Start a new chat, e.g., called "Hello from 3rd party browser"

**Android**
- [x] Open duck chat; verify you see the chat from 3rd party browser syncing in after a few seconds. Open it, and add a reply.
- [x] Start a new duck chat, e.g., called "Hello from Android"

**3rd party browser (e.g., Chrome desktop)**
- [x] After a few seconds to sync, make sure you see the new "Hello from Android" chat
- [x] Open the original "Hello from 3rd party browser" chat and verify you see the reply you typed on Android

> [!TIP]
> If all of the above is working, then you have established duck chat between Android and a 3rd party browser 🎉. That's largely it for this PR for now, other than ensuring nothing in prod is broken (see below)

#### Ensuring prod behavior is ok
Below is some of the main smoke testing of prod that is required, but isn't exhaustive. Any further smoke testing of typical prod sync pairing is helpful to ensure no regressions.

> [!NOTE]
> You'll need two emulators/devices for this

- [ ] Fresh install `internal` variants on two emulators/devices
- On **Device A**: 
    - [ ] Open production Sync Settings, and tap on `Sync and Back Up This Device`
    - [ ] Complete the flow and verify it works successfully
    - [ ] Tap on `Sync With Another Device`
- On **Device B**
    - [ ] Open production Sync Settings, and tap on on `Sync With Another Device` 
    - [ ] Connect the two, ensuring they connect successfully and are syncing

---

> [!NOTE]
> **High Risk**
> Large changes to sync authentication, multi-step credential upgrade, and cryptographic key handling; production impact depends on feature flags but mistakes when enabled would affect account pairing and secrets.
> 
> **Overview**
> Adds the **v2 sync connect/exchange stack** behind `canUseV2ConnectFlow` (default off): a `SyncCodeDispatcher` routes pasted/scanned codes to legacy v1 or v2 linking/recovery flows, drives relay pairing via `ExchangeV2Runner`, and maps outcomes (confirmations, login, failures) for UI callers.
> 
> **Scoped access credentials** extend the account layer: JWE/HKDF helpers (`SyncJweCrypto`, jjwt), 3party credential create/refresh/recovery codes, protected RSA keys for purposes like `ai_chats`, native **3party→ddg upgrade** (`joinAccountFromThirdPartyRecoveryCode`), and `entries_v2` device list decryption (ddg vs 3party). Sync API/Retrofit gains v2 exchange relay, access-credentials, and protected-keys endpoints; signup/login can send `credential_id` / `scope` when the flag is on.
> 
> New **`sync_setup_*` pixel definitions** instrument v1/v2 setup screens and outcomes. With flags disabled, code routing and device listing stay on the existing v1 paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6987020fa2ff8e6f4d087c3704e0522c3465782b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ProtectedKeyManager.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ProtectedKeyManager.kt
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.sync.impl
+
+import android.util.Base64
+import androidx.annotation.WorkerThread
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.sync.crypto.SyncLib
+import com.duckduckgo.sync.impl.Result.Error
+import com.duckduckgo.sync.impl.Result.Success
+import com.duckduckgo.sync.impl.crypto.SyncJweCrypto
+import com.duckduckgo.sync.store.SyncStore
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import logcat.LogPriority.ERROR
+import logcat.logcat
+import java.util.UUID
+import javax.inject.Inject
+
+/**
+ * Creates and stores RSA keypairs ("protected keys") on the account, wrapped with the ddg
+ * credential's secret. Used to mint per-purpose keys (e.g. for ai_chats) that other devices on
+ * the account can decrypt to receive end-to-end-encrypted payloads addressed to a known kid.
+ */
+interface ProtectedKeyManager {
+
+    /**
+     * Generates an RSA keypair for [purpose] (e.g. "ai_chats"), libsodium-encrypts the private key
+     * with the account secret key, and POSTs the entry via /sync/keys/.../set-if-absent.
+     *
+     * No-op success if the server already has a key for [purpose] (set-if-absent semantics).
+     */
+    fun create(purpose: String): Result<Boolean>
+}
+
+@ContributesBinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
+@WorkerThread
+class RealProtectedKeyManager @Inject constructor(
+    private val syncStore: SyncStore,
+    private val syncApi: SyncApi,
+    private val syncJweCrypto: SyncJweCrypto,
+    private val nativeLib: SyncLib,
+    private val syncFeature: SyncFeature,
+) : ProtectedKeyManager {
+
+    override fun create(purpose: String): Result<Boolean> {
+        if (!syncFeature.canUseV2ConnectFlow().isEnabled()) {
+            return Error(reason = "Scoped access credentials feature is disabled")
+        }
+        val token = syncStore.token.takeUnless { it.isNullOrEmpty() }
+            ?: return Error(reason = "CreateProtectedKey: not signed in")
+        val secretKey = syncStore.secretKey
+            ?: return Error(reason = "CreateProtectedKey: no secret key for libsodium encryption")
+
+        logcat { "Sync-ScopedToken: creating protected key for purpose=$purpose" }
+
+        val rsaKeyPair = kotlin.runCatching {
+            syncJweCrypto.generateRsaKeyPair()
+        }.getOrElse { return it.asLoggedError("CreateProtectedKey: failed to generate RSA keypair") }
+        val (n, e) = kotlin.runCatching {
+            syncJweCrypto.extractJwkComponents(rsaKeyPair.publicKeyBase64)
+        }.getOrElse { return it.asLoggedError("CreateProtectedKey: failed to extract JWK components") }
+
+        // ddg-side wrap: libsodium-secretbox(privateKey, secretKey), base64url-encoded.
+        // Matches duck.ai's EncryptWithSyncMasterKeyHandler so native clients read keys uniformly.
+        val privateKeyBytes = kotlin.runCatching {
+            Base64.decode(rsaKeyPair.privateKeyBase64, Base64.URL_SAFE or Base64.NO_WRAP or Base64.NO_PADDING)
+        }.getOrElse { return it.asLoggedError("CreateProtectedKey: failed to decode private key bytes") }
+        val encryptedPrivateKey = kotlin.runCatching {
+            val result = nativeLib.encryptData(privateKeyBytes, secretKey).also {
+                it.checkResult("CreateProtectedKey: libsodium encryption of private key failed")
+            }
+            Base64.encodeToString(result.encryptedData, Base64.NO_WRAP).applyUrlSafetyFromB64()
+        }.getOrElse { return it.asErrorResult() }
+
+        val kid = UUID.randomUUID().toString()
+        val key = ProtectedKeyEntry(
+            kid = kid,
+            purpose = purpose,
+            encryptedWith = CREDENTIAL_ID_DDG,
+            encryptedPrivateKey = encryptedPrivateKey,
+            publicKey = RsaJwk(n = n, e = e),
+        )
+
+        return when (val result = syncApi.setProtectedKeyIfAbsent(token, purpose, key)) {
+            is Success -> {
+                logcat { "Sync-ScopedToken: protected key for $purpose created (kid=$kid)" }
+                Success(true)
+            }
+            is Error -> {
+                logcat(ERROR) { "Sync-ScopedToken: failed to create protected key: ${result.reason}" }
+                result
+            }
+        }
+    }
+}
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ProtectedKeyManager.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ProtectedKeyManager.kt
@@ -42,9 +42,14 @@ interface ProtectedKeyManager {
      * Generates an RSA keypair for [purpose] (e.g. "ai_chats"), libsodium-encrypts the private key
      * with the account secret key, and POSTs the entry via /sync/keys/.../set-if-absent.
      *
-     * No-op success if the server already has a key for [purpose] (set-if-absent semantics).
+     * Returns the authoritative [ProtectedKeyEntry] taken from the server's response. In the
+     * we-wrote-it case the returned entry equals the one we POSTed; in the race-loser case the
+     * server's `set-if-absent` returns its pre-existing entry instead, and we surface that so
+     * the caller reflects the truth on the server rather than the local key we minted.
+     *
+     * Returns the entry for [purpose] only — not all keys for the account.
      */
-    fun create(purpose: String): Result<Boolean>
+    fun create(purpose: String): Result<ProtectedKeyEntry>
 }
 
 @ContributesBinding(AppScope::class)
@@ -58,7 +63,7 @@ class RealProtectedKeyManager @Inject constructor(
     private val syncFeature: SyncFeature,
 ) : ProtectedKeyManager {
 
-    override fun create(purpose: String): Result<Boolean> {
+    override fun create(purpose: String): Result<ProtectedKeyEntry> {
         if (!syncFeature.canUseV2ConnectFlow().isEnabled()) {
             return Error(reason = "Scoped access credentials feature is disabled")
         }
@@ -97,10 +102,21 @@ class RealProtectedKeyManager @Inject constructor(
             publicKey = RsaJwk(n = n, e = e),
         )
 
-        return when (val result = syncApi.setProtectedKeyIfAbsent(token, purpose, key)) {
+        return when (val result = syncApi.setProtectedKeyIfAbsent(token, purpose, listOf(key))) {
             is Success -> {
-                logcat { "Sync-ScopedToken: protected key for $purpose created (kid=$kid)" }
-                Success(true)
+                val entry = result.data.firstOrNull { it.purpose == purpose && it.encryptedWith == CREDENTIAL_ID_DDG }
+                    ?: return Error(reason = "CreateProtectedKey: server response missing created key for $purpose")
+                if (entry.kid != kid) {
+                    // Server returned a pre-existing entry — another device created the key for this
+                    // purpose between our (implicit) check and our POST. The server's entry wins.
+                    logcat {
+                        "Sync-ScopedToken: protected key for $purpose already existed on server " +
+                            "(server kid=${entry.kid}, our kid=$kid) — adopting server entry"
+                    }
+                } else {
+                    logcat { "Sync-ScopedToken: protected key for $purpose created (kid=$kid)" }
+                }
+                Success(entry)
             }
             is Error -> {
                 logcat(ERROR) { "Sync-ScopedToken: failed to create protected key: ${result.reason}" }
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ScopedCredentialsEncoding.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ScopedCredentialsEncoding.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.sync.impl
+
+import android.util.Base64
+import com.duckduckgo.sync.impl.Result.Error
+import com.duckduckgo.sync.impl.crypto.SyncJweCrypto
+import logcat.LogPriority.ERROR
+import logcat.asLog
+import logcat.logcat
+
+/**
+ * HKDF-SHA-256 over the decoded `base64Key` bytes, returning raw bytes for direct use as an AES key.
+ * Used to derive the credential MEK (info="Main Key") per Encryption Algorithms TD (Asana 1214802412121967).
+ */
+internal fun SyncJweCrypto.hkdfDeriveBytes(
+    base64Key: String,
+    salt: ByteArray,
+    info: String,
+    outBytes: Int = 32,
+): ByteArray {
+    val raw = Base64.decode(base64Key, Base64.NO_WRAP)
+    return hkdfSha256SingleBlock(raw, salt, info, outBytes)
+}
+
+/**
+ * HKDF-SHA-256 over the decoded `base64Key` bytes, re-encoded as base64url (no padding).
+ * Used for 3party HKDF derivations: `credential_hashed_password` (info="Password") per
+ * Encryption Algorithms TD (Asana 1214802412121967).
+ *
+ * Not for ddg auth — that uses libsodium BLAKE2b via `nativeLib.prepareForLogin`.
+ */
+internal fun SyncJweCrypto.hkdfDeriveBase64Url(
+    base64Key: String,
+    salt: ByteArray,
+    info: String,
+    outBytes: Int = 32,
+): String {
+    val raw = Base64.decode(base64Key, Base64.NO_WRAP)
+    val out = hkdfSha256SingleBlock(raw, salt, info, outBytes)
+    return Base64.encodeToString(out, Base64.URL_SAFE or Base64.NO_PADDING or Base64.NO_WRAP)
+}
+
+/** Convert a base64url SP from the wire into the standard-base64 form used locally. */
+internal fun base64UrlStringToStandardBase64(b64Url: String): String {
+    val raw = Base64.decode(b64Url, Base64.URL_SAFE or Base64.NO_PADDING or Base64.NO_WRAP)
+    return Base64.encodeToString(raw, Base64.NO_WRAP)
+}
+
+/** Log the throwable's full stack trace, then return an [Error] tagged with [reason]. */
+internal fun Throwable.asLoggedError(reason: String): Error {
+    logcat(ERROR) { "Sync-ScopedToken: $reason: ${asLog()}" }
+    return Error(reason = "$reason: $message")
+}
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
@@ -41,6 +41,7 @@ import com.duckduckgo.sync.impl.SyncAuthCode.Connect
 import com.duckduckgo.sync.impl.SyncAuthCode.Exchange
 import com.duckduckgo.sync.impl.SyncAuthCode.Recovery
 import com.duckduckgo.sync.impl.SyncAuthCode.Unknown
+import com.duckduckgo.sync.impl.crypto.SyncJweCrypto
 import com.duckduckgo.sync.impl.metrics.ConnectedDevicesObserver
 import com.duckduckgo.sync.impl.pixels.*
 import com.duckduckgo.sync.impl.ui.qrcode.SyncBarcodeUrl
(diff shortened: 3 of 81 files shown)


CDRussell:
* **#8683** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8683?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#8695](https://github.com/duckduckgo/Android/pull/8695) <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8695?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#8795](https://github.com/duckduckgo/Android/pull/8795) <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8795?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8683?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/crypto/SyncJweCrypto.kt
```diff
@@ -0,0 +1,153 @@
...
+            .decryptWith(secretKey)
+            .build()
+
+        return runCatching {
+            parser.parseEncryptedContent(jweCompact).payload
+        }.getOrElse { contentError ->
+            logcat { "Sync-ScopedToken: parseEncryptedContent failed (${contentError.message}); falling back to parseEncryptedClaims" }
+            val claims = parser.parseEncryptedClaims(jweCompact).payload
+            JSONObject(claims).toString().toByteArray()
+        }
```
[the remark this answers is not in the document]
CDRussell: Valid; fixed by removing fallback as after reviewing the TD and spec there is no valid path that should land us there.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyCredentialManager.kt
```diff
@@ -0,0 +1,370 @@
...
+                Success(true)
+            }
+            is Error -> {
+                logcat(ERROR) { "Sync-ScopedToken: failed to create 3party credential: ${result.reason}" }
+                if (result.code == API_CODE.COUNT_LIMIT.code) {
+                    logcat { "Sync-ScopedToken: 409 conflict - another device created the credential first" }
+                }
+                result
+            }
+        }
```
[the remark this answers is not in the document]
CDRussell: Valid; fixed by implementing the 409 (which is correct per the spec)

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
```diff
@@ -755,6 +811,10 @@ class AppSyncAccountRepository @Inject constructor(
 
             is Success -> {
                 syncStore.storeCredentials(account.userId, deviceId, deviceName, account.primaryKey, account.secretKey, result.data.token)
+                if (syncFeature.canUseV2ConnectFlow().isEnabled()) {
```
cmonfortep, line 1162: we should move this above `storeCredentials` or inside it. `storeCrendentials` will emit signed in and best if all the store has all the expected data before emit.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
```diff
@@ -811,6 +871,37 @@ class AppSyncAccountRepository @Inject constructor(
                 }
 
                 syncStore.storeCredentials(userId, deviceId, deviceName, preLogin.primaryKey, decryptResult.decryptedData, result.data.token)
+
+                if (syncFeature.canUseV2ConnectFlow().isEnabled()) {
```
cmonfortep, line 1330: we should move this above `storeCredentials` or inside it. `storeCrendentials` will emit signed in and best if all the store has all the expected data before emit.

I think is better to compute values before, and passing new params to the method so we can do the write internally.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/crypto/SyncJweCrypto.kt
```diff
@@ -0,0 +1,146 @@
...
+     */
+    fun hkdfSha256SingleBlock(input: ByteArray, salt: ByteArray, info: String, outBytes: Int = 32): ByteArray
+}
+
+data class RsaKeyPair(
+    val publicKeyBase64: String,
+    val privateKeyBase64: String,
+)
+
+@ContributesBinding(AppScope::class)
```
cmonfortep, line 92: Add `@SingleInstanceIn(AppScope::class)` since this is stateless and it can be a singleton

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/crypto/SyncJweCrypto.kt
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    override fun jweEncryptSymmetric(plaintext: ByteArray, symmetricKey: ByteArray, kid: String?): String {
+        logcat { "Sync-ScopedToken: JWE encrypting with symmetric key (dir + A256GCM, kid=$kid)" }
+        val secretKey = SecretKeySpec(symmetricKey, "AES")
+
+        val builder = Jwts.builder()
+        if (kid != null) builder.header().keyId(kid).and()
+        return builder
+            .content(plaintext, JWE_CONTENT_TYPE_OCTET_STREAM)
```
cmonfortep, line 116: This was raised by claude, posting here for review. Seems not harmful to keep, but checking if intentional (if so, we should guard with test)

>  jweEncryptSymmetric calls Jwts.builder()...content(plaintext, JWE_CONTENT_TYPE_OCTET_STREAM).... Per jjwt's API, the
   two-argument .content(bytes, contentType) form sets the cty header. The TD reference envelope header in tdTest3 is:
  {"alg":"dir","enc":"A256GCM","kid":"ddg"}
  — no cty. Android's produced header is likely:
  {"alg":"dir","enc":"A256GCM","kid":"ddg","cty":"application/octet-stream"}
  
  This isn't a decryption problem — the protected-header bytes are AAD for AES-GCM, and the decryptor uses the
  received bytes (not an expected shape). So FE decrypting an Android-produced envelope will compute the same AAD
  Android wrote, and decryption succeeds, if the FE impl tolerates unknown headers (which RFC 7516 requires).

  But: tdTest3_envelopeHeaderHasExpectedFields only asserts alg/enc/kid are present + segment count = 5. It doesn't
  assert cty is absent, nor that the keyset is exactly those three. The "Android-produces an envelope FE can decrypt"
  direction isn't unit-tested anywhere — only verified at integration level via the PR's manual test plan with a
  3rd-party browser.

  Two cheap improvements:
  - Switch to single-argument content(bytes) so jjwt doesn't emit cty, making Android's header byte-for-byte identical
   to the TD reference.
  - Tighten tdTest3_envelopeHeaderHasExpectedFields to assert the exact keyset (e.g.
  assertEquals(setOf("alg","enc","kid"), header.keys().asSequence().toSet())) so any future header drift is caught.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/crypto/SyncJweCrypto.kt
```diff
@@ -0,0 +1,146 @@
...
+import java.security.spec.X509EncodedKeySpec
+import java.util.Base64
+import javax.crypto.Mac
+import javax.crypto.spec.SecretKeySpec
+import javax.inject.Inject
+
+private const val JWE_CONTENT_TYPE_OCTET_STREAM = "application/octet-stream"
+
+interface SyncJweCrypto {
+    fun generateRsaKeyPair(): RsaKeyPair
```
cmonfortep, line 46: Add a @WorkerThread annotation on generateRsaKeyPair() (or the whole interface)

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ScopedCredentialsEncoding.kt
```diff
@@ -0,0 +1,68 @@
...
+
+import android.util.Base64
+import com.duckduckgo.sync.impl.Result.Error
+import com.duckduckgo.sync.impl.crypto.SyncJweCrypto
+import logcat.LogPriority.ERROR
+import logcat.asLog
+import logcat.logcat
+
+/**
+ * HKDF-SHA-256 over the decoded `base64Key` bytes, returning raw bytes for direct use as an AES key.
```
cmonfortep, line 27: Let's add this lines to make the expectation clear

```
   * @param base64Key the secret in **standard** base64 form (local storage format, e.g. [SyncStore.primaryKey] or 
  [ScopedPassword.raw]).
   * @return raw HKDF output bytes — 32 bytes when [outBytes] is left at default, suitable for direct use as an AES-256 key.
```

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ScopedCredentialsEncoding.kt
```diff
@@ -0,0 +1,68 @@
...
+): ByteArray {
+    val raw = Base64.decode(base64Key, Base64.NO_WRAP)
+    return hkdfSha256SingleBlock(raw, salt, info, outBytes)
+}
+
+/**
+ * HKDF-SHA-256 over the decoded `base64Key` bytes, re-encoded as base64url (no padding).
+ * Used for 3party HKDF derivations: `credential_hashed_password` (info="Password") per
+ * Encryption Algorithms TD (Asana 1214802412121967).
+ *
```
cmonfortep, line 44: Similar here

```
  /**
   * ...
   * @param base64Key the secret in **standard** base64 form (local storage format, e.g. [SyncStore.primaryKey] or 
  [ScopedPassword.raw]).
   * @return HKDF output re-encoded as **base64url, no padding** (wire format). Suitable for upload as 
  `credential_hashed_password`.
   */
```

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncServiceRemote.kt
```diff
@@ -503,6 +523,66 @@ class SyncServiceRemote @Inject constructor(
...
+    override fun getProtectedKeys(token: String): Result<List<ProtectedKeyEntry>> {
+        val response = runCatching {
+            val call = syncService.getProtectedKeys("Bearer $token")
+            call.execute()
+        }.getOrElse { throwable ->
+            return Result.Error(reason = throwable.message.toString())
+        }
+
+        return onSuccess(response) {
+            val keys = response.body()?.keys ?: emptyList()
```
cmonfortep, line 565: In `getAccessCredentials` we do return error if keys missing. Is returning emptyList here intended? If so, document with one comment why.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncService.kt
```diff
@@ -151,6 +175,7 @@ data class Signup(
     @field:Json(name = "device_id") val deviceId: String,
     @field:Json(name = "device_name") val deviceName: String,
     @field:Json(name = "device_type") val deviceType: String,
+    @field:Json(name = "credential_id") val credentialId: String? = null,
```
cmonfortep, line 203: is it safe to include `null` in this object? I assume with flag on this will always have value, with flag disabled this will be empty/null. BE is happy with both, should be excluded?

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncService.kt
```diff
@@ -210,6 +237,66 @@ data class TokenRescopeResponse(
...
+ * One entry in GET /sync/access-credentials. Each entry represents an access credential the
+ * account holds (id "ddg" for the native credential, "3party" for the scoped-access one shared
+ * with browser surfaces). `encryptedCredential` is the credential's secret encrypted with the
+ * companion credential's MEK (e.g. for id="3party", encrypted with the DDG MEK).
+ *
+ */
+data class AccessCredentialEntry(
+    val id: String,
+    val scope: String? = null,
+    @field:Json(name = "encrypted_3party_credential") val encryptedCredential: String? = null,
```
cmonfortep, line 303: is `encryptedCredential` a 3partycrendential? or based on id? if it's always 3rd party, I would include it in the name for clarity.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncServiceRemote.kt
```diff
@@ -503,6 +523,66 @@ class SyncServiceRemote @Inject constructor(
         }
     }
 
+    override fun setProtectedKeyIfAbsent(token: String, purpose: String, key: ProtectedKeyEntry): Result<Boolean> {
```
cmonfortep: do we need to do any handling if keys were already present or similar?
or if the keys are present, does this error or goes through success flow?

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyCredentialManager.kt
```diff
@@ -0,0 +1,377 @@
...
+        return Success(CreateInputs(token, primaryKey, userId, userId.toByteArray(Charsets.UTF_8)))
+    }
+
+    /** If another device already created the 3party credential, decrypt its SP and store locally. */
+    private fun tryAdoptExistingCredential(inputs: CreateInputs): AdoptResult {
+        val existing = when (val r = syncApi.getAccessCredentials(inputs.token)) {
+            is Success -> r.data.find { it.id == CREDENTIAL_ID_3PARTY } ?: return AdoptResult.NotFound
+            is Error -> {
+                logcat(ERROR) { "Sync-ScopedToken: failed to check existing credentials: ${r.reason}" }
+                return AdoptResult.Failed(r.copy(code = GENERIC_ERROR.code))
```
cmonfortep, line 131: do we want to override the code received from `syncApi.getAccessCredentials`? Is there any cleanup we need to perform based on the server response, if so we will lose visibility.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyCredentialManager.kt
```diff
@@ -0,0 +1,377 @@
...
+        logcat { "Sync-ScopedToken: refreshing 3party credential from server" }
+
+        val existing = when (val r = syncApi.getAccessCredentials(token)) {
+            is Success -> r.data.find { it.id == CREDENTIAL_ID_3PARTY } ?: run {
+                logcat { "Sync-ScopedToken: no 3party credential on server — nothing to refresh" }
+                return Success(false)
+            }
+            is Error -> {
+                logcat(ERROR) { "Sync-ScopedToken: failed to fetch access credentials: ${r.reason}" }
+                return r.copy(code = GENERIC_ERROR.code)
```
cmonfortep, line 337: ditto

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyCredentialManager.kt
```diff
@@ -0,0 +1,377 @@
...
+        if (!syncFeature.canUseV2ConnectFlow().isEnabled()) {
+            return Error(reason = "Scoped access credentials feature is disabled")
+        }
+        val token = syncStore.token.takeUnless { it.isNullOrEmpty() }
+            ?: return Error(reason = "RefreshThirdPartyCredential: not signed in")
+        logcat { "Sync-ScopedToken: refreshing 3party credential from server" }
+
+        val existing = when (val r = syncApi.getAccessCredentials(token)) {
+            is Success -> r.data.find { it.id == CREDENTIAL_ID_3PARTY } ?: run {
+                logcat { "Sync-ScopedToken: no 3party credential on server — nothing to refresh" }
```
cmonfortep, line 332: If the server says "no 3party credential exists" but syncStore.scopedPassword is non-null locally, the local state is stale. A subsequent
  getRecoveryCode() would emit a recovery code that no browser could use to sync with the account (the credential is gone server-side).

  Probably should clear local SP on this branch:
  syncStore.scopedPassword = null
  return Success(false)

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
```diff
@@ -878,27 +969,41 @@ class AppSyncAccountRepository @Inject constructor(
         companion object {
             private val moshi = Moshi.Builder().build()
             val recoveryCodeAdapter: JsonAdapter<LinkCode> = moshi.adapter(LinkCode::class.java)
+            val thirdPartyRecoveryCodeAdapter: JsonAdapter<ThirdPartyRecoveryCodeWrapper> = moshi.adapter(ThirdPartyRecoveryCodeWrapper::class.java)
```
cmonfortep, line 1450: This is unused as you have the lazy adapter in `ThirdPartyCredentialManager`

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyCredentialManager.kt
```diff
@@ -0,0 +1,377 @@
...
+        val decryptedBase64Url = kotlin.runCatching {
+            String(syncJweCrypto.jweDecryptSymmetric(encryptedSp, ddgMek), Charsets.UTF_8)
+        }.getOrElse {
+            logcat(ERROR) { "Sync-ScopedToken: failed to decrypt SP from server: ${it.message}" }
+            return Error(reason = "$errorPrefix: failed to decrypt SP: ${it.message}")
+        }
+        return Success(base64UrlStringToStandardBase64(decryptedBase64Url))
+    }
+
+    override fun refresh(): Result<Boolean> {
```
cmonfortep, line 322: is refresh only expected to refresh scoped password? or also the keys? while this is documented, I'd consider adding it into the name. But I'm nor sure if SP and keys should be in sync, meaning they need to be refreshed at the same time.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ProtectedKeyManager.kt
```diff
@@ -0,0 +1,111 @@
...
+            syncJweCrypto.generateRsaKeyPair()
+        }.getOrElse { return it.asLoggedError("CreateProtectedKey: failed to generate RSA keypair") }
+        val (n, e) = kotlin.runCatching {
+            syncJweCrypto.extractJwkComponents(rsaKeyPair.publicKeyBase64)
+        }.getOrElse { return it.asLoggedError("CreateProtectedKey: failed to extract JWK components") }
+
+        // ddg-side wrap: libsodium-secretbox(privateKey, secretKey), base64url-encoded.
+        // Matches duck.ai's EncryptWithSyncMasterKeyHandler so native clients read keys uniformly.
+        val privateKeyBytes = kotlin.runCatching {
+            Base64.decode(rsaKeyPair.privateKeyBase64, Base64.URL_SAFE or Base64.NO_WRAP or Base64.NO_PADDING)
```
cmonfortep, line 87: I think will be best if we can reuse `Base64.URL_SAFE or Base64.NO_WRAP or Base64.NO_PADDING` and give it a name. Also to understand why sometimes we do `Base64.NO_WRAP).applyUrlSafetyFromB64()`

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
```diff
@@ -811,6 +871,37 @@ class AppSyncAccountRepository @Inject constructor(
...
+                if (syncFeature.canUseV2ConnectFlow().isEnabled()) {
+                    logcat { "Sync-ScopedToken: login response has v2 data, storing" }
+                    syncStore.credentialId = CREDENTIAL_ID_DDG
+                    result.data.accessCredentials?.let { credentials ->
+                        logcat { "Sync-ScopedToken: ${credentials.size} access credential(s) in response" }
+                        val thirdParty = credentials.find { it.id == CREDENTIAL_ID_3PARTY }
+                        val encryptedSp = thirdParty?.encryptedCredential
+                        if (encryptedSp != null) {
+                            // Decrypt with the DDG MEK; convert wire base64url to standard base64 for storage.
+                            val decryptedSp = kotlin.runCatching {
```
cmonfortep, line 1339: All the logic related to SP decrypt should live inside `ThirdPartyCredentialManager`, which is where you already have similar logic. we should expose a method for that.

Asked claude recommendation for this:

  Better shape: have the manager expose applyLoginResponse(response, primaryKey, userId) that handles SP decrypt, keys filter, and (in the
  proposed atomic-refactor world) returns a V2Credentials bundle. The repo then:
  val v2 = thirdPartyCredentialManager.buildV2FromLoginResponse(result.data, preLogin.primaryKey, userId)
  syncStore.storeCredentials(userId, deviceId, deviceName, preLogin.primaryKey, decryptResult.decryptedData, result.data.token, v2 = v2)

  This consolidates the SP-decrypt logic, makes the manager the single source of truth, and naturally enables the atomic-write refactor.

cmonfortep (approved):
Testing happy path works, and approving as this is first version and nothing blocking. 

There are a few things we need to discuss and decide next steps, so they might be easier on follow up PRs. Not specific suggestions as they rely on specifications and here we are just doing foundation work:
- I'm interested on error handling and how to clear store. Right now we have a few new endpoints, they can error, or BE can return some specific errors. It's not clear how to react to each of them.
- I'm curious what's the plan after enabling the flag for accounts that already exist. They won't have credential Id, SP, etc.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/crypto/SyncJweCrypto.kt
CDRussell: 👍 whole interface works for me

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
CDRussell: Good spot; makes sense to re-order as you propose.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncAccountRepository.kt
CDRussell: That sounds good to me.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/crypto/SyncJweCrypto.kt
CDRussell: I think it shouldn't matter, but that dropping `cty` would be fine. as long as it works against the reference impl i'm good with it

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncServiceRemote.kt
CDRussell: hmmm, probably not intended. best to convert this to copy `getAccessCredentials()` approach and error on null rather than empty list (need to test to confirm, but sounds right)

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncService.kt
CDRussell: though one of the few things to not be contained behind a FF, this should be fine

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/SyncService.kt
CDRussell: always 3rd party; renaming sounds good

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyCredentialManager.kt
CDRussell: quick thoughts: 

from the response we probably do want to:

- propagate the error so that the likes of `401` can be handled separately from other errors
- treat the server's response as a the source of truth (i.e., insert or update)

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ProtectedKeyManager.kt
CDRussell: Summarizing from our call for posterity: 

depends what you're starting out with, whether you have a plain string needing encoded or already starting out with a base64 (not base64url) encoded string:
- `applyUrlSafetyFromB64` was added for when we already receive a base64 encoded string that we want to take that one step further and convert to base64url encoding
- if you're starting with the plain string rather than a base64-encoded one already, this should one-shot it: `Base64.URL_SAFE or Base64.NO_WRAP or Base64.NO_PADDING`

PR #8692, opened 2026-05-27
Fix onboarding flow background issues when native input is enabled

YoussefKeyrouz:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1212810093780571/task/1214882981544707?focus=true 

### Description

NOTE: To test this PR you can use the [do_not_merge/youssef/test_onboarding_native_input](https://github.com/duckduckgo/Android/tree/do_not_merge/youssef/test_onboarding_native_input) branch where I enabled the native input by default. 

Two onboarding-flow issues that surface when the native input widget is enabled:

1. **Onboarding bubble leaks behind the native input widget on a webpage.** After completing onboarding and navigating to a page, tapping the address bar opens the native input widget over the new tab layout. The previous NTP bubble reappears under the widget, most visibly when there are no autocomplete suggestions covering the gap below the input card.

2. **Tapping a dax CTA option doesn't dismiss the widget, leaving the address bar with stale text.** The widget stays open after submit, showing whatever was in its EditText, while the omnibar underneath updates to the new URL. 

#### Root cause / fix

- **Renderer fix** in `BrowserTabFragment.renderCtaViewState`: clear the previous bubble's view state when the renderer fires, and skip `showCta` for `DaxBubbleCta`s arriving while the browser is showing. The `DaxBubbleCta` family is NTP-only; the contextual `OnboardingDaxDialogCta` family (trackers blocked, fire button, SERP, etc.) paints into a different container inside `browserLayout` and continues to render normally in browser mode.
- **CTA option click fix**: route the three dax option-click sites through the existing `submitQuery` instead of `userEnteredQuery`. `submitQuery` calls `hideNativeInput` before `onUserSubmittedQuery`.

### Steps to test this PR

_Onboarding bubble doesn't leak behind the native input widget_

- [ ] Fresh install the app.
- [ ] Complete onboarding choosing **Search only** when prompted about the address bar toggle.
- [ ] On the NTP, observe the "Try a search!" bubble.
- [ ] Submit a search (or tap one of the suggestion options).
- [ ] Once the SERP loads, tap the address bar.
- [ ] Clear the URL/text in the widget (or leave it as-is).
- [ ] **Verify:** The "Try a search!" bubble is **not** visible behind the widget, neither when text is present nor when cleared.
- [ ] Dismiss the widget and confirm you return to the SERP cleanly.

_Tapping a dax bubble option closes the widget and the address bar updates_

- [ ] Fresh install the app and complete onboarding.
- [ ] On the NTP with the "Try a search!" bubble visible, tap the address bar to open the widget.
- [ ] Tap one of the bubble's suggestion options.
- [ ] **Verify:** The widget closes immediately and the address bar shows the submitted query/URL while the SERP loads.

_Contextual onboarding bubbles in the browser still work_

- [ ] Continue the onboarding.
- [ ] Contextual bubbles (`trackers blocked`, `fire button`, `visit a site` if applicable) all appear and dismiss normally.
- [ ] Finish the onboarding sequence and reach the `DaxEnd` bubble on the NTP — it appears as expected.

### UI changes

https://github.com/user-attachments/assets/c4285a84-3acd-4ccd-a73f-fdc1f20a7c0d

---

> [!NOTE]
> **Low Risk**
> Localized UI/onboarding behavior in BrowserTabFragment with no security, auth, or data-handling changes.
> 
> **Overview**
> Fixes two onboarding issues when the **native address-bar input** is enabled: NTP Dax bubbles reappearing under the widget after leaving the new tab page, and the widget staying open with stale text after tapping a bubble suggestion.
> 
> **`renderCtaViewState`** now clears the previous **`DaxBubbleCta`** view state whenever the CTA renderer runs (including when switching to browser mode), and skips **`showCta`** for NTP-only bubbles while the browser is visible—contextual **`OnboardingDaxDialogCta`** flows in **`browserLayout`** are unchanged.
> 
> Onboarding suggestion taps (intro bubble, in-context, and brand-design dialogs) call **`submitQuery`** instead of **`userEnteredQuery`**, so **`hideNativeInput`** runs before navigation and the omnibar stays in sync.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0caf18fc24313a8a71c039a03bc725462a59a0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -5915,15 +5915,24 @@ class BrowserTabFragment :
                 when {
                     viewState.cta != null -> {
                         hideNewTab()
-                        showCta(viewState.cta)
+                        // Bubble lives inside newTabLayout, which is shown outside this renderer.
+                        // Hiding the parent doesn't reset the bubble's visibility, so we clear it here.
+                        if (previousCta is DaxBubbleCta) hideDaxBubbleCta(previousCta)
+
+                        // Non-bubble CTAs paint outside newTabLayout, so they're never at risk of leaking.
+                        // If we're on the NTP, the bubbles paint normally.
+                        if (viewState.cta !is DaxBubbleCta || !viewState.isBrowserShowing) {
+                            showCta(viewState.cta)
+                        }
                     }
 
                     viewState.isBrowserShowing -> {
+                        if (previousCta is DaxBubbleCta) hideDaxBubbleCta(previousCta)
                         hideNewTab()
                     }
 
                     viewState.isOnboardingCompleteInNewTabPage && !viewState.isErrorShowing -> {
-                        hideDaxBubbleCta(previousCta as? DaxBubbleCta)
+                        if (previousCta is DaxBubbleCta) hideDaxBubbleCta(previousCta)
                         showNewTab()
                     }
                 }
@@ -5983,7 +5992,7 @@ class BrowserTabFragment :
                         onboardingExperimentEnabled = false,
                         configuration = configuration,
                     ) { option, index ->
-                        userEnteredQuery(option.link)
+                        submitQuery(option.link)
                         viewModel.onUserSelectedOnboardingDialogOption(configuration, index)
                     }
                 }
@@ -6064,7 +6073,7 @@ class BrowserTabFragment :
                 }
             val onSuggestedOptionsSelected: ((DaxDialogIntroOption) -> Unit)? =
                 if (configuration is OnboardingDaxDialogCta.DaxSiteSuggestionsCta) {
-                    { option: DaxDialogIntroOption -> userEnteredQuery(option.link) }
+                    { option: DaxDialogIntroOption -> submitQuery(option.link) }
                 } else {
                     null
                 }
@@ -6102,7 +6111,7 @@ class BrowserTabFragment :
                 // CTA-type-specific notifications (e.g. DaxTrackersBlocked) live on the
                 // subclass override, not on the fragment.
                 onTypingAnimationFinished = { viewModel.onOnboardingDaxTypingAnimationFinished() },
-                onSuggestedOptionClicked = { option: DaxDialogIntroOption -> userEnteredQuery(option.link) },
+                onSuggestedOptionClicked = { option: DaxDialogIntroOption -> submitQuery(option.link) },
                 onDismissCtaClicked = { viewModel.onUserClickCtaDismissButton(configuration) },
                 instantShow = instantShow,
             )
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
@@ -4187,6 +4187,10 @@ class BrowserTabFragment :
         brandDesignDialogScrollView.gone()
     }
 
+    private fun hideIfDaxBubble(cta: Cta?) {
+        if (cta is DaxBubbleCta) hideDaxBubbleCta(cta)
+    }
+
     @SuppressLint("AddDocumentStartJavaScriptUsage")
     private fun configureWebViewForBlobDownload(webView: DuckDuckGoWebView) {
         lifecycleScope.launch(dispatchers.main()) {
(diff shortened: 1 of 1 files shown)


YoussefKeyrouz:
* **#8692** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5915,15 +5915,24 @@ class BrowserTabFragment :
                 when {
                     viewState.cta != null -> {
                         hideNewTab()
-                        showCta(viewState.cta)
+                        // Bubble lives inside newTabLayout, which is shown outside this renderer.
+                        // Hiding the parent doesn't reset the bubble's visibility, so we clear it here.
+                        if (previousCta is DaxBubbleCta) hideDaxBubbleCta(previousCta)
```
malmstein: nit: this `if (previousCta is DaxBubbleCta) hideDaxBubbleCta(previousCta)` is now repeated 3 times in the same `when`. worth extracting to a one-liner `clearPreviousBubble(previousCta)` to keep the invariant in one place — easy to drift on the next branch added.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5983,7 +5992,7 @@ class BrowserTabFragment :
                         onboardingExperimentEnabled = false,
                         configuration = configuration,
                     ) { option, index ->
-                        userEnteredQuery(option.link)
+                        submitQuery(option.link)
```
malmstein, line 5999: what's the practical difference between `userEnteredQuery` and `submitQuery` here? swap makes sense for the native-input case (we don't want the input widget to pop on a curated onboarding link), but worth a sentence in the PR description about what each does and why. also — are these three sites the only places we needed to swap, or are there other onboarding flows still on `userEnteredQuery(option.link)` we missed?

malmstein (approved):
did not install or smoke-test the APK. the fix looks sensible — bubble CTAs leak when transitioning between NTP and browser surfaces with native input on, and the three pre-existing branches in the `when` weren't all clearing the prior bubble. couple of points worth chatting through: a duplicated bubble-clear pattern that begs for a one-liner helper, and the `userEnteredQuery` → `submitQuery` swap which is meaningful behaviour but isn't explained in the PR description. no tests — understandable for a visibility/state-machine bug in a fragment, but worth verifying the fix on a device before merging.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
YoussefKeyrouz: Not all branches should clear the bubble, “when” branches are not exhaustive so there might be scenarios where none of the branches match and we don’t want to clear it. Unless you mean extract it to a function to call instead of having 3 ifs. I can do that if you like. Before the change, the function was called with a null which perform useless operations I only added the “if” to mini-optimize.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
YoussefKeyrouz: submitQuery hides the native input then calls userEnteredQuery. We do want to hide/dismiss the native input when we submit a query. No other difference. 

All onboarding branches where we do want to hide the native input are handled. If you’re asking whether there are non-onboarding branches where this should be called, I’m not sure, I kept the scope to the onboarding process

As for description, I had this line: 

```
CTA option click fix: route the three dax option-click sites through the existing submitQuery instead of userEnteredQuery. submitQuery calls hideNativeInput before onUserSubmittedQuery.
```

let me know if unclear

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
malmstein: yeah I meant a helper — `clearPreviousBubble(previousCta)` (or similar) that owns the `is DaxBubbleCta` check internally. each branch still decides whether to call it, but the type guard lives in one place, so if a new bubble subtype lands we edit it once. happy if you want to do it.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
malmstein: thanks, that's the answer — and I missed your line in the description, sorry. scope to onboarding is fine; we can grep non-onboarding sites separately if a pattern emerges. all good here.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
YoussefKeyrouz: Added a helper function “hideIfDaxBubble” which I think is better named than “clearPreviousBubble” because we’re passing the bubble in the params

PR #8725, opened 2026-05-30
Fire Mode: Duck.ai and input widget theme overlay

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1215242220281929?focus=true

### Description

Adds Fire mode theme overlays for Duck.ai screens and widgets.

Figma link: https://www.figma.com/design/9S33EK9qIxZdN1FOIFHM5W/%F0%9F%94%A5-Fire-Mode-Tabs--iOS-Android-?node-id=1846-13998&m=dev

### Steps to test this PR

- [ ] Enable the Fire tabs FF in dev settings
- [ ] Make sure Search & Duck.ai option is selected in AI settings
- [ ] Make sure the Light mode is set on a test device
- [ ] In the Tab switcher, switch to Fire mode
- [ ] Go back to the browser
- [ ] Verify that the browser is using the light mode but the native input widget is using dark mode colors
- [ ] Verify the action buttons are using the Fire mode orange colors
- [ ] Switch to dark mode on the device
- [ ] Verify the dark mode theme is used everywhere

---

> [!NOTE]
> **Low Risk**
> Theming and layout attribute changes across browser and Duck.ai UI with no auth or data logic; regression risk is visual (Fire vs normal themes, light Fire card contrast).
> 
> **Overview**
> Introduces **`daxInputModeCardThemeOverlay`** so Duck.ai / native input **cards** can pick up Fire styling independently of the rest of the screen. In **light Fire mode**, light Fire sets **`ThemeOverlay.DuckDuckGo.InputModeCard.DarkFire`** on the card while the activity stays on the light Fire theme—matching the Figma split (light browser chrome, dark input card, orange accents).
> 
> Fire accents are centralized on new **`daxColorFireModeAccent*`** and **`daxColorOmnibarAccentContent`** theme attrs; Fire themes and omnibar overlays now reference those instead of hard-coded mandarin colors. Input cards and send/outline controls switch from blue primary button colors to **`daxColorOmnibarAccent`** / **`daxColorOmnibarAccentContent`**.
> 
> **`InputScreenActivity`** applies the Fire theme when **`BrowserMode.FIRE`**; bottom-bar **`InputScreenButtons`** are inflated with the card’s themed context via new **`InputModeWidget.cardContext`**. Omnibar tracker animations and privacy shield rendering force the dark variant in Fire mode; the tab switcher icon uses **`?attr/daxIconTabSwitcher`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9465d5bd2775956df65bceb036dbbd5943159e75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


0nko:
* **#9078** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9078?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9053** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9053?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9021** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9021?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9001** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9001?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8917** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8902** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8902?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8792** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8792?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8791** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8791?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8748** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8748?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8742** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8725** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8725?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8725?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8714** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

android-design-system/design-system/src/main/res/values/design-system-colors-rebrand.xml
```diff
@@ -54,6 +54,8 @@
     <color name="mandarin80">#671907</color>
     <color name="mandarin90">#47140B</color>
     <color name="mandarin100">#290E0A</color>
+    <color name="mandarin40_translucent">#33FF8D5C</color>
```
malmstein, line 57: we moved `omnibar_card_fire_light/dark` into `design-system-colors.xml` but the `mandarin*_translucent` pair landed here in the rebrand file — can we keep them together so the fire palette lives in one place?

malmstein (commented):
did not install or smoke-test the APK — static-only. nice follow-up to #8714: pulling the hardcoded `mandarin50/40` values into semantic `daxColorFireModeAccent*` attrs and adding the `daxInputModeCardThemeOverlay` "dark island" overlay for the Duck.ai and input-widget cards is clean and consistent with the omnibar overlay pattern. the `NativeInputOmnibarController` switch to resolving `daxColorWindow` from the card's themed `context` (instead of the outer `ctx`) is the right fix to make the overlay actually take. this also addresses the raw-hex-in-the-theming-file nit I raised on #8714.

no blockers. one question worth a reply before merge: the input-mode-switch card's focus stroke stays `daxColorAccentBlue` under the fire overlay — looks like it'd read blue in fire mode. the tab-button dimension flip and the palette-file split are minor.

since I haven't run it, I'd smoke-test the input widget and Duck.ai card in both light/dark fire mode before approving — especially the stroke colour.

browser/browser-ui/src/main/res/layout/view_new_toolbar_tab_switcher_button.xml
```diff
@@ -26,12 +26,12 @@
 
     <ImageView
         android:id="@+id/tabsImageView"
-        android:layout_width="@dimen/bottomNavIcon"
-        android:layout_height="@dimen/bottomNavIcon"
+        android:layout_width="@dimen/toolbarIcon"
```
malmstein, line 29: deleting the `:app` copy means this browser-ui layout now wins, and it also changes the ImageView from `bottomNavIcon` (24dp) to `toolbarIcon` (40dp). `scaleType="center"` keeps the drawable at its intrinsic size so it's mostly invisible, but is the dimension change intentional or just fallout from dropping the duplicate?

duckchat/duckchat-impl/src/main/res/layout/view_input_mode_switch_widget.xml
```diff
@@ -96,6 +96,7 @@
             android:layout_height="wrap_content"
             android:clipChildren="false"
             android:clipToPadding="false"
+            android:theme="?attr/daxInputModeCardThemeOverlay"
             app:cardElevation="0dp"
             app:strokeColor="?attr/daxColorAccentBlue"
```
malmstein: the card now gets `daxInputModeCardThemeOverlay`, but `strokeColor` is still pinned to `daxColorAccentBlue` and the DarkFire overlay only remaps `daxColorOmnibarAccent`, not `daxColorAccentBlue` — so the outline stays blue in fire mode while the rest of the card goes orange. is that intended, or should the stroke follow the fire accent too?

android-design-system/design-system/src/main/res/values/design-system-colors-rebrand.xml
0nko: the split is following the existing two-file convention rather than being accidental: `mandarin40/50_translucent` are mandarin-scale colours (the mandarin hue at 20% alpha), so they sit in the Mandarin block next to `mandarin40/50` in the rebrand palette file. `omnibar_card_fire_light/dark` aren't part of any named scale (they're bespoke surface greys), so they live in `design-system-colors.xml` alongside the other raw one-offs like `background_omnibar_shadow`. moving either group to join the other would pull it out of its conventional home — but if you'd prefer discoverability over the convention, i can add a `<!-- Fire mode -->` grouping comment or co-locate them. let me know.

browser/browser-ui/src/main/res/layout/view_new_toolbar_tab_switcher_button.xml
0nko: intentional — it's needed for the fire-mode icon rather than fallout from dropping the duplicate. `daxIconTabSwitcher` resolves to `ic_tab_fire_mode_24` in fire mode, which is 24×30dp (taller than the 24×24 `ic_tab_24`). with `scaleType="center"` the drawable renders at its intrinsic size, so at `bottomNavIcon` (24dp) the 30dp-tall fire icon clips ~3dp top and bottom; `toolbarIcon` (40dp) gives it room. the default and highlighted icons are 24×24, which is why the change looks like a no-op for them. the view is `match_parent` inside a `toolbarIcon`-sized button either way, so the tab count doesn't shift.

duckchat/duckchat-impl/src/main/res/layout/view_input_mode_switch_widget.xml
0nko: good catch — fixed. the stroke now uses `?attr/daxColorOmnibarAccent` instead of `daxColorAccentBlue`. in the base themes `daxColorOmnibarAccent` already resolves to `daxColorAccentBlue`, so non-fire modes are unchanged, and the `InputModeCard.DarkFire` overlay remaps `daxColorOmnibarAccent` to the fire accent — so the outline now follows the card into orange. i applied the same change to the two sibling cards that get the overlay in this PR (`fragment_contextual_duck_ai.xml` and `view_native_input_mode_switch_widget.xml`, the latter via its `:app` host) since they had the identical blue-pinned stroke.

malmstein:
The toggle in the Tab Manager doesn’t look right. We discussed this offline and @0nko  will fix it in a separate PR
Screenshot_20260611_141441

browser/browser-ui/src/main/res/layout/view_new_toolbar_tab_switcher_button.xml
malmstein, line 29: still leaving this from last round so it isn't lost: deleting the `:app` copy of this layout means this browser-ui one wins, and it also changes the ImageView from `bottomNavIcon` (24dp) to `toolbarIcon` (40dp). `scaleType="center"` keeps the drawable at its intrinsic size so it's mostly invisible, but is the dimension change intentional or just fallout from dropping the duplicate?

malmstein (approved):
re-review after the five new commits, approving. built the branch locally and installed on a Pixel 7 — it launches healthy with no crash; the device dropped off adb before I could drive into Fire mode, so I didn't get to eyeball the fire theming on-device (stacked PR, no CI APK, fire surfaces need an internal build with the flag on anyway). both of my substantive comments are addressed: the input-card focus strokes now follow `daxColorOmnibarAccent` so they go orange in Fire mode, and the fire palette colors moved into their semantically correct files. the new `daxColorFireModeAccent*` / `daxColorOmnibarAccentContent` attrs are a clean way to handle the accent and the icon-contrast-in-dark-fire case, and the InputModeCard.DarkFire overlay overriding the content colour back to white is exactly right for the dark-card-in-light-app case.

only leftover is the minor one inlined: the tab-switcher button's 24dp→40dp dimension flip from deleting the `:app` layout — muted by `scaleType="center"`, just worth confirming it's intentional. nothing blocking.

browser/browser-ui/src/main/res/layout/view_new_toolbar_tab_switcher_button.xml
0nko: intentional, not fallout. `daxIconTabSwitcher` resolves to `ic_tab_fire_mode_24` in fire mode, which is 24×30dp — taller than the 24×24 `ic_tab_24`/`ic_tab_24_highlighted`. with `scaleType="center"` the drawable draws at its intrinsic size, so a 24dp (`bottomNavIcon`) container clips the fire icon ~3dp top and bottom; `toolbarIcon` (40dp) gives it room. the default/highlighted icons are 24×24, which is why it reads as a no-op for them. (re-posting here since the threaded reply from last round landed on the now-outdated thread.)

0nko:
The toggle and the animation background color have both been fixed.

PR #8748, opened 2026-06-02
Fire Mode: Empty state pages

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1215327489156969?focus=true

Figma: [Tab switcher](https://www.figma.com/design/9S33EK9qIxZdN1FOIFHM5W/%F0%9F%94%A5-Fire-Mode-Tabs--iOS-Android-?node-id=1851-24549&m=dev)
Figma: [Browser](https://www.figma.com/design/9S33EK9qIxZdN1FOIFHM5W/%F0%9F%94%A5-Fire-Mode-Tabs--iOS-Android-?node-id=1851-26427&m=dev)

### Description

This PR adds Fire mode empty state for Fire tabs in the browser and the Tab switcher.

### Steps to test this PR

- [x] Enable the Fire tabs FF in the dev settings (and restart the app)
- [x] Go to the Tab switcher
- [x] Switch to Fire mode
- [x] Verify the new empty state screen is displayed (compare with Figma design)
- [x] Tap on the New Fire tab button
- [x] Verify a new Fire tab is opened
- [x] Verify the new empty tab screen is displayed instead of Dax (compare with Figma design)
- [x] Go back to the Tab switcher
- [x] Close all Fire tabs using the X button
- [x] After the last Fire tab is closed, verify that the app automatically changes modes to Regular

| Tab switch | Browser |
|--------|--------|
| image | image |

---

> [!NOTE]
> **Medium Risk**
> Changes tab close and mode-switch behavior for Fire mode and tab auto-creation edge cases, but scope is limited to Fire mode with new unit tests covering the main paths.
> 
> **Overview**
> Adds **Fire mode empty states** on the tab switcher and Fire new-tab page, replacing the normal Dax/NTP chrome when Fire mode is active.
> 
> In the **tab switcher**, when Fire mode has zero tabs, a dedicated empty state (feature bullets, explainer, **New Fire Tab**) is shown instead of the tab grid; **New Fire Tab** opens a tab via the existing new-tab flow. **Mode-switch fade** skips animating a hidden recycler so toggling modes from the empty state still recreates cleanly.
> 
> On the **Fire new tab page**, Fire mode shows a simpler empty state and hides logo, messages, favourites, and related NTP sections.
> 
> **Closing the last or all Fire tabs** now emits **`SwitchToRegularMode`** (activity recreates in Regular mode) instead of the close-all undo snackbar path used for regular tabs.
> 
> **`DefaultTabManager`** stops auto-adding a default tab when the Fire tab list is empty but global browser mode is already Regular, avoiding a stray tab during mode transitions. Tab switcher loading uses a **`TabItems`** loaded vs not-initialized distinction so the Fire empty state only appears after tabs have loaded, not during initial load.
> 
> New layouts, strings, and design-system icons support the UI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e04a96c25644519930e4f62ac465cf0e3e7d0145. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabPageView.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabPageView.kt
@@ -272,6 +272,8 @@ class NewTabPageView @JvmOverloads constructor(
             return
         }
         binding.fireTabEmptyState.root.gone()
+        binding.indonesiaNewTabSectionView.show()
+        binding.appTrackingProtectionStateView.show()
 
         if (viewState.shouldShowLogo) {
             homeBackgroundLogo.showLogo()
--- a/app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabPageView.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabPageView.kt
@@ -262,6 +262,17 @@ class NewTabPageView @JvmOverloads constructor(
 
         onHasContent?.invoke(viewState.hasContent)
 
+        if (viewState.showFireTabEmptyState) {
+            binding.fireTabEmptyState.root.show()
+            binding.indonesiaNewTabSectionView.gone()
+            binding.appTrackingProtectionStateView.gone()
+            binding.ddgLogo.gone()
+            binding.messageCta.gone()
+            binding.focusedFavourites.gone()
+            return
+        }
+        binding.fireTabEmptyState.root.gone()
+
         if (viewState.shouldShowLogo) {
             homeBackgroundLogo.showLogo()
         } else {
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
@@ -505,6 +505,15 @@ class TabSwitcherActivity :
         if (fadingOutForRecreate) return
         fadingOutForRecreate = true
 
+        // In the Fire tabs empty state the recycler is hidden and empty, so there is nothing to
+        // fade out. Animate nothing and just switch mode and recreate immediately; the recreated
+        // activity still fades the new mode's tabs in via the saved fadingInAfterRecreate flag.
+        if (tabsRecycler.visibility != View.VISIBLE) {
+            viewModel.onBrowserModeToggled(newMode)
+            recreate()
+            return
+        }
+
         tabsRecycler.animate()
             .alpha(0f)
             .setDuration(MODE_SWITCH_FADE_OUT_MS)
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
@@ -304,6 +304,7 @@ class TabSwitcherActivity :
         setupToolbar(toolbar)
         configureRecycler()
         configureNavigationBar()
+        configureFireTabsEmptyState()
 
         if (edgeToEdgeEnabled) {
             configureEdgeToEdgeInsets()
@@ -364,6 +365,12 @@ class TabSwitcherActivity :
             binding.navigationBar.show()
         } else {
             binding.navigationBar.gone()
+        }
+    }
+
+    private fun configureFireTabsEmptyState() {
+        binding.fireTabsEmptyState.newFireTabButton.setOnClickListener {
+            onNewTabRequested(fromOverflowMenu = false)
         }
     }
 
@@ -562,30 +569,37 @@ class TabSwitcherActivity :
     private fun configureObservers() {
         lifecycleScope.launch {
             viewModel.viewState.flowWithLifecycle(lifecycle).collectLatest {
-                it.layoutType?.let(::updateLayoutType)
+                if (it.showFireTabsEmptyState && !fadingOutForRecreate) {
+                    binding.fireTabsEmptyState.root.show()
+                    tabsRecycler.gone()
+                } else {
+                    binding.fireTabsEmptyState.root.gone()
 
-                tabsRecycler.invalidateItemDecorations()
+                    it.layoutType?.let(::updateLayoutType)
 
-                val shouldTryScroll = it.tabs.isNotEmpty() &&
-                    (firstTimeLoadingTabsList || fadingInAfterRecreate)
+                    tabsRecycler.invalidateItemDecorations()
 
-                tabsAdapter.updateData(it.tabSwitcherItems) {
-                    pendingToggleScrollAnchor?.let { anchor -> scrollPositionToTop(anchor) }
+                    val shouldTryScroll = it.tabs.isNotEmpty() &&
+                        (firstTimeLoadingTabsList || fadingInAfterRecreate)
 
-                    val scrolled = shouldTryScroll && scrollToActiveTab(it.tabSwitcherItems)
-                    if (scrolled) firstTimeLoadingTabsList = false
+                    tabsAdapter.updateData(it.tabSwitcherItems) {
+                        pendingToggleScrollAnchor?.let { anchor -> scrollPositionToTop(anchor) }
 
-                    if (fadingInAfterRecreate && !fadeInAnimationStarted && it.tabs.isNotEmpty()) {
-                        fadeInAnimationStarted = true
-                        tabsRecycler.show()
-                        tabsRecycler.animate()
-                            .alpha(1f)
-                            .setDuration(MODE_SWITCH_FADE_IN_MS)
-                            .withEndAction {
-                                fadingInAfterRecreate = false
-                                tabsRecycler.itemAnimator = DefaultItemAnimator()
-                            }
-                            .start()
+                        val scrolled = shouldTryScroll && scrollToActiveTab(it.tabSwitcherItems)
+                        if (scrolled) firstTimeLoadingTabsList = false
+
+                        if (fadingInAfterRecreate && !fadeInAnimationStarted && it.tabs.isNotEmpty()) {
+                            fadeInAnimationStarted = true
+                            tabsRecycler.show()
+                            tabsRecycler.animate()
+                                .alpha(1f)
+                                .setDuration(MODE_SWITCH_FADE_IN_MS)
+                                .withEndAction {
+                                    fadingInAfterRecreate = false
+                                    tabsRecycler.itemAnimator = DefaultItemAnimator()
+                                }
+                                .start()
+                        }
                     }
                 }
 
@@ -787,6 +801,7 @@ class TabSwitcherActivity :
             DismissAnimatedTileDismissalDialog -> tabSwitcherAnimationTileRemovalDialog!!.dismiss()
             Command.ShowFireBottomSheet -> onFireButtonClicked()
             Command.DismissSnackbar -> lastSnackbar?.dismiss()
+            Command.SwitchToRegularMode -> fadeOutTabsThenRecreate(BrowserMode.REGULAR)
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
@@ -73,11 +73,11 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.combine
-import kotlinx.coroutines.flow.conflate
 import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flatMapLatest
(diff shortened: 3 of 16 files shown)


0nko:
* **#9078** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9078?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9053** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9053?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9021** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9021?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9001** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9001?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8917** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8902** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8902?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8792** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8792?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8791** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8791?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8748** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8748?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8748?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8742** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8725** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8714** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
```diff
@@ -524,29 +531,38 @@ class TabSwitcherActivity :
     private fun configureObservers() {
         lifecycleScope.launch {
             viewModel.viewState.flowWithLifecycle(lifecycle).collectLatest {
-                // Apply layout type first so the LayoutManager is in place before the adapter
```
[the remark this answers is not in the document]
0nko: Partially valid — fixed in b5b34190b1.

The literal "end action may never run" premise is dubious: a GONE-but-attached view's `ViewPropertyAnimator` still drives its `ValueAnimator` off the Choreographer, so `withEndAction` does fire and the toggle wouldn't actually get stuck. But the underlying point is real — in the Fire empty state `tabsRecycler` is `gone()` and empty, so the fade animates nothing, the visible empty-state view isn't animated, and we just eat a pointless 180ms delay before `recreate()`.

Fix: in `fadeOutTabsThenRecreate`, when `tabsRecycler.visibility != VISIBLE`, switch mode and `recreate()` immediately instead of fading. `fadingOutForRecreate` stays `true` so the recreated activity still fades the new mode's tabs in. Also covers the `Command.SwitchToRegularMode` path, which routes through the same method.

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
```diff
@@ -475,29 +482,38 @@ class TabSwitcherActivity :
...
-                                tabsRecycler.itemAnimator = DefaultItemAnimator()
-                            }
-                            .start()
+                // Skip the empty state while a mode-switch recreate is in flight so it doesn't flash
+                // over the fading-out grid.
+                if (it.showFireTabsEmptyState && !fadingOutForRecreate) {
+                    binding.fireTabsEmptyState.root.show()
+                    tabsRecycler.gone()
+                } else {
+                    binding.fireTabsEmptyState.root.gone()
```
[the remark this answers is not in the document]
0nko: False positive — no change needed.

The `else` branch doesn't only re-show `tabsRecycler` in the `fadingInAfterRecreate` path; it runs `it.layoutType?.let(::updateLayoutType)`, and `updateLayoutType` always ends with `tabsRecycler.show()` (both the early-return branch and the normal branch). So leaving the empty state through the normal path already restores the recycler.

The only theoretical gap is `layoutType == null`, but at that point there are no tabs to render anyway, and the one reachable exit from the empty state (the "new fire tab" button) calls `removeObservers()` first, so the observer never processes a `showFireTabsEmptyState` true->false flip. No blank-screen-with-content scenario exists.

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
```diff
@@ -449,7 +445,11 @@ class TabSwitcherViewModel @Inject constructor(
         swipeGestureUsed: Boolean = false,
     ) {
         viewModelScope.launch {
-            if (tabs.size == 1) {
+            val isLastTab = tabs.size == 1
```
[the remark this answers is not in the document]
0nko: Valid — fixed in 0d9f43930d.

The single-tab close path drops to Regular when the last Fire tab is closed, but `onCloseTabsConfirmed` (the "Close all tabs" menu, plus any selection-close that empties the list) still emitted `CloseAndShowUndoMessage` in Fire mode, leaving the user in Fire mode with no tabs — contrary to the stated intent.

Fix: when `tabs.size == tabIds.size && fireModeAvailable && currentMode == FIRE`, mark all tabs deletable and emit `SwitchToRegularMode` instead of the undo flow, mirroring the single-tab Fire path (which also forgoes undo). Non-Fire close-all is unchanged. Added a unit test (`when closing all fire tabs then switches to regular mode instead of showing undo`).

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
```diff
@@ -687,9 +695,12 @@ class TabSwitcherViewModel @Inject constructor(
         val regularTabCount: Int? = null,
         val isBrowserModeToggleVisible: Boolean = false,
     ) {
+        val tabSwitcherItems: List<TabSwitcherItem> = tabItems.itemsOrEmpty
         val tabs: List<Tab> = tabSwitcherItems.filterIsInstance<Tab>()
         val numSelectedTabs: Int = (mode as? Selection)?.selectedTabs?.size ?: 0
 
+        val showFireTabsEmptyState: Boolean = browserMode == BrowserMode.FIRE && tabItems is TabItems.Loaded && tabs.isEmpty()
```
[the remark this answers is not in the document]
0nko: By design — no change needed.

The `NotInitialized` → `Loaded` split is intentional precisely so the empty state appears only once Fire tabs are *known* empty, not while the debounced flow is still loading. Gating `showFireTabsEmptyState` on `Loaded` is the point: the alternative (treating the initial state as an empty list) would flash the designed empty state during the ~100ms load window and then replace it with tabs whenever Fire tabs do exist.

Showing the (empty) recycler for that brief load window is the same behaviour as any other mode during load, and is the deliberate tradeoff over flashing the empty state. No regression vs. the prior `onStart { emit(emptyList()) }`, which rendered an empty list during the same window.

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherActivity.kt
```diff
@@ -562,30 +578,37 @@ class TabSwitcherActivity :
...
+                            fadeInAnimationStarted = true
+                            tabsRecycler.show()
+                            tabsRecycler.animate()
+                                .alpha(1f)
+                                .setDuration(MODE_SWITCH_FADE_IN_MS)
+                                .withEndAction {
+                                    fadingInAfterRecreate = false
+                                    tabsRecycler.itemAnimator = DefaultItemAnimator()
+                                }
+                                .start()
```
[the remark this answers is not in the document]
0nko: No visible impact — no change needed.

The recycler's alpha is only restored in the fade-in block, gated on `tabs.isNotEmpty()` — but tab content is *only* ever rendered through the same `updateData` callback that runs that fade-in. So content is never displayed at alpha 0: the recycler is either (empty + alpha 0, nothing to see) or (has tabs → fade-in → alpha 1). The Fire empty-state view is a separate sibling shown at full opacity, unaffected by recycler alpha.

When a recreate lands on a screen with no tabs to fade in, `fadingInAfterRecreate`/the suppressed `itemAnimator` do linger, but they self-heal: the next tab addition is non-empty, so the fade-in block runs (alpha → 1, `fadingInAfterRecreate` cleared, `DefaultItemAnimator` restored). Regular mode always retains an NTP tab, so there's no genuinely-stuck blank screen. At most this is a harmless lingering flag; adding a reset would touch the working fade choreography for no observable benefit.

app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabPageView.kt
```diff
@@ -262,6 +262,17 @@ class NewTabPageView @JvmOverloads constructor(
 
         onHasContent?.invoke(viewState.hasContent)
 
+        if (viewState.showFireTabEmptyState) {
+            binding.fireTabEmptyState.root.show()
+            binding.indonesiaNewTabSectionView.gone()
```
LukasPaczos, line 267: This is hidden but there's no other code path that would show this view again. Can that cause any issues?

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
```diff
@@ -425,9 +421,16 @@ class TabSwitcherViewModel @Inject constructor(
                 pixel.fire(AppPixelName.TAB_MANAGER_MENU_CLOSE_ALL_TABS_CONFIRMED)
                 pixel.fire(AppPixelName.TAB_MANAGER_MENU_CLOSE_ALL_TABS_CONFIRMED_DAILY, type = Daily())
 
-                // mark tabs as deletable, the undo snackbar will be displayed when the tab switcher is closed
                 tabRepository.markDeletable(tabIds)
-                command.value = Command.CloseAndShowUndoMessage(tabIds)
+
+                if (fireModeAvailable && currentMode.value == BrowserMode.FIRE) {
```
LukasPaczos, line 426: If we're already in `currentMode.value == BrowserMode.FIRE`, checking `fireModeAvailable` seems unnecessary, or even dangerous. If the user is in fire mode, but in parallel the `fireModeAvailable` changes to `false`, they might get stuck.

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
```diff
@@ -449,7 +452,11 @@ class TabSwitcherViewModel @Inject constructor(
         swipeGestureUsed: Boolean = false,
     ) {
         viewModelScope.launch {
-            if (tabs.size == 1) {
+            val isLastTab = tabs.size == 1
+            if (isLastTab && fireModeAvailable && currentMode.value == BrowserMode.FIRE) {
```
LukasPaczos, line 456: Same as above ref. fire mode availability.

LukasPaczos (commented):
If i repeatably click the mode switcher, it eventually freezes. Not sure if this is a regression from this PR, or whether it's a preexisting issue. If the latter, are we tracking it already? If not, could you create a follow up?

https://github.com/user-attachments/assets/11b8f775-3c18-470c-87de-21ac9be145dd

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
0nko: Fire mode availability is checked and cached on app startup so it can never change mid-run and the default mode on startup is always regular mode. I guess the `fireModeAvailable` is redundant here so I can remove it if you want but I don't think it changes anything.

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
0nko: See my answer above.

app/src/main/java/com/duckduckgo/app/tabs/ui/TabSwitcherViewModel.kt
LukasPaczos: I see :+1: This just looked like a code smell to me - since it's behind the interface, the view model shouldn't need to know that it's evaluated only once. Either way, I'll let you make the call on next steps.

app/src/main/java/com/duckduckgo/app/browser/newtab/NewTabPageView.kt
0nko: Yeah, good catch but it doesn't cause a problem today because `showFireTabEmptyState` is derived from `isFireMode`, which is fixed when the ViewModel is constructed and never changes, and the NTP view is fully recreated on mode switch. But I've made the hide reversible by showing both views in the non-fire path.

0nko:
Thanks @LukasPaczos!

> If i repeatably click the mode switcher, it eventually freezes. Not sure if this is a regression from this PR, or whether it's a preexisting issue. If the latter, are we tracking it already? If not, could you create a follow up?

The only thing I've experienced before was an occasional lag when changing the modes. I've never seen a complete freeze like you did here. This is not related to this PR so I'll follow up with it in another task.

LukasPaczos (approved):

PR #8758, opened 2026-06-03
Fix Duck.ai tab logo flash and vertical shift

aibrahim-:
Task/Issue URL:  https://app.asana.com/1/137249556945/project/1212608036467427/task/1215279708895831?focus=true

### Description
Two issues fixed

Tab switch blinking: showed the chat-suggestions overlay synchronously to cover the morphing logo during the async fetch, then cleared it ~200ms later when an empty input returned no content, flashing the list and the logo behind it. Now only cover when there is input text (content is then guaranteed), via shouldShowChatSuggestionsCoverOnSelect.

Logo vertical shift between tabs: isLogoOnlyContent treated the NewTabReturnHatchView as present whenever it was VISIBLE, but that container is always VISIBLE and collapses to zero height when empty, so the logo was pinned below the taller Duck.ai widget. Now detect a real hatch via height > 0 (extracted as isLogoOnly), keeping the logo at a fixed vertical position.

### Steps to test this PR
- enable unified input toggle
- switch between tabs and observe smooth animation
- try the same but with existing recent chats in duck.ai

### UI changes
output


joshliebe (approved):
It still moves around when the hatch is showing but looks much better for the default case

https://github.com/user-attachments/assets/d6e48767-b5e1-4f3d-a3d2-2face66a76fa

PR #8762, opened 2026-06-03
Add AI config CI check (rules, skills, refs, symlinks)

catalinradoiu:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211724162604201/task/1214927988557607

### Description

Adds an **AI config CI check** that fails the build when the repository's AI configuration drifts. The core is a pure-Kotlin `AiConfigChecker` in `buildSrc` performing four validations:

- **CLAUDE.md is the Claude entry point** — Cursor auto-loads `.cursor/rules/*.mdc` natively, but Claude can't read the `.mdc` extension, so the rules only reach Claude through `CLAUDE.md`. The check enforces that `CLAUDE.md` (a) imports `AGENTS.md` via `@AGENTS.md` so it inherits the shared base config, and (b) indexes every `.cursor/rules/*.mdc` in its "Detailed Rules" table — bidirectionally (every rule has a row, every indexed path exists). `@file` includes are resolved first, so index rows may live in an imported file.
- **Dangling references** — file paths and Gradle module references in `AGENTS.md` + the rule files resolve. Fenced code blocks, external/anchor/placeholder links, non-repo-rooted paths, and file-relative markdown links are handled to avoid false positives.
- **Rule symlink parity** — every `.cursor/rules/*.mdc` has a correct `.claude/rules/<name>.md` symlink, and every `.claude/rules/` entry is a symlink (no regular files).
- **Skill symlink parity** — skills are authored once under `.claude/skills/<name>/SKILL.md` and mirrored to `.cursor/skills/<name>/SKILL.md` as symlinks (Cursor 2.4+ reads `.cursor/skills` natively). The check requires a correct mirror symlink for every source skill and rejects orphaned mirrors or regular files. This absorbs and replaces `.github/workflows/check-rules-symlinks.yml` (deleted).

Enforcement:
- **CI:** a dedicated `.github/workflows/ai-config-check.yml` runs `./gradlew aiConfigCheck` (+ the checker's unit tests) on every PR.
- **Local:** wired into `jvm_checks` (`./gradlew jvm_checks`).

This PR also fixes the existing drift so the check ships green:
- Converts `CLAUDE.md` from a bare symlink to `AGENTS.md` into a real file that imports `AGENTS.md` (`@AGENTS.md`) and hosts the **Detailed Rules** table (moved out of `AGENTS.md`), now also indexing `pixels.mdc` and `pixel-definitions.mdc`.
- Adds `.cursorignore` so Cursor ignores `CLAUDE.md` and reads `AGENTS.md` directly — avoiding loading the shared base config twice.
- Mirrors the project skills into `.cursor/skills` via symlinks.
- Refreshes `dependency-updates.mdc` to stop hardcoding the Kotlin version (points to `version.kotlin` in `versions.properties` instead) and updates the Kotlin-incompatible-library list (e.g. `io.coil-kt.coil3` 3.4.0 requires Kotlin 2.3.x).

### Steps to test this PR

_AI config check_
- [ ] Run `./gradlew aiConfigCheck` → passes with `AI config check passed: 0 violations.`
- [ ] Run `./gradlew :buildSrc:test --tests "com.duckduckgo.gradle.AiConfigCheckerTest"` → 30 tests pass.
- [ ] Add a `.cursor/rules/foo.mdc` without a row in the `CLAUDE.md` "Detailed Rules" table, run `./gradlew aiConfigCheck` → fails reporting the unindexed rule.
- [ ] Remove the `@AGENTS.md` import from `CLAUDE.md`, run `./gradlew aiConfigCheck` → fails reporting the missing import.
- [ ] Replace a `.claude/rules/*.md` symlink with a regular file (or delete it), run `./gradlew aiConfigCheck` → fails reporting the rule symlink violation.
- [ ] Delete a `.cursor/skills/<name>/SKILL.md` mirror symlink, run `./gradlew aiConfigCheck` → fails reporting the skill symlink violation.

### UI changes
| Before  | After |
| ------ | ----- |
No UI changes | No UI changes |

---

> [!NOTE]
> **Low Risk**
> Changes are limited to AI documentation, build verification, and CI; no app runtime or production code paths are affected.
> 
> **Overview**
> Adds a **`buildSrc` `AiConfigChecker`** and root **`aiConfigCheck`** Gradle task that fails when AI repo config drifts. It validates **`CLAUDE.md`** imports **`AGENTS.md`** and bidirectionally indexes every **`.cursor/rules/*.mdc`**, scans **`AGENTS.md`** and rule files for dangling **paths** and **Gradle module** refs (with guards for fences, URLs, and non-repo paths), and enforces **`.claude/rules` ↔ `.cursor/rules`** plus **`.cursor/skills` ↔ `.claude/skills`** symlink parity.
> 
> **Enforcement:** new **`.github/workflows/ai-config-check.yml`** runs `./gradlew aiConfigCheck` and **`AiConfigCheckerTest`** on PRs; **`app` `jvm_checks`** depends on the same task. The standalone **`check-rules-symlinks.yml`** workflow is **removed** in favor of this check.
> 
> **Docs/layout:** the **Detailed Rules** table moves from **`AGENTS.md`** to **`CLAUDE.md`** (with **`@AGENTS.md`**); **`.cursor/skills/*/SKILL.md`** symlinks point at **`.claude/skills`**; **`.cursorignore`** ignores **`CLAUDE.md`**. **`dependency-updates.mdc`** is updated for current Kotlin metadata guidance (e.g. Coil 3.4 / 2.3.x).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c1aeb942453e2f71b52648a94a260b167ecfa80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


catalinradoiu:
* **#8762** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8762?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8762?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

buildSrc/src/main/java/com/duckduckgo/gradle/AiConfigChecker.kt
```diff
@@ -0,0 +1,213 @@
...
+    private fun resolveModule(ref: String, modules: Set<String>): Boolean {
+        val name = ref.trimStart(':').substringAfterLast(':')
+        return name in modules
+    }
+
+    private fun discoverModules(): Set<String> {
+        return repoRoot.walkTopDown()
+            .maxDepth(2)
+            .onEnter { it.name != "build" && it.name != ".git" && it.name != "node_modules" }
+            .filter { it.isDirectory && File(it, "build.gradle").exists() }
```
malmstein: `discoverModules()` only picks up dirs with a `build.gradle`, so a module defined with `build.gradle.kts` wouldn't be discovered and any `:module` ref to it would get flagged as a dangling reference. we're all-Groovy today so it's latent, but a `|| File(it, "build.gradle.kts").exists()` would future-proof it cheaply.

.cursor/rules/dependency-updates.mdc
```diff
@@ -33,36 +33,37 @@ It is required by the refreshVersions plugin at build time. If missing, the buil
 
 This is the most important rule when evaluating a library update.
 
-**The project currently uses Kotlin 1.9.24.** Many libraries have silently started publishing
-releases compiled with Kotlin 2.x. These cause a build failure at KSP/kapt time:
+**The project currently uses Kotlin 2.2.21** (`version.kotlin` in `versions.properties`;
```
malmstein: we're restating the exact Kotlin version (2.2.21) in prose again here, which is the kind of fact that goes stale — and aiConfigCheck validates structure/links/symlinks but not version freshness, so nothing actually guards this number. since you've added the `version.kotlin` pointer right beside it, could we lean on the pointer and drop the literal? heads up too: this section overlaps my in-flight #8811, happy to reconcile so we don't both edit it.

malmstein (approved):
really nice work — the checker is clean, pure-Kotlin, and the test coverage is excellent (both directions for rules and skills, plus all the false-positive sources: fenced blocks, urls/anchors/placeholders, xml namespaces, file-relative links, abbreviated `...` paths). folding the old check-rules-symlinks.yml shell logic into Kotlin and deleting it is a good cleanup.

one conceptual note, not blocking: this guards structural drift (unindexed rules/skills, dead links, broken symlinks) but not factual drift — whether a stated version or convention still matches the code. that's complementary to the guard/agentic approach in my #8809/#8811, and worth being explicit that the two cover different halves of the problem rather than overlapping.

two housekeeping things: the description still says it "corrects stale facts in AGENTS.md (Gradle 7.6 → 8.14.4, Kotlin 1.9.24 → 2.2.21)", but develop's AGENTS.md was already trimmed of those, so the diff only adds the Skills/pixels index there — the Kotlin correction actually lives in dependency-updates.mdc now; worth updating the description. and since this lands on top of that trim and overlaps #8811 on dependency-updates.mdc, let's coordinate the merge order.

AGENTS.md
```diff
@@ -12,6 +12,19 @@ The following rule files contain detailed guidance for specific topics. Read the
...
+## Skills
+
+Claude Code skills (model-invoked capabilities, one directory each under `.claude/skills/` with a `SKILL.md`). These are Claude-Code-only; Cursor has no equivalent.
+
+| Skill | Purpose |
+|---|---|
+| `.claude/skills/check-translations-pr/` | Verify a translations PR provides full language coverage |
+| `.claude/skills/review-public-api/` | Review a DuckDuckGo Android public API proposal |
+| `.claude/skills/run-maintenance-task/` | Execute a task from the Android Agentic Maintenance Backlog |
+| `.claude/skills/scope-maintenance-task/` | Scope a maintenance idea into a backlog task |
```
mikescamell: ❓ Why did we decide to list all the skills available in the AGENTS.md file? I'm not sure of the benefit seeing as the LLM should be able to discover them itself and this bloats context.

malmstein: Shouldn’t be there, has no benefit to either Claude or any other LLMs. I didn’t see it, thanks for spotting @mikescamell

.cursor/rules/dependency-updates.mdc
catalinradoiu: fixed

buildSrc/src/main/java/com/duckduckgo/gradle/AiConfigChecker.kt
catalinradoiu: fixed

AGENTS.md
catalinradoiu: I removed them

PR #8775, opened 2026-06-04
Fix empty config breaking C-S-S

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202552961248957/task/1215422278523665?focus=true

### Description
C-S-S retrieves configuration through plugins and then concatenates its contents separated by a comma. The code adds the separator based on the accumulator already being non-empty, not on the current plugin's output, sp if one plugin returns an empty config, we'll still add a comma, breaking the config: 

```
if (config.isNotEmpty()) { config += "," }   // accumulator, not the plugin's output
config += plugin.config()  // if this is empty, nothing follows the comma
```

So an empty plugin leaves a dangling/trailing comma, and the next non-empty plugin adds its own separator → you end up with a double comma (,,), which is the actual malformed token that breaks the JSON.

This is not an issue nowadays, but it's purely a coincidence: `AccessibilityForcedZoomContentScopeConfigPlugin` returns an empty config, but since it's the first one provided (alphabetical sort), the overall config is still empty when it runs, so we don't append a comma. However, the moment we have a plugin that sorts before `AccessibilityForcedZoomContentScopeConfigPlugin`, it breaks the format, and therefore C-S-S stops working.

### Steps to test this PR

_Feature 1_
- [ ] Fresh install, open youtube.com, search for any video, check Duck Player overlay is shown

### UI changes
n/a

---

> [!NOTE]
> **Low Risk**
> Small, localized change to string assembly for injected script config; behavior for all-non-empty plugins is unchanged and tests cover the edge cases.
> 
> **Overview**
> Fixes **Content Scope Scripts (C-S-S)** JSON assembly in `getPluginParameters()` so commas are only added when a plugin actually returns non-empty **config** or **preferences**, instead of inserting a separator whenever the accumulator was already non-empty.
> 
> That prevents **dangling commas** and `,,` in the `features` object (and the same pattern for user preferences), which could break parsing and disable C-S-S when a plugin returns an empty string but is not first in plugin order.
> 
> Adds unit tests for empty config in the middle or at the start of the plugin list and for empty preferences between non-empty ones.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc7ec763617e570d88d4aaa1fc3920c59111904b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/RealContentScopeScripts.kt
+++ b/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/RealContentScopeScripts.kt
@@ -126,16 +126,22 @@ class RealContentScopeScripts @Inject constructor(
         var preferences = ""
         val plugins = pluginPoint.getPlugins()
         plugins.forEach { plugin ->
-            if (config.isNotEmpty()) {
-                config += ","
+            plugin.config().let { pluginConfig ->
+                if (pluginConfig.isNotEmpty()) {
+                    if (config.isNotEmpty()) {
+                        config += ","
+                    }
+                    config += pluginConfig
+                }
             }
-            config += plugin.config()
 
             plugin.preferences()?.let { pluginPreferences ->
-                if (preferences.isNotEmpty()) {
-                    preferences += ","
+                if (pluginPreferences.isNotEmpty()) {
+                    if (preferences.isNotEmpty()) {
+                        preferences += ","
+                    }
+                    preferences += pluginPreferences
                 }
-                preferences += pluginPreferences
             }
         }
         return PluginParameters(config, preferences)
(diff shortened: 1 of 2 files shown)


CrisBarreiro:
* **#8775** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8775?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8775?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

aibrahim- (approved):
LGTM, I think it might be worth adding some documentation in `ContentScopeConfigPlugin` that the config string should have a certain format, e.g. a comma in that string would break and hence not allowed.

CrisBarreiro:
> LGTM, I think it might be worth adding some documentation in `ContentScopeConfigPlugin` that the config string should have a certain format, e.g. a comma in that string would break and hence not allowed.

It can have a comma, it just needs to be a valid JSON, which is already reflected in JavaDocs `@return a [String] containing the JSON config`. At that point, we should probably make the function return a JSONObject and not a String, though

PR #8806, opened 2026-06-08
Stop the local build-cache from invalidating the configuration cache

aibrahim-:
Task/Issue URL:  https://app.asana.com/1/137249556945/project/1212608036467427/task/1215492528760456?focus=true

### Description
The local Gradle build cache lives at ./build-cache, but settings.gradle's module auto-discovery scanned that directory during settings evaluation. Its contents change on every build, so Gradle fingerprinted the directory listing as a configuration-cache input and discarded the configuration cache on every build (Develocity miss reason: "directory 'build-cache' has changed").

Exclude 'build-cache' from the auto-discovery ignore list so the cache can stay in the project dir without being scanned. Also switch module discovery to check for a build file directly instead of listing each directory's full contents, keeping the configuration-cache input surface minimal.

### Steps to test this PR
Verified: included module set unchanged before/after the fix (182 projects); using `./gradlew projects`

Make changes and ensure config cache is reuesed. 

---

> [!NOTE]
> **Medium Risk**
> Touches settings evaluation for all modules; removing the vpn `buildFileName` override could break that project if the custom filename is still in use.
> 
> **Overview**
> Fixes **configuration cache** invalidation on every build by keeping Gradle from treating the local **`build-cache`** directory as a module-discovery input.
> 
> **Module auto-discovery** now uses a **`hasBuildFile`** check (`build.gradle` / `build.gradle.kts` via `isFile()`) instead of listing each folder’s contents, so configuration-cache inputs stay narrow and stray names like `build.gradle.orig` are not matched. **`build-cache`** is added to the discovery ignore list, with a comment tying it to the local cache path under `buildCache { local { directory ... } }`.
> 
> The post-discovery loop that set **`vpn`**’s `buildFileName` to `vpn-build.gradle` is **removed** (worth confirming that module still resolves as intended if that override was still required).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 262ea10f1fe37e1c67d923cf3c67990f59d1dcab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -36,43 +36,28 @@ develocity {
     }
 }
 
+// Discover modules by checking for a build file directly rather than listing each directory's full
+// contents. A full listing gets fingerprinted as a configuration-cache input (so any new top-level
+// file in a module dir would invalidate it); statting the specific build file narrows the input to
+// exactly what matters and avoids reading large directories. It also drops the old
+// `name.contains("build.gradle")` match, which would also match stray files like build.gradle.orig.
+def hasBuildFile = { File dir -> new File(dir, "build.gradle").isFile() || new File(dir, "build.gradle.kts").isFile() }
+
 rootDir.eachFile(groovy.io.FileType.DIRECTORIES) { File parent ->
 
-    String[] ignoreFolders = ["buildSrc", "fastlane", "submodules", "node_modules", "gradle", "build", ".maestro"]
+    String[] ignoreFolders = ["buildSrc", "fastlane", "submodules", "node_modules", "gradle", "build", "build-cache", ".maestro"]
     if (!parent.name.startsWith(".") && !ignoreFolders.contains(parent.name)) {
-        Boolean shouldAddProject = false
-        parent.eachFile {
-            if (it.name.contains("build.gradle")) {
-                shouldAddProject = true
-                return
-            }
-        }
-        if (shouldAddProject) {
+        if (hasBuildFile(parent)) {
             include ":${parent.name}"
         } else {
-            parent.eachFile(groovy.io.FileType.DIRECTORIES) { child -> // We only one level deep
-                Boolean shouldAddSubProject = false
-                child.eachFile {
-                    if (it.name.contains("build.gradle")) {
-                        shouldAddSubProject = true
-                        return
-                    }
-                }
-                if (shouldAddSubProject) {
+            parent.eachFile(groovy.io.FileType.DIRECTORIES) { child -> // only one level deep
+                if (hasBuildFile(child)) {
                     include ":${child.name}"; project(":${child.name}").projectDir = new File("${parent.name}/${child.name}")
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


settings.gradle
```diff
@@ -36,29 +36,22 @@ develocity {
...
     if (!parent.name.startsWith(".") && !ignoreFolders.contains(parent.name)) {
-        Boolean shouldAddProject = false
-        parent.eachFile {
-            if (it.name.contains("build.gradle")) {
-                shouldAddProject = true
-                return
-            }
-        }
-        if (shouldAddProject) {
+        if (hasBuildFile(parent)) {
```
landomen, line 50: From Claude: `settings.gradle:62-67` still contains module-specific handling for a vpn project whose build file is named `vpn-build.gradle`:

```
  rootProject.children.each { subproject ->
      if (subproject.name == "vpn") {
          subproject.buildFileName = "${subproject.name}-build.gradle"
      }
  }
```
  
  `buildFileName` only configures the build script for a project that has already been included by auto-discovery — it does not include it. Under the old code, a vpn/ directory containing only `vpn-build.gradle` was discovered because i`t.name.contains("build.gradle")` matched `vpn-build.gradle`. 

The new `hasBuildFile` helper (`settings.gradle:44`) checks only for exact `build.gradle` / `build.gradle.kts`, so such a directory would not be discovered, `include ":vpn"` would never run, and the rename block at line 64 would silently never match.

 **Failure scenario**: if a top-level vpn/ module whose sole build file is `vpn-build.gradle` is (re)introduced, it is dropped from the build with no error — Gradle just won't know the project exists.

  Why this matters even though it's dormant: there is no vpn/ directory in the current tree (only `vpn-network/`, whose submodules use plain `build.gradle`), so this does not bite today and the verification was correct. But the file itself carries code (lines 62-67) that exists specifically to support a non-standard `vpn-build.gradle` name — which is exactly the case the new exact-match logic can no longer discover. The commit message notes dropping the contains match `"would also match stray files like build.gradle.orig"` but doesn't mention that the same contains match was the load-bearing mechanism for `vpn-build.gradle`.

**Suggested resolution (pick one):**
  - If the vpn module is genuinely gone, remove the now-dead vpn handling at lines 62-67 in this same commit, so the file is internally consistent and no one is misled into thinking `vpn-build.gradle` is still discoverable.
  - If vpn may return, have `hasBuildFile` also accept the module-named variant (e.g. also check `new File(dir, "${dir.name}-build.gradle").isFile())`, preserving the prior capability without reintroducing the broad contains match.

landomen: I think we can remove that special handling

landomen (approved):
LGTM, left one comment

settings.gradle
aibrahim-: Done.

PR #8820, opened 2026-06-08
Speed up tab state update on return-card close

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1157893581871903/task/1214802869599812

### Description

Closing a tab from the new-tab-page "return to" card used to defer the actual deletion until the "tab closed" snackbar timed out, so the tab count and the tab switcher only reflected the close several seconds later.

This makes the close take effect immediately while keeping it undoable:

- On close the tab is marked deletable, so it leaves the tab list / switcher right away; the delete is committed only when the snackbar is dismissed, and undo restores it.
- The card is now driven by a snapshot of the last-accessed tab, captured once per idle-return, instead of the live last-accessed flow. This keeps the shown tab stable across the close/undo toggle and stops it jumping to a different tab while the card is up.
- The floating native-input content offset is re-applied on new-tab-page layout passes so the card re-appearing on undo no longer overlaps the input.

### Steps to test this PR

_Return-card close updates state immediately_
- [x] Enable the native input field (internal build), open a few tabs, then background the app and return so the NTP "return to" card appears
- [x] Close the tab from the card → the tab count drops immediately and the tab is gone from the tab switcher straight away (no multi-second delay)
- [ ] Tap Undo on the snackbar → the tab is restored and the card re-appears showing the same tab, without overlapping the input
- [x] Let the snackbar time out instead → the tab is permanently closed

### UI changes
| Before  | After |
| ------ | ----- |

---

> [!NOTE]
> **Medium Risk**
> Changes tab lifecycle (mark deletable vs delete) and NTP/native-input layout timing; behavior is covered by updated ViewModel tests but touches user-visible tab state.
> 
> **Overview**
> Closing a tab from the new-tab **return** card now **marks it deletable immediately** (same pattern as tab switcher undo), so tab count and the switcher update right away; **permanent removal** happens on snackbar dismiss via `purgeDeletableTabs`, and **undo** calls `undoDeletable`.
> 
> The hatch no longer follows the **live last-accessed tab** flow—it **snapshots** `getLastAccessedTab()` once per idle return so the offered tab stays fixed through close/undo and doesn’t jump to another tab while the card is visible.
> 
> The **“tab closed” snackbar** is delegated to the host: `BrowserActivity.showUndoableSnackbar` (and Duck.ai input screen with activity-root parent + anchor) instead of a snackbar inside the hatch view, so it sits above the floating native input and can dismiss the keyboard.
> 
> **Native input layout** on NTP re-applies content padding on global layout passes (with cached hatch/logo lookups) so when the card reappears after undo it doesn’t overlap the input widget.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebec70eec884b876a2975545dd13421a372bd726. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
@@ -20,6 +20,7 @@ import android.animation.LayoutTransition
 import android.view.Gravity
 import android.view.View
 import android.view.ViewGroup
+import android.view.ViewTreeObserver
 import android.widget.FrameLayout
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.view.updateLayoutParams
@@ -273,6 +274,19 @@ class NativeInputLayoutCoordinator(
             }
         widgetView.addOnLayoutChangeListener(layoutListener)
         rootView.addOnLayoutChangeListener(layoutListener)
+
+        // NTP content can change visibility/size asynchronously without moving rootView/widgetView
+        // — e.g. the return hatch re-appearing on undo. Those passes never re-fire the layout
+        // listeners above, leaving a stale offset that overlaps the floating input. Observe the NTP
+        // subtree and re-apply the offset on its layout passes. Guarded by isWidgetAnimating like
+        // layoutListener, and idempotent via applyPadding's no-op-when-unchanged check.
+        val ntpContentView = newTabContent?.takeIf { !isBottom }
+        val globalLayoutListener =
+            ViewTreeObserver.OnGlobalLayoutListener {
+                if (isWidgetAnimating) return@OnGlobalLayoutListener
+                applyOffset()
+            }
+        ntpContentView?.viewTreeObserver?.addOnGlobalLayoutListener(globalLayoutListener)
         widgetView.addOnAttachStateChangeListener(
             object : View.OnAttachStateChangeListener {
                 override fun onViewAttachedToWindow(v: View) = Unit
@@ -287,6 +301,7 @@ class NativeInputLayoutCoordinator(
                     }
                     v.removeOnLayoutChangeListener(layoutListener)
                     rootView.removeOnLayoutChangeListener(layoutListener)
+                    ntpContentView?.viewTreeObserver?.takeIf { it.isAlive }?.removeOnGlobalLayoutListener(globalLayoutListener)
                     v.removeOnAttachStateChangeListener(this)
                 }
             },
--- a/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
@@ -177,6 +177,13 @@ class NativeInputLayoutCoordinator(
         if (targets.isEmpty()) return
         val anchor = widgetView.findViewById(R.id.inputModeWidgetCard) ?: widgetView
 
+        // Resolved once here instead of per call inside isLogoOnlyContent: that runs from the global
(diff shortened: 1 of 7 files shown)


browser/browser-ui/src/main/java/com/duckduckgo/browser/ui/newtab/hatch/NewTabReturnHatchViewModel.kt
```diff
@@ -152,15 +169,21 @@ class NewTabReturnHatchViewModel @Inject constructor(
...
+            tabRepository.undoDeletable(listOf(tabId))
+        }
         pendingClose.value = false
     }
 
     fun onTabClosedSnackbarDismissed(tabId: String) {
+        // The tab was already marked deletable on close; commit the deletion now.
         viewModelScope.launch(dispatchers.io()) {
-            tabRepository.deleteTabs(listOf(tabId))
+            tabRepository.purgeDeletableTabs()
```
YoussefKeyrouz, line 183: If other tabs are marked deletable from other places this would purge them all. Wouldn’t it be better to have a targeted purge? Like a version that takes the tabId and only purge that one?

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
```diff
@@ -273,6 +274,19 @@ class NativeInputLayoutCoordinator(
...
+        // listeners above, leaving a stale offset that overlaps the floating input. Observe the NTP
+        // subtree and re-apply the offset on its layout passes. Guarded by isWidgetAnimating like
+        // layoutListener, and idempotent via applyPadding's no-op-when-unchanged check.
+        val ntpContentView = newTabContent?.takeIf { !isBottom }
+        val globalLayoutListener =
+            ViewTreeObserver.OnGlobalLayoutListener {
+                if (isWidgetAnimating) return@OnGlobalLayoutListener
+                applyOffset()
+            }
+        ntpContentView?.viewTreeObserver?.addOnGlobalLayoutListener(globalLayoutListener)
```
YoussefKeyrouz, line 300: OnGlobalLayoutListener registers on the window's shared ViewTreeObserver, not just this view, so  applyOffset() will on every global layout pass anywhere in the window (autocomplete list updating, keyboard showing, User scrolling etc.), not only when the NTP changes. Do we actually need to recompute the offset on every layout pass, or can we jsut scope this to the specific event we care about (the hatch reappearing)?

browser/browser-ui/src/main/java/com/duckduckgo/browser/ui/newtab/hatch/NewTabReturnHatchViewModel.kt
```diff
@@ -141,6 +153,11 @@ class NewTabReturnHatchViewModel @Inject constructor(
         pixel.fire(NewTabReturnHatchPixelName.OPTION_SELECTED_CLOSE_TAB_DAILY, type = Daily())
         val tabId = viewState.value.currentTabId
         if (tabId.isEmpty()) return
+        // Mark the tab deletable now so it disappears from the tab list/switcher immediately
+        // (recoverable via undo); the actual delete is committed when the snackbar is dismissed.
+        viewModelScope.launch(dispatchers.io()) {
+            tabRepository.markDeletable(listOf(tabId))
+        }
```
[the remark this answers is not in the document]
YoussefKeyrouz: I think this is worth resolving.

YoussefKeyrouz (approved):
Delete works well. There’s one edge case that you could cover part of this or in a new PR, up to you:

Steps: Close the hatch, then undo. When you first switch to the Duck.ai tab, it does not switch the view: the hatch and favourites are still showing up and you can't see suggestions

PR #8834, opened 2026-06-10
Fix onboarding decoration sizing on tablets; unify waving-Dax sizing

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207908166761516/task/1215297293922286?focus=true

### Description

Fixes several onboarding layout issues in the brand-design-update flow and unifies how the new-tab waving Dax is sized.

- **Decoration sizing & fit (welcome, comparison, address-bar, input; phone + tablet):** wings and the bobbing/waving Dax now shrink-to-fit the room below the card via a shared clamp, with a post-layout corrector that re-checks on every pass and only ever shrinks or hides it (the card scrolls only as a last resort). Fixes the tablet welcome-screen scrollbar, decorations that over-reserved height or ignored live font-scale changes, the decoration vanishing mid dialog-transition, and the input-screen left wing not showing.
- **Dialog card vs navigation/gesture bar (large fonts):** the card reserves the bottom-bar inset only when it is the bottom-most element, so its buttons no longer fall under the bar at large fonts while decorations still extend to the screen edge.
- **Input demo screen scrollbar and shift (tablet, keyboard open):** on the "Ready to get started? Try a search or AI chat!" screen the card no longer shows a scrollbar, and no longer shifts position, when the keyboard opens. It stays put and grows to fit its content, scrolling only when the content genuinely exceeds the space available (small screens or very large fonts).
- **Subscription Dax + fin (tablet landscape):** the rotated Dax's bounding box is inset back to the artwork before the fit check, keeping both visible instead of being hidden by the empty rotated corners.
- **Unified waving-Dax sizing:** replaces the per-CTA `autoSize` flag and geometric collision engine with the same room-based clamp the welcome screen uses (`minHeightDp`/`maxHeightDp`, where `min == max` means fixed size).

The new-tab waving Dax and fin are gated by the pre-existing `onboardingImprovements` kill switch (default ON), which this branch does not change; when it is OFF they render as develop does (top-anchored, fixed height, shown immediately).

Everything this branch adds is gated behind the new `onboardingImprovementsV2` kill switch (default ON): the welcome-page decoration fit engine plus corrector, the new-tab waving-Dax room-based sizing, and the smaller welcome-layout tweaks (the card-clear-of-nav-bar bottom inset, the `layout_constrainedHeight` change on the dialog card, and `adjustViewBounds` on the bottom wing, plus the input-demo card staying scrollbar-free and steady when the keyboard opens). When `onboardingImprovementsV2` is OFF the branch behaves exactly like develop: develop's welcome decoration sizing with no reactive shrink and no corrector, develop's fit-based show/hide for the new-tab waving Dax (the pre-V2 geometric sizing, with no room-based resize), and develop's original XML values. `onboardingImprovements` still controls whether the waving Dax appears at all.

### Steps to test this PR

Patches below are valid unified diffs. Save one to a file and run `git apply <file>.patch`, or use Android Studio -> **Git -> Apply Patch from Clipboard**. Reinstall after applying. To revert, run `git apply -R <file>.patch` or discard the change.

**Turn the `onboardingImprovementsV2` flag OFF** (apply turns it off; revert/`-R` turns it back on). Reverts this branch's sizing refinements to develop while `onboardingImprovements` keeps the waving Dax present:

```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
--- a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
@@ -52,6 +52,6 @@ interface OnboardingBrandDesignUpdateToggles {
      * and the new-tab waving-Dax room-based sizing. Off restores develop's sizing behavior.
      * onboardingImprovements still gates whether the waving Dax appears at all.
      */
-    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
+    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
     fun onboardingImprovementsV2(): Toggle
 }
```

**Optional: force the subscription onboarding bubble to show** without Privacy Pro eligibility (debug-only testing aid, **do not commit**):

```diff
diff --git a/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt b/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
index 6011cf87a2..293c695422 100644
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/CtaViewModel.kt
@@ -113,7 +113,7 @@ class CtaViewModel @Inject constructor(
             }
 
     private suspend fun isSubscriptionCtaAvailable(): Boolean =
-        subscriptions.isEligible() && hasNoSubscription() && extendedOnboardingFeatureToggles.privacyProCta().isEnabled()
+        true
 
     private suspend fun isBrandDesignUpdateEnabled(): Boolean = withContext(dispatchers.io()) {
         onboardingBrandDesignUpdateToggles.brandDesignUpdate().isEnabled()
@@ -446,6 +446,7 @@ class CtaViewModel @Inject constructor(
 
     @WorkerThread
     private suspend fun canShowSubscriptionCta(): Boolean {
+        if (true) return isSubscriptionCtaAvailable()
         if (hideTips() || daxDialogSubscriptionShown() || !isSubscriptionCtaAvailable()) return false
 
         return if (onboardingStore.isDuckAiOnboardingFlow()) {
```

_Tablet welcome screen (the scrollbar fix)_
- [ ] On a tablet, reonboard and view the welcome screen in portrait and landscape.
- [ ] Confirm no scrollbar appears and the wings + bobbing Dax sit fully within the visible area, above the navigation bar.
- [ ] Repeat at large font scales (1.5 / 1.8): decorations shrink further to fit rather than overflowing or showing a scrollbar.
- [ ] Increase the system font scale **while the welcome/comparison screen is visible**: the decoration shrinks (and if there's still not enough room, hides so the card fills the space) with no scrollbar.
- [ ] Decrease the font scale back (or rotate the device): the decoration returns to full size when there's room again.
- [ ] On a phone at the largest font scale, confirm that only once the decoration has hidden does the card itself scroll: scrolling is the last resort, not the default.
- [ ] Step forward through welcome → comparison → address-bar → input and confirm the decoration never briefly flickers or vanishes during the transition or entrance animation.
- [ ] Repeat the large-font checks (up to the maximum system font scale) in **both gesture and 3-button navigation**, on a phone and a tablet, and include the "Compare browsers" screen where the card is tallest: the card and its buttons always stay fully above the navigation/gesture bar and are never clipped by it.
- [ ] At a font scale large enough that the decoration hides, the card fills the freed space but still stops short of the bar, and the bottom button stays fully visible and tappable.

_Address-bar & input-screen decorations (phone + tablet)_
- [ ] Reonboard and step forward to the "Where should I put your address bar?" screen: on first arrival the bobbing Dax appears below the card and animates.
- [ ] Continue to the "Want easy access to private AI chat?" screen: the wing is visible at the bottom-left and sits clear of the card (a small gap remains; it never touches the card).
- [ ] Repeat in both gesture and 3-button navigation, on a phone and a tablet: the Dax and the wing stay visible in each case.
- [ ] Repeat the above at large font scales: the bobbing Dax (address-bar screen) and the left wing (input screen) still appear, and the card's buttons stay above the navigation/gesture bar.

_Input demo screen scrollbar ("Ready to get started?"; phone + tablet, portrait + landscape)_
- [ ] Reonboard and step forward to the final input demo screen ("Ready to get started? Try a search or AI chat!"); the keyboard opens automatically.
- [ ] On a tablet, in both portrait and landscape: confirm no scrollbar appears on the card, and the card does not shift or jump when the keyboard opens.
- [ ] On a phone, in both portrait and landscape: confirm the same (no scrollbar when the content fits, and the card stays put as the keyboard opens).
- [ ] On a small-screen device (or any phone at a large system font scale, 1.5 and above): raise the font until the content no longer fits the space above the keyboard, and confirm the card then scrolls with a scrollbar so the suggestions and the button stay reachable. Scrolling should appear only once the content genuinely overflows, and this overflow case is the important one to cover.
- [ ] Repeat that large-font overflow check in both gesture and 3-button navigation, and confirm the card stays clear of the navigation/gesture bar.

_New-tab bubbles (default config: `onboardingImprovements` and `onboardingImprovementsV2` both ON)_
- [ ] Advance to the new-tab onboarding bubbles (visit-site, end, and subscription with the optional patch).
- [ ] Confirm the waving Dax + shark fin appear only when there's room below the bubble and shrink with the available space; the horizontal peek stays in proportion as it scales.
- [ ] Raise the keyboard: the Dax + fin disappear immediately as the keyboard rises and never linger or flash over it; dismiss the keyboard and they return.
- [ ] Increase the font scale until the Dax no longer fits: it shrinks then hides; reduce it again and it returns.
- [ ] **Tablet landscape, subscription bubble:** confirm the rotated Dax and fin remain visible.
- [ ] Phone landscape: Dax hidden.

_`onboardingImprovementsV2` OFF (this branch's sizing reverts to develop)_
- [ ] Apply the `onboardingImprovementsV2` flag-OFF patch, reinstall, reonboard. Leave `onboardingImprovements` ON.
- [ ] Welcome and comparison screens: decorations use develop's sizing and do not reactively shrink or hide as the card grows (no live font-scale shrink, no corrector behaviour).
- [ ] Address-bar and input screens: decorations behave as on develop.
- [ ] Input demo screen ("Ready to get started?"): the card behaves as on develop when the keyboard opens (no keyboard-aware bottom inset and no overflow-only height clamp).
- [ ] New-tab waving Dax (visit-site, end, subscription): reverts to develop's fit-based show/hide (the pre-V2 geometric check). The Dax still starts hidden and appears only when it fits below the bubble, and still hides when the keyboard is up, but it no longer resizes to fill the available room. (Fully legacy top-anchored, shown-immediately rendering only happens with `onboardingImprovements` OFF, which this PR does not change.)
- [ ] With V2 OFF the welcome layout matches develop exactly, so confirm each of these reverts to develop rather than the branch's V2-on behaviour: the card uses develop's bottom spacing with no added navigation-bar inset, the dialog card keeps develop's `layout_constrainedHeight`, and the bottom wing has no `adjustViewBounds`.
- [ ] Revert the patch and confirm the branch's default V2 behaviour returns.

### UI changes
See task [comment](https://app.asana.com/1/137249556945/task/1215297293922286/comment/1215628391635161?focus=true)

---

> [!NOTE]
> **Medium Risk**
> Touches onboarding layout across phones/tablets, keyboard, and font scale with dual code paths behind a remote toggle; regressions would be visible but not security-critical.
> 
> **Overview**
> Adds **`onboardingImprovementsV2`** (default ON) so welcome-page wings/Dax and new-tab **waving Dax** use shared **room-based** sizing (`minHeightDp`/`maxHeightDp` via `OnboardingDecorationSizing`), with V2 OFF preserving prior develop behavior while **`onboardingImprovements`** still controls whether the waving Dax shows at all.
> 
> On the **brand-design welcome flow**, decorations are sized with `calculateDecorationHeight` and kept in sync by a new **`OnboardingDecorationFitCorrector`** (pre-draw shrink/hide, measured heights during transitions, bottom-bar inset only when the card is bottom-anchored with no decoration below). **`WavingDaxController`** switches to viewport/card-bottom height fitting and proportional horizontal peek when V2 is on; bubble CTAs **bottom-anchor** the Lottie Dax under V2 instead of fixed top-anchored height.
> 
> Smaller tweaks: input-demo title uses `.html()`, comparison rows drop `preventWidows()`; **`preventWidows()`** only replaces the last space when the string has multiple spaces.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b75686c4846166d16886de1f574d5f23c99718bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
@@ -1625,6 +1625,7 @@ sealed class DaxBubbleCta(
         open val isLightTheme: Boolean,
         open val deviceInfo: DeviceInfo,
         open val onboardingImprovementsEnabled: Boolean = true,
+        open val onboardingImprovementsV2Enabled: Boolean = true,
     ) : DaxBubbleCta(
         ctaId = ctaId,
         title = title,
@@ -1676,7 +1677,7 @@ sealed class DaxBubbleCta(
 
         private val wavingDaxController: WavingDaxController? by lazy {
             if (onboardingImprovementsEnabled && this is ShowsWavingDax) {
-                WavingDaxController(showArrow, deviceInfo, wavingDaxSpec)
+                WavingDaxController(showArrow, deviceInfo, wavingDaxSpec, improvementsV2Enabled = onboardingImprovementsV2Enabled)
             } else {
                 null
             }
@@ -1716,7 +1717,7 @@ sealed class DaxBubbleCta(
                     showsWavingDax.configureWavingDax(
                         dax = dax,
                         deviceInfo = deviceInfo,
-                        improvementsEnabled = onboardingImprovementsEnabled,
+                        improvementsEnabled = onboardingImprovementsEnabled && onboardingImprovementsV2Enabled,
                     )
                     if (onboardingImprovementsEnabled) {
                         dax.isInvisible = true
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
@@ -1567,19 +1567,34 @@ sealed class DaxBubbleCta(
         val restartWavingDax: Boolean get() = false
         val wavingDaxSpec: WavingDaxSpec
 
-        fun configureWavingDax(dax: LottieAnimationView, deviceInfo: DeviceInfo) {
+        fun configureWavingDax(
+            dax: LottieAnimationView,
+            deviceInfo: DeviceInfo,
+            improvementsEnabled: Boolean = false,
+        ) {
             val spec = wavingDaxSpec
             val density = dax.resources.displayMetrics.density
             dax.rotation = spec.rotationDegrees
-            dax.translationX = spec.translationXDp * density
-            dax.translationY = spec.translationYDp * density
+            if (improvementsEnabled) {
+                dax.translationY = spec.bottomTranslationYDp * density
+            } else {
+                dax.translationX = spec.translationXDp * density
+                dax.translationY = spec.translationYDp * density
+            }
             (dax.layoutParams as? ConstraintLayout.LayoutParams)?.let { lp ->
                 lp.startToStart = if (spec.anchorToCardOnTablet && deviceInfo.isTablet()) {
                     R.id.brandDesignCardView
                 } else {
                     ConstraintLayout.LayoutParams.PARENT_ID
                 }
-                lp.height = (spec.heightDp * density).toInt()
+                if (improvementsEnabled) {
+                    lp.topToBottom = ConstraintLayout.LayoutParams.UNSET
+                    lp.bottomToBottom = ConstraintLayout.LayoutParams.PARENT_ID
+                } else {
+                    lp.bottomToBottom = ConstraintLayout.LayoutParams.UNSET
+                    lp.topToBottom = ConstraintLayout.LayoutParams.PARENT_ID
+                    lp.height = (spec.maxHeightDp * density).toInt()
+                }
                 dax.layoutParams = lp
             }
         }
@@ -1589,9 +1604,13 @@ sealed class DaxBubbleCta(
         val rotationDegrees: Float,
         val translationXDp: Float,
         val translationYDp: Float,
-        val heightDp: Float,
+        val minHeightDp: Float,
+        val maxHeightDp: Float,
         val anchorToCardOnTablet: Boolean,
-    )
+    ) {
+        /** Bottom-anchored resting offset: the legacy top-anchored position re-expressed against the parent bottom. */
+        val bottomTranslationYDp: Float get() = translationYDp + maxHeightDp
+    }
 
     abstract class BrandDesignUpdateBubbleCta(
         ctaId: CtaId,
@@ -1658,7 +1677,7 @@ sealed class DaxBubbleCta(
 
         private val wavingDaxController: WavingDaxController? by lazy {
             if (onboardingImprovementsEnabled && this is ShowsWavingDax) {
-                WavingDaxController(showArrow, deviceInfo)
+                WavingDaxController(showArrow, deviceInfo, wavingDaxSpec)
             } else {
                 null
             }
@@ -1695,7 +1714,11 @@ sealed class DaxBubbleCta(
         internal fun applyWavingDaxState(container: View, showsWavingDax: ShowsWavingDax?) {
             container.findViewById<LottieAnimationView>(R.id.wavingDax)?.let { dax ->
                 if (showsWavingDax != null && !container.isPhoneLandscape()) {
-                    showsWavingDax.configureWavingDax(dax, deviceInfo)
+                    showsWavingDax.configureWavingDax(
+                        dax = dax,
+                        deviceInfo = deviceInfo,
+                        improvementsEnabled = onboardingImprovementsEnabled,
+                    )
                     if (onboardingImprovementsEnabled) {
                         dax.isInvisible = true
                     } else {
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
@@ -1608,7 +1608,6 @@ sealed class DaxBubbleCta(
         val maxHeightDp: Float,
         val anchorToCardOnTablet: Boolean,
     ) {
-        /** Bottom-anchored resting offset: the legacy top-anchored position re-expressed against the parent bottom. */
         val bottomTranslationYDp: Float get() = translationYDp + maxHeightDp
     }
 
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/DaxEndBrandDesignUpdateBubbleCta.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/DaxEndBrandDesignUpdateBubbleCta.kt
@@ -57,7 +57,8 @@ data class DaxEndBrandDesignUpdateBubbleCta(
         // so we need to nudge Dax to the right to ensure both his legs are behind the address bar.
         translationXDp = if (isOmnibarBottom) -30f else -40f,
         translationYDp = -150f,
-        heightDp = 178f,
+        minHeightDp = 178f,
+        maxHeightDp = 178f,
         anchorToCardOnTablet = true,
     )
 
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/DaxEndBrandDesignUpdateBubbleCta.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/DaxEndBrandDesignUpdateBubbleCta.kt
@@ -33,6 +33,7 @@ data class DaxEndBrandDesignUpdateBubbleCta(
     override val isLightTheme: Boolean,
     override val deviceInfo: DeviceInfo,
     override val onboardingImprovementsEnabled: Boolean,
+    override val onboardingImprovementsV2Enabled: Boolean,
     val isOmnibarBottom: Boolean,
 ) : DaxBubbleCta.BrandDesignUpdateBubbleCta(
     ctaId = CtaId.DAX_END,
@@ -47,6 +48,7 @@ data class DaxEndBrandDesignUpdateBubbleCta(
     isLightTheme = isLightTheme,
     deviceInfo = deviceInfo,
     onboardingImprovementsEnabled = onboardingImprovementsEnabled,
+    onboardingImprovementsV2Enabled = onboardingImprovementsV2Enabled,
 ),
     DaxBubbleCta.ShowsWavingDax {
     override val activeIncludeId: Int = R.id.primaryCta
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/WavingDaxController.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/WavingDaxController.kt
@@ -92,7 +92,11 @@ internal class WavingDaxController(
     )
 
     private fun decideFit(container: View) {
-        val height = computeDaxFitHeight(container) ?: return
+        val height = computeDaxFitHeight(container)
+        if (height == null) {
+            applyFitResult(container, false)
+            return
+        }
         applyDaxHeight(container, height)
         applyFitResult(container, true)
     }
--- a/app/src/main/java/com/duckduckgo/app/cta/ui/WavingDaxController.kt
+++ b/app/src/main/java/com/duckduckgo/app/cta/ui/WavingDaxController.kt
@@ -18,17 +18,17 @@ package com.duckduckgo.app.cta.ui
 
 import android.animation.ValueAnimator
 import android.content.res.Configuration
-import android.graphics.RectF
+import android.graphics.Rect
 import android.view.View
 import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
 import androidx.interpolator.view.animation.FastOutSlowInInterpolator
 import com.airbnb.lottie.LottieAnimationView
 import com.duckduckgo.app.browser.R
+import com.duckduckgo.app.onboarding.ui.page.OnboardingDecorationSizing
 import com.duckduckgo.common.ui.view.shape.DaxOnboardingBubbleBrandDesignUpdateCardView
 import com.duckduckgo.common.utils.device.DeviceInfo
 import com.duckduckgo.common.utils.device.isTablet
-import kotlin.math.abs
 
 /**
  * Keeps the waving Dax below a new-tab bubble in sync with the space available: decides whether the
@@ -41,6 +41,7 @@ import kotlin.math.abs
 internal class WavingDaxController(
     private val showArrow: Boolean,
     private val deviceInfo: DeviceInfo,
+    private val wavingDaxSpec: DaxBubbleCta.WavingDaxSpec,
 ) {
 
     private var fitArrowAnimator: ValueAnimator? = null
@@ -51,11 +52,12 @@ internal class WavingDaxController(
     fun applyFit(container: View) {
         this.container = container
         container.removeCallbacks(fitRunnable)
-        val fits = computeDaxFits(container) ?: return
-        if (fits) {
-            container.postDelayed(fitRunnable, FIT_SETTLE_DELAY_MS)
-        } else {
+        val height = computeDaxFitHeight(container)
+        if (height == null) {
             applyFitResult(container, false)
+        } else {
+            applyDaxHeight(container, height)
+            container.postDelayed(fitRunnable, FIT_SETTLE_DELAY_MS)
         }
     }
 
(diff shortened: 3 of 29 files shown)


app/src/main/java/com/duckduckgo/app/cta/ui/Cta.kt
```diff
@@ -1570,9 +1585,13 @@ sealed class DaxBubbleCta(
         val rotationDegrees: Float,
         val translationXDp: Float,
         val translationYDp: Float,
-        val heightDp: Float,
+        val minHeightDp: Float,
+        val maxHeightDp: Float,
         val anchorToCardOnTablet: Boolean,
-    )
+    ) {
+        /** Bottom-anchored resting offset: the legacy top-anchored position re-expressed against the parent bottom. */
```
mikescamell: 🔍  I don't think we need this comment

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
```diff
@@ -0,0 +1,105 @@
...
+
+        if (target == null) {
+            deco.isGone = true
+            dialog.updateLayoutParams<ConstraintLayout.LayoutParams> {
+                verticalBias = 0f
+                bottomToTop = ConstraintLayout.LayoutParams.UNSET
+                bottomToBottom = ConstraintLayout.LayoutParams.PARENT_ID
+            }
+            onDecorationHidden()
+            return false
```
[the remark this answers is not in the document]
mikescamell: This is the intended shrink-only behaviour, not an oversight. Onboarding is a forward-only flow: each step tracks a distinct decoration view and explicitly hides the previous step's, so the corrector never re-evaluates a view it has already hidden. The only events that increase vertical space — orientation / screen-size / font-scale changes — recreate `OnboardingActivity` (it declares no `android:configChanges`), and the fresh layout path rebuilds full-size decorations from scratch (device-verified: font scale 2.5 → 1.0 restored full size). There is also no in-app back navigation between steps.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
```diff
@@ -0,0 +1,105 @@
...
+        val deco = decoration ?: return true
+        if (deco.isGone) return true
+        if (root.height == 0 || dialog.height == 0) return true
+        if (BrandDesignUpdateOnboardingLayoutHelper.isInScrollableContainer(dialog, root)) return true
+
+        val viewport = cardContainer.parent as? View ?: return true
+        val overflow = (cardContainer.height - viewport.height).coerceAtLeast(0)
+        val available = root.height - root.paddingTop - root.paddingBottom
+        val dialogParams = dialog.layoutParams as ViewGroup.MarginLayoutParams
+        val dialogSpace = dialog.height + overflow + dialogParams.topMargin + dialogParams.bottomMargin
```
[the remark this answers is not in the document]
mikescamell: Added a `viewport.height == 0` early-return so a not-yet-laid-out ScrollView viewport can't be read as full-content overflow, plus a Robolectric case (`whenViewportNotLaidOutThenNoOp`). This closes the zero-viewport path specifically; the shrink path stays monotonic by design, so the held `!isAnimating` lever (ADR 0001) is what would cover a transient mid-animation frame if one ever surfaced — not applied here since no jank was observed on device.

mikescamell:
* **#8851** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8851?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8834** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8834?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8834?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/cta/ui/WavingDaxController.kt
```diff
@@ -72,50 +75,75 @@ internal class WavingDaxController(
...
+
+    private fun applyDaxHeight(container: View, height: Int) {
+        val dax = container.findViewById<LottieAnimationView>(R.id.wavingDax) ?: return
+        val density = dax.resources.displayMetrics.density
+        val maxHeightPx = (wavingDaxSpec.maxHeightDp * density).toInt()
+        val scale = daxHorizontalScale(height, maxHeightPx)
+        dax.translationX = wavingDaxSpec.translationXDp * density * scale
+        if (dax.layoutParams.height != height) {
+            dax.layoutParams = dax.layoutParams.apply { this.height = height }
+        }
```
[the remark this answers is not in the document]
mikescamell: The waving Dax is bottom-anchored and intended to shrink from the top with its bottom fixed under the fin, so the fixed `translationY` is deliberate. Only `translationX` scales, to keep the horizontal peek proportional. Verified on device (Galaxy S25) across font scales: the Dax stays aligned to the fin at shrunk heights, with no misalignment.

app/src/main/java/com/duckduckgo/app/cta/ui/WavingDaxController.kt
```diff
@@ -72,50 +75,75 @@ internal class WavingDaxController(
...
+        val visibleFrame = Rect()
+        container.getWindowVisibleDisplayFrame(visibleFrame)
+        val usableBottom = visibleFrame.bottom
+
+        return daxFitHeight(
+            usableBottom = usableBottom,
+            cardBottom = cardBottom,
+            marginPx = marginPx,
+            minHeightPx = minHeightPx,
+            maxHeightPx = maxHeightPx,
```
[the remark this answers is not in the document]
mikescamell: The Dax is bottom-anchored (`bottomToBottom=parent`) with a `translationY` set once in `configureWavingDax`. `applyDaxHeight` only ever writes `height` and `translationX`, never `translationY`, so its bottom edge is layout-pinned independent of the fit height. A larger height only extends the top upward; it can't move anything toward the IME edge.

The fin/card overlap is the same observable I verified on device (Galaxy S25) for this subscription bubble across font scales at shrunk heights, where the Dax stays aligned to the fin with no misalignment.

I'm not subtracting the card's 36dp `marginBottom` here. It has no bottom-constrained sibling in this layout, so it isn't extending the container, and subtracting it speculatively would shrink or hide the Dax too eagerly.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
```diff
@@ -2192,6 +2222,65 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
...
+            decorationView.updateLayoutParams { this.height = height }
+            decorationFitCorrector?.track(
+                decorationView,
+                minHeightPx = minHeightDp.toPx(),
+                maxHeightPx = maxHeightDp.toPx(),
+                bottomOverlapPx = bottomOverlapPx,
+            )
+        } else {
+            decorationFitCorrector?.clear()
+        }
```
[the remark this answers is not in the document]
mikescamell: Fixed in 43be7e81d4. The corrector now gates its bottom-inset reservation on an enabled flag that the fragment sets from its ViewState collector, defaulting off. With onboardingImprovementsV2 off the corrector stays inert, so syncCardBottomInset leaves the card bottom margin at 0 and V2-off layout matches legacy on the phone address-bar path.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
```diff
@@ -529,9 +526,21 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
         binding.daxDialogCta.cardContainer.setOnClickListener { viewModel.onDialogTapped() }
         binding.root.setOnClickListener { viewModel.onBackgroundTapped() }
 
+        decorationFitCorrector = OnboardingDecorationFitCorrector(
+            root = binding.root,
+            dialog = binding.daxDialogCta.root,
+            cardContainer = binding.daxDialogCta.cardContainer,
+            onDecorationHidden = { binding.daxDialogCta.cardView.setArrowDepthFraction(0f) },
+            cardBottomInsetPx = { cardBottomInsetPx },
+        ).also { it.attach() }
```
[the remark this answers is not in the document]
mikescamell: This is intentional. The `onboardingImprovementsV2` flag only reaches the view through `ViewState`, which the ViewModel populates asynchronously on the IO dispatcher (the read is `suspend` because onboarding is a cold-start first-reader, so a synchronous main-thread read would risk disk IO on first launch). The window insets, though, dispatch during the first layout pass in `onViewCreated`, well before that flag resolves.

Gating the tablet root inset on the flag would mean either acting on the default value before it settles (wrong on a V2-off device, with no guaranteed re-dispatch to correct it afterwards) or adding machinery to force `requestApplyInsets` once the ViewState arrives. We chose not to take that on. The only consequence is on the V2-off tablet path, which is the kill-switch rollback state rather than the rollout target, so the minor divergence there is acceptable.

LukasPaczos:
@mikescamell, the input demo screen still doesn't scale correctly:
Screenshot_20260703_114048
Screenshot_20260703_114058

It might be related to the fact that on this specific screen, the whole card isn't scrollable (as is with other screen) but rather its content. We've had a [separate task to look into it](https://app.asana.com/1/137249556945/project/72649045549333/task/1215976514691679?focus=true), but perhaps this is the right place to do so? What do you think?

mikescamell:
> @mikescamell, the input demo screen still doesn't scale correctly: Screenshot_20260703_114048 Screenshot_20260703_114058
> 
> It might be related to the fact that on this specific screen, the whole card isn't scrollable (as is with other screen) but rather its content. We've had a [separate task to look into it](https://app.asana.com/1/137249556945/project/72649045549333/task/1215976514691679?focus=true), but perhaps this is the right place to do so? What do you think?

Yeah makes sense, let me get this one fixed as well so we leave onboarding in a good place 👍

app/src/main/java/com/duckduckgo/app/cta/ui/DaxEndBrandDesignUpdateBubbleCta.kt
```diff
@@ -57,7 +59,8 @@ data class DaxEndBrandDesignUpdateBubbleCta(
         // so we need to nudge Dax to the right to ensure both his legs are behind the address bar.
         translationXDp = if (isOmnibarBottom) -30f else -40f,
         translationYDp = -150f,
-        heightDp = 178f,
+        minHeightDp = 178f,
+        maxHeightDp = 178f,
```
LukasPaczos, line 63: Should this also be applied to `DaxDuckAiEndBrandDesignUpdateBubbleCta`?

LukasPaczos (approved):
LGTM, I tested various combinations, including font sizes and navigation buttons, on a Pixel 9 Pro and Samsung Tab S9+, all looks generally fine. Minor issues I noticed:

1. On max font size and with navigation buttons, the browser comparison screen becomes scrollable and title jumps to new line (not sure what the expectations is).

Image

2. On same settings, the input options text gets clipped, but interestingly recovers after rotation:

| Before rotation  | After rotation |
| ------ | ----- |
Image|Image|

3. The 'Try input' screen is not scrollable with keyboard up:

https://github.com/user-attachments/assets/d22e0639-1ec7-451c-89c5-21928e28f020

None of these are blockers to me - 1 and 2 are an unlikely worst-case scenario. 3 would be highest prio to me. Would be great to understand what the changes to improve should look like, but that can happen in follow-ups.

app/src/main/java/com/duckduckgo/app/cta/ui/DaxEndBrandDesignUpdateBubbleCta.kt
mikescamell: Unless I misunderstand your question, it is applied [here](https://github.com/duckduckgo/Android/pull/8834/changes/BASE..a483faa98e1d8895582f6552ba28361565a1ab1f#diff-99839c4e55fd2b37b5e3f822e5b4983c22a237c96b7b48c5369cc579d056a3ffR64) 👍

mikescamell:
> LGTM, I tested various combinations, including font sizes and navigation buttons, on a Pixel 9 Pro and Samsung Tab S9+, all looks generally fine. Minor issues I noticed:
> 
> 1. On max font size and with navigation buttons, the browser comparison screen becomes scrollable and title jumps to new line (not sure what the expectations is).
> 
> Image
> 2. On same settings, the input options text gets clipped, but interestingly recovers after rotation:
> 
> Before rotation	After rotation
> Image	Image
> 3. The 'Try input' screen is not scrollable with keyboard up:
> 
>  Screen_recording_20260707_150859.mp4 
> None of these are blockers to me - 1 and 2 are an unlikely worst-case scenario. 3 would be highest prio to me. Would be great to understand what the changes to improve should look like, but that can happen in follow-ups.

1.  I think this is an issue with the preventWidows function, I'll take a look because it hopefully should be a quick fix.
2. I have a quick look but leave it if it's going to take too long
3. I'll investigate this, I focused on getting the scrollbars removed like the other screens when there is actually enough room to display content, I misunderstood your comment 👍

app/src/main/java/com/duckduckgo/app/cta/ui/DaxEndBrandDesignUpdateBubbleCta.kt
LukasPaczos: Ah yes, missed it.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
```diff
@@ -2749,7 +2749,7 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
                     false,
                 ) as LinearLayout
                 rowView.findViewById<ImageView>(R.id.rowIcon).setImageResource(row.iconRes)
-                rowView.findViewById<DaxTextView>(R.id.rowText).text = getString(row.textRes).preventWidows()
+                rowView.findViewById<DaxTextView>(R.id.rowText).text = getString(row.textRes)
```
LukasPaczos, line 2779: I think we want to keep this for comparison chart rows, unless there's a specific reason to drop it?

mikescamell: @LukasPaczos  For the same reason as you highlighted on the title, and in your screenshot. Words can break mid word

Screenshot 2026-07-09 at 14 56 25

LukasPaczos: Hmm, by this logic, `preventWidows` should never be used as it always bares the risk of mid-word break happening in edge cases. That said, imo a widowed word is less harmful than a split word from accessibility perspective, so let's keep this specific change.

LukasPaczos (approved):

LukasPaczos (approved):

PR #8839, opened 2026-06-10
Android: App missing pixels for voice chat service and notification

karlenDimla:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1215567316447344?focus=true

### Description
Adds the missing pixels related to the voice chat service

### Steps to test this PR                                                                                                                                                                                  ### Setup                                                                                                                                                                                                          
  - [ ] Install an internal build with pixel logging visible (logcat filter `Pixel sent`)                                                                                                           
  - [ ] Open Duck.ai and start a Duck.ai voice chat session in a tab (so the microphone foreground service starts and the "Duck.ai voice mode active" notification appears in the shade)                             
  - [ ] Confirm the notification shows both **Open Chat** and **End Chat** actions                                                                                                                                   
  - [ ] Clear any prior pixel log entries before each scenario below                                                                                                                                                 
                                                                                                                                                                                                                     
### 1. Voice session start — service `onCreate`                                                                                                                                                                    
  - [ ] When the voice session begins, pixel `m_aichat_voice_service_started` IS fired exactly once        
  - [ ] The Duck.ai voice notification IS visible in the system notification shade                                                                                                                                   
                                                                                                                                                                                                                     
  ### 2. Tap "Open Chat" notification action — app already in foreground (warm path via `onNewIntent`)                                                                                                               
  - [ ] With the Duck.ai voice session active, switch to another app so Duck.ai goes to background but is NOT killed                                                                                               
  - [ ] Pull down the shade and tap the **Open Chat** action on the Duck.ai voice notification                                                                                                                       
  - [ ] The Duck.ai chat tab IS brought to the foreground (existing tab, not a new one)                                                                                                                              
  - [ ] Pixel `m_aichat_voice_notification_open_chat_tapped` IS fired exactly once                                                                                                                                   
  - [ ] Pixel `m_aichat_voice_service_killed` is **NOT** fired (the session is still active)   
  - [ ] Switch to another tab and background the app
  - [ ] Pull down the shade and tap the **Open Chat** action on the Duck.ai voice notification             
  - [ ]  The App with Duck.ai chat tab IS brought to the foreground           
  - [ ]  Pixel `m_aichat_voice_notification_open_chat_tapped` IS fired exactly once                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
  ### 3. Tap "End Chat" notification action                                                                                                                                                                        
  - [ ] With the voice session active, pull down the shade and tap the **End Chat** action                                                                                                                           
  - [ ] Pixel `m_aichat_voice_notification_end_chat_tapped` IS fired exactly once                                                                                                                                    
  - [ ] The voice session ends (notification is dismissed; microphone indicator disappears)
  - [ ] Pixel `m_aichat_voice_service_killed` IS fired exactly once shortly after, as the foreground service is destroyed                                                                                            
                                                                                                                                                                                                                     
  ### 4. Voice session ended from the Duck.ai UI (not the notification)                                                                                                                                              
  - [ ] Start a voice session, then end it from inside the Duck.ai chat UI (not via the notification action)                                                                                                         
  - [ ] Pixel `m_aichat_voice_service_killed` IS fired exactly once                                                                                                                                                  
  - [ ] Pixel `m_aichat_voice_notification_end_chat_tapped` is **NOT** fired (this pixel is specific to the notification action)                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                           

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Low Risk**
> Telemetry-only changes with no auth, data, or session logic changes beyond firing pixels at existing lifecycle and notification entry points.
> 
> **Overview**
> Adds **analytics pixels** for Duck.ai **voice** foreground service lifecycle and **notification** actions that were previously missing.
> 
> Four new metrics are registered in pixel definitions and wired through `DuckChatPixels`: **`m_aichat_voice_service_started`** / **`m_aichat_voice_service_killed`** fire when the microphone foreground service is created and destroyed; **`m_aichat_voice_notification_end_chat_tapped`** fires when the End Chat broadcast is handled; **`m_aichat_voice_notification_open_chat_tapped`** is passed on the Open Chat activity intent and fired via existing `BrowserActivity` / `onLaunchedFromNotification` handling (including the new **`onNewIntent`** path for warm resume).
> 
> Receiver unit tests now assert the end-chat pixel fires only on valid end-session intents.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbf8a0fb10b027d7b558189ebe10727e390bf80b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/pixel/DuckChatPixels.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/pixel/DuckChatPixels.kt
@@ -114,7 +114,11 @@ import kotlinx.coroutines.withContext
 import javax.inject.Inject
 
 interface DuckChatPixels {
-    fun sendReportMetricPixel(reportMetric: ReportMetric, modelTier: ModelTier? = null)
+    fun sendReportMetricPixel(
+        reportMetric: ReportMetric,
+        modelTier: ModelTier? = null,
+    )
+
     fun reportOpen()
     fun reportContextualSheetOpened()
     fun reportContextualSheetDismissed()
@@ -146,7 +150,7 @@ interface DuckChatPixels {
     fun reportNativeStorageDeletionUsed(native: Boolean)
     fun reportVoiceSessionStarted()
     fun reportVoiceNotificationOpenChatTapped()
-    fun reportVoiceNotificationEndSessionTapped()
+    fun reportVoiceNotificationEndChatTapped()
     fun reportVoiceServiceStarted()
     fun reportVoiceServiceKilled()
 }
@@ -162,7 +166,10 @@ class RealDuckChatPixels @Inject constructor(
     private val termsOfServiceHandler: DuckChatTermsOfServiceHandler,
 ) : DuckChatPixels {
 
-    override fun sendReportMetricPixel(reportMetric: ReportMetric, modelTier: ModelTier?) {
+    override fun sendReportMetricPixel(
+        reportMetric: ReportMetric,
+        modelTier: ModelTier?,
+    ) {
         appCoroutineScope.launch(dispatcherProvider.io()) {
             var refreshAtb = false
             val sessionParams = mapOf(
@@ -173,10 +180,12 @@ class RealDuckChatPixels @Inject constructor(
                     refreshAtb = true
                     DUCK_CHAT_SEND_PROMPT_ONGOING_CHAT to sessionParams
                 }
+
                 USER_DID_SUBMIT_FIRST_PROMPT -> {
                     refreshAtb = true
                     DUCK_CHAT_START_NEW_CONVERSATION to sessionParams
                 }
+
                 USER_DID_OPEN_HISTORY -> DUCK_CHAT_OPEN_HISTORY to sessionParams
                 USER_DID_SELECT_FIRST_HISTORY_ITEM -> DUCK_CHAT_OPEN_MOST_RECENT_HISTORY_CHAT to sessionParams
                 USER_DID_CREATE_NEW_CHAT -> DUCK_CHAT_START_NEW_CONVERSATION_BUTTON_CLICKED to sessionParams
@@ -405,11 +414,8 @@ class RealDuckChatPixels @Inject constructor(
         pixel.fire(DuckChatPixelName.DUCK_CHAT_VOICE_NOTIFICATION_OPEN_CHAT_TAPPED)
     }
 
-    override fun reportVoiceNotificationEndSessionTapped() {
-        pixel.fireCountAndDaily(
-            DuckChatPixelName.DUCK_CHAT_VOICE_NOTIFICATION_END_SESSION_TAPPED_COUNT,
-            DuckChatPixelName.DUCK_CHAT_VOICE_NOTIFICATION_END_SESSION_TAPPED_DAILY,
-        )
+    override fun reportVoiceNotificationEndChatTapped() {
+        pixel.fire(DuckChatPixelName.DUCK_CHAT_VOICE_NOTIFICATION_END_CHAT_TAPPED)
     }
 
     override fun reportVoiceServiceStarted() {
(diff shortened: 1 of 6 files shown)


karlenDimla:
* **#8839** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8839?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8839?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/pixel/DuckChatPixels.kt
```diff
@@ -145,6 +149,10 @@ interface DuckChatPixels {
     fun reportNativeStorageReaderUsed(native: Boolean)
     fun reportNativeStorageDeletionUsed(native: Boolean)
     fun reportVoiceSessionStarted()
+    fun reportVoiceNotificationOpenChatTapped()
```
malmstein: `reportVoiceNotificationOpenChatTapped()` is added to the interface and implemented, but nothing calls it — the open-chat pixel actually fires through the string-name path (`service putExtra → BrowserViewModel.onLaunchedFromNotification → pixel.fire(name)`). the other three go through typed methods, so this one reads like it should too. shouldn't we either wire it or drop the unused method to avoid the impression this is the firing path?

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/DuckChatVoiceMicrophoneService.kt
```diff
@@ -143,6 +156,9 @@ class DuckChatVoiceMicrophoneService : Service() {
         private const val REQUEST_CODE_FALLBACK = 3
         private const val EXTRA_TAB_ID = "EXTRA_TAB_ID"
 
+        // Must match BrowserActivity.LAUNCH_FROM_NOTIFICATION_PIXEL_NAME — :app reads this extra to fire a pixel on resume.
+        private const val LAUNCH_FROM_NOTIFICATION_PIXEL_NAME_EXTRA = "LAUNCH_FROM_NOTIFICATION_PIXEL_NAME"
```
malmstein, line 160: this key has to stay byte-identical to `BrowserActivity.LAUNCH_FROM_NOTIFICATION_PIXEL_NAME`, and the only thing keeping them in sync is the comment — if either drifts the pixel silently stops firing. since impl can't see `:app`, could the key live in a shared place (browser-api, wherever the launch-from-notification contract belongs) so both sides reference one constant?

malmstein (approved):
LGTM, approving. did not install or smoke-test the APK — static-only. solid pixel coverage addition: all four pixels are defined in the json5, the enum, and the `removeAtb` param-removal plugin in lockstep, and the service/receiver lifecycle hooks are sensible (service started/killed in onCreate/onDestroy, end-chat in the receiver). good call extending the existing notification-launch infra to `onNewIntent` so warm-resume taps are covered too, and the end-chat receiver test was updated nicely including the negative `verifyNoInteractions` paths.

one thing worth resolving before merge: `reportVoiceNotificationOpenChatTapped()` is dead — the open-chat pixel fires via the string path, so the typed method should be wired or removed. and the duplicated extra key is a quiet drift risk. nothing rule-level. since I didn't run it, worth confirming on a device that the open-chat notification tap actually emits the pixel on both cold and warm resume.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/pixel/DuckChatPixels.kt
karlenDimla: Ah actually the pixel is being emitted through the BrowserActivity’s support for firing pixels when opened through notifications. I will remove this method.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/DuckChatVoiceMicrophoneService.kt
karlenDimla: Will address on a separate PR since that would require changes on other classes too. But good point.

PR #8877, opened 2026-06-15
Preserve per-site desktop mode preferences

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214749231578703/task/1212541545715038?focus=true
API Proposal: https://app.asana.com/1/137249556945/project/1207418217763355/task/1215944643003573?focus=true

### Description

This PR adds a per-site desktop mode preference persistence using eTLD+1 as a key (or "localhost" or an IP address). When a desktop mode is enabled on a site, it will be always used until it's either disabled or the site is cleared using the Fire button.

### Steps to test this PR

- [x] Load some URL
- [x] Enable the Desktop mode
- [x] Notice the site is reloaded in a desktop mode
- [x] Open a new tab and load the same site
- [ ] Notice it loads using the desktop mode
- [x] Use the fire button and delete this tab
- [x] On a NTP, load the same site
- [x] Notice it now loads in a mobile mode

---

> [!NOTE]
> **Medium Risk**
> Touches WebView UA selection, navigation/reload paths, and data-clearing/fireproof behavior; incorrect site-key or cache priming could load wrong mode or drop preferences on burn.
> 
> **Overview**
> Adds **site-preferences** (`site-preferences-api` / `site-preferences-impl`) with Room-backed storage and a **`rememberDesktopMode`** remote feature toggle. When enabled, toggling “Desktop site” **remembers or forgets** the choice per site key (eTLD+1, or raw host for IPs/`localhost` via new **`toTldPlusOneOrSelf`**).
> 
> The browser **reads persisted mode** on navigation and in **`WebViewRequestInterceptor`** (`isDesktopSiteEnabled(url)`), **resets UI per domain** on `pageChanged`, and **full-reloads (Navigate)** on user refresh when stored mode drifted (e.g. another tab). Desktop toggle logic is refactored into **`reloadInMode`**.
> 
> **Fire / data clearing** wires **`SitePreferencesDataClearer`**: full clears keep fireproofed sites; per-domain burns clear preferences for non-fireproof domains, with fireproof matching aligned to the same host-fallback keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5cb15853c443242cfb03419aab02e25ee0eb1b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/src/main/java/com/duckduckgo/app/global/view/ClearPersonalDataAction.kt
```diff
@@ -262,6 +265,18 @@ class ClearPersonalDataAction(
...
 
+    // Shared preamble for full-fire paths: clears cookies and the per-site stores that must keep
+    // fireproofed domains. Site permissions match on raw host; site preferences on eTLD+1.
+    private suspend fun clearFireproofExemptData() {
+        val fireproofEntities = fireproofWebsiteRepository.fireproofWebsitesSync()
+        cookieManager.flush()
+        sitePermissionsManager.clearAllButFireproof(fireproofEntities.map { it.domain })
+        sitePreferencesRepository.clearAllButFireproofed(
+            fireproofEntities.mapNotNull { it.domain.toTldPlusOne() }.toSet(),
+        )
```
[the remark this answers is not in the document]
0nko: Fixed in 51dde37765. The full-fire keep-list now uses the same site-key derivation as persistence — toTldPlusOne() ?: host — so a fireproofed host without a registrable domain (IP, localhost) keeps its remembered desktop preference through the fire button, matching the key it was stored under.

app/src/main/java/com/duckduckgo/app/browser/sitepreferences/SitePreferencesRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+                "example.com".toTldPlusOne()
+                sitePreferencesDao.desktopModeDomainsFlow().collect { domains ->
+                    desktopModeDomains.clear()
+                    desktopModeDomains.addAll(domains)
+                }
+            }
+        }
+    }
+
+    override fun isDesktopModeRemembered(domain: String): Boolean = desktopModeDomains.contains(domain)
```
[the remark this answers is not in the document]
0nko: Addressed in 20ab42e85b. RealSitePreferencesRepository now implements MainProcessLifecycleObserver and primes the in-memory cache from the DB in onCreate (on the IO dispatcher), so it’s loaded at app startup before the first navigation rather than lazily on first read. The prime runs off the main thread (and warms the public-suffix DB there too). Worst case — a navigation so early the prime hasn’t finished — still fails safe to mobile and self-corrects on the next navigation.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3309,6 +3335,20 @@ class BrowserTabViewModel @Inject constructor(
 
         val uri = site?.uri ?: return
 
+        // Persist the choice per-site. uri.host is the pre-rewrite host (e.g. m.example.com), whose eTLD+1
+        // (example.com) matches the key the desktop-rewritten URL produces; hosts without a registrable
+        // domain (IPs, localhost) fall back to the raw host. The optimistic cache update means the reload
+        // triggered below already sees the new value.
```
landomen: nit: I'd recommend shorter comments or even skip them if code is clear

app/src/main/java/com/duckduckgo/app/global/db/AppDatabase.kt
```diff
@@ -118,6 +120,7 @@ import com.duckduckgo.savedsites.store.SavedSitesRelationsDao
         Entity::class,
         Relation::class,
         DefaultBrowserPromptsAppUsageEntity::class,
+        SitePreferencesEntity::class,
```
landomen: Have you maybe considered creating a new module for site-preferences and a separate database? ;) 

Asking because this relates to our Modularization AOI and adding new features that use AppDatabase will make it harder later to remove them. I'm not saying you need to do the new module now, just something to keep in mind.
(cc @aibrahim- )

landomen (commented):
@0nko It happened multiple times that I had to force refresh the site to re-render in desktop mode after opening it. See video:

https://github.com/user-attachments/assets/91b74c8e-ac19-4062-af5c-7d89aec5c022

app/src/main/java/com/duckduckgo/app/global/db/AppDatabase.kt
aibrahim-: Do we really need a room DB for this? it seems a simple preferences store should be enough, e.g. `androidx.datastore` or even a SharedPreferences

0nko: @aibrahim- It seemed like a natural choice, since the settings is per-domain, which can grow indefinitely. Plus, this would allow to store other prefs per-domain in the future. But I supposed storing it as a list of domains in a DataStore would work if the list was reasonable.

0nko: @landomen Yeah, I suppose it could be a separate module. I can refactor it.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
0nko: I'll let Claude know.. 😁

0nko:
@landomen @aibrahim- I moved the site preferences to a separate module and tried to address the issue you noticed. It is now better but a site set to desktop mode occasionally still renders mobile because the desktop-UA reload is served from the HTTP cache, which can return a previously-fetched mobile copy (a forced refresh fixes it). This is a pre-existing limitation not introduced here, it is just easier to surface now that desktop mode is applied automatically on every initial load of a remembered site.

landomen (commented):
@0nko Thanks for moving this to new module. I like the new API but just to follow the process, I think we need an official API proposal for this 😅

0nko:
@landomen Thanks! For sure, just wanted you to take a look first to make sure it made sense.

Proposal: https://app.asana.com/1/137249556945/project/1207418217763355/task/1215944643003573?focus=true

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -2078,6 +2106,13 @@ class BrowserTabViewModel @Inject constructor(
         cleanupBlobDownloadReplyProxyMaps(url)
 
         hasCtaBeenShownForCurrentPage.set(false)
+
+        if (rememberDesktopModeFeature.self().isEnabled()) {
+            browserViewState.value = currentBrowserViewState().copy(
+                isDesktopBrowsingMode = isDesktopModeRememberedForUrl(url),
+            )
```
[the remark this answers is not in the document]
0nko: This is the known cold-start priming window, and the impact is narrower than the description suggests.

- The per-request UA decision (`isDesktopSiteEnabled` → suspend `isDesktopModeRemembered`) falls back to a direct DB read while the cache is priming, so the main document is correctly requested with the **desktop** UA even on a cold start.
- `pageChanged` uses the synchronous cache-only read because it runs on the main thread, where Room forbids a DB query — so it can read `false` during the brief priming window.
- The progress==100 `RefreshUserAgent` then sets the UA string to mobile, but `refreshUserAgent()` only updates `WebSettings.userAgentString`; it does **not** reload. So the already-loaded desktop page is not reverted — only the menu label and UA string are transiently inconsistent, and both self-correct on the next navigation once the cache primes.

This is the intended tradeoff for the synchronous main-thread read (covered in the tech design), and the feature is behind the `rememberDesktopMode` kill-switch. Leaving as-is.

site-preferences/site-preferences-impl/src/main/java/com/duckduckgo/site/preferences/impl/RealSitePreferencesRepository.kt
```diff
@@ -0,0 +1,107 @@
...
+) : DesktopModeSettings, SitePreferencesDataClearer, MainProcessLifecycleObserver {
+
+    private val desktopModeDomains = CopyOnWriteArraySet<String>()
+
+    @Volatile private var cachePrimed = false
+
+    override fun onCreate(owner: LifecycleOwner) {
+        if (!isMainProcess) return
+        appCoroutineScope.launch(dispatcherProvider.io()) {
+            "example.com".toTldPlusOne()
```
landomen, line 60: nit: it would be good to have a comment here that this is about pre-warming the list

site-preferences/site-preferences-impl/src/main/java/com/duckduckgo/site/preferences/impl/RealSitePreferencesRepository.kt
```diff
@@ -0,0 +1,107 @@
...
+            sitePreferencesDao.desktopModeDomainsFlow().collect { domains ->
+                desktopModeDomains.clear()
+                desktopModeDomains.addAll(domains)
```
landomen: Could it happen that another read happens on the main thread between the `clear` and `addAll` calls, falsely reporting desktop mode is not enabled?

site-preferences/site-preferences-impl/src/main/java/com/duckduckgo/site/preferences/impl/store/SitePreferencesDao.kt
```diff
@@ -0,0 +1,48 @@
...
+    fun deleteAllExcept(domains: Set<String>)
+
+    @Query("SELECT domain FROM site_preferences WHERE desktopModeEnabled = 1")
+    fun desktopModeDomainsFlow(): Flow<List<String>>
+
+    @Query("SELECT EXISTS(SELECT 1 FROM site_preferences WHERE domain = :domain AND desktopModeEnabled = 1)")
+    fun isDesktopModeEnabled(domain: String): Boolean
+
+    @Query("DELETE FROM site_preferences")
+    fun deleteAll()
```
landomen: Is this used anywhere?

landomen (approved):
LGTM, just left a few minor comments

site-preferences/site-preferences-impl/src/main/java/com/duckduckgo/site/preferences/impl/RealSitePreferencesRepository.kt
0nko: Probably unlikely, but I updated the code to make sure.

site-preferences/site-preferences-impl/src/main/java/com/duckduckgo/site/preferences/impl/store/SitePreferencesDao.kt
0nko: No, removed.

site-preferences/site-preferences-impl/src/main/java/com/duckduckgo/site/preferences/impl/RealSitePreferencesRepository.kt
0nko: Done.

PR #8882, opened 2026-06-15
Fix app startup ANRs

CrisBarreiro:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1215718265244661?focus=true 

### Description
Fixes ANRs caused by building a moshi adapter on the main thread during app init 

### Steps to test this PR
qa-optional

---

> [!NOTE]
> **Low Risk**
> Threading-only change for startup and config refresh with the same parsing logic; risk is mainly timing of when `scriptletsSettings` first emits relative to other startup observers.
> 
> **Overview**
> Moves **ad-blocking extension settings parsing** off the main thread during app startup to address ANRs from building Moshi adapters and parsing JSON on init.
> 
> `RealAdBlockingExtensionConfigProvider` no longer refreshes in its constructor; it implements `MainProcessLifecycleObserver` and kicks off `refresh()` from `onCreate`, with parsing on `dispatchers.io()` and in-flight work cancelled when privacy config updates arrive again. `ScriptletDownloadWorkerScheduler` now starts its `combine`/`collect` loop on the IO dispatcher as well.
> 
> Unit tests were updated to trigger `onCreate`, inject test coroutine scopes/dispatchers, and use `CoroutineTestRule` instead of ad-hoc test scopes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6f96314e3b599beae5e825a191a07d336714729. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/remoteconfig/AdBlockingExtensionConfigProvider.kt
+++ b/ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/remoteconfig/AdBlockingExtensionConfigProvider.kt
@@ -16,6 +16,8 @@
 
 package com.duckduckgo.adblocking.impl.remoteconfig
 
+import com.duckduckgo.app.di.AppCoroutineScope
+import com.duckduckgo.common.utils.DispatcherProvider
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.privacy.config.api.PrivacyConfigCallbackPlugin
 import com.squareup.anvil.annotations.ContributesBinding
@@ -25,9 +27,11 @@ import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import dagger.SingleInstanceIn
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
 import logcat.LogPriority.WARN
 import logcat.asLog
 import logcat.logcat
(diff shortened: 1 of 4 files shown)


CrisBarreiro:
* **#8882** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8882?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8882?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/remoteconfig/AdBlockingExtensionConfigProvider.kt
```diff
@@ -56,20 +60,22 @@ interface AdBlockingExtensionConfigProvider {
...
+    @AppCoroutineScope private val appScope: CoroutineScope,
+    private val dispatchers: DispatcherProvider,
 ) : AdBlockingExtensionConfigProvider, PrivacyConfigCallbackPlugin {
 
     private val settingsAdapter by lazy { buildJsonAdapter() }
     private val scriptletsFlow = MutableStateFlow<AdBlockingExtensionSettings?>(null)
 
     init {
-        refresh()
+        appScope.launch(dispatchers.io()) { refresh() }
```
cmonfortep: what if we do this from app lifecycle oncreate? that way we are 100% out of putting logic inside `init` blocks?

CrisBarreiro: No strong opinions, I think that way this will execute very slightly after `init`, but now what we're launching a coroutine, this shouldn't be an issue anymore, so I don't see the benefit of coupling the class to the Android lifecycle

cmonfortep: By default I try to avoid using `init` as it's a coupling logic to object/class creation and brings it's own risks. For example `init` position matters, which is also easy to miss. App lifecycle provides the same without the risks and also makes it part of knowing the app and graph has been created.

cmonfortep (approved):
LGTM

Left a comment about avoiding `init` in favor of app lifecycle for you to consider.

PR #8861, opened 2026-06-12
Add brand-design Duck.AI input-screen end CTA

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205648422731273/task/1215654188045007?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/1205648422731273/task/1215654373627388?focus=true (InputScreenActivityParams API proposal)

### Description

Adds a brand-design-update variant of the Duck.AI onboarding **end** CTA
(`CtaId.DAX_DUCK_AI_END`) rendered on the Input Screen.

The legacy end CTA lives in `InputScreenFragment` with its own layout (not as a
contextual `OnboardingDaxDialogCta`); the redesigned variant follows the same model —
a parallel include in `fragment_input_screen.xml`, picked at show-time.

Changes:
- **Public API** (`duckchat-api`, see [API proposal](https://app.asana.com/1/137249556945/project/1205648422731273/task/1215654373627388?focus=true)):
  `InputScreenActivityParams.showDuckAiOnboardingEndCta: Boolean` is replaced by
  `duckAiOnboardingEndCta: DuckAiOnboardingEndCtaVariant` (new enum:
  `NONE` / `LEGACY` / `BRAND_DESIGN_UPDATE`), so the launch path carries the three-way
  state explicitly instead of a sibling boolean. `Command.LaunchInputScreen` carries the
  same variant.
- `CtaViewModel.prepareAndMarkDuckAiEndCtaForInputScreen()` now returns the variant,
  choosing `BRAND_DESIGN_UPDATE` when the Duck.AI brand-design FF
  (`onboardingDuckAiExperimentMay26BrandDesignUpdate` + `onboardingBrandDesignUpdate`)
  is on, otherwise `LEGACY`.
- `InputScreenFragment` renders the matching variant: the `BRAND_DESIGN_UPDATE` variant
  uses a flat `?attr/onboardingSurfaceBackdrop` colour with the bottom-anchored
  `bg_onboarding_end` image layered on top, matching the brand-design home tab; `LEGACY`
  keeps the original onboarding bitmap.
- The `bg_onboarding_end` webps move from `:app` to `:android-design-system` so
  `duckchat-impl` can reference them (existing brand-design end CTAs now point at the
  shared `CommonR` drawable). The `onboardingBackgroundMaxHeight` dimen stays module-local:
  `:app` keeps its copy (used by `include_new_browser_tab.xml`) and `:duckchat-impl` gets
[...]

PR #8857, opened 2026-06-12
Add brand-design Duck.AI fire button onboarding dialog

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205648422731273/task/1215626471231067?focus=true
Tech Design URL (if applicable):

### Description

Adds a brand-design-update variant of the Duck.AI in-context **fire button** onboarding
CTA (`CtaId.DAX_DUCK_AI_FIRE_BUTTON`).

The new CTA mirrors `DaxFireButtonBrandDesignUpdateContextualCta` but preserves the legacy
Duck.AI UX: no in-card primary button and no dismiss button — the user dismisses by tapping
the actual fire button.

Supporting changes:
- `BrandDesignContextualDaxDialogCta` now supports CTAs that render **title + description
  only**: `activeIncludeId` becomes optional (`open val … = 0`) and the render pipeline
  gracefully skips slot work when no content slot is present. A new empty content-include
  slot (`contextualBrandDesignNoCtaContent`) and a `showDismiss` flag (default `true`)
  back this.
- Four `BrowserTab` call sites that drove fire-button behaviour (pulse highlight,
  `isDuckAiOnboarding` flag, `fireFireButtonPressed` metric, dismissal) only matched the
  legacy `DaxDuckAiFireButtonCta`; they now treat the brand-design variant equivalently.
- The brand-design variant is gated behind a **new** `onboardingDuckAiExperimentMay26BrandDesignUpdate`
  toggle (default `INTERNAL`).

### Steps to test this PR
_Prerequisites_
- [x] Apply patch included in asana task 

_Brand-design Duck.AI fire-button dialog_
- [x] Run the Duck.AI onboarding flow until the in-context fire-button CTA appears; confirm
      it renders in the **brand-design** style (title + description, no in-card button, no
      dismiss button).
- [x] Tap the fire button the arrow points at; confirm the fire flow runs and the CTA
      dismisses.

_Legacy fallback_
- [ ] The toggle defaults to `INTERNAL` (on in internal builds). Override
      `onboardingDuckAiExperimentMay26BrandDesignUpdate` to disabled; confirm the
      **legacy** Duck.AI fire-button CTA renders instead.

### UI changes
| Before  | After |
| ------ | ----- |
Screenshot_20260612_110935|Screenshot_20260612_105603|

---

> [!NOTE]
> **Low Risk**
> Onboarding UI and CTA selection only; legacy path remains when brand-design is off, with broad test coverage for parity.
> 
> **Overview**
[...]

PR #8888, opened 2026-06-15
Make Duck.ai onboarding compatible with native input field

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1205648422731273/task/1215660165459344?focus=true

### Description

The Duck.ai onboarding flow (the in-context fire-button CTA and the end CTA) was built for the
classic omnibar/input-screen layout. When the `duckChat.nativeInputField` feature is enabled, the
fire button and chat controls move into the bottom native-input widget and the omnibar shows
different chrome, which broke the onboarding experience. This PR makes the flow work correctly in
that mode.

Changes:

- **End CTA as a home bubble.** With `nativeInputField` on, the legacy `InputScreenActivity` path
  no longer fires, so the `DAX_DUCK_AI_END` CTA is rendered as a home-tab bubble on the NTP instead
  (legacy + brand-design variants), and dismisses correctly on the "High Five" tap.
- **Fire-button CTA highlight.** The fire button now lives in the native-input widget, so the
  onboarding pulse is applied there instead of the (now-hidden) omnibar fire icon
  (`setFireButtonSpotlightMode`).
- **Lock everything but the fire button during the fire CTA.** The native-input widget is dimmed and
  made non-interactive except for the fire button, and the omnibar controls that aren't covered by
  the existing onboarding lock are disabled too: the new-chat **+** icon, the **Free plan** pill, and
  the Duck.ai **close (X)** button.
- **Positioning.** The end-CTA bubble is offset so the expanded native-input widget no longer covers it.

Legacy behaviour (with `nativeInputField` off) is unchanged.

### Steps to test this PR

Apply this diff to make the duck.ai onboarding path available. The native input field is already
enabled in internal builds, so no other change is needed.

```diff
 override suspend fun enroll(): DuckAiOnboardingExperimentVariant? = withContext(dispatcherProvider.io()) {
     if (!arePrerequisitesMet()) return@withContext null

+    return@withContext DuckAiOnboardingExperimentVariant.TREATMENT_WITH_SEARCH_DEFAULT
+
     val toggle = extendedOnboardingFeatureToggles.onboardingDuckAiExperimentMay26()

 private fun arePrerequisitesMet(): Boolean =
     !appBuildConfig.isDefaultVariantForced &&
         browserConfig.singleTabFireDialog().isEnabled() &&
-        !onboardingBrandDesignUpdateToggles.brandDesignUpdate().isEnabled() &&
         !deviceInfo.isTablet()
```

_Fire-button CTA_
- [x] Run the Duck.ai onboarding flow and navigate to Duck.ai so the fire-button CTA appears
- [x] The fire button **in the native-input widget** pulses (not the omnibar)
- [x] The rest of the input widget is dimmed and unresponsive; the **+**, **Free plan**, and **X** buttons are dimmed and do nothing; only the fire button works
- [x] The input-bar background stays colour-uniform with the Duck.ai page (no darkened bar)
- [x] Tapping the fire button proceeds with the fire animation

_End CTA_
- [x] After the fire op, the end CTA appears as a **bubble on the NTP** (not a separate input screen)
- [x] The bubble is fully visible and not covered by the native-input widget
- [x] Toggle the input between **Search** and **Duck.ai** and confirm the bubble repositions to stay clear of the widget as it resizes
- [x] Tapping **"High Five"** dismisses the bubble and completes onboarding

_Regression_
- [ ] Ensure the `duckChat.nativeInputField` flag is disabled, then do a clean install
- [ ] Confirm the legacy onboarding / input-screen flow is unchanged

### UI changes
| Before  | After |
| ------ | ----- |
|(Upload before screenshot)|(Upload after screenshot)|

---

> [!NOTE]
> **Medium Risk**
> Touches onboarding gating, back navigation, and native-input interaction during a guided flow; legacy paths are feature-flagged but the new bubble and lock wiring spans browser, CTA, and duckchat modules.
> 
> **Overview**
> Makes Duck.ai onboarding work when the **native input field** is on instead of the legacy input screen.
> 
> **End CTA on the NTP:** If `DuckAiFeatureState.showInputScreen` is false, `DAX_DUCK_AI_END` is shown as home-tab bubbles (`DaxDuckAiEndBubbleCta` / brand-design variant) with the same impression and completion side effects as the input-screen path; custom-AI copy moves off generic `DaxEndBrandDesignUpdateBubbleCta`. Legacy input-screen mode still suppresses home CTAs until that flow runs.
> 
> **Locked demo UX:** Browser view state drives `NativeInputState.InteractionLock` and fire-button highlight on the widget (full lock vs lock except leading fire). Back is blocked while omnibar onboarding lock is active. Omnibar also disables **+**, **Free plan**, and Duck.ai **back** during lock; fire pulse targets visible fire chrome in Duck.ai/native-input mode.
> 
> **Layout:** `isLogoOnly` treats visible onboarding CTA bubbles as non-logo-only so NTP content offsets below the taller widget and bubbles stay visible.
> 
> `PulseAnimation` moves to `browser-ui`; pixel CTA params centralize in `Pixel.PixelValues`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54861131f08ac75b38be52d214f56f7c98f5a234. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmac012:
* **#8888** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8888?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8888?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8861** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8861?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8857** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8857?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
```diff
@@ -1562,7 +1564,10 @@ class OmnibarLayout @JvmOverloads constructor(
         applyEnabled(clearTextButton, nonFireEnabled)
         applyEnabled(duckAISidebar, nonFireEnabled)
         applyEnabled(duckAIHeader, nonFireEnabled)
+        applyEnabled(duckAIFreePill, nonFireEnabled)
```
LukasPaczos, line 1567: Not a blocker but something I wanted to consider - any ideas how we could make this more robust so that it doesn't regress in the future? Instead of disabling each view independently (and people contributing new UI elements would need to remember to do so), could we apply a mask with semi-transparent foreground on the whole container and capture clicks so that they are not passed down to children? Or something similar?

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5953,6 +5953,7 @@ class BrowserTabFragment :
 
                 browserNavigationBarIntegration.configureFireButtonHighlight(highlighted = viewState.fireButton.isHighlighted())
                 browserNavigationBarIntegration.configureLockForOnboarding(locked = viewState.isOmnibarLockedForOnboarding)
+                nativeInputManager.setFireButtonSpotlightMode(viewState.fireButton.isHighlighted())
```
LukasPaczos: We don't need to check `viewState.isOmnibarLockedForOnboarding` here?

app/src/main/java/com/duckduckgo/app/cta/ui/DaxDuckAiEndBrandDesignUpdateBubbleCta.kt
```diff
@@ -0,0 +1,66 @@
...
+    override val isLightTheme: Boolean,
+    override val deviceInfo: DeviceInfo,
+) : DaxBubbleCta.BrandDesignUpdateBubbleCta(
+    ctaId = CtaId.DAX_DUCK_AI_END,
+    title = R.string.onboardingDuckAiEndCtaTitle,
+    description = R.string.onboardingDuckAiEndCtaDescription,
+    backgroundRes = CommonR.drawable.bg_onboarding_end,
+    shownPixel = AppPixelName.ONBOARDING_DAX_CTA_SHOWN,
+    okPixel = AppPixelName.ONBOARDING_DAX_CTA_OK_BUTTON,
+    ctaPixelParam = "duck_ai_end_cta",
```
LukasPaczos: The original end CTA uses two different pixels for in-browser and NTP end CTAs, but here we use the same pixel for both. Should we use separate ones? Also, could you move it to a constant?

app/src/main/java/com/duckduckgo/app/cta/ui/DaxDuckAiEndBubbleCta.kt
```diff
@@ -0,0 +1,60 @@
...
+    override val onboardingStore: OnboardingStore,
+    override val appInstallStore: AppInstallStore,
+) : DaxBubbleCta(
+    ctaId = CtaId.DAX_DUCK_AI_END,
+    title = R.string.onboardingDuckAiEndCtaTitle,
+    description = R.string.onboardingDuckAiEndCtaDescription,
+    primaryCta = R.string.onboardingDuckAiEndCtaButton,
+    shownPixel = AppPixelName.ONBOARDING_DAX_CTA_SHOWN,
+    okPixel = AppPixelName.ONBOARDING_DAX_CTA_OK_BUTTON,
+    ctaPixelParam = "duck_ai_end_cta",
```
LukasPaczos: Same as above.

LukasPaczos (commented):
On a Pixel 9 Pro, in landscape, the CTA takes so much space that you can't see or scroll to reveal the response. Could we decrease the empty space at the bottom a little bit?
image

LukasPaczos:
Using back button on the Duck.ai demo screen breaks the flow:

https://github.com/user-attachments/assets/2573c857-e3de-4c3b-b55a-d558371e93f8

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayout.kt
lmac012: Makes sense, I can look into it while working on the follow up task to hide `duckAIFreePill` entirely. As the requirements kept evolving, it wasn't clear we'd end up disabling everything, so I initially went with a simple yet flexible approach.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
lmac012: For the most part `viewState.isOmnibarLockedForOnboarding` is equal to  `viewState.fireButton.isHighlighted()`. However, there is a tiny time window at the beginning before we start showing the fire button CTA + highlight, where we currently don't lock the native input field, but ideally we should, so I will look into improving this.

app/src/main/java/com/duckduckgo/app/cta/ui/DaxDuckAiEndBrandDesignUpdateBubbleCta.kt
lmac012: Unlike the original CTA, the duck ai end CTA is only shown in the context of NTP. I'm ignoring `InputScreen`, since that path is going to be disabled anyway. 

I will extract this to a constant (it's not great that other CTAs keep these constants in the `statistics-api` module, but we can prioritize consistency here).

lmac012:
Thanks for the review @LukasPaczos, I pushed a few commits to address (most of) your feedback. For the remaining items (CTA spacing in landscape orientation, disabling views in bulk) - I propose dealing with them as follow-ups.

The landscape issue seems to be shared with existing (reskinned) onboarding CTAs (but it is not as visible unless address bar is set to split). Ideally this should be solved adjusting the scrolling behavior (make CTAs scroll with the page content), but I'll have look if we can reduce the empty space below this particular CTA.

The back button is definitely a blocker, thanks for catching that. Fortunately, the fix seems to be simple.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -5953,6 +5957,13 @@ class BrowserTabFragment :
 
                 browserNavigationBarIntegration.configureFireButtonHighlight(highlighted = viewState.fireButton.isHighlighted())
                 browserNavigationBarIntegration.configureLockForOnboarding(locked = viewState.isOmnibarLockedForOnboarding)
+                nativeInputManager.setEnabledState(
+                    when {
+                        !viewState.isOmnibarLockedForOnboarding -> EnabledState.ENABLED
+                        viewState.fireButton.isHighlighted() -> EnabledState.FIRE_BUTTON_SPOTLIGHT
+                        else -> EnabledState.DISABLED
+                    },
+                )
```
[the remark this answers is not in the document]
lmac012: Cursor doesn't get the fact that NTP is opened in a new tab which has `isOmnibarLockedForOnboarding` set to false.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
```diff
@@ -4993,6 +4994,9 @@ class BrowserTabFragment :
 
     fun onBackPressed(isCustomTab: Boolean = false): Boolean {
         if (!isAdded) return false
+        // During the locked Duck.ai onboarding demo, don't navigate within the app — returning false
+        // lets BrowserActivity exit (close the app). The user progresses by tapping the fire button.
+        if (viewModel.isOmnibarLockedForOnboarding()) return false
```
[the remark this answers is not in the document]
lmac012: `The flag (...) is not cleared when onboarding finishes` - this is not true, the fire effectively clears the flag because we open a new tab and tear down the previous one.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -5130,6 +5134,8 @@ class BrowserTabViewModel @Inject constructor(
             }
             is DaxBubbleCta.DaxEndCta,
             is DaxEndBrandDesignUpdateBubbleCta,
+            is DaxDuckAiEndBubbleCta,
+            is DaxDuckAiEndBrandDesignUpdateBubbleCta,
```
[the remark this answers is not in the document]
lmac012: `onDuckAiEndCtaInteraction()` is only meant to be called in the input-screen path (when native input is disabled). Bubble implementation has a generic handling of ok/dismiss pixels.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabFragment.kt
LukasPaczos, line 5042: This was flagged as a known issue for the initial experiment, but after the app closes and restarts here, we end up in an unlocked state, so user can interact with Duck.ai or even close it while the fire animation plays etc.

Longer term, I think this flow should move to the linear onboarding orchestrator, which forces the Duck.ai browser CTAs into a linear flow and handles aborting correctly (restarts from scratch).

Either way, since we already deemed this an edge case, it's not a blocker here.

LukasPaczos (approved):
LGTM

LukasPaczos:
btw. I tested with forced native input, as it's rolling out now:
```diff
diff --git a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/feature/DuckChatFeature.kt b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/feature/DuckChatFeature.kt
index 151e8b0c21..6043dee436 100644
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/feature/DuckChatFeature.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/feature/DuckChatFeature.kt
@@ -142,6 +142,7 @@ interface DuckChatFeature {
      * @return `true` when the Native Input Field should be used instead of the web-based input.
      */
     @Toggle.DefaultValue(DefaultFeatureValue.INTERNAL)
+    @InternalAlwaysEnabled
     fun nativeInputField(): Toggle
```

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
```diff
@@ -120,6 +125,16 @@ interface NativeInputManager {
...
+
+    /** Apply an [EnabledState] to the widget. No-op outside Duck.ai mode. */
+    fun setEnabledState(state: EnabledState)
+
+    /**
+     * Which of the widget's controls are interactive: [EnabledState.ENABLED] all,
+     * [EnabledState.DISABLED] none, [EnabledState.FIRE_BUTTON_SPOTLIGHT] only the fire button
+     * (highlighted with a pulse).
+     */
+    enum class EnabledState { ENABLED, DISABLED, FIRE_BUTTON_SPOTLIGHT }
```
malmstein: why is this here? There’s a `NativeInputState` that represents the current state. Please use the existing pattern, there’s no need to duplicate it for the sake of onboarding.

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
```diff
@@ -673,13 +692,47 @@ class RealNativeInputManager @Inject constructor(
...
         applyWindowChrome(widgetView, isBottom)
+        applyEnabledState(widgetView)
 
         if (!startEnterAnimation(widgetView, isBottom)) {
             animator.applyLayoutTransitions(widgetView, isBottom)
             onEnterComplete(widgetView)
         }
     }
 
+    override fun setEnabledState(state: EnabledState) {
```
malmstein: same as above, there’s a `NativeInputState` and a `NativeInputStatePublisher` that are in charge of setting the state of the input field. What happens here is exclusively related to onboarding and pulse animations, so I’d prefer that you either use the existing path to change the input field state or create a proper separation that only affects the Onboarding path. 

Calling this `enabledState` is not correct.

malmstein (changes requested):
This change not only affects Onboarding, because you are setting a “state” that is computed outside onboarding too. 

I’d like this to either use the existing path to update the input field state or create a separation that only affects onboarding. 

The suggestion to disabling views in bulk needs to go through O-J since we are about refactor much of the tech debt added.

malmstein (approved):
This looks great @lmac012 . Thanks for taking the time to work through this, our future refactor will thank you!

PR #8893, opened 2026-06-16
Extract referrer parsing into a :referral module behind a plugin point

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1215753390280439?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/481882893211075/task/1215446535130661?focus=true

### Description
Moves `:app`'s `referral` package into a new `:referral` module. To facilitate the move, a number of interfaces need to be exposed so the existing consumers in `:app` can still reach the features.

The initial plan in the TD was to expose the interfaces as-is. Looking closer, a number of functions and fields on the previous interfaces don't need to be exposed at all, so I took a pass to compress the exposed API surface down to what's actually used today and drop the old code smells.

Below is a table (scrolls horizontally :upside_down_face:) of the original interfaces and their updated state in the new module.

| Interface (in `:app` `referral`) | Before | After | Notes |
|---|---|---|---|
| `AppInstallationReferrerStateListener` | `initialiseReferralRetrieval()`, `waitForReferrerCode()`, `MAX_REFERRER_WAIT_TIME_MS` | **Exposed** in `:referral-api`, unchanged | The entry point consumers await. Only added KDoc. |
| `ParsedReferrerResult` | `sealed class` carrying `fromCache` on every subtype; markers were `data class` | **Exposed** in `:referral-api`, slimmed | Dropped `fromCache` (it was set but never read anywhere). The now-parameterless markers became `data object`. |
| `ParseFailureReason` | `sealed class` of failure reasons | **Exposed** in `:referral-api`, unchanged | Carried inside `ParsedReferrerResult.ParseFailure`. |
| `ReferrerParserPlugin` | did not exist | **New, exposed** in `:referral-api`: `process(Map<String, String>)` | The extension point that lets features parse their own referrer params. Takes a parsed key→value map rather than raw `&`-split parts, so plugins don't each re-parse. |
| `AppReferrerDataStore` | `interface` with 4 mutable `var`s, read/written app-wide | **Kept private** in `:referral-impl` | A mutable persistence store. Exposing it would leak impl-only fields (`campaignSuffix`, `referrerCheckedPreviously`) and hand out write access nothing outside the subsystem needs. The one cross-module read now goes through `AppReferrer` (below). |
| `AppInstallationReferrerParser` | `interface { parse(referrer): ParsedReferrerResult }` | **Kept private** in `:referral-impl` (impl `QueryParamReferrerParser`) | Only the referrer subsystem parses. Nothing in `:app` calls it directly. |
| `ReferrerOriginAttributeHandler` | bespoke `interface { process(List<String>) }`, called directly by the parser | **Removed**, reborn as `ReferrerOriginAttributeParserPlugin` (a `ReferrerParserPlugin`), private in `:referral-impl` | Folded the one-off handler into the generic plugin mechanism instead of a dedicated interface. |
| `AppReferrer` | `getOriginAttributeCampaign()`, `setOriginAttributeCampaign()` | **Exposed**, moved from `:browser-api` to `:referral`, added `isInstalledFromEuAuction()` | Gives `:app`'s one remaining EU-auction read a clean home so `AppReferrerDataStore` can stay private. |

Two `-api` relocations the move required, both unrelated to the referral surface itself:
- `VerificationCheckPlayStoreInstall` moved from `verified-installation-impl` to `verified-installation-api` (the origin plugin depends on it).
- The reserved EU-auction variant keys moved to `experiments-api`.

Implementations moved into `:referral-impl` with no API impact: `EmptyReferrerStateListener`, `PlayStoreAppReferrerStateListener`, and `AppReferenceSharePreferences` (renamed to `AppReferrerSharedPreferences`, also fixing the typo).

### Steps to test this PR
See [this](https://app.asana.com/1/137249556945/project/1207137509162935/task/1207930443876573?focus=true) for context.

- [ ] Uninstall the app
- [ ] Apply this patch:
```diff
diff --git a/app/build.gradle b/app/build.gradle
index bfa8c0d877..ebb49b3e94 100644
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -142,7 +142,7 @@ android {
     }
     buildTypes {
         debug {
-            applicationIdSuffix ".debug"
+//            applicationIdSuffix ".debug"
             pseudoLocalesEnabled false
             manifestPlaceholders = [
                     appIcon     : "@mipmap/ic_launcher_blue",
```

- [ ] Change build type to `PlayDebug`
- [ ] Filter logcat by "m_android_install"
- [ ] Open this link on the test device: https://play.google.com/store/apps/details?id=com.duckduckgo.mobile.android&referrer=utm_campaign%3Dappnosupport-atb-appnosupport%26origin%3Dfunnel_appnosupport_website
- [ ] Install the app (remember to use `PlayDebug`)
- [ ] Verify `m_android_install` pixel was sent with `origin=funnel_appnosupport_website`

Above steps focus on manually verifying the `origin` attribute, as that's the only actual implementation detail that changes (migration to a plugin). Everything else stays unchanged, besides exposing under refreshed interfaces.

### UI changes
No UI changes.

---

> [!NOTE]
> **Medium Risk**
> Touches install attribution, SERP source params, experiment variants, and analytics pixels; behavior is intended to be equivalent but the module boundary and origin parsing path changed.
> 
> **Overview**
> **Play Store install-referrer logic moves out of `:app` into new `:referral-api` / `:referral-impl` modules**, with `:app`, attributed-metrics, and installation wired to depend on them instead of in-app types. The Play Install Referrer dependency is removed from `:app` and lives on the `:referral-impl` **play** flavor only.
> 
> **Public surface is narrowed and relocated:** `AppReferrer` and install-referrer coordination types move from `:browser-api` / `:app` to `com.duckduckgo.referral.api`, including **`isInstalledFromEuAuction()`** so callers no longer touch the internal data store. **`ReferrerParserPlugin`** is new—referrer params are parsed once into a map and each plugin persists its own slice. Origin attribution is no longer a dedicated handler in `:app`; it runs as **`ReferrerOriginAttributeParserPlugin`**. **`ParsedReferrerResult`** drops unused **`fromCache`**; parameterless outcomes become **`data object`**.
> 
> **Supporting moves:** EU auction variant constants go to **`VariantManager`’s companion** in `experiments-api`; **`VerificationCheckPlayStoreInstall`** is exposed from `verified-installation-api`. Persistence is renamed to **`AppReferrerSharedPreferences`** and uses **`SharedPreferencesProvider`**. In-app referral tests/handlers are deleted or relocated under `:referral-impl`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48f2887cf97d2bdf69905145958572286ff99b50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verified-installation/verified-installation-api/src/main/java/com/duckduckgo/verifiedinstallation/installsource/VerificationCheckPlayStoreInstall.kt
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.verifiedinstallation.installsource
+
+/**
+ * Reports whether this app was installed from the Google Play Store, based on the recorded install
+ * source.
+ */
+interface VerificationCheckPlayStoreInstall {
```
LukasPaczos, line 23: This interface was already illegally used in `:app`, so making this formal here.

referral/referral-impl/src/test/java/com/duckduckgo/referral/impl/QueryParamReferrerParserTest.kt
```diff
@@ -59,10 +64,18 @@ class QueryParamReferrerParserTest {
 
     @Test
     fun whenReferrerContainsTargetButNoSuffixThenNoReferrerFound() {
-        val result = testee.parse("DDGRAX")
+        val result = testee.parse("DDGRA")
```
LukasPaczos, line 67: This test was broken, it expects no suffix.

LukasPaczos:
bugbot run

marcosholgado (approved):
LGTM. Also checked that FDroid builds just in case.

PR #8955, opened 2026-06-22
Fix Dax legs cut off on the onboarding end CTA

lmac012:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1215937292229504

### Description

The waving Dax on the onboarding end CTA has no legs in its asset, so it needs to sit flush against the bottom input field to read correctly. With a bottom address bar it floated above the input with its body cut off, because the new-tab content offset reserved the full input-widget height — including the part of the widget below the NTP container — leaving a dead gap.

Now the offset reserves only the height by which the input card overlaps the content, so NTP content (and the Dax) sits at the input card's top. The `DAX_END` Dax is also nudged right (bottom address bar only) so both legs tuck behind the card. The `DUCK_AI_DAX_END` Dax doesn't need similar nudge, because it is shown on top of input toggle, which is wider and covers his legs already.

### Steps to test this PR

_Onboarding end CTA — bottom address bar_
- [ ] Set the address bar to the bottom, then run onboarding to the end ("You've got this") CTA
- [ ] Confirm the Dax sits flush against the input field with both legs behind it — no gap, no cut-off body

_Top address bar (regression)_
- [ ] Repeat with the address bar at the top and confirm the Dax is unchanged

### UI changes
| Before  | After |
| ----- | ----- |
| Screenshot_20260622_161555 | Screenshot_20260622_145921 |

---

> [!NOTE]
> **Medium Risk**
> Touches shared native-input layout padding used by NTP and browser chrome during animations; wrong overlap math could affect bottom-bar browsing layout beyond onboarding.
> 
> **Overview**
> Fixes the onboarding **end** CTA with a **bottom address bar**: waving Dax no longer floats above the input with his body cut off.
> 
> **Native input content padding** — For bottom omnibar, bottom padding no longer uses the full `widgetView` height. It now reserves only the overlap between NTP/browser content and the input **card** top (same idea as autocomplete offset). Steady-state layout, layout listeners, and widget enter/exit animation frames all pass both the card’s top and bottom in window coordinates.
> 
> **Dax positioning** — `DaxEndBrandDesignUpdateBubbleCta` gets `isOmnibarBottom` from settings (`OmnibarType.SINGLE_BOTTOM`). When the bar is at the bottom, horizontal nudge changes from `-40dp` to `-30dp` so both legs sit behind the narrower bottom card.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8a0e1a3b52d0de3b6cbcc700005945b3235053b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
@@ -236,15 +236,17 @@ class NativeInputLayoutCoordinator(
             return maxOf(0, anchorBottomInWindow - viewLocation[1])
         }
 
-        fun computeDeltaBottom(): Int {
+        fun computeDeltaBottom(view: View, anchorTopInWindow: Int): Int {
             if (!isBottom) return 0
-            return maxOf(0, widgetView.height)
+            val viewLocation = IntArray(2).also { view.getLocationInWindow(it) }
+            val viewBottomInWindow = viewLocation[1] + view.height
+            return maxOf(0, viewBottomInWindow - anchorTopInWindow)
         }
 
-        fun applyOffsetWithBottom(anchorBottomInWindow: Int) {
-            val deltaBottom = computeDeltaBottom()
+        fun applyOffsetWithBottom(anchorTopInWindow: Int, anchorBottomInWindow: Int) {
             targets.forEach { target ->
                 val deltaTop = computeDeltaTop(target.view, anchorBottomInWindow)
+                val deltaBottom = computeDeltaBottom(target.view, anchorTopInWindow)
                 applyPadding(target.view, target.basePadding, deltaTop, deltaBottom)
             }
         }
@@ -256,7 +258,7 @@ class NativeInputLayoutCoordinator(
             }
             val anchorLocation = IntArray(2).also { anchor.getLocationInWindow(it) }
             val anchorBottomInWindow = anchorLocation[1] + anchor.height
-            applyOffsetWithBottom(anchorBottomInWindow)
+            applyOffsetWithBottom(anchorTopInWindow = anchorLocation[1], anchorBottomInWindow = anchorBottomInWindow)
         }
 
         // Called from the enter/exit animators' onUpdate, BEFORE the layout pass that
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
```diff
@@ -236,15 +236,17 @@ class NativeInputLayoutCoordinator(
             return maxOf(0, anchorBottomInWindow - viewLocation[1])
         }
 
-        fun computeDeltaBottom(): Int {
+        fun computeDeltaBottom(view: View, anchorTopInWindow: Int): Int {
```
malmstein, line 239: this changes `computeDeltaBottom` from a blanket `widgetView.height` to a per-target `viewBottomInWindow - anchorTopInWindow`, which re-pads *every* target the coordinator manages in bottom mode — not just the onboarding end CTA. that's the more correct geometry, but since it touches general bottom-mode padding, did we confirm the other targets still sit right (the blanket height was presumably over-padding)? worth a device pass beyond just the CTA.

malmstein (approved):
LGTM, approving. did not install or smoke-test the APK — static-only, and this is a visual/layout fix so it really wants eyeballing on a device. the two changes are sensible: the `computeDeltaBottom` rewrite computes the actual overlap of each target's bottom against the anchor top instead of blanket-padding by the widget height, and the `DaxEndBrandDesignUpdateBubbleCta` `translationXDp` nudge (-30 vs -40 when `isOmnibarBottom`) is well-commented. the `isOmnibarBottom = omnibarType == SINGLE_BOTTOM` check is the right complement to the existing `isTop = SINGLE_TOP || SPLIT` convention, so SPLIT correctly stays non-bottom.

one thing worth a reply: the coordinator change is general bottom-mode padding, not CTA-specific, so I'd confirm the other padded targets still look right. no blockers.

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputLayoutCoordinator.kt
lmac012: I tested favorites, unusually long RMF, etc. - I didn't see any issues with this fix.

PR #8987, opened 2026-06-24
Add the new instrumentation pixels for the linear onboarding

catalinradoiu:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1211724162604201/task/1215956973978692?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/1211724162604201/task/1215956973978711?focus=true

### Description

### Description                                                                                                                                                                                                    
                                                                                                                                                                                                                     
  Adds the new linear onboarding instrumentation pixels (shown / clicked / confirmed, per the onboarding instrumentation standard). Two parts:

1. Adds `OnboardingPixelSender` with a single `fire(pixelName, action)` entry point. The pixel **name lives on the step** — `pixelName: OnboardingPixelName?` on `NewUserOnboardingActivityStep`/ `NewUserBrowserActivityStep`, required with no default (a step that fires no pixel sets it to `null` explicitly) — and `OnboardingPixelAction` is a sealed type describing *what happened*: `Shown`, `Clicked(engaged)`, and per-screen payload variants (`AddressBarClicked`, `SearchExperienceClicked`, `TryASearchClicked`, `QuickSetupClicked`, `SetDefaultConfirmed`, `NotificationsConfirmed`).`RealOnboardingPixelSender` assembles the standard params (`it`, `source`, `flow`, `variant`, `d`, `pixelSource`) and owns the wire format; the exhaustive `when` over `OnboardingPixelAction` makes it a compile error to add an action shape without handling it. The former `QuickSetupPixelSender` is folded in. All pixel firing lives in the orchestrator step definitions (`NewUserOnboardingPlanProvider`), not the ViewModel — the ViewModel only emits a pixel-unaware `NewUserOnboardingEvent.Presented` signal when a step is rendered. A `firingShownPixels()` wrapper reads each step's `pixelName` and fires the `Shown` pixel on `Presented`; clicked/confirmed pixels are fired from the relevant step transitions using data the orchestrator events already carry. This reaches steps hosted in both `OnboardingActivity` and `BrowserActivity` (e.g. the Duck.ai demo), which the previous `PreOnboardingDialogType` anchor structurally could not. 

2. Adds the new onboarding standard pixels (shown/clicked/confirmed, following the onboarding instrumentation standard, defined in `PixelDefinitions/pixels/definitions/onboarding.json5`). The names live in a dedicated `OnboardingPixelName` type (kept out of `AppPixelName`) and are registered in `PixelParamRemovalInterceptor` for atb stripping.

### Steps to test this PR

_Testing the onboarding instrumentation pixels_

- [ ] Go through the linear onboarding and check that the pixels fire according to the specs: https://app.asana.com/1/137249556945/project/72649045549333/task/1213601955483546.
- [ ] Test the sync restore and quick setup as well (apply Patch C for quick setup and then skip the onboarding)

These flows are gated, so apply the testing patches below first:

1. To go through the **Duck.ai onboarding**, apply the `CustomAiOnboardingStore.kt` patch (Patch A below) — it forces the custom Duck.ai onboarding path.
2. To test the **input screen preview**, apply the `ExtendedOnboardingFeatureToggles.kt` patch (Patch B below) — it enables the `duckAiOnboarding` feature toggle.
3. Build & install, run through onboarding, and confirm the pixels fire as expected (sync-restore shown/clicked, input-preview shown/clicked with the right value, ai-comparison shown/clicked).

<details>
<summary>Patch A — Duck.ai onboarding (apply to test the Duck.ai flow)</summary>

```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt b/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt
index 9da0ed066d..68308c5c59 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt
@@ -113,13 +113,15 @@ class CustomAiOnboardingStoreImpl @Inject constructor(
             val resolution = referrerExists && customAiOnboardingEnabled && orchestratorEnabled && brandDesignEnabled
             preferences.edit { putBoolean(PREFS_KEY_ENABLED, resolution) }
 
-            return@withContext resolution
+            // return@withContext resolution
+            return@withContext true
         }
     }
 
     override suspend fun isEnabled(): Boolean = resolveMutex.withLock {
         withContext(dispatcherProvider.io()) {
             preferences.getBoolean(PREFS_KEY_ENABLED, false)
+            true
         }
     }
```
</details>

<details>
<summary>Patch B — Input screen preview (apply to test the input-screen preview)</summary>

```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/extendedonboarding/ExtendedOnboardingFeatureToggles.kt b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/extendedonboarding/ExtendedOnboardingFeatureToggles.kt
index 1a9607ac44..48af91eda4 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/extendedonboarding/ExtendedOnboardingFeatureToggles.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/extendedonboarding/ExtendedOnboardingFeatureToggles.kt
@@ -50,6 +50,6 @@ interface ExtendedOnboardingFeatureToggles {
     @Experiment
     fun freeTrialCopy(): Toggle
 
-    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
     fun duckAiOnboarding(): Toggle
 }
```
</details>

<details>
<summary>Patch C — Quick Setup (apply to test the quick setup flow)</summary>

```
diff --git a/app/src/main/java/com/duckduckgo/app/onboardingquicksetup/OnboardingQuickSetupExperimentManager.kt b/app/src/main/java/com/duckduckgo/app/onboardingquicksetup/OnboardingQuickSetupExperimentManager.kt
--- a/app/src/main/java/com/duckduckgo/app/onboardingquicksetup/OnboardingQuickSetupExperimentManager.kt	(revision 1a17392d88d5240f79da4b2381cacafcaeafde03)
+++ b/app/src/main/java/com/duckduckgo/app/onboardingquicksetup/OnboardingQuickSetupExperimentManager.kt	(date 1782291823249)
@@ -45,19 +45,19 @@
 ) : OnboardingQuickSetupExperimentManager {
 
     override suspend fun enroll(): QuickSetupExperimentVariant? = withContext(dispatcherProvider.io()) {
-        toggles.onboardingQuickSetupExperimentJun3().enroll()
-        when {
-            toggles.onboardingQuickSetupExperimentJun3().isEnrolledAndEnabled(QuickSetupCohorts.TREATMENT) -> {
+        // toggles.onboardingQuickSetupExperimentJun3().enroll()
+        // when {
+        //     toggles.onboardingQuickSetupExperimentJun3().isEnrolledAndEnabled(QuickSetupCohorts.TREATMENT) -> {
                 QuickSetupExperimentVariant.TREATMENT
-            }
-
-            toggles.onboardingQuickSetupExperimentJun3().isEnrolledAndEnabled(QuickSetupCohorts.CONTROL) -> {
-                QuickSetupExperimentVariant.CONTROL
-            }
-
-            else -> {
-                null
-            }
-        }
+        //     }
+        //
+        //     toggles.onboardingQuickSetupExperimentJun3().isEnrolledAndEnabled(QuickSetupCohorts.CONTROL) -> {
+        //         QuickSetupExperimentVariant.CONTROL
+        //     }
+        //
+        //     else -> {
+        //         null
+        //     }
+        // }
     }
 }
```

</details>

### UI changes
| Before  | After |
| ------ | ----- |





<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Large orchestrator and telemetry touch area with dual legacy and new pixel paths during rollout; incorrect wiring could skew onboarding analytics but does not affect auth or user data.
> 
> **Overview**
> Introduces **standard onboarding instrumentation** (`onboarding_*` pixels with `shown` / `clicked` / `confirmed`) and wires them through the linear orchestrator instead of ad hoc VM senders.
> 
> **Pixel definitions & types:** `onboarding.json5` gains per-step pixels (welcome, set-default, address bar, search experience, skip, notifications, search-chat toggle, AI intro, fire button, quick setup). Shared params drop `source`; `flow` now includes `duckai`. `OnboardingPixelName` replaces `AppPixelName.ONBOARDING_QUICK_SETUP`, with ATB stripping via `OnboardingPixelParamRemovalPlugin`.
> 
> **Sending:** `QuickSetupPixelSender` is removed in favor of `OnboardingPixelSender` + `OnboardingPixelAction`, which builds standard params (`it`, `flow`, `variant`, `d`, `pixelSource`, etc.) and unique tags. Search vs chat branch is persisted for later `variant` on Duck.ai paths.
> 
> **Orchestrator:** Each activity/browser step carries optional `pixelName`. `firingShownPixels()` fires `Shown` on `NewUserOnboardingEvent.Presented`; step transitions fire clicked/confirmed actions (including default-browser result, notification grant/deny, demo query with `fromSuggestion`). Browser Duck.ai demo also emits `Presented` when the demo opens.
> 
> **UI layer:** The welcome VM forwards `Presented` when dialogs render and when the notification prompt shows; permission deny is tracked; input demo passes suggestion vs custom. Legacy `m_preonboarding_*` impression pixels in `fireDialogShownPixel` are largely unchanged—new standard pixels are orchestrator-owned.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4d682c81d152f3770793d402dae2b3e38b959f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModel.kt
```diff
@@ -290,29 +299,66 @@ class BrandDesignUpdatePageViewModel @Inject constructor(
...
+        fireBrandDesignSecondaryClickedPixel(currentDialog)
         flow.onSecondaryCta(currentDialog)
     }
 
+    private fun fireBrandDesignPrimaryClickedPixel(dialog: PreOnboardingDialogType) {
+        val action = OnboardingAction.PrimaryClick(
+            addressBarPosition = _viewState.value.selectedAddressBarPosition,
+            withAi = _viewState.value.inputScreenSelected,
+        )
+        dialog.toEvent(action)?.let { brandDesignOnboardingPixelSender.fire(pixelContext(), it) }
```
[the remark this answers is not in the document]
catalinradoiu: Not applicable since we don’t show the split omnibar option if it’s disabled, so the user cannot select it.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModel.kt
```diff
@@ -290,29 +299,66 @@ class BrandDesignUpdatePageViewModel @Inject constructor(
...
+
+    // Builds the shared onboarding pixel context from current viewState. The install-type override
+    // exists for the notification pixels, which fire before loadDaxDialog resolves reinstall status
+    // into viewState and therefore pass the value resolved via isReinstallDeferred.
+    private fun pixelContext(isReinstallUser: Boolean = _viewState.value.isReinstallUser): OnboardingPixelContext =
+        OnboardingPixelContext(
+            isReinstallUser = isReinstallUser,
+            // source/flow are resolved inside the sender from CustomAiOnboardingStore.
+            variant = _viewState.value.onboardingBranchVariant,
+        )
```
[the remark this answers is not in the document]
catalinradoiu: The sync-restore pixel has been removed.

app/src/main/java/com/duckduckgo/app/global/api/PixelParamRemovalInterceptor.kt
```diff
@@ -166,6 +166,12 @@ object PixelInterceptorPixelsRequiringDataCleaning : PixelParamRemovalPlugin {
             AppPixelName.NEW_ADDRESS_BAR_PICKER_V2_CONFIRMED_DAILY.pixelName to PixelParameter.removeAtb(),
             AppPixelName.PREONBOARDING_SPLIT_ADDRESS_BAR_SELECTED_UNIQUE.pixelName to PixelParameter.removeAtb(),
             AppPixelName.ONBOARDING_QUICK_SETUP.pixelName to PixelParameter.removeAtb(),
+            AppPixelName.ONBOARDING_WELCOME.pixelName to PixelParameter.removeAtb(),
+            AppPixelName.ONBOARDING_SET_DEFAULT.pixelName to PixelParameter.removeAtb(),
+            AppPixelName.ONBOARDING_ADDRESS_BAR_POSITION.pixelName to PixelParameter.removeAtb(),
+            AppPixelName.ONBOARDING_SEARCH_EXPERIENCE.pixelName to PixelParameter.removeAtb(),
+            AppPixelName.ONBOARDING_SKIP_ONBOARDING.pixelName to PixelParameter.removeAtb(),
+            AppPixelName.ONBOARDING_NOTIFICATIONS.pixelName to PixelParameter.removeAtb(),
```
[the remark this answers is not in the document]
catalinradoiu: Done

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
```diff
@@ -0,0 +1,404 @@
...
+                },
+            )
+        }
+        val params = mutableMapOf(
+            PIXEL_PARAM_INSTALL_TYPE to if (context.isReinstallUser) INSTALL_TYPE_REINSTALL else INSTALL_TYPE_NEW,
+            PIXEL_PARAM_SOURCE to if (isCustomAiFlow) SOURCE_DUCKAI_CPP else ONBOARDING_DEFAULT,
+            PIXEL_PARAM_FLOW to if (isCustomAiFlow) FLOW_DUCKAI else ONBOARDING_DEFAULT,
+            PIXEL_PARAM_PIXEL_SOURCE to deviceInfo.formFactor().description,
+        )
+        variant?.let { params[PIXEL_PARAM_VARIANT] = it }
```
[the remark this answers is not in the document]
catalinradoiu: The branching between the 2 variants can happen only after we try a search or a chat in the input screen preview. This is the intended behavior. We don’t have a variant before that. 
Also in the custom duck.ai plan we are using the `flow` and `source` that will tell us we’re in the variant of the onboarding.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModel.kt
```diff
@@ -271,13 +272,9 @@ class BrandDesignUpdatePageViewModel @Inject constructor(
...
             INPUT_SCREEN -> pixel.fire(PREONBOARDING_CHOOSE_SEARCH_EXPERIENCE_IMPRESSIONS_UNIQUE, type = Unique())
-            INPUT_SCREEN_PREVIEW -> {
-            }
-
-            QUICK_SETUP -> {
-                quickSetupPixelSender.fireShown(isReinstallUser = _viewState.value.isReinstallUser)
-            }
+            INPUT_SCREEN_PREVIEW, QUICK_SETUP -> Unit
         }
+        viewModelScope.launch { orchestrator.onEvent(NewUserOnboardingEvent.Presented) }
```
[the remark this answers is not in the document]
catalinradoiu: This will be removed in the future. Currently only the `OrchestratorFlow` is active and the plan is to remove the legacy. @LukasPaczos I don’t think we should implement this for the legacy, unless you believe otherwise.

app/src/main/java/com/duckduckgo/app/global/api/PixelParamRemovalInterceptor.kt
```diff
@@ -166,7 +166,15 @@ object PixelInterceptorPixelsRequiringDataCleaning : PixelParamRemovalPlugin {
...
-            AppPixelName.ONBOARDING_QUICK_SETUP.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_QUICK_SETUP.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_WELCOME.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_SET_DEFAULT.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_ADDRESS_BAR_POSITION.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_SEARCH_EXPERIENCE.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_SKIP_ONBOARDING.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_NOTIFICATIONS.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_SEARCH_CHAT_TOGGLE.pixelName to PixelParameter.removeAtb(),
+            OnboardingPixelName.ONBOARDING_AI_INTRO.pixelName to PixelParameter.removeAtb(),
```
LukasPaczos: nit: you could create a separate removal plugin that lives closer to your `OnboardingPixelName` (or even in the same file)

app/src/main/java/com/duckduckgo/app/pixels/OnboardingPixelName.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.duckduckgo.app.statistics.pixels.Pixel
+
+/**
+ * Pixel names for the linear-onboarding instrumentation. Kept separate from [AppPixelName] so the
+ * onboarding pixel catalogue stays self-contained; the wire strings are unchanged.
+ */
+enum class OnboardingPixelName(override val pixelName: String) : Pixel.PixelName {
+    ONBOARDING_QUICK_SETUP("onboarding_quick-setup"),
+    ONBOARDING_WELCOME("onboarding_welcome"),
+    ONBOARDING_SET_DEFAULT("onboarding_set-default"),
```
LukasPaczos, line 24:
```suggestion
    ONBOARDING_SET_DEFAULT_BROWSER("onboarding_set-default-browser"),
```
I'm worried there might be many more 'defaults' to set in the future 🙃

app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
```diff
@@ -277,220 +303,307 @@ class NewUserOnboardingPlanProvider @Inject constructor(
...
-        },
-    )
+            },
+        )
+    }
 
     private fun defaultBrowserPromptStep() = NewUserOnboardingActivityStep(
         id = NewUserOnboardingStepIds.DEFAULT_BROWSER_PROMPT,
+        // No shown pixel of its own; the confirmed result belongs to the set-default pixel (started on the comparison-chart step).
+        pixelName = null,
```
LukasPaczos, line 449: Wondering why shouldn't this be a separate step. Conceptually it's connected to the comparison chart screen right now, but technically it could be moved to any place in the onboarding. There should be no problems to connect the dots on the data visualization side, but we'd retain the granularity and avoid coupling the steps.

app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
```diff
@@ -277,220 +303,307 @@ class NewUserOnboardingPlanProvider @Inject constructor(
...
+                    }
+                    else -> Stay
+                }
+            },
+        )
+    }
 
     private fun duckAiDemoStep(ctx: NewUserOnboardingPlanContext) = NewUserBrowserActivityStep(
         id = NewUserOnboardingStepIds.DUCK_AI_DEMO,
+        pixelName = null,
```
LukasPaczos: Is there are reason not to track this one?

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModel.kt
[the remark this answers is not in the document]
LukasPaczos: I think that's fine 👍

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
```diff
@@ -0,0 +1,259 @@
...
+    data object Shown : OnboardingPixelAction
+
+    /** A click/tap. [engaged] null = no value param; true/false = engage/dismiss. */
+    data class Clicked(val engaged: Boolean? = null) : OnboardingPixelAction
+
+    data class SetDefaultConfirmed(val isDdgDefault: Boolean) : OnboardingPixelAction
+    data class NotificationsConfirmed(val granted: Boolean) : OnboardingPixelAction
+    data class AddressBarClicked(val position: OmnibarType) : OnboardingPixelAction
+    data class SearchExperienceClicked(val withAi: Boolean) : OnboardingPixelAction
+    data class TryASearchClicked(
```
LukasPaczos:
```suggestion
    data class TryInputClicked(
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
```diff
@@ -0,0 +1,259 @@
...
+        event: String,
+        value: String? = null,
+    ) {
+        appCoroutineScope.launch {
+            val params = buildStandardParams().toMutableMap()
+            params[PIXEL_PARAM_EVENT] = event
+            value?.let { params[PIXEL_PARAM_VALUE] = it }
+            val tag = buildString {
+                append(pixelName.pixelName).append("_").append(event)
+                value?.let { append("_").append(it) }
```
LukasPaczos, line 178: Should the value be appended here? This would assume a step can be shown multiple times and produce different values. Also, we're not appending the value in quick setup step.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
```diff
@@ -0,0 +1,259 @@
...
+                when (onboardingStore.getOnboardingVariant()) {
+                    "search" -> VARIANT_SEARCH
+                    "chat" -> VARIANT_CHAT
+                    else -> null
+                },
+            )
+        }
+        val params = mutableMapOf(
+            PIXEL_PARAM_INSTALL_TYPE to if (reinstall) INSTALL_TYPE_REINSTALL else INSTALL_TYPE_NEW,
+            PIXEL_PARAM_SOURCE to if (isCustomAiFlow) SOURCE_DUCKAI_CPP else ONBOARDING_DEFAULT,
```
LukasPaczos: If `PIXEL_PARAM_SOURCE` is meant to represent the store listing user downloaded from, it's impossible to determine on Android, and the current `SOURCE_DUCKAI_CPP` would not be reliable. If user downloads from the Duck.ai store listing by discovering it organically, without a referral link, we would not have the signal.

app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
```diff
@@ -277,220 +303,307 @@ class NewUserOnboardingPlanProvider @Inject constructor(
...
+            showsStepIndicator = true,
+            resolveDialog = { NewUserOnboardingActivityDialog.InputScreenPreview(isSearchDefault = false) },
+            transition = { event ->
+                when {
+                    event is NewUserOnboardingEvent.InputDemoQuerySubmitted -> {
+                        if (event.isChat) {
+                            onboardingStore.setChatOnboardingVariant()
+                        } else {
+                            onboardingStore.setSearchOnboardingVariant()
+                        }
```
LukasPaczos: The input screen preview step in the Custom AI flow has no variant, it's always chat-only.

app/src/main/java/com/duckduckgo/app/onboarding/store/OnboardingStoreImpl.kt
```diff
@@ -226,11 +226,27 @@ class OnboardingStoreImpl @Inject constructor(
...
+        preferences.edit { putString(KEY_ONBOARDING_VARIANT, ONBOARDING_VARIANT_SEARCH) }
+    }
+
+    override fun setChatOnboardingVariant() {
+        preferences.edit { putString(KEY_ONBOARDING_VARIANT, ONBOARDING_VARIANT_CHAT) }
+    }
+
+    override fun getOnboardingVariant(): String? {
+        return preferences.getString(KEY_ONBOARDING_VARIANT, null)
+            ?.takeIf { it == ONBOARDING_VARIANT_SEARCH || it == ONBOARDING_VARIANT_CHAT }
```
LukasPaczos: Why do we need this check?

app/src/main/java/com/duckduckgo/app/onboarding/store/OnboardingStore.kt
```diff
@@ -30,4 +30,7 @@ interface OnboardingStore {
     fun setInputScreenSelectionOverriddenByUser()
     fun setDuckAiOnboardingFlow()
     fun isDuckAiOnboardingFlow(): Boolean
+    fun setSearchOnboardingVariant()
+    fun setChatOnboardingVariant()
+    fun getOnboardingVariant(): String?
```
LukasPaczos: Does this need to be in the `OnboardingStore` and accessible to everyone? Couldn't this info be persisted and used internally by `OnboardingPixelSender`?

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
```diff
@@ -96,10 +96,12 @@ class OnboardingViewModel @Inject constructor(
                     // Arm the in-browser Duck.ai demo (sets the flow + silences the standard DAX CTAs).
                     // Shared with the linear-onboarding duck_ai_demo step so both paths arm identically.
                     duckAiOnboardingDemo.arm()
+                    onboardingStore.setChatOnboardingVariant()
```
LukasPaczos: This is already set in the onboarding plan, no?

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
```diff
@@ -96,10 +96,12 @@ class OnboardingViewModel @Inject constructor(
...
                 }
 
                 DEFAULT_WITHOUT_INTRO_CTA -> {
                     dismissedCtaDao.insert(DismissedCta(CtaId.DAX_INTRO))
+                    onboardingStore.setSearchOnboardingVariant()
```
LukasPaczos: Same as above.

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
LukasPaczos, line 1: Aren't we missing a test that verifies `Presented` even is sent?

PixelDefinitions/pixels/definitions/onboarding.json5
```diff
@@ -76,13 +76,13 @@
                 "key": "source",
                 "type": "string",
                 "description": "Onboarding entry point",
-                "enum": ["default"]
+                "enum": ["default", "duckai_cpp"]
```
LukasPaczos: Same as above, we can't reliably send `duckai_cpp`.

catalinradoiu:
* **#9027** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9027?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8987** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8987?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/8987?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
catalinradoiu: As per the instrumentation standard https://app.asana.com/1/137249556945/project/72649045549333/task/1213601955483546?focus=true, every pixel should be unique per user per combination of step, e value, that’s why I added the value as well. I agree that it can produce different values but the alternative of not adding the value could lead us to missing pixels. For example someone goes through the onboarding, gets to the address bar positon selection screen and select `top`. Then closes the app, goes through the onboarding again as it wasn’t completed and this time the user selects `bottom` omnibar. If I wouldn’t add the value I would miss the second event. 

For the quick setup we don’t really need the value as we will send the pixel only when the user clicks “Start browsing” and by that time the onboarding is already completed. I can it there though for consistency.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
catalinradoiu: I have opened a discussion around this: https://app.asana.com/1/137249556945/task/1213629555396912/comment/1216050288368419?focus=true

PixelDefinitions/pixels/definitions/onboarding.json5
catalinradoiu: Started the dicussion here: https://app.asana.com/1/137249556945/task/1213629555396912/comment/1216050288368419?focus=true

app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
catalinradoiu: Currently the specs define the set as default and browser comparison screen as a single step. The set as default is basically a system prompt and all of the events that are system prompts/ actions took effect outside the app are defined as `confirmed`. We currently don’t send any `shown` event for the default browser prompt and I am not sure if we should. We can get that information from the `clicked` event in the `set_as_default` pixel.

I can also understand your point that this could be a separate step, but we somehow need to correlate it with the current one (in this case the browser comparison) as this is a system prompt rather than a standalone screen. 

An alternative would be to set the `pixelName` as `ONBOARDING_SET_DEFAULT` and rely on the fact that each pixel even is unique. But I am not sure if this would be the right thing to do, as in this case we would have the same name for 2 steps, which I don’t see as the better solution.

app/src/main/java/com/duckduckgo/app/pixels/OnboardingPixelName.kt
catalinradoiu: I see your point and it makes sense. However changing this name here would require changing it on the other platforms as well, as these events have already been implemented. Are we fine keeping it as it is for now and changing this in the future if we’ll end up with more defaults or should we start a discussion and change it for every other platform?

LukasPaczos: Fine to keep as-is if this was already agreed on :+1:

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
LukasPaczos: Adding it for consistency makes sense, in case quick step ever shifts to not be the final step.

catalinradoiu:
The pixel for the fire button usage on the chat tab, in the duck ai onboarding, will be added in the next PR.

app/src/main/java/com/duckduckgo/app/global/api/PixelParamRemovalInterceptor.kt
catalinradoiu: done

app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
catalinradoiu: I missed that initially. I added it now, thanks for raising this

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
catalinradoiu: Done

app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
catalinradoiu: Remove the if check

app/src/main/java/com/duckduckgo/app/onboarding/store/OnboardingStoreImpl.kt
catalinradoiu: leftover, remove it

app/src/main/java/com/duckduckgo/app/onboarding/store/OnboardingStore.kt
catalinradoiu: I think it makes sense. I moved it to the `OnboardingPixelSender`

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
catalinradoiu: leftover from another solution, removed it

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
catalinradoiu: removed it

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
catalinradoiu: Added them, missed them initially

app/src/main/java/com/duckduckgo/app/pixels/OnboardingPixelName.kt
```diff
@@ -32,5 +32,5 @@ enum class OnboardingPixelName(override val pixelName: String) : Pixel.PixelName
     ONBOARDING_NOTIFICATIONS("onboarding_notifications"),
     ONBOARDING_SEARCH_CHAT_TOGGLE("onboarding_search-chat-toggle"),
     ONBOARDING_AI_INTRO("onboarding_ai-intro"),
-    ONBOARDING_AI_CHAT("onboarding_ai-chat"),
+    ONBOARDING_FIRE_BUTTON("onboarding_fire-button"),
```
catalinradoiu, line 31: Renamed the pixel key `onboarding_ai-chat` to `onboarding_fire-button`, matching the iOS behavior. This is being sent from the ai demo in the duck.ai onboarding. When the fire button is used we are sending with `event=clicked`.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
LukasPaczos:
```suggestion
            PIXEL_PARAM_SOURCE to null
```
I understand we want to keep the option to send the source, but as things stand, there's no source we can reliably determine given existing definition (where `source` is expected to be the store listing variant). So, for now, I think we should just send `null`.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
```diff
@@ -0,0 +1,286 @@
...
+        variant?.let { params[PIXEL_PARAM_VARIANT] = it }
+        if (days in 0..MAX_DAYS_SINCE_INSTALL_REPORTED) {
+            params[PIXEL_PARAM_DAYS_SINCE_INSTALL] = days.toString()
+        }
+        return params
+    }
+
+    private fun engageOrDismiss(engaged: Boolean): String = if (engaged) VALUE_ENGAGE else VALUE_DISMISS
+
+    private fun tryASearchValue(fromSuggestion: Boolean, isChat: Boolean): String {
```
LukasPaczos: nit
```suggestion
    private fun tryInputValue(fromSuggestion: Boolean, isChat: Boolean): String {
```

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
```diff
@@ -852,22 +873,30 @@ class BrandDesignUpdatePageViewModelTest {
...
+        advanceUntilIdle()
         verify(mockPixel).fire(
             AppPixelName.NOTIFICATIONS_ENABLED,
             mapOf(PixelParameter.FROM_ONBOARDING to true.toString()),
         )
     }
 
+    // Notification brand-design pixels (NotificationsShown/Confirmed) are now fired from the
+    // orchestrator plan provider, not the ViewModel. The VM only fires the legacy NOTIFICATION_RUNTIME_PERMISSION_SHOWN
+    // pixel and records the grant outcome for forwarding via NotificationPermissionFinished(granted).
```
LukasPaczos: Is this comment needed here? Feels like a little bit of bloat.

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
```diff
@@ -1080,13 +1109,16 @@ class BrandDesignUpdatePageViewModelTest {
...
-            testee.onInputModeDemoQuerySubmitted("search query", isChat = false)
+            testee.onInputModeDemoQuerySubmitted("search query", isChat = false, fromSuggestion = false)
             val command = awaitItem()
             assertTrue(command is Command.FinishAndSubmitSearchQuery)
             assertEquals("search query", (command as Command.FinishAndSubmitSearchQuery).query)
         }
     }
 
+    // TryInputClicked brand-design pixel and onboarding variant setting are now fired from the
+    // orchestrator plan provider (inputScreenPreviewStep / customAiInputScreenPreviewStep), not the ViewModel.
```
LukasPaczos: Same as above.

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
```diff
@@ -1175,39 +1207,9 @@ class BrandDesignUpdatePageViewModelTest {
...
-
-        verify(mockQuickSetupPixelSender).fireClicked(
-            isReinstallUser = true,
-            addressBarPosition = OmnibarType.SINGLE_BOTTOM,
-            inputScreenSelected = false,
-        )
-    }
+    // Brand-design pixels (QuickSetupShown/Clicked, AddressBarPositionClicked, SearchExperienceClicked,
+    // SkipOnboardingClicked, SetDefaultClicked, SetDefaultConfirmed) are now fired from the orchestrator
+    // plan provider step definitions, not the ViewModel. Tests for these moved to NewUserOnboardingPlanProviderTest.
```
LukasPaczos: Same as above.

LukasPaczos (approved):
LGTM, thanks for addressing the comments.

Couple of nits below. Also, I'd still recommend not sending the `source` param for now, at least if its expected set of values remains "the store listing user came from", since we can't determine that. Just sending `default` would also be misleading. The custom AI flow is already covered by the `flow` param.

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
catalinradoiu: removed

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
catalinradoiu: removed

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdatePageViewModelTest.kt
catalinradoiu: removed

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingPixelSender.kt
catalinradoiu: done

PR #9015, opened 2026-06-25
Fire Mode: Fix the UI glitch when switching modes

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1216049501259336?focus=true

### Description

This PR fixes 2 bugs that happen while switching modes in the Tab switcher:

1. While [rapidly switching between modes](https://github.com/duckduckgo/Android/pull/8748#pullrequestreview-4504401491), the app eventually freezes. This caused by app trying to recreate the activity multiple times. User is no prevented from doing that.
2. Sometimes when switching modes, there is lag lasting a few seconds. Because switching modes in the tab switcher and changing the omnibar position recreate the activity via `Activity.recreate()`, it briefly drops the started-activity count to 0. Previously this fired the full app foreground/background pipeline (`BrowserLifecycleObserver.onClose()`/`onOpen()` — e.g. `AutomaticDataClearer`, ANR-supervisor teardown, session pixels) on every switch. Running the foreground pipeline on every switch intermittently stalled the main thread ~0.6–1.6s (observed `Davey! duration=1647ms`), which is the switcher lag/"freeze". After the fix, `BrowserApplicationStateInfo` now treats a stop→start caused by a configuration-change recreate (`Activity.isChangingConfigurations`) as *not* a real background→foreground transition, and skips `onClose()`/`onOpen()` for it. A genuine background still reports `isChangingConfigurations == false`, so `onClose()`/`onOpen()` fire exactly as before — verified on-device across HOME, screen-off, and app-switch (see table).

image

### Steps to test this PR

**Setup**
- Internal build. Enable the `fireTabs` flag, then force-stop and relaunch

**Mode-switch responsiveness (freeze/lag fix)**
- [ ] Open the tab switcher and rapidly toggle Regular↔Fire many times (tap-storm). The toggle stays responsive and the switcher never freezes, blanks, or leaves the tab grid hidden.
- [ ] Toggle into the Fire empty state (no Fire tabs) and back to Regular. The toggle is still responsive in the empty state.
- [ ] Switching modes is smooth, with no multi-second stall/jank on the switch.

**Auto data-clearing still fires on a real background (key regression check — this PR changes the onClose/onOpen path)**
- [ ] Settings → set automatic clearing to "Tabs and data" with a short "when" (e.g. 5s).
- [ ] Browse, then background the app via HOME / app-switch / screen-off, wait past the timer, and reopen. Data is cleared as configured (fresh state).
- [ ] "Show on app launch" and any after-inactivity screen still trigger after a genuine background for the configured duration.

**Mode switch must NOT behave like a background**
- [ ] With auto-clearing configured, switch Regular↔Fire in the tab switcher: it must NOT clear data or restart the app (a switch is not a real background→foreground).
- [ ] Rotate the device / change system font size while browsing: no spurious data-clear, no freeze.

**Automated**
- [ ] `./gradlew :app:testPlayDebugUnitTest --tests "com.duckduckgo.app.browser.state.BrowserApplicationStateInfoTest"` (8/8 pass).

---

> [!NOTE]
> **Medium Risk**
> Changes global activity lifecycle signaling used by auto data-clear and related observers; mitigated by isChangingConfigurations gating, a remote kill switch, and expanded unit tests, but real-background regression testing is important.
> 
> **Overview**
> Fixes tab switcher lag/freezes when toggling **Fire ↔ Regular** by treating `Activity.recreate()` (mode switch, rotation) as **not** a real background→foreground in `BrowserApplicationStateInfo`, so `BrowserLifecycleObserver` **`onClose()` / `onOpen()`** no longer run on every switch. Adds remote kill switch **`recreateAwareLifecycle`** (default on); when off, behavior reverts to firing the full pipeline on recreate.
> 
> **Tab switcher hardening:** blocks overlapping user mode switches while post-recreate fade-in is settling; **`BrowserModeToggleView`** reports tap vs drag; drag switches skip a misleading toggle “reverse” animation; **`onResume`** forces the tab grid visible if fade-in races leave it hidden; Fire empty state clears stuck post-recreate flags so the toggle stays usable.
> 
> **BrowserActivity** only calls **`modeSwitchRecreateSignal.markPending()`** when **`recreateAwareLifecycle`** is disabled (legacy path). Unit tests cover recreate vs real background/foreground and the kill switch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09708c473c70200d5a2e38aa4605aa7636497efb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
@@ -1345,7 +1345,11 @@ open class BrowserActivity : DuckDuckGoActivity() {
                 .flowWithLifecycle(lifecycle, Lifecycle.State.RESUMED)
                 .collect { mode ->
                     if (mode != currentBrowserMode) {
-                        modeSwitchRecreateSignal.markPending()
+                        // Only needed on the old path: with recreateAwareLifecycle on,
+                        // the recreate's onOpen is suppressed so FirstScreenHandler never runs
+                        if (!androidBrowserConfigFeature.recreateAwareLifecycle().isEnabled()) {
+                            modeSwitchRecreateSignal.markPending()
+                        }
                         recreate()
                     }
                 }
--- a/app/src/main/java/com/duckduckgo/app/browser/state/BrowserApplicationStateInfo.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/state/BrowserApplicationStateInfo.kt
@@ -19,6 +19,7 @@ package com.duckduckgo.app.browser.state
 import android.app.Activity
 import android.os.Bundle
 import com.duckduckgo.app.browser.BrowserActivity
+import com.duckduckgo.app.pixels.remoteconfig.AndroidBrowserConfigFeature
 import com.duckduckgo.browser.api.ActivityLifecycleCallbacks
 import com.duckduckgo.browser.api.BrowserLifecycleObserver
 import com.duckduckgo.di.DaggerSet
@@ -31,6 +32,7 @@ import javax.inject.Inject
 @SingleInstanceIn(AppScope::class)
 class BrowserApplicationStateInfo @Inject constructor(
     private val observers: DaggerSet<BrowserLifecycleObserver>,
+    private val androidBrowserConfigFeature: AndroidBrowserConfigFeature,
 ) : ActivityLifecycleCallbacks {
     private var created = 0
     private var started = 0
@@ -75,8 +77,8 @@ class BrowserApplicationStateInfo @Inject constructor(
     override fun onActivityStopped(activity: Activity) {
         if (started > 0) (--started)
         if (started == 0) {
-            if (activity.isChangingConfigurations) {
-                // recreate()/rotation, not a real background
+            if (activity.isChangingConfigurations && androidBrowserConfigFeature.recreateAwareLifecycle().isEnabled()) {
+                // recreate()/rotation, not a real background (kill switch: recreateAwareLifecycle)
                 stoppedForRecreate = true
             } else {
                 observers.forEach { it.onClose() }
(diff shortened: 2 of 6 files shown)


0nko:
* **#9015** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9015?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9015?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

CDRussell:
@0nko , not quite sure how but it got it into a weird state with this branch where sliding the toggle resulted in it animating in from the wrong side after i slid it into position. e.g., i slide it to the right and release and then it immediately moves it left and animates it back to the right

Screen Recording 2026-06-26 at 15 54 21

0nko:
@CDRussell The animation issue you saw should be fixed now. Ready for another round.

CDRussell (approved):
@0nko approving as this works from my testing, though I'll admit I'm a bit worried how wide the change could be if there's a problem since the changes affects the whole app, and there's no FF guarding it. 

Is there no local solution to this instead that would only require changes to `TabSwitcherActivity` (like debouncing the toggle) that might be safer?

0nko:
@CDRussell The toggle lag was a single `recreate()` running the entire app foreground/background pipeline (automatic data-clearing check, ANR supervisor, session pixels), because the app-level lifecycle observer can't tell a config-change recreate from a real background. The `TabSwitcherActivity` itself can't prevent that. There was already one workaround for the escape hatch (`ModeSwitchRecreateSignal`) but that only patched one observer, while the rest still fire. Fixing it at the source with `isChangingConfigurations` handles every observer and every `recreate()` caller in one place.

To limit the risk I added a kill-switch (`recreateAwareLifecycle`, default on), which can restore the existing behavior.

CDRussell:
> To limit the risk I added a kill-switch (`recreateAwareLifecycle`, default on), which can restore the existing behavior.

Ok, that works for me, thanks!

PR #9044, opened 2026-06-29
Compose: Add DaxSnackbar to ADS

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1211670072973944?focus=true

### Description

Adds the Compose `DaxSnackbar` to the ADS. Both overloads are thin wrappers over Material3 `Snackbar` with DDG design tokens (surface background, primary text, accent-blue action, 8dp corners per the component Figma):

- `DaxSnackbar(message, action?)` — the visual component (message + optional trailing action). For previews, the ADS catalog, or any caller that owns its own visibility.
- `DaxSnackbar(snackbarData)` — for use inside a Compose `SnackbarHost` (`SnackbarHost(hostState) { DaxSnackbar(it) }`); reads the message/action label and wires `performAction()`. Duration, placement and swipe-to-dismiss are owned by the host's `SnackbarHostState`.

Also adds the `NoRawM3SnackbarUsage` lint rule (errors on raw Material3 `Snackbar` outside the design-system modules; the legacy View `Snackbar.make` API is untouched) and a showcase in the ADS internal catalog.

### Steps to test this PR

_ADS catalog_
- [ ] Open Settings → "Android Design System Preview" → **Messaging** tab 
- [ ] Scroll to the **Snackbar** section
- [ ] Under the "COMPOSE" label, verify two `DaxSnackbar` cards render — message-only, and message + "Action"
- [ ] Confirm the above in both light and dark themes
- [ ] Tap the Compose "Action" button
- [ ] Verify an "Action pressed" toast appears

---

> [!NOTE]
> **Low Risk**
> Design-system and lint-only surface area; DaxAction relocation is import-level with no API change for consumers already using the data class.
> 
> **Overview**
> Introduces **`DaxSnackbar`** for Compose: a message-only or message + optional **`DaxAction`** snackbar styled with DDG surface, typography, and ghost-button actions (two-line cap with ellipsis). A **`SnackbarData`** overload wires **`SnackbarHost`** / **`showSnackbar`** flows without rendering M3 dismiss actions.
> 
> **`DaxAction`** moves from the message package to **`com.duckduckgo.common.ui.compose`** so snackbars and message components share one type; message call sites only change imports.
> 
> The internal ADS catalog **Snackbar** section now labels **View** vs **Compose**, keeps the legacy **`Snackbar.make`** demo, and adds themed **`DaxSnackbar`** examples (with/without action and long text). **`NoRawM3SnackbarUsage`** lint errors on raw **`androidx.compose.material3.Snackbar`** outside design-system modules (View snackbars unchanged).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c967660e8c3029c03472b778f1a920f907293d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
```diff
@@ -0,0 +1,176 @@
...
+fun DaxSnackbar(
+    message: String,
+    modifier: Modifier = Modifier,
+    action: DaxAction? = null,
+) {
+    Snackbar(
+        modifier = modifier,
+        action = action?.let { daxAction ->
+            {
+                TextButton(
```
mikescamell: ❓I'm thinking this should really be one of our own buttons. But if none are suitable then I think we should make a DaxTextButton for the DaxSnackbar, WDYT?

0nko: Good point! Fixed.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
```diff
@@ -0,0 +1,176 @@
...
+ * not rendered — the DDG snackbar design exposes only the message and an optional action.
+ *
+ * @param snackbarData The data provided by the host's [SnackbarHostState].
+ * @param modifier The [Modifier] to be applied to this snackbar.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1215496415658080/task/1211670072973944
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=6550-55049
+ */
+@Composable
+fun DaxSnackbar(
```
mikescamell, line 106: 🔍 I think we should at least have a preview of this Composable variant 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
```diff
@@ -0,0 +1,176 @@
...
+    val contentColor: Color
+        @Composable
+        @ReadOnlyComposable
+        get() = DuckDuckGoTheme.textColors.primary
+
+    val actionColor: Color
+        @Composable
+        @ReadOnlyComposable
+        get() = DuckDuckGoTheme.colors.brand.accentBlue
+}
```
mikescamell: 🔍 These properties should all be PascalCase to match Compose guidelines (yes we're being a bit inconsistent with existing Composables, I've made a task to address it)

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
```diff
@@ -0,0 +1,176 @@
...
+ * @param action Optional trailing action (e.g. "Undo"). When null, only the message is shown.
+ *
+ * Asana Task: https://app.asana.com/1/137249556945/project/1215496415658080/task/1211670072973944
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=6550-55049
+ */
+@Composable
+fun DaxSnackbar(
+    message: String,
+    modifier: Modifier = Modifier,
+    action: DaxAction? = null,
```
mikescamell, line 60: ❓Why did we opt to use DaxAction here? If we need something similar I think ideally it's specific to the Snackbar, otherwise if we really do need it here then DaxAction needs to be extracted from DaxMessage

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
```diff
@@ -0,0 +1,176 @@
...
+        DaxSnackbar(
+            message = "Text of the Snackbar can span to several lines.",
+            action = DaxAction(text = "Undo", onClick = {}),
+        )
+    }
+}
+
+@PreviewLightDark
+@Composable
+private fun DaxSnackbarLongMessagePreview() {
```
mikescamell: ❓Any idea why the bottom padding for this long snackbar is particuarly small compared to the other snackbars?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
```diff
@@ -0,0 +1,163 @@
...
+ * [SnackbarData] overload of [DaxSnackbar] for use inside a [SnackbarHost]:
+ *
+ * ```
+ * SnackbarHost(hostState) { data -> DaxSnackbar(data) }
+ * ```
+ *
+ * Reads the message and optional action label from [snackbarData] and delegates to the visual
+ * [DaxSnackbar] overload; tapping the action invokes [SnackbarData.performAction]. Trigger snackbars
+ * via [SnackbarHostState.showSnackbar]. A dismiss action ([SnackbarData]'s `withDismissAction`) is
+ * not rendered — the DDG snackbar design exposes only the message and an optional action.
```
mikescamell: 🔍 The suppression itself is correct, and matches the View `Snackbar`, which never had a dismiss "X". It's worth making the consequence explicit, since it diverges from the View world. The View `Snackbar` is swipe-dismissable by default (DDG even opts out via `NonDismissibleBehavior`), whereas a Compose  
  `SnackbarHost` has no swipe-to-dismiss, so an indefinite snackbar with no action can only be dismissed in code.                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                    
  ```suggestion                                                                                                                                                                                                                                                                                                     
   * not rendered. The DDG snackbar design exposes only the message and an optional action. Unlike the                                                                                                                                                                                                              
   * View `Snackbar`, a Compose `SnackbarHost` has no swipe-to-dismiss, so a `SnackbarDuration.Indefinite`                                                                                                                                                                                                          
   * snackbar with no `actionLabel` can only be dismissed in code via [SnackbarHostState.currentSnackbarData].                                                                                                                                                                                                      
  ```

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
0nko: Yeah, makes sense 👍. The nice thing about `DaxAction` is that it's really just a tuple that contains a title and a handler and you can prevent a situation for optional parameters where you'd supply one and not the other. I'll extract it because it's not specific to messages and can be used elsewhere.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/snackbar/DaxSnackbar.kt
0nko: Actually the [guidelines state that the snackbars should be at most 2 lines](https://m3.material.io/components/snackbar/guidelines#0ce711aa-151b-43fd-87f7-c49c7ca8ae12). So the M3 composable doesn't support 2+ and the padding is off. I added a limit to 2 lines and it looks fine.

0nko:
Thanks @mikescamell! I've addressed your comments, ready for another round.

mikescamell:
@0nko Looks like our Compose versions of the snackbar have more rounded corners than the View version. I'm guessing this is an M3 default that we need to change 👍  We might change this again in future but I think it's best we match for now 👍 

image

mikescamell (approved):
I think other than addressing https://github.com/duckduckgo/Android/pull/9044#issuecomment-5043522764 if we can, LGTM! Great work 🙌

0nko:
Just adding a note -- we discussed it on MM and the `8dp` radius is used by the [new design system guidelines](https://www.figma.com/design/jHLwh4erLbNc2YeobQpGFt/Design-System-Guidelines?node-id=8796-21590&t=GotUbPpzwDc0WfE0-4), so we'll leave it as it is.

PR #9062, opened 2026-06-30
Fire Mode: Mode-specific CookieManager

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1216110164786817?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/1207418217763355/task/1214558513236397?focus=true

### Description

Cookie access is now **mode-aware**. Each browser mode has its own cookie jar (Regular → the default WebView profile, Fire → the Fire profile), and `CookieManagerProvider` hands out the `CookieManager` for a given mode:

- `forMode(mode)` — the jar for a specific mode
- `setCookieForAllModes(url, cookie)` — seeds every mode's jar

This started as a crash fix: calling `setAcceptThirdPartyCookies` on the default `CookieManager` against a Fire-profile WebView SIGSEGVs natively. The wider change is routing every cookie call site to the jar that is correct for what it does.

**There is no live "current mode" read at any cookie call site.** The mode is either injected and frozen at construction (WebView clients) or carried with a deferred request (downloads). That is what makes the routing race-free — a mid-operation mode switch can't redirect a call to the wrong jar.

### Which jar each call site uses

**1. The WebView's own mode — `forMode(mode)`**, where `mode` is the tab/WebView's profile, obtained without a live global read:

| Component | How it gets the mode | What it does |
|---|---|---|
| `ThirdPartyCookieManager` | passed by `BrowserTabFragment` | enable/check 3rd-party cookies on the active WebView — **a profile mismatch here was the SIGSEGV** (always on the main thread) |
| `BrowserWebViewClient.flushCookies` | injected (the tab's mode; `REGULAR` in `WebViewActivity`, an in-app viewer that shares this client and stays on the default profile) | flush the tab's cookies after page events |
| `UrlExtractingWebViewClient.flushCookies` | injected | flush the AMP URL-extraction WebView's cookies |
| `InlinePdfHandler` | passed by `BrowserTabViewModel` | forward the page's cookies to the inline-PDF request |
| Duck.ai WebView surfaces (`DuckChatWebViewFragment`, `DuckChatContextualFragment`, `ChatHistoryReader`) | injected (fragment graph) | forward the chat WebView's cookies |
| `downloads/CookieManagerWrapper.getCookie` | carried on `PendingFileDownload` through WorkManager | forward cookies to the download request |

**2. The persistent (Regular) jar — `forMode(REGULAR)`** — durable, mode-independent state, or the Regular data-clearing path that must never touch the ephemeral Fire jar:

| Component | What it does |
|---|---|
| `WebViewCookieManager` + `SQLCookieRemover` / `CookieManagerRemover` | Regular Fire-button cookie clear (the Fire jar has its own burn) |
| `RealDuckChat.isStandaloneMigrationCompleted` | reads the one-time `migration_status` cookie — a durable global flag |
| `TakeoutZipDownloader` | forwards Google session cookies to authorize a Takeout export (Settings/import runs in Regular) |
| `SerpPromo` / `RedditBlockWorkaround` (reads) | checks the existing DDG/Reddit session, which lives in the persistent jar |

**3. All jars — `setCookieForAllModes()`** — seed a cookie every mode must see:

| Component | What it does |
|---|---|
| `SerpPromo.injectCookie` | seeds the Privacy Pro access token (a Regular and an active Fire SERP tab both need it) |
| `RedditBlockWorkaround` | seeds/expires the Reddit dummy session for VPN users |

Both run off the main thread (as before). The default jar is always seeded; the Fire jar is seeded only when its manager is already resolved — a non-default jar is never silently replaced by the wrong one.

### Implementation notes

- The Fire profile's `CookieManager` is resolved via `ProfileStore` (`@UiThread`) and **cached**. Off-main callers read the cached instance; off-main with a cold cache returns `null` (skip) rather than the wrong default jar — never a crash
- The third-party-cookie setup launched from `BrowserWebViewClient.onPageStarted` runs on the **WebView's `lifecycleScope`**. It is cancelled when the WebView is torn down so `setAcceptThirdPartyCookies` can't run on a stale WebView

### Steps to test this PR

Use an **internal** build with Fire mode enabled (`fireTabs` remote feature on).

_Crash fix (the original report)_
- [x] Enter Fire mode
- [x] Open a tab, tap the URL bar and browse a few sites
- [x] Verify no native crash (previously a SIGSEGV in `libwebviewchromium`)

_Cookie isolation_
- [x] In Fire mode
- [x] Sign in to / set cookies on a site 
- [x] Switch to Regular
- [x] Verify you are **not** signed in there 
- [x] Burn all tabs 
- [x] Re-enter Fire mode 
- [x] Verify the earlier Fire cookies are gone

_In-app web viewer_
- [x] Switch to Fire mode
- [x] Open an in-app page (e.g. a link that opens `WebViewActivity` from the About screen)
- [x] Verify loads, no crash

_Downloads (cookies forwarded to the download request) (Optional)_
- [x] In Fire mode, sign in so a site sets a session cookie in the Fire profile
- [x] Trigger an account-gated download
- [x] Verify it completes (would fail / bounce to login if the empty/Regular jar were used)

_Inline PDF (page cookies forwarded to the PDF fetch) (Optional)_
- [ ] In Fire mode, sign into a site
- [ ] On a site you signed into, open a PDF served behind that session
- [ ] Verify it renders inline, no login prompt

_AMP link de-cloaking (extraction runs on the active mode's profile) (Optional)_
- [ ] In Fire mode, open a Google AMP / `…/amp/…` link (e.g. `https://www-bbc-com.cdn.ampproject.org/c/s/www.bbc.com/news/articles/`)
- [ ] Verify it resolves to the canonical (non-AMP) URL, no crash
- [ ] Switch to Regular 
- [ ] Verify no session from the AMP page leaked there

---

> [!NOTE]
> **High Risk**
> Touches authentication cookies, third-party cookie policy, downloads, and WebView profile binding—core browsing and session isolation—with many call sites across the app.
> 
> **Overview**
> Replaces the singleton **`CookieManagerProvider.get()`** with **`forMode(BrowserMode)`** so Regular and Fire each use their own WebView cookie jar (Fire via **`ProfileStore`**), plus **`setCookieForAllModes`** for cookies that must land in every profile.
> 
> **WebView and tab flows** now pass a frozen **`browserMode`** into third-party cookie handling, cookie flush, AMP URL extraction (profile bind + injectable client), inline PDF fetch, and Duck.ai surfaces. **`BrowserWebViewClient`** runs third-party cookie work on the WebView **`lifecycleScope`** instead of app scope. **`WebViewActivity`** is pinned to **`REGULAR`** like Custom Tabs; **`TabSwitcherActivity`** guards **`onResume`** when views were not initialized after an aborted recreate.
> 
> **Downloads** carry **`browserMode`** on **`PendingFileDownload`** through WorkManager; download and PDF paths resolve Fire cookies on a worker with a main-thread fallback when the Fire manager is not warmed yet.
> 
> **Clearing and global work** keep **`REGULAR`** for Fire-button / SQL cookie removal and durable reads (e.g. migration cookie, Takeout). **Serp promo** and **Reddit VPN workaround** write or read per resolved profile so one jar does not overwrite another session.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 543608fcda0ee62d0bca56ce1a7be493cb2cf668. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -76,6 +76,7 @@ import com.duckduckgo.autoconsent.api.Autoconsent
 import com.duckduckgo.autofill.api.BrowserAutofill
 import com.duckduckgo.autofill.api.InternalTestUserChecker
 import com.duckduckgo.browser.api.JsInjectorPlugin
+import com.duckduckgo.browsermode.api.BrowserMode
 import com.duckduckgo.common.utils.AppUrl.ParamKey.QUERY
 import com.duckduckgo.common.utils.CurrentTimeProvider
 import com.duckduckgo.common.utils.DispatcherProvider
@@ -138,6 +139,7 @@ class BrowserWebViewClient @Inject constructor(
     private val contentScopeExperiments: ContentScopeExperiments,
     private val appSchemeInterceptionFeature: AppSchemeInterceptionFeature,
     private val forceWebViewRecompositeFeature: ForceWebViewRecompositeFeature,
+    private val browserMode: BrowserMode,
 ) : WebViewClient() {
     var webViewClientListener: WebViewClientListener? = null
     var clientProvider: ClientBrandHintProvider? = null
@@ -538,8 +540,10 @@ class BrowserWebViewClient @Inject constructor(
             handleMediaPlayback(webView, it)
             autoconsent.injectAutoconsent(webView, url)
             adClickManager.detectAdDomain(url)
-            appCoroutineScope.launch(dispatcherProvider.io()) {
-                thirdPartyCookieManager.processUriForThirdPartyCookies(webView, url.toUri())
+
+            val scope = webView.findViewTreeLifecycleOwner()?.lifecycleScope ?: return@let
+            scope.launch(dispatcherProvider.io()) {
+                thirdPartyCookieManager.processUriForThirdPartyCookies(webView, url.toUri(), browserMode)
             }
         }
         val navigationList = webView.safeCopyBackForwardList() ?: return
@@ -702,8 +706,9 @@ class BrowserWebViewClient @Inject constructor(
      */
     private fun flushCookies(webView: WebView) {
         val scope = webView.findViewTreeLifecycleOwner()?.lifecycleScope ?: return
+        val cookieManager = cookieManagerProvider.forMode(browserMode)
         scope.launch(dispatcherProvider.io()) {
-            cookieManagerProvider.forCurrentBrowserMode()?.flush()
+            cookieManager?.flush()
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -703,7 +703,7 @@ class BrowserWebViewClient @Inject constructor(
     private fun flushCookies(webView: WebView) {
         val scope = webView.findViewTreeLifecycleOwner()?.lifecycleScope ?: return
         scope.launch(dispatcherProvider.io()) {
-            cookieManagerProvider.get()?.flush()
+            cookieManagerProvider.forCurrentBrowserMode()?.flush()
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/browser/urlextraction/UrlExtractingWebView.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/urlextraction/UrlExtractingWebView.kt
@@ -21,7 +21,11 @@ import android.content.Context
 import android.webkit.WebSettings
 import android.webkit.WebView
 import com.duckduckgo.app.browser.BuildConfig
+import com.duckduckgo.browsermode.api.BrowserMode
+import com.duckduckgo.browsermode.api.WebViewModeInitializer
 import com.duckduckgo.user.agent.api.UserAgentProvider
+import logcat.LogPriority.WARN
+import logcat.logcat
 
 @SuppressLint("SetJavaScriptEnabled", "ViewConstructor")
 class UrlExtractingWebView(
@@ -29,12 +33,17 @@ class UrlExtractingWebView(
     webViewClient: UrlExtractingWebViewClient,
     userAgentProvider: UserAgentProvider,
     urlExtractor: DOMUrlExtractor,
+    webViewModeInitializer: WebViewModeInitializer,
+    browserMode: BrowserMode,
 ) : WebView(context) {
 
     var urlExtractionListener: UrlExtractionListener? = null
     lateinit var initialUrl: String
 
     init {
+        webViewModeInitializer.bind(this, browserMode).onFailure {
+            logcat(WARN) { "URL extractor WebView profile bind failed for $browserMode: ${it.message}" }
+        }
         settings.apply {
             userAgentString = userAgentProvider.userAgent()
             javaScriptEnabled = true
--- a/cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
+++ b/cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
@@ -17,11 +17,31 @@
 package com.duckduckgo.cookies.api
 
 import android.webkit.CookieManager
+import com.duckduckgo.browsermode.api.BrowserMode
 
 /** Public interface for CookieManagerProvider */
 interface CookieManagerProvider {
     /**
-     * This method returns the [CookieManager] instance. This is a singleton instance.
+     * Returns the [CookieManager] for the browser mode that is currently active.
      */
-    fun get(): CookieManager?
+    fun forCurrentBrowserMode(): CookieManager?
+
+    /**
+     * Returns the [CookieManager] for the given [mode], regardless of which mode is currently active.
+     */
+    fun forMode(mode: BrowserMode): CookieManager? = forCurrentBrowserMode()
+}
+
+/**
+ * Sets [cookieString] for [url] in every browser mode's cookie jar (the default profile plus any
+ * non-default ones), flushing each.
+ */
+fun CookieManagerProvider.setCookieForAllModes(url: String, cookieString: String) {
+    BrowserMode.entries
+        .mapNotNull { forMode(it) }
+        .distinct()
+        .forEach { cookieManager ->
+            cookieManager.setCookie(url, cookieString)
+            cookieManager.flush()
+        }
 }
--- a/cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
+++ b/cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
@@ -22,14 +22,10 @@ import com.duckduckgo.browsermode.api.BrowserMode
 /** Public interface for CookieManagerProvider */
 interface CookieManagerProvider {
     /**
-     * Returns the [CookieManager] for the browser mode that is currently active.
+     * Returns the [CookieManager] for the given [mode]. Callers always specify the mode explicitly — the
+     * tab/activity's frozen mode for WebView-bound work, or the mode carried with a deferred request.
      */
-    fun forCurrentBrowserMode(): CookieManager?
-
-    /**
-     * Returns the [CookieManager] for the given [mode], regardless of which mode is currently active.
-     */
-    fun forMode(mode: BrowserMode): CookieManager? = forCurrentBrowserMode()
+    fun forMode(mode: BrowserMode): CookieManager?
 }
 
 /**
--- a/cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/DefaultCookieManagerProvider.kt
+++ b/cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/DefaultCookieManagerProvider.kt
@@ -21,7 +21,6 @@ import android.os.Looper
 import android.webkit.CookieManager
 import androidx.webkit.ProfileStore
 import com.duckduckgo.browsermode.api.BrowserMode
-import com.duckduckgo.browsermode.api.BrowserModeStateHolder
 import com.duckduckgo.browsermode.api.FireModeAvailability
 import com.duckduckgo.browsermode.api.profileName
 import com.duckduckgo.cookies.api.CookieManagerProvider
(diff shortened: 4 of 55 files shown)


cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
```diff
@@ -17,11 +17,27 @@
...
      */
-    fun get(): CookieManager?
+    fun forMode(mode: BrowserMode): CookieManager?
+}
+
+/**
+ * Sets [cookieString] for [url] in every browser mode's cookie jar (the default profile plus any
+ * non-default ones), flushing each.
+ */
+fun CookieManagerProvider.setCookieForAllModes(url: String, cookieString: String) {
```
CDRussell, line 35: Do we need this? Since it's only called from network protection and subscriptions I'd double check with someone closer to that work to confirm this is what we want to do re cookies and modes.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -538,8 +540,10 @@ class BrowserWebViewClient @Inject constructor(
             handleMediaPlayback(webView, it)
             autoconsent.injectAutoconsent(webView, url)
             adClickManager.detectAdDomain(url)
-            appCoroutineScope.launch(dispatcherProvider.io()) {
-                thirdPartyCookieManager.processUriForThirdPartyCookies(webView, url.toUri())
+
+            val scope = webView.findViewTreeLifecycleOwner()?.lifecycleScope ?: return@let
```
CDRussell, line 544: The PR description states that: 
>The third-party-cookie setup launched from BrowserWebViewClient.onPageStarted runs on the WebView's lifecycleScope (app scope only as a fallback)

But the code isn't using `app scope` fallback?

app/src/main/java/com/duckduckgo/app/browser/urlextraction/UrlExtractingWebView.kt
```diff
@@ -21,20 +21,29 @@ import android.content.Context
...
+    webViewModeInitializer: WebViewModeInitializer,
+    browserMode: BrowserMode,
 ) : WebView(context) {
 
     var urlExtractionListener: UrlExtractionListener? = null
     lateinit var initialUrl: String
 
     init {
+        webViewModeInitializer.bind(this, browserMode).onFailure {
+            logcat(WARN) { "URL extractor WebView profile bind failed for $browserMode: ${it.message}" }
```
CDRussell, line 45: nit: bump to `ERROR`

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -702,8 +706,9 @@ class BrowserWebViewClient @Inject constructor(
      */
     private fun flushCookies(webView: WebView) {
         val scope = webView.findViewTreeLifecycleOwner()?.lifecycleScope ?: return
+        val cookieManager = cookieManagerProvider.forMode(browserMode)
```
CDRussell, line 709: should this be inside the `launch` block, too?

cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
```diff
@@ -17,11 +17,27 @@
 package com.duckduckgo.cookies.api
 
 import android.webkit.CookieManager
+import com.duckduckgo.browsermode.api.BrowserMode
 
 /** Public interface for CookieManagerProvider */
 interface CookieManagerProvider {
```
CDRussell, line 23: is there an API proposal for these changes?

cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/SQLCookieRemover.kt
```diff
@@ -42,7 +43,7 @@ import kotlin.coroutines.suspendCoroutine
 class CookieManagerRemover @Inject constructor(private val cookieManagerProvider: CookieManagerProvider) : CookieRemover {
     override suspend fun removeCookies(): Boolean {
         suspendCoroutine { continuation ->
-            cookieManagerProvider.get()?.removeAllCookies {
+            cookieManagerProvider.forMode(BrowserMode.REGULAR)?.removeAllCookies {
```
CDRussell, line 47: how do you know you've to use `REGULAR` here? this is as much of a general question as it is specific to this case, but do you have to find everywhere this class is used to then work out which mode to hardcode here?  (and is hardcoding necessary?)

cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/WebViewCookieManager.kt
```diff
@@ -72,7 +73,7 @@ class WebViewCookieManager @Inject constructor(
 
     private suspend fun storeCookie(cookie: String, host: String) {
         suspendCoroutine { continuation ->
-            cookieManager.get()?.setCookie(host, cookie) { success ->
+            cookieManager.forMode(BrowserMode.REGULAR)?.setCookie(host, cookie) { success ->
```
CDRussell: similar question to elsewhere, is this class only ever applicable to `REGULAR` mode, such that it's safe to hardcode it? and if so, is there a better way to make that clearer that it's deliberate (e.g., by injecting the right one in, or at least defining it in one place in the class with a comment)

CDRussell (commented):
Still reviewing, sharing comments so far

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/serp_promo/SerpPromo.kt
```diff
@@ -110,15 +116,12 @@ private class CookieManagerWrapperImpl constructor(
...
+        return cookieManagerProvider.forMode(BrowserMode.REGULAR)?.getCookie(url)
     }
 
-    override fun setCookie(domain: String, cookie: String) {
-        logcat { "Setting cookie $cookie for domain $domain" }
-        cookieManager?.setCookie(domain, cookie)
-        cookieManager?.flush()
+    override fun setCookieOnAllProfiles(url: String, cookieString: String) {
+        logcat { "Setting cookie $cookieString for domain $url" }
+        cookieManagerProvider.setCookieForAllModes(url, cookieString)
```
CDRussell: we've got a mixture of hardcoded profile (`REGULAR`) as well as setting cookies on all modes.

cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/DefaultCookieManagerProvider.kt
```diff
@@ -16,25 +16,53 @@
...
     @Volatile
-    private var instance: CookieManager? = null
+    private var defaultInstance: CookieManager? = null
+
+    @Volatile
+    private var fireInstance: CookieManager? = null
+
+    override fun forMode(mode: BrowserMode): CookieManager? = when (mode) {
+        BrowserMode.REGULAR -> defaultManager()
+        BrowserMode.FIRE -> if (fireModeAvailability.get().isAvailable()) fireManager() else defaultManager()
```
CDRussell, line 50: should we return `null` if `FIRE` is requested but not available?

cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
0nko: As we discussed on the call, this is just used by call sites that need to set some cookie to fix something or remember something while browsing, and with the new mode, it will now be done in Fire mode, too.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
0nko: The PR description is stale, fixed.

app/src/main/java/com/duckduckgo/app/browser/urlextraction/UrlExtractingWebView.kt
0nko: We talked about this on a call, I'd prefer we keep it as warning.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
0nko: No, because profile must resolve on the main thread.

cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/SQLCookieRemover.kt
0nko: Discussed on a call, this is Regular mode data clearing. Added a code comment.

cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/WebViewCookieManager.kt
0nko: Added a common property with a comment.

cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
0nko: Will post it shortly.

CDRussell (approved):

cookies/cookies-impl/src/main/java/com/duckduckgo/cookies/impl/DefaultCookieManagerProvider.kt
0nko: When Fire mode is unavailable the WebView runs on the default profile, so else `defaultManager()` hands back the jar that actually backs that WebView, whereas null would give callers nothing to operate on. This also keeps the cookie provider consistent with `WebStorageProvider`, which likewise returns the default store for any mode when Fire mode is off. The only caller that passes `FIRE` regardless of availability, setCookieForAllModes, collapses to a single default-jar write under either design.

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/serp_promo/SerpPromo.kt
0nko: On further inspection, the `getCookie()` was a dead code not used anywhere. I removed it.

cookies/cookies-api/src/main/java/com/duckduckgo/cookies/api/CookieManagerProvider.kt
0nko: https://app.asana.com/1/137249556945/project/1207418217763355/task/1216202300648745

PR #9126, opened 2026-07-07
Add immediate and data parameter templates to EventHub

mark-szabo:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1215978258235171?focus=true
Tech Design URL (if applicable): [Tech Design: Add immediate firing pixels to the EventHub specification](https://app.asana.com/1/137249556945/project/481882893211075/task/1215327441118832)

> Recreated from #8942, moved off my fork onto an in-repo branch so CI can run. Same commits, same diff.

### Description

Brings the Android EventHub framework up to the cross-platform spec additions made while implementing EventHub on Windows. Closes the postmortem action item: *"Implement data property and immediate templates on Android."*

- **`immediate` trigger type** — `trigger: { "type": "immediate", "source": "<event>" }`. Fires one pixel per matching `webEvent`, immediately, with no period, persistence, dedup, or foreground gating. `trigger.type` defaults to `"period"` when omitted, so existing configs are unaffected.
- **`data` parameter template** — `{ "template": "data", "dataKey": "<key>", "source": "<event>" }`. Forwards a value from `webEvent.data` as its compact JSON, percent-encoded once (RFC 3986) and sent as an already-encoded pixel parameter to avoid double-encoding. Absent keys are omitted. Immediate pixels use the triggering message's data; aggregate (period) pixels keep the last value seen on a matching `source`.

The third spec change (config-generation collapsing durations to integer `seconds`) needs no Android change — it already lands for all platforms in privacy-configuration, and Android already reads `trigger.period.seconds`.

### Steps to test this PR

_EventHub immediate + data templates_
- [ ] Run `./gradlew :event-hub-impl:testDebugUnitTest --tests "com.duckduckgo.eventhub.impl.pixels.*"` and confirm it passes, including the new `EventHubImmediatePixelTest` and `EventHubDataParameterTest` (ported from the Windows EventHub suites).
- [ ] Confirm existing counter/period EventHub behaviour is unchanged (`RealEventHubPixelManagerTest`, `EventHubConfigParserTest`).

### UI changes
| Before  | After |
| ------ | ----- |

---

> [!NOTE]
> **Medium Risk**
> Changes web-event-to-pixel firing paths and telemetry payload shape (encoded parameters); behavior is well-tested but affects statistics emission when new remote configs roll out.
> 
> **Overview**
> Aligns Android EventHub telemetry with the cross-platform spec by adding **immediate** triggers and **`data`** parameters, while refactoring config types into sealed classes.
> 
> **Immediate pixels** (`trigger.type: "immediate"` + `source`) fire on each matching `webEvent` via `fireImmediatePixels`, with no period state, dedup, or foreground gating. Omitted `type` still defaults to **period**, so existing remote configs behave the same; period pixels are excluded from `initMissingPixels` / `startNewPeriod`.
> 
> **Data parameters** (`template: "data"` + `dataKey`, optional `source` for aggregates) read `webEvent.data`, serialize to compact JSON, RFC 3986 **percent-encode** once, and pass values through `Pixel.enqueueFire` as `encodedParameters`. Period pixels keep the **last** value per matching source in `ParamState.lastDataValue`.
> 
> The parser and serializer now handle `trigger.type` / `source` and counter vs data templates. New unit tests cover immediate firing and data encoding (ported from Windows EventHub suites); existing manager/parser tests were updated for the sealed types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d1cfc8711c23ba38ac026e08febceddc980d822. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
+++ b/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
@@ -26,8 +26,13 @@ data class TelemetryPixelConfig(
 }
 
 data class TelemetryTriggerConfig(
-    val period: TelemetryPeriodConfig,
-)
+    val type: String = "period",
+    val period: TelemetryPeriodConfig = TelemetryPeriodConfig(),
+    val source: String? = null,
+) {
+    val isImmediate: Boolean get() = type == "immediate"
+    val isPeriod: Boolean get() = type == "period"
+}
 
 data class TelemetryPeriodConfig(
     val seconds: Int = 0,
@@ -41,10 +46,12 @@ data class TelemetryPeriodConfig(
 
 data class TelemetryParameterConfig(
     val template: String,
-    val source: String,
-    val buckets: Map<String, BucketConfig>,
+    val source: String? = null,
+    val dataKey: String? = null,
+    val buckets: Map<String, BucketConfig> = emptyMap(),
 ) {
     val isCounter: Boolean get() = template == "counter"
+    val isData: Boolean get() = template == "data"
 }
 
 data class BucketConfig(
@@ -52,7 +59,11 @@ data class BucketConfig(
     val lt: Int?,
 )
 
-data class ParamState(val value: Int, val stopCounting: Boolean = false)
+data class ParamState(
+    val value: Int,
+    val stopCounting: Boolean = false,
+    val lastDataValue: String? = null,
+)
 
 data class PixelState(
     val pixelName: String,
--- a/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
+++ b/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
@@ -25,13 +25,12 @@ data class TelemetryPixelConfig(
     val isEnabled: Boolean get() = state == "enabled"
 }
 
-data class TelemetryTriggerConfig(
-    val type: String = "period",
-    val period: TelemetryPeriodConfig = TelemetryPeriodConfig(),
-    val source: String? = null,
-) {
-    val isImmediate: Boolean get() = type == "immediate"
-    val isPeriod: Boolean get() = type == "period"
+sealed class TelemetryTriggerConfig {
+    /** Fires once per [period], accumulating parameter values across all matching events in the window. */
+    data class Period(val period: TelemetryPeriodConfig) : TelemetryTriggerConfig()
+
+    /** Fires immediately on each matching [source] event: no period, persistence, dedup, or foreground gating. */
+    data class Immediate(val source: String) : TelemetryTriggerConfig()
 }
 
 data class TelemetryPeriodConfig(
@@ -44,14 +43,21 @@ data class TelemetryPeriodConfig(
         get() = seconds.toLong() + minutes.toLong() * 60 + hours.toLong() * 3600 + days.toLong() * 86400
 }
 
-data class TelemetryParameterConfig(
-    val template: String,
-    val source: String? = null,
-    val dataKey: String? = null,
-    val buckets: Map<String, BucketConfig> = emptyMap(),
-) {
-    val isCounter: Boolean get() = template == "counter"
-    val isData: Boolean get() = template == "data"
+sealed class TelemetryParameterConfig {
+    /** Counts matching [source] events into [buckets]; emitted as a bucket-name pixel parameter. */
+    data class Counter(
+        val source: String,
+        val buckets: Map<String, BucketConfig>,
+    ) : TelemetryParameterConfig()
+
+    /**
+     * Forwards `webEvent.data[dataKey]` into the pixel. Immediate pixels read the triggering event's data and
+     * omit [source]; aggregate (period) pixels set [source] to select which event supplies the value.
+     */
+    data class Data(
+        val dataKey: String,
+        val source: String? = null,
+    ) : TelemetryParameterConfig()
 }
 
 data class BucketConfig(
--- a/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
+++ b/event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
@@ -136,12 +136,14 @@ class RealEventHubPixelManager @Inject constructor(
 
             for (pixelState in repository.getAllPixelStates()) {
                 if (nowMillis >= pixelState.periodEndMillis) continue
+                // Only period pixels are persisted as states; immediate pixels fire above and never reach here.
+                if (pixelState.config.trigger !is TelemetryTriggerConfig.Period) continue
 
                 val updatedParams = pixelState.params.toMutableMap()
                 var changed = false
 
                 for ((paramName, paramConfig) in pixelState.config.parameters) {
-                    if (paramConfig.isCounter && paramConfig.source == eventType) {
+                    if (paramConfig is TelemetryParameterConfig.Counter && paramConfig.source == eventType) {
                         val paramState = updatedParams[paramName] ?: ParamState(0)
                         if (paramState.stopCounting) continue
                         if (isDuplicateEvent(pixelState.pixelName, paramName, eventType, webViewId)) continue
@@ -156,7 +158,7 @@ class RealEventHubPixelManager @Inject constructor(
                         val newValue = paramState.value + 1
                         updatedParams[paramName] = paramState.copy(value = newValue)
                         logcat(VERBOSE) { "EventHub: ${pixelState.pixelName}.$paramName incremented to $newValue" }
-                    } else if (paramConfig.isData && paramConfig.source == eventType) {
+                    } else if (paramConfig is TelemetryParameterConfig.Data && paramConfig.source == eventType) {
                         // Aggregate data params keep the most recent value seen on the matching source event.
                         val paramState = updatedParams[paramName] ?: ParamState(0)
                         updatedParams[paramName] = paramState.copy(lastDataValue = extractDataParam(eventData, paramConfig.dataKey))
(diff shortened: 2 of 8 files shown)


event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -150,6 +156,11 @@ class RealEventHubPixelManager @Inject constructor(
                         val newValue = paramState.value + 1
                         updatedParams[paramName] = paramState.copy(value = newValue)
                         logcat(VERBOSE) { "EventHub: ${pixelState.pixelName}.$paramName incremented to $newValue" }
+                    } else if (paramConfig.isData && paramConfig.source == eventType) {
+                        // Aggregate data params keep the most recent value seen on the matching source event.
+                        val paramState = updatedParams[paramName] ?: ParamState(0)
+                        updatedParams[paramName] = paramState.copy(lastDataValue = extractDataParam(eventData, paramConfig.dataKey))
+                        changed = true
```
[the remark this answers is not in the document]
mark-szabo: This is by design.

mark-szabo (commented):
This is ready for review. (Tactical comment to re-trigger CI.)

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -302,25 +328,58 @@ class RealEventHubPixelManager @Inject constructor(
...
+        val compactJson = if (eventData.isNull(dataKey)) {
+            "null"
+        } else {
+            when (val value = eventData.get(dataKey)) {
+                is JSONObject, is JSONArray -> value.toString()
+                is String -> JSONObject.quote(value)
+                else -> value.toString()
+            }
+        }
+        return percentEncode(compactJson)
```
GuiltyDolphin, line 380: I've got a feeling that Android will automatically %-encode the value (seems to do this for buckets like `40+` -> `40%2B` when I look at the actual pixel fired).

Worth double checking - but if these do get %-encoded automatically then could just pass along the raw compact string.

marcosholgado: I was wondering the same, has this been tested prior raising the PR?

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
```diff
@@ -26,8 +26,13 @@ data class TelemetryPixelConfig(
 }
 
 data class TelemetryTriggerConfig(
-    val period: TelemetryPeriodConfig,
-)
+    val type: String = "period",
```
CrisBarreiro: The TD mentions 
> New trigger type `immediate` in the `eventHub` remote-config telemetry entry: `trigger: { type: "immediate", source: "<event name>" }`. No `period` object. If `type` is omitted, clients should default to` "period"`.

Taking that into account, why do we need to expose `type` and `period` at the same time? Can we have a sealed class that better enforces this separation?

That way, we could also avoid having 3 fields `type`, `isImmediate`, and `isPeriod` that are effectively the same

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
```diff
@@ -41,18 +46,24 @@ data class TelemetryPeriodConfig(
...
 data class TelemetryParameterConfig(
     val template: String,
-    val source: String,
-    val buckets: Map<String, BucketConfig>,
+    val source: String? = null,
+    val dataKey: String? = null,
+    val buckets: Map<String, BucketConfig> = emptyMap(),
 ) {
     val isCounter: Boolean get() = template == "counter"
+    val isData: Boolean get() = template == "data"
```
CrisBarreiro: Similar to my comment in `TelemetryTriggerConfig`, can we have a sealed class here? if `template` is `data`, we don't need `buckets` or `source`, and if template is `counter`, we don't need `dataKey`

This way, the model can actually enforce the valid combinations

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
```diff
@@ -125,7 +126,12 @@ class RealEventHubPixelManager @Inject constructor(
 
         if (!isEnabled()) return
 
+        val eventData = data.optJSONObject("data")
+
         appCoroutineScope.launch(pixelDispatcher) {
+            // Immediate pixels fire right away on a matching event: no period, persistence, dedup, or foreground gating.
+            fireImmediatePixels(eventType, eventData)
```
CrisBarreiro, line 133: We're not short-circuiting here, so is it possible for immediate pixels to also be handled using the branches that follow? I'm assuming `if (!pixelConfig.trigger.isPeriod) continue` prevents it, but I think we should make it more explicit to avoid future bugs

CrisBarreiro (changes requested):

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
mark-szabo: Done, both configs are now sealed classes (`Period/Immediate`, `Counter/Data`), so the model enforces the valid combinations and the `type/isImmediate/isPeriod` fields are gone.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubConfig.kt
mark-szabo: Done.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
mark-szabo: Made it explicit by adding `if (... !is TelemetryTriggerConfig.Period) continue` at the top of the loop.

event-hub/event-hub-impl/src/main/java/com/duckduckgo/eventhub/impl/pixels/EventHubPixelManager.kt
mark-szabo: Thanks both, fair questions.

On the encoding, I think it *should* be handled correctly, though it'd be worth a sanity check on-device to be sure. As far as I can tell from the pixel-sending code, the two cases go out differently: bucket/counter values ride the normal `parameters` map, which Retrofit URL-encodes for us (that's the `40+` → `40%2B` you're seeing, @GuiltyDolphin). The data-template values instead go through `encodedParameters`, which reach `PixelService.fire` as `@QueryMap(encoded = true)`, and Retrofit sends those verbatim (treating them as already-encoded). So my read is that the manual `percentEncode` is deliberately doing the encoding Retrofit skips on that path, and passing the raw compact string would leave `{ } " :` unescaped in the URL. But I've only reasoned this from the code, so it'd be great if someone with an on-device setup could confirm it looks right on the wire.

@marcosholgado, honestly no, not end-to-end. I'm not an Android dev and don't have a proper e2e environment set up, so I've only run the test projects. The unit tests do cover the encoding (`EventHubDataParameterTest` pins `{"a":true}` → `%7B%22a%22%3Atrue%7D`, strings → `%22…%22`, `null`, and absent-key omission), but that's unit-level only.

One thing worth flagging: the `data` params can't really be tested e2e yet regardless: nothing in content-scope-scripts populates `webEvent.data` today, so there's no producer for the value these params read. That would need a C-S-S change first.

The immediate trigger itself *is* e2e-testable though (it just fires on a matching event, no data needed). A minimal config to exercise it (similar to captchas on Windows in duckduckgo/privacy-configuration#5407):

```jsonc
// webDetection: emit an event when a visible reCAPTCHA is on the page
"detectors": { "captcha": { "recaptcha": {
  "match":    { "element": { "selector": [".g-recaptcha", "iframe[src*='recaptcha']"], "visibility": "visible" } },
  "triggers": { "auto": { "state": "enabled", "when": { "intervalMs": [500, 1000, 3000, 5000] } } },
  "actions":  { "fireEvent": { "state": "enabled", "type": "captcha_recaptcha" } }
}}}
```
```jsonc
// eventHub: fire an immediate pixel on that event
"webTelemetry_captcha_recaptcha": {
  "state": "enabled",
  "trigger": { "type": "immediate", "source": "captcha_recaptcha" },
  "parameters": {}
}
```

Then load a page showing a visible reCAPTCHA and the pixel should fire immediately.

For a bit of context on why this is coming from me rather than a task: I originally reached out to @cmonfortep to help get this built on Android to match the Windows implementation and the updated spec, not just for a review. I could have filed an Asana task for the Android team, but figured it'd be lower overhead for everyone if I did the implementation with Claude, sanity-checked it since I have the context, and opened the PR for you all to take the rest of the way. Thanks for the careful look!

CrisBarreiro: @mark-szabo, regardless of whether the implementation was done with Claude, we should have confidence in what we're suggesting to merge, and have the same levels of scrutiny than if not using it. That includes being able to fully test our own changes, and give reviewers clear steps to validate E2E. Otherwise, we're not only decreasing our safety layer for merging, but we're increasing the reviewer's overload, since, as you said, you're the one most familiar with the context, but we aren't that much, and gaining that context might as well take us the same time that was saved by getting someone else opening the PR for us. 

> One thing worth flagging: the data params can't really be tested e2e yet regardless: nothing in content-scope-scripts populates webEvent.data today, so there's no producer for the value these params read. That would need a C-S-S change first.

If we need a C-S-S change first, then I think we should wait for that C-S-S change to be available, and then test E2E before merging. This can be via a C-S-S release, or pointing towards a C-S-S branch

So TL;DR: For this PR to be merged, it needs to be tested E2E by the author, as well as contain a defined set of steps to test it.

mark-szabo:
Closing this one. After talking with @cmonfortep, we agreed the right path is a help request so the Android team can staff and own the implementation and end-to-end testing, in line with the expectations.

[Help Request (O-B): Bring Android EventHub up to the current cross-platform spec (immediate + data templates)](https://app.asana.com/1/137249556945/project/1213311565674947/task/1216509643482854?focus=true)

PR #9190, opened 2026-07-15
Quick Win: Android: Disable ATB by default for pixels

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214749231578703/task/1214650222771756?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/481882893211075/task/1216528078778145

### Description
Adds an opt-in mechanism so that ATB is no longer attached to most pixels, addressing the privacy goal of not sending the ATB cohort identifier on telemetry that doesn't need it.

Behaviour, gated by a new remote feature `pixelSenderAtbOptIn (@Toggle.DefaultValue(DefaultFeatureValue.TRUE))`:

  - Flag ON (default): ATB is attached only to pixels declared by a PixelRequiringAtbPlugin. Every other pixel is sent without ATB.
  - Flag OFF: legacy behaviour — ATB is attached to every pixel (and later removed for most pixels).

Key changes:

  - PixelRequiringAtbPlugin (new plugin point in common-utils) — features declare, via a multibinding, the pixel names that must keep ATB.
  - AnrCrashPixelsRequiringAtb — opts the ANR and crash pixels into ATB (`m_anr_exception`, `m_d_ac_g`, `m_app_crashed_on_verified_play_store_install`), referenced by their enum constants so the list stays in sync with the pixel definitions.
  - PixelService.fireWithoutAtb — new Retrofit endpoint (same URL template as fire, minus the atb query param) used for non-opted-in pixels.
  - RxPixelSender — a shouldSendAtb(pixelName) gate routes each pixel to fire / fireWithoutAtb on both the immediate and the pending/offline send paths. The opt-in prefix set is built once (by lazy) so there's no per-pixel allocation. PixelSenderFeature is injected via dagger.Lazy to break a DI cycle (the remote-feature proxy depends on VariantManager, which transitively depends on the PixelSender).
  - Removed VerifiedCrashPixelDataCleaner — it existed solely to strip ATB from the verified-install crash pixel via PixelParamRemovalPlugin. That pixel is now opted into ATB, so the cleaner directly contradicted the new behaviour and has been deleted.

Tests: 
- RxPixelSenderTest covers flag on/off, opted-in vs not, the pending-pixel path (with and without ATB), prefix vs. contains matching, and matching across multiple contributed plugins.
- AnrCrashPixelsRequiringAtbTest asserts the real ANR/crash pixel names are covered and that an unrelated pixel is not.

### Steps to test this PR

Filter logcat by "Pixel url request:"

  - [ ] Flag ON (default) — non-opted pixel has no ATB: Fresh internal build, trigger any ordinary pixel (e.g. ml - app launch, m_addressbar_focus_ntp). Confirm the request has no atb query param.
  - [ ] Flag ON — opted-in pixel keeps ATB: Force an ANR or crash (or fire the crash pixels), let the offline pixels flush on next app start. Confirm m_anr_exception / m_d_ac_g include atb=<atb><variant> (You can't really test m_app_crashed_on_verified_play_store_install requests).
  - [ ] Flag OFF — legacy behaviour: Disable pixelSenderAtbOptIn, trigger an ordinary pixel (e.g. ml - app launch, m_addressbar_focus_ntp), confirm the atb param is present again on all pixels.
  - [ ] Pending/offline path: Enqueue a non-opted pixel while offline, then reconnect and foreground the app; confirm it drains without ATB.
  - [ ] Unit tests are passing.

### NO UI changes

---

> [!NOTE]
> **Medium Risk**
> Changes core pixel sending and what cohort data leaves the device on most events; behavior is gated and tested but misconfigured opt-in lists could drop ATB from pixels that still need it.
> 
> **Overview**
> Adds a **pixel ATB opt-in** path so most telemetry can be sent **without** the ATB cohort identifier when remote feature **`pixelSenderAtbOptIn`** is on (default). With the flag on, only pixel names declared by **`PixelRequiringAtbPlugin`** (prefix/`startsWith` match) use **`PixelService.fire`**; everything else uses new **`fireWithoutAtb`**. Flag off keeps legacy behavior: ATB on every pixel.
> 
> **`RxPixelSender`** routes immediate and pending/offline sends through **`shouldSendAtb`**, aggregates plugin prefixes once, and injects **`PixelSenderFeature`** via **`Lazy`** to avoid a DI cycle. **`AnrCrashPixelsRequiringAtb`** replaces the old **`VerifiedCrashPixelDataCleaner`** (ATB stripping for verified-install crash) and opts ANR/crash pixels into ATB via enum pixel names.
> 
> Tests cover flag on/off, opted-in vs not, pending flush, prefix vs contains matching, and multiple plugins; **`AnrCrashPixelsRequiringAtbTest`** locks the ANR/crash pixel list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ca12738ff70f5b97b8d21ff429628b6a551eec2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


anikiki:
* **#9190** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9190?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9190?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

marcosholgado:
Don't really have a strong opinion but was wondering if would make sense to also add the ff to `PixelParamRemovalInterceptor` as part of `val shouldRemoveAtb = atbs.any { pixel.startsWith(it) }` to return a false earlier if the FF is enabled because the interceptor will also try to remove atb even when we won't be sending it with the new noAtb API.

anikiki:
> Don't really have a strong opinion but was wondering if would make sense to also add the ff to `PixelParamRemovalInterceptor` as part of `val shouldRemoveAtb = atbs.any { pixel.startsWith(it) }` to return a false earlier if the FF is enabled because the interceptor will also try to remove atb even when we won't be sending it with the new noAtb API.

@marcosholgado I lean towards leaving it as is. I agree we can return earlier in the `PixelParamRemovalInterceptor`. However, that would mean wiring `PixelSenderFeature` into the interceptor just to skip a noop, and it feels like more coupling than it's worth.

marcosholgado (approved):
LGTM

PR #9207, opened 2026-07-16
Quick Win: Android: Reenabling Microphone permissions in Settings doesn’t grant permissions to DDG app 

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200581511062568/task/1216071378153650?focus=true

### Description

Fixes a bug where voice search kept showing the "enable microphone access in Settings" dialog even after the user had re-granted the permission from system settings.

**Root cause:** The decision to show the "go to Settings" dialog was gated by a persisted `permissionDeclinedForever` flag. That flag was written `true` when a permission request was denied while `shouldShowRequestPermissionRationale` was false, and was **never reset**. When the permission was later re-enabled as "Ask every time", the OS still reports it as _not_ _currently granted_, so the app fell into the flag-gated path and trusted its stale cache instead of re-checking the OS. Only "Allow every time" appeared to fix it, because a granted permission short-circuits before the flag is ever read.

**Fix:** Removed the sticky `permissionDeclinedForever` flag entirely and now decide from live OS state:

- `launch()` always re-requests the permission when the rationale was previously accepted, so the OS can surface its prompt (or auto-deny). It no longer consults any persisted decline flag.
- The "no mic access" dialog is now driven solely by the permission-result callback: shown only when a request comes back denied **and** `shouldShowRequestPermissionRationale` is now false (the live "permanently denied" signal). This self-heals when the user changes the permission from system settings.
- The settings deep-link now uses the shared `launchApplicationInfoSettings()` util, and the dialog is skipped if the host Activity is finishing/destroyed (avoids a `BadTokenException`).

### Steps to test this PR

- [ ]  Install from this branch.
- [ ]  Enable voice search and grant the microphone permission when prompted.
- [ ]  Go to system Settings → App permissions and **disable** the microphone permission for DDG.
- [ ]  Re-enable the permission, choosing **"Ask every time"**.
- [ ]  Return to the app and tap the voice search icon.
- [ ]  **Expected:** the system permission prompt appears
- [ ]  Regression check: permanently denied still works: deny the mic permission twice (until "Ask every time"/"Allow" no longer prompts), tap the voice icon, and confirm the "enable microphone access in Settings" dialog appears and its button opens app settings.

### NO UI changes

---

> [!NOTE]
> **Medium Risk**
> Touches microphone permission UX and persistence; behavior changes for users who previously hit the sticky decline path, but logic is narrower and OS-driven with added crash guard.
> 
> **Overview**
> Fixes voice search staying on the “enable microphone in Settings” flow after the user re-grants mic access (e.g. **Ask every time**) from system settings.
> 
> **`permissionDeclinedForever` is removed** from the datastore/repository and is no longer written on deny. **`launch()`** no longer branches on that flag: after the privacy rationale was accepted it always fires the system permission request again. The **no-mic-access** dialog is shown only from the permission result when the request is denied and **`shouldShowRequestPermissionRationale`** is false (live “don’t ask again” signal).
> 
> **`showNoMicAccessDialog`** skips showing if the activity is finishing/destroyed, and app settings opens via **`launchApplicationInfoSettings()`** instead of a hand-rolled intent. Tests are updated for the new flow, including regression coverage for re-request after settings changes and the destroyed-activity case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebbfd1b9a8ea2534f8bde35f4b45469b1c75bb89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/voice-search/voice-search-impl/src/main/java/com/duckduckgo/voice/impl/PermissionRequest.kt
+++ b/voice-search/voice-search-impl/src/main/java/com/duckduckgo/voice/impl/PermissionRequest.kt
@@ -18,11 +18,10 @@ package com.duckduckgo.voice.impl
 
 import android.app.Activity
 import android.content.Context
-import android.content.Intent
-import android.net.Uri
-import android.provider.Settings
 import androidx.activity.result.ActivityResultCaller
+import androidx.appcompat.app.AppCompatActivity
 import com.duckduckgo.app.statistics.pixels.Pixel
+import com.duckduckgo.common.utils.extensions.launchApplicationInfoSettings
 import com.duckduckgo.di.scopes.ActivityScope
 import com.duckduckgo.voice.impl.ActivityResultLauncherWrapper.Action.LaunchPermissionRequest
 import com.duckduckgo.voice.impl.ActivityResultLauncherWrapper.Request
@@ -49,10 +48,6 @@ class MicrophonePermissionRequest @Inject constructor(
     private val activityResultLauncherWrapper: ActivityResultLauncherWrapper,
     private val permissionRationale: PermissionRationale,
 ) : PermissionRequest {
-    companion object {
-        private const val SCHEME_PACKAGE = "package"
-    }
-
     private lateinit var voiceSearchDisabled: () -> Unit
 
     override fun registerResultsCallback(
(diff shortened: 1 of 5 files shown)


anikiki:
* **#9207** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9207?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9207?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

voice-search/voice-search-impl/src/main/java/com/duckduckgo/voice/impl/PermissionRequest.kt
```diff
@@ -63,44 +58,36 @@ class MicrophonePermissionRequest @Inject constructor(
...
-                voiceSearchPermissionDialogsLauncher.showPermissionRationale(
-                    activity,
-                    { handleRationaleAccepted() },
-                    { handleRationaleCancelled(activity) },
-                )
-            }
         }
     }
 
-    private fun Context.launchDuckDuckGoSettings() {
```
anikiki, line 99: This was removed as we already had `launchApplicationInfoSettings()` that could be used.

karlenDimla (approved):

PR #9233, opened 2026-07-17
Investigate RequestBlocklistTest failure on the Metro nightly

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200581511062568/task/1216654721701993?focus=true

### Description

RequestBlocklistTest was intermittently failing on the nightly. Root cause: the requestBlocklist rules are not bundled with the app — they arrive via an asynchronous remote privacy-config download and are then loaded into memory off the main thread. On a slow boot, the test page could issue its requests before blocking was live, so the "Blocked …" checks saw the resources load and the page reported status: failed. There's no observable signal for in-memory rule readiness, so the test now waits on the real end condition instead of assuming a single page load is enough.

Changes (test infrastructure only — no production code):

- RequestBlocklistTest: retry up to MAX_ATTEMPTS; on a failing run, wait for the config download to land, reload the page so requests are re-issued with blocking active, and re-check. On final failure it now reports every mismatching check (description, expected, actual) instead of an opaque aggregate status. window.results is cleared synchronously before each reload so a retry can't scrape the previous run.
- JsObjectIdlingResource: added a configurable timeout and a failOnTimeout flag. RequestBlocklistTest opts out of the throw-on-timeout behavior so a slow attempt degrades into a retry rather than crashing the test from the resource's Handler. The default (true) preserves existing behavior for the other callers (GpcTest, FingerprintProtectionTest).

### Steps to test this PR

- [ ] Confirm the test passes, and re-run it a few times to check it's stable
- [ ] (Optional) Watch logcat for the per-attempt breadcrumb to see the retry behavior: adb logcat | grep "RequestBlocklistTest: attempt"

### NO UI changes

---

> [!NOTE]
> **Low Risk**
> Android instrumentation test changes only; production privacy/blocklist behavior is unchanged and default JsObjectIdlingResource timeout behavior remains for other tests.
> 
> **Overview**
> Fixes intermittent **RequestBlocklistTest** failures on nightly runs where the privacy test page could run before remotely downloaded request-blocklist rules were loaded in memory.
> 
> **RequestBlocklistTest** now retries up to three times: wait for `window.results`, evaluate checks, and on failure wait for the config download, clear `window.results`, reload the WebView, and try again. Assertions list every mismatched check (description, expected, actual) instead of only an overall status.
> 
> **JsObjectIdlingResource** gains configurable `timeoutMillis` and `failOnTimeout` (default still throws on timeout). This test uses `failOnTimeout = false` so a slow page load can fall through to the retry loop instead of failing from the idling resource’s Handler.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d241e983b953f569a7704541d7749b05e43abe96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
@@ -31,13 +31,13 @@ import com.duckduckgo.espresso.privacy.preparationsForPrivacyTest
 import com.duckduckgo.privacy.config.impl.network.JSONObjectAdapter
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
-import java.util.concurrent.CountDownLatch
-import java.util.concurrent.TimeUnit
 import logcat.logcat
 import org.junit.After
 import org.junit.Assert.assertTrue
 import org.junit.Rule
 import org.junit.Test
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
 
 class RequestBlocklistTest {
 
--- a/app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
+++ b/app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
@@ -16,12 +16,10 @@
 
 package com.duckduckgo.espresso
 
-import android.view.View
 import android.webkit.WebView
+import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.IdlingRegistry
-import androidx.test.espresso.IdlingResource
-import androidx.test.espresso.UiController
-import androidx.test.espresso.ViewAction
+import androidx.test.espresso.matcher.ViewMatchers.isRoot
 import androidx.test.espresso.web.model.Atoms
 import androidx.test.espresso.web.sugar.Web
 import androidx.test.ext.junit.rules.activityScenarioRule
@@ -33,10 +31,11 @@ import com.duckduckgo.espresso.privacy.preparationsForPrivacyTest
 import com.duckduckgo.privacy.config.impl.network.JSONObjectAdapter
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
-import org.hamcrest.Matcher
-import org.hamcrest.Matchers.instanceOf
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import logcat.logcat
 import org.junit.After
-import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
 import org.junit.Rule
 import org.junit.Test
 
@@ -51,11 +50,11 @@ class RequestBlocklistTest {
         ),
     )
 
-    private val registeredResources = mutableListOf<IdlingResource>()
-
     @After
     fun tearDown() {
-        registeredResources.forEach { IdlingRegistry.getInstance().unregister(it) }
+        IdlingRegistry.getInstance().resources.toList().forEach {
+            IdlingRegistry.getInstance().unregister(it)
+        }
     }
 
     @Test @InternalPrivacyTest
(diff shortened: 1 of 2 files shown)


anikiki:
* **#9233** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
```diff
@@ -66,28 +65,76 @@ class RequestBlocklistTest {
...
+        // This only compensates for config-download *timing*. A config that genuinely lacks rules
+        // for the test domain, or the test site being unreachable, will (correctly) still fail
+        // once the attempts are exhausted.
+        var attempts = 0
+        var overallStatus: String? = null
+        var checks: List<RequestBlocklistResult> = emptyList()
+        var failures: List<RequestBlocklistResult> = emptyList()
+
+        while (true) {
+            attempts++
```
landomen, line 86: nit: Could be simplified using `for (attempt in 1..MAX_ATTEMPTS)`

app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
```diff
@@ -97,20 +144,16 @@ class RequestBlocklistTest {
...
-    }
-
     companion object {
         const val SCRIPT = "return window.results;"
-        const val SUCCESS = "success"
+
+        private const val MAX_ATTEMPTS = 3
+
+        // Time given to the background config download between attempts before reloading.
+        private const val DOWNLOAD_WAIT_MILLIS = 30_000L
```
landomen, line 153: Wondering if it would be better to have a higher number of attempts but lower wait time? Because right now if download happens quickly we still wait for 30 seconds either way.

app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
```diff
@@ -66,28 +65,76 @@ class RequestBlocklistTest {
...
+            IdlingRegistry.getInstance().unregister(resultsIdling)
+
+            val testJson: TestJson? = getTestJson(results.toJSONString())
+            overallStatus = testJson?.value?.status
+            checks = testJson?.value?.results.orEmpty()
+            failures = checks.filter { it.actual != it.expected }
+            logcat { "RequestBlocklistTest: attempt $attempts/$MAX_ATTEMPTS, status=$overallStatus, ${failures.size}/${checks.size} checks failing" }
+
+            if (checks.isNotEmpty() && failures.isEmpty()) break
+            if (attempts >= MAX_ATTEMPTS) break
```
landomen, line 108: One downside is that if the test does fail for valid reason it will take a lot more time + we have retry setup on global level for all tests so this will run twice.

landomen (approved):
Couple of non-blocking comments, otherwise looks good

app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
anikiki: Yes, that's true. I'm not concerned about this yet.

app/src/androidTest/java/com/duckduckgo/espresso/RequestBlocklistTest.kt
anikiki: I tried that and it worked better with bigger wait time (30s, I tried other values).

anikiki:
Thanks @landomen for checking and for the suggestions. I'll merge as is for now and will come back to it if we need to do the tweaks.

PR #9249, opened 2026-07-21
Add Compose list item family (one line, two line, settings)

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1215496415658080/task/1211659112661214?focus=true

### Description

Adds the Compose list item family to the design system, following on from the buttons and text inputs.

- `DaxOneLineListItem`, `DaxTwoLineListItem` and `DaxSettingsListItem`. All are thin variants over an internal `DaxListItem` core with receiver-scoped leading/trailing slots, so callers can only place content the design system allows
- Leading icons support size and background independently, with an optional tint. Trailing content supports decorative or clickable icons, a switch, a button and a status indicator. There's also an optional inline pill via `pillText`
- `DaxPill`, a standalone status pill matching the View `DaxYellowPill`. The caller supplies the (localised) label so the design system owns no product copy
- Two new lint rules: one keeps the leading/trailing slots to their scoped composables, the other requires `primaryTextColor`/`secondaryTextColor` to resolve to `DuckDuckGoTheme` semantic colours
- The internal ADS showcase now shows each View list item next to its Compose equivalent, captioned "View" and "Compose". Rows the Compose API intentionally does not cover are flagged "View (no Compose equivalent)"
- A new `PreviewSurface` for full-width components that carry their own internal padding
- A Maestro test asserts the new showcase sections

Intentional divergences from the View components and Figma: disabled alpha is 0.4 per Figma rather than the 0.38 in the legacy XML, there is no leading icon click handler and `DaxSettingsListItem` has no `AnnotatedString` overload, since no call sites need them. These will also be captured in the design system Asana entry for list items.

### Steps to test this PR

_List items showcase_
- [ ] Install an internal debug build
- [ ] Go to Settings -> Android Design System Preview -> LIST ITEMS
- [ ] In the Single Line Items, Two Line Items and Settings sections, check each Compose row renders the same as the View row above it, in light and dark mode
- [ ] Check the switches and trailing icons in the Compose rows respond to taps

_Lint rules_
- [ ] Run `./gradlew :lint-rules:test` and check the new detector tests pass

_Maestro_
- [ ] Run `maestro test .maestro/ads_preview_flows/1-_design-system-components.yaml` with the app installed

### UI changes
| Before  | After |
| ------ | ----- |

---

> [!NOTE]
> **Low Risk**
> Changes are confined to design-system components, internal showcase, lint, and Maestro; no production settings or browser flows are migrated to the new APIs in this diff.
> 
> **Overview**
> Introduces the **Compose list item family** (`DaxOneLineListItem`, `DaxTwoLineListItem`, `DaxSettingsListItem`) built on a shared internal `DaxListItem` with scoped leading, trailing, and inline slots (icons/images, switches, status indicators, pills). Adds **`DaxPill`** for Beta/New-style labels and extends **`DaxIconButton`** with optional `iconSize` for list-item trailing icons.
> 
> **Lint** now warns when list items use arbitrary text colors or put non-design-system content in those slots. **`DialogsFragment`** bottom-sheet samples adopt the new `primaryText` / `leadingContent` API.
> 
> The **internal design-system preview** pairs each legacy View row with a Compose equivalent (captioned View vs Compose), passes dark theme into those holders, and documents View-only cases (e.g. deprecated 16dp/40dp leading sizes). **`PreviewSurface`** supports full-width previews; Maestro scroll/assert steps were updated for the longer LIST ITEMS section.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29bfbde85c52529abc0fb8b1620c85bc6ff708c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mikescamell:
* **#9249** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9249?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9249?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/ComponentViewHolder.kt
```diff
@@ -537,11 +548,33 @@ sealed class ComponentViewHolder(val view: View) : RecyclerView.ViewHolder(view)
...
+                R.id.composeOneLineDisabled to { ComposeOneLineDisabled() },
+                R.id.composeOneLineDestructive to { ComposeOneLineDestructive() },
+                R.id.composeOneLineLongText to { ComposeOneLineLongText() },
+                R.id.composeOneLineLongTextTruncated to { ComposeOneLineLongTextTruncated() },
+                R.id.composeOneLineNewPill to { ComposeOneLineNewPill() },
+                R.id.composeOneLineExtras to { ComposeOneLineExtras() },
+            )
+            composeContent.forEach { (id, content) ->
+                view.setupThemedComposeView(id, isDarkTheme) { Column(modifier = Modifier.fillMaxWidth()) { content() } }
+            }
```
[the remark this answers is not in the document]
mikescamell: This is the existing pattern for every Compose showcase holder in this screen (switches, checkboxes, radios, info panels all call setContent from bind() on develop). Keeping demo toggle state across recycling isn't a goal of the showcase, so I'm leaving it consistent with the rest of the file rather than special-casing the list items.

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/listitems/ComponentListItemsElementsFragment.kt
```diff
@@ -27,6 +27,13 @@ import com.duckduckgo.common.ui.internal.ui.component.ComponentFragment
...
     override fun getComponents(): List<Component> {
-        return listOf(SECTION_HEADER_LIST_ITEM, SINGLE_LINE_LIST_ITEM, TWO_LINE_LIST_ITEM, SETTINGS_LIST_ITEM, MENU_ITEM, POPUP_MENU_ITEM)
+        return listOf(
+            SECTION_HEADER_LIST_ITEM,
+            SINGLE_LINE_LIST_ITEM,
+            TWO_LINE_LIST_ITEM,
+            SETTINGS_LIST_ITEM,
+            MENU_ITEM,
+            POPUP_MENU_ITEM,
+        )
```
mikescamell: Why was this re-formatted?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
```diff
@@ -0,0 +1,94 @@
...
+@Composable
+fun DaxPill(
+    text: String,
+    modifier: Modifier = Modifier,
+) {
+    Box(
+        modifier = modifier
+            .height(DaxPillDefaults.Height)
+            .clip(RoundedCornerShape(DaxPillDefaults.CornerRadius))
+            .background(Yellow50)
```
landomen: This uses Yellow50 directly, but should we go through `DuckDuckGoPillColors` to match other components?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
```diff
@@ -0,0 +1,94 @@
...
+            .padding(horizontal = DaxPillDefaults.HorizontalPadding),
+        contentAlignment = Alignment.Center,
+    ) {
+        Text(
+            text = text.uppercase(),
```
landomen, line 69: This could in theory be the callers responsibility, but considering existing `DaxYellowPill` already does this it's fine.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.duckduckgo.common.ui.compose.theme.asTextStyle
+import com.duckduckgo.common.ui.compose.tools.PreviewBox
+
+/**
+ * A reusable yellow pill for short status labels (e.g. "Beta", "New").
+ *
+ * Fixed palette in both light and dark (DuckDuckGo yellow background, black text) — it is not
+ * theme-tokened and draws its own text colour, matching the View [DaxYellowPill]. Callers supply
+ * their own (localised) label.
+ *
```
landomen, line 48: Missing Figma and Asana links

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
```diff
@@ -0,0 +1,253 @@
...
+import com.duckduckgo.common.ui.compose.tools.PreviewSurface
+import com.duckduckgo.mobile.android.R
+
+/**
+ * Two-line list item for the DuckDuckGo design system.
+ *
+ * The primary label defaults to a single ellipsised line; the secondary caption is unbounded (wraps),
+ * matching the View. Sibling overloads accept [AnnotatedString] for HTML / inline-styled text.
+ *
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=6032-13665
```
landomen, line 44: Missing Asana link like other components have

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
```diff
@@ -0,0 +1,253 @@
...
+ * @param enabled Whether the item is enabled and interactive; defaults to true.
+ */
+@Composable
+fun DaxTwoLineListItem(
+    primaryText: String,
+    secondaryText: String,
+    modifier: Modifier = Modifier,
+    primaryTextColor: Color = DuckDuckGoTheme.textColors.primary,
+    secondaryTextColor: Color = DuckDuckGoTheme.textColors.secondary,
+    pillText: String? = null,
```
landomen, line 67: Would it make sense to expose this as a slot where callers put `DaxPill` vs a `String`?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
```diff
@@ -0,0 +1,253 @@
...
+ * @param onClick Optional click handler; when set the item becomes clickable.
+ * @param onLongClick Optional long-click handler.
+ * @param primaryMaxLines Maximum lines for the primary label; defaults to 1.
+ * @param secondaryMaxLines Maximum lines for the secondary caption; defaults to [Int.MAX_VALUE].
+ * @param enabled Whether the item is enabled and interactive; defaults to true.
+ */
+@Composable
+fun DaxTwoLineListItem(
+    primaryText: String,
+    secondaryText: String,
```
landomen, line 63: Would it make sense to expose this as a slot where callers put DaxText vs a String? That also removes the need for `primaryTextColor` and `secondaryTextColor`. 

I get the benefit of keeping it like now, it also matches existing XML. Moving to a more slot-based component makes it more Compose-like, but also adds more work for the caller and enables invalid usages (for which we'd need more lint rules). 

I'm okay either way, just want to open a discussion since we went with more open slot approach for some other components.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxSettingsListItem.kt
```diff
@@ -0,0 +1,110 @@
...
+import androidx.compose.ui.unit.dp
+import com.duckduckgo.common.ui.compose.Status
+import com.duckduckgo.common.ui.compose.tools.PreviewSurface
+import com.duckduckgo.mobile.android.R
+
+/**
+ * Settings row: one-line list item whose trailing element is a status indicator.
+ *
+ * The trailing slot is fixed to a [DaxListItemTrailingScope.StatusIndicator] driven by [status]; it is not a free slot like the other variants.
+ *
```
landomen, line 35: Missing Figma and Asana links

android-design-system/design-system-internal/src/main/java/com/duckduckgo/common/ui/internal/ui/component/ComponentViewHolder.kt
```diff
@@ -801,3 +880,577 @@ sealed class ComponentViewHolder(val view: View) : RecyclerView.ViewHolder(view)
...
+    )
+}
+
+@Composable
+private fun ComposeTwoLineSwitchTruncated() {
+    ComposeCaption()
+    var checked by remember { mutableStateOf(false) }
+    DaxTwoLineListItem(
+        primaryText = "Two Line Item Two Line Item Two Line Item Two Line Item",
+        secondaryText = "In disabled state",
```
landomen, line 1254: Technically, this one is not in a disabled state since there's no `enabled=false` 🕵️

landomen (approved):
Left a few non-blocking comments otherwise it looks good

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+import com.duckduckgo.common.ui.compose.Status
+import com.duckduckgo.common.ui.compose.button.DaxGhostButton
+import com.duckduckgo.common.ui.compose.button.DaxIconButton
+import com.duckduckgo.common.ui.compose.pill.DaxPill
+import com.duckduckgo.common.ui.compose.switch.DaxSwitch
+import com.duckduckgo.common.ui.compose.text.DaxText
+import com.duckduckgo.common.ui.compose.theme.DuckDuckGoTheme
+
+@Composable
+internal fun DaxListItem(
```
catalinradoiu, line 82: We should add KDocs for this composable as well, even if it’s internal

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+    Row(
+        modifier = modifier
+            .fillMaxWidth()
+            .then(interaction)
+            .heightIn(min = minHeight)
+            .padding(horizontal = DaxListItemDefaults.HorizontalPadding),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        if (leadingContent != null) {
+            leadingScope.leadingContent()
```
catalinradoiu, line 123: This is the way to enforce that the leading content is part of `DaxListItemLeadingScope`, right?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+        }
+        if (trailingContent != null) {
+            Spacer(Modifier.width(DaxListItemDefaults.TrailingGap))
+            trailingScope.trailingContent()
+        }
+    }
+}
+
+@Stable
+class DaxListItemLeadingScope internal constructor(private val parentEnabled: Boolean) {
```
catalinradoiu, line 173: Should we add KDocs here as well?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+                contentDescription = contentDescription,
+                tint = tint ?: Color.Unspecified,
+                modifier = Modifier.size(iconDp),
+            )
+        }
+    }
+}
+
+@Stable
+class DaxListItemTrailingScope internal constructor(private val parentEnabled: Boolean) {
```
catalinradoiu, line 232: Should we add KDocs?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/**
+ * Two-line list item with [AnnotatedString] primary **and** secondary text — the HTML-in-both-lines case.
+ *
+ * @see DaxTwoLineListItem for the parameters.
+ */
+@Composable
+fun DaxTwoLineListItem(
+    primaryText: AnnotatedString,
```
catalinradoiu, line 138: Just making sure, we don’t need a variant with primary text as `AnnotatedString` and secondary text as `String`, right? From a design perspective I don’t think it makes sense but just want to make sure that we didn’t miss this component variant.

catalinradoiu (approved):
Just some small comments, but overall it looks good to me.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+            enabled = enabled,
+            onClick = { onClick?.invoke() },
+            onLongClick = onLongClick,
+        )
+    } else {
+        Modifier
+    }
+    val leadingScope = DaxListItemLeadingScope(enabled)
+    val trailingScope = DaxListItemTrailingScope(enabled)
+    Row(
```
mikescamell, line 114: No space between Row and declared properties. It would be easier to read if there were. Is there anything we could apply to Spotless to enforce this?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+                onClick = onClick,
+                iconPainter = painter,
+                contentDescription = contentDescription,
+                enabled = effectiveEnabled,
+                modifier = modifier
+                    .size(iconDp)
+                    .alpha(if (effectiveEnabled) 1f else DaxListItemDefaults.DisabledAlpha),
+            )
+        } else {
+            androidx.compose.material3.Icon(
```
mikescamell: Why do we have the full import here? It should just be `Icon`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
mikescamell: I don't believe there are any usages of ListItem's without text, therefore I think I prefer to lock it down and only allow Strings to enforce the API without the need of Lint. If at some point we need a ListItem that at it's core does not have text, then I think it would be appropriate to move to slots as you mention 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
mikescamell: All [examples](https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=6032-13783&t=yoKio931hpgujxWL-4) of Pill in the design system use capitalisation, so I'd like to enforce it and if something changes, we can update it 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
mikescamell: Yes I agree, same for Text color.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
mikescamell: Yeah are there any examples of this in the codebase where the current View version's primary text needs an Annotated string?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
mikescamell: Yeah this is a good point, does it make sense to make this slottable and do something similar to DaxListItemTrailingScope to enforce that for now we can only add a DaxPill?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+            .then(interaction)
+            .heightIn(min = minHeight)
+            .padding(horizontal = DaxListItemDefaults.HorizontalPadding),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        if (leadingContent != null) {
+            leadingScope.leadingContent()
+            Spacer(Modifier.width(DaxListItemDefaults.LeadingGap))
+        }
+        Column(Modifier.weight(1f).alpha(if (enabled) 1f else DaxListItemDefaults.DisabledAlpha)) {
```
mikescamell: No space between Column and declared properties. It would be easier to read if there were. Is there anything we could apply to Spotless to enforce this?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+                    color = primaryTextColor,
+                    maxLines = primaryMaxLines,
+                    modifier = Modifier.weight(1f, fill = false),
+                )
+                if (pillText != null) {
+                    Spacer(Modifier.width(DaxListItemDefaults.PillGap))
+                    DaxPill(text = pillText)
+                }
+            }
+            if (secondaryText != null) {
```
mikescamell, line 144: No space between the if statement and previous Row code.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+                DaxText(
+                    text = secondaryText,
+                    style = DuckDuckGoTheme.typography.body2,
+                    color = secondaryTextColor,
+                    maxLines = secondaryMaxLines,
+                )
+            }
+        }
+        if (trailingContent != null) {
```
mikescamell, line 154: No space between the if statement and previous code.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,232 @@
...
+        size: DaxListItemTrailingIconSize = DaxListItemTrailingIconSize.Medium,
+        tint: Color? = null,
+        enabled: Boolean = true,
+    ) {
+        val effectiveEnabled = enabled && parentEnabled
+        val iconDp = when (size) {
+            DaxListItemTrailingIconSize.Small -> DaxListItemDefaults.TrailingIconSmall
+            DaxListItemTrailingIconSize.Medium -> DaxListItemDefaults.TrailingIconMedium
+        }
+        if (onClick != null) {
```
mikescamell: No space between the if statement and previous code.

0nko (commented):
I went over the code and aside from the existing comments I didn't find any issues. Nice work! 🏅 

When I was looking at the View vs. Compose differences, I noticed some UI and behavior differences, like: 
- Different leading icon sizes and color (globe) (smaller in Compose)
- Different ripple size when tapping the menu icon (smaller in Compose)
- No leading icon click handler in Compose
- Smaller padding between primary and secondary text in Compose

I was looking for the documentation for the differences but I couldn't find any README and [this](https://app.asana.com/1/137249556945/project/1215496415658080/task/1214523588264771?focus=true) is empty. Is it somewhere else?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+     * @param status Status to display.
+     * @param modifier Modifier applied to the indicator.
+     */
+    @Composable
+    fun StatusIndicator(
+        status: Status,
+        modifier: Modifier = Modifier
+    ) {
+        DaxStatusIndicator(status = status, modifier = modifier.alpha(if (parentEnabled) 1f else DaxListItemDefaults.DisabledAlpha))
+    }
```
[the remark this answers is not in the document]
mikescamell: Fixed in 6bd0075b. Every trailing element now ends 16dp from the row edge, matching Figma. The status indicator had no end inset at all, and the trailing icon sat at 14dp because it inherited the View component's geometry of a 24dp icon centred in a 52dp container.

One correction on the report: the ghost button case was already right. It carries 16dp of its own content padding, so a 0dp spacer is correct there and its label lands on the same 16dp grid.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,342 @@
...
+
+            if (onClick != null) {
+                DaxIconButton(
+                    onClick = onClick,
+                    iconPainter = painter,
+                    contentDescription = contentDescription,
+                    enabled = effectiveEnabled,
+                    modifier = modifier
+                        .alpha(if (effectiveEnabled) 1f else DaxListItemDefaults.DisabledAlpha),
+                )
```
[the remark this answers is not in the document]
mikescamell: Fixed. DaxIconButton now takes an iconSize param, defaulting to Dp.Unspecified so all existing callers render exactly as before, and the list item passes the resolved size through. The end padding is derived from the icon size rather than hardcoded, so Small (16dp) and Medium (24dp) both land 16dp from the row edge.

This also fixes the two showcase rows without touching them. Both pass a 16dp drawable, so Small renders at 16dp and Medium scales up to 24dp, which is what the paired View rows already do.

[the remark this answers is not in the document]
mikescamell: Fixed in 6bd0075b. Dropped the extra alpha from the clickable branch, since DaxIconButton already swaps in the disabled icon colour when enabled is false. The decorative branch keeps its alpha because a plain Icon has no disabled state of its own.

mikescamell:
Every row on the List Items tab has a Compose pair apart from the ones below. All 7 Settings rows are paired. Figma is the reason for most of this.

**Leading icon at 16dp (View `IconSize.Small`)**
- Not in Figma. The two sizes it has are 24dp and 32dp, both built in Compose.
- No layout in the app sets `leadingIconSize="small"`.
- Worth knowing: Compose `DaxListItemIconSize.Small` is 24dp, which is the View's Medium. The names don't line up between the two.

**Leading icon at 40dp (View `IconSize.ExtraLarge`)**
- Not in Figma, and unused outside the showcase.
- The background variant can't render anyway. Icon and circular container are both 40dp, so the background is never visible.
- Figma's closest thing is the large full bleed treatment, which is artwork filling the container. That's the `Image` member of the Compose leading scope, so the intent is covered.

**Leading icon click**
- Not ported. `setLeadingIconClickListener` has no callers in the app, only the showcase.
- It also binds to the icon rather than its container, so with a circular background only the inner 24dp is tappable, under the 48dp minimum.
- If a screen needs it later I'd rather add it then, as a parameter with a proper touch target.

**Things this PR changed rather than skipped**
- Rounded leading icon background: this was a gap, not a decision. Figma has a rectangle background alongside the circle, so `DaxListItemIconBackground.Rounded` is in now and both rows are paired.
- `DaxSettingsListItem` was one line only with the trailing element hardcoded to a status indicator. Figma treats every settings row as a settings list item whatever its trailing element, so the required `status` param is gone, the trailing slot is open, and it takes a `secondaryText`.
- I removed the trailing button from the trailing scope. Neither Figma nor the View has one and nothing called it.

One deliberate difference to flag: the circular and rounded backgrounds use 6% black on both platforms, Figma has them at 3%. That's a theme level thing being picked up separately, so Compose stays in step with the View for now.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,405 @@
...
+            .padding(start = HorizontalPadding),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        if (leadingContent != null) {
+            leadingScope.leadingContent()
+            Spacer(Modifier.width(DaxListItemDefaults.LeadingGap))
+        }
+
+        Column(
+            verticalArrangement = Arrangement.spacedBy(2.dp),
```
GerardPaligot: Suggestion: Put `2.dp` in `DaxListItemDefaults`

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
```diff
@@ -0,0 +1,405 @@
...
+     */
+    @Composable
+    fun Switch(
+        checked: Boolean,
+        onCheckedChange: ((Boolean) -> Unit)?,
+        modifier: Modifier = Modifier,
+        enabled: Boolean = true,
+    ) {
+        Row {
+            DaxSwitch(checked = checked, onCheckedChange = onCheckedChange, modifier = modifier, enabled = enabled && parentEnabled)
```
GerardPaligot, line 316: Question: It could be strange to specify explicitly the switch with true for `enabled` parameter but the switch is disabled because the parent is disabled. What do you think to use an 'OR' instead an 'AND'? (same question for all enabled parameter in this scope)

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
GerardPaligot: My 2 cents: If we have a variant which is using `AnnotatedString` only, we can do everything with a very low extra cost for these edge cases.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
```diff
@@ -0,0 +1,114 @@
...
+ * A reusable pill for short status labels (e.g. "Beta", "New").
+ *
+ * The palette is intentionally identical in light and dark.
+ * Callers supply their own (localised) label.
+ *
+ * @param text Pill label; rendered upper-cased.
+ * @param modifier Modifier applied to the pill container.
+ *
+ * Asana task: https://app.asana.com/1/137249556945/project/1202857801505092/task/1217020584206032
+ * Figma reference: https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=6032-13783
```
GerardPaligot, line 53: Suggestion: In DaxListItem components, these links are at the top but for all others, there are at the bottom of kdoc comments. We should be consistant here.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
```diff
@@ -0,0 +1,114 @@
...
+        modifier = modifier
+            .height(DaxPillDefaults.Height)
+            .clip(RoundedCornerShape(DaxPillDefaults.CornerRadius))
+            .background(DaxPillDefaults.colors.background)
+            .padding(horizontal = DaxPillDefaults.HorizontalPadding),
+        contentAlignment = Alignment.Center,
+    ) {
+        Text(
+            text = text.uppercase(),
+            color = DaxPillDefaults.colors.text,
```
GerardPaligot, line 70: Question: For DaxListItem components, you don't use the default class to declare colors. Is there any reason?

GerardPaligot (approved):
Congrats for this component, the contract is great! Can't wait to use it!
I just posted few questions

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxListItem.kt
mikescamell: The View samples do the same: disabling the item also disables its switch, while `switchEnabled=false` disables only the switch. I’ll keep the AND for parity. We can document this and revisit it later if it doesn’t make sense any more.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/listitem/DaxTwoLineListItem.kt
mikescamell: I don’t think we need another overload until there’s a caller for it. The all-`AnnotatedString` overload covers the edge case if it comes up.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/pill/DaxPill.kt
mikescamell: Nice catch! Moving to it's own color class in the Defaults object

PR #9253, opened 2026-07-21
Add comments guidance into LLM rules

cmonfortep:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214901934989258/task/1216760582882444?focus=true 

### Description
Adds guidance to agents to instruct our comments guidance.
Tries to address LLMs creating verbose and non useful comments.

### Steps to test this PR

_Feature 1_
- [ ]
- [ ]

### UI changes
| Before  | After |
| ------ | ----- |

---

> [!NOTE]
> **Low Risk**
> Documentation-only change to agent guidelines; no runtime or build behavior affected.
> 
> **Overview**
> Adds a **Code Comments** section to `AGENTS.md` so agent and reviewer instructions align with how the team wants comments written.
> 
> The guidance favors **why** over **what**, defaults to no comment when code is clear, and allows KDoc on `-api` contracts as consumer documentation. It lists disallowed patterns (trivial narration, temporal residue, process/plan references) with Kotlin examples, and tells reviewers to flag new comments that break these rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8160738289f5d49d9f0b7567583dc87a2706a7f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmonfortep:
* **#9253** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9253?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9253?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

catalinradoiu:
The `aiConfigCheck` Gradle task scans backtick code spans in `AGENTS.md` and treats any token that looks like a path as a file reference it needs to resolve. The problem is that a span starting with `//` (for example `// fixes the bug from the previous task`) gets read as a repo root path, so the checker goes looking for that file, doesn't find it, and reports a dangling reference.

Worth noting: `count++ // increment the counter` passes fine, because that span doesn't start with `//`, so the checker treats it as illustrative and skips it.

The cleanest fix is to move the example comments into a fenced code block. The checker ignores everything inside fences, so the `//` strings stop being parsed as paths:

````markdown
## Code Comments

Comments explain the **why** (intent, assumptions, non-obvious decisions), not the **what** the code already shows.
Default to no comment: prefer self-documenting code (clear names, small functions), and add one only where code alone can't carry the reasoning, and only if it still answers *"why was this done this way?"* for someone reading it two years from now.
`-api` module declarations may carry KDoc describing the contract, that's documentation for consumers, not narration.

Never add:
- **narration** of trivial code
- **conversational / temporal residue**
- **process / plan references**

Examples of comments to avoid:

```kotlin
count++ // increment the counter          // narration of trivial code
val user = repo.load() // fixes the bug from the previous task   // temporal residue
val state = flow // changed from LiveData                        // temporal residue
cache.clear() // step 3 of the plan       // process / plan reference
retryCount = 3 // per the Asana task      // process / plan reference
```

When reviewing a diff, flag any added comment that doesn't follow these rules.
````

If you'd rather keep the inline bullet style, the other option is to put a bit of real code in front of each example so the span no longer starts with `//` (for example `val user = repo.load() // fixes the bug from the previous task`).

catalinradoiu (approved):
LGTM 👍

PR #9270, opened 2026-07-22
Progress Bar: Indeterminate fallback state

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207418217763355/task/1216760636921126?focus=true

### Description

Implements the indeterminate progress bar fallback state, which kicks in after a 5s stall is detected. Once the loading resumes, the normal progress bar tracking is resumed, as well.

There is a kill-switch (`progressBarUpgradeFeature.indeterminateFallback`) that can be used to disable the feature.

### Steps to test this PR

- [x] Load a website that takes >5s to load
- [x] Notice how the indeterminate state kicks in
- [x] After loading resumes, verify that the regular progress tacking resumes, as well

### UI changes

https://github.com/user-attachments/assets/5c3b49f2-47a1-4a2f-8382-ae97604eb3e4

---

> [!NOTE]
> **Low Risk**
> UI-only omnibar progress behavior behind feature toggles, with broad unit test coverage and no auth or data-path changes.
> 
> **Overview**
> Adds an **indeterminate fallback** for the upgraded omnibar page-load bar when WebView progress stalls for **5s** without forward updates. Gated by **`progressBarUpgrade.behaviourUpdate`** and a new **`indeterminateFallback`** kill-switch; `OmnibarLayoutViewModel` exposes `isProgressBarIndeterminateEnabled` and the layout calls `setStallDetectionEnabled` on `PageLoadProgressBar`.
> 
> `ProgressPhaseEngine` gains **`INDETERMINATE`** and **`RESUMING`**: stall detection freezes the determinate fill, then either animates catch-up back to tracking or defers completion until the sweep exits. New **`IndeterminateSweepGeometry`** / **`IndeterminateSweepRenderer`** draw fast-then-slow accent sweeps seeded from the current fill. **`PageLoadProgressBar`** wires the sweep into the frame loop, requests finish on resume or page complete, and pauses Choreographer callbacks when the window is not visible.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b31df4b44663bb71ac9d1eba321fdcb525357eb0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/src/main/java/com/duckduckgo/app/browser/progressbar/IndeterminateSweepGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+     * exited. Used to finish the current sequence gracefully when leaving the indeterminate state.
+     */
+    fun calculateSequenceEnd(startTime: Long, now: Long, cycleMs: Long): Long {
+        if (cycleMs <= 0L) return now
+        val cycleIndex = (now - startTime).coerceAtLeast(0L) / cycleMs
+        return startTime + cycleIndex * cycleMs + getSlowSweepStart(cycleMs) + getSlowSweepDuration(cycleMs)
+    }
+
+    fun getCycleIndex(startTime: Long, now: Long, cycleMs: Long): Long =
+        if (cycleMs <= 0L) 0L else (now - startTime).coerceAtLeast(0L) / cycleMs
```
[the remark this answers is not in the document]
0nko: Fixed.

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -127,6 +127,8 @@ class OmnibarLayoutViewModel @Inject constructor(
 
     private val isSplitOmnibarEnabled = settingsDataStore.omnibarType == OmnibarType.SPLIT
     private val isProgressBarUpgradeEnabled = progressBarUpgradeFeature.behaviourUpdate().isEnabled()
+    private val isProgressBarIndeterminateEnabled =
```
mikescamell, line 130: ❓Is there any reason we can't check this on a background dispatcher rather than on ViewModel init? Similar to isSetFavouriteEasterEggLogoFeatureEnabled

0nko: Yeah, good point. Also the flags are actually redundant. Fixed.

app/src/main/java/com/duckduckgo/app/browser/progressbar/ProgressPhaseEngine.kt
```diff
@@ -40,12 +40,19 @@ class ProgressPhaseEngine(
     var displayProgress: Float = 0f
         private set
 
-    private var realProgress: Float = 0f
+    var stallDetectionEnabled: Boolean = false
+
+    var realProgress: Float = 0f
```
mikescamell: 🔍I think we can make this private, right? I don't think any external class is using it

app/src/main/java/com/duckduckgo/app/browser/progressbar/IndeterminateSweepRenderer.kt
```diff
@@ -0,0 +1,88 @@
...
+        startTime = 0L
+        endTime = 0L
+        finalCycleIndex = null
+    }
+
+    /**
+     * Stop looping: finish the current fast-slow sequence, then report done via [hasFinished].
+     * Sweeps from the next sequence are suppressed while the final slow sweep exits.
+     */
+    fun requestFinish(now: Long) {
```
mikescamell, line 63: 🔍 The engine and geometry are well covered, but this class and the hand-off in `PageLoadProgressBar` aren't tested directly, and this is exactly where the wrong finish cycle bug lived. Everything here except `draw` is Canvas-free and takes `now` as a parameter, so a small JVM test covering the `start` / `requestFinish` / `hasFinished` sequencing (plus double `requestFinish` and the final cycle suppression) should be cheap to add.

app/src/main/java/com/duckduckgo/app/browser/progressbar/ProgressPhaseEngine.kt
```diff
@@ -93,6 +131,8 @@ class ProgressPhaseEngine(
             Phase.IDLE -> {} // no-op
             Phase.FAST_START -> tickFastStart()
             Phase.TRACKING -> tickTracking(dt)
+            Phase.INDETERMINATE -> {} // hold displayProgress frozen; the view renders the sweep
```
mikescamell: 🔍 Nice to have: flipping the kill-switch off only stops us entering INDETERMINATE. A page that's already mid-sweep and never progresses again keeps sweeping until completion. Probably rare, but the kill-switch exists for when the sweep itself misbehaves, so bailing out here would cover it:
```suggestion
            Phase.INDETERMINATE -> if (!stallDetectionEnabled) onSweepFinished() // otherwise hold frozen; the view renders the sweep
```

app/src/main/java/com/duckduckgo/app/browser/progressbar/ProgressBarConfig.kt
```diff
@@ -22,6 +22,9 @@ data class ProgressBarConfig(
     val springStiffness: Float = 2.0f,
     val dampingRatio: Float = 8.5f,
     val creepVelocity: Float = 0.04f,
+    val stallTimeoutMs: Long = 5000L,
+    val indeterminateCycleMs: Long = 2000L,
+    val resumeDuration: Long = 2000L,
```
mikescamell: 🔍 The other new fields use the `Ms` suffix (`stallTimeoutMs`, `indeterminateCycleMs`) but this one doesn't:
```suggestion
    val resumeDurationMs: Long = 2000L,
```
(`tickResuming` in `ProgressPhaseEngine` would need the same rename)

app/src/main/java/com/duckduckgo/app/browser/progressbar/IndeterminateSweepRenderer.kt
```diff
@@ -0,0 +1,88 @@
...
+        IndeterminateSweepGeometry.calculateSegments(elapsed, config.indeterminateCycleMs, seedLeading)
+            .asSequence()
+            .filter { segment -> lastCycle == null || segment.cycleIndex <= lastCycle }
+            .forEach { segment ->
+                val left = segment.edges.trailing * trackWidth
+                val right = segment.edges.leading * trackWidth
+                if (right - left > 0f) {
+                    canvas.drawRoundRect(left, top, right, bottom, radius, radius, paint)
+                }
+            }
```
mikescamell: 🔍 This runs every frame while the sweep is showing. The list is tiny so it's not a real problem, but the `asSequence` / `filter` / `forEach` chain allocates on every draw and a plain loop doesn't:
```suggestion
        for (segment in IndeterminateSweepGeometry.calculateSegments(elapsed, config.indeterminateCycleMs, seedLeading)) {
            if (lastCycle != null && segment.cycleIndex > lastCycle) continue
            val left = segment.edges.trailing * trackWidth
            val right = segment.edges.leading * trackWidth
            if (right - left > 0f) {
                canvas.drawRoundRect(left, top, right, bottom, radius, radius, paint)
            }
        }
```

app/src/main/java/com/duckduckgo/app/browser/progressbar/ProgressPhaseEngine.kt
```diff
@@ -115,6 +155,14 @@ class ProgressPhaseEngine(
     }
 
     private fun tickTracking(dt: Float) {
+        if (stallDetectionEnabled &&
+            timeProvider.elapsedRealtime() - lastForwardProgressTime >= config.stallTimeoutMs
```
mikescamell, line 161: 🔍 `elapsedRealtime()` keeps counting while the app is backgrounded, so backgrounding mid-load for more than 5s will kick off the sweep on the first frame back even though the page wasn't actually stalled from the user's point of view. `tick` caps `dt` for the spring but the stall timer has no equivalent. Refreshing `lastForwardProgressTime` when the view becomes visible again would cover it. Nice to have, happy to leave it if we think it's rare enough.

app/src/main/java/com/duckduckgo/app/browser/progressbar/PageLoadProgressBar.kt
```diff
@@ -85,13 +94,25 @@ class PageLoadProgressBar @JvmOverloads constructor(
...
+                indeterminateSweepRenderer.stop()
+            }
+
             invalidate()
 
-            if (state.phase == Phase.DONE) {
+            if (engine.phase == Phase.DONE) {
                 onDone()
-            } else if (state.shouldInvalidate) {
+            } else if (engine.phase != Phase.IDLE) {
```
mikescamell: 🔍 `FrameState.shouldInvalidate` is computed on every frame but nothing reads it any more, this line was its only consumer. Having the same "keep animating" condition expressed in two places invites them drifting apart later. I would either keep using the snapshot here or delete the field:
```suggestion
            } else if (state.shouldInvalidate) {
```

app/src/main/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModel.kt
```diff
@@ -140,6 +142,7 @@ class OmnibarLayoutViewModel @Inject constructor(
             showTabsMenu = !isSplitOmnibarEnabled,
             showBrowserMenu = !isSplitOmnibarEnabled,
             isProgressBarUpgradeEnabled = isProgressBarUpgradeEnabled,
+            isProgressBarIndeterminateEnabled = isProgressBarIndeterminateEnabled,
```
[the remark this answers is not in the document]
0nko: That's fine.

0nko:
Thanks @mikescamell! All comments addressed. Ready for another round.

app/src/test/java/com/duckduckgo/app/browser/omnibar/OmnibarLayoutViewModelTest.kt
```diff
@@ -3115,4 +3116,28 @@ class OmnibarLayoutViewModelTest {
...
+    fun whenUpgradeEnabledButIndeterminateFallbackDisabledThenIndeterminateFlagIsFalse() = runTest {
+        fakeProgressBarUpgradeFeature.behaviourUpdate().setRawStoredState(Toggle.State(enable = true))
+        fakeProgressBarUpgradeFeature.indeterminateFallback().setRawStoredState(Toggle.State(enable = false))
+        initializeViewModel()
+
+        testee.viewState.test {
+            val viewState = awaitItem()
+            assertFalse(viewState.isProgressBarIndeterminateEnabled)
+        }
+    }
```
mikescamell, line 3142: 🔍 The revert dropped the upgrade-off + fallback-on case, so the AND gating isn't covered any more. It works fine with the synchronous reads:
```suggestion
    }

    @Test
    fun whenUpgradeDisabledButIndeterminateFallbackEnabledThenIndeterminateFlagIsFalse() = runTest {
        fakeProgressBarUpgradeFeature.behaviourUpdate().setRawStoredState(Toggle.State(enable = false))
        fakeProgressBarUpgradeFeature.indeterminateFallback().setRawStoredState(Toggle.State(enable = true))
        initializeViewModel()

        testee.viewState.test {
            val viewState = awaitItem()
            assertFalse(viewState.isProgressBarIndeterminateEnabled)
        }
    }
```

app/src/test/java/com/duckduckgo/app/browser/progressbar/ProgressPhaseEngineTest.kt
```diff
@@ -268,4 +268,237 @@ class ProgressPhaseEngineTest {
...
+        engine.tick(0.016f)
+
+        assertEquals(Phase.TRACKING, engine.phase)
+        time.advance(config.stallTimeoutMs - 1)
+        engine.tick(0.016f)
+        assertEquals(Phase.TRACKING, engine.phase)
+        time.advance(1)
+        engine.tick(0.016f)
+        assertEquals(Phase.INDETERMINATE, engine.phase)
+    }
```
mikescamell, line 332: 🔍 The kill-switch mid-sweep bail-out doesn't have a test yet. Something like:
```suggestion
    }

    @Test
    fun `disabling stall detection mid-sweep exits the indeterminate phase`() {
        enterIndeterminate(real = 50f)

        engine.stallDetectionEnabled = false
        engine.tick(0.016f)

        assertEquals(Phase.RESUMING, engine.phase)
    }
```

mikescamell (approved):
Nice work! LGTM :shipit: 

Tested on a Pixel 4a and Tablet emulator.

PR #9284, opened 2026-07-23
[Brand design updates - Core App Components] Update Button styles & Add New Brand Button

mikescamell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1207908166761516/task/1216143677003477?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/1207908166761516/task/1216195196427041

### Description

Adds the rebrand button system behind the `appBrandDesignUpdate` flag. `DuckDuckGoActivity` applies a `ThemeOverlay.Rebrand` that re-points every `daxButton*` theme attribute at new rebrand styles, so existing `DaxButton*` widgets pick up the new metrics, typography and colour selectors without per-screen changes.

Also adds `DaxButtonBrand` for promotional surfaces. It is un-gated and always uses the rebrand look, with `daxButtonBrandTypography` as the only sanctioned route to DuckSans, enforced by a new lint issue. Onboarding CTAs and the Fire Tabs empty state CTA move onto the shared components so they no longer carry local button overrides.

One behaviour change worth calling out: under the overlay, `daxButtonGhost` and `daxButtonGhostAlt` both resolve to the same rebrand Ghost style, so Ghost Alt is no longer a distinct look when the flag is on.

### Designs

[Project spec
](https://www.figma.com/design/6tcKR2xDTfq80O2FqU6Qsi/%F0%9F%93%90-Mobile--New-Design-Language?node-id=1448-22564&t=Dfe42CcBvWBWhB5s-4)

[Update Design System Components page with all new buttons 
](https://www.figma.com/design/BOHDESHODUXK7wSRNBOHdu/%F0%9F%A4%96-Android-Components?node-id=6550-28349&t=Q0Wtwygug0VkY5Rk-4)

### Steps to test this PR

_Design system preview_
- [x] Internal settings > Android Design System Preview > toggle `Brand design update` on if it is not already
- [x] Open the Buttons component and check every variant against the designs: Primary, Secondary, Ghost, Ghost Alt, Destructive, Destructive Secondary, Ghost Destructive, and the Brand small/large/disabled/DuckSans examples
- [x] Ghost and Ghost Alt should now render identically, since both attributes point at the single rebrand Ghost style
- [x] Toggle `Brand design update` off and on and confirm the buttons switch between the legacy and rebrand looks

_Onboarding (Primary and Secondary)_
- [x] Fresh install and walk through onboarding
- [x] The CTAs should be visually unchanged from the current release apart from now using the brand button, so compare side by side with a build from `develop`
- [x] The secondary CTAs are now a filled secondary. At rest they should match the previous version; when held down the pressed highlight should differ

_Fire mode new tab (Secondary)_
- [x] Activate fire mode and open the Fire Tabs new tab screen
- [x] The empty state CTA should be the shared filled secondary rather than the previous hand restyled button, and should look the same with the flag on or off

_Fire dialog (Destructive and Destructive Secondary)_
- [x] Open a tab, tap the fire button and check the single tab fire dialog
- [ ] `Delete this tab` is new Destructive Primary, `Delete all` is Destructive Secondary

_Ghost in an alert dialog_
- [x] Settings > Passwords, open a saved password and delete it
- [x] The confirmation dialog buttons are Ghost, so they should pick up the rebrand Ghost style

_Exploratory pass_
- [x] Move through the rest of the app with the flag on. The overlay is applied at activity level, so any screen using `DaxButton*` is affected. Look for clipped text, wrong heights or buttons that no longer match the designs

### UI changes

See Asana [task](https://app.asana.com/1/137249556945/project/1207908166761516/task/1217194782052987?focus=true) for updated button screenshots

---

> [!NOTE]
> **Medium Risk**
> App-wide activity theming changes every `DaxButton*` screen when the flag is on, though rollout is gated; onboarding and promo CTAs change unconditionally via `DaxButtonBrand`.
> 
> **Overview**
> Introduces the rebrand button stack behind the **`appBrandDesignUpdate`** remote flag. **`DuckDuckGoActivity`** applies **`ThemeOverlay.Rebrand`** when the flag is on (only on themes that expose rebrand attrs), so existing **`DaxButton*`** widgets pick up new styles, metrics, and color selectors without per-screen edits. **`DaxButton`** now resolves padding, height, vertical inset, and large-button typography from theme attributes so the overlay can override legacy dimens.
> 
> Adds **`DaxButtonBrand`** for promotional/onboarding CTAs (always rebrand styling, not flag-gated), with **`daxButtonBrandTypography="duckSans"`** as the only supported DuckSans path. Onboarding, quick-setup sheets, contextual brand CTAs, and several layouts swap **`DaxButtonPrimary`** for **`DaxButtonBrand`** or use **`ThemeOverlay.Rebrand`** on secondary buttons; local onboarding button styles and hand-tuned Fire Tabs secondary overrides are removed. With the overlay on, **`daxButtonGhost`** and **`daxButtonGhostAlt`** both map to the same rebrand Ghost style.
> 
> **`AppBrandDesignUpdateToggles`** moves into **`common.ui.store`**, with a pre-warm lifecycle observer for the synchronous toggle read in **`onCreate`**. Rebrand palette tokens, selectors, and unit/Robolectric tests are added; lint adds **`InvalidDaxButtonDuckSans`**. Internal design-system preview gains a brand-design toggle and brand button samples.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce4a99da81ed9c33fd4ac7395f793fb2dabbc339. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/src/main/res/layout/view_fire_tabs_empty_state.xml
```diff
@@ -109,26 +108,16 @@
...
-                android:textColor="?attr/daxColorPrimaryText"
-                app:backgroundTint="?attr/daxColorContainer"
+                android:theme="@style/ThemeOverlay.Rebrand"
                 app:daxButtonSize="small"
-                app:icon="@drawable/ic_add_16"
-                app:iconTint="?attr/daxColorPrimaryText"
-                app:rippleColor="?attr/daxColorRipple"
-                app:strokeWidth="0dp"
-                tools:ignore="InvalidDaxButtonProperty" />
+                app:icon="@drawable/ic_add_16" />
```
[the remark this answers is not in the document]
mikescamell: Intentional, this was a custom button previously and now users the new rebrand secondary

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/button/DaxButton.kt
```diff
@@ -91,6 +134,11 @@ enum class Size {
     }
 }
 
+internal fun Size.resolveMinHeight(
+    tokenHeight: Int,
+    verticalInset: Int?,
+): Int = if (this == Size.Large && verticalInset != null) tokenHeight + verticalInset * 2 else tokenHeight
```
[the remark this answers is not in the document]
mikescamell: Intentional, the small button is 40dp and we add 4dp insets

mikescamell:
@LukasPaczos This is ready for review 👍

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/store/AppBrandDesignThemeProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Exposes the appBrandDesignUpdate feature flag to the theming layer.
+ * :design-system cannot read feature flags directly; :app binds the implementation.
+ * Interim scaffolding for the brand design update, deleted with the flag by the
+ * follow-up theme project.
+ */
+interface AppBrandDesignThemeProvider {
+    fun isAppBrandDesignUpdateEnabled(): Boolean
+}
```
LukasPaczos: Why don't we move `AppBrandDesignUpdateToggles` to `/android-design-system` instead?

android-design-system/design-system/src/test/java/com/duckduckgo/common/ui/ThemingRebrandOverlayTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    ): Boolean {
+        val value = TypedValue()
+        activity.theme.resolveAttribute(attr, value, true)
+        return value.data != 0
+    }
+
+    private fun fixedThemeActivity(themeResId: Int = R.style.Theme_DuckDuckGo_Dark): AppCompatActivity =
+        Robolectric.buildActivity(SyncConnectActivity::class.java).get().apply {
+            setTheme(themeResId)
+        }
```
LukasPaczos, line 63:
```suggestion
    class FixedThemeActivity : AppCompatActivity() {
        override fun getLocalClassName(): String = Theming.Constants.FIXED_THEME_ACTIVITIES.first()
    }

    private fun fixedThemeActivity(themeResId: Int = R.style.Theme_DuckDuckGo_Dark): AppCompatActivity {
        return Robolectric.buildActivity(FixedThemeActivity::class.java).get().apply {
            setTheme(themeResId)
        }
    }
```

android-design-system/design-system/src/test/java/com/duckduckgo/sync/impl/ui/SyncConnectActivity.kt
LukasPaczos, line 1: This can be deleted with the suggestion above.

app/src/main/res/layout/activity_onboarding.xml
```diff
@@ -41,17 +41,16 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
-    <com.duckduckgo.common.ui.view.button.DaxButtonPrimary
+    <com.duckduckgo.common.ui.view.button.DaxButtonSecondary
```
LukasPaczos, line 44: Nice improvement 👍

LukasPaczos (commented):
1.
There's no ripple effect on the colored buttons (brand/primary/destructive), the pressed state color snaps in on touch and fades out on release. The default, grey-based buttons like 'secondary' do have the ripple, same as the colored non-brand-design buttons.

https://github.com/user-attachments/assets/202d795c-b71d-488b-b796-491ba52f34cf

A related issue seems to be that ghost button has the ripple, but keeps the pressed state background color, while the old one didn't.

https://github.com/user-attachments/assets/d2c4ab6d-2a77-4f73-88a9-04d5887f3337

2.
> The secondary CTAs are now a filled secondary. At rest they should match the previous version

I see both the new primary and secondary have a slightly darker tint in dark mode:
_left old, right new_

Image

Is this expected?

3.
> Delete this tab and Delete all are destructive buttons and the cancel style action is a destructive secondary, all against the designs

Could you clarify this step? Should the buttons in the dialog go against the design and use legacy? Right now I see them updated to brand design buttons (more corner radius etc).

android-design-system/design-system/src/main/res/values/design-system-theming.xml
```diff
@@ -259,6 +266,26 @@
         <item name="daxColorButtonGhostAltTextPressed">@color/white84</item>
         <item name="daxColorButtonDestructiveGhostText">@color/alertRedOnDarkDefault</item>
         <item name="daxColorButtonDestructiveGhostTextPressed">@color/alertRedOnDarkDefault</item>
+        <item name="daxColorButtonBrandContainer">@color/pollen40</item>
+        <item name="daxColorButtonBrandContainerPressed">@color/pollen60</item>
+        <item name="daxColorButtonBrandText">@color/pollen100</item>
+        <item name="daxColorRebrandButtonPrimaryContainer">@color/pondwater40</item>
+        <item name="daxColorRebrandButtonPrimaryContainerPressed">@color/pondwater60</item>
+        <item name="daxColorRebrandButtonPrimaryText">#242323</item>
```
malmstein: 21 of these attrs are inline hex, and three duplicate colors that already have names: `#242323` is `@color/eggshell90`, `#1F75B6EB` and `#1F1074CC` are `pondwater40` and `pondwater60` at 12% alpha. the rest are new reds sitting next to `@color/rb_red40` and `rb_red60`, so the ramp is half tokenised. the alpha ones can't reference a `@color` directly, but we already handle that with named entries like `mandarin40_translucent`. can we name these in the rebrand colors file and reference them here?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/button/DaxButtonBrand.kt
```diff
@@ -0,0 +1,76 @@
...
+        } else {
+            Small
+        }
+        daxButtonAttrs.recycle()
+
+        val brandAttrs = context.obtainStyledAttributes(attrs, R.styleable.DaxButtonBrand, 0, 0)
+        val duckSans = brandAttrs.getInt(R.styleable.DaxButtonBrand_daxButtonBrandTypography, 0) == 1
+        brandAttrs.recycle()
+
+        val sidePadding = resources.getDimensionPixelSize(
```
malmstein, line 48: `DaxButton` was reworked in this PR to resolve side padding, top padding, height and inset from theme attrs with dimen fallbacks, which is what lets the rebrand overlay retarget them. this reimplements the same four with hardcoded `R.dimen.rebrand*` instead. if the brand button is always rebrand and never needs overlaying that's reasonable, but it duplicates the sizing logic in two places. could it declare the `daxButton*` attrs in its own style and reuse the parent's path?

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/button/DaxButtonBrand.kt
```diff
@@ -0,0 +1,76 @@
...
+        insetBottom = inset
+
+        val tokenHeight = resources.getDimensionPixelSize(
+            if (buttonSize == Small) R.dimen.rebrandButtonSmallHeight else R.dimen.rebrandButtonLargeHeight,
+        )
+        minHeight = buttonSize.resolveMinHeight(tokenHeight, inset)
+
+        setPadding(sidePadding, topPadding, sidePadding, topPadding)
+
+        setTextAppearance(
```
malmstein, line 67: `DaxButton` deliberately skips its Large default when the layout already set `textAppearance`, but this sets it every time, so a `DaxButtonBrand` with `android:textAppearance` in XML silently loses it. worth the same `hasLayoutTextAppearance` guard for consistency. related: that helper only inspects direct attributes, so a `textAppearance` arriving via `style="..."` isn't detected either.

android-design-system/design-system/src/test/java/com/duckduckgo/sync/impl/ui/SyncConnectActivity.kt
```diff
@@ -0,0 +1,26 @@
...
+package com.duckduckgo.sync.impl.ui
+
+import androidx.appcompat.app.AppCompatActivity
+
+/**
+ * Test double, not the production class. Its fully qualified name deliberately matches the entry in
+ * Theming.Constants.FIXED_THEME_ACTIVITIES so that localClassName resolves to a fixed-theme
+ * activity. Safe because design-system does not depend on sync-impl.
+ */
+class SyncConnectActivity : AppCompatActivity()
```
malmstein: the comment explains the trick well and the no-dependency argument holds. the fragile part is that it hardcodes a name that lives elsewhere, so if the production `SyncConnectActivity` is renamed or moved this keeps passing while `FIXED_THEME_ACTIVITIES` no longer matches anything. can we assert against the constant itself, for instance drive the test from an entry read out of `FIXED_THEME_ACTIVITIES` rather than restating one?

malmstein (approved):
Smoke-tested on the attached device and it works as expected.

Solid design-system work: `DaxButton` now resolves padding, height and inset from theme attrs with dimen fallbacks so the rebrand overlay can retarget them without touching the component, the new `DaxButtonBrand` slots into `ButtonType`, and the lint rule gains an `InvalidDaxButtonDuckSans` check that stops DuckSans being applied through a font family, a text appearance or a style. Good test coverage across the overlay, the theme attrs and the brand button, and `applyStatusBarInsetsAsMargin` with its "pill grows with padding" rationale is a nice catch. CI green.

Four non-blocking items inline.

mikescamell:
> There's no ripple effect on the colored buttons (brand/primary/destructive), the pressed state color snaps in on touch and fades out on release. The default, grey-based buttons like 'secondary' do have the ripple, same as the colored non-brand-design buttons.
> 
>  Screen_recording_20260811_145525.mp4 
> A related issue seems to be that ghost button has the ripple, but keeps the pressed state background color, while the old one didn't.
> 
>  Screen_recording_20260811_150216.mp4 
> 
> > The secondary CTAs are now a filled secondary. At rest they should match the previous version
> 
> I see both the new primary and secondary have a slightly darker tint in dark mode: _left old, right new_
> 
> Image
> Is this expected?
> 
> 
> > Delete this tab and Delete all are destructive buttons and the cancel style action is a destructive secondary, all against the designs
> 
> Could you clarify this step? Should the buttons in the dialog go against the design and use legacy? Right now I see them updated to brand design buttons (more corner radius etc).

1. I'll get that fixed, can't believe I missed the lack of a ripple 😅 

2. Yes these buttons now match the brand design changes, with the updated colours.

3. Sorry that wasn't clear, I've updated to "`Delete this tab` is new Destructive Primary, `Delete all` is Destructive Secondary"

app/src/main/res/layout/activity_onboarding.xml
mikescamell: Yes the giant blue button was driving me nuts 😅  I was going to move it to the bottom but I feel like it obscured the animations too much

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/button/DaxButtonBrand.kt
mikescamell: Good spot, and you're right that it's duplicated.

I did build this to see what it actually takes. To get the brand style feeding the parent's sizing, `DaxButton` has to read via `obtainStyledAttributes(attrs, styleable, defStyleAttr, 0)` instead of `theme.resolveAttribute`, because a style is invisible to the latter. That means adding all eight `daxButton*` dimension attrs to the `DaxButton` styleable, and the side effect is that they then become settable per instance from any layout. Today they're theme only and no single layout can override them.

I decided that wasn't a good trade. Adding overridable surface to the shared base button to remove a sizing block from one subclass felt the wrong way round, and you called it yourself in your comment: the brand button is always rebrand and never needs overlaying.

So I've left the duplication in. If we do ever need to overlay it, the change is well understood and we can make it then.

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/view/button/DaxButtonBrand.kt
mikescamell: I looked at this but I think the branch is unreachable.

`DaxButtonStylingDetector` has `ATTR_TEXT_APPEARANCE` in `INVALID_ATTRIBUTES` and `DaxButtonBrand` is in `DAX_BUTTONS`, so `android:textAppearance` on a brand button fails the build. There's a test covering that exact case in `DaxButtonStylingDetectorTest`. Your point about `style="..."` is covered too, since `visitElement` flags any `style=` on any `DaxButton`.

So the guard would be dead code. Happy to add it for symmetry anyway if you'd prefer, it's only three lines. Do you want it?

app/src/main/res/layout/activity_onboarding.xml
```diff
@@ -41,16 +41,16 @@
...
-        app:daxButtonSize="large"
-        app:layout_constraintTop_toTopOf="parent"
         android:text="@string/skipOnboarding"
+        android:visibility="gone"
+        app:daxButtonSize="small"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintEnd_toEndOf="parent" />
+        tools:visibilty="visible" />
```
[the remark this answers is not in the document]
mikescamell: The `tools:visibilty` typo doesn't exist anywhere under `app/src/main/res/` any more, so resolving this one.

mikescamell:
@LukasPaczos Ready for round 2 👍 Thanks for the suggestions!

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/RebrandColor.kt
```diff
@@ -486,5 +486,5 @@ val RbRed100
 val Green40
     @Composable
     @ReadOnlyComposable
-    get() = colorResource(R.color.green40)
+    get() = colorResource(R.color.rb_green40)
```
LukasPaczos, line 489: Is this change intentional? I don't think you planned updating compose components just yet?

android-design-system/design-system/build.gradle
```diff
@@ -57,6 +62,7 @@ dependencies {
     implementation project(path: ':anvil-annotations')
     implementation project(path: ':app-build-config-api')
     implementation project(':internal-features-api')
+    api project(':feature-toggles-api')
```
LukasPaczos:
```suggestion
    implementation project(':feature-toggles-api')
```

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/store/AppBrandDesignUpdatePreWarmObserver.kt
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Pre-warms the `appBrandDesignUpdate` toggle's in-memory cache off the main thread so the
+ * synchronous read in `DuckDuckGoActivity.onCreate` is usually a cache hit, not a first-read disk hit.
+ *
+ * Interim scaffolding for the brand design update. The follow-up theme project deletes this observer
+ * with the flag.
+ */
+@ContributesMultibinding(AppScope::class)
+@SingleInstanceIn(AppScope::class)
+class AppBrandDesignUpdatePreWarmObserver @Inject constructor(
```
LukasPaczos, line 39: This could be moved alongside the flag as well.

LukasPaczos (commented):
In dark mode, the destructive and ghost buttons change the text color on touch, which they didn't do in legacy. Is that expected? Feels a little distracting.

https://github.com/user-attachments/assets/034e4ad4-ebae-4536-9a46-478a21747e81

android-design-system/design-system/src/main/java/com/duckduckgo/common/ui/compose/theme/RebrandColor.kt
mikescamell: Yeah just aligning the palette, this color is not used in Compose yet 👍

mikescamell:
> In dark mode, the destructive and ghost buttons change the text color on touch, which they didn't do in legacy. Is that expected? Feels a little distracting.
> 
>  Screen_recording_20260813_132638.mp4

We decided to remove the selected text colour changes for these buttons so it doesn't look glitchy. Design agreed the ripple is enough and we match our old buttons in this behaviour

LukasPaczos (approved):
LGTM

PR #9286, opened 2026-07-23
Set TRUNCATE journaling mode for the history.db and tab_visited_sites.db

0nko:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1202552961248957/task/1216780670689923?focus=true

### Description

This PR sets the `TRUNCATE` journaling mode for `tab_visited_sites.db` and `history.db`, which removes the data residue after clearing data with the Fire button.

### Steps to test this PR

QA-optional

Smoke test data clearing and make sure everything works as expected.

---

> [!NOTE]
> **Low Risk**
> Small builder-only change aligned with existing TRUNCATE usage on other Fire-related databases; no schema or query logic changes.
> 
> **Overview**
> Configures Room to use **`TRUNCATE`** journaling for **`history.db`** and **`tab_visited_sites.db`**, matching **`app.db`** and **`fire_mode.db`**.
> 
> The change is only in the Dagger `Room.databaseBuilder` setup in `HistoryModule` and `TabVisitedSitesModule` via `.setJournalMode(RoomDatabase.JournalMode.TRUNCATE)`. That reduces leftover SQLite journal/WAL residue after Fire-driven data clearing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1da690711e5b86bbdaa75049744537e8bc2bf0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/fire/store/TabVisitedSitesModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/fire/store/TabVisitedSitesModule.kt
@@ -18,6 +18,7 @@ package com.duckduckgo.app.fire.store
 
 import android.content.Context
 import androidx.room.Room
+import androidx.room.RoomDatabase
 import com.duckduckgo.di.scopes.AppScope
 import com.squareup.anvil.annotations.ContributesTo
 import dagger.Module
@@ -32,6 +33,7 @@ class TabVisitedSitesModule {
     @Provides
     fun provideDatabase(context: Context): TabVisitedSitesDatabase {
         return Room.databaseBuilder(context, TabVisitedSitesDatabase::class.java, "tab_visited_sites.db")
+            .setJournalMode(RoomDatabase.JournalMode.TRUNCATE)
             .fallbackToDestructiveMigration()
             .build()
     }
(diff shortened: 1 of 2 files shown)


MiSikora (commented):
I think there's one thing to consider. `TRUNCATE` only takes care of journal residue. Deleted row content can still linger in the free pages of the main `.db` file until a `VACUUM` runs or something eventually overwrites it. So if the goal is complete residue removal after the Fire button, that's probably something worth thinking about.

From what I see we have `DatabaseCleaner.cleanDatabase()` which runs `VACUUM`, but it's currently not invoked for `history.db` or `tab_visited_sites.db`. Might be worth using it here or in a follow-up task.

But vacuuming also isn't free. I don't know all the details, but if the databases can still have a lot of live rows after deleting, then vacuuming can take some time. In that case it becomes a tradeoff between the cost of vacuuming and how much data is actually purged. From my understanding, with Fire Mode the whole content of the databases is purged (except for typical Android/Room metadata, etc.), so it shouldn't be an issue if we go this route.

MiSikora (approved):

0nko:
Thanks @MiSikora! `DatabaseCLeaner.cleanDatabase()` is currently only run with `http_auth.db`. As you said, this isn't free and for example the `tab_visited_sites.db` is used for single tab burning, where only some rows associated with a specific tab are deleted, leaving the rest untouched. 

I'll let @CDRussell and @karlenDimla weigh in.

karlenDimla:
I honestly think that `TRUNCATE` is good enough for the privacy audit perspective. We don’t do any vacuum for the app.db today. The risks are higher for auth related data hence why we enforce the vacuum there.

PR #9308, opened 2026-07-27
Add PT_GNU_RELRO 16 KB check to check_elf_alignment.sh

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208589375218614/task/1216907917649404?focus=true

### Description

Play Console flags our release bundle with:

```
PT_GNU_RELRO start: 0x001ecc20 end: 0x00203000 align: 0x00000001
RELRO is not a suffix and its end is not 16 KB aligned
```

`scripts/check_elf_alignment.sh` (a vendored copy of the AOSP script) reported the same APK as completely clean, because it only checks two things: zip alignment of  `.so` entries inside the APK, and `PT_LOAD` `p_align`. Every native library we ship passes both — all 40 are `2**14` aligned. `PT_GNU_RELRO` is a third, independent condition that nothing in our tooling looked at, so the only way to see this finding
was to upload a build and read the Play Console report.

This adds that check. Libraries are now additionally reported as `RELRO MISALIGNED`, appended to the existing per-library line so output length is unchanged.  
  
Two further notes:

- **`objdump`** **is used rather than** **`llvm-readelf`**, matching what the script already calls, so it keeps working with no NDK on `PATH`.
- **The exit code now covers both checks.** A library can pass `PT_LOAD` while failing RELRO — exactly our case — so the script previously exited 0 while printing findings. This also fixes a pre-existing bug where the unaligned-libs branch called `cleanup_trap` directly and skipped the closing `=====` separator.

Additions are fenced with `>>> DDG local addition >>>` markers. The file is otherwise a verbatim copy of the upstream AOSP script, so a future refresh from `android.googlesource.com` will need them re-applied — the markers make that conflict obvious rather than silent.

This PR only improves _detection_; it does not fix the warning. 

### Steps to test this PR

- [ ] `./scripts/check_elf_alignment.sh <ADD_HERE_AAB>` reports `ALIGNED (2**14) RELRO MISALIGNED (end=0x203000)` and exits **1**. The `end` value matches the address in the Play Console report exactly.

### NO UI changes

---

> [!NOTE]
> **Low Risk**
> Changes only a host-side validation script used in CI; no app runtime, auth, or data-path impact.
> 
> **Overview**
> Extends the vendored AOSP **`check_elf_alignment.sh`** so local runs can surface the same **Play Console** finding that **`PT_GNU_RELRO`** is not 16 KB–safe when writable LOAD data sits in the padding range after RELRO (“RELRO is not a suffix”).
> 
> A new **`relro_misaligned_end`** helper parses **`objdump -p`** (no NDK) and flags libs as **`RELRO MISALIGNED`** on the existing per-library line, with a separate summary count. **Exit behavior** is unified: RELRO failures set **`exit_code=1`**, the script always prints the closing separator, and **`cleanup_trap`** runs with that code—fixing the case where PT_LOAD looked clean but RELRO failed yet the script exited 0, and the old early exit that skipped the footer on unaligned LOAD libs.
> 
> DDG-local blocks are marked for future AOSP refreshes. This is **detection only**; it does not relink or fix native libs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ef2b254eb6eb73630bfc2b96ad76601bb47d2ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/check_elf_alignment.sh
+++ b/scripts/check_elf_alignment.sh
@@ -12,6 +12,8 @@ usage() {
   echo "Host side script to check the ELF alignment of shared libraries."
   echo "Shared libraries are reported ALIGNED when their ELF regions are"
   echo "16 KB or 64 KB aligned. Otherwise they are reported as UNALIGNED."
+  echo "Libraries whose PT_GNU_RELRO segment cannot be protected at a 16 KB"
+  echo "page size are additionally reported as RELRO MISALIGNED."
   echo
   echo "Usage: ${progname} [input-path|input-APK|input-APEX]"
 }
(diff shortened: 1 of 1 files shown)


anikiki:
* **#9308** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9308?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9308?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

cmonfortep (approved):
LGTM Thanks @anikiki

PR #9335, opened 2026-07-29
Improve RealContentScopeScripts.getScript()

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200581511062568/task/1216821003646786?focus=true

 ### Description

  Reduces the per-navigation cost of `getScript()` without changing its output, behind the `optimizeContentScopeInjection` kill-switch (`DefaultFeatureValue.INTERNAL`). Every input is still read on every call, so nothing can go stale — the savings are all downstream, in skipping work whose inputs didn't change and doing the rest more cheaply.

  Optimized path:

  - Plugin `config()` and `preferences()` accumulate into `StringBuilder`s instead of reassigning a `String` per plugin, which copied the whole accumulator each time — quadratic in the assembled config size, across ~22 plugins, on the main thread.
  - The contentScope JSON is rebuilt only when the assembled config or the exception list actually changed, reusing the already-serialized exceptions instead of re-serializing and re-comparing per call.
  - `runBlocking` is skipped entirely when no contentScope experiment is active — `getActiveExperiments()` returns only enabled toggles, so the result is then a constant.
  - `Toggle.getCohort()` is called once per toggle instead of twice; it can enrol to self-heal a stale cohort, so the second call wasn't a free read.
  - The feature flag is read once per call rather than at each use site, so a flag flip can't mix the two paths mid-call.
  - Messaging parameters and the version/platform constants are assembled once, and the messaging parameters are embedded when building the preferences JSON instead of being substituted from a placeholder during assembly.

  An earlier revision cached the assembled plugin config and invalidated it on privacy-config persistence.
  That was removed: a feature repo that hasn't finished its async load reports its default JSON rather than an
  empty string, so the startup stabilisation latch could settle on defaults and serve them for the rest of the
  process.

  Deliberately unchanged: allow list, unprotected temporary, plugin `config()` and `preferences()`, cohorts and
  desktop mode are evaluated on every call, so config-driven and user-facing settings (GPC, forced zoom,
  allowlisting) apply immediately.

  Rollback: with the flag off, every call runs `optimized = false` end to end through the original helpers —
  throwaway Moshi instances, the original quadratic parameter assembly, the original cohort filter/map, and the
  chained-replace assembly. The legacy path is byte-and-behaviour identical to before and is deleted with the
  flag once rollout completes. `whenOptimizeInjectionEnabledThenOutputIsByteIdenticalToFallbackPath` compares
  the two paths in full, and two further tests pin the freeze via the per-flag `getCohort()` and
  `appBuildConfig` read counts.

### Steps to test this PR
  
  - [x] Tests are passing.

  flag ON (default on internal)

  - [x] Fresh install, browse several pages in a row, confirm protections stay applied.
  - [x] Override the privacy config from dev settings, confirm a config-driven change takes effect.
  - [x] Toggle GPC in Settings, reload, confirm `globalPrivacyControlValue` takes effect (`sec-gpc: 1` is not
        sent when GPC is off).
  - [x] Toggle Accessibility → force zoom, reload, confirm behaviour changes.
  - [x] Tap the shield to allowlist a site, confirm protections drop, then re-protect it.
  - [x] Switch to desktop site and confirm the change is reflected.

  Rollback is clean
  
  - [x] Turn the flag OFF, restart, repeat the above. Behaviour must be identical to develop.
  - [x] Flip it back ON without restarting, load a page, confirm protections are still correct.

### NO UI changes

---

> [!NOTE]
> **Medium Risk**
> Changes how often privacy-config-driven script JSON is rebuilt and when plugin config updates apply (persist vs startup latch), which could affect protection behavior if invalidation or settling logic is wrong; legacy path preserved behind a kill-switch.
> 
> **Overview**
> Adds an **optimized** `getScript()` path behind `optimizeContentScopeInjection`, while keeping the prior logic as `getLegacyScript()` for rollback.
> 
> **Caching & invalidation:** Assembled plugin `config()` is cached and only rebuilt when `onPrivacyConfigPersisted()` sets `pluginConfigNeedsRebuild`, or during startup until the config stabilizes (empty configs never latch). `RealContentScopeScripts` now implements `PrivacyConfigCallbackPlugin`; `onPrivacyConfigDownloaded()` is intentionally a no-op.
> 
> **Per-call work:** Allow list, plugin `preferences()`, desktop mode, and cohorts still update every navigation. Content-scope JSON is rebuilt only when cached plugin config or unprotected-temporary exceptions change. Messaging secrets, version/platform strings, reusable Moshi adapters, and single-pass template assembly are reused on the optimized path; cohort reads go from two `getCohort()` calls to one per toggle.
> 
> **Flag behavior:** Mid-session flag flips route entirely through one path; legacy sets `pluginConfigNeedsRebuild` so switching back to optimized does not resurrect stale plugin config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d54e2870bc6d78529de2443f1bcfb9d681d77c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/RealContentScopeScripts.kt
+++ b/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/RealContentScopeScripts.kt
@@ -71,10 +71,6 @@ class RealContentScopeScripts @Inject constructor(
     private val experimentsAdapter: JsonAdapter<List<Experiment>> =
         reusableMoshi.adapter(Types.newParameterizedType(List::class.java, Experiment::class.java))
 
-    // Compared against a freshly assembled config on every call, not used to skip assembling it: the
-    // feature repos load their persisted config into memory asynchronously and expose no completion
-    // signal, so any attempt to decide up front that the config cannot have changed risks serving
-    // default protections for the rest of the process.
     private var cachedPluginConfig: String = ""
 
     private var cachedContentScopeJson: String = getContentScopeJson("", emptyList(), optimized = true)
@@ -121,7 +117,7 @@ class RealContentScopeScripts @Inject constructor(
     ): String {
         var updateJS = false
 
-        val pluginParameters = getPluginParameters()
+        val pluginParameters = getLegacyPluginParameters()
 
         if (cachedUnprotectTemporaryExceptions != unprotectedTemporary.unprotectedTemporaryExceptions) {
             cacheUserUnprotectedTemporaryExceptions(unprotectedTemporary.unprotectedTemporaryExceptions, optimized = false)
@@ -161,7 +157,7 @@ class RealContentScopeScripts @Inject constructor(
         var updateJS = false
         var contentScopeChanged = false
 
-        val pluginParameters = getPluginParameters()
+        val pluginParameters = getOptimizedPluginParameters()
 
         if (cachedPluginConfig != pluginParameters.config) {
             cachedPluginConfig = pluginParameters.config
@@ -207,7 +203,7 @@ class RealContentScopeScripts @Inject constructor(
 
     private fun getInterfaceKeyValuePair() = "\"javascriptInterface\":\"$javascriptInterface\""
 
-    private fun getPluginParameters(): PluginParameters {
+    private fun getLegacyPluginParameters(): PluginParameters {
         var config = ""
         var preferences = ""
         val plugins = pluginPoint.getPlugins()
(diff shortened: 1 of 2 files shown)


anikiki:
* **#9335** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9335?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9335?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

CrisBarreiro (approved):

PR #9336, opened 2026-07-29
[Config-driven linear onboarding dialogs] Unify title handling

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1216978524863339?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/72649045549333/task/1216854264994244

### Description
The title machinery each step layout copy-pasted — typing `TypeAnimationTextView`, invisible sizing twin, `preventWidows` at the call site — becomes one `OnboardingDialogTitleView`, dropped into all eight step includes plus the address bar picker bottom sheet, which shares the input screen layout.

Three behavior changes:

1. **`preventWidows` on every title**, not the three that happened to call it. Five titles gain a non-breaking space before their last word.
2. **`includeFontPadding="false"` on every title.** Four layouts set it, four left the platform default.
3. **Input screen preview sizing fix.** Its animated path sized the twin from the raw string, so a literal `<br/>` reserved one line instead of two and the card grew while the title typed. Now decoded uniformly, matching what the snap path already did.

@mikescamell, all of the above look to me like unintentional drift, so addressing here. Let me know if that's not the case though!

### Steps to test this PR

- [ ] Apply below patch:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt b/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
index df50491818..446dd61735 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
@@ -155,10 +155,10 @@ class NewUserOnboardingPlanProvider @Inject constructor(
                 add(initialStep(firstDialog))
                 add(comparisonChartStep())
                 add(defaultBrowserPromptStep())
-                if (showDock) {
+                if (true) {
                     add(addToDockStep())
                 }
-                if (showWidget) {
+                if (true) {
                     add(widgetPromptStep(ctx))
                     add(addWidgetStep(ctx))
                 }
```

- [x] Clean install.
- [x] Walk the linear flow end to end. Verify titles type in, the card does not resize while typing, content below fades in afterwards.
- [x] Tap the card mid-typing. Content snaps into place.
- [x] Rotate on each step. Title appears complete with no animation, card correctly sized.
- [x] Background and re-enter mid-step. No stale title from the previous step.

### UI changes
The only visual difference is caused by the application of universal font padding handling. On screens that previously didn't disable font padding, this results in a minor shift of the whole content up. A before -> after example below:

onboarding_font_padding_before_after_small

---

> [!NOTE]
> **Medium Risk**
> Touches first-run and reinstaller onboarding UI across many steps; visual/timing changes are intentional but need end-to-end verification. Logic is centralized with tests, limiting regression risk outside onboarding.
> 
> **Overview**
> Introduces **`OnboardingDialogTitleView`** (with **`OnboardingDialogTitleController`**) so onboarding dialog titles share one pattern: invisible sizing text for stable card height, typed **`TypeAnimationTextView`**, and centralized **`preventWidows`** + HTML decoding via **`setTitle`**, **`typeTitle`**, **`snapTitle`**, and **`finishTyping`**.
> 
> **`BrandDesignUpdateWelcomePage`**, the address bar picker bottom sheet, and eight step includes swap duplicated FrameLayout twins and local **`startOnboardingTypingAnimation`** helpers for the new view. Skip-mid-animation now calls **`finishTyping()`** on each title instead of **`performClick()`** on views that were typing.
> 
> **Behavior alignment:** all titles get **`preventWidows`** (non-breaking space before the last word) and **`includeFontPadding="false"`**; the input-screen preview title sizes from decoded HTML so the card no longer grows when a `<br/>` splits the title.
> 
> Maestro flows update regex assertions for the new spacing (e.g. **`address.bar`**, **`customize.anything`**). Unit tests cover **`OnboardingDialogTitleController`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf7728e53e3bdbc6274e7c0228f3f74db0ebddad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
@@ -100,7 +100,6 @@ import com.duckduckgo.app.onboardingquicksetup.ui.RemoveWidgetInstructionsBottom
 import com.duckduckgo.app.widget.AddWidgetLauncher
 import com.duckduckgo.appbuildconfig.api.AppBuildConfig
 import com.duckduckgo.common.ui.store.AppTheme
-import com.duckduckgo.common.ui.view.TypeAnimationTextView
 import com.duckduckgo.common.ui.view.addBottomShadow
 import com.duckduckgo.common.ui.view.text.DaxTextView
 import com.duckduckgo.common.ui.view.toPx
@@ -903,9 +902,9 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
                     binding.daxDialogCta.secondaryCta.visibility = if (showSecondaryCta) View.INVISIBLE else View.GONE
 
                     val titleRes = if (isSyncRestore) R.string.syncRestoreDialogBrandDesignTitle else R.string.preOnboardingWelcomeDialogTitle
+                    binding.daxDialogCta.welcomeContent.titleText.setTitle(getString(titleRes))
                     if (isSyncRestore) {
                         // SYNC_RESTORE reuses welcomeContent with its own copy and the sync-restore CTAs.
-                        binding.daxDialogCta.welcomeContent.hiddenTitleText.text = getString(titleRes)
                         binding.daxDialogCta.welcomeContent.bodyText1.text =
                             getString(R.string.syncRestoreDialogBrandDesignBody1).preventWidows().html(requireContext())
                         binding.daxDialogCta.primaryCta.text = getString(R.string.syncRestoreDialogPrimaryCta)
@@ -933,9 +932,7 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
                         },
                         onAnimationEnd = {
                             fadeInDialog {
-                                binding.daxDialogCta.welcomeContent.titleText.startOnboardingTypingAnimation(
-                                    getString(titleRes),
-                                ) {
+                                binding.daxDialogCta.welcomeContent.titleText.typeTitle {
                                     val animators = mutableListOf<Animator>(
                                         ObjectAnimator.ofFloat(binding.daxDialogCta.welcomeContent.bodyText1, View.ALPHA, 1f)
                                             .setDuration(DIALOG_CONTENT_FADE_IN_DURATION),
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
```diff
@@ -1732,15 +1724,14 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
...
                 binding.daxDialogCta.welcomeContent.root.alpha = 1f
-                binding.daxDialogCta.welcomeContent.titleText.cancelAnimation()
                 val titleString = if (isSyncRestore) {
                     getString(R.string.syncRestoreDialogBrandDesignTitle)
                 } else {
                     getString(R.string.preOnboardingWelcomeDialogTitle)
                 }
-                binding.daxDialogCta.welcomeContent.titleText.text = titleString
+                binding.daxDialogCta.welcomeContent.titleText.setTitle(titleString)
+                binding.daxDialogCta.welcomeContent.titleText.snapTitle()
```
mikescamell, line 1733: ❓ Could we create a function on that both sets and snaps? Or bundle setting into the snapTitle function? We do it quite a lot in this class.

mikescamell (commented):
@LukasPaczos 

1. Are we confident we don't need a kill-switch for these changes? Now that onboarding is released any changes we make that could have an adverse affect would have an impact on users. I realise these changes are relatively minor in the grand scheme of things but I think it's worth asking. 

2. I think this is unrelated, but I noticed that when when pressing back from any of the steps, and then opening the app, we get a brief flash of the welcome screen before the step we were on is re-opened, is that related to the LinearOnboardingOrchestrator?

https://github.com/user-attachments/assets/f1b6f43c-8b52-4caa-aaae-eba261d90a6f

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
LukasPaczos: Conceptually, these are separate actions. Setting the title always needs to happen upfront for the correct sizing. Whether we later snap it immediately (like during rotation), or snap it on tap-to-skip, is a separate decision. This might look a tiny bit clunky in the current fragment, but as soon as it's moved to the render engine which is planned next, multiple call sites will disappear, so this shouldn't be an issue.

I'd keep as-is for now and adjust after render engine lands, if needed.

LukasPaczos:
Thanks @mikescamell!

1. I'm not sure a flag buys us much, and it would mean keeping a lot of dead code around. If this breaks, we'd need a patch anyway, given how early in the lifecycle the title views are used. The change is simple, patching or reverting it will be simple too, so the cost of the flag imo isn't justified by a minor gain in breakage-mitigation convenience.
2. Nice catch! The initial state check in the fragment is probably lagging behind orchestrator state after activity is recreated, or something along those lines. Not related to this PR, I'll cut a task to investigate.

mikescamell (approved):
LGTM! :shipit: Tested on a Samsung S25 and Tablet emulator

PR #9341, opened 2026-07-30
PIR: Preserve and forward broker-supplied profile fields native doesn't model

noisysocks:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216773212741753
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/481882893211075/task/1216773212741757

### Description

Native now preserves and forwards new broker-supplied profile fields it doesn't understand, so adding a profile field no longer needs an app release.

Two things worth flagging for review:

- **Behaviour change, not just a refactor.** Profile identity is narrowed to the tuple the `pir_extracted_profiles` unique index actually covers. Previously every scraped field was compared, so any change to one marked the live opt-out `REMOVED` and then silently dropped the replacement insert. Profiles that today flip to `REMOVED` on a scrape change will stop doing so.
- **DB v17 with an auto-migration.** The factory uses `fallbackToDestructiveMigration()`, so the migration is mandatory rather than optional. `17.json` is committed and there's a Robolectric migration test.

Existing rows pick up extras on their next scan, so no backfill is needed — but only for brokers that still return the profile.

### Steps to test this PR

_Extras are stored and forwarded_
- [ ] Add a `PIR` profile and run a scan against a broker whose config returns extra profile fields.
- [ ] Confirm the scan finds profiles and the opt-out proceeds as before.
- [ ] Confirm the `fillForm` payload for the opt-out carries the extra fields through, at both the profile and the address level, and that empty extras are omitted rather than sent as `{}`.

_Upgrade preserves data_
- [ ] Install a build from `develop`, add a profile and run a scan so extracted profiles are stored.
- [ ] Install this branch over the top, without clearing app data.
- [ ] Confirm the previously found profiles and their opt-out state are all still present — i.e. the DB migrated rather than being wiped.

_Re-scan refreshes instead of marking removed_
- [ ] With a profile already found and its opt-out in flight, run a scan again for the same broker.
- [ ] Confirm the profile is refreshed in place rather than duplicated, and that its opt-out is **not** marked `REMOVED`.

### UI changes
| Before  | After |
| ------ | ----- |


noisysocks:
Superseded by #9353, reopened from a branch on this repo rather than a fork.

PR #9347, opened 2026-07-30
Drop defensive CopyOnWriteArrayList from the optimized RealContentScopeScripts

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200905986587319/task/1217019532582642?focus=true

### Description

Two privacy-config lists were rebuilt in place on a worker thread while readers read them on the main thread, and both repositories handed out the live instance.

`RealUnprotectedTemporaryRepository` did `clear()` then one `add()` per row, so a reader could observe any prefix — a millisecond-wide window on every config persist, with ~300 exception domains.
`RealContentScopeScripts` reads that list on every navigation and serializes it into the injected script, so a truncated read meant content-scope features running on sites meant to be exempt.

Both repositories now build the list aside and publish it in a single assignment to a `@Volatile` field, following `RealTrackerAllowlistRepository` in the same module. `UnprotectedTemporaryRepository.exceptions`widens from `CopyOnWriteArrayList<FeatureException>` to `List<FeatureException>`; no `-api` change, and existing test stubs are unaffected. `RealUserAllowListRepository` used a single `addAll()` so it could never tear, and is included for the second reason below.

Because a published list is now immutable, the optimized path in `RealContentScopeScripts` drops its two defensive `CopyOnWriteArrayList` copies and holds the references instead, comparing `!==` then `!=`: identity settles the unchanged case in O(1), and equals still stops a reload that changed nothing from reassembling the script. The legacy path keeps its copies and now clears the optimized baselines, so a mid-session flag flip followed by an input returning to its pre-flip value cannot leave the shared JSON stale.

### Steps to test this PR

- [x] Tests are passing.

With `optimizeContentScopeInjection` ON (default on internal)

- [x] Browse several pages in a row, confirm protections stay applied.
- [ ] Override the privacy config from dev settings, confirm an `unprotectedTemporary` change takes effect.
- [x] Tap the shield to allowlist a site, confirm protections drop, then re-protect it.
- [x] Toggle GPC and Accessibility → force zoom, reload, confirm both take effect. (i.e. YouTube)
- [x] Turn the flag OFF, restart, repeat the above. Behaviour must be identical to develop.
- [x] Flip it back ON without restarting, load a page, confirm protections are still correct.

### NO UI changes

---

> [!NOTE]
> **Medium Risk**
> Changes threading and caching for privacy exemptions and injected content-scope JS; incorrect behavior would mis-apply protections, but the change is well-tested around flag flips and list immutability.
> 
> **Overview**
> Fixes a **race** where `RealUnprotectedTemporaryRepository` and `RealUserAllowListRepository` rebuilt `CopyOnWriteArrayList` in place while readers (including content-scope script injection on every navigation) could see **truncated** exception or allowlist data.
> 
> Both repositories now **build aside and assign** a new `List` to a `@Volatile` field. `UnprotectedTemporaryRepository.exceptions` is typed as `List<FeatureException>` instead of the live `CopyOnWriteArrayList`. Failed reloads keep the previous snapshot; returned lists are not mutated after publish.
> 
> In **`RealContentScopeScripts`**, the optimized path drops defensive `CopyOnWriteArrayList` copies and uses **reference + equality** change detection (`differsFrom`) so equal content on a new list instance does not reassemble the script. **`resetCaches()`** runs when `optimizeContentScopeInjection` flips mid-session so legacy and optimized baselines cannot leave shared JSON stale after flag toggles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 605346c575be42582856b122ae0ca50c4adc5f47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/privacy-config/privacy-config-store/src/main/java/com/duckduckgo/privacy/config/store/features/unprotectedtemporary/UnprotectedTemporaryRepository.kt
+++ b/privacy-config/privacy-config-store/src/main/java/com/duckduckgo/privacy/config/store/features/unprotectedtemporary/UnprotectedTemporaryRepository.kt
@@ -27,7 +27,6 @@ import kotlinx.coroutines.launch
 interface UnprotectedTemporaryRepository {
     fun updateAll(exceptions: List<UnprotectedTemporaryEntity>)
 
-    /** Immutable snapshot. A caller may hold the reference: it is replaced, never mutated. */
     val exceptions: List<FeatureException>
 }
 
@@ -41,9 +40,6 @@ class RealUnprotectedTemporaryRepository(
     private val unprotectedTemporaryDao: UnprotectedTemporaryDao =
         database.unprotectedTemporaryDao()
 
-    // Reloaded on a worker thread while readers are on the main thread, so it is published in a single
-    // assignment: a reader sees either the previous list or the complete new one. Refilling in place would
-    // let a reader observe any prefix of the list mid-reload.
     @Volatile
     private var snapshot: List<FeatureException> = emptyList()
 
--- a/privacy-config/privacy-config-store/src/main/java/com/duckduckgo/privacy/config/store/features/unprotectedtemporary/UnprotectedTemporaryRepository.kt
+++ b/privacy-config/privacy-config-store/src/main/java/com/duckduckgo/privacy/config/store/features/unprotectedtemporary/UnprotectedTemporaryRepository.kt
@@ -23,11 +23,12 @@ import com.duckduckgo.privacy.config.store.UnprotectedTemporaryEntity
 import com.duckduckgo.privacy.config.store.toFeatureException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
-import java.util.concurrent.CopyOnWriteArrayList
 
 interface UnprotectedTemporaryRepository {
     fun updateAll(exceptions: List<UnprotectedTemporaryEntity>)
-    val exceptions: CopyOnWriteArrayList<FeatureException>
+
+    /** Immutable snapshot. A caller may hold the reference: it is replaced, never mutated. */
+    val exceptions: List<FeatureException>
 }
 
 class RealUnprotectedTemporaryRepository(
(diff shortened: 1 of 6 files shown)


anikiki:
* **#9347** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9347?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9347?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

privacy-config/privacy-config-store/src/main/java/com/duckduckgo/privacy/config/store/features/unprotectedtemporary/UnprotectedTemporaryRepository.kt
```diff
@@ -39,7 +39,11 @@ class RealUnprotectedTemporaryRepository(
...
     private val unprotectedTemporaryDao: UnprotectedTemporaryDao =
         database.unprotectedTemporaryDao()
-    override val exceptions = CopyOnWriteArrayList<FeatureException>()
+
+    @Volatile
+    private var snapshot: List<FeatureException> = emptyList()
+
+    override val exceptions get() = snapshot
 
     init {
```
CrisBarreiro, line 48: While we're here, can we move this from `init` to `onCreate` from `MainProcessLifecycleObserver`?

CrisBarreiro (approved):

privacy-config/privacy-config-store/src/main/java/com/duckduckgo/privacy/config/store/features/unprotectedtemporary/UnprotectedTemporaryRepository.kt
anikiki: @CrisBarreiro I think it's bigger than it looks, and it's better to do it as a follow-up as part of the modularization AoI.
 
`privacy-config-store` currently has no di and no lifecycle dependencies. Implementing `MainProcessLifecycleObserve`r means adding :di + lifecycle-runtime to a pure persistence module. Other classes in the same module (ContentBlocking, Gpc, Https, Drm, etc) use the identical init pattern, so doing just this one makes the module inconsistent.

PR #9351, opened 2026-07-30
Stop resolving contentScope experiments on every navigation

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1217017440666479?focus=true

### Description

Content scope scripts are injected from onPageStarted, and that injection first awaits ContentScopeExperiments.getActiveExperiments(). Resolving it materialises every feature toggle in the app through a reflective proxy (~580), splits each toggle key twice (~1,140 string splits), then enrols and evaluates the 22 contentScopeExperiments sub-toggles.

Injection already races the page's own scripts, so this is protections coverage as well as page-load cost: it's time the page runs without our protections.

The result only changes when a new privacy config lands, so it's now resolved once and held until then, invalidated via PrivacyConfigCallbackPlugin. Cache hits return without suspending — the caller awaits this on the main thread, so dispatching to io() and back would cost a main looper turn even for a hit. Concurrent callers serialise on a mutex so tabs navigating together resolve once between them.

Behind cacheContentScopeExperiments. The flag is sampled rather than read per call (it's itself a toggle, so reading it per navigation would put back part of the cost). With the flag off, resolve runs straight through without taking the mutex, exactly as it behaves today.

Behaviour change: enrolment is re-evaluated when a config is persisted rather than on every navigation.

### Steps to test this PR

The flag defaults to enabled on internal builds. Both the cached result and the sampled flag are cleared on privacy config updates, so restart the app after changing either.

- [x] Browse normally across several sites and confirm no behaviour change: pages load, protections apply, no crashes.
- [x] Confirm enrolment still works: enable a contentScopeExperiment* sub-toggle under contentScopeExperiments in the internal feature-toggle inventory, restart the app, browse, then check that a cohort has been assigned (stored state in the inventory, or the experiment_enroll pixel).
- [x] Confirm invalidation: relaunch so a fresh privacy config is downloaded and persisted, and check experiments are re-resolved rather than serving a stale list.
- [x] Turn cacheContentScopeExperiments off, restart, and repeat 1–2 — behaviour should be identical, resolving on every navigation.
- [x] Unit tests are passing. `./gradlew :content-scope-scripts-impl:testDebugUnitTest` — `RealContentScopeExperimentsTest` covers resolve-once, re-resolve after persisted/downloaded, and resolve-per-call with the flag off.

Two notes for reviewers. With the cache on, flipping an experiment sub-toggle by hand in the toggle inventory won't take effect until the next config download or restart, because a manual toggle edit doesn't fire the config callback — that's expected, not a bug. And the latency win isn't observable in-app; the wide-event intervals that make it measurable in the field come in a follow-up PR.

### NO UI changes

---

> [!NOTE]
> **Medium Risk**
> Changes when experiment cohorts are resolved relative to navigation and privacy config, which affects content-scope script injection timing and experiment assignment; mitigated by a kill-switch and invalidation on config updates.
> 
> **Overview**
> **Caches** the resolved active content-scope experiment toggles so `getActiveExperiments()` no longer walks and enrols every app toggle on each navigation—work that currently blocks document-start script injection on page start.
> 
> When the new **`cacheContentScopeExperiments`** kill-switch is on (default **internal**), `RealContentScopeExperiments` keeps the list in memory, returns cache hits without extra suspension, and uses a mutex so concurrent navigations resolve once. **`PrivacyConfigCallbackPlugin`** clears the cache on config download and persist; an invalidation counter avoids storing a result if config changes mid-resolve. With the flag off, behavior matches today (resolve every call, no mutex).
> 
> **Behavior change:** experiment enrolment is re-evaluated when privacy config is persisted/downloaded, not on every navigation—manual toggle edits in the inventory won’t refresh until the next config event or restart.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e41d027d9a149aea55d6a2c31be67bbdd6affc4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/features/contentscopeexperiments/RealContentScopeExperiments.kt
+++ b/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/features/contentscopeexperiments/RealContentScopeExperiments.kt
@@ -55,14 +55,6 @@ class RealContentScopeExperiments @Inject constructor(
     @Volatile
     private var activeExperiments: List<Toggle>? = null
 
-    /**
-     * Sampled rather than read per call because the flag is itself a toggle: reading it on every navigation would put
-     * back a chunk of the cost this is here to remove. Cleared alongside the result so a remote change to the
-     * kill-switch takes effect at the next privacy config update rather than needing a restart.
-     */
-    @Volatile
-    private var cachingEnabled: Boolean? = null
-
     private val resolveMutex = Mutex()
 
     override suspend fun getActiveExperiments(): List<Toggle> {
@@ -85,13 +77,12 @@ class RealContentScopeExperiments @Inject constructor(
 
     private fun invalidate() {
         activeExperiments = null
-        cachingEnabled = null
     }
 
     private suspend fun cachingEnabled(): Boolean =
-        cachingEnabled ?: withContext(dispatcherProvider.io()) {
+        withContext(dispatcherProvider.io()) {
             contentScopeScriptsFeature.cacheContentScopeExperiments().isEnabled()
-        }.also { cachingEnabled = it }
+        }
 
     private suspend fun resolve(): List<Toggle> =
         withContext(dispatcherProvider.io()) {
--- a/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/features/contentscopeexperiments/RealContentScopeExperiments.kt
+++ b/content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/features/contentscopeexperiments/RealContentScopeExperiments.kt
@@ -29,6 +29,7 @@ import dagger.SingleInstanceIn
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
+import java.util.concurrent.atomic.AtomicInteger
 import javax.inject.Inject
 
 @SingleInstanceIn(AppScope::class)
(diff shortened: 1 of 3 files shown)


anikiki:
* **#9351** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9351?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/features/contentscopeexperiments/RealContentScopeExperiments.kt
```diff
@@ -18,20 +18,82 @@ package com.duckduckgo.contentscopescripts.impl.features.contentscopeexperiments
...
+    @Volatile
+    private var activeExperiments: List<Toggle>? = null
+
+    /**
+     * Sampled rather than read per call because the flag is itself a toggle: reading it on every navigation would put
+     * back a chunk of the cost this is here to remove. Cleared alongside the result so a remote change to the
+     * kill-switch takes effect at the next privacy config update rather than needing a restart.
+     */
+    @Volatile
+    private var cachingEnabled: Boolean? = null
```
CrisBarreiro: RC flag reads are cached internally by default, so from a performance perspective, we don't need to cache. We could also observe `contentScopeScriptsFeature.cacheContentScopeExperiments().enabled()`  if we wanted, so we always have the latest value

CrisBarreiro:
@anikiki 

>  Confirm enrolment still works: enable a contentScopeExperiment* sub-toggle under contentScopeExperiments in the internal feature-toggle inventory, restart the app, browse, then check that a cohort has been assigned (stored state in the inventory, or the experiment_enroll pixel).


this isn't working 100% of the time for me, even if restarting the app. I can see it work if I disable the optimization, though. I think there might be a race condition between page load (experiments retrieved and stored for the first time), and cache invalidation

CrisBarreiro (changes requested):

content-scope-scripts/content-scope-scripts-impl/src/main/java/com/duckduckgo/contentscopescripts/impl/features/contentscopeexperiments/RealContentScopeExperiments.kt
anikiki: Good catch! Updated 👍 
Also added one more test.

CrisBarreiro (approved):

PR #9353, opened 2026-07-31
PIR: Forward extras on extracted profiles

noisysocks:
Task/Issue URL:
https://app.asana.com/1/137249556945/task/1216773212741753
https://app.asana.com/1/137249556945/project/481882893211075/task/1216773212741757

### Description

Support `extras` in extracted profiles and addresses, pass them along to `fillForm`, and update stored profiles with the freshly scraped profile. This lessens the need for an app release when an opt out form requires a new field we don't have.

### Steps to test this PR

1. Point C-S-S to [Support extracting extras from profiles and addresses content-scope-scripts#2907](https://github.com/duckduckgo/content-scope-scripts/pull/2907):

```
npm install @duckduckgo/content-scope-scripts@github:duckduckgo/content-scope-scripts#pr-releases/randerson/pir-extras-open-field
```

2. Build and run.
4. Open Settings → Pir Dev Settings → Broker Config.
5. Select `cyberbackgroundchecks.com` and replace the JSON with: https://dub.duckduckgo.com/duckduckgo/dbp-api/blob/2fb2bb29f5d3f67bc6c70098ca36db7ca4a100b9/dbp-json/data/json/cyberbackgroundchecks.com.json
6. Complete a scan and opt out: Debug Scan → Run opt out → PIR Email → Debug opt out. 
7. The street and zip fields in the post-confirmation opt out form should be filled from data that was extracted during the scan.

---

> [!NOTE]
> **Medium Risk**
> Touches PIR local storage, scan refresh semantics, and opt-out identity matching; changes are well-tested but incorrect merge or key logic could affect pending opt-outs or removal detection.
> 
> **Overview**
> Adds **`extras`** on extracted profiles and addresses so broker configs can scrape arbitrary fields without a native release, and passes the fuller profile through to C-S-S for **`fillForm`**.
> 
> On repeat scans, **`saveExtractedProfiles`** refreshes rows that already match the DB unique key (profile query, broker, name, URL, identifier) instead of ignoring them, merging extras so keys missing from a new scrape keep stored values. Profile identity for removals/reappearances uses a narrower key and **`matches`** ignores address extras so new scraped fields don’t look like a different person.
> 
> **`ExtractProfileSelectors`** becomes a passthrough map so extract profile config isn’t dropped at parse time. Room bumps to **v17** with an **`extras`** column, Moshi map converters, and migration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71d26d75b1cfb1de6e6c01f9ed8005f44fc8a619. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirRunStateHandler.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirRunStateHandler.kt
@@ -668,6 +668,7 @@ class RealPirRunStateHandler @Inject constructor(
                                     city = item.city,
                                     state = item.state,
                                     fullAddress = item.fullAddress,
+                                    extras = item.extras.orEmpty(),
                                 )
                             },
                             phoneNumbers = it.phoneNumbers,
@@ -676,13 +677,14 @@ class RealPirRunStateHandler @Inject constructor(
                             reportId = it.reportId.orEmpty(),
                             email = it.email.orEmpty(),
                             fullName = it.fullName.orEmpty(),
+                            extras = it.extras.orEmpty(),
                         )
                     }.also {
                         /**
                          * For every locally stored extractedProfile for the broker x profile that is not part of the newly received
                          * extracted Profiles, or no extracted Profiles were found on the broker:
                          * - We update the optOut status to REMOVED
-                         * - We store the new extracted profiles (if profile query is not deprecated). We ignore the ones that already exist.
+                         * - We store the new extracted profiles (if profile query is not deprecated) and refresh the ones we already have.
                          * - Update the corresponding ScanJobRecord
                          * For every stored extractedProfile that reappears in the results and whose opt-out was previously
                          * REMOVED, we revert it back to REQUESTED and report a reappearance pixel.
@@ -695,7 +697,7 @@ class RealPirRunStateHandler @Inject constructor(
                                     pixelSender.reportBrokerOptOutProfileReappeared(state.broker.url)
                                 }
                             jobRecordUpdater.updateScanMatchesFound(it, brokerName, state.profileQueryId)
-                            repository.saveNewExtractedProfiles(it)
+                            repository.saveExtractedProfiles(it)
                         } else {
                             jobRecordUpdater.updateScanNoMatchFound(brokerName, state.profileQueryId)
                         }
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
@@ -46,6 +46,7 @@ internal fun <T> List<T>.splitIntoParts(parts: Int): List<List<T>> {
 internal fun ExtractedProfile.toParams(fullName: String): ExtractedProfileParams {
     return ExtractedProfileParams(
         name = this.name.ifEmpty { null },
+        alternativeNames = this.alternativeNames,
         profileUrl = this.profileUrl.ifEmpty { null },
         fullName = fullName.ifEmpty { null },
         email = this.email.ifEmpty { null },
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
@@ -16,8 +16,10 @@
 
 package com.duckduckgo.pir.impl.common
 
+import com.duckduckgo.pir.impl.models.AddressCityState
 import com.duckduckgo.pir.impl.models.ExtractedProfile
 import com.duckduckgo.pir.impl.scripts.models.ExtractedProfileParams
+import com.duckduckgo.pir.impl.scripts.models.ExtractedProfileParams.AddressParams
 
 internal fun <T> List<T>.splitIntoParts(parts: Int): List<List<T>> {
     if (this.isEmpty()) {
@@ -47,8 +49,54 @@ internal fun ExtractedProfile.toParams(fullName: String): ExtractedProfileParams
         profileUrl = this.profileUrl.ifEmpty { null },
         fullName = fullName.ifEmpty { null },
         email = this.email.ifEmpty { null },
+        age = this.age.ifEmpty { null },
+        addresses = this.addresses.map {
+            AddressParams(
+                city = it.city,
+                state = it.state,
+                extras = it.extras,
+            )
+        },
+        phoneNumbers = this.phoneNumbers,
+        relatives = this.relatives,
+        identifier = this.identifier.ifEmpty { null },
+        extras = this.extras,
     )
 }
+
+/**
+ * Two addresses describe the same place when their city and state match. Everything else on an
+ * address is detail scraped alongside it. Mirrors `sameCityState` in C-S-S.
+ */
+internal fun AddressCityState.sameCityState(other: AddressCityState): Boolean =
+    this.city == other.city && this.state == other.state
+
+/**
+ * Returns [scraped] with the attributes we own locally carried over from the stored record.
+ *
+ * Extras are merged rather than replaced: a key absent from the new scrape keeps the value we
+ * already hold, so a broker changing its page layout doesn't strip fields that a pending opt-out
+ * still needs. Address extras are merged onto the address with the same city and state.
+ */
+internal fun ExtractedProfile.refreshedWith(scraped: ExtractedProfile): ExtractedProfile {
+    val storedAddresses = this.addresses
+    return scraped.copy(
+        dbId = this.dbId,
+        dateAddedInMillis = this.dateAddedInMillis,
+        deprecated = this.deprecated,
+        extras = this.extras + scraped.extras,
+        addresses = scraped.addresses.map { address ->
+            val stored = storedAddresses.firstOrNull { it.sameCityState(address) }
+            address.copy(extras = stored?.extras.orEmpty() + address.extras)
+        },
+    )
+}
+
+/**
+ * Extras are excluded wherever we compare profiles for identity: a broker config that starts
+ * scraping a new field must not make a record we already hold look like a different one.
+ */
+internal fun List<AddressCityState>.withoutExtras(): List<AddressCityState> = this.map { it.copy(extras = emptyMap()) }
 
 internal fun ExtractedProfile.hasMatchingProfileOnParent(extractedProfiles: List<ExtractedProfile>): Boolean {
     return extractedProfiles.any {
@@ -60,7 +108,7 @@ internal fun ExtractedProfile.matches(extractedProfile: ExtractedProfile): Boole
     return this.name == extractedProfile.name && this.age == extractedProfile.age &&
         this.alternativeNames.isASubSetOrSuperSetOf(extractedProfile.alternativeNames) &&
         this.relatives.isASubSetOrSuperSetOf(extractedProfile.relatives) &&
-        this.addresses.isASubSetOrSuperSetOf(extractedProfile.addresses)
+        this.addresses.withoutExtras().isASubSetOrSuperSetOf(extractedProfile.addresses.withoutExtras())
 }
 
 private fun <T> List<T>.isASubSetOrSuperSetOf(other: List<T>): Boolean {
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
@@ -98,6 +98,23 @@ internal fun ExtractedProfile.refreshedWith(scraped: ExtractedProfile): Extracte
  */
 internal fun List<AddressCityState>.withoutExtras(): List<AddressCityState> = this.map { it.copy(extras = emptyMap()) }
 
+internal data class ExtractedProfileKey(
+    val profileQueryId: Long,
+    val brokerName: String,
+    val name: String,
+    val profileUrl: String,
+    val identifier: String,
+)
+
+internal fun ExtractedProfile.toKey(): ExtractedProfileKey =
+    ExtractedProfileKey(
+        profileQueryId = this.profileQueryId,
+        brokerName = this.brokerName,
+        name = this.name,
+        profileUrl = this.profileUrl,
+        identifier = this.identifier,
+    )
+
 internal fun ExtractedProfile.hasMatchingProfileOnParent(extractedProfiles: List<ExtractedProfile>): Boolean {
     return extractedProfiles.any {
         it.brokerName == this.brokerName && this.matches(it)
@@ -106,9 +123,9 @@ internal fun ExtractedProfile.hasMatchingProfileOnParent(extractedProfiles: List
 
 internal fun ExtractedProfile.matches(extractedProfile: ExtractedProfile): Boolean {
     return this.name == extractedProfile.name && this.age == extractedProfile.age &&
-        this.alternativeNames.isASubSetOrSuperSetOf(extractedProfile.alternativeNames) &&
-        this.relatives.isASubSetOrSuperSetOf(extractedProfile.relatives) &&
-        this.addresses.withoutExtras().isASubSetOrSuperSetOf(extractedProfile.addresses.withoutExtras())
+            this.alternativeNames.isASubSetOrSuperSetOf(extractedProfile.alternativeNames) &&
+            this.relatives.isASubSetOrSuperSetOf(extractedProfile.relatives) &&
+            this.addresses.withoutExtras().isASubSetOrSuperSetOf(extractedProfile.addresses.withoutExtras())
 }
 
 private fun <T> List<T>.isASubSetOrSuperSetOf(other: List<T>): Boolean {
--- a/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
+++ b/pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
@@ -123,9 +123,9 @@ internal fun ExtractedProfile.hasMatchingProfileOnParent(extractedProfiles: List
 
 internal fun ExtractedProfile.matches(extractedProfile: ExtractedProfile): Boolean {
     return this.name == extractedProfile.name && this.age == extractedProfile.age &&
-            this.alternativeNames.isASubSetOrSuperSetOf(extractedProfile.alternativeNames) &&
-            this.relatives.isASubSetOrSuperSetOf(extractedProfile.relatives) &&
-            this.addresses.withoutExtras().isASubSetOrSuperSetOf(extractedProfile.addresses.withoutExtras())
+        this.alternativeNames.isASubSetOrSuperSetOf(extractedProfile.alternativeNames) &&
+        this.relatives.isASubSetOrSuperSetOf(extractedProfile.relatives) &&
+        this.addresses.withoutExtras().isASubSetOrSuperSetOf(extractedProfile.addresses.withoutExtras())
 }
 
 private fun <T> List<T>.isASubSetOrSuperSetOf(other: List<T>): Boolean {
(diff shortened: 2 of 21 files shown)


noisysocks:
**CI status:** `Android CI tests` failed on one test — `AtbIntegrationTest#whenNoStatisticsStoredThenAtbInitializationSuccessfullyStoresAtb` (Pixel2 API 28, `java.lang.AssertionError`). Not caused by this diff; I've re-run the failed jobs.

Why it's unrelated:

- This branch changes only `pir/pir-impl` and `pir/pir-internal` files — nothing in `:app` or `:statistics`.
- `AtbIntegrationTest` talks to the live ATB server (the class KDoc calls out that it fails without internet access).
- `StatisticsRequester.initializeAtb()` calls `store.clearAtb()` in its error handler, so any failure of the `/atb` or `/exti` request makes `assertTrue(statisticsStore.hasInstallationStatistics)` fail.
- The other two tests in the class passed. They don't hit `/exti`, which points at a transient failure on that request rather than a general outage.

All other checks are green: Unit tests, Unit tests (Metro), Lint, Lint (Metro), Code Formatting, and both debug APK builds. `develop` is currently green, so this isn't a pre-existing base-branch failure either.

I'll keep watching and will follow up if the re-run doesn't come back clean.

---
_Generated by [Claude Code](https://claude.ai/code)_

noisysocks:
Correcting my previous comment: I said `develop` was green and this wasn't a pre-existing base-branch failure. That was wrong — **it is a pre-existing base-branch failure**, and I'd missed a red `develop` run when I checked.

`AtbIntegrationTest` is flaky against the live ATB server. Evidence across three runs of the same `Android CI tests` job:

| Run | Device | Failing tests |
|---|---|---|
| `develop` @ [`7a9170b`](https://github.com/duckduckgo/Android/actions/runs/30606355641/job/91079377685) (Update content scope scripts to 15.23.0, #9354) | MediumPhone API 30 | all 3 |
| This PR, [attempt 1](https://github.com/duckduckgo/Android/actions/runs/30786061464/job/91599753916) | Pixel2 API 28 | `whenNoStatisticsStoredThenAtbInitializationSuccessfullyStoresAtb` |
| This PR, [attempt 2](https://github.com/duckduckgo/Android/actions/runs/30786061464/job/91601972348) | MediumPhone API 30 | the *other two* — and the one from attempt 1 passed |

The failing subset rotates between runs and moves across devices, and it already failed on `develop` on a commit unrelated to this branch. All failures are bare `java.lang.AssertionError` from the network-dependent assertions. `flakyTestAttempts = 1`, so each of these is already a post-retry failure.

This also drops the `/exti`-specific theory from my earlier comment: the two tests that failed in attempt 2 never call `/exti`, so the flakiness spans the ATB requests generally rather than one endpoint.

Nothing here touches this PR's diff, which is confined to `pir/pir-impl` and `pir/pir-internal`. I've re-run once; I'm not going to keep re-rolling the dice, since the fix belongs in `AtbIntegrationTest` (in `:app`) and not in a PIR change. Everything else on this PR is green.

---
_Generated by [Claude Code](https://claude.ai/code)_

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/db/ExtractedProfileDao.kt
```diff
@@ -51,6 +52,9 @@ interface ExtractedProfileDao {
     @Insert(onConflict = OnConflictStrategy.IGNORE)
     fun insertNewExtractedProfiles(extractedProfiles: List<StoredExtractedProfile>)
 
+    @Update
```
landomen, line 55: I think we will need to use `OnConflictStrategy.REPLACE` here. This defines what happens if you try to insert a record that has the same id as an existing record. 

The default is `OnConflictStrategy.ABORT` which simply rejects the update. `OnConflictStrategy.REPLACE` will update it like we want.

The primary key of `StoredExtractedProfile` is an auto-incremented `id` which we copy over when updating the extracted profile, so this update will fail because keys will already exist.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
```diff
@@ -869,6 +887,26 @@ class RealPirRepository(
             },
         )
 
+    /**
+     * Identifies a record within the unique index on the extracted profiles table.
+     */
+    private data class ExtractedProfileStorageKey(
```
landomen: nit: move the data class to bottom of the class

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
```diff
@@ -556,12 +564,22 @@ class RealPirRepository(
...
+                .associateBy { it.storageKey() }
+
+            val (alreadyStored, newlyFound) = extractedProfiles.partition { it.storageKey() in storedProfiles }
+
+            dao.insertNewExtractedProfiles(newlyFound.map { it.toStoredExtractedProfile() })
+            dao.updateExtractedProfiles(
+                alreadyStored.map { scraped ->
+                    storedProfiles.getValue(scraped.storageKey()).refreshedWith(scraped).toStoredExtractedProfile()
+                },
+            )
```
landomen: This is doing three separate queries to read, write and update the table. Wrap it in a transaction so if an error happens the state is still correct.

Example: User could mark a profile as deprecated from the dashboard while we are updating that same profile, overwriting users change.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirRunStateHandler.kt
```diff
@@ -695,7 +697,7 @@ class RealPirRunStateHandler @Inject constructor(
                                     pixelSender.reportBrokerOptOutProfileReappeared(state.broker.url)
                                 }
                             jobRecordUpdater.updateScanMatchesFound(it, brokerName, state.profileQueryId)
-                            repository.saveNewExtractedProfiles(it)
+                            repository.saveExtractedProfiles(it)
```
landomen, line 700: If we've previously submitted an opt-out for a profile with age=35 and a later scan returns the same profile with age=36, removal/reappearance detection (toKey(), all modeled fields) won't match it, while the storage refresh (storageKey(), name/profileUrl/identifier only) will. Result: scan N marks the opt-out REMOVED and refreshes the row → scan N+1 matches again → reverts to REQUESTED and fires the reappearance pixel (line 697) for a profile that never went away.

> To be fair, this is better than before this PR — previously the row was never refreshed, so the record stayed falsely REMOVED forever. But any modeled-field drift from a broker now produces one cycle of false-REMOVED → REQUESTED plus a spurious reappearance pixel, so the reappearance metric will pick up noise correlated with broker markup changes.

Aligning toKey() with storageKey() would close the gap (and is consistent with how storage already treats these as the same profile) — though it does loosen what counts as "removed", so worth checking against what the other platforms do.

landomen (changes requested):
@noisysocks The changes generally look good, and PIR state was kept after installing updated build 👍 

Left a few comments that we need to address, most importantly the transaction one since it can cause data inconsistencies.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
noisysocks: Good catch — fixed in 492df89e. The read, insert and update now run inside `db.runInTransaction { }`, along with the deprecated-user-profile check so that read is part of the same snapshot.

Your example is the exact hazard: `refreshedWith` carries `deprecated` over from the stored row, so a dashboard edit landing between the read and the write was being clobbered.

One knock-on: `RealPirRepositoryTest` mocks `PirDatabase`, so a mocked `runInTransaction` would never invoke the lambda and the existing save tests would pass vacuously. I stubbed it in `setUp()` to run the `Runnable` so those tests still exercise the real path.

---
_Generated by [Claude Code](https://claude.ai/code)_

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
noisysocks: Done in 492df89e — `ExtractedProfileStorageKey` now sits at the bottom of the class, just above the companion object.

---
_Generated by [Claude Code](https://claude.ai/code)_

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/db/ExtractedProfileDao.kt
noisysocks: I dug into this one and I don't think we need it — and I think `REPLACE` here would actively hurt. Happy to be argued out of it though.

`OnConflictStrategy` on `@Update` isn't about the primary key. Room generates `UPDATE OR ABORT … WHERE id = ?` — the PK is the *lookup* key, not a conflict source. Reusing the stored `id` is what makes the update find the right row, so it can't collide with itself.

The strategy only comes into play if the new column values would violate a **unique index**. `StoredExtractedProfile` has one, on `(profileQueryId, brokerName, name, profileUrl, identifier)` — and `storageKey()` is exactly those five columns. We only take the update path when the scraped profile's `storageKey()` matched a stored one, so all five index columns are byte-identical to what's already in that row. Nothing to conflict with.

What changes on the update path is `age`, `addresses`, `extras`, etc. — none of them indexed.

The reason I'd avoid `REPLACE`: `UPDATE OR REPLACE` doesn't mean "update instead of abort", it means "on conflict, **DELETE** the conflicting rows and carry on". So if a genuine unique-index conflict ever did arise, `REPLACE` would silently delete other extracted-profile rows rather than surfacing the bug. `ABORT` is the safer default.

I've added `ExtractedProfileDaoTest` in 492df89e to pin this down against real SQLite rather than a mocked DAO — it inserts, then updates via the same path with changed `age`/`extras`, and asserts the row is updated in place with its id intact and neighbouring rows untouched. If that passes in CI, the current strategy is fine; if it throws, you're right and I'll switch it.

Worth noting the existing `RealPirRepositoryTest` mocks `ExtractedProfileDao`, so it could never have caught this either way — that gap is what prompted the new test.

---
_Generated by [Claude Code](https://claude.ai/code)_

noisysocks: Apologies, I was trying out `/autofix-pr` in Claude Code… Not a fan of how it spits long replies out at unsuspecting reviewers! 😅

My reading here is that with an update the only way we’d get a conflict is if we’re violating the unique index:

```kt
Index(value = ["profileQueryId", "brokerName", "name", "profileUrl", "identifier"], unique = true)
```

But this isn’t possible since this `storageKey()` tuple is exactly what we’re partitioning on:

```kt
val (alreadyStored, newlyFound) = extractedProfiles.partition { it.storageKey() in storedProfiles }

dao.insertNewExtractedProfiles(newlyFound.map { it.toStoredExtractedProfile() })
dao.updateExtractedProfiles(
    alreadyStored.map { scraped →
        storedProfiles.getValue(scraped.storageKey()).refreshedWith(scraped).toStoredExtractedProfile()
    },
)
```

It’s possible a future code path might call `updateExtractedProfiles` with something that would violate the index. But I’m not sure `OnConflictStrategy.REPLACE` is what we want here since that would delete the existing row. `pir_optout_job_record` / `pir_email_confirmation_job_record` are keyed on `extractedProfileId` with no foreign key.

So I’m not sure. What do you suggest?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
```diff
@@ -542,26 +550,42 @@ class RealPirRepository(
...
+                        .map { it.toExtractedProfile() }
+                        .associateBy { it.storageKey() }
+
+                    val (alreadyStored, newlyFound) = extractedProfiles.partition { it.storageKey() in storedProfiles }
+
+                    dao.insertNewExtractedProfiles(newlyFound.map { it.toStoredExtractedProfile() })
+                    dao.updateExtractedProfiles(
+                        alreadyStored.map { scraped ->
+                            storedProfiles.getValue(scraped.storageKey()).refreshedWith(scraped).toStoredExtractedProfile()
+                        },
```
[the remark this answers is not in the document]
noisysocks: Already raised in https://github.com/duckduckgo/Android/pull/9353#discussion_r3722739852.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirRunStateHandler.kt
noisysocks: What other platforms do:

|             | Reconciliation (removed/reappeared)                              | Storage identity                                   |
| ----------- | ---------------------------------------------------------------- | -------------------------------------------------- |
| **Apple**   | `identifier` only (falls back to `profileUrl`)                   | **same function** — one identity check drives both |
| **Windows** | `Identifier` only, case-insensitive (falls back to `ProfileUrl`) | full `Profile` JSON blob unique index              |
| **Android** | `toKey()` — all 13 modeled fields                                | `storageKey()` — 5 fields                          |

So yes aligning `toKey()` with `storageKey()` wouldn’t make what counts as “removed” looser than other platforms. In fact it’d still be far more strict.

What do you think? Should I combine `toKey()` and `storageKey()`? Or go further and align with Apple/Windows by looking only at `identifier`? The latter feels like might be a rabbit hole – but happy to go down it if you suggest.

noisysocks:
@landomen: Addressed / responded to your notes. I’m no Android dev so leaning on asking what you’d suggest more than I ordinarily would 😅

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scripts/models/BrokerAction.kt
```diff
@@ -160,6 +160,7 @@ data class ElementSelector(
     val min: String?,
     val max: String?,
     val failSilently: Boolean?,
+    val format: String? = null,
```
noisysocks, line 145: Late addition to support https://github.com/duckduckgo/content-scope-scripts/pull/2936.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/PirRepository.kt
```diff
@@ -542,26 +550,42 @@ class RealPirRepository(
...
-            }
 
-            extractedProfiles
-                .map {
-                    it.toStoredExtractedProfile()
-                }.also {
-                    extractedProfileDao()?.insertNewExtractedProfiles(it)
-                }
+            db.runInTransaction(
+                Runnable {
```
landomen: nit: I think you can remove Runnable and just have `db.runInTransaction { ... }`

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/store/db/ExtractedProfileDao.kt
landomen: Ah, good catch, thanks for checking. You're right, I missed that we're partitioning the profiles before we update. 
And correct that the resolution strategy checks the indexes since id has to match 😅 

I think no changes are needed then 👍

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirRunStateHandler.kt
landomen: I agree with you that changing to `identifier` only is too risky at this point. Let's align the `toKey()` and `storageKey()` to have a single check if that's possible.

landomen (commented):
@noisysocks I see unit tests are failing because of an unrelated issue. Could you try to rebase the branch with latest develop to see if it resolves it?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/scripts/models/BrokerAction.kt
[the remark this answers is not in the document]
noisysocks: Hmmm good point. Fixed.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
```diff
@@ -47,9 +49,72 @@ internal fun ExtractedProfile.toParams(fullName: String): ExtractedProfileParams
         profileUrl = this.profileUrl.ifEmpty { null },
         fullName = fullName.ifEmpty { null },
         email = this.email.ifEmpty { null },
+        age = this.age.ifEmpty { null },
+        addresses = this.addresses.map {
+            AddressParams(
+                city = it.city,
+                state = it.state,
+                extras = it.extras,
```
landomen, line 58: We don't forward `fullAddress` here, is that intentional?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
```diff
@@ -47,9 +49,72 @@ internal fun ExtractedProfile.toParams(fullName: String): ExtractedProfileParams
...
+            )
+        },
+        phoneNumbers = this.phoneNumbers,
+        relatives = this.relatives,
+        identifier = this.identifier.ifEmpty { null },
+        extras = this.extras,
```
landomen, line 64: What about `alternativeNames`?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
```diff
@@ -47,9 +49,72 @@ internal fun ExtractedProfile.toParams(fullName: String): ExtractedProfileParams
...
+ * already hold, so a broker changing its page layout doesn't strip fields that a pending opt-out
+ * still needs. Address extras are merged onto the address with the same city and state.
+ */
+internal fun ExtractedProfile.refreshedWith(scraped: ExtractedProfile): ExtractedProfile {
+    val storedAddresses = this.addresses
+    return scraped.copy(
+        dbId = this.dbId,
+        dateAddedInMillis = this.dateAddedInMillis,
+        deprecated = this.deprecated,
+        extras = this.extras + scraped.extras,
```
landomen, line 88: Can this result in duplicated extras?

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
```diff
@@ -47,9 +49,72 @@ internal fun ExtractedProfile.toParams(fullName: String): ExtractedProfileParams
...
+        addresses = scraped.addresses.map { address ->
+            val stored = storedAddresses.firstOrNull { it.sameCityState(address) }
+            address.copy(extras = stored?.extras.orEmpty() + address.extras)
```
landomen, line 91: Same question here about duplicated addresses

landomen (approved):
Left few minor comments, otherwise I ran a scan on this branch and on develop and got same results so I think we're good 👍

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
noisysocks: C-S-S only consumes city, state, and extras.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
noisysocks: I’ll add it!

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
noisysocks: No and whenRefreshedWithThenMergesProfileExtrasKeepingKeysMissingFromTheScrape tests this.

pir/pir-impl/src/main/java/com/duckduckgo/pir/impl/common/PirUtils.kt
noisysocks: No and whenRefreshedWithNewAddressThenKeepsOnlyItsOwnExtras tests this.

PR #9365, opened 2026-07-31
[Config-driven linear onboarding dialogs] Extract render engine and controllers

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1216978525560082?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/1208671518894266/task/1216854264994244?focus=true

### Description

Step 2 of the config-driven onboarding dialogs work.

- Each onboarding step is now described as a value-comparable `DialogConfig`.
- A slim `ViewModel` resolves the orchestrator's current step into the config, and `DialogRenderEngine` diffs it
  against the previous config to drive seven collaborators — one per visual axis, plus the shared card
  choreography.
  - The VM already contains some Fragment callbacks and commands that are directly copy-pasted from the legacy impl and currently unused. These will be used by screens migrated in follow-ups.
- `COMPARISON_CHART` (stateless, with a bespoke content entrance) and `ADDRESS_BAR_POSITION` (stateful) are
  ported end to end. The whole arm sits behind `OnboardingBrandDesignUpdateToggles#configDrivenDialogs`. The
  legacy renderer and its `ViewModel` are untouched.
- Unported steps (including intro animation) auto-advance and apply default side effects (settings writes, click pixels).
- "Presented" pixels are not universally sent, will be addressed when all steps are ported.

#### Divergences from the tech design

| Change | Why |
|---|---|
| `ContentHandle` gains **`onContentReady`** alongside `afterFade` | `afterFade` is an `Animator` the engine owns end to end. Some entrance work can't be modelled that way (an unbounded loop, or animation driven outside the animator framework, for example `INPUT_SCREEN`'s toggle animation), so it needs a side-effect hook that still runs exactly once per render on the animated, snapped and skipped paths. |
| Added `CardArrowController` | TD already defined `DialogConfig.cardArrow`, so this is the place for that logic. |
| Interactive content revealed in `afterFade`/`onContentReady` must be **gated by the binder** | The design assumed `isAnimating` → `interceptChildTouches` covered the whole entrance. It doesn't: interception is released as `afterFade` starts (so that the already visible CTA buttons are clickable). Anything interactive revealed in `afterFade` is tappable while invisible (for example, `INPUT_SCREEN_PREVIEW` step's suggestion buttons), so binders need to gate that directly. Documented on `ContentHandle` and will impact only the `INPUT_SCREEN_PREVIEW` step later on. |

#### Behaviour changes vs the legacy renderer (ported parts)

| Change | Detail |
|---|---|
| When content animations are skipped, so are background/Dax | In legacy, skipping animations meant only snapping card content into place while background/Dax animations kept playing. This is aligned now but to be confirmed with @mikescamell if the original behavior wasn't intentional. |
| Card anchoring has a **single owner** | Legacy's vertical bias on tablets disagreed between steps (bias 0.5 in quick setup and add-to-dock, 0 elsewhere), and the fit corrector overrode them again whenever a decoration stopped fitting. `CardAnchorResolver` now aligns the bias across all steps. To let the resolver be the only owner, the shared `OnboardingDecorationFitCorrector` no longer writes the anchor itself. That reset moved into `BrandDesignUpdateWelcomePage#onDecorationHidden`, setting the same three properties on the same view, so the legacy renderer behaves as before. |
| Undecorated screens **reserve a band** to maintain height | Legacy decided per screen. `ADD_TO_DOCK` [held on to the outgoing bottom wing's anchor](https://github.com/duckduckgo/Android/blob/cbe8848963a6459215a032f83dc42c197d68d08c/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt#L1089) so the card stayed put, even though the wing itself was hidden, and `INPUT_SCREEN_PREVIEW` [pinned it to the parent bottom](https://github.com/duckduckgo/Android/blob/cbe8848963a6459215a032f83dc42c197d68d08c/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt#L1472-L1481). Both assumed a fixed step order. `Embellishment.None` now declares its own independent placement, backed by a `Space`, so the card keeps a comparable height on every undecorated screen. No anchoring rule reads which screen rendered before it, which matters because step order is meant to be reshuffled freely. |

Example of undecorated card's position on a tablet:
- On the left is the legacy behavior, which `ADD_TO_DOCK` and `INPUT_SCREEN_PREVIEW` had to work around.
- On the right, the new behavior for undecorated cards that maintains height similar to a decorated card, which makes transitions smoother.

side_by_side

This can be tested now by applying this diff:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt
index 0685e14a4d..080b0f734c 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt
@@ -36,8 +36,8 @@ class DialogConfigResolver @Inject constructor() {
 
         is NewUserOnboardingActivityDialog.AddressBarPosition -> DialogConfig(
             background = OnboardingBackgroundStep.AddressBar,
-            embellishment = Embellishment.BobbingDax,
-            cardArrow = CardArrowConfig.AtEnd,
+            embellishment = Embellishment.None,
+            cardArrow = CardArrowConfig.Hidden,
             content = ContentConfig.AddressBar(
                 title = TextConfig.Resource(R.string.preOnboardingAddressBarTitle),
                 initialPosition = OmnibarType.SINGLE_TOP,

```
but will be easier to exercise when rest of the steps are migrated.

### Steps to test this PR

_flag off_
- [ ] Clean install.
- [ ] Go through onboarding, verify all steps are handled.

_transitions and skips_
- [ ] Apply below patch:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
index b6a1e4d352..41a8d85ad4 100644
--- a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
@@ -58,6 +58,6 @@ interface OnboardingBrandDesignUpdateToggles {
     /**
      * Selects the config-driven renderer for the brand-design onboarding dialogs.
      */
-    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
     fun configDrivenDialogs(): Toggle
 }

```
- [x] Clean install.
- [x] Verify the permissions prompt, browser comparison chart, and address bar position steps execute.
- [x] Verify animations:
  - [x] Fade transitions.
  - [x] Comparison table staggered animation.
  - [x] Step indicator transition.
  - [x] On tablet: once any transition finishes, the card stays in its final position, doesn't snap to new anchor
- [x] Tap the card mid-animation. Content snaps into place.
- [x] Rotate on each step. Content appears complete with no animation, card correctly sized.
- [x] Rotate on each step while animating. Content appears complete with no animation, card correctly sized.
- [x] Background and re-enter mid-step. No stale animations, content snaps into place.
- [x] Select bottom address bar position and click "next". Verify the settings is applied.

_fit calculation_
- [x] Use a phone in portrait orientation.
- [x] Pause on the browser comparison screen.
- [x] Keep increasing the font until the **wing barely doesn't fit and hides**.
- [x] Continue to the next step and verify that the card transitions and bobbing Dax re-appears smoothly.

_arrow slide_
- [ ] Apply this diff:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt
index 0685e14a4d..eedd2d785d 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/DialogConfigResolver.kt
@@ -67,7 +67,7 @@ class DialogConfigResolver @Inject constructor() {
     private fun comparisonChart(chart: ComparisonChartConfig) = DialogConfig(
         background = OnboardingBackgroundStep.ComparisonChart,
         embellishment = Embellishment.BottomWing,
-        cardArrow = CardArrowConfig.AtEnd,
+        cardArrow = CardArrowConfig.AtStart,
         content = ContentConfig.ComparisonChart(title = TextConfig.Resource(chart.titleRes), config = chart),
         primaryCta = CtaConfig(
             text = TextConfig.Resource(chart.primaryCtaTextRes),

```
- [ ] Verify that the arrow slides when transitioning between browser comparison and address bar position steps.

---

> [!NOTE]
> **Medium Risk**
> Large new onboarding UI stack on a feature flag; first-run UX and animation/fit behavior differ from legacy in several documented ways until all steps are migrated.
> 
> **Overview**
> Adds a **config-driven onboarding path** gated by `configDrivenDialogs()` (default off). When brand design update and that toggle are on, onboarding uses `ConfigDrivenWelcomePageFragment` and `buildConfigDrivenPageBlueprints()` instead of the legacy brand-design welcome page.
> 
> Each step is described as value-comparable **`DialogConfig`**; `ConfigDrivenOnboardingPageViewModel` maps orchestrator steps via `DialogConfigResolver`, and **`DialogRenderEngine`** diffs configs and drives collaborators (background, embellishments, card anchor/arrow/stage, step indicator, content binders). Comparison chart and address bar position are ported end-to-end; other dialogs auto-advance with existing side effects until follow-ups land.
> 
> **Layout/fit:** `OnboardingDecorationFitCorrector` no longer repins the card when a decoration hides—hosts handle that (`BrandDesignUpdateWelcomePage`, config-driven embellishment path). An **`undecoratedBand`** `Space` plus `Embellishment.None` placement keeps undecorated screens at a stable card height. **`CardAnchorResolver`** centralizes card anchoring and arrow depth.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4d0f59a0af9dd74e17b8a76e5d9f4b0caf22bfa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
@@ -60,10 +60,12 @@ class OnboardingViewModel @Inject constructor(
         val isBrandDesignUpdateEnabled = withContext(dispatchers.io()) {
             onboardingBrandDesignUpdateToggles.brandDesignUpdate().isEnabled()
         }
-        if (isBrandDesignUpdateEnabled) {
-            pageLayoutManager.buildBrandDesignUpdatePageBlueprints()
-        } else {
-            pageLayoutManager.buildPageBlueprints()
+        val isConfigDrivenDialogsEnabled = isBrandDesignUpdateEnabled &&
+            withContext(dispatchers.io()) { onboardingBrandDesignUpdateToggles.configDrivenDialogs().isEnabled() }
+        when {
+            isConfigDrivenDialogsEnabled -> pageLayoutManager.buildConfigDrivenPageBlueprints()
+            isBrandDesignUpdateEnabled -> pageLayoutManager.buildBrandDesignUpdatePageBlueprints()
+            else -> pageLayoutManager.buildPageBlueprints()
         }
     }
 
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
@@ -57,16 +57,18 @@ class OnboardingViewModel @Inject constructor(
     val viewState = _viewState.asStateFlow()
 
     suspend fun initializePages() {
-        val isBrandDesignUpdateEnabled = withContext(dispatchers.io()) {
-            onboardingBrandDesignUpdateToggles.brandDesignUpdate().isEnabled()
+        val renderer = withContext(dispatchers.io()) { resolveRenderer() }
+        when (renderer) {
+            OnboardingRenderer.ConfigDriven -> pageLayoutManager.buildConfigDrivenPageBlueprints()
+            OnboardingRenderer.BrandDesignUpdate -> pageLayoutManager.buildBrandDesignUpdatePageBlueprints()
+            OnboardingRenderer.Legacy -> pageLayoutManager.buildPageBlueprints()
         }
-        val isConfigDrivenDialogsEnabled = isBrandDesignUpdateEnabled &&
-            withContext(dispatchers.io()) { onboardingBrandDesignUpdateToggles.configDrivenDialogs().isEnabled() }
-        when {
-            isConfigDrivenDialogsEnabled -> pageLayoutManager.buildConfigDrivenPageBlueprints()
-            isBrandDesignUpdateEnabled -> pageLayoutManager.buildBrandDesignUpdatePageBlueprints()
-            else -> pageLayoutManager.buildPageBlueprints()
-        }
+    }
+
+    private fun resolveRenderer(): OnboardingRenderer = when {
+        !onboardingBrandDesignUpdateToggles.brandDesignUpdate().isEnabled() -> OnboardingRenderer.Legacy
+        onboardingBrandDesignUpdateToggles.configDrivenDialogs().isEnabled() -> OnboardingRenderer.ConfigDriven
+        else -> OnboardingRenderer.BrandDesignUpdate
     }
 
     fun pageCount(): Int {
@@ -156,4 +158,6 @@ class OnboardingViewModel @Inject constructor(
         DUCK_AI_FOCUSED,
         DEFAULT_WITHOUT_INTRO_CTA,
     }
+
+    private enum class OnboardingRenderer { Legacy, BrandDesignUpdate, ConfigDriven }
 }
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
@@ -546,7 +546,14 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
             root = binding.root,
             dialog = binding.daxDialogCta.root,
             cardContainer = binding.daxDialogCta.cardContainer,
-            onDecorationHidden = { binding.daxDialogCta.cardView.setArrowDepthFraction(0f) },
+            onDecorationHidden = {
+                binding.daxDialogCta.cardView.setArrowDepthFraction(0f)
+                binding.daxDialogCta.root.updateLayoutParams<ConstraintLayout.LayoutParams> {
+                    verticalBias = 0f
+                    bottomToTop = ConstraintLayout.LayoutParams.UNSET
+                    bottomToBottom = ConstraintLayout.LayoutParams.PARENT_ID
+                }
+            },
             cardBottomInsetPx = { cardBottomInsetPx },
         ).also { it.attach() }
 
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
@@ -24,11 +24,6 @@ import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.core.view.isGone
 import androidx.core.view.updateLayoutParams
 
-/**
- * @param onDecorationHidden the tracked decoration no longer fits and has been hidden. The card's own constraints
- *   are the host's to write, including re-anchoring it to the parent bottom now the decoration is gone: a second
- *   writer here would compete with whatever rule the host applies at render time.
- */
 class OnboardingDecorationFitCorrector(
     private val root: View,
     private val dialog: View,
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
@@ -24,6 +24,11 @@ import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.core.view.isGone
 import androidx.core.view.updateLayoutParams
 
+/**
+ * @param onDecorationHidden the tracked decoration no longer fits and has been hidden. The card's own constraints
+ *   are the host's to write, including re-anchoring it to the parent bottom now the decoration is gone: a second
+ *   writer here would compete with whatever rule the host applies at render time.
+ */
 class OnboardingDecorationFitCorrector(
     private val root: View,
     private val dialog: View,
@@ -116,11 +121,6 @@ class OnboardingDecorationFitCorrector(
 
         if (target == null) {
             deco.isGone = true
-            dialog.updateLayoutParams<ConstraintLayout.LayoutParams> {
-                verticalBias = 0f
-                bottomToTop = ConstraintLayout.LayoutParams.UNSET
-                bottomToBottom = ConstraintLayout.LayoutParams.PARENT_ID
-            }
             onDecorationHidden()
             return false
         }
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
@@ -44,6 +44,7 @@ import com.duckduckgo.app.onboarding.ui.page.OnboardingBackgroundAnimator
 import com.duckduckgo.app.onboarding.ui.page.OnboardingPageFragment
 import com.duckduckgo.app.onboarding.ui.page.configdriven.engine.BackgroundControllerImpl
 import com.duckduckgo.app.onboarding.ui.page.configdriven.engine.CardAnchorControllerImpl
+import com.duckduckgo.app.onboarding.ui.page.configdriven.engine.CardAnchorResolver
 import com.duckduckgo.app.onboarding.ui.page.configdriven.engine.CardArrowControllerImpl
 import com.duckduckgo.app.onboarding.ui.page.configdriven.engine.CardStageImpl
 import com.duckduckgo.app.onboarding.ui.page.configdriven.engine.ContentControllerImpl
@@ -142,7 +143,7 @@ class ConfigDrivenWelcomePage : OnboardingPageFragment(R.layout.content_onboardi
             windowInsets
         }
 
-        val cardAnchor = CardAnchorControllerImpl(binding, deviceInfo.isTablet())
+        val cardAnchor = CardAnchorControllerImpl(binding, CardAnchorResolver(deviceInfo.isTablet()))
 
         engine = DialogRenderEngine(
             content = ContentControllerImpl(
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
@@ -142,6 +142,8 @@ class ConfigDrivenWelcomePage : OnboardingPageFragment(R.layout.content_onboardi
             windowInsets
         }
 
+        val cardAnchor = CardAnchorControllerImpl(binding, deviceInfo.isTablet())
+
         engine = DialogRenderEngine(
             content = ContentControllerImpl(
                 binding = binding.daxDialogCta,
@@ -157,10 +159,12 @@ class ConfigDrivenWelcomePage : OnboardingPageFragment(R.layout.content_onboardi
             ),
             embellishments = EmbellishmentControllerImpl(
                 binding = binding,
-                onDecorationHidden = { binding.daxDialogCta.cardView.setArrowDepthFraction(0f) },
+                // A decoration that stops fitting leaves the card anchored to a hidden view, so re-run the same
+                // anchor rule the render applies, which drops the arrow's depth along with it.
+                onDecorationHidden = { cardAnchor.apply(null) },
                 cardBottomInsetPx = { cardBottomInsetPx },
             ),
-            cardAnchor = CardAnchorControllerImpl(binding, deviceInfo.isTablet()),
+            cardAnchor = cardAnchor,
             cardArrow = CardArrowControllerImpl(binding.daxDialogCta.cardView),
             stepIndicator = StepIndicatorControllerImpl(binding.daxDialogCta.stepIndicator),
             emit = viewModel::onEvent,
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
@@ -93,8 +93,6 @@ class ConfigDrivenWelcomePage : OnboardingPageFragment(R.layout.content_onboardi
     /** Fed to the embellishment controller's fit corrector; kept in sync by the window-insets listener below. */
     private var cardBottomInsetPx = 0
 
-    private var hasRenderedOnce = false
-
     private val requestNotificationPermission = registerForActivityResult(ActivityResultContracts.RequestPermission()) { granted ->
         if (view?.windowVisibility == View.VISIBLE) {
             viewModel.notificationPermissionFlowFinished(granted)
@@ -200,25 +198,19 @@ class ConfigDrivenWelcomePage : OnboardingPageFragment(R.layout.content_onboardi
 
     private fun renderConfig(state: ConfigDrivenOnboardingPageViewModel.ViewState) {
         val engine = engine ?: return
+        settleIntroViews()
+
+        // A retained view model emits before a recreated view has been laid out, and the decoration fit
+        // check measures the root's height, so measuring at 0 would hide the decoration for good.
+        if (!binding.root.isLaidOut) {
+            binding.root.doOnLayout { renderConfig(viewModel.viewState.value) }
+            return
+        }
+
         val stepId = state.stepId ?: return
         val config = state.config ?: return
-
-        if (!hasRenderedOnce) {
-            hasRenderedOnce = true
-            settleIntroViews()
-            // A retained view model emits before a recreated view has been laid out, and the decoration fit
-            // check measures the root's height — measuring at 0 would hide the decoration for good.
-            binding.root.doOnLayout {
-                val live = viewModel.viewState.value
-                val liveStepId = live.stepId ?: return@doOnLayout
-                val liveConfig = live.config ?: return@doOnLayout
-                engine.render(liveStepId, liveConfig, live.animateEntry)
-                viewModel.onDialogRendered(liveStepId)
-            }
-        } else {
-            engine.render(stepId, config, state.animateEntry)
-            viewModel.onDialogRendered(stepId)
-        }
+        engine.render(stepId, config, state.animateEntry)
+        viewModel.onDialogRendered(stepId)
     }
 
     private fun handleCommand(command: ConfigDrivenOnboardingPageViewModel.Command) {
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
@@ -1,0 +1,270 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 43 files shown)


LukasPaczos:
bugbot run

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
```diff
@@ -116,11 +116,6 @@ class OnboardingDecorationFitCorrector(
 
         if (target == null) {
             deco.isGone = true
-            dialog.updateLayoutParams<ConstraintLayout.LayoutParams> {
-                verticalBias = 0f
-                bottomToTop = ConstraintLayout.LayoutParams.UNSET
-                bottomToBottom = ConstraintLayout.LayoutParams.PARENT_ID
-            }
```
LukasPaczos, line 123: This is the only change to the "flag off" code path. This code is moved directly to `BrandDesignUpdateWelcomePage`, so no behavior change.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/BrandDesignUpdateWelcomePage.kt
```diff
@@ -546,7 +546,14 @@ class BrandDesignUpdateWelcomePage : OnboardingPageFragment(R.layout.content_onb
...
             cardContainer = binding.daxDialogCta.cardContainer,
-            onDecorationHidden = { binding.daxDialogCta.cardView.setArrowDepthFraction(0f) },
+            onDecorationHidden = {
+                binding.daxDialogCta.cardView.setArrowDepthFraction(0f)
+                binding.daxDialogCta.root.updateLayoutParams<ConstraintLayout.LayoutParams> {
+                    verticalBias = 0f
+                    bottomToTop = ConstraintLayout.LayoutParams.UNSET
+                    bottomToBottom = ConstraintLayout.LayoutParams.PARENT_ID
+                }
+            },
```
LukasPaczos, line 556: Refs https://github.com/duckduckgo/Android/pull/9365/changes#r3690849481, this is the only change to production codebase, just a moved functional call.

LukasPaczos:
bugbot run

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/engine/DialogRenderEngine.kt
```diff
@@ -0,0 +1,210 @@
...
+            stepIndicator.skipRunning()
+            cardArrow.skipRunning()
+            isAnimating = false
+        } finally {
+            settling = false
+        }
+    }
+
+    /** Teardown: suppresses everything still pending instead of settling it. */
+    fun release() {
```
mikescamell, line 141: ❓Should we have a cardArrow release function that's called here as well?

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePage.kt
```diff
@@ -0,0 +1,266 @@
...
+    private val viewModel by lazy {
+        ViewModelProvider(this, viewModelFactory)[ConfigDrivenOnboardingPageViewModel::class.java]
+    }
+
+    private var engine: DialogRenderEngine? = null
+
+    /** Fed to the embellishment controller's fit corrector; kept in sync by the window-insets listener below. */
+    private var cardBottomInsetPx = 0
+
+    private var hasRenderedOnce = false
```
mikescamell: 🤖 Companion to the `renderConfig` suggestion below. With the layout check in place this flag has no readers left.

```suggestion
    private var cardBottomInsetPx = 0
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePageFragment.kt
```diff
@@ -0,0 +1,266 @@
...
+                val liveStepId = live.stepId ?: return@doOnLayout
+                val liveConfig = live.config ?: return@doOnLayout
+                engine.render(liveStepId, liveConfig, live.animateEntry)
+                viewModel.onDialogRendered(liveStepId)
+            }
+        } else {
+            engine.render(stepId, config, state.animateEntry)
+            viewModel.onDialogRendered(stepId)
+        }
+    }
```
mikescamell, line 225: 🤖 `hasRenderedOnce` is a one-shot flag standing in for a question the view can answer directly. If a second config arrives before the first `doOnLayout` fires, the `else` branch renders against an unlaid-out root, which is the exact failure the flag exists to prevent: the fit check measures 0 and hides the decoration for good. It is narrow, since a second render normally needs user input and that needs layout, but `isLaidOut` is symmetric and self-correcting, and it is already the idiom in `CardStageImpl.morph`.

The recursion is depth 1: `doOnLayout` fires after a layout pass, so the retry takes the direct branch. Repeated pre-layout emissions each register their own listener, but each one reads live state and `render`'s identity guard makes the duplicates no-ops.

One behaviour note: `settleIntroViews()` now runs on every render rather than once. It only sets `isVisible = false` and `alpha = 0f`, so it is idempotent.

```suggestion
    private fun renderConfig(state: ConfigDrivenOnboardingPageViewModel.ViewState) {
        val engine = engine ?: return
        settleIntroViews()

        // A retained view model emits before a recreated view has been laid out, and the decoration fit
        // check measures the root's height, so measuring at 0 would hide the decoration for good.
        if (!binding.root.isLaidOut) {
            binding.root.doOnLayout { renderConfig(viewModel.viewState.value) }
            return
        }

        val stepId = state.stepId ?: return
        val config = state.config ?: return
        engine.render(stepId, config, state.animateEntry)
        viewModel.onDialogRendered(stepId)
    }
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/binders/AddressBarBinder.kt
```diff
@@ -0,0 +1,59 @@
...
+            state.collect { addressBarPicker.setSelection(it.position, animate = true) }
+        }
+
+        addressBarTitle.setTitle(content.title.resolve(root.context))
+
+        ContentHandle(
+            title = addressBarTitle,
+            fadeTargets = listOf(addressBarPicker),
+            result = { NewUserOnboardingEvent.AddressBarConfirmed(state.value.position) },
+        )
```
mikescamell, line 58: 🤖 The picker keeps this listener after unbind, closing over the store-owned flow. `CardStageImpl.bindCta` clears its click listener first for exactly this reason ("cleared first so a listener closing over an already-unbound handle is never retained"), so the two paths read inconsistently today.

`setOnSelectionChangedListener` takes a non-null lambda, so a no-op is the clear.

```suggestion
        ContentHandle(
            title = addressBarTitle,
            fadeTargets = listOf(addressBarPicker),
            result = { NewUserOnboardingEvent.AddressBarConfirmed(state.value.position) },
            unbind = { addressBarPicker.setOnSelectionChangedListener {} },
        )
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/engine/ContentController.kt
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.onboarding.ui.page.configdriven.engine
+
+import android.view.View
+import androidx.core.view.isVisible
```
mikescamell, line 21: 🤖 Companion to the `resetStage` suggestion below.

```suggestion
import androidx.core.view.children
import androidx.core.view.isVisible
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/engine/ContentController.kt
```diff
@@ -0,0 +1,94 @@
...
+            binding.welcomeContent.root,
+            binding.comparisonChartContent.root,
+            binding.addressBarContent.root,
+            binding.inputScreenContent.root,
+            binding.inputScreenPreviewContent.root,
+            binding.reinstallerQuickSetupContent.root,
+            binding.addToDockContent.root,
+            binding.widgetPromptContent.root,
+        ).forEach { it.isVisible = false }
+    }
```
mikescamell, line 63: 🤖 This list has no compile-time guard and the KDoc spells out what a missed entry costs, so it will drift as binders land in the follow-ups. The includes are all siblings in `cardContainer`, so deriving the list removes the drift.

Worth knowing while you are in here: only `welcomeContent` actually defaults to visible. The other seven are already `android:visibility="gone"` in the layout, so the hand-written list is doing less work than it looks.

`resetStage()` runs before `showCtaButtons`, so excluding the two CTAs is not strictly required, but leaving them out keeps this off the card stage's turf.

```suggestion
    /**
     * Covers every content include, not only the ones with a binder: `welcomeContent` defaults to visible in the
     * card layout, so a first render of any other screen would otherwise leave it stacked above, reserving blank
     * height inside the card. The CTAs share the container but belong to the card stage.
     */
    override fun resetStage() {
        binding.cardContainer.children
            .filter { it !== binding.primaryCta && it !== binding.secondaryCta }
            .forEach { it.isVisible = false }
    }
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePageFragment.kt
```diff
@@ -0,0 +1,270 @@
...
+import javax.inject.Inject
+import com.duckduckgo.mobile.android.R as CommonR
+
+/**
+ * Wires [ConfigDrivenOnboardingPageViewModel] to [DialogRenderEngine]: it owns the views, builds the engine's
+ * collaborators from the inflated binding, and routes commands. Every rendering decision belongs to the engine.
+ */
+@InjectWith(FragmentScope::class)
+class ConfigDrivenWelcomePage : OnboardingPageFragment(R.layout.content_onboarding_welcome_page_update) {
+
```
mikescamell, line 70: 🔍 As we're making these changes I wonder if it's finally worth adding Fragment to the end of these Pages, seeing as we append Activity/Fragment elsewhere in the app

app/src/main/java/com/duckduckgo/app/onboarding/ui/OnboardingViewModel.kt
```diff
@@ -60,10 +60,12 @@ class OnboardingViewModel @Inject constructor(
...
-        } else {
-            pageLayoutManager.buildPageBlueprints()
+        val isConfigDrivenDialogsEnabled = isBrandDesignUpdateEnabled &&
+            withContext(dispatchers.io()) { onboardingBrandDesignUpdateToggles.configDrivenDialogs().isEnabled() }
+        when {
+            isConfigDrivenDialogsEnabled -> pageLayoutManager.buildConfigDrivenPageBlueprints()
+            isBrandDesignUpdateEnabled -> pageLayoutManager.buildBrandDesignUpdatePageBlueprints()
+            else -> pageLayoutManager.buildPageBlueprints()
         }
     }
```
mikescamell: 🤖 The second toggle read is a suspend call on the right-hand side of `&&`, so it needs a `withContext` of its own and the flow dispatches to IO twice. Resolving the whole three-way choice inside one IO block collapses that to a single dispatch, and the short-circuit survives: `configDrivenDialogs()` is still only read when `brandDesignUpdate()` is on, so `whenInitializePagesCalledAndConfigDrivenDialogsEnabledButBrandDesignUpdateDisabledThenBuildPageBlueprints` keeps meaning what it means.

Deliberately leaves the `build*Blueprints()` calls on the caller's thread. Wrapping the whole `when` in one `withContext` would be a smaller diff, but it would move `buildPageBlueprints()` and `shouldShowDefaultBrowserPage()` off the main thread on the legacy path too, and that route wants to stay identical to develop.

Keeping IO for the reads is right either way: `isEnabled()` is cache-backed and thread-safe, but onboarding is a guaranteed first reader, so the cache miss can really hit disk here.

You may prefer the enum to sit next to `ExtendedOnboardingFlow` rather than here.

```suggestion
        val renderer = withContext(dispatchers.io()) { resolveRenderer() }
        when (renderer) {
            OnboardingRenderer.ConfigDriven -> pageLayoutManager.buildConfigDrivenPageBlueprints()
            OnboardingRenderer.BrandDesignUpdate -> pageLayoutManager.buildBrandDesignUpdatePageBlueprints()
            OnboardingRenderer.Legacy -> pageLayoutManager.buildPageBlueprints()
        }
    }

    private fun resolveRenderer(): OnboardingRenderer = when {
        !onboardingBrandDesignUpdateToggles.brandDesignUpdate().isEnabled() -> OnboardingRenderer.Legacy
        onboardingBrandDesignUpdateToggles.configDrivenDialogs().isEnabled() -> OnboardingRenderer.ConfigDriven
        else -> OnboardingRenderer.BrandDesignUpdate
    }

    private enum class OnboardingRenderer { Legacy, BrandDesignUpdate, ConfigDriven }
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/engine/CardStage.kt
```diff
@@ -0,0 +1,220 @@
...
+        fade.start()
+    }
+
+    override fun settle() {
+        pendingMorph?.let { continuation ->
+            pendingMorph = null
+            continuation()
+        }
+        drain { it.end() }
+    }
```
mikescamell, line 172: 🤖 `settle()` runs the continuation and clears the slot, but nothing ends the `ChangeBounds` itself. `beginDelayedTransition` pauses and resumes transitions already running on the scene root rather than ending them, so the superseded one keeps going and fires `onTransitionEnd` later. By then `pendingMorph` holds the *next* render's continuation, and line 104 has no way to tell whose transition just ended, so that render's chain advances before its bounds land: the title types and the content fades in during the card resize.

The same gap has a second, more visible effect. `skipRunningAnimations()` promises to settle every axis, but the bounds tween carries on because only its callback was consumed.

Ending the transition covers both, and it lands before the new `pendingMorph` is installed since `render()` calls `skipRunningAnimations()` first. The listener at 103 then fires inside this call and runs the continuation, so the explicit block below becomes the no-op fallback for a snapped render.

`release()` does not need the same treatment: it nulls `pendingMorph`, so a late `onTransitionEnd` finds nothing to run, and the hierarchy is going away regardless.

```suggestion
    override fun settle() {
        // Ends the transition, not just its continuation: a superseded ChangeBounds is paused and resumed by the
        // next beginDelayedTransition rather than ended, so it would otherwise fire into the next render's slot.
        TransitionManager.endTransitions(binding.root as ViewGroup)
        pendingMorph?.let { continuation ->
            pendingMorph = null
            continuation()
        }
        drain { it.end() }
    }
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/engine/EmbellishmentController.kt
```diff
@@ -0,0 +1,534 @@
...
+    ): SettledDecoration? {
+        // An earlier transition's exit may still be running. Draining it now keeps two exits off the stage at once.
+        drainInFlight()
+
+        if (previous == next) {
+            // The drain may have cut this decoration's own entrance short, so snap it to where that entrance was
+            // heading before reporting the fit.
+            decorations[next]?.snap?.invoke()
+            return applyFit(next)
+        }
```
mikescamell, line 116: 🤖 This branch snaps the decoration visible and then returns whatever `applyFit` says, so a reused decoration that no longer fits stays on stage at its old height while `cardAnchor` moves the card down to the parent bottom. The changed-embellishment branch below already handles it (`decorations[next]?.hide?.invoke()` when `settled == null`), so the two paths disagree.

Reachable on a consecutive pair of dialogs sharing an embellishment, when the second one's card is tall enough to squeeze the decoration out: a large font scale or a short viewport.

Keeping `snap()` before `applyFit` since the fit measurement wants the view at its settled size.

```suggestion
        if (previous == next) {
            // The drain may have cut this decoration's own entrance short, so snap it to where that entrance was
            // heading before reporting the fit.
            decorations[next]?.snap?.invoke()
            val settled = applyFit(next)
            if (settled == null) decorations[next]?.hide?.invoke()
            return settled
        }
```

mikescamell:
@LukasPaczos Overall looking good! Added some comments by me and from a Claude review (indicated by 🤖 prefix)

1. I noticed we get a white screen flash when rotating on either the browser comparison or address bar dialogs. Samsung S25

https://github.com/user-attachments/assets/81be78ce-5712-46a1-98d8-be075b874827

2. The browser comparison background seems to jump up on the y axis now before it transitions out and down. Tablet emulator

https://github.com/user-attachments/assets/5ab99fd4-d037-4ee1-90e8-a075aa42c462

3. This is live as well, but Bobbing Dax seems to be sitting too high in landscape phone on my Samsung S25, could we add it as a follow up to fix after this refactor is complete? Should be an easy fix.

image

LukasPaczos:
1. I've seen this one before, but was never able to consistently reproduce. Can't now either:

https://github.com/user-attachments/assets/bc3628f8-0c9d-42cb-8a22-e2f0d659e705

Is that something you can reliably trigger?

2. Fixed in https://github.com/duckduckgo/Android/pull/9365/commits/b4d0f59a0af9dd74e17b8a76e5d9f4b0caf22bfa.

3. Tracking in https://app.asana.com/1/137249556945/project/1202552961248957/task/1217145974942449. 

This is ready for another round!

mikescamell:
> 1. I've seen this one before, but was never able to consistently reproduce. Can't now either:
> 
>  Screen_recording_20260803_184729.mp4 
> Is that something you can reliably trigger?
> 
> 2. Fixed in [b4d0f59](https://github.com/duckduckgo/Android/commit/b4d0f59a0af9dd74e17b8a76e5d9f4b0caf22bfa).
> 3. Tracking in [app.asana.com/1/137249556945/project/1202552961248957/task/1217145974942449](https://app.asana.com/1/137249556945/project/1202552961248957/task/1217145974942449).
> 
> This is ready for another round!

1. I see the black screen in your video, I guess which is what I was seeing but it was White for me as I was in light mode. But now I cannot reproduce on my Samsung S25 with the up to date branch. Let's keep an eye out 👍

mikescamell (approved):
RE-tested on Samsung S25 and Tablet emulator, all looks good! LGTM

I'm sure we might notice other things as we get further into the refactor so didn't spend too long re-testing post review changes 👍

PR #9360, opened 2026-07-31
Sync: Extract shared crypto/transport helpers

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/608920331025315/task/1217068100697968?focus=true
Tech Design URL (if applicable): 
API Proposals URL(s) (if applicable):

### Description
Refactor to make future work easier / cleaner to land.

Extracts reusable helpers from existing `account_info` code for
- `ProtectedKeyUnwrapper`
- a shared `toUnparsedError()` function
- some crypto constants that were starting to get repeated in a lot of places

- [ ] Fresh install `internal` build
- [ ] Visit `Sync Dev Settings`
- [ ] Scroll down and tap on `Create account` (one-tap enabling of sync on a new account)
- [ ] Scroll up and tap `Create & Register account_info Key` (ensures there is an existing ddg-wrapped protected key)
- [ ] Then tap on `Create 3party credential` and verify that succeeds 

---

> [!NOTE]
> **Medium Risk**
> Touches protected-key decryption and 3party credential key re-wrap logic; behavior is intended to be equivalent but errors in shared crypto paths could break credential creation or key adoption.
> 
> **Overview**
> Refactors sync crypto and HTTP error handling so upcoming unified-device work can reuse the same paths instead of duplicating them.
> 
> **Protected key unwrapping** is centralized in a new `ProtectedKeyUnwrapper` that recovers raw private bytes from `ProtectedKeyEntry` for **ddg** (libsodium secretbox under the account secret key) and **3party** (HKDF + symmetric JWE under the scoped password). `RealThirdPartyCredentialManager` now calls `protectedKeyUnwrapper.unwrap` when re-encrypting existing ddg keys for a new 3party credential, replacing inline Base64 decode + `nativeLib.decryptData`.
> 
> **HKDF main-key derivation** is aligned via shared constants `HKDF_INFO_MAIN_ENCRYPTION_KEY` and `MAIN_ENCRYPTION_KEY_LENGTH_BYTES` in `ScopedCredentialsEncoding`; `DeviceFieldDecryptor` and the unwrapper use them instead of scattered `"Main Key"` / `32` literals.
> 
> **Failed Retrofit responses** that do not parse error JSON share `Response.toUnparsedError()` on `SyncServiceRemote`, used from `SetKeysIfAbsentCall` and token rescope error mapping (replacing duplicate `mapError` / `mapRescopeTokenError` body reads).
> 
[...]

PR #9369, opened 2026-07-31
Sync: Write & verify device_info via dev settings

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/608920331025315/task/1217068100697970?focus=true

### Description
Dev-only plumbing for the unified device list's `device_info` capability. Accessible only from `Sync Dev Settings` screen for now. 

#### Rename via device_info PATCH works
- [ ] Fresh install `internal` build
- [ ] Open `Sync Dev Settings`, and tap `Create account` to quickly sign into sync
- [ ] Tap `Create & Register account_info Key` button
- [ ] Tap `Rename device (PATCH device_info)` button. Call it `test` and verify it shows `PATCH ok` with the chosen name
- [ ] Tap `Fetch & decrypt devices` and verify both legacy and device_info has `name=test`

#### 2nd device confirmation
- [ ] Install on a 2nd device, and sync with the first device
- [ ] Visit production `Sync & Backup` screen and verify you see `test` as the other device's name

---

> [!NOTE]
> **High Risk**
> Touches sync cryptography, protected key unwrap, and server-side device identity; incorrect encryption or PATCH semantics could break cross-device names or leak mishandled key material paths.
> 
> **Overview**
> Adds the **unified device list** write path: when this device’s name changes, the client encrypts a cross-credential `device_info` JWE (under the cached `account_info` public key) and still encrypts legacy `name`/`type` for older clients, then **PATCHes** `/sync/devices` with all three fields every time (the server clears `info` if omitted).
> 
> New sync-impl pieces include **`AccountInfoPrivateKeyProvider`** (unwrap `account_info` private key for decrypt), **`DeviceInfoEncryptor`/`DeviceInfoDecryptor`** (RSA-OAEP JWE round-trip with a decrypt session), **`DeviceFieldEncryptor`** (mirror of existing field decrypt), and **`DeviceInfoUpdater`** orchestrating the PATCH. **`DeviceV2`** gains an `info` field; **`SyncJweCrypto`** adds **`rsaSpkiFromJwkComponents`** for JWK→SPKI encryption. **`SyncGzipInterceptor`** skips gzip on PATCH `/sync/devices` until the backend accepts it.
> 
> **Sync Dev Settings** exposes rename (dialog → PATCH) and fetch & decrypt devices (legacy vs `device_info` side by side). Unit tests cover the new encrypt/decrypt/update and remote PATCH behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71bcd4a892cb35ac3b4438a7e7bffd853d29f83d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/DeviceFieldEncryptor.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/DeviceFieldEncryptor.kt
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.sync.impl
+
+import androidx.annotation.WorkerThread
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.sync.crypto.SyncLib
+import com.duckduckgo.sync.impl.Result.Error
+import com.duckduckgo.sync.impl.Result.Success
+import com.duckduckgo.sync.impl.crypto.SyncJweCrypto
+import com.duckduckgo.sync.store.SyncStore
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import javax.inject.Inject
+
+/**
+ * Encrypts the legacy `name`/`type` device fields for the current device's credential — the inverse
+ * of [DeviceFieldDecryptor], so the values stay readable by clients that don't understand
+ * `device_info` yet. 3party encrypts under a main encryption key derived from the scoped password;
+ * ddg/null encrypts with the local primary key.
+ */
+@WorkerThread
+interface DeviceFieldEncryptor {
+    fun encrypt(name: String, type: String): Result<EncryptedDeviceFields>
+}
+
+data class EncryptedDeviceFields(
+    val name: String,
+    val type: String,
+)
+
+@SingleInstanceIn(AppScope::class)
+@ContributesBinding(AppScope::class)
+class RealDeviceFieldEncryptor @Inject constructor(
+    private val syncStore: SyncStore,
+    private val syncJweCrypto: SyncJweCrypto,
+    private val nativeLib: SyncLib,
+) : DeviceFieldEncryptor {
+
+    override fun encrypt(name: String, type: String): Result<EncryptedDeviceFields> {
+        return when (syncStore.credentialId) {
+            CREDENTIAL_ID_3PARTY -> encryptThirdParty(name, type)
+            CREDENTIAL_ID_DDG, null -> encryptDdg(name, type)
+            else -> Error(reason = "DeviceFieldEncryptor: unknown credential_id=${syncStore.credentialId}")
+        }
+    }
+
+    private fun encryptDdg(name: String, type: String): Result<EncryptedDeviceFields> {
+        val primaryKey = syncStore.primaryKey?.takeUnless { it.isEmpty() }
+            ?: return Error(reason = "DeviceFieldEncryptor: primaryKey missing")
+        return runCatching {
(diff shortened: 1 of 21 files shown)


CDRussell:
* **#9380** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9380?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9369** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9360** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9360?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/engine/SyncGzipInterceptor.kt
```diff
@@ -54,8 +54,11 @@ class SyncGzipInterceptor @Inject constructor(
             return chain.proceed(chain.request())
         }
 
+        // The /sync/devices endpoint rejects gzip bodies as invalid_json, so skip compression here until the backend supports it.
+        val isDevicesEndpoint = chain.request().url.encodedPath.endsWith("/sync/devices")
+
         // check if it's http operation is PATCH
-        if (chain.request().method == "PATCH" && syncFeature.gzipPatchRequests().isEnabled()) {
+        if (chain.request().method == "PATCH" && !isDevicesEndpoint && syncFeature.gzipPatchRequests().isEnabled()) {
```
CDRussell, line 61: This is temporary; should be able to remove this soon as backend will soon support it

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/DeviceInfoDecryptor.kt
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * The private key is fetched and unwrapped once per [openSession];
+ * reuse the returned [Session] to decrypt every blob from a device list off that single fetch.
+ */
+@WorkerThread
+interface DeviceInfoDecryptor {
+    fun openSession(): Result<Session>
+
+    /** Decrypts `device_info` blobs with a private key held only for this session's lifetime. */
+    interface Session {
```
CDRussell, line 39: Adding `Session` as a performance improvement instead of doing a `GET` network call for every device, instead it can be done once and that will last to get all the devices decrypted.

A better solution is to cache this so a network call isn't needed at all. Will confirm this is possible and do that in a later branch if so.

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/DeviceFieldEncryptor.kt
```diff
@@ -0,0 +1,91 @@
...
+interface DeviceFieldEncryptor {
+    fun encrypt(name: String, type: String): Result<EncryptedDeviceFields>
+}
+
+data class EncryptedDeviceFields(
+    val name: String,
+    val type: String,
+)
+
+@SingleInstanceIn(AppScope::class)
```
MiSikora: I've noticed some stateless classes in this PR are marked with `@SingleInstanceIn` while others aren't, even though there isn't an obvious cost to creating them. This class is one example, and `RealDeviceInfoDecryptor` is another that goes the opposite way. Is there a convention or reasoning we're following for when to apply `@SingleInstanceIn`?

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/DeviceInfoDecryptor.kt
MiSikora, line 39: Shouldn't this be also marked as `@WorkerThread`? I don't think that the outer class annotation is inherited.

MiSikora (approved):

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/DeviceFieldEncryptor.kt
CDRussell: Good observation on the inconsistency. 

Looking again, it's being applied here when it isn't needed, and in a few other of these related classes too. Unless there's shared state in the class, it's guarding access around a shared resource, or there is a real cost for construction, it's probably not needing to be a singleton.

Will remove it from a few of these classes.

CDRussell:
### Merge activity

* **Aug 4, 9:04 AM UTC**: @CDRussell merged this pull request with [Graphite](https://app.graphite.com/github/pr/duckduckgo/Android/9369).

PR #9389, opened 2026-08-04
[Config-driven linear onboarding dialogs] Extract intro animators

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1216987945331691?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/481882893211075/task/1216854264994244?focus=true

### Description

Ports the welcome intro/outro choreography from `BrandDesignUpdateWelcomePage` to the config-driven page, behind the same remote toggle as the rest of the config-driven work.

Two pieces:

1. **`OnboardingIntroChoreographer`** owns the intro views and everything that happens to them: play, snap to end state, fade out for the first dialog, tear down. The animators are a straight port, same durations, interpolators and Lottie frames.
2. **`ViewState.screen`** replaces the flat `stepId` / `config` / `animateEntry` triple with a sealed `Screen` (`Intro.Play`, `Intro.Restore`, `Dialog`, `None`). The view model states which screen is current instead of the view inferring it from flags.

#### What changed vs legacy

- **Rotating mid-intro keeps the first dialog's background cross-fade.** Legacy snapped it after a mid-intro recreation. Bug fix.
_before -> after (ignore different step used, important is the transition)_

transition_after_rotation

- **Any first dialog fades the intro out and no longer waits for the background transition to end.** Legacy faded only into the welcome dialog and snapped the intro away for every other dialog. Now, no step order is assumed. Welcome steps' binders will come with a card reveal delay configured instead.
- **The tablet `enterStartX` override is dropped.** Legacy shortened the entering background's travel on tablets for the first dialog. The entering view is transparent for most of its travel, so the difference is not visible. Verified on a Galaxy Tab S9+.
_before -> after_ (ignore different step used, important is the **intro background** transition)

tablet_intro_background_fade

### Steps to test this PR

_Fresh onboarding_
- [ ] Apply below patch:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
index b6a1e4d352..41a8d85ad4 100644
--- a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
@@ -58,6 +58,6 @@ interface OnboardingBrandDesignUpdateToggles {
     /**
      * Selects the config-driven renderer for the brand-design onboarding dialogs.
      */
-    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
     fun configDrivenDialogs(): Toggle
 }

```
- [ ] Clear app data, launch.
- [ ] Logo drops, background slides up, title fades and scales in.
- [ ] First dialog arrives: logo and title fade out while the background cross-fades to the dialog backdrop.

_duck.ai variant_
- [ ] Additionally apply below patch:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt b/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
index df50491818..cf6a6196ba 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/orchestrator/NewUserOnboardingPlanProvider.kt
@@ -101,7 +101,7 @@ class NewUserOnboardingPlanProvider @Inject constructor(
         onCompleted: suspend () -> Unit,
         onSkipped: suspend () -> Unit,
     ): LinearOnboardingPlan =
-        if (customAiOnboardingResolver.resolve()) {
+        if (true) {
             // in custom AI onboarding path, the input toggle is enabled by default
             duckChat.setCosmeticInputScreenUserSetting(enabled = true)
             onboardingStore.storeInputScreenSelection(selected = true)

```
- [ ] Enable the custom duck.ai onboarding flow, clear data, launch.
- [ ] The duck.ai lockup plays after the title settles, in the correct font and text colour, in both light and dark mode.
- [ ] First dialog arrives: the lockup fades out along with the logo and title.

_Rotation_
- [ ] Rotate mid-intro. Intro finishes and transitions to the dialog without any artifacts.
- [ ] Rotate after intro settles. Intro finishes and transitions to the dialog without any artifacts.
- [ ] Rotate on the first dialog. Dialog redraws with no entry animation and no intro visuals reappearing.

---

> [!NOTE]
> **Medium Risk**
> Touches first-run onboarding UI and orchestrator handoff timing; mistakes could skip steps or show broken transitions, but changes are localized and covered by new ViewModel/state tests.
> 
> **Overview**
> Ports the brand-design welcome **logo/title/background intro** from the legacy page into **config-driven onboarding**, gated behind the same remote toggle as the rest of that work.
> 
> **View model** replaces flat `stepId` / `config` / `animateEntry` with a sealed **`Screen`** (`Intro.Play`, `Intro.Restore`, `Dialog`, `None`). The intro step is handled explicitly (play vs restore after rotation), `IntroAnimationFinished` is emitted when the choreographer completes, and steps with no UI (system prompts) keep the last screen or publish `None` instead of auto-skipping the intro.
> 
> **Fragment** wires **`OnboardingIntroChoreographer`** for play/restore/outro and hands `backgroundPrimary` to **`DialogRenderEngine`** on the first dialog via `clearForDialog()`, using a new **`animateBackground`** flag so the backdrop can cross-fade from the intro when appropriate.
> 
> **Behavior vs legacy:** first dialog always fades intro visuals (not only welcome); rotation mid-intro preserves background cross-fade; optional duck.ai Lottie after the title uses theme text color and Duck Sans.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 620a168fe3465c57f515668654c81cf9b9171905. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenOnboardingPageViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenOnboardingPageViewModel.kt
@@ -128,8 +128,6 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
     /** Last step id a [DialogConfig] was published for; drives the [ViewState.animateEntry] policy. */
     private var lastPresentedStepId: LinearOnboardingStepId? = null
 
-    private var introStarted = false
-
     private var notificationPermissionFlowStarted = false
 
     private var addWidgetPromptFlowStarted = false
@@ -156,7 +154,6 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
     }
 
     fun onIntroAnimationStarted() {
-        introStarted = true
         _viewState.update { state ->
             val screen = state.screen
             if (screen is Screen.Intro.Play) {
@@ -274,7 +271,7 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
         if (dialog is NewUserOnboardingActivityDialog.IntroAnimation) {
             _viewState.update {
                 it.copy(
-                    screen = if (introStarted) {
+                    screen = if (it.screen is Screen.Intro.Restore) {
                         Screen.Intro.Restore(withDuckAi = dialog.withDuckAi)
                     } else {
                         Screen.Intro.Play(withDuckAi = dialog.withDuckAi)
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenOnboardingPageViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenOnboardingPageViewModel.kt
@@ -75,10 +75,35 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
 ) : ViewModel() {
 
     data class ViewState(
-        val stepId: LinearOnboardingStepId? = null,
-        val config: DialogConfig? = null,
-        val animateEntry: Boolean = true,
+        val screen: Screen? = null,
     )
+
+    sealed interface Screen {
+
+        sealed interface Intro : Screen {
+            val withDuckAi: Boolean
+
+            data class Play(override val withDuckAi: Boolean) : Intro
+
+            /**
+             * A view already started the intro.
+             * Ignore this state if the view played the intro, snap to end state if the view hasn't (for example, when recreated).
+             */
+            data class Restore(override val withDuckAi: Boolean) : Intro
+        }
+
+        data class Dialog(
+            val stepId: LinearOnboardingStepId,
+            val config: DialogConfig,
+            val animateEntry: Boolean,
+        ) : Screen
+
+        /**
+         * The flow is on a step this view draws nothing for and there is no earlier screen to keep showing, which
+         * only happens on a fresh view model, for example after a mid-flow re-entry.
+         */
+        data object None : Screen
+    }
 
     sealed interface Command {
         data object RequestNotificationPermissions : Command
@@ -103,6 +128,8 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
     /** Last step id a [DialogConfig] was published for; drives the [ViewState.animateEntry] policy. */
     private var lastPresentedStepId: LinearOnboardingStepId? = null
 
+    private var introStarted = false
+
     private var notificationPermissionFlowStarted = false
 
     private var addWidgetPromptFlowStarted = false
@@ -116,12 +143,31 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
     fun onContentInteraction(interaction: ContentInteraction) = Unit // No-op until dialogs with local state are implemented in follow-ups.
 
     fun onDialogRendered(stepId: LinearOnboardingStepId) {
-        _viewState.update {
+        _viewState.update { state ->
+            val screen = state.screen
             // The dialog for this step has been rendered.
             // Disable entry animation for potential re-draws (like config change/rotation).
-            if (it.stepId == stepId) it.copy(animateEntry = false) else it
+            if (screen is Screen.Dialog && screen.stepId == stepId) {
+                state.copy(screen = screen.copy(animateEntry = false))
+            } else {
+                state
+            }
         }
     }
+
+    fun onIntroAnimationStarted() {
+        introStarted = true
+        _viewState.update { state ->
+            val screen = state.screen
+            if (screen is Screen.Intro.Play) {
+                state.copy(screen = Screen.Intro.Restore(withDuckAi = screen.withDuckAi))
+            } else {
+                state
+            }
+        }
+    }
+
+    fun onIntroAnimationFinished() = emit(NewUserOnboardingEvent.IntroAnimationFinished)
 
     fun onResume() {
         checkAddWidgetPromptResult()
@@ -224,18 +270,37 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
         state: LinearOnboardingState.InProgress,
     ) {
         val dialog = step.resolveDialog()
+
+        if (dialog is NewUserOnboardingActivityDialog.IntroAnimation) {
+            _viewState.update {
+                it.copy(
+                    screen = if (introStarted) {
+                        Screen.Intro.Restore(withDuckAi = dialog.withDuckAi)
+                    } else {
+                        Screen.Intro.Play(withDuckAi = dialog.withDuckAi)
+                    },
+                )
+            }
+            return
+        }
+
         val config = dialogConfigResolver.resolve(dialog, customAiOnboardingStore.isEnabled())
         if (config != null) {
             _viewState.update {
                 it.copy(
-                    stepId = step.id,
-                    config = config.copy(stepIndicator = state.stepIndicatorProgress()),
-                    animateEntry = step.id != lastPresentedStepId,
+                    screen = Screen.Dialog(
+                        stepId = step.id,
+                        config = config.copy(stepIndicator = state.stepIndicatorProgress()),
+                        animateEntry = step.id != lastPresentedStepId,
+                    ),
                 )
             }
             lastPresentedStepId = step.id
             emit(NewUserOnboardingEvent.Presented)
         } else {
+            // If there's no new dialog to draw (e.g. we're displaying a system prompt), keep the current one.
+            // None only when there was never anything to keep.
+            _viewState.update { state -> if (state.screen == null) state.copy(screen = Screen.None) else state }
             advancePastUnrenderedDialog(dialog)
         }
     }
@@ -248,8 +313,6 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
      */
     private suspend fun advancePastUnrenderedDialog(dialog: NewUserOnboardingActivityDialog) {
         when (dialog) {
-            is NewUserOnboardingActivityDialog.IntroAnimation -> emit(NewUserOnboardingEvent.IntroAnimationFinished)
-
             NewUserOnboardingActivityDialog.NotificationPermission -> {
                 if (!notificationPermissionFlowStarted) {
                     notificationPermissionFlowStarted = true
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePageFragment.kt
```diff
@@ -187,41 +198,52 @@ class ConfigDrivenWelcomePageFragment : OnboardingPageFragment(R.layout.content_
...
+        // Once the first dialog arrives, if:
+        // - intro visuals on screen: clear them and the background cross-fades from them
+        // - no intro visual on screen (like a mid-flow re-entry from another activity): nothing to animate from, snap new background
+        // Later renders always animate the background.
+        val animateBackground = if (hasRenderedOnce) {
+            screen.animateEntry
+        } else {
+            hasRenderedOnce = true
+            (intro?.clearForFirstDialog() ?: false) && screen.animateEntry
+        }
```
mikescamell: 🤖 nit: `backgroundPrimary` has two writers now. The choreographer sets `translationY`, scale, `minFrame` and `progress` on it and plays its Lottie, and `OnboardingBackgroundAnimator` owns the same view via `BackgroundController`.

This is correct today, but only because `clearForFirstDialog()` is the left operand of the `&&`. Reorder it to `screen.animateEntry && intro?.clearForFirstDialog() == true` and a first render with `animateEntry = false` never settles the intro. `BackgroundController.cancel()` only cancels the enter/exit set and the `.animate()` animators, so `backgroundIntroAnimatorSet` would keep sliding the background up under the dialog's cross-fade.

Suggest the choreographer answers the question rather than the fragment ordering the calls:

```kotlin
private var firstDialogHandled = false

/**
 * Clears the intro for an arriving dialog. Only the first call clears anything.
 *
 * @return true when the dialog's background can cross-fade from what is on screen
 */
fun clearForDialog(): Boolean {
    if (firstDialogHandled) return true
    firstDialogHandled = true
    if (visualsOnScreen) {
        settleRunningIntro()
        playOutro()
    } else {
        snapToOutroEndState()
    }
    return visualsOnScreen
}
```

`render()` then becomes `val animateBackground = intro?.clearForDialog() != false && screen.animateEntry`, and `hasRenderedOnce` plus its `onDestroyView` reset go away. Truth table matches current behaviour.

If you'd rather not move it, the class KDoc stating the handover (choreographer owns `backgroundPrimary` until the first dialog, `BackgroundController` after) would cover the next reader.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/OnboardingIntroChoreographer.kt
```diff
@@ -0,0 +1,423 @@
...
+        return AnimatorSet().apply {
+            playTogether(animators)
+        }
+    }
+
+    fun play(withDuckAi: Boolean, onFinished: () -> Unit) {
+        visualsOnScreen = true
+        // The animators measure the title's laid-out width and the screen height.
+        binding.root.doOnLayout {
+            if (!released) startIntroAnimation(withDuckAi, onFinished)
```
mikescamell: 🤖 nit: the deferred start consults only one of the two flags that mean "don't start".

```suggestion
            if (!released && !cleared) startIntroAnimation(withDuckAi, onFinished)
```

No behaviour change that I can find. `Screen.None` can't overwrite `Intro.Play`, and a `Dialog` needs `IntroAnimationFinished` first, so neither clear path is reachable before layout. Just keeps the guard aligned with the class's state.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/OnboardingIntroChoreographer.kt
```diff
@@ -0,0 +1,423 @@
...
+import com.airbnb.lottie.LottieProperty
+import com.airbnb.lottie.model.KeyPath
+import com.duckduckgo.app.browser.databinding.ContentOnboardingWelcomePageUpdateBinding
+import com.duckduckgo.fonts.R as FontsR
+import com.duckduckgo.mobile.android.R as CommonR
+
+/**
+ * One-time welcome intro/outro choreography for [ConfigDrivenWelcomePageFragment].
+ */
+class OnboardingIntroChoreographer(
```
mikescamell, line 46: 🤖 The flag contract here is what the fragment's `animateBackground` decision reads, and none of it is covered. The orderings are the interesting part rather than the animators:

- `restore()` returns `false` while this view is playing, `true` when it isn't
- `clearForFirstDialog()` is idempotent and reports whether there was anything to fade
- `dismissUnplayed()` after a clear is a no-op
- `release()` mid-play stops the deferred start

Those four are what a rotation lands on, and they're cheap to pin down compared to the animator construction, which I'd leave alone.

One thought while you're in here: every other collaborator on this branch is `interface` plus `Impl(binding)`, which is what lets `DialogRenderEngineTest` assert sequencing against fakes. An `OnboardingIntroChoreographer` interface would do the same for these four methods and give the fragment a seam. Not asking for it in this PR, but it might be the cheaper route to the tests than working around the binding.

mikescamell:
@LukasPaczos Looking good! 👨‍🍳 💋 Some Claude review comments that I thought might be worth a quick look, but nothing major.

1. Do we want to take the opportunity to snap the welcome screen animation on rotation rather than replaying it constantly? Or maybe it's a good follow up task? WDYT?

2. I managed to get into a state where the welcome screen did not progress after doing multiple rotations via tablet emulator. Very edgey.

https://github.com/user-attachments/assets/d8c46126-5fde-47a4-aec6-c6e6eb070f05

3. I managed to get into some odd behaviour rotating on the Samsung S25 device while rotating, the flash came back 🙈 I got some long held white screen, then the welcome animation snapped in, then rotating in what seemingly was a transisition snapped the browser into place before it had place.

https://github.com/user-attachments/assets/c4718f79-423b-46c6-af9a-4c615d1c9dc3

Maybe we add these as follow ups as I don't think anyone is rotating like a madman in onboarding 👍

LukasPaczos:
1. Done in https://github.com/duckduckgo/Android/pull/9389/commits/4ff208478f79fff63d71aae3d66ede5e33ea212e.
2. I can't reproduce but there was a view visibility check guarding flow progression that was carried over all the way back from the [legacy-legacy WelcomePage](https://github.com/duckduckgo/Android/blob/2326d387865dcefee26f947cdd9b52424c2b2743/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/WelcomePage.kt#L122-L124). In there, it used to guard genuine animation code, but now it was only guarding state progression, so I removed it.
3. Ah, I forgot to mention in the original task, but it's honestly best to test the intro animation in a release build. In debug, launch is so slow that I can see it impacting intro animations _a lot_. If you add screen rotations to it, jank can be pretty bad. Could you give it another pass in release flavor and see how it feels?

Ready for another round!

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenOnboardingPageViewModel.kt
```diff
@@ -103,6 +128,8 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
     /** Last step id a [DialogConfig] was published for; drives the [ViewState.animateEntry] policy. */
     private var lastPresentedStepId: LinearOnboardingStepId? = null
 
+    private var introStarted = false
+
```
mikescamell: 🤖 `introStarted` duplicates a fact `Screen.Intro.Restore` already encodes. `onIntroAnimationStarted()` sets both, and `applyStep` reads only the boolean. They cannot disagree today, but nothing enforces that. Deriving it from the state removes the second source of truth.

The path this guards is a `Stay` transition re-emitting the intro step. Screen is already `Restore` by then, so the derived form gives the same answer and still will not replay. Both tests that cover it call `onIntroAnimationStarted()` before the re-emission, so they hold as written.

This comment plus the two below are one change.

```suggestion
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenOnboardingPageViewModel.kt
```diff
@@ -116,13 +143,32 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
...
+        introStarted = true
+        _viewState.update { state ->
+            val screen = state.screen
+            if (screen is Screen.Intro.Play) {
+                state.copy(screen = Screen.Intro.Restore(withDuckAi = screen.withDuckAi))
+            } else {
+                state
+            }
         }
     }
```
mikescamell, line 165: 🤖 Keep the `Play` to `Restore` flip, drop the assignment.

```suggestion
    fun onIntroAnimationStarted() {
        _viewState.update { state ->
            val screen = state.screen
            if (screen is Screen.Intro.Play) {
                state.copy(screen = Screen.Intro.Restore(withDuckAi = screen.withDuckAi))
            } else {
                state
            }
        }
    }
```

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenOnboardingPageViewModel.kt
```diff
@@ -224,18 +270,37 @@ class ConfigDrivenOnboardingPageViewModel @Inject constructor(
...
+        if (dialog is NewUserOnboardingActivityDialog.IntroAnimation) {
+            _viewState.update {
+                it.copy(
+                    screen = if (introStarted) {
+                        Screen.Intro.Restore(withDuckAi = dialog.withDuckAi)
+                    } else {
+                        Screen.Intro.Play(withDuckAi = dialog.withDuckAi)
+                    },
+                )
+            }
```
mikescamell, line 280: 🤖 Read it off the state being updated rather than the field.

```suggestion
            _viewState.update {
                it.copy(
                    screen = if (it.screen is Screen.Intro.Restore) {
                        Screen.Intro.Restore(withDuckAi = dialog.withDuckAi)
                    } else {
                        Screen.Intro.Play(withDuckAi = dialog.withDuckAi)
                    },
                )
            }
```

mikescamell (approved):
@LukasPaczos LGTM, tested again on Samsung S25 and Tablet Emulator. Couple of optional suggestions 👍

PR #9403, opened 2026-08-05
[Config-driven linear onboarding dialogs] Port existing screens (forward flows)

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1217183142109149?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/481882893211075/task/1216854264994244

### Description

Step 3 of the tech design: ports the remaining screens to the config-driven onboarding renderer. This PR covers below binders:
- Welcome
- Input selection
- Input preview

**Divergences from the legacy implementation** (`BrandDesignUpdateWelcomePage.kt`):
- The input screen's "with AI" Lottie flourish always starts after the 2s initial delay. Legacy's snapped render (rotation, process restore) started the loop immediately. This fixes an unintentional divergence.
- The input preview's card resizes (mode switch, suggestion reveal) go through the render engine. Same visual, but the tween is suppressed while a snapped or skipped entrance is still landing the screen, where legacy always ran it:

input_preview_snap

**Divergences from the tech design:**
- Shown pixels are a 1:1 port used only by the new view model, not the shared utility the TD proposed for both arms. Parity is guarded by the exhaustive mapping and tests instead; the legacy path is untouched, which keeps this PR flag-off safe. We can look into the alternative approach of moving these pixels into plan provider in a follow up, since we'd need to make changes to both old and new code paths.
- `BindScope` gained `animateCardBounds`, which the TD didn't have. A binder only reaches its own layout, so it had no way to animate a card resize, if one was needed. The binder requests the tween but the engine owns it, so the request is safely ignored during snapped transitions.
- `DialogConfig` gained a `CardEntry` param (`Immediate` / `AfterBackgroundTransition`). The welcome dialogs arrive while the intro background is still crossing over, and without it the card faded in over the moving background. The timing precisely matches legacy.

### Steps to test this PR
- [x] Apply below patch:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
index b6a1e4d352..41a8d85ad4 100644
--- a/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboardingbranddesignupdate/OnboardingBrandDesignUpdateToggles.kt
@@ -58,6 +58,6 @@ interface OnboardingBrandDesignUpdateToggles {
     /**
      * Selects the config-driven renderer for the brand-design onboarding dialogs.
      */
-    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
     fun configDrivenDialogs(): Toggle
 }

```
_base flow_
- [x] Clean install.
- [x] Walk the full flow: welcome, comparison chart, address bar, input screen, input preview, — each renders with its background, Dax embellishment, card arrow and CTAs.
- [x] Rotate on each screen — the snapped re-render keeps in-progress state (address bar selection, switches, input mode).
- [x] Tap during a transition to skip animations.
- [x] Input preview: switch the search/AI tabs (card resizes smoothly), submit a typed query and a suggestion.

_sync restore flow_
- [x] Finish onboarding, go to settings, and enable sync on the device.
- [x] Uninstall the app (do not clear caches).
- [ ] Reinstall the app.
- [ ] Verify restore content is displayed on the welcome step.

_Custom AI flow_
- [x] Additionally apply this diff:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt b/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt
index fe3752cbb2..af220bb09d 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/CustomAiOnboardingStore.kt
@@ -100,6 +100,7 @@ class CustomAiOnboardingStoreImpl @Inject constructor(
     }
 
     override suspend fun resolve() = resolveMutex.withLock {
+        return@withLock true
         withContext(dispatcherProvider.io()) {
             // Ensure the install referrer (and therefore the processing function) has resolved before reading.
             withTimeoutOrNull(MAX_REFERRER_WAIT_TIME_MS) { referrerStateListener.get().waitForReferrerCode() }
@@ -116,6 +117,7 @@ class CustomAiOnboardingStoreImpl @Inject constructor(
     }
 
     override suspend fun isEnabled(): Boolean = resolveMutex.withLock {
+        return@withLock true
         withContext(dispatcherProvider.io()) {
             preferences.getBoolean(PREFS_KEY_ENABLED, false)
         }

```

- [x] Clean install.
- [x] Walk the full flow: welcome, AI comparison, input preview, Duck.ai demo, browser comparison, address bar

---

> [!NOTE]
> **Medium Risk**
> Touches shared decoration fit correction and onboarding layout/keyboard behavior across many screens; regressions could show as clipped cards, hidden Dax art, or wrong sync-restore flow, though coverage is heavy in unit tests.
> 
> **Overview**
> Ports **welcome**, **sync restore**, **input selection**, and **input preview** onto the config-driven onboarding renderer, with new binders, `DialogConfigResolver` mappings, and `ContentConfig` types wired through `ContentController` and `DialogRenderEngine`.
> 
> **Layout and decorations:** `OnboardingDecorationFitCorrector` gains `reservesInsetAboveDecoration` and revised inset/clamp math so cards above decorations can clear the keyboard without shrinking decorations; the embellishment controller enables this for the config-driven path. The welcome fragment applies layout overrides (drops default `constrainedHeight`, wing `adjustViewBounds`) and keeps bottom inset when focus is inside the card during IME.
> 
> **Render engine UX:** `CardEntry.AfterBackgroundTransition` delays card fade-in after welcome backgrounds; `BindScope.animateCardBounds` tweens card resizes (mode switch, suggestions) with suppression during snapped/skipped entrances. Intro handover always clears choreographer visuals on every dialog render; shown pixels fire via exhaustive `OnboardingDialogShownPixels`.
> 
> **View model:** Config-driven path renders sync restore instead of skipping; input preview submits forward `InputDemoQuerySubmitted`; several orchestrator dialogs no longer auto-continue at activity level (handled by config UI).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69e3560cdea1f9e33b592689b1965ab9159d966d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
@@ -107,11 +107,11 @@ class OnboardingDecorationFitCorrector(
         val overflow = (cardContainerHeight - viewportHeight).coerceAtLeast(0)
         val available = root.height - root.paddingTop - root.paddingBottom
         val dialogParams = dialog.layoutParams as ViewGroup.MarginLayoutParams
-        // A keyboard overlays the window rather than shrinking it, so `available` already covers the room the card
-        // gave up above the decoration. Counting it would shrink the decoration, which grows the inset, which
-        // shrinks the decoration again, until it is gone.
-        val cardBottomMargin = if (reservesInsetAboveDecoration) 0 else dialogParams.bottomMargin
-        val dialogSpace = dialogHeight + overflow + dialogParams.topMargin + cardBottomMargin
+        // The card's bottom margin never counts against the decoration's room. With a decoration shown it is
+        // either zero (synced above) or a reserved inset, and the keyboard that inset clears overlays the
+        // window rather than shrinking it, so `available` already covers the room the card gave up. Counting
+        // it would shrink the decoration, which grows the inset, which shrinks it again, until it is gone.
+        val dialogSpace = dialogHeight + overflow + dialogParams.topMargin
         val decorationParams = deco.layoutParams as ViewGroup.MarginLayoutParams
 
         val target = BrandDesignUpdateOnboardingLayoutHelper.computeDecorationHeight(
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
@@ -83,10 +83,6 @@ class OnboardingDecorationFitCorrector(
 
         val deco = decoration
         val decorationShown = deco != null && !deco.isGone
-        // The card reserves the bottom-bar inset when it is the bottom-most element: bottom-anchored AND no
-        // decoration shown below it. A shown decoration is at least its min height, which exceeds any bar
-        // inset, so it covers the bar for the card; reserving the inset then would feed dialogSpace and hide
-        // that very decoration.
         if (syncCardBottomInset(decorationShown)) return false
 
         // While onboardingImprovementsV2 is off the corrector stays inert: syncCardBottomInset above
@@ -155,7 +151,12 @@ class OnboardingDecorationFitCorrector(
     ): Int {
         if (!enabled) return 0
         if (params.bottomToBottom == ConstraintLayout.LayoutParams.PARENT_ID) {
-            return if (decorationShown) 0 else cardBottomInsetPx()
+            // On the legacy path a shown decoration exceeds any bar inset and covers it for the card;
+            // reserving the inset then would feed dialogSpace and hide that very decoration. On the
+            // config-driven path a decoration that leaves the card bottom-anchored sits beside or over
+            // it, covering nothing below it, so the card still has to clear the inset — the keyboard.
+            if (decorationShown && !reservesInsetAboveDecoration) return 0
+            return cardBottomInsetPx()
         }
         val deco = decoration?.takeIf { decorationShown && reservesInsetAboveDecoration } ?: return 0
         return (cardBottomInsetPx() - roomBelowCard(deco)).coerceAtLeast(0)
--- a/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
@@ -40,6 +40,9 @@ class OnboardingDecorationFitCorrector(
      * does not already cover. Only the config-driven renderer sets this: its undecorated band anchors the card on
      * screens the legacy renderer pinned to the parent bottom, so without this the card has no way to clear an
      * inset taller than the band, i.e. the keyboard.
+     *
+     * The whole reservation hangs off this flag, so the legacy renderer's geometry is untouched: the reserved
+     * inset stays out of the decoration's fit, and the card's span above the decoration is what clamps it.
      */
     var reservesInsetAboveDecoration: Boolean = false
 
@@ -93,9 +96,10 @@ class OnboardingDecorationFitCorrector(
         // has already reverted any reserved inset, and the shrink/hide logic below must not run.
         if (!enabled) return true
 
-        if (deco == null) return !syncBottomAnchoredClamp()
+        if (deco == null) return !syncCardClamp()
         if (deco.isGone) return true
         if (BrandDesignUpdateOnboardingLayoutHelper.isInScrollableContainer(dialog, root)) return true
+        if (reservesInsetAboveDecoration && syncCardClamp()) return false
 
         val viewport = cardContainer.parent as? View ?: return true
 
@@ -110,7 +114,12 @@ class OnboardingDecorationFitCorrector(
         val overflow = (cardContainerHeight - viewportHeight).coerceAtLeast(0)
         val available = root.height - root.paddingTop - root.paddingBottom
         val dialogParams = dialog.layoutParams as ViewGroup.MarginLayoutParams
-        val dialogSpace = dialogHeight + overflow + dialogParams.topMargin + dialogParams.bottomMargin
+        // An inset reserved above the decoration buys room the decoration is not competing for: a keyboard
+        // overlays the window rather than shrinking it, so `available` already covers the room the card gave up.
+        // Counting it would shrink the decoration, which grows the inset, which shrinks the decoration, until it
+        // is gone. Without that reservation the margin is 0 here anyway, a decoration being shown.
+        val cardBottomMargin = if (reservesInsetAboveDecoration) 0 else dialogParams.bottomMargin
+        val dialogSpace = dialogHeight + overflow + dialogParams.topMargin + cardBottomMargin
         val decorationParams = deco.layoutParams as ViewGroup.MarginLayoutParams
 
         val target = BrandDesignUpdateOnboardingLayoutHelper.computeDecorationHeight(
@@ -164,11 +173,19 @@ class OnboardingDecorationFitCorrector(
     }
 
     // Clamp only on overflow: constrainedHeight rounds a fitting wrap down a pixel → a permanent scrollbar.
-    private fun syncBottomAnchoredClamp(): Boolean {
+    private fun syncCardClamp(): Boolean {
         val params = dialog.layoutParams as? ConstraintLayout.LayoutParams ?: return false
-        if (params.bottomToBottom != ConstraintLayout.LayoutParams.PARENT_ID) return false
+        val roomBelowCard = when {
+            params.bottomToBottom == ConstraintLayout.LayoutParams.PARENT_ID -> 0
+            // Anchored above a decoration: unclamped the card overruns it instead of scrolling, so a card lifted
+            // clear of the keyboard would still spill under it.
+            params.bottomToTop != ConstraintLayout.LayoutParams.UNSET ->
+                decoration?.takeIf { !it.isGone }?.let { roomBelowCard(it) } ?: return false
+            else -> return false
+        }
         val content = cardContainer.measuredHeight.takeIf { it > 0 } ?: return false
-        val span = root.height - root.paddingTop - root.paddingBottom - params.topMargin - params.bottomMargin
+        val span = root.height - root.paddingTop - root.paddingBottom -
+            params.topMargin - params.bottomMargin - roomBelowCard
         val shouldClamp = dialog.paddingTop + content > span
         if (params.constrainedHeight == shouldClamp) return false
         dialog.updateLayoutParams<ConstraintLayout.LayoutParams> { constrainedHeight = shouldClamp }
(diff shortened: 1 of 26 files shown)


mikescamell:
@LukasPaczos Noticed something on with the dax walking animation on the welcome screen where he slides in rather than walking. Almost looks like it's not playing the full animation?

https://github.com/user-attachments/assets/20dc2ac1-1915-403a-8214-c81cc35ff0a1

mikescamell:
@LukasPaczos We also have a sliding bobbing Dax on resume 😁 

https://github.com/user-attachments/assets/1a53b42a-5427-4458-b522-a915781625e4

LukasPaczos:
> dax walking animation on the welcome screen where he slides in rather than walking

Fixed!

> We also have a sliding bobbing Dax on resume

Argh, I over-optimized in the previous PR. There was a change to snap background and fade in only embellishment on fragment recreation. I think it generally looks better, but I didn't consider that some embellishments _require_ the background to move as well to make sense. Instead of adding conditions, I just dropped that change and now the background always animates, same as prod.

Additionally, I pushed changes to:
1. Fix issues reported in https://github.com/duckduckgo/Android/pull/9404#issuecomment-5215195269, they were not specific to that following PR and it'd be better to test them for all dialogs here as well.
2. Fix issue with scrolling of the Input Preview step when keyboard is visible, now behaves exactly like prod.
3. Make the Welcome card on tablets float in the middle of the screen instead of being anchored to Dax, now matches prod.

Ready for another round!

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/engine/DialogRenderEngine.kt
```diff
@@ -101,8 +114,15 @@ class DialogRenderEngine(
         val settledDecoration = embellishments.transition(previous?.embellishment, config.embellishment, animate)
         cardAnchor.apply(settledDecoration)
 
+        // A snapped background is already in place, so there is nothing for the card to wait on.
+        val revealDelayMs = if (config.cardEntry == CardEntry.AfterBackgroundTransition && animate) {
```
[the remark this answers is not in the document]
LukasPaczos: Background now genuinely animates so the card waiting on it is the contract, rotation is unaffected (`onDialogRendered` pins `animateEntry = false`).

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
```diff
@@ -130,19 +143,44 @@ class OnboardingDecorationFitCorrector(
...
         return true
     }
 
+    private fun desiredCardBottomInset(
+        params: ConstraintLayout.LayoutParams,
+        decorationShown: Boolean,
+    ): Int {
+        if (!enabled) return 0
+        if (params.bottomToBottom == ConstraintLayout.LayoutParams.PARENT_ID) {
+            return if (decorationShown) 0 else cardBottomInsetPx()
```
mikescamell, line 158: 🤖 This returns 0 whenever a decoration is shown, but the reasoning above only holds for a bar inset. Since the IME backport `cardBottomInsetPx()` can carry the keyboard, and a decoration that does not anchor the card (`LeftWing`, `BobbingDax`) leaves the card at `PARENT_ID` while sitting beside it, covering nothing below it. The card then sits under the keyboard.

Proof: with `reservesInsetAboveDecoration = true`, a 422px decoration shown, the card at `PARENT_ID` and an 863px inset, asserting the card clears its inset fails `expected:<863> but was:<0>`. This change turns it green with all 40 onboarding test classes still passing.

```suggestion
            // A decoration the card is not stacked above sits beside or over it and covers nothing below it, so
            // it only stands in for the inset on the legacy path, where the reservation still feeds dialogSpace.
            if (decorationShown && !reservesInsetAboveDecoration) return 0
            return cardBottomInsetPx()
```

Scoping it on the flag is safe because the loop this guard protected against (a reservation feeding `dialogSpace` and hiding the decoration it reserved against) is already broken for the config-driven path by `cardBottomMargin = 0` on line 117. Flag-off stays identical, and `whenCardBottomAnchoredButDecorationShownThenNoBottomInset` passes untouched.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
```diff
@@ -102,7 +111,11 @@ class OnboardingDecorationFitCorrector(
         val overflow = (cardContainerHeight - viewportHeight).coerceAtLeast(0)
         val available = root.height - root.paddingTop - root.paddingBottom
         val dialogParams = dialog.layoutParams as ViewGroup.MarginLayoutParams
-        val dialogSpace = dialogHeight + overflow + dialogParams.topMargin + dialogParams.bottomMargin
+        // A keyboard overlays the window rather than shrinking it, so `available` already covers the room the card
+        // gave up above the decoration. Counting it would shrink the decoration, which grows the inset, which
+        // shrinks the decoration again, until it is gone.
+        val cardBottomMargin = if (reservesInsetAboveDecoration) 0 else dialogParams.bottomMargin
```
mikescamell, line 117: 🤖 🔍 This reads as though it preserves legacy arithmetic, but control only reaches here with the flag on, and `syncCardBottomInset` has already forced `bottomMargin` to 0 in every non-reserving path with a shown decoration. So it is always 0 unless the params are not `ConstraintLayout.LayoutParams`. Either simplify to `0` or say why it stays.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrector.kt
```diff
@@ -130,19 +143,44 @@ class OnboardingDecorationFitCorrector(
...
+            return if (decorationShown) 0 else cardBottomInsetPx()
+        }
+        val deco = decoration?.takeIf { decorationShown && reservesInsetAboveDecoration } ?: return 0
+        return (cardBottomInsetPx() - roomBelowCard(deco)).coerceAtLeast(0)
+    }
+
+    private fun roomBelowCard(decoration: View): Int {
+        val params = decoration.layoutParams as ViewGroup.MarginLayoutParams
+        // The applied height, not the laid-out one: the corrector runs before the frame that lays the shrink out.
+        val height = params.height.takeIf { it > 0 } ?: decoration.measuredHeight
```
mikescamell, line 167: 🤖 🔍 Unreachable in production: `applyFit` always assigns an explicit height before `track()`. Fine as belt and braces, but this is where a wrong `roomBelowCard` would silently produce a wrong reservation, so it deserves a test or deletion.

app/src/test/java/com/duckduckgo/app/onboarding/ui/page/OnboardingDecorationFitCorrectorTest.kt
```diff
@@ -504,6 +507,223 @@ class OnboardingDecorationFitCorrectorTest {
         assertEquals(0, (h.dialog.layoutParams as ViewGroup.MarginLayoutParams).bottomMargin)
     }
 
+    @Test
+    fun whenInsetExceedsTheDecorationBelowThenCardReservesTheRemainder() {
```
mikescamell, line 511: 🤖 Every new reservation test starts at `bottomMargin == 0` and asserts a reservation is taken. Nothing asserts it comes back when the inset drops, which is the regression this commit's own message describes ("the card parked at the top of the screen and never came back").

Proof it is only untested rather than broken: threading a mutable inset supplier through the harness and flipping 863 to 0 gives 441 to 0, with the clamp releasing on the next pass. Green today, worth pinning.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/OnboardingIntroState.kt
```diff
@@ -66,20 +66,17 @@ class OnboardingIntroState {
...
-        SnapAway(canCrossFadeBackground = false),
+        SnapAway,
 
         /** The intro was already snapped away unplayed. */
-        AlreadyDismissed(canCrossFadeBackground = false),
+        AlreadyDismissed,
 
         /** An earlier dialog already took the background over. */
-        AlreadyHandedOver(canCrossFadeBackground = true),
+        AlreadyHandedOver,
```
mikescamell, line 80: 🤖 🔍 `canCrossFadeBackground` was the only thing separating these two, and the choreographer now maps both to `Unit`. Two constants and two tests encoding a distinction with no consequence. Collapsing them would let `cleared` and `handedOver` merge too.

app/src/main/java/com/duckduckgo/app/onboarding/ui/page/configdriven/ConfigDrivenWelcomePageFragment.kt
```diff
@@ -227,18 +252,13 @@ class ConfigDrivenWelcomePageFragment : OnboardingPageFragment(R.layout.content_
...
         screen: ConfigDrivenOnboardingPageViewModel.Screen.Dialog,
     ) {
-        // We assume that if intro is played, it's always done so before any dialog is rendered.
-        // Once the first dialog arrives, if:
-        // - intro visuals on screen: clear them and the background cross-fades from them
-        // - no intro visual on screen (like a mid-flow re-entry from another activity): nothing to animate from, snap new background
-        // Later renders always animate the background. The handover is unconditional so that a render which does not
-        // animate still takes the background off the choreographer.
-        val canCrossFadeBackground = intro?.clearForDialog() == true
+        // We assume that if intro is played, it's always done so before any dialog is rendered. The handover is
```
mikescamell, line 255: 🤖 🔍 The second sentence defends a choice no longer visible in the code now that `clearForDialog()` returns `Unit` and is always called.

mikescamell (approved):
A few minor things, but once they are addressed on resolved I think we're good to go! :shipit:

mikescamell:
@LukasPaczos 

Some nits that we can consider for future improvements:

1. When reopening that app while on a step, we replay the outgoing background, I think it probably makes sense just to show the incoming background

https://github.com/user-attachments/assets/4d4201df-9d2a-4a5a-bfa5-6b35ab744296

2. On the input selection screen, if you dismiss the keyboard and then rotate, the keyboard pops back up. My concern here is more in regards to remembering the ViewState, if the keyboard is down, I'd expect it to still be down when I rotate

https://github.com/user-attachments/assets/f8505357-91a4-4910-b35f-ff551ab275c2

PR #9460, opened 2026-08-11
Fix broken site refresh detection timing

laghee:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671677432066/task/1217360550002208?focus=true

### Description

We show the 3x refresh prompt to report site breakage 2.5x less often on Android than iOS. To address that, we're starting with this change to evaluate refresh patterns when reload occurs rather than after page load completes to prevent slow page loads from shrinking our detection window.

### Steps to test this PR

1. Install the internal debug build and skip/complete onboarding:
     `./gradlew installInternalDebug`

2. Run:
     `adb reverse tcp:8765 tcp:8765`

3. Start the local slow server using a Python script:
```
python3 - <<'PY'
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
from time import sleep
class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        if self.path.startswith("/slow"):
            sleep(15)
        self.send_response(200)
        self.send_header("Cache-Control", "no-store")
        self.end_headers()
        try:
            self.wfile.write(b"Reload detection test page")
        except BrokenPipeError:
            pass
    def log_message(self, *_):
        pass
ThreadingHTTPServer(("127.0.0.1", 8765), Handler).serve_forever()
PY
```

#### URL mismatch

  1. Open http://127.0.0.1:8765/slow.
  2. After loading, refresh 3x quickly.
  3. Immediately open http://127.0.0.1:8765/fast.
  4. Verify the broken-site prompt does not appear.

#### Slow load detection

  1. Open http://127.0.0.1:8765/slow again and wait for it to load.
  2. Refresh 3x quickly.
  3. Wait for the final load to finish.
  4. Verify “Site not working? Let us know” appears.

  Run the reset commands before repeating after the prompt has appeared.

```
 adb shell am force-stop com.duckduckgo.mobile.android.debug
 adb shell run-as com.duckduckgo.mobile.android.debug rm -f files/datastore/broken_site_prompt.preferences_pb
 adb shell monkey -p com.duckduckgo.mobile.android.debug 1
```

 That should clear only prompt cooldown/dismissal state, preserving onboarding.

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -3696,13 +3696,19 @@ class BrowserTabViewModel @Inject constructor(
             if (hasCtaBeenShownForCurrentPage.get() && isBrowserShowing) return null
             val detectedRefreshPatterns = brokenSitePrompt.getUserRefreshPatterns()
             handleBreakageRefreshPatterns(detectedRefreshPatterns)
+            val currentSite = siteLiveData.value
+            val currentNavigationUrl = webNavigationState?.currentUrl
+            val patternsForCurrentSite =
+                detectedRefreshPatterns.takeIf {
+                    currentNavigationUrl != null && currentSite?.url == currentNavigationUrl
+                } ?: emptySet()
             val cta =
                 withContext(dispatchers.io()) {
                     ctaViewModel.refreshCta(
                         dispatchers.io(),
                         isBrowserShowing && !isErrorShowing,
-                        siteLiveData.value,
-                        detectedRefreshPatterns,
+                        currentSite,
+                        patternsForCurrentSite,
                         suppressDuckAiOnboardingCta,
                     )
                 }
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -1817,7 +1817,8 @@ class BrowserTabViewModel @Inject constructor(
 
         if (triggeredByUser) {
             site?.realBrokenSiteContext?.onUserTriggeredRefresh()
-            site?.uri?.let {
+            val refreshedUrl = webNavigationState?.currentUrl?.toUri() ?: site?.uri
+            refreshedUrl?.let {
                 brokenSitePrompt.pageRefreshed(it)
             }
         }
@@ -3698,18 +3699,15 @@ class BrowserTabViewModel @Inject constructor(
             handleBreakageRefreshPatterns(detectedRefreshPatterns)
             val currentSite = siteLiveData.value
             val currentNavigationUrl = webNavigationState?.currentUrl
-            val patternsForCurrentSite =
-                detectedRefreshPatterns.takeIf {
-                    currentNavigationUrl != null && currentSite?.url == currentNavigationUrl
-                } ?: emptySet()
             val cta =
                 withContext(dispatchers.io()) {
                     ctaViewModel.refreshCta(
                         dispatchers.io(),
                         isBrowserShowing && !isErrorShowing,
                         currentSite,
-                        patternsForCurrentSite,
+                        detectedRefreshPatterns,
                         suppressDuckAiOnboardingCta,
+                        brokenSitePromptUrl = currentNavigationUrl,
                     )
                 }
             val contextDaxDialogsShown =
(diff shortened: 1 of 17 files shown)


broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSiteRefreshesInMemoryStore.kt
```diff
@@ -25,106 +25,97 @@ import java.time.LocalDateTime
...
-        if (doubleRefreshes == null && tripleRefreshes == null) {
-            return emptySet()
-        }
-        val thriceRefreshCount = tripleRefreshes?.time?.count {
-            it.isAfter(currentDateTime.minusSeconds(THRICE_REFRESH_WINDOW_IN_SECS))
-        } ?: 0
-
-        val twiceRefreshCount = doubleRefreshes?.time?.count {
-            it.isAfter(currentDateTime.minusSeconds(TWICE_REFRESH_WINDOW_IN_SECS))
-        } ?: 0
```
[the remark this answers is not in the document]
laghee: To avoid making an API change here, I'm going with an in-memory token per `RealBrokenSitePrompt`

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3697,13 +3696,19 @@ class BrowserTabViewModel @Inject constructor(
             if (hasCtaBeenShownForCurrentPage.get() && isBrowserShowing) return null
             val detectedRefreshPatterns = brokenSitePrompt.getUserRefreshPatterns()
             handleBreakageRefreshPatterns(detectedRefreshPatterns)
+            val currentSite = siteLiveData.value
+            val currentNavigationUrl = webNavigationState?.currentUrl
+            val patternsForCurrentSite =
+                detectedRefreshPatterns.takeIf {
+                    currentNavigationUrl != null && currentSite?.url == currentNavigationUrl
+                } ?: emptySet()
```
lmac012: I have 2 questions here:

1. The implementation fires pixels / increments experiment metrics right before this code is executed (inside `handleBreakageRefreshPatterns()`) - if `currentNavigationUrl` is different from  `currentSite?.url`, we effectively suppress the prompt but still fire the pixels - is that disconnect intended?
2. Does this logic have to live inside `BrowserTabViewModel`? If `currentNavigationUrl` holds a more up-to-date url value, shouldn't we just feed that into `BrokenSitePrompt.pageRefreshed()` and `BrokenSitePrompt.shouldShowBrokenSitePrompt()` instead of `currentSite?.url`?

broken-site/broken-site-impl/src/main/java/com/duckduckgo/brokensite/impl/BrokenSiteRefreshesInMemoryStore.kt
```diff
@@ -25,106 +25,124 @@ import java.time.LocalDateTime
...
 interface BrokenSiteRefreshesInMemoryStore {
-    fun resetRefreshCount(pattern: RefreshPattern)
-    fun addRefresh(url: Uri, localDateTime: LocalDateTime)
-    fun getRefreshPatterns(currentDateTime: LocalDateTime): Set<RefreshPattern>
+    fun addRefresh(owner: RefreshPatternOwner, url: Uri, localDateTime: LocalDateTime)
+    fun getRefreshPatterns(owner: RefreshPatternOwner, currentDateTime: LocalDateTime): Set<RefreshPattern>
+    fun isRefreshPatternDetectionValid(url: Uri, currentDateTime: LocalDateTime): Boolean
 }
 
+class RefreshPatternOwner
```
lmac012, line 37: I think this class would benefit from a comment explaining why it exists. At least to me, it wasn't obvious at first glance - specifically, the fact that we rely on having exactly one instance of `BrokenSitePrompt` per tab is not explicit enough.

lmac012 (commented):
Test scenarios work as expected 👍 I posted a couple of comments/questions.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
laghee: 1. Yes, the separation is intentional. There's a legitimate scenario where we want to fire the pixel even when we're suppressing the prompt -- if a user hits a slow-loading page, refreshes 3 times, then gives up and navigates to another URL before the last load completes, that's still data we want to capture (and is closer to iOS behavior, which doesn't wait until full page load to send the pixel or prompt) even if it doesn't make sense to prompt them on the new URL.
2. No, I was leery of this too, and your suggestion makes sense 👍

lmac012 (approved):

PR #9491, opened 2026-08-13
Stop pinning the inferred Duck.ai default model

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1174433894299346/task/1216424634539240?focus=true

### Description

`/duckchat/v1/models` exposes no default marker, so we infer the default by taking the first model the user's tier can access. The problem was that we then saved that inferred value to disk as if the user had picked it, on the first successful fetch. Later fetches only re-resolved when the saved model disappeared from the response or stopped being accessible, so reordering the list, or promoting a new default, never reached an install that already had one.

That write also fired on the first fetch after *upgrading* to a build containing the model picker, not just on new installs, so the pre-existing install base had its default snapshotted in one go shortly after #8157 shipped. The saved id is sent with every prompt, so it also overrides whatever default the backend would have applied. Net effect: most Android users are pinned to `gpt-5.4-mini` and no server-side default change can move them, while iOS follows the current default (see the investigation notes and Pete's confirmation of the iOS behaviour on the Asana task).

Changes:

- Persist a model only when the user explicitly picks one, and re-derive the default from every response. This matches iOS.
- Drop a saved pick that is gone from the response or no longer accessible, rather than replacing it with our fallback, so we never turn a fallback into a choice the user did not make.
- One-off migration clearing a saved `gpt-5.4-mini` so the affected cohort picks up the current default. A flag means it runs once, so a later deliberate pick of that model survives.

Wire behaviour is unchanged: we still send the resolved model id with the prompt. Existing chats are unaffected, they already resolve their own stored model and never read the global preference.

Two things worth knowing while reviewing:

- The migration hardcodes `gpt-5.4-mini`, deliberately, as a dated one-off repair for a known bad write. It rests on mini being the wrongly pinned value, which is still being confirmed against pixel data (`m_aichat_unified_input_prompt_submitted_count` carries `model_id`, `m_aichat_unified_input_model_selected` fires only on an explicit pick). Users who genuinely chose mini are reset too, since the stored value records no intent.
- Clearing the model preference orphans the saved reasoning mode, which then gets dropped by the existing orphan handling. Those users get the new default's first accessible reasoning mode.

### Steps to test this PR

_Default follows the server_
- [x] Fresh install, open the omnibar and switch to Duck.ai, confirm the picker chip shows `5.4-nano`, the first free model in the current list
- [x] Submit a prompt without touching the picker, confirm the chat runs on nano

_Explicit picks still stick_
- [x] Tap the chip, pick a different free model, submit a prompt, confirm that model is used
- [x] Kill and relaunch the app, confirm the picked model is still selected

_Migration_
- [x] On a build from `develop`, select `5.4-mini` so it is written to disk, then install this branch over it
- [x] Confirm the chip now shows `5.4-nano` rather than `5.4-mini`
- [x] Pick `5.4-mini` again, kill and relaunch, confirm it is still selected, the migration does not run twice

### UI changes
No UI changes. The picker renders the same, it just resolves the default from the current response instead of a value frozen on first run.

---

> [!NOTE]
> **Medium Risk**
> Changes global Duck.ai model resolution and prefs for all users, including a hardcoded one-time migration that may reset users who truly chose mini; prompt wiring still sends the resolved id.
> 
> **Overview**
> Fixes Android Duck.ai installs that were **pinned to `gpt-5.4-mini`** because the inferred default (first accessible model from `/duckchat/v1/models`) was written to prefs on fetch, as if the user had chosen it.
> 
> **Selection logic:** `fetchModels` now uses **`resolveSelection`** instead of validating and re-persisting the default. Only an explicit **`selectModel`** call saves a model id. On each fetch, users with no valid persisted pick get the **first accessible model** from the current response, so list reordering and new server defaults can land without a stored override.
> 
> **Invalid picks:** If a saved id is missing from the response or no longer accessible, prefs are **cleared** (`setSelectedModel(null)`); the UI still falls back to the first accessible model in memory, but that fallback is **not** written back as a user choice.
> 
> **One-off migration:** On startup, **`clearPinnedDefaultModel`** runs once (flag `DUCK_AI_CLEARED_PINNED_DEFAULT_MODEL`) and removes a stored **`gpt-5.4-mini`** selection so the affected cohort picks up the current default. Deliberate picks of that model after migration are kept.
> 
> Tests cover migration edge cases, reorder behavior, and that fetch no longer calls `setSelectedModel` for inferred defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e11935337f4fa7b4f48595436463d69ed3172d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/DuckAiModelManager.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/models/DuckAiModelManager.kt
@@ -97,6 +97,12 @@ class RealDuckAiModelManager @Inject constructor(
     init {
         appCoroutineScope.launch(dispatcherProvider.io()) {
             try {
+                // Separate from the restore below so a failed migration cannot cost the user their selection.
+                stateMutex.withLock { clearPinnedDefaultModel() }
+            } catch (e: Exception) {
+                logcat { "Duck.ai Model Manager: failed to clear pinned default model: ${e.message}" }
+            }
+            try {
                 stateMutex.withLock { restoreCachedSelection() }
             } catch (e: Exception) {
                 logcat { "Duck.ai Model Manager: failed to restore cached selection: ${e.message}" }
@@ -108,6 +114,21 @@ class RealDuckAiModelManager @Inject constructor(
                     fetchModels()
                 }
         }
+    }
+
+    /**
+     * Earlier versions persisted the inferred default as if the user had picked it, which pinned most
+     * of the install base to [PINNED_DEFAULT_MODEL_ID] and stopped later server-side default changes
+     * from reaching them. Clearing that one id lets [resolveSelection] derive the current default
+     * again. Users who deliberately picked it are reset too, as the stored value records no intent.
+     */
+    private suspend fun clearPinnedDefaultModel() {
+        if (dataStore.hasClearedPinnedDefaultModel()) return
+        if (dataStore.getSelectedModel()?.id == PINNED_DEFAULT_MODEL_ID) {
+            dataStore.setSelectedModel(null)
+            logcat { "Duck.ai Model Manager: cleared pinned default model" }
+        }
+        dataStore.setClearedPinnedDefaultModel()
     }
 
     private suspend fun restoreCachedSelection() {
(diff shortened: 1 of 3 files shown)


joshliebe (approved):
Tested, works as expected now 🛳️

PR #9499, opened 2026-08-14
Document the Smartling translation process

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1214901934989258/task/1217484868332131

### Description

Our Smartling translation process was documented only in Asana, so nothing in the repo told an engineer — or an agent working in it — which file a new string belongs in, why a placeholder needs an `instruction`, or how to change copy that is already translated in 24 languages.

Adds `.claude/docs/translations.md`, distilled from the [Android Smartling Translation Guide](https://app.asana.com/1/137249556945/project/1202561462274611/task/1203224618541800) and the [Typical Development Flow and FAQs](https://app.asana.com/1/137249556945/project/1202561462274611/task/1211223880022672):

- Where strings live: `donottranslate.xml` while copy is in flux, `values/strings-<module>.xml` once it's final and ready to be picked up by the repository connector
- The rules the connector and lint enforce: required Smartling directives, an `instruction` on every placeholder, positional `%1$s`, plurals, entity escaping, character limits, and why lint annotations must not live inside a translated string
- Recipes for adding, changing and removing already-translated copy — notably that an English change is delete-plus-new-key rather than an edit in place, so no locale keeps a stale translation
- Links back to the Asana guides and the Translation/Localization AOR

Also adds a row to the CLAUDE.md table so the doc gets read whenever a change touches user-facing strings.

Docs only — no product code, no strings.

### Steps to test this PR

_Documentation review_
- [ ] Read `.claude/docs/translations.md` and check it matches how we actually use Smartling
- [ ] Confirm the locale list and file mapping it describes match `smartling-config.json`
- [ ] Confirm the CLAUDE.md row reads consistently with the rest of the table

### UI changes
| Before  | After |
| ------ | ----- |

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no product code or string resource edits.
> 
> **Overview**
> Adds **in-repo documentation** for the Smartling workflow (previously only in Asana) so engineers and agents know how to handle user-facing strings.
> 
> **`.claude/docs/translations.md`** covers where English copy lives (`donottranslate.xml` during development vs `strings-<module>.xml` for connector pickup), required Smartling directives and lint rules (`MissingSmartlingRequiredDirectives`, placeholders, `instruction`, plurals, escaping, character limits), and workflows for **add / change / remove** already-translated strings—including **delete old key + new key** for English copy changes so locales don’t keep stale translations. It also points to Asana references and the Translation/Localization AOR.
> 
> **`CLAUDE.md`** gains a table row so this doc is consulted whenever work touches user-facing strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 644c3ff61d2a0ac029c7033730daeeba61712ee6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/docs/translations.md
+++ b/.claude/docs/translations.md
@@ -24,8 +24,8 @@ lint rule fails the build otherwise:
 ```
 
 **During development, add new strings to `donottranslate.xml`.** Move them into the real strings file
-only when the copy is final and you're ready to request translations — see
-`.claude/skills/request-translations/SKILL.md` for that flow.
+only when the copy is final and you're ready to request translations: pushing half-finished copy costs
+days, since each round trip through Smartling takes a few.
 
 ## Rules for translatable strings
 
--- a/.claude/docs/translations.md
+++ b/.claude/docs/translations.md
@@ -29,19 +29,25 @@ days, since each round trip through Smartling takes a few.
 
 ## Rules for translatable strings
 
-1. **Every placeholder needs an `instruction`** explaining what it stands for. Translators see the
-   `instruction` attribute and nothing else — not the surrounding code, not the screen.
+1. **Add an `instruction` only where the string isn't self-evident** — always for a string with a
+   placeholder, and for copy whose meaning depends on the screen around it. Translators see the
+   `instruction` attribute and nothing else, not the surrounding code and not the screen. Plain
+   unambiguous copy ("Settings", "Done") needs none; don't add one for its own sake.
 
    ```xml
    <string name="authenticationDialogMessage" instruction="Placeholder is the name of a website">%1$s requires a username and password.</string>
    ```
 
-2. **Use positional placeholders** — `%1$s`, `%2$s`, never bare `%s`. Word order differs by language, so
+2. **Reuse copy that is already translated** instead of adding a key for it. Common words like "Cancel"
+   or "Done" live in the design system's `strings-common-ui.xml` and are translated in all 24 locales,
+   so point at the existing string. A duplicate key means paying for a translation we already have.
+
+3. **Use positional placeholders** — `%1$s`, `%2$s`, never bare `%s`. Word order differs by language, so
    translators need to be able to reorder them.
 
-3. **Don't skip plurals.** Use `<plurals>` where the English has a count; quantity sets differ per language.
+4. **Don't skip plurals.** Use `<plurals>` where the English has a count; quantity sets differ per language.
 
-4. **Entity escaping is off globally.** A string that needs escaping has to flip the directive around
+5. **Entity escaping is off globally.** A string that needs escaping has to flip the directive around
    itself, and flip it back afterwards:
 
    ```xml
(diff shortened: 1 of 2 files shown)


.claude/docs/translations.md
```diff
@@ -0,0 +1,83 @@
...
+<!-- smartling.instruction_attributes = instruction -->
+```
+
+**During development, add new strings to `donottranslate.xml`.** Move them into the real strings file
+only when the copy is final and you're ready to request translations: pushing half-finished copy costs
+days, since each round trip through Smartling takes a few.
+
+## Rules for translatable strings
+
+1. **Every placeholder needs an `instruction`** explaining what it stands for. Translators see the
```
catalinradoiu: Is this a rule of thumb? So far we were not adding instructions to all of the strings, right?

.claude/docs/translations.md
```diff
@@ -0,0 +1,83 @@
...
+reorder and inflect, and a fragment gives the translator nothing to work with.
+
+## Changing strings that are already translated
+
+| Goal | What to do |
+|---|---|
+| Add a string | Add the English string to the strings file |
+| Change English copy | Delete the old key, add a **new key** with the new copy, update references. Leave the other languages alone — Smartling prunes them |
+| Remove a string | Delete the English string only. Smartling removes the translations |
+| Fix a bad translation | Edit it in the Smartling dashboard (DuckDuckGo Android project → language → search the key → Edit Translation). It lands with the next job; to ship sooner, also edit the `values-<locale>` file directly |
```
catalinradoiu: Do we need this for the LLM? Or is the developer going to handle this?

.claude/docs/translations.md
```diff
@@ -0,0 +1,83 @@
...
+
+Reusing a key with different copy leaves every locale holding a stale translation of the old text,
+which is why an English change is delete-plus-new-key rather than an edit in place.
+
+## Reference
+
+- [Android Smartling Translation Guide](https://app.asana.com/1/137249556945/project/1202561462274611/task/1203224618541800)
```
catalinradoiu: Do we need all these references? Will the LLM use all of them or are all of them useful for it?

.claude/docs/translations.md
malmstein: There should be an instruction for each string that's not self explanatory. Copy like "Cancel" should not be added again as new strings since they are already translated.

.claude/docs/translations.md
malmstein: Developer should, LLM won't have access unless unless added via MCP. This instruction brings it to the user's attention.

.claude/docs/translations.md
malmstein: LLM uses them, and the real source of truth for how translations work. This is a local skill, so access to Asana is enabled.

.claude/docs/translations.md
catalinradoiu: Yes but in this case the LLM is always going to add it when moving the copy from donottranslate.xml to strings.xml. Should we mention this explicitly that we need the instruction only when the string is not already self explanatory? Otherwise the LLM is going to always add it.

.claude/docs/translations.md
catalinradoiu: So the LLM is going to flag this based on this explanation?

.claude/docs/translations.md
catalinradoiu: But some of them are actually needed by the LLM? For example will it need the `Translation/Localization AOR`? Could it end up using more tokens by trying to gather more information from all of these Asana links?

catalinradoiu (approved):

PR #9477, opened 2026-08-12
Move AndroidBrowserConfigFeature to new browser-feature-toggles module

landomen:
Task/Issue URL: https://app.asana.com/1/137249556945/project/488551667048375/task/1216757351941573?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/481882893211075/task/1216757351941571?focus=true
API Proposals URL(s) (if applicable): https://app.asana.com/1/137249556945/project/1212087397361015/task/1216757351941572?focus=true

### Description
Moves `AndroidBrowserConfigFeature` to new `browser-feature-toggles` module, updates all usages of it and adds a lint detector to prevent new feature toggles from being added to the module.

### Steps to test this PR

_Smoke check_
- [x] Smoke test the app
- [x] (optional) Add a new dummy feature toggle to `AndroidBrowserConfigFeature`, run Lint and verify it fails

### UI changes
No UI changes

---

> [!NOTE]
> **Medium Risk**
> Wide import and Gradle dependency churn across app and tests with no intended runtime change; miswired module dependency could break build or toggle injection at compile time.
> 
> **Overview**
> **Moves `AndroidBrowserConfigFeature` out of `:app` into a new `:browser-feature-toggles` library** and rewires `:app` plus every consumer to import `com.duckduckgo.browser.feature.toggles.AndroidBrowserConfigFeature` instead of `com.duckduckgo.app.pixels.remoteconfig`.
> 
> The new module is a standard Anvil/KSP library (namespace `com.duckduckgo.browser.feature.toggles`) holding the `@ContributesRemoteFeature` interface; behavior of existing toggles is unchanged—this is packaging and dependency graph cleanup.
> 
> **Freezes the module with lint:** `NoNewBrowserFeatureToggleDetector` errors on any new toggle method under `com.duckduckgo.browser.feature.toggles` that is not in a hardcoded grandfather list of existing `AndroidBrowserConfigFeature` methods, steering new flags into feature-owned modules instead of growing the catch-all interface again.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d811ed33b8cf610c1504ccb58bc3d5cff00bde1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9522, opened 2026-08-18
Add feature flag for the Suggest redirect feature

juandiana:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1216807998862658/task/1217572324167492?focus=true

### Description
Add the feature flag interface inside the package where the feature will reside. Default value set to `false`.

### Steps to test this PR
1. Tap on Settings > Feature Flag Inventory.
2. Search for “suggest”.
3. Verify that both the `suggestRedirectOnUnresolvedError` toggle and its `suggestRedirect` sub-feature toggle are present.

### UI changes
N/A

---

> [!NOTE]
> **Low Risk**
> Toggle-only addition with no navigation or error-page logic changes; risk is limited to feature inventory registration.
> 
> **Overview**
> Introduces **`SuggestRedirectOnUnresolvedErrorFeature`** in `app/browser/suggestredirect` as a remote-controllable flag for an upcoming behavior: on unresolved bare-domain errors, optionally offer a link to the `www` subdomain when it resolves.
> 
> The feature is registered as **`suggestRedirectOnUnresolvedError`** with **`self()`** and a **`suggestRedirect()`** sub-toggle; both default to **off**. No user-facing behavior is wired in this PR—only inventory/toggle plumbing for later rollout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0879433e0f396954e5bfd610f55e5b1ba7134a47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/suggestredirect/SuggestRedirectOnUnresolvedErrorFeature.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/suggestredirect/SuggestRedirectOnUnresolvedErrorFeature.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.suggestredirect
+
+import com.duckduckgo.anvil.annotations.ContributesRemoteFeature
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.feature.toggles.api.Toggle
+import com.duckduckgo.feature.toggles.api.Toggle.DefaultFeatureValue
+
+/**
+ * Controls whether the error page shown for an unresolved bare domain (e.g. example.com) offers a
+ * suggestion link to its "www" subdomain (e.g. www.example.com) when such subdomain resolves.
+ */
+@ContributesRemoteFeature(
+    scope = AppScope::class,
+    featureName = "suggestRedirectOnUnresolvedError",
+)
+interface SuggestRedirectOnUnresolvedErrorFeature {
+    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    fun self(): Toggle
+
+    @Toggle.DefaultValue(DefaultFeatureValue.FALSE)
+    fun suggestRedirect(): Toggle
+}


app/src/main/java/com/duckduckgo/app/browser/suggestredirect/SuggestRedirectOnUnresolvedErrorFeature.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Controls whether the error page shown for an unresolved bare domain (e.g. example.com) offers a
+ * suggestion link to its "www" subdomain (e.g. www.example.com) when such subdomain resolves.
+ */
+@ContributesRemoteFeature(
+    scope = AppScope::class,
+    featureName = "suggestRedirectOnUnresolvedError",
+)
+interface SuggestRedirectOnUnresolvedErrorFeature {
```
malmstein, line 32: this should live inside `AndroidBrowserConfigFeature` and not as a separate flag.

malmstein (changes requested):
Move the flag inside `AndroidBrowserConfigFeature`, it’s where are all the browser related feature flags live and where you’ll see other error page related flags too.

app/src/main/java/com/duckduckgo/app/browser/suggestredirect/SuggestRedirectOnUnresolvedErrorFeature.kt
juandiana: That was my initial choice. However, a new lint, `NoNewBrowserFeatureToggle`, was introduced last week that prevents new toggles from being added to the `AndroidBrowserConfigFeature` interface. You may see it ([here](https://github.com/duckduckgo/Android/pull/9477/changes#diff-07dcb069bd0c92cf2d80a85ece667487441e5a7024e46b539a65600cdecb60f5R142)).

juandiana: cc @landomen

landomen: @malmstein @juandiana  Correct, as written in https://app.asana.com/1/137249556945/project/1202561462274611/task/1217462888888561?focus=true we recently moved `AndroidBrowserConfigFeature` from `:app` module to a new module to unblock extraction of several features from the `:app` module. 

We want to avoid adding new feature flags to the new module and the guidance is that they should be owned by whatever feature actually needs it. If we strongly believe that a new feature flag belongs in `AndroidBrowserConfigFeature` then an exception can be made. Do we think this is such a case?

cc @aibrahim- @anikiki

malmstein: I missed that task Domen, my bad. That makes total sense then, let's follow the new rules instead and keep this outside.

juandiana: I believe the PR is already in good shape. It follows the first case from the article @landomen cited:

> One -impl module reads it → put the @ContributesRemoteFeature interface in that -impl module.

Please let me know if I missed something otherwise.

landomen: No, looks good @juandiana

landomen (approved):

malmstein (approved):
Discussed with @landomen  and @anikiki . Moving forward with this

PR #9542, opened 2026-08-19
Retire wirePluginView() in native input

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1157893581871903/task/1217549691747336
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/task/1217490861467320

### Description

`NativeInputModeWidget.wirePluginView()` handed dependencies and callbacks straight to concrete plugin views, which is what forced every control migration through the widget, and it is now gone: a shared `EffectiveModelProvider` resolves the effective model, options derives its own tool visibility, the picker's menu events moved onto `NativeInputHost`, and `AttachmentView` injects its own dependencies. The non-obvious part is that submission still needs the staged attachments, so the widget resolves `AttachmentViewModel` through its own `ViewModelStoreOwner` instead of holding the plugin's view, which avoided both an attachment store and a `duckchat-api` change. Because state and the resolved model arrive on separate flows, both options and the picker chip now ignore an answer resolved for a different chat, otherwise a tab switch could clear a tool the arriving chat supports or name the previous tab's model. This PR also touches `:app`, because testing surfaced a Duck.ai tab with no input field at all: the bottom host's offset was measured against a stale layout when the keyboard hid, and it now recomputes when the host itself is laid out (`cc6f3b5037`).

### Steps to test this PR

_Model and tools across tabs_
- [ ] Open a Duck.ai chat and select Web Search, then switch to a second Duck.ai tab and pick a model that does not support it
- [ ] Return to the first tab, confirm the tool list matches that tab's model, and confirm a tool that tab's model does support is not dropped
- [ ] Switch between two Duck.ai tabs on different models and confirm the chip always names the model of the tab you are on
- [ ] Start a model change in one tab and then in a second tab, switch between them, and confirm each keeps its own pick
- [ ] Dismiss the model picker without choosing a model and confirm the chip reverts

_Input position and presence_
- [ ] On a Duck.ai tab, open and close the keyboard, switch tabs, and confirm the input stays flush at the bottom and never disappears

_Attachments_
- [ ] Attach an image from the camera and from the file picker, exceed the attachment limit, remove one, then submit and confirm the prompt carries the attachments
- [ ] In the contextual sheet, confirm page context attaches, can be removed, and that "ask about page" works
- [ ] From the edit-message screen, confirm existing attachments are adopted

### UI changes

No UI changes. Controls render as before, and the `:app` fix restores the input's original resting position.

### DIFF
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
@@ -41,9 +41,9 @@ interface EffectiveModelProvider {
 
     /**
      * Records the model picked during the FE model-change recovery flow. It wins over the chat's stored
-     * model until that window closes, because the FE syncs the new model back to us asynchronously.
+     * model while that window is open, because the FE syncs the new model back to us asynchronously.
      */
-    fun onRecoveryModelPicked(chatId: String?, modelId: String)
+    fun onRecoveryModelPicked(modelId: String)
 }
 
 @OptIn(ExperimentalCoroutinesApi::class)
@@ -55,15 +55,13 @@ class RealEffectiveModelProvider @Inject constructor(
     private val duckAiChatStore: DuckAiChatStore,
 ) : EffectiveModelProvider {
 
-    private data class RecoverySelection(val chatId: String?, val modelId: String)
-
     private data class ActiveChat(
         val chatId: String?,
         val modelChangeMode: Boolean,
         val chatModel: String?,
     )
 
-    private val recoverySelection = MutableStateFlow<RecoverySelection?>(null)
+    private val recoveryModelId = MutableStateFlow<String?>(null)
 
     // mapLatest: a chatId flip cancels an in-flight lookup, so a slow read can't resolve into a stale chat.
     private val activeChat: Flow<ActiveChat> = nativeInputStateProvider.state
@@ -80,17 +78,17 @@ class RealEffectiveModelProvider @Inject constructor(
     override val effectiveModelId: Flow<String?> = combine(
         modelManager.modelState,
         activeChat,
-        recoverySelection,
+        recoveryModelId,
     ) { modelState, chat, recovery ->
         val modelIds = modelState.models.mapTo(HashSet()) { it.id }
+        // Honoured only while the model-change window is open; it closes per tab, so a stale pick cannot leak.
         val recovered = recovery
-            ?.takeIf { chat.modelChangeMode && it.chatId == chat.chatId }
-            ?.modelId
+            ?.takeIf { chat.modelChangeMode }
             ?.takeIf { it in modelIds }
         recovered ?: chat.chatModel?.takeIf { it in modelIds } ?: modelState.selectedModelId
     }.distinctUntilChanged()
 
-    override fun onRecoveryModelPicked(chatId: String?, modelId: String) {
-        recoverySelection.value = RecoverySelection(chatId, modelId)
+    override fun onRecoveryModelPicked(modelId: String) {
+        recoveryModelId.value = modelId
     }
 }
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
@@ -16,7 +16,7 @@
 
 package com.duckduckgo.duckchat.impl.nativeinput
 
-import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.di.scopes.ActivityScope
 import com.duckduckgo.duckchat.api.nativeinput.NativeInputStateProvider
 import com.duckduckgo.duckchat.impl.models.DuckAiModelManager
 import com.duckduckgo.duckchat.store.impl.DuckAiChatStore
@@ -47,8 +47,10 @@ interface EffectiveModelProvider {
 }
 
 @OptIn(ExperimentalCoroutinesApi::class)
-@SingleInstanceIn(AppScope::class)
-@ContributesBinding(AppScope::class)
+// ActivityScope, not AppScope: the unqualified DuckAiChatStore is bound per activity so it can resolve
+// the browser mode. One instance per activity is what the plugins sharing a widget need anyway.
+@SingleInstanceIn(ActivityScope::class)
+@ContributesBinding(ActivityScope::class)
 class RealEffectiveModelProvider @Inject constructor(
     private val modelManager: DuckAiModelManager,
     private val nativeInputStateProvider: NativeInputStateProvider,
--- a/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
+++ b/duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.duckchat.impl.nativeinput
+
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.duckchat.api.nativeinput.NativeInputStateProvider
+import com.duckduckgo.duckchat.impl.models.DuckAiModelManager
+import com.duckduckgo.duckchat.store.impl.DuckAiChatStore
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import javax.inject.Inject
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.mapLatest
+
+/**
+ * The model whose capabilities the native input controls should reflect for the active tab. Shared so
+ * the model picker and the options menu resolve it identically instead of one asking the other.
+ */
+interface EffectiveModelProvider {
+
+    val effectiveModelId: Flow<String?>
+
+    /**
+     * Records the model picked during the FE model-change recovery flow. It wins over the chat's stored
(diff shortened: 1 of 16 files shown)


duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
```diff
@@ -0,0 +1,96 @@
...
+        // Honoured only while the model-change window is open; it closes per tab, so a stale pick cannot leak.
+        val recovered = recovery
+            ?.takeIf { chat.modelChangeMode }
+            ?.takeIf { it in modelIds }
+        recovered ?: chat.chatModel?.takeIf { it in modelIds } ?: modelState.selectedModelId
+    }.distinctUntilChanged()
+
+    override fun onRecoveryModelPicked(modelId: String) {
+        recoveryModelId.value = modelId
+    }
```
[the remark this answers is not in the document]
malmstein: good catch, this is real. gating on modelChangeMode only hid the stale pick, so a second window in the same activity would reapply it while the picker's tick showed something else.

the picker already owns that lifecycle (it resets its local copy when the window closes) so it now clears the provider at the same point. covered at both levels in 1e8e463b5d: the provider contract, and the picker case that actually regressed, which fails without the clear.

this is the hole I opened when I dropped the chatId key from onRecoveryModelPicked, so worth having caught it here.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/AttachmentView.kt
```diff
@@ -87,16 +91,13 @@ class AttachmentView(
...
-        nativeInputStateProvider: NativeInputStateProvider,
-        faviconManager: FaviconManager,
-    ) {
+    override fun onAttachedToWindow() {
+        AndroidSupportInjection.inject(this)
+        super.onAttachedToWindow()
         val owner = findViewTreeViewModelStoreOwner() ?: return
-        val vm = ViewModelProvider(owner, factory)[AttachmentViewModel::class.java]
+        val scope = findViewTreeLifecycleOwner()?.lifecycleScope ?: return
+        val vm = ViewModelProvider(owner, viewModelFactory)[AttachmentViewModel::class.java]
```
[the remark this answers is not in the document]
malmstein: correct, fixed in 7423218116. the job is kept and cancelled next to nativeInputStateJob on detach.

worth noting this only became reachable because the view now injects itself and collects in onAttachedToWindow, where before bind() was called once per created view, so a re-attach really would have stacked collectors.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/AttachmentView.kt
```diff
@@ -33,45 +33,49 @@ import android.widget.ScrollView
...
     var isEditMode: Boolean = false
-    var onAskAboutPage: (() -> Unit)? = null
-    var onPageContextRemoved: (() -> Unit)? = null
 
     private var viewModel: AttachmentViewModel? = null
-    private var faviconManager: FaviconManager? = null
+
+    /** The contextual sheet is the only surface that offers page context. */
+    private val isContextual: Boolean
+        get() = lastNativeInputState?.inputContext == NativeInputState.InputContext.DUCK_AI_CONTEXTUAL
```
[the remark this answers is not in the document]
malmstein: agreed, this one is a behaviour change I should not have made. the sticky flag existed precisely because the sheet and the omnibar widget share a per-tab state slot that can briefly read BROWSER.

7423218116 puts it back to sticky, but asked of the host (isContextualSurface()) rather than pushed into the view, which keeps the plugin off the widget's setters. same shape as isEditSurface().

for the record isEditMode is unaffected: it was previously assigned inside the same plugin-attach loop, so reading it from the host at creation has identical timing.

malmstein:
### Verification

**Unified-input Maestro suite, green on a local API 34 emulator** (`--include-tags unifiedInputTest`, internal release build of `7423218116`, 6/6 flows, each across top/bottom/split omnibar):

| Flow | |
|---|---|
| duck.ai tool pills select and remove | Passed (1m 33s) |
| duck.ai web search prompt sticky-then-clears | Passed (2m 47s) |
| open duck.ai via icon (AI off) and back to NTP | Passed (2m 2s) |
| ask duck.ai suggestion opens chat | Passed (1m 44s) |
| seeded favorites visible across omnibar × AI combos | Passed (52s) |
| search suggestions visible and dismiss | Passed (56s) |

The tool-pills flow is the relevant one for this PR: it exercises the path where options now derives tool visibility from `EffectiveModelProvider` instead of asking the model picker's view.

**Attachments verified manually on device** (attaching photos, file limits, and related paths). There are no attachment flows in `.maestro`, so this could only be done by hand, and it is the area two of the three Bugbot findings landed in.

**Unit:** `:duckchat-impl:testDebugUnitTest` and `:duckchat-impl:spotlessCheck` green; `:app` compiles for both play and internal. The recovery-pick regression test was checked properly, by reverting only the production line and confirming `ModelPickerViewModelTest` fails without it.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/ModelPickerViewModel.kt
```diff
@@ -94,28 +95,16 @@ class ModelPickerViewModel @Inject constructor(
             nativeInputStateProvider.state.collect { state ->
                 modelChangeMode = state.modelChangeMode
                 pixelSurface = DuckChatPixelSurface.from(state.inputContext)
-                if (!state.modelChangeMode) recoverySelectedModelId.value = null
+                if (!state.modelChangeMode) {
+                    recoverySelectedModelId.value = null
+                    effectiveModelProvider.clearRecoveryModelPick()
+                }
```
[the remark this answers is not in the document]
malmstein: right, and this is the same defect as the first finding seen from the other side: I stored one activity-scoped pick for what is per-tab state. clearing hid the stale-reuse case but broke the still-open-window case.

225495d4d2 keys the pick by the chatId whose window it belongs to. it applies only while that chat's window is open, and a clear only drops a pick for the same chat, so one tab's window closing cannot touch another's. the picker takes the key from published state rather than currentChat, which can still be resolving at pick time (that was what broke my first attempt at keying it).

two provider tests cover both directions, and I checked they fail with the clear unkeyed.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    // mapLatest: a chatId flip cancels an in-flight lookup, so a slow read can't resolve into a stale chat.
+    private val activeChat: Flow<ActiveChat> = nativeInputStateProvider.state
+        .map { it.chatId to it.modelChangeMode }
+        .distinctUntilChanged()
+        .mapLatest { (chatId, modelChangeMode) ->
+            ActiveChat(
+                chatId = chatId,
+                modelChangeMode = modelChangeMode,
+                chatModel = chatId?.let { duckAiChatStore.getChatById(it)?.model },
```
YoussefKeyrouz: q: The FE writes the newly selected model asynchronously right? So If I am in model recovery flow, I select a new model and submit immediately, is there a possibility where the chatModel would be stale here because it's reading from a snapshot instead of the flow (so nothing updates after write)?

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    private data class ActiveChat(
+        val chatId: String?,
+        val modelChangeMode: Boolean,
+        val chatModel: String?,
+    )
+
+    private data class RecoveryPick(val chatId: String?, val modelId: String)
+
+    private val recoveryPick = MutableStateFlow<RecoveryPick?>(null)
```
YoussefKeyrouz: q: Should this be per chat? I don't know if we can get into a state where multiple tabs are in recovery mode at the same time, but if we do then selecting on one tab could override the other selection.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/OptionsViewModel.kt
```diff
@@ -46,21 +51,28 @@ class OptionsViewModel @Inject constructor(
...
-     * through [com.duckduckgo.duckchat.impl.nativeinput.NativeInputHost.toolSelected]).
+     * Emits when the selected tool is no longer supported by the effective model. Deliberately not a
+     * pixel: the user did not deselect it, the model change did.
      */
-    fun updateVisibleTools(tools: Set<Tool>): Boolean {
-        _visibleTools.value = tools
-        val current = selectedTool.value
-        return current != null && current !in tools
-    }
+    val toolSelectionCleared: Flow<Unit> = combine(selectedTool, visibleTools) { selected, visible ->
```
YoussefKeyrouz: I think there could be a mismatch between selectedTool and visibleTools because they update at different times. (Selected changes imediately when tab is switched while visibleTools is a flow that waits for the model to be loaded from storage). So it could still contain the previous tab's tool on load. I reproduced by:
- Select Web Search on a tab
- Switch to another duck.ai tab and select a model that doesn't support Web Search
- Go back to first tab, the model is different but the Web Search is still there. 

See video:

https://github.com/user-attachments/assets/938b9a95-e706-4dd4-ab65-6766a35c8050

YoussefKeyrouz (approved):
The scenarios I tested looks good. I have 2 questions, and found n issue with switching tabs. But none of them are blocking. Feel free to address whichever you find important or as a follow up.

Approved.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
malmstein: yes, that was a real gap. the chat model came from a one-shot getChatById per chatId change, so a model the FE wrote back later was never picked up until the chatId changed. it now comes from getChatsFlow(), covered by whenTheChatsFlowReportsANewModelThenTheEffectiveModelFollowsIt in 7bb5304ae4.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/nativeinput/EffectiveModelProvider.kt
malmstein: it was keyed per chat already, but held in a single slot, so two chats in a window at once meant the second pick evicted the first. now a map keyed by chatId, with whenTwoChatsAreInRecoveryThenEachKeepsItsOwnPick covering exactly that case.

duckchat/duckchat-impl/src/main/java/com/duckduckgo/duckchat/impl/ui/nativeinput/views/OptionsViewModel.kt
malmstein: reproduced and fixed, thanks for the video. your read was right, and the consequence is worse than the stale pill: selectedTool comes from state immediately while the tool set waited on the chat lookup, so the auto-clear could fire against the previous tab's model and drop a tool the arriving tab's own model supports. the provider now reports Unresolved until the active tab's chat is read, and options offers every tool and clears nothing in that window. whenTabIsSwitchedThenTheClearDoesNotFireAgainstTheOtherTabsModel failed before the fix and passes now.

PR #9545, opened 2026-08-19
Clear URL when switching to AI mode

joshliebe:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200204095367872/task/1217646113585153?focus=true

### Description

- Clears an URL when toggling to Duck.ai in the native input (and puts it back if you toggle back to Search)
- Also fixes a bug where the wrong URL is shown in the input: If you enter a URL and focus the omnibar quickly, it still shows the old URL
- Adds kill switches for both issues

### Steps to test this PR

_URL clearing_
- [x] Type a URL in Search and submit
- [x] Tap the omnibar and toggle to Duck.ai
- [x] Verify that the URL is cleared
- [x] Toggle back to Search
- [x] Verify that the URL is repopulated
- [x] Type a query in Search
- [x] Toggle to Duck.ai
- [x] Verify that the query persists
- [x] Type a query in Duck.ai
- [x] Toggle to Search
- [x] Verify that the query persists
- [x] Type a URL in Search
- [x] Toggle to Duck.ai
- [x] Verify that the URL persists

_Prefill bug_
- [x] Enter a URL in Search and submit
- [x] Tap the omnibar and enter a new URL/query
- [x] Tap the omnibar again quickly
- [x] Verify that it shows the new URL/query

_Kill switches_
- [ ] In feature flag inventory, disable `omnibarPreFillKillSwitch`
- [ ] Verify old behavior (old URL shows when tapping quickly)
- [ ] In feature flag inventory, disable`nativeInputUrlClearingFeature`
- [ ] Verify old behavior (URL not cleared when toggling)

### UI changes

Before (URL clearing)

https://github.com/user-attachments/assets/c24bef35-bb98-40a8-9c26-c6c7be4ac9b5

After (URL clearing)

https://github.com/user-attachments/assets/6e65ec4a-7594-4266-b94b-1ae0c36b2f6f

Before (Prefill bug)

https://github.com/user-attachments/assets/4ff62cb4-6152-4cc0-bd44-283247d1a710

After (Prefill bug)

https://github.com/user-attachments/assets/dfabbc52-20fd-40af-809d-7b9924b9ac6b

---

> [!NOTE]
> **Medium Risk**
> Touches native input and omnibar prefill, which can hide or restore URLs incorrectly if toggle/cache logic is wrong. Both behaviors are behind remote flags.
> 
> **Overview**
> When switching native input from Search to Duck.ai, an **unedited prefilled URL is now cleared** (and restored if the user switches back to Search). Search queries and user-edited text stay put. Behavior is gated by `nativeInputUrlClearingFeature` (internal default).
> 
> Also fixes native-input **prefill** so a newly typed URL/query is not overwritten by the previously loaded page URL when the omnibar is tapped quickly. Prefill uses the loaded URL only when the displayed text still matches it. `omnibarPreFillKillSwitch` (on by default) can restore the old always-use-loaded-URL behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19c09b6aaa01de365c2b75415673c4e1a60260c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/lint-baseline.xml
+++ b/app/lint-baseline.xml
@@ -883,6 +883,17 @@
 
     <issue
         id="NoImplImportsInAppModule"
+        message="Modules should not import from :*-impl modules.&#xA;Consider using the public API instead: com.duckduckgo.duckchat.impl.ui.nativeinput.views.NativeInputWidget&#xA;If the API doesn&apos;t expose what you need, extend the API module first."
+        errorLine1="import com.duckduckgo.duckchat.impl.ui.nativeinput.views.NativeInputWidget"
+        errorLine2="~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="src/test/java/com/duckduckgo/app/browser/nativeinput/RealNativeInputManagerTest.kt"
+            line="45"
+            column="1"/>
+    </issue>
+
+    <issue
+        id="NoImplImportsInAppModule"
         message="Modules should not import from :*-impl modules.&#xA;Consider using the public API instead: com.duckduckgo.duckchat.impl.wideevents.InputScreenOnboardingWideEvent&#xA;If the API doesn&apos;t expose what you need, extend the API module first."
         errorLine1="import com.duckduckgo.duckchat.impl.wideevents.InputScreenOnboardingWideEvent"
         errorLine2="~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
--- a/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
@@ -188,7 +188,7 @@ class RealNativeInputManager @Inject constructor(
     private val duckChatInputModeState: DuckChatInputModeState,
     private val pixel: Pixel,
     private val nativeInputStateBugKillSwitch: NativeInputStateBugKillSwitch,
-    private val nativeInputUrlClearingKillSwitch: NativeInputUrlClearingKillSwitch,
+    private val nativeInputUrlClearingFeature: NativeInputUrlClearingFeature,
     private val nativeInputOmnibarFeature: NativeInputOmnibarFeature,
     private val nativeInputEventListener: NativeInputEventListener,
     private val edgeToEdgeProvider: EdgeToEdgeProvider,
@@ -977,7 +977,7 @@ class RealNativeInputManager @Inject constructor(
     }
 
     internal fun bindUrlCaching(widgetView: View) {
-        if (!nativeInputUrlClearingKillSwitch.self().isEnabled() || omnibarController.isDuckAiMode()) return
+        if (!nativeInputUrlClearingFeature.self().isEnabled() || omnibarController.isDuckAiMode()) return
         val widget = widgetFrom(widgetView) ?: return
         val text = widget.text
 
--- a/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
@@ -188,6 +188,7 @@ class RealNativeInputManager @Inject constructor(
     private val duckChatInputModeState: DuckChatInputModeState,
     private val pixel: Pixel,
     private val nativeInputStateBugKillSwitch: NativeInputStateBugKillSwitch,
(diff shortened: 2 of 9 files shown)


joshliebe:
* **#9545** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9545?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputManager.kt
```diff
@@ -971,6 +976,39 @@ class RealNativeInputManager @Inject constructor(
...
+        widget.onChatSelected = { animate ->
+            cacheUrl(widget)
+            onChatSelected?.invoke(animate)
+        }
+
+        val onSearchSelected = widget.onSearchSelected
+        widget.onSearchSelected = { animate ->
+            restoreUrl(widget)
+            onSearchSelected?.invoke(animate)
+        }
```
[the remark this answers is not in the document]
joshliebe: I don’t think this is an issue

app/lint-baseline.xml
```diff
@@ -881,6 +881,17 @@
...
+    <issue
+        id="NoImplImportsInAppModule"
+        message="Modules should not import from :*-impl modules.&#xA;Consider using the public API instead: com.duckduckgo.duckchat.impl.ui.nativeinput.views.NativeInputWidget&#xA;If the API doesn&apos;t expose what you need, extend the API module first."
+        errorLine1="import com.duckduckgo.duckchat.impl.ui.nativeinput.views.NativeInputWidget"
+        errorLine2="~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="src/test/java/com/duckduckgo/app/browser/nativeinput/RealNativeInputManagerTest.kt"
+            line="45"
+            column="1"/>
+    </issue>
```
joshliebe, line 893: Added this so I can import the widget into the test so I can test interactions with it. It is already imported into the manager, so I think this is fine. When the import is removed from the manager, we can also remove it from the test.

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputUrlClearingFeature.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.duckduckgo.feature.toggles.api.Toggle.DefaultFeatureValue
+
+@ContributesRemoteFeature(
+    scope = AppScope::class,
+    featureName = "nativeInputUrlClearingKillSwitch",
+)
+interface NativeInputUrlClearingKillSwitch {
+
+    @Toggle.DefaultValue(DefaultFeatureValue.INTERNAL)
+    fun self(): Toggle
```
malmstein, line 31: `INTERNAL` is the right default for staging this, but every other `*KillSwitch` in the repo defaults `TRUE` (`NativeInputStateBugKillSwitch`, and `OmnibarPreFillKillSwitch` in this same PR), so the name implies default-on with a remote off. can we keep the `INTERNAL` default and rename to something like `NativeInputUrlClearingFeature`, so it reads as the rollout flag it is?

malmstein (approved):
Installed the APK on the attached device.

Both behaviours are cleanly gated and the polarity is right in each case. `OmnibarPreFillKillSwitch` disabled restores the exact previous expression, so the kill path is a true revert. The pre-fill logic is also genuinely better: comparing `omnibarText` against the displayed url (short or full, via `isFullUrlEnabled`) means a user who typed something keeps their text instead of having the url substituted underneath them. On the clearing side, capturing the bind-time text and only caching when the widget is not dirty is the right guard, and `restoreUrl` only restores into an empty field so it cannot clobber typing. `RealNativeInputManager` is `FragmentScope`, so `cachedUrl` is per tab rather than shared, and widgets are recreated per `showNativeInput` so the handler wrapping cannot stack.

On the lint baseline: the new entry is a test file importing `NativeInputWidget`, and that file's production counterpart already imports it. The baseline carries 204 `NoImplImportsInAppModule` entries, 137 of them tests, so this follows the established pattern rather than hiding something new.\n\nOne naming suggestion on the new flag, inline.

app/src/main/java/com/duckduckgo/app/browser/nativeinput/NativeInputUrlClearingFeature.kt
joshliebe: Done: https://github.com/duckduckgo/Android/pull/9545/commits/19c09b6aaa01de365c2b75415673c4e1a60260c2

PR #9547, opened 2026-08-20
Implement HostnameResolver for the Suggest redirect feature

juandiana:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1216807998862658/task/1217615568331049?focus=true
Tech Design URL: https://app.asana.com/1/137249556945/project/1216807998862658/task/1217494645371352?focus=true

### Description
This PR adds the `HostnameResolver` implementation for the Suggest redirect feature, together with unit tests for it and integration tests for the two `DnsLookup` implementations (one for API 29+ and one for API 28), which are dependencies of the main component.

The `HostnameResolver` component is wired up to DI but is not yet used in the app. The third PR, which culminates this feature's implementation will make use of it.

### UI changes
N/A

---

> [!NOTE]
> **Medium Risk**
> Introduces DNS lookups against the active network and relocates the app-scoped ConnectivityManager binding. The resolver is unused in product code, so runtime impact is limited until a follow-up.
> 
> **Overview**
> Adds a `HostnameResolver` that reports whether a hostname resolves on the active network within 2s, for the upcoming Suggest redirect feature. It is wired in DI but not used yet.
> 
> Lookups skip IP literals and invalid domains. API 29+ uses `DnsResolver` with cancellation; older APIs use `Network.getAllByName`. `ConnectivityManager` is now provided from `SystemComponentsModule` instead of the VPN module.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5543f1ada581d6d18be737086b3ff30afbf628d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
@@ -367,7 +367,7 @@ class BrowserModule {
     ): HostnameResolver {
         val lookupTimeout = HostnameResolver.LOOKUP_TIMEOUT_MS.milliseconds
         val dnsLookup = if (appBuildConfig.sdkInt >= 29) {
-            @Suppress("NewApi")
+            @Suppress("NewApi") // we use appBuildConfig
             DnsLookupApi29Impl(dispatcherProvider, DnsResolver.getInstance())
         } else {
             DnsLookupPreApi29Impl(dispatcherProvider)
--- a/app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
@@ -19,6 +19,8 @@ package com.duckduckgo.app.browser.di
 import android.content.ClipboardManager
 import android.content.Context
 import android.content.pm.PackageManager
+import android.net.ConnectivityManager
+import android.net.DnsResolver
 import androidx.datastore.core.DataStore
 import androidx.datastore.preferences.core.Preferences
 import androidx.datastore.preferences.preferencesDataStore
@@ -49,6 +51,10 @@ import com.duckduckgo.app.browser.menu.BrowserMenuHighlightPlugin
 import com.duckduckgo.app.browser.menu.TopInContextSection
 import com.duckduckgo.app.browser.pageloadpixel.PageLoadedPixelDao
 import com.duckduckgo.app.browser.pageloadpixel.firstpaint.PagePaintedPixelDao
+import com.duckduckgo.app.browser.suggestredirect.DnsLookupApi29Impl
+import com.duckduckgo.app.browser.suggestredirect.DnsLookupPreApi29Impl
+import com.duckduckgo.app.browser.suggestredirect.HostnameResolver
+import com.duckduckgo.app.browser.suggestredirect.HostnameResolverImpl
 import com.duckduckgo.app.browser.tabpreview.FileBasedWebViewPreviewGenerator
 import com.duckduckgo.app.browser.tabpreview.FileBasedWebViewPreviewPersister
 import com.duckduckgo.app.browser.tabpreview.WebViewPreviewGenerator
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
```diff
@@ -349,6 +357,23 @@ class BrowserModule {
     fun provideSiteErrorCodeHandler(): HttpCodeSiteErrorHandler {
         return HttpCodeSiteErrorHandlerImpl()
     }
+
+    @SingleInstanceIn(AppScope::class)
+    @Provides
+    fun provideHostnameResolver(
+        appBuildConfig: AppBuildConfig,
+        connectivityManager: ConnectivityManager,
```
malmstein, line 365: q: the only ConnectivityManager provider in the graph is VpnAppModule.providesConnectivityManager in :vpn-impl, so browser code now leans on a binding the AppTP module owns. works today since :app depends on vpn-impl, but should we provide it here instead?

app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
```diff
@@ -349,6 +357,23 @@ class BrowserModule {
...
+        dispatcherProvider: DispatcherProvider,
+    ): HostnameResolver {
+        val lookupTimeout = HostnameResolver.LOOKUP_TIMEOUT_MS.milliseconds
+        val dnsLookup = if (appBuildConfig.sdkInt >= 29) {
+            @Suppress("NewApi")
```
malmstein: nit: the other five @Suppress("NewApi") in :app carry // we use appBuildConfig so the next reader knows why lint can't see the gate. worth adding the same here.

app/src/main/java/com/duckduckgo/app/browser/suggestredirect/HostnameResolver.kt
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.browser.suggestredirect
+
+import android.net.ConnectivityManager
+import com.duckduckgo.app.browser.suggestredirect.DnsLookup.LookupResult.Failure
+import com.duckduckgo.app.browser.suggestredirect.DnsLookup.LookupResult.Success
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.wireguard.config.InetAddresses
```
malmstein: q: no build file in the repo declares wireguard, so this class is on :app's classpath transitively, and the only other users are network-protection. what made InetAddresses.isHostname the right call over a local check? asking mainly because a netP dependency change would break the browser build.

malmstein (approved):
two non-blocking questions below worth a reply: where ConnectivityManager comes from, and the wireguard import in :app. the rest is nits you can take or leave.

app/src/main/java/com/duckduckgo/app/browser/suggestredirect/HostnameResolver.kt
juandiana: Correct.

> Should we move this utility class to a common module (I believe this requires an API proposal though) or rather implement my own utility class for this? I figure the latter would be easier, but we may want to be consistent across the app on aspects such as hostname validation.

It’s mostly a question of what we value more. If we don’t feel so strong about how we perform hostname validation across the codebase, I’m happy to have a local implementation on the browser module for this feature. Let know and I’ll do the change.

app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
juandiana: That was my question as well :)

`ConnectivityManager` cannot be provided by both modules (`BrowserModule` and `VpnAppModule`). Dagger will detect the duplicate provision and fail the build. My question was what do you generally do in these cases in this codebase? A reasonable approach would be to find a common place to put it where both modules can access it.

A good candidate for this seems to be `SystemComponentsModule`. It also exposes a similar system service `LocationManager`. If you agree, I’ll move the `ConnectivityManager` provide method from `VpnAppModule` to `SystemComponentsModule`. I’ll reach out to the maintainers of that feature on how to verify the change, just in case.

juandiana: Done

app/src/main/java/com/duckduckgo/app/browser/di/BrowserModule.kt
juandiana: Done

app/src/main/java/com/duckduckgo/app/browser/suggestredirect/HostnameResolver.kt
juandiana: Done. I actually found `PatternsCompat.DOMAIN_NAME` & `PatternsCompat.IP_ADDRESS` from `androidx.core.util`, which I believe are a much better solution.

PR #9549, opened 2026-08-20
Pin GpcTest interactions to the tab's WebView

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1217673990175146
Tech Design URL (if applicable):
API Proposals URL(s) (if applicable): None

### Description

`GpcTest.whenProtectionsAreEnableGpcSetCorrectly` fails with `AmbiguousViewMatcherException`: the matcher `an instance of android.webkit.WebView and webView.getSettings().getJavaScriptEnabled() is <true>` matched two `DuckDuckGoWebView`s, both with `res-name=browserWebView` and both visible. See the failure in [this run](https://github.com/duckduckgo/Android/actions/runs/32133443253).

Same root cause as #9534. The `top frame header` case on `privacy-test-pages.site/privacy-protections/gpc/` calls `window.open(...)`, and `BrowserChromeClient.onCreateWindow` forces `isGesture = true` under instrumentation, so the popup always becomes a real tab. Its fragment sits in the tab pager alongside the original one, and Espresso's default web matcher selects purely on type and id, so whenever it resolved inside that window the test failed.

Both interactions now match the `WebView` instance the test already captured, which is the one holding the page's `results` object, and the idling resources are built from that same instance. How many tabs exist at that moment no longer matters.

`HttpsUpgradesTest` needed no change, #9534 already covers it.

### Steps to test this PR

_GpcTest passes_
- [x] Check the Privacy Tests workflow on this PR is green (it runs the `@PrivacyTest` suite, including this test)
- [x] Optionally, run it locally against an API 34 emulator, since Espresso 3.6.1 cannot inject events on API 36+: `./gradlew :app:connectedPlayDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.duckduckgo.espresso.privacy.GpcTest`
- [x] While the test runs, note the second tab that appears when the test page opens its popup, and that the assertions still read the original tab

### UI changes
| Before  | After |
| ------ | ----- |
| No UI changes | No UI changes |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Instrumentation-test-only change; no production, auth, or data-handling code is modified.
> 
> **Overview**
> Fixes flaky `GpcTest.whenProtectionsAreEnableGpcSetCorrectly` when the GPC test page’s `window.open` creates a second tab.
> 
[...]

PR #9534, opened 2026-08-19
Pin HttpsUpgradesTest interactions to the tab's WebView

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1217636326375070
Tech Design URL (if applicable):
API Proposals URL(s) (if applicable): None

### Description

`HttpsUpgradesTest.whenProtectionsAreEnabledHttpsUpgradedCorrectly` has been failing with `AmbiguousViewMatcherException`: the matcher `an instance of android.webkit.WebView and webView.getSettings().getJavaScriptEnabled() is <true>` matched two `DuckDuckGoWebView`s, both with `res-name=browserWebView` and both visible.

The second WebView belongs to a second tab. The `upgrade-navigation` case on `privacy-test-pages.site/privacy-protections/https-upgrades/` calls `window.open(...)`, and `BrowserChromeClient.onCreateWindow` forces `isGesture = true` under instrumentation, so the popup always becomes a real tab. Its fragment sits in the tab pager alongside the original one until the page calls `otherWindow.close()`, and Espresso's default web matcher selects purely on type and id, so whenever it resolved inside that window the test failed. Nothing changed recently on either side to cause this: the `window.open` call has been on the test page since it was created in 2020 and the tab pager has been in place since early 2025, so this is a latent race whose timing on the test runner has shifted.

Both interactions now match the `WebView` instance the test already captured and registers its idling resources against, which is the one holding the page's `results` object. How many tabs exist at that moment no longer matters.

### Steps to test this PR

_HttpsUpgradesTest passes_
- [x] Check the Privacy Tests workflow on this PR is green (it runs the `@PrivacyTest` suite, including this test)
- [x] Optionally, run it locally against an API 34 emulator, since Espresso 3.6.1 cannot inject events on API 36+: `./gradlew :app:connectedPlayDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.duckduckgo.espresso.privacy.HttpsUpgradesTest`
- [x] While the test runs, note the second tab that appears when the test page opens its popup, and that the assertions still read the original tab

### UI changes
| Before  | After |
| ------ | ----- |
| No UI changes | No UI changes |

---

> [!NOTE]
> **Low Risk**
> Android instrumentation test change only; no production code or security-sensitive paths affected.
> 
> **Overview**
[...]

PR #9554, opened 2026-08-20
Wait for async results in HttpsUpgradesTest

malmstein:
Task/Issue URL: https://app.asana.com/1/137249556945/project/414730916066338/task/1217673990175146

> [!NOTE]
> Stacked on #9549. Review that one first, this PR targets its branch so a single Privacy Tests run covers both fixes.

### Description

`HttpsUpgradesTest.whenProtectionsAreEnabledHttpsUpgradedCorrectly` fails with `java.lang.AssertionError: Url for upgrade-navigation should be https`. Seen in [this run](https://github.com/duckduckgo/Android/actions/runs/32343711433).

This is a different bug from #9534 and #9549, which were both about `AmbiguousViewMatcherException`. Now that the matcher is pinned, the test no longer throws when the popup tab is open, so it proceeds and reads results that are not ready yet.

The only gate between the click and the read was `WebViewIdlingResource`, which is idle once `webView.progress == 100`. The top frame never navigates while the page's tests run, so progress was already 100 and that gate opened immediately. Meanwhile the page fills each entry asynchronously: `runTests()` pushes `{id, value}` up front and `upgrade-navigation` only resolves once the popup it opens postMessages its URL back (retried on a 500ms interval). Reading before that gives no value, and `it.value?.toUri()?.isHttps ?: false` is then false.

Evidence that the results are still being filled after the gate opens, logged from an instrumented local run 10s after the click:

```
{"id":"upgrade-navigation","value":"https://good.third-party.site/..."},
{"id":"upgrade-iframe","value":"http://..."},
{"id":"upgrade-subrequest","value":"http://..."},
{"id":"upgrade-websocket"}   <- still no value
```

Whether the read wins the race is pure timing, which is why this test passed, flaked and failed on consecutive CI runs with no change on either side.

This PR gates the read on the asserted results being populated. Two properties worth noting:

- a truthy but `http` value still satisfies the condition, so a genuine https-upgrade regression fails the assertion exactly as it does today rather than being masked by the wait
- a result that never arrives now fails naming the unmet condition, instead of the misleading "should be https"

`JsObjectIdlingResource` takes the condition as an optional argument defaulting to its current object-exists check, so `GpcTest`, `FingerprintProtectionTest` and `RequestBlocklistTest` are unaffected.

The condition uses the bare `results` identifier rather than `window.results` because the page declares it with `const`, which creates a global binding and not a window property.

### Steps to test this PR

_HttpsUpgradesTest passes_
- [x] Check the Privacy Tests workflow on this PR is green (it runs the `@PrivacyTest` suite, including this test and GpcTest from #9549)
- [x] Optionally, run it locally against an API 34 emulator, since Espresso 3.6.1 cannot inject events on API 36+: `./gradlew :app:connectedPlayDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.duckduckgo.espresso.privacy.HttpsUpgradesTest`

Note on local verification: the race does not reproduce on a local emulator. The host is fast enough, and `frame.html` is served from the WebView cache, that the handshake always completes inside the Espresso click round-trip, so the read is already populated and the new gate is a no-op of a few ms. Network throttling does not widen the window for the same caching reason. Local runs therefore only confirm the fix is green and cheap when timing is good; the protection itself shows on the slower CI runner.

### UI changes
| Before  | After |
| ------ | ----- |

---

> [!NOTE]
> **Low Risk**
> Instrumented-test-only change; no production, auth, or data-handling code is modified.
> 
> **Overview**
> Stops `HttpsUpgradesTest` from reading results as soon as the page hits 100% progress. The `upgrade-navigation` case only fills in after a popup posts its URL back, so the old `WebViewIdlingResource` gate often opened too early and the https assertion failed on empty values.
> 
> `JsObjectIdlingResource` now accepts an optional JS `condition` (default still “object exists”). The test waits until `results` has a truthy value for the asserted ids, so a real http upgrade still fails the assertion rather than being hidden by the wait. Other privacy tests keep the default object-exists check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42193ef179a1804feb799227e231c9d422d965b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/androidTest/java/com/duckduckgo/espresso/JsObjectIdlingResource.kt
+++ b/app/src/androidTest/java/com/duckduckgo/espresso/JsObjectIdlingResource.kt
@@ -28,6 +28,8 @@ class JsObjectIdlingResource(
     private val objectName: String,
     private val timeoutMillis: Long = DEFAULT_TIMEOUT_MILLIS,
     private val failOnTimeout: Boolean = true,
+    // Defaults to waiting for the object to exist; pass an expression to wait for a specific state instead
+    private val condition: String = "typeof $objectName !== 'undefined'",
 ) : IdlingResource {
 
     @Volatile
@@ -54,15 +56,15 @@ class JsObjectIdlingResource(
             isIdle = true
             callback?.onTransitionToIdle()
             if (failOnTimeout) {
-                // fail test if the object is not found within the timeout if opted-in
-                throw AssertionError("JS object '$objectName' did not appear within ${timeoutMillis}ms.")
+                // fail test if the condition is not met within the timeout if opted-in
+                throw AssertionError("JS condition '$condition' was not met within ${timeoutMillis}ms.")
             }
-            logcat { "JsObjectIdlingResource: '$objectName' did not appear within ${timeoutMillis}ms; proceeding" }
+            logcat { "JsObjectIdlingResource: '$condition' was not met within ${timeoutMillis}ms; proceeding" }
             return
         }
 
         webView.evaluateJavascript(
-            "(typeof $objectName !== 'undefined')",
+            "($condition)",
         ) { result ->
             if (result == "true") {
                 isIdle = true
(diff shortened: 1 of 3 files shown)


joshliebe (approved):
Ran the test locally, passing 👍

PR #9574, opened 2026-08-21
Quick Win: Android: Improve PageLoadWideEvent reporting to recover 21.6% unknown events

anikiki:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1200581511062568/task/1217240546184827?focus=true

### Description

The page load wide event was ending a large share of flows at the five-minute cleanup as `Unknown`instead of at a real outcome. Five separate causes:

- **Flows are keyed by** **`navigationId`, not url.** A redirect changed the url under the flow, so its finish no longer matched and was dropped.
- **A start for an untracked url now ends the tab's open flow**, instead of leaving it to the cleanup policy.
- **Every main-frame error reports a failure.** Reporting was gated on the error having a user-facing error screen, so most failures were indistinguishable from abandoned loads.
- **The measured load ends independently of the page load cycle**, so a terminal callback for an already-replaced load no longer orphans the flow that replaced it.
- **The fixed-progress gate resets per load**, closing its interval on same-host redirects (which compare as `UrlUpdated`, so the previous url-driven reset never ran).

Failures are matched strictly on the load's url, because marking a load failed also suppresses its finish — an error reported against a different url is declined rather than guessed at.

No pixel definition change: `error_code` is already a free-form string documented as any WebViewClient error constant.

**Known limitation, deliberate.** Every main-frame page start opens its own measured load, including a redirect hop, so a redirecting load is measured from its last hop: `elapsed_time_to_finish` excludes the redirect leg the user also waited through, and aborts count hops rather than loads the user abandoned. Confirmed on device — `reddit.com/` → `reddit.com/?rdt=…` drops 505 ms. Spanning a chain with one flow is possible via `WebResourceRequest.isRedirect`, but it would move the content-scope and JS-injection steps onto the document the redirect left behind and change which urls are sampled. Both are decisions about what the metric should mean, so they are out of scope here.

### Steps to test this PR

**Setup - IMPORTANT**:

- [x]  Replace the current`isFeatureEnabled()` function in `PageLoadWideEvent.kt` with `private suspend fun isFeatureEnabled(): Boolean = true `so you can see events.
- [x] Only sites in `PageLoadedSites.perfSites` are tracked (`bbc.co.uk`, `reddit.com`, `foxnews.com`, `amazon.com`, `apnews.com`, …). Any other domain logs nothing.
- [x] Use this grep for easy checks:
`adb logcat -v time | grep -E ￼'Page load measured as navigationId|Page load flow started|Cancelling previous flow|Page visible recorded|Exited max progress threshold|Recorded elapsed_time|Page load finished|Ignoring repeat|Ignoring load failure|Dropping event from navigation|No active flow found|Failed to start page load￼flow'`

_Successful load reaches a terminal outcome_

- [x] Load https://www.bbc.com/news. Expect `Page load measured as navigationId [...]` `Page load finished [...]`  Flow id should match for the above logs. If https://www.bbc.com/news redirects to https://www.bbc.co.uk/news you should see 2 events, one for each page loaded.
- [x]  Check the wide event pixel as well. Open another logcat console and filter by `wide_page-load_c`

_Multiple_ _page_ _loads_, where one is cancelled should not record the cancelled one

- [x]  Load https://www.reddit.com/  and, before it finishes, navigate away to ebay.com in the same tab. Let ebay.com finish. Within 5 minutes, load reddit.com again in that same tab.
- [x]  Expect to see Page load finished only for ebay.com and for the second load of reddit. The fist one should be cancelled.
- [x]  Check the wide event pixel as well. Open another logcat console and filter by wide_page-load_c. A wide event should not be sent for the cancelled page load.

_Cancelling a load_
- [ ] Load bbc.com/news and before it finished loading load foxnews.com
- [ ] Notice 2 wide event pixels containing:
`wide_page-load_c with params: {... feature.status=CANCELLED, ..., feature.data.ext.cancellation_reason=superseded_different_url, ...}`
`wide_page-load_c with params: {... feature.status=SUCCESS, ...}`

- [ ] Load bbc.com/news and before it finished loading do a pull to refresh 
- [ ] Notice 2 wide event pixels containing:
`wide_page-load_c with params: {... feature.status=CANCELLED, ..., feature.data.ext.cancellation_reason=superseded_same_url, ...}`
`wide_page-load_c with params: {... feature.status=SUCCESS, ...}`

### NO UI changes

---

> [!NOTE]
> **Medium Risk**
> Telemetry-only, but it rewires page-load flow identity, cancellation, and error matching in WebView navigation. Mis-attribution would distort success/failure/cancel rates without affecting browsing itself.
> 
> **Overview**
> Stops page-load wide events from aging out as **Unknown** by identifying each measured load with `navigationId` instead of URL, so redirects, replacements, and late callbacks no longer steal or drop another load’s events.
> 
> `BrowserWebViewClient` now opens a new measured load on every main-frame `onPageStarted` (including redirect hops), ends it independently of the page-load cycle, and reports **every** main-frame error as a failure when the URL matches the load being measured. Superseded in-flight flows finish as **Cancelled** with `cancellation_reason` (`superseded_same_url` / `superseded_different_url`), including when the replacement URL is untracked. Progress-threshold tracking resets per load via `onMainFrameLoadStarted`.
> 
> Pixel definition patch 0.1 documents the new cancellation field and that finish-only metadata is absent on cancelled loads. Redirect chains are still measured from the last hop by design.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0435b9645ebb32549e52fc91d5a58cfdbc53ed38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
@@ -149,6 +149,14 @@ class BrowserWebViewClient @Inject constructor(
     // WebView clears hasGesture() on redirect hops, but the App Link rules assume Chromium's per-navigation gesture flag.
     private var mainFrameGestureOriginUrl: String? = null
     private var start: Long? = null
+
+    // Needed for PageLoadWideEvent: it identifies the page load the wide event is currently measuring,
+    // so every callback reports against the flow that load opened and no later one can claim them.
+    private var navigationId: Long? = null
+
+    // Needed for PageLoadWideEvent: it identifies the url the measured load started with, so a main frame error
+    // can be told apart from one reported against some other url.
+    private var navigationUrl: String? = null
     private var lastInterceptedAppSchemeUrl: String? = null
     private var pageCommitVisibleFired: Boolean = false
     private var recompositeScheduled: Boolean = false
@@ -177,6 +185,7 @@ class BrowserWebViewClient @Inject constructor(
 
     private var currentLoadOperationId: String? = null
     private var parallelRequestsOnStart = 0
+    private var parallelRequestsOnMeasuredLoadStart = 0
 
     init {
         appCoroutineScope.launch {
@@ -496,8 +505,10 @@ class BrowserWebViewClient @Inject constructor(
         if (webView.url == url) {
             val navigationList = webView.safeCopyBackForwardList() ?: return
             webViewClientListener?.onPageCommitVisible(WebViewNavigationState(navigationList), url)
-            webViewClientListener?.getCurrentTabId()?.let { tabId ->
-                pageLoadWideEvent.onPageVisible(tabId, url, webView.progress)
+            navigationId?.let { loadNavigationId ->
+                webViewClientListener?.getCurrentTabId()?.let { tabId ->
+                    pageLoadWideEvent.onPageVisible(tabId, loadNavigationId, webView.progress)
+                }
             }
         }
     }
@@ -546,15 +557,15 @@ class BrowserWebViewClient @Inject constructor(
         var wideEventNavigation: Pair<String, Long>? = null
         url?.let {
             // See https://app.asana.com/0/0/1206159443951489/f (WebView limitations)
-            if (it != ABOUT_BLANK && start == null) {
-                start = currentTimeProvider.elapsedRealtime()
-                webViewClientListener?.getCurrentTabId()?.let { tabId ->
-                    val navigationId = wideEventNavigationIdCounter.incrementAndGet()
-                    pageLoadWideEvent.onPageStarted(tabId, it, navigationId)
-                    wideEventNavigation = tabId to navigationId
+            if (it != ABOUT_BLANK) {
+                if (start == null) {
+                    start = currentTimeProvider.elapsedRealtime()
+                    incrementAndTrackLoad() // increment the request counter
+                    requestInterceptor.onPageStarted(url)
                 }
-                incrementAndTrackLoad() // increment the request counter
-                requestInterceptor.onPageStarted(url)
+                // A page start arriving while another load is in flight opens its own measured load, so a redirect
+                // chain is measured from its last hop.
+                wideEventNavigation = startWideEventPageLoad(it)
             }
 
             handleMediaPlayback(webView, it)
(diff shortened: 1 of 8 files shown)


anikiki:
* **#9574** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9574?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9574?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -138,20 +140,18 @@ class RealPageLoadWideEvent @Inject constructor(
...
-                if (isInProgress(tabId, url)) return@launch
 
-                val existingState = activeFlows[tabId]
-                if (existingState != null && existingState.url != url) {
-                    logcat { "Cancelling previous flow for tabId=$tabId (${existingState.url} → $url)" }
-                    activeFlows.remove(tabId)
-                    wideEventClient.flowAbort(existingState.flowId)
+                activeFlows.remove(tabId)?.let { previous ->
+                    logcat { "Cancelling previous flow for tabId=$tabId, flowId=${previous.flowId} (${previous.url} → $url)" }
+                    wideEventClient.flowAbort(previous.flowId)
```
GerardPaligot: This change can have a significant impact if we come from an observed page to another one. `flowAbort` sends nothing to our telemetry; what do you think about using `flowFinish(status = FlowStatus.Cancelled)` instead to capture the event, while making it clear that the page load was superseded by another one?

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -149,6 +149,14 @@ class BrowserWebViewClient @Inject constructor(
...
     private var mainFrameGestureOriginUrl: String? = null
     private var start: Long? = null
+
+    // Needed for PageLoadWideEvent: it identifies the page load the wide event is currently measuring,
+    // so every callback reports against the flow that load opened and no later one can claim them.
+    private var navigationId: Long? = null
+
+    // Needed for PageLoadWideEvent: it identifies the url the measured load started with, so a main frame error
+    // can be told apart from one reported against some other url.
+    private var navigationUrl: String? = null
```
GerardPaligot, line 159: Nit: `BrowserWebViewClient` is becoming increasingly complex (with these fields, but all related functions too). I'm pretty sure we can extract the page load logic elsewhere to handle the navigation ID/URL. What do you think about creating a dedicated task for this outside the Quick Win scope?

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
```diff
@@ -921,22 +968,15 @@ class BrowserWebViewClient @Inject constructor(
 
             val parsedError = parseErrorResponse(webResourceError)
             if (request?.isForMainFrame == true) {
+                // Reported for every main-frame error, not just ones shown to the user. Otherwise, other failures
+                // never close properly and later get marked Unknown, making failed loads look like abandoned ones.
+                reportMeasuredLoadFailed(
```
GerardPaligot, line 973: For an `OMITTED` main-frame error, `reportMeasuredLoadFailed` ends the measured load while `start` stays non-null, so `onPageFinished` still fires `pageLoadedHandler.onPageLoaded` and the pixel counts a completed load, while the wide event recorded `error`.

I don't know if the existing pixel is still used by someone but we should consider to drop it in a dedicated task.

GerardPaligot (approved):
Few optional comments but it works pretty well, I've tested additional scenarios:

1. Load a website in error state. Expectation: Cancel the page load.
2. Load the same website in the same tab before the end of the load. Expectation: Cancelling the previous one and send only the last one
3. Load the same website in different tab. Expectation: Capture all events.

Everything is working fine!

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
anikiki: Good point 👍  Created https://app.asana.com/1/137249556945/project/1200905986587319/task/1217779171531443?focus=true

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
anikiki: Makes sense. Updated and added a reason for the cancellation as cancellations due to reloading the same URL might signal a page load issue.

app/src/main/java/com/duckduckgo/app/browser/BrowserWebViewClient.kt
anikiki: Before this change the wide event never reported the OMITTED failure at all and aged came as Unknown. The existing pixel behaved identically. We'll be looking into the existing pixel to see if it's still needed.

app/src/main/java/com/duckduckgo/app/browser/pageload/PageLoadWideEvent.kt
```diff
@@ -353,37 +351,40 @@ class RealPageLoadWideEvent @Inject constructor(
...
+            return null
         }
-        return state?.flowId
+        if (state.isStale()) return null
+        return state
     }
+    private suspend fun isFeatureEnabled(): Boolean = true
+    //     private suspend fun isFeatureEnabled(): Boolean = withContext(dispatchers.io()) {
+    //     androidBrowserConfigFeature.get().sendPageLoadWideEvent().isEnabled()
+    // }
```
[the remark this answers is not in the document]
anikiki: Updated 😓

PR #9580, opened 2026-08-21
Add UI and logic for displaying redirect suggestion on the BAD_URL error page

juandiana:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1216807998862658/task/1217666997020568?focus=true
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/1216807998862658/task/1217494645371353?focus=true

### Description
This PR completes the changes for displaying a redirect suggestion on `BAD_URL` error page. The feature is behind a feature flag, which is disabled by default at the moment.

### Steps to test this PR
1. Make sure the feature flag `suggestRedirct` is enabled.
2. Perform the following steps using both a normal tab and a custom tab:
    1. Navigate to `airnow.gov/test`
    2. Verify that the redirect suggestion is displayed with only `www.airnow.gov` underlined
    3. Tap on the underlined link
    4. Verify that the browser navigated to `www.airnow.gov/test`, loading successfully (minus the expected page not found)
3. On a normal tab:
    1. Try to navigate to `airnow.gov`
    2. Verify that the `BAD_URL` error page together with the redirect suggestion is displayed
    3. Turn on airplane mode
    4. Tap on the redirect link
    5. Verify that the offline error page is displayed and that the redirect suggestion is no longer displayed
    6. Turn off airplane mode
    7. Try to navigate again to the current page, which should be `www.airnow.gov`
    8. Verify that it loads successfully

### UI changes
https://github.com/user-attachments/assets/f39eee57-aede-409d-b807-a065a091f9f8

---

> [!NOTE]
> **Medium Risk**
> Changes error-page navigation and async host lookups on failed loads; risk is moderate because mis-suggestions or stale async results could send users to the wrong host, though feature-flag gating and job cancellation limit exposure.
> 
> **Overview**
> When a page fails with a **BAD_URL** (host not found) error and the feature flag is on, the browser now evaluates whether a `www.` variant of the bare registrable domain resolves and, if so, shows a **“Did you mean …?”** line on the error UI with a tappable link to that URL (path, query, port preserved).
> 
> **BrowserTabViewModel** drives this via `redirectSuggestion` on `BrowserViewState`, async evaluation through `SuggestRedirectEvaluator`, and clears or cancels in-flight work when the user navigates, goes home, resets/refreshes the error, or gets a different error type. Tapping the link clears the error state and issues navigation to the suggested URL.
> 
> **BrowserTabFragment** renders the new `redirect_suggestion_message` view and reacts when the suggestion changes while an error is shown. **SuggestRedirectEvaluator** encapsulates the www-prefix + DNS lookup rules. Unit tests cover the evaluator and view-model gating, race cancellation, and click handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb16d8f16afc4c7a591e05198c0df755397d76ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -4568,7 +4568,10 @@ class BrowserTabViewModel @Inject constructor(
             command.value = WebViewError(errorType, url)
             suggestRedirectJob.cancel() // Cancel previous in-flight job as the new errorType might not be BAD_URL
         }
-        if (errorType == BAD_URL && suggestRedirectOnUnresolvedErrorFeature.suggestRedirect().isEnabled()) {
+        if (errorType == BAD_URL &&
+            suggestRedirectOnUnresolvedErrorFeature.self().isEnabled() &&
+            suggestRedirectOnUnresolvedErrorFeature.suggestRedirect().isEnabled()
+        ) {
             suggestRedirectJob += viewModelScope.launch {
                 suggestRedirectEvaluator.suggestRedirect(url)?.let { suggestion ->
                     browserViewState.value = currentBrowserViewState().copy(redirectSuggestion = suggestion)
--- a/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
+++ b/app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
@@ -5797,6 +5797,11 @@ class BrowserTabViewModel @Inject constructor(
         }
     }
 
+    fun onRedirectSuggestionClicked(url: String) {
+        resetBrowserError()
+        command.value = NavigationCommand.Navigate(url, getUrlHeaders(url))
+    }
+
     private fun trackersCount(): String =
         siteLiveData.value?.trackerCount?.takeIf { it > 0 }?.toString() ?: ""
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -4548,12 +4559,21 @@ class BrowserTabViewModel @Inject constructor(
...
+                    redirectSuggestion = null,
                     showPrivacyShield = HighlightableButton.Visible(enabled = false),
                 )
             if (androidBrowserConfig.errorPagePixel().isEnabled()) {
                 pixel.enqueueFire(AppPixelName.ERROR_PAGE_SHOWN)
             }
             command.value = WebViewError(errorType, url)
+            suggestRedirectJob.cancel() // Cancel previous in-flight job as the new errorType might not be BAD_URL
+        }
+        if (errorType == BAD_URL && suggestRedirectOnUnresolvedErrorFeature.suggestRedirect().isEnabled()) {
```
malmstein: suggestion: this only reads suggestRedirect(), so the parent self() toggle is inert. isEnabled reads its own state only (FeatureToggles.kt:509), which is why the other 30 call sites in the repo do self().isEnabled() && sub().isEnabled(). without it we can't kill the feature from the parent switch.

malmstein (approved):
lgtm, one thing worth fixing before rollout: the parent self() toggle is inert here, so the feature can't be killed from the parent switch.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
juandiana: Good catch, fixed and tested it.

PR #9590, opened 2026-08-24
Refactor Exchange v2 envelope handling

MiSikora:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1216103556496795/task/1217777614176413
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/project/1216103556496795/task/1217626160385689
API Proposals URL(s) (if applicable): N/A

### Description

Prepares the Exchange v2 envelope for v2.1 by making the envelope carry the version required by the message it wraps, instead of a hard-coded "2". This is groundwork for sending messages that need a newer minor version, and there is no change to the pairing flow a user sees today.

- The v2 channel and envelope now take a typed `ExchangeV2Message` instead of a raw JSON string, so the message being sent is the same object all the way down to encryption.
- The version written into an outgoing envelope comes from the message's own `protocolVersion`, which is `2` for every message the app sends today, so codes and messages stay compatible with existing devices.
- Inbound envelopes are parsed with the shared `ExchangeProtocolVersion` type. A version with a higher major than we support is still rejected as too new, a v1 envelope is still rejected as obsolete, and a version that cannot be parsed at all is still rejected as malformed.
- The confirmation messages sent during a session (`recovery_code_awaiting_confirmation`, `recovery_code_confirmed`, `recovery_code_denied`) are built as typed messages rather than inline JSON literals, and go on the wire unchanged.
- The `OUR_MAJOR` and `OUR_VERSION_STRING` constants are gone, replaced by `ExchangeProtocolVersion.V2.MAJOR`.

### Steps to test this PR

_Pair with a production app, v2.1 disabled_
- [ ] Install this branch as an internal build on one device.
- [ ] Install the current Play Store release on a second device.
- [ ] On the internal build, leave `canUseExchangeV2Point1` disabled.
- [ ] On the internal build, start the flow that shows the sync QR code.
- [ ] Scan that code with the production app.
- [ ] Verify both devices end up signed in to the same sync account.
- [ ] Show the sync QR code on the production app.
- [ ] Scan that code with the internal build.
- [x] Verify pairing completes with no error.

_Pair with a production app, v2.1 enabled_
- [ ] On the internal build, enable `canUseExchangeV2Point1`.
- [ ] On the internal build, start the flow that shows the sync QR code.
- [ ] Scan that code with the production app.
- [ ] Verify both devices end up signed in to the same sync account.
- [ ] Show the sync QR code on the production app.
- [ ] Scan that code with the internal build.
- [x] Verify pairing completes with no error.

### UI changes
N/A

---

---

> [!NOTE]
> **Medium Risk**
> Touches sync pairing encryption and envelope negotiation on the relay; behavior for current v2.0 traffic should stay compatible, but mistakes in version handling could break cross-device pairing.
> 
> **Overview**
> Prepares Exchange v2 for **v2.1** by tying relay envelope `version` to each outgoing message instead of a fixed `"2"`.
> 
> **Wire path:** `ExchangeV2Channel` and `ExchangeV2Envelope` now accept **`ExchangeV2Message`** (not raw JSON). Sealing encrypts `message.rawJson` and sets `ExchangeEnvelope.version` from **`message.protocolVersion`**, so future messages can advertise a newer minor while today’s traffic still uses the versions the app already sends.
> 
> **Inbound envelopes:** `open()` parses versions via **`ExchangeProtocolVersion`**—v2 majors (including minors like `2.5`) still decrypt; v1 is obsolete; unknown higher majors still raise **`EnvelopeVersionTooNew`** with a typed version. **`OUR_MAJOR` / `OUR_VERSION_STRING`** are removed.
> 
> **Runner/tests:** `sendOnWireAndRecord` passes typed messages through; Mockito checks use message types instead of JSON substring matching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c39e7ec192445adeec4fba1c29afa75abe0cfc8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/exchange/v2/ExchangeV2Envelope.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/exchange/v2/ExchangeV2Envelope.kt
@@ -19,6 +19,7 @@ package com.duckduckgo.sync.impl.exchange.v2
 import com.duckduckgo.di.scopes.AppScope
 import com.duckduckgo.sync.impl.ExchangeEnvelope
 import com.duckduckgo.sync.impl.crypto.SyncJweCrypto
+import com.duckduckgo.sync.impl.exchange.ExchangeProtocolVersion
 import com.squareup.anvil.annotations.ContributesBinding
 import javax.inject.Inject
 
@@ -35,10 +36,10 @@ import javax.inject.Inject
 interface ExchangeV2Envelope {
 
     /**
-     * Build an outbound envelope encrypting [messageJson] to the recipient's [peerPublicKeyBase64].
+     * Build an outbound envelope encrypting [message] to the recipient's [peerPublicKeyBase64].
      * [senderChannelId] is written as the JWE `kid` header so the recipient knows the sender.
      */
-    fun seal(messageJson: String, peerPublicKeyBase64: String, senderChannelId: String): ExchangeEnvelope
+    fun seal(message: ExchangeV2Message, peerPublicKeyBase64: String, senderChannelId: String): ExchangeEnvelope
 
     /**
      * Decrypt an inbound envelope with our ephemeral private key, returning the inner message JSON.
@@ -49,8 +50,8 @@ interface ExchangeV2Envelope {
 }
 
 /** Thrown when an envelope requires a protocol version we don't support. */
-class EnvelopeVersionTooNew(val version: String) : RuntimeException(
-    "Envelope requires protocol v$version; we only support v$OUR_MAJOR",
+class EnvelopeVersionTooNew(val version: ExchangeProtocolVersion) : RuntimeException(
+    "Envelope requires protocol v$version; we only support up to v${ExchangeProtocolVersion.V2.MAJOR}",
 )
 
 /**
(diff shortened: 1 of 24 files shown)


sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/exchange/v2/ExchangeV2Envelope.kt
```diff
@@ -62,34 +63,37 @@ class EnvelopeDecryptFailure(val seq: Int, cause: Throwable) : RuntimeException(
...
+
+            is ExchangeProtocolVersion.V1 -> {
+                throw IllegalArgumentException("Obsolete envelope version $protocol")
+            }
 
-    private fun parseMajor(version: String): Int {
-        val majorPart = version.substringBefore('.')
-        return majorPart.toIntOrNull()
-            ?: throw IllegalArgumentException("Malformed envelope version: $version")
+            else -> {
```
CDRussell, line 94: nit: what about changing this to another `is` statement, so that if we add any future versions here there will be a compilation error and we'll be forced to handle it explicitly instead of falling into the `else` implicitly?

e.g., 
```
return when (protocol) {
            is ExchangeProtocolVersion.V2 -> {
                val decrypted = jweCrypto.jweDecryptRsaOaep(envelope.payload, ownPrivateKeyBase64)
                String(decrypted, Charsets.UTF_8)
            }

            is ExchangeProtocolVersion.V1 -> {
                throw IllegalArgumentException("Obsolete envelope version $protocol")
            }

            is ExchangeProtocolVersion.Unsupported -> {
                throw EnvelopeVersionTooNew(protocol)
            }
        }
```

CDRussell (approved):

sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/exchange/v2/ExchangeV2Envelope.kt
MiSikora: Normally, I would agree. But here I don't think the code would benefit from a compiler check. This a V2 envelope so anything `else` is too new. If we had a generic envelope type I think it would be correct to use `is`.

PR #9609, opened 2026-08-25
[Segmented Onboarding] No AI flow

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1216829123556663?focus=true
Tech Design URL (if applicable): 
API Proposals URL(s) (if applicable): https://app.asana.com/1/137249556945/project/1208671518894266/task/1217778222182956?focus=true

### Description

Adds the third segmented onboarding path: the one a user lands on by picking **"No AI"** as their
download reason.

The path runs comparison chart → default browser prompt → preference selector → Duck.ai on/off →
address bar position, and reuses the existing config-driven screen machinery. New pieces:

- **`SEARCH_ASSIST` and `HIDE_AI_GENERATED_IMAGES` onboarding preferences.** Both are offered on the
  preference selector and written straight to the SERP settings blob, so the web reflects them on its
  next `getNativeSettings`. Both rows seed to the position the no-AI path wants (Search Assist off,
  AI-generated images hidden) rather than to whatever is stored.
- **A Duck.ai on/off screen**, backed by a new `OnboardingSingleChoiceDataPlugin` in `duckchat-impl`
  that commits the pick to the same user setting the Duck.ai settings screen writes.
- **Typed SERP settings in `settings-api`.** `SafeSearch`, `SearchAssistVisibility` and
  `HideAiGeneratedImages` now share a `SerpSetting` contract, with `setSetting(setting)` and
  `observeSetting(default)` extensions that encode and decode the SERP's raw key/code pairs in one
  place. `SearchAssistVisibility` and `HideAiGeneratedImages` moved out of `duckchat-impl` so
  onboarding can reach them; `SafeSearch` replaces the `kp` literals that were inline in
  `OnboardingPreferenceApplier`. Duck.ai settings and the AI-features state pixel now read through the
  typed accessors.
- **A pre-title fade phase in the render engine.** A screen can now declare content that sits *above*
  its title (`ContentHandle.preTitleFadeTargets`), which fades in before the title types instead of
  with the body. The CTAs stay with the body fade. The Duck.ai state screen's pictogram is the first
  user.

### Steps to test this PR

_No-AI path, Duck.ai left off_
- [x] Apply below diff:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/SegmentedOnboardingExperimentManager.kt b/app/src/main/java/com/duckduckgo/app/onboarding/SegmentedOnboardingExperimentManager.kt
[...]

PR #9651, opened 2026-08-28
[Segmented Onboarding] Block Ads flow

LukasPaczos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1208671518894266/task/1216829123556664?focus=true
API Proposals URL(s) (if applicable): https://app.asana.com/1/137249556945/project/1212087397361015/task/1217861601448570?focus=true

### Description

Adds the block-ads path to segmented onboarding, the fourth and last of the download-reason paths.

The preference selector on this path offers three switches — ad blocking, rejecting optional cookies, and accepting the cookies that can't be rejected. The last one is a dependent row: it only shows while rejecting optional cookies is on.

`OnboardingPreferenceApplier` is replaced by `OnboardingPreferenceCatalog`, a single table where each preference names its row, its availability, the value to seed and how a pick is committed. The table only holds lambdas, so nothing is evaluated until the step that offers it is reached.

`OnboardingBooleanPreferencePlugin` is introduced to allow modules provide preferences. Onboarding still decides which preferences exist and where their rows sit, so it asks for a specific plugin by id rather than rendering whatever is contributed. Three plugins contribute today: ad blocking, reject optional cookies, accept non-opt-out cookies.

### Steps to test this PR

- [x] Apply below diff:
```diff
diff --git a/app/src/main/java/com/duckduckgo/app/onboarding/SegmentedOnboardingExperimentManager.kt b/app/src/main/java/com/duckduckgo/app/onboarding/SegmentedOnboardingExperimentManager.kt
index 5cecc0bb48..908629ae24 100644
--- a/app/src/main/java/com/duckduckgo/app/onboarding/SegmentedOnboardingExperimentManager.kt
+++ b/app/src/main/java/com/duckduckgo/app/onboarding/SegmentedOnboardingExperimentManager.kt
@@ -47,2 +47,3 @@ class SegmentedOnboardingExperimentManagerImpl @Inject constructor(
     override suspend fun enroll(): SegmentedOnboardingExperimentVariant? = withContext(dispatcherProvider.io()) {
+        return@withContext SegmentedOnboardingExperimentVariant.TREATMENT
         if (onboardingPrivacyConfigPersistedGate.awaitPersisted() && checkPrerequisites()) {

```

_Block ads path, toggle on, search_
- [x] Clean install.
- [x] Walk to the download reason screen and pick the **Block ads** option.
- [x] Confirm the comparison chart says "Ad blocking activated".
- [x] On the preference selector, confirm:
  - [x] "Block ads on YouTube" starts **on**
  - [x] Reject optional cookies starts **on**
  - [x] Accept some cookies starts **off**
- [x] Continue without changing them.
- [x] On the input pref screen keep toggle on and continue.
- [x] Verify you see input preview with toggle available.
- [x] Submit search query, continue onboarding, and verify you see a generic "High five!" End CTA.
- [x] Verify Settings -> Ad blocking -> Block ads on YouTube is on.
- [x] Verify Settings -> Cookie Pop-up Protection -> Reject optional cookies is on.
- [x] Verify Settings -> Cookie Pop-up Protection -> Accept cookies that can't be rejected is off.

_Block ads path, toggle off_
- [x] Clean install.
- [x] Walk to the download reason screen and pick the **Block ads** option.
- [x] On the preference selector, switch:
  - [x] Block ads on YouTube **off**
  - [x] Reject optional cookies **off**
  - [x] Verify "Accept some cookies starts" hides
- [x] Continue.
- [x] On the input pref screen select "Search-only".
- [x] Verify you see input preview with search only title.
- [x] Submit search query, continue onboarding, and verify you see a generic "High five!" End CTA.
- [x] Verify Settings -> Ad blocking -> Block ads on YouTube is off.
- [x] Verify Settings -> Cookie Pop-up Protection -> Reject optional cookies is off.
- [x] Verify Settings -> Cookie Pop-up Protection -> Accept cookies that can't be rejected is gone.

_Block ads path, toggle on, chat_
- [x] Clean install.
- [x] Walk to the download reason screen and pick the **Block ads** option.
- [x] On the preference selector, switch:
  - [x] Block ads on YouTube **on**
  - [x] Reject optional cookies **on**
  - [x] Accept some cookies starts **on**
- [x] Continue.
- [x] On the input pref screen keep toggle on and continue.
- [x] Verify you see input preview with toggle available.
- [x] Switch to Duck.ai and submit a prompt, continue onboarding. 
- [x] Verify you see "Use the address bar to search and visit sites or toggle to Duck.ai..." End CTA, with toggle visible on Search tab.
- [x] Verify Settings -> Ad blocking -> Block ads on YouTube is on.
- [x] Verify Settings -> Cookie Pop-up Protection -> Reject optional cookies is on.
- [x] Verify Settings -> Cookie Pop-up Protection -> Accept cookies that can't be rejected is on.

_Block ads path, autoconsent FF off_
- [x] Create  `privacy-config/privacy-config-internal/local-config-patches/autoconsent_off.json`:
```json
[
  { "op": "replace", "path": "/features/autoconsent/state", "value": "disabled" },
  { "op": "remove", "path": "/features/autoconsent/hash" },
  { "op": "replace", "path": "/version", "value": 90000000000001 }
]
```
- [x] Create `privacy-config/privacy-config-internal/local.properties`:
```
config_patches=privacy-config/privacy-config-internal/local-config-patches/autoconsent_off.json
```
- [x] Clean install the app.
- [x] Walk to the download reason screen and pick the **Block ads** option.
- [x] Continue.
- [x] Verify only "Block ads on YouTube" pref is shown.

_Block ads path, ad blocking `enabledByDefault` FF off_
- [ ] Create  `privacy-config/privacy-config-internal/local-config-patches/ad_blocking_extension_enabled_by_default.json`:
```json
[
  { "op": "replace", "path": "/features/adBlockingExtension/features/enabledByDefault/state", "value": "disabled" },
  { "op": "remove", "path": "/features/adBlockingExtension/features/enabledByDefault/hash" },
  { "op": "replace", "path": "/version", "value": 90000000000001 }
]
```
- [ ] Create `privacy-config/privacy-config-internal/local.properties`:
```
config_patches=privacy-config/privacy-config-internal/local-config-patches/ad_blocking_extension_enabled_by_default.json
```
- [ ] Clean install the app.
- [ ] Walk to the download reason screen and pick the **Block ads** option.
- [ ] Continue.
- [ ] Leave ad blocking as enabled
- [ ] Finish onboarding
- [ ] Open ad blocking settings
- [ ] Check no data collection disclaimer is shown

### UI changes

https://github.com/user-attachments/assets/5ad5a34d-ce19-4dd0-b3e9-30f56ed6e72c

---

> [!NOTE]
> **Medium Risk**
> Touches first-run onboarding orchestration and how ad blocking and autoconsent settings are written, with careful no-op logic to avoid overwriting remote defaults; scope is large but well covered by tests.
> 
> **Overview**
> Adds the **block ads** segmented onboarding path: choosing that download reason now runs a dedicated plan (comparison chart, default browser, preference selector, input screen, address bar, preview) instead of staying on the current step.
> 
> Onboarding preference handling is refactored: **`OnboardingPreferenceApplier` is removed** and replaced by **`OnboardingPreferenceCatalog`**, which builds fully resolved selector rows (copy, icons, seeds, apply) only when the step is reached. **`OnboardingBooleanPreferencePlugin`** lets ad-blocking and autoconsent modules supply rows and persist settings without the app owning their storage; onboarding still maps fixed **`OnboardingPreference`** values to plugin ids.
> 
> The block-ads selector offers **block ads**, **reject optional cookies**, and **accept non-opt-out cookies**, with the accept row **dependent** on reject being on (show/hide with fade and card resize). Rows can omit secondary text or icons; an optional **“adjust later in Settings”** caption is supported.
> 
> UI updates include **`PreferenceSelectorBinder`** dependent-row behavior, **`QuickSetupTwoLineSwitchRow`** nullable icon/secondary text, new cookie icons in the design system, and dialog resolution that passes catalog rows through instead of rebuilding them in the resolver.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a692a0c17680a4b2eff66a4e570fdce15f1e11b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/src/main/java/com/duckduckgo/app/onboarding/OnboardingPreferenceCatalog.kt
LukasPaczos, line 1: @mikescamell, in https://github.com/duckduckgo/Android/pull/9609#discussion_r3870242489 we discussed cleaning up the `OnboardingPreferenceApplier` last, after all feature PRs land. However, in this PR I also had to add logic to fetch presentation data (icon/copy) from external plugins which complicated the old `*Applier` even more, so opted to clean up the class in this PR directly instead.

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/onboarding/OnboardingAcceptNonOptOutCookiesPreferencePluginImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+        if (!autoconsentFeature.self().isEnabled() || !autoconsentFeature.cookiePopUpPreferenceSetting().isEnabled()) return null
+
+        return Preference(
+            primaryText = context.getString(R.string.autoconsent_onboarding_accept_non_opt_out_cookies_primary),
+            secondaryText = context.getString(R.string.autoconsent_onboarding_accept_non_opt_out_cookies_secondary),
+            iconRes = CommonR.drawable.cookie_color_24,
+        )
+    }
+
+    override suspend fun apply(enabled: Boolean) = autoconsent.changeClickAcceptEnabled(enabled)
```
LukasPaczos, line 59: @CrisBarreiro could you double-check the condition and the way the pref is applied is correct?

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/onboarding/OnboardingRejectOptionalCookiesPreferencePluginImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+            iconRes = CommonR.drawable.cookie_blocked_color_24,
+        )
+    }
+
+    override suspend fun apply(enabled: Boolean) {
+        // Persisting a value that already applies would turn the remote default into an explicit user
+        // choice, which can impact later default change or pixels.
+        if (enabled == autoconsent.isSettingEnabled()) return
+        autoconsent.changeSetting(enabled)
+    }
```
LukasPaczos, line 60: @CrisBarreiro could you double-check the condition and the way the pref is applied is correct?

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/onboarding/OnboardingAdBlockingPreferencePluginImpl.kt
```diff
@@ -0,0 +1,72 @@
...
+        // choice, which can impact later default change or pixels.
+        val alreadyApplies = when (statusChecker.observeState().firstOrNull()) {
+            Enabled.UserEnabled, Enabled.Default -> enabled
+            Disabled.Permanent -> !enabled
+            Disabled.UntilRelaunch, Uninitialized, null -> false
+        }
+        if (!alreadyApplies) {
+            settingsRepository.setEnabled(enabled)
+        }
+    }
```
LukasPaczos, line 71: @CrisBarreiro could you double-check the condition and the way the pref is applied is correct?

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/onboarding/OnboardingAcceptNonOptOutCookiesPreferencePluginImpl.kt
CrisBarreiro: Hm, so there's an interesting caveat here, and that's part of why preference plugins return a view, instead of parts of it. If you look at autoconsent settings, if the main setting for autoconsent is disabled, the row for non opt-out disappears. We should either do that or, when enabling this one, if the other is disabled, enable it as well

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/onboarding/OnboardingRejectOptionalCookiesPreferencePluginImpl.kt
CrisBarreiro: Looks correct

ad-blocking/ad-blocking-impl/src/main/java/com/duckduckgo/adblocking/impl/onboarding/OnboardingAdBlockingPreferencePluginImpl.kt
CrisBarreiro: Looks correct

autoconsent/autoconsent-impl/src/main/java/com/duckduckgo/autoconsent/impl/onboarding/OnboardingAcceptNonOptOutCookiesPreferencePluginImpl.kt
LukasPaczos: That condition is actually baked directly into onboarding logic, see https://github.com/duckduckgo/Android/pull/9651/changes#diff-725692fa525e7553d630590ba89346b7e097fdc6c60bfe1d541bdfa2c0eb715fR143:
```
OnboardingPreference.ACCEPT_NON_OPT_OUT_COOKIES to fromPlugin(
    id = OnboardingBooleanPreferencePlugin.Id.AcceptNonOptOutCookies,
    dependsOn = OnboardingPreference.REJECT_OPTIONAL_COOKIES,
    seed = { false },
```

We could in theory encode this relationship in the `OnboardingBooleanPreferencePlugin` but I'm not sure if it's worth doing just for this one case. What do you think @CrisBarreiro?

CrisBarreiro: But that's only for the initial state, right? Or does it change with user choices?

LukasPaczos: @CrisBarreiro it changes with user choices, exactly the same as in the settings activity:

https://github.com/user-attachments/assets/bb66ac49-6174-4d57-97e4-748cb94a085b

The `dependsOn` requires the "parent" to be both present and enabled.

CrisBarreiro: All good, then

Ideally, the CPM module would handle the dependency, but I'm ok with not rearchitecting everything now

LukasPaczos: Thanks @CrisBarreiro :+1: 

If another case like this emerges, we can consider folding this logic into the plugin contract in some way.

mikescamell:
@LukasPaczos Looks good! One minor issue

1. When navigating from the preference selector to the address bar picker, the left wing does not animate in but rather it just appears

https://github.com/user-attachments/assets/048eb0f8-4f9a-46f8-9ee1-8a4ceb1d4666

LukasPaczos:
@mikescamell nice find! This bug is unrelated to the PR, an it actually is worked around by https://github.com/duckduckgo/Android/pull/9659 anyway, would you mind if we tackled it separately?

https://github.com/user-attachments/assets/845bffd0-0671-44e9-8c32-e954c652e299

As far as I can tell, the repro steps are:
1. Have font large enough that the preference picker card hides the wing.
2. Move on to the toggle picker, which is a smaller card and has enough space for the wing.
3. Since _we're showing left wing again_, the render engine thinks it's already visible so just blinks it into place. Seems like it doesn't have the right signal that the collision happened.
  a. This stops being an issue after https://github.com/duckduckgo/Android/pull/9659 because we're not showing left wing twice anymore.

There's an associated small bug that, on the preference picker screen, if the CPM pref is disabled and collapses one of the rows, there's clearly enough space for the wing but it doesn't re-appear. We can look into the two issues together.

mikescamell:
> @mikescamell nice find! This bug is unrelated to the PR, an it actually is worked around by #9659 anyway, would you mind if we tackled it separately?
> 
>  Screen_recording_20260901_112528.mp4 
> As far as I can tell, the repro steps are:
> 
> 1. Have font large enough that the preference picker card hides the wing.
> 2. Move on to the toggle picker, which is a smaller card and has enough space for the wing.
> 3. Since _we're showing left wing again_, the render engine thinks it's already visible so just blinks it into place. Seems like it doesn't have the right signal that the collision happened.
>    a. This stops being an issue after [[Segmented Onboarding] Update backgrounds on segmented path #9659](https://github.com/duckduckgo/Android/pull/9659) because we're not showing left wing twice anymore.
> 
> There's an associated small bug that, on the preference picker screen, if the CPM pref is disabled and collapses one of the rows, there's clearly enough space for the wing but it doesn't re-appear. We can look into the two issues together.

I don't think it has to do with font size, just if you are on a device that's screen doesn't have enough room to show the wing as well as the dialog (classic Samsung S25 screen issue 😁)

And yep follow up sounds good! 👍

mikescamell (approved):
:shipit:

CrisBarreiro (changes requested):

CrisBarreiro (changes requested):
@LukasPaczos, when `enabledByDefault` is disabled and the user leaves the toggle as "on" during onboarding, ad blocking settings show the consent message, which they shouldn't

Also, unsure if it's expected that if `enabledByDefault` is disabled, we still show the toggle as enabled during onboarding

LukasPaczos:
@CrisBarreiro

> when enabledByDefault is disabled and the user leaves the toggle as "on" during onboarding, ad blocking settings show the consent message, which they shouldn't

This is expected here and fixed in https://github.com/duckduckgo/Android/pull/9672. We can retarget the branch and merge into this one instead to make it safer, but the diff will increase. Let me know whichever you prefer.

> Also, unsure if it's expected that if enabledByDefault is disabled, we still show the toggle as enabled during onboarding

This is expected, we aim to seed the value in onboarding.

CrisBarreiro:
> @CrisBarreiro
> 
> > when enabledByDefault is disabled and the user leaves the toggle as "on" during onboarding, ad blocking settings show the consent message, which they shouldn't
> 
> This is expected here and fixed in #9672. We can retarget the branch and merge into this one instead to make it safer, but the diff will increase. Let me know whichever you prefer.
> 
> > Also, unsure if it's expected that if enabledByDefault is disabled, we still show the toggle as enabled during onboarding
> 
> This is expected, we aim to seed the value in onboarding.

All good then

CrisBarreiro (approved):

LukasPaczos:
> I don't think it has to do with font size, just if you are on a device that's screen doesn't have enough room to show the wing as well as the dialog (classic Samsung S25 screen issue 😁)

Right @mikescamell, I just meant to restrict the space so it's reproducible :sweat_smile: Will create a follow up task!

PR #9708, opened 2026-09-03
Stop a first device_info write triggering a redundant repair PATCH

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/project/72649045549333/task/1218149406867552?focus=true

### Description
There is a narrow window where we might think a repair to re-publish `device_info` is needed and do one where it isn't necessary. Even without this  PR the state would end up ok, but the repair is redundant. 

**Cause**
Essentially, it happens when the request to get devices was issued before the PATCH publishing `device_info` landed and the get devices response is handled after the published marker is set. i.e., the get devices request fires and before it comes back we've patched `device_info`.

Repair is not triggered just by “device_info being missing”. It is triggered by device_info missing and we already marked this account as published. 
- If not marked as published  + no blob → reason NOT_PUBLISHED_YET → no repair.
- If marked as published + no blob → reason BLOB_ABSENT → repair PATCH.

The published marker is set immediately after the first-write PATCH succeeds. So any list read whose payload predates the PATCH (but which is decrypted after the latch is set) will classify as BLOB_ABSENT and fire a redundant repair.

This PR removes that redundant repair, by having a count of `device_info` publishes in memory and snapshotting the count before issuing the get-devices request.

### Steps to test this PR

_The genuine repair still fires when the blob really is missing_

Needs an `internal` build; no second device required.

- [ ] Open Sync Dev Settings and enable sync
- [ ] Turn `canWriteUnifiedDeviceList` OFF
- [ ] Confirm the Unified Devices Migration section still reads `migrated: true` — the repair only applies once this device has published
- [ ] Tap **Rename device (PATCH device_info)** and give it any new name. With write off this PATCH omits `device_info`, so the server clears the blob
- [ ] Tap **Fetch & decrypt devices**. This device's entry must read `device_info: (no device_info)`. If it still shows a blob the rename did not take effect — stop here
- [ ] Turn `canWriteUnifiedDeviceList` back ON
- [ ] Force-stop the app. The repair runs at most once per account per process, so an earlier attempt in the same session would mask the result
- [ ] Launch the app and open Settings → Sync (the signed-in devices screen)
- [ ] Expect in logcat: `this device's device_info did not resolve on read; re-publishing it` followed by `device_info re-published`
- [ ] Back in Dev Settings → Sync, tap **Fetch & decrypt devices** again. This device's entry now shows `device_info: name=…`, and the devices screen shows the correct name

Logcat filter: `message~:"Sync-UnifiedDevices"`

---

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit b38269be08a0e8a931edabccb31e7a3a3fa9a462. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyDeviceListDecryptor.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyDeviceListDecryptor.kt
@@ -39,8 +39,9 @@ import javax.inject.Inject
 interface ThirdPartyDeviceListDecryptor {
     /**
      * @param thisDeviceId this device's sync id, used to report back whether our own `device_info` resolved
+     * @param publishSnapshot [DeviceInfoPublishWatcher.snapshot] taken before [entries] were fetched, so a read that raced our first write isn't reported as a missing blob
      */
-    fun decryptAll(entries: List<DeviceV2>, thisDeviceId: String?): DecryptAllResult
+    fun decryptAll(entries: List<DeviceV2>, thisDeviceId: String?, publishSnapshot: Int): DecryptAllResult
 
     companion object {
         const val FALLBACK_TYPE_3PARTY = "Browser"
@@ -83,9 +84,10 @@ class RealThirdPartyDeviceListDecryptor @Inject constructor(
     private val deviceInfoDecryptor: DeviceInfoDecryptor,
     private val syncFeature: SyncFeature,
     private val syncStore: SyncStore,
+    private val deviceInfoPublishWatcher: DeviceInfoPublishWatcher,
 ) : ThirdPartyDeviceListDecryptor {
 
-    override fun decryptAll(entries: List<DeviceV2>, thisDeviceId: String?): DecryptAllResult {
+    override fun decryptAll(entries: List<DeviceV2>, thisDeviceId: String?, publishSnapshot: Int): DecryptAllResult {
         if (entries.isEmpty()) return DecryptAllResult(emptyList(), emptyList())
 
         val readEnabled = syncFeature.canReadUnifiedDeviceList().isEnabled()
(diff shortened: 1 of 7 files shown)


sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/ThirdPartyDeviceListDecryptor.kt
```diff
@@ -168,14 +171,17 @@ class RealThirdPartyDeviceListDecryptor @Inject constructor(
...
         val ownDeviceId = thisDeviceId ?: return null
         val ownEntry = entries.firstOrNull { it.deviceId == ownDeviceId } ?: return null
         if (viaDeviceInfo.any { it.deviceId == ownDeviceId }) return OwnDeviceReadOutcome.ResolvedDeviceInfo
 
+        val publishedDuringFetch = deviceInfoPublishWatcher.publishedSince(publishSnapshot)
         val reason = when {
             ownEntry.deviceId in failedDeviceInfoIds -> DeviceInfoReadFailureReason.BLOB_DECRYPT_FAILED
-            syncStore.unifiedDeviceListMigratedForUserId == syncStore.userId -> DeviceInfoReadFailureReason.BLOB_ABSENT
+            syncStore.unifiedDeviceListMigratedForUserId == syncStore.userId && !publishedDuringFetch ->
+                DeviceInfoReadFailureReason.BLOB_ABSENT
```
CDRussell, line 184: This is the core of the race condition fix 
- if there was a request sent to get devices
- and while that was inflight, we published `device_info`
- when the first request finishes, we don't want its missing `device_info` to trigger us into thinking we're needing to self-repair and republish the `device_info` again.

MiSikora (approved):

PR #9720, opened 2026-09-04
Add camera pixel for the sync flow

MiSikora:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218142887748447?focus=true
Tech Design URL (if applicable): N/A
API Proposals URL(s) (if applicable): N/A

### Description

Moves the "Go to Settings" action on the sync scanner's no camera permission screen from the fragment into the view model, so the view model can measure the camera permission state around it. Tapping "Go to Settings" on the no camera permission screen now routes through the view model's `onGoToPermissionSettingsClicked`, which sends the `OpenPermissionSettings` command.

### Steps to test this PR

_Go to settings from the no camera permission screen_
- [x] Have camera permission not granted.
- [x] Go to Sync & Backup.
- [x] Tap "Sync With Another Device".
- [x] Tap "I'm Ready to Scan".
- [x] Deny the camera permission when prompted.
- [x] Check the logs for the `sync_scanner_camera_permission_state` pixel.
- [x] Verify it fired once with `before_requesting=false` and `after_requesting=false`.
- [x] Tap "Go to Settings" on the no camera permission screen.
- [x] Verify the app's system settings screen opens.
- [x] Grant the camera permission in settings.
- [x] Return to the app.
- [x] Verify the camera scanner is shown.
- [x] Check the logs for the `sync_scanner_camera_permission_state` pixel.
- [x] Verify it fired again with `before_requesting=false` and `after_requesting=true`.

### UI changes
N/A

---

---

> [!NOTE]
> **Low Risk**
> Analytics and a small MVVM refactor on the sync scanner permission UI; no changes to sync auth or data handling.
> 
> **Overview**
> Adds analytics for **camera permission on the sync QR scanner**: a new `sync_scanner_camera_permission_state` pixel (with `before_requesting` / `after_requesting` booleans) and `SyncPixels.fireScannerCameraPermissionState`.
> 
> **Go to Settings** on the no-permission screen is routed through `ReadSyncCodeCameraIntroViewModel` (`onGoToPermissionSettingsClicked` → `OpenPermissionSettings` command) so the view model can record permission before/after the dialog or system settings. Reporting fires **once** per “resolution” path, with a second report allowed after the user opens settings from that screen.
> 
> Unit tests cover pixel wiring and the view model’s reporting rules (grant/deny, pre-granted, settings return).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6908df91d0f519133515f0749a3a020662b6449. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixelParamRemovalPlugin.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixelParamRemovalPlugin.kt
@@ -49,6 +49,7 @@ class SyncPixelParamRemovalPlugin @Inject constructor() : PixelParamRemovalPlugi
             SyncPixelName.SYNC_SETUP_DEEP_LINK_FLOW_ABANDONED.pixelName to PixelParameter.removeAtb(),
 
             SyncPixelName.SYNC_SETUP_BARCODE_SCREEN_SHOWN.pixelName to PixelParameter.removeAtb(),
+            SyncPixelName.SYNC_SCANNER_CAMERA_PERMISSION_STATE.pixelName to PixelParameter.removeAtb(),
             SyncPixelName.SYNC_SETUP_BARCODE_SCANNER_SUCCESS.pixelName to PixelParameter.removeAtb(),
             SyncPixelName.SYNC_SETUP_BARCODE_SCANNER_FAILED.pixelName to PixelParameter.removeAtb(),
             SyncPixelName.SYNC_SETUP_BARCODE_CODE_COPIED.pixelName to PixelParameter.removeAtb(),
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixelParamRemovalPlugin.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixelParamRemovalPlugin.kt
@@ -49,7 +49,6 @@ class SyncPixelParamRemovalPlugin @Inject constructor() : PixelParamRemovalPlugi
             SyncPixelName.SYNC_SETUP_DEEP_LINK_FLOW_ABANDONED.pixelName to PixelParameter.removeAtb(),
 
             SyncPixelName.SYNC_SETUP_BARCODE_SCREEN_SHOWN.pixelName to PixelParameter.removeAtb(),
-            SyncPixelName.SYNC_SCANNER_CAMERA_PERMISSION_STATE.pixelName to PixelParameter.removeAtb(),
             SyncPixelName.SYNC_SETUP_BARCODE_SCANNER_SUCCESS.pixelName to PixelParameter.removeAtb(),
             SyncPixelName.SYNC_SETUP_BARCODE_SCANNER_FAILED.pixelName to PixelParameter.removeAtb(),
             SyncPixelName.SYNC_SETUP_BARCODE_CODE_COPIED.pixelName to PixelParameter.removeAtb(),
--- a/PixelDefinitions/pixels/definitions/sync_setup.json5
+++ b/PixelDefinitions/pixels/definitions/sync_setup.json5
@@ -32,6 +32,25 @@
             }
         ]
     },
+    "sync_scanner_camera_permission_state": {
+        "description": "Reports the camera permission state observed on the sync scanner screen: whether it was already granted before and after asking for it, via the permission dialog or the system settings",
+        "owners": ["MiSikora"],
+        "triggers": ["other"],
+        "suffixes": ["form_factor"],
+        "parameters": [
+            "appVersion",
+            {
+                "key": "before_requesting",
+                "type": "boolean",
+                "description": "Whether camera permission was already granted before any request was made"
+            },
+            {
+                "key": "after_requesting",
+                "type": "boolean",
+                "description": "Whether camera permission ended up granted, via the permission dialog or the system settings"
+            }
+        ]
+    },
     "sync_setup_barcode_screen_shown": {
         "description": "The sync setup barcode (QR) screen was shown to the user",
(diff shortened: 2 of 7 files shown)


sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/pixels/SyncPixelParamRemovalPlugin.kt
```diff
@@ -49,6 +49,7 @@ class SyncPixelParamRemovalPlugin @Inject constructor() : PixelParamRemovalPlugi
             SyncPixelName.SYNC_SETUP_DEEP_LINK_FLOW_ABANDONED.pixelName to PixelParameter.removeAtb(),
 
             SyncPixelName.SYNC_SETUP_BARCODE_SCREEN_SHOWN.pixelName to PixelParameter.removeAtb(),
+            SyncPixelName.SYNC_SCANNER_CAMERA_PERMISSION_STATE.pixelName to PixelParameter.removeAtb(),
```
CrisBarreiro: Nit: Atb is no longer added by default, so no need to remove it

MiSikora: Changed in [c6908df](https://github.com/duckduckgo/Android/pull/9720/commits/c6908df91d0f519133515f0749a3a020662b6449)

CrisBarreiro:
Looks good and works as expected according to the test plan, but some comments:
* Should we feature-flag the changes?
* If the user goes to settings and selects "Ask every time" instead of "Allow only while using the app" we won't re-check when going back to the app, and just keep showing the "Camera Permission is Required" message. Is this intentional?

MiSikora:
> Should we feature-flag the changes?

What would we feature-flag here? The only change to the actual logic is that `OpenPermissionSettings` triggers opening settings from the command, instead of directly from the button's listener. I think the change is trivial enough that it bears no risk.

> If the user goes to settings and selects "Ask every time" instead of "Allow only while using the app" we won't re-check when going back to the app, and just keep showing the "Camera Permission is Required" message. Is this intentional?

It was never designed for that case, so I'd say it's unintentional. But to me, this feels like an edge case scenario anyway. Users already have to go out of their way to open system settings and still not grant the permission.

This pixel is intended for us to understand camera handling better and optimize the flow if needed. I can note it down in Asana and bring this up when discussing improvements to this flow.

CrisBarreiro:
> > Should we feature-flag the changes?
> 
> What would we feature-flag here? The only change to the actual logic is that `OpenPermissionSettings` triggers opening settings from the command, instead of directly from the button's listener. I think the change is trivial enough that it bears no risk.
> 
> > If the user goes to settings and selects "Ask every time" instead of "Allow only while using the app" we won't re-check when going back to the app, and just keep showing the "Camera Permission is Required" message. Is this intentional?
> 
> It was never designed for that case, so I'd say it's unintentional. But to me, this feels like an edge case scenario anyway. Users already have to go out of their way to open system settings and still not grant the permission.
> 
> This pixel is intended for us to understand camera handling better and optimize the flow if needed. I can note it down in Asana and bring this up when discussing improvements to this flow.

Agree it sounds low risk enough

>It was never designed for that case, so I'd say it's unintentional. But to me, this feels like an edge case scenario anyway. Users already have to go out of their way to open system settings and still not grant the permission.

That was my suspicion, but I worry this might be a gap in design, so regardless of this PR, I think it makes sense to bring it up on your project in case it was indeed overlooked. I'll leave it up to your judgement whether it makes sense to merge as is or wait for a response

CrisBarreiro (approved):

MiSikora:
> That was my suspicion, but I worry this might be a gap in design, so regardless of this PR, I think it makes sense to bring it up on your project in case it was indeed overlooked. I'll leave it up to your judgement whether it makes sense to merge as is or wait for a response

I don't disagree about that. But I also wouldn't change it in this PR since this is only about adding a pixel.

In general to your point I think that Android stands in our way a bit here. There is no way to distinguish between "Ask every time" from "Denied once". Probably the best we could do is something like:

_<fresh_user>_
1. Tap "I'm Ready to Scan".
2. Deny.
3. Instead of "Go to settings" show them something like "Grant camera access".
4. They tap "Grant camera access".
5. Deny.
6. At this point we show "Go to settings".
7. When coming back to the screen or opening it again they see "Grant camera access" or "Go to settings" depending on their choices.

CrisBarreiro:
> > > Should we feature-flag the changes?
> > 
> > 
> > What would we feature-flag here? The only change to the actual logic is that `OpenPermissionSettings` triggers opening settings from the command, instead of directly from the button's listener. I think the change is trivial enough that it bears no risk.
> > > If the user goes to settings and selects "Ask every time" instead of "Allow only while using the app" we won't re-check when going back to the app, and just keep showing the "Camera Permission is Required" message. Is this intentional?
> > 
> > 
> > It was never designed for that case, so I'd say it's unintentional. But to me, this feels like an edge case scenario anyway. Users already have to go out of their way to open system settings and still not grant the permission.
> > This pixel is intended for us to understand camera handling better and optimize the flow if needed. I can note it down in Asana and bring this up when discussing improvements to this flow.
> 
> Agree it sounds low risk enough
> 
> > It was never designed for that case, so I'd say it's unintentional. But to me, this feels like an edge case scenario anyway. Users already have to go out of their way to open system settings and still not grant the permission.
> 
> That was my suspicion, but I worry this might be a gap in design, so regardless of this PR, I think it makes sense to bring it up on your project in case it was indeed overlooked. I'll leave it up to your judgement whether it makes sense to merge as is or wait for a response

Yea, the pixel makes sense regardless. Other than the UX, we treat "Deny" and "Ask" as not having the permission, but that's consistent with the experience users see

PR #9757, opened 2026-09-09
Open the performance-optimized paywall URLs when the flag is on

Kyriakos-Georgiopoulos:
Task/Issue URL: https://app.asana.com/1/137249556945/project/1217319487385906/task/1217750626830941?focus=true

### Description
Opens the performance-optimized paywall page when the `performanceOptimizedPaywalls` flag is on, so the front end can test the **faster** entry points on Android.

- The paths come from the flag settings, so they can change in remote config without an app release https://github.com/duckduckgo/privacy-configuration/pull/5809
- The URL carries `trial=true` or `trial=false` from free trial eligibility, and `pir=false` when PIR is not part of the offer, so the page renders without asking the app what the user is being offered.
- Params the app already sets, such as `origin`, are carried onto the new path unchanged.
- With the flag off, the old URL is loaded and nothing changes.

### Steps to test this PR
- [x] Apply the pinned staging patch from https://app.asana.com/1/137249556945/project/1209991789468715/task/1210448620621729?focus=true and install the internal debug build on a device with Google Play.
- [x] Launch the app once with network and confirm the remote config has landed: Settings > Internal Features > Privacy Config Internal Settings, check "Latest Loaded Date" is today. The paths come from the config, not from the app, and they ship in https://github.com/duckduckgo/privacy-configuration/pull/5809.
- [ ] Settings > Internal Features > Feature Flag Inventory, search `performanceOptimizedPaywalls`, turn it on.
- [ ] Filter logcat on `SubscriptionsWebViewActivity`. It logs the URL every time the paywall opens. Every check below just reads that line.
- [ ] Optional, to see the page render: take the current front end tunnel URL from https://app.asana.com/1/137249556945/task/1216295712277568, put it in Settings > Internal Features > Subscriptions Dev Settings > "Enter the base URL for subscriptions", and Save. The tunnel is temporary.

_VPN paywall_
- [ ] Open from Settings > Privacy Pro. URL should be `/subscriptions/new/mobile/vpn`.

_Duck.ai paywall_
- [ ] Open from a Duck.ai entry point. URL should be `/subscriptions/new/mobile/duckai`, with no `featurePage` in the query.

_Query params_
- [ ] Trial eligible account: `trial=true`. Account that already used a trial: `trial=false`.
- [ ] Region without PIR: `pir=false` present. Region with PIR: no `pir` param.
- [ ] Open from app settings: `origin=funnel_appsettings_android` and `environment=staging` both survive onto the new path.

_Flag off_
- [ ] Turn `performanceOptimizedPaywalls` off, open the paywall. URL should be `/subscriptions` with no `trial` or `pir`, and the old page loads.

_Fallback_
- [ ] Force stop the app, reopen, go to the paywall before Google Play has loaded products. Old URL loads, no error page, no crash.

### UX changes
| Before  | After |
| ------ | ----- |
https://github.com/user-attachments/assets/5f6a6240-da74-4a85-9e98-c0974be6841e|https://github.com/user-attachments/assets/43d3462f-8e03-4f73-9c88-84e3a6998baf|

---

> [!NOTE]
> **Medium Risk**
> Changes which URL loads at paywall open and depends on Play offer/trial state; fallbacks keep the legacy URL, but wrong rewrite could affect purchase funnel and analytics origins.
> 
> **Overview**
> When **`performanceOptimizedPaywalls`** is enabled, the subscriptions webview no longer loads the generic buy URL directly. A new **`PaywallUrlResolver`** rewrites matching paywall URLs to remote-config entry paths (VPN vs Duck.ai from `featurePage`), injects **`trial`** from free-trial eligibility and **`pir=false`** when PIR isn’t on the offer, and preserves other query params like **`origin`**. If the flag is off, offers aren’t ready, or rewrite fails, the original URL is used unchanged.
> 
> **`SubscriptionWebViewViewModel`** resolves that URL once and exposes it via **`initialUrl`**; **`SubscriptionsWebViewActivity`** waits for the first resolved URL before **`loadUrl`**. Shared constants for feature pages and query keys replace a local Duck.ai string in JS messaging. Unit tests cover resolver behavior and view-model URL resolution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6efc658430a974ff401c09ebaea98a776dfb07ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/internal/PaywallUrlResolver.kt
+++ b/subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/internal/PaywallUrlResolver.kt
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.subscriptions.impl.internal
+
+import android.net.Uri
+import androidx.core.net.toUri
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.subscriptions.api.Product
+import com.duckduckgo.subscriptions.impl.SubscriptionsConstants.FEATURE_PAGE_QUERY_PARAM_KEY
+import com.duckduckgo.subscriptions.impl.SubscriptionsConstants.PIR_QUERY_PARAM_KEY
+import com.duckduckgo.subscriptions.impl.SubscriptionsConstants.TRIAL_QUERY_PARAM_KEY
+import com.duckduckgo.subscriptions.impl.SubscriptionsConstants.VPN_FEATURE_PAGE
+import com.duckduckgo.subscriptions.impl.SubscriptionsFeature
+import com.duckduckgo.subscriptions.impl.SubscriptionsManager
+import com.squareup.anvil.annotations.ContributesBinding
+import kotlinx.coroutines.withContext
+import javax.inject.Inject
+
+interface PaywallUrlResolver {
+    suspend fun resolve(url: String): String
+}
+
+@ContributesBinding(AppScope::class)
+class RealPaywallUrlResolver @Inject constructor(
+    private val subscriptionsUrlProvider: SubscriptionsUrlProvider,
+    private val subscriptionsManager: SubscriptionsManager,
+    private val subscriptionsFeature: SubscriptionsFeature,
+    private val paywallPathProvider: PaywallPathProvider,
+    private val dispatcherProvider: DispatcherProvider,
+) : PaywallUrlResolver {
+
+    override suspend fun resolve(url: String): String = withContext(dispatcherProvider.io()) {
+        runCatching {
+            optimizedUrl(url)
+        }.getOrNull() ?: url
+    }
+
+    private suspend fun optimizedUrl(url: String): String? {
+        if (!subscriptionsFeature.performanceOptimizedPaywalls().isEnabled()) return null
+        val uri = url.toUri()
+        if (!isPaywallUrl(uri)) return null
+        return rewrite(uri)
+    }
+
+    private fun isPaywallUrl(uri: Uri): Boolean {
+        val buyUri = subscriptionsUrlProvider.buyUrl.toUri()
+        return uri.host == buyUri.host && uri.path == buyUri.path
+    }
+
+    private suspend fun rewrite(uri: Uri): String? {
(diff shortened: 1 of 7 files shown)


subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/internal/PaywallUrlResolver.kt
```diff
@@ -0,0 +1,104 @@
...
+    override suspend fun resolve(url: String): String = withContext(dispatcherProvider.io()) {
+        runCatching {
+            optimizedUrl(url)
+        }.getOrNull() ?: url
+    }
+
+    private suspend fun optimizedUrl(url: String): String? {
+        if (!subscriptionsFeature.performanceOptimizedPaywalls().isEnabled()) return null
+        val uri = url.toUri()
+        if (!isPaywallUrl(uri)) return null
```
nalcalag: NIT: It's cheaper to make _isPaywallUrl_ check, before the disk-backed toggle read
```suggestion
        val uri = url.toUri()
        if (!isPaywallUrl(uri)) return null
        if (!subscriptionsFeature.performanceOptimizedPaywalls().isEnabled()) return null
```

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/ui/SubscriptionWebViewViewModel.kt
```diff
@@ -172,6 +174,12 @@ class SubscriptionWebViewViewModel @Inject constructor(
             .launchIn(viewModelScope)
     }
 
+    fun loadInitialUrl(url: String) {
+        viewModelScope.launch {
+            command.send(LoadUrl(paywallUrlResolver.resolve(url)))
```
nalcalag: `command` is a single slot `DROP_OLDEST` channel read by a `STARTED`-scoped collector, so this one shot `LoadUrl` can be evicted by another command or lost if the user backgrounds the screen mid-resolution with a blank WebView and nothing to retry from.
Could we expose the resolved URL as state rather than an event, so it can't be dropped and replays when the collector restarts?

nalcalag (commented):
It works as expected. Main thing I'd revisit is how the resolved URL reaches the WebView , now that resolving isn't instant, a fire-and-forget event leaves a window where it can go missing and the user gets an empty screen. Holding it as state the activity can read at any point (a `MutableStateFlow` rather than a `Command`) would be safer. Something like:
   
    private val _initialUrl = MutableStateFlow<String?>(null)
    val initialUrl = _initialUrl.asStateFlow()

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/internal/PaywallUrlResolver.kt
Kyriakos-Georgiopoulos: Good catch, done in [2a247c23d1](https://github.com/duckduckgo/Android/pull/9757/changes/2a247c23d1cbeb2673f4866b35d46c8606cf8fae).

subscriptions/subscriptions-impl/src/main/java/com/duckduckgo/subscriptions/impl/ui/SubscriptionWebViewViewModel.kt
Kyriakos-Georgiopoulos: Done, `initialUrl` is now a `StateFlow` as you suggested so it can't be dropped.

nalcalag (approved):
LGTM

PR #9760, opened 2026-09-09
Duck.Ai/ChatHistory: Add context menu

karlenDimla:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218238186970938?focus=true
Tech Design URL (if applicable): 
API Proposals URL(s) (if applicable):

### Description
 - Chat history screen gains an overflow menu (gated by the flag) with Select chats and Chats protection actions; Chats protection opens the Duck.ai chat-protection URL (new RealDuckChat.getChatProtectionUrl()), in a new tab unless already on a chat tab.

### Setup
  - [x] Internal build installed
  - [x] At least three saved chat exists in history

_`nativeDuckAiSidebar` enabled, default_
- [x] Open the native chat history screen (Duck.ai sidebar button with `nativeDuckAiSidebar` on) from a duck.ai chat tab
- [x] Overflow (⋮) icon **IS** visible in the chat history toolbar
- [x] Tapping ⋮ opens a popup with **Select chats** and **Chats protection*
- [x] Tapping **Select chats** enters selection mode (toolbar switches to select-mode actions)
- [x] Verify items are selectable, selecting/deselecting ALL works and Overflow (⋮) icon and search icon are NOT visible 
- [x] Press back and verify you exit the select state
- [x] Tapping Overflow (⋮) icon and select **Chat protection** from the  menu
- [x] Verify that history screen is dismissed and on the previously opened tab, the chat protection sheet becomes visible
- [x] Go to New Tab Page
- [x] Select menu and press "Chats"
- [x] Tapping Overflow (⋮) icon and select **Chat protection** from the menu
- [x] Verify a new duck.ai chat tab opens and the  chat protection sheet becomes visible

_`nativeDuckAiSidebar` disabled (via FF inventory)_
- [x] Go to New Tab Page
- [x] Select menu and press "Chats"
- [x] Verify that Overflow (⋮) icon is **NOT** visible while the history search bar is showing

---

> [!NOTE]
> **Low Risk**
> UI and navigation-only changes behind an existing feature flag, with unit tests for URL/tab routing; no auth or data-layer changes.
> 
> **Overview**
> Adds a **toolbar overflow (⋮) menu** on native Duck.ai chat history, shown only when **`nativeDuckAiSidebar`** is enabled. The menu offers **Select Chats** (enters bulk selection; hidden when history isn’t loaded) and **Chat Protection**.
> 
[...]

PR #9752, opened 2026-09-09
DuckAi/Sidebar: Implement replace Chat Sidebar with chats history

karlenDimla:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218228567702514?focus=true
Tech Design URL (if applicable): 
API Proposals URL(s) (if applicable): https://app.asana.com/1/137249556945/project/1201462763415876/task/1218312857360677?focus=true

### Description
This is the main PR to Implement replace Chat Sidebar with chats history and include all change outlined in the stack:
- https://github.com/duckduckgo/Android/pull/9760
- https://github.com/duckduckgo/Android/pull/9761
- https://github.com/duckduckgo/Android/pull/9773
### Steps to test this PR
_nativeDuckAiSidebar enabled, default_
- [x] Open a duck.ai chat
- [x] Select the sidebar
- [x] Verify that chats history opens

_nativeDuckAiSidebar disabled_
- [x] Open a duck.ai chat
- [x] Select the sidebar
- [ ] Verify that the FE sidebar opens

---

> [!NOTE]
> **Medium Risk**
> Changes omnibar Duck.ai navigation behind a remote flag; incorrect rollout could send users to the wrong sidebar/history surface, though legacy behavior is preserved when the flag is off.
> 
> **Overview**
> **Routes the omnibar Duck.ai sidebar tap through a feature-flagged path** so users with `nativeDuckAiSidebar` enabled open native chat history (`openDuckChatHistory` / `LaunchDuckChatHistory`) instead of the legacy web sidebar JS bridge.
> 
> Adds **`nativeDuckAiSidebar`** to `DuckAiFeatureState`, wired from remote config via new `DuckChatFeature.nativeDuckAiSidebar()` (default internal) in `RealDuckChat`. `BrowserTabFragment` now calls `onDuckChatSidebarButtonPressed()` instead of `openDuckChatSidebar()` directly.
> 
> Also **fixes swapped native actions**: `onCustomizeResponsesClicked` now emits `NativeAction.CUSTOMIZE_RESPONSES` and `openDuckChatSidebar` emits `NativeAction.SIDEBAR` (they were reversed). Unit tests cover the flag branches, history vs sidebar fallback, and the corrected actions.
> 
[...]

PR #9761, opened 2026-09-09
DuckAi/ChatTab: Add New search and remove New Tab option from + menu 

karlenDimla:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218238186970934?focus=true

### Description
 - New "+" plus menu on the omnibar with New Chat, New Voice Chat, New Image (sidebar-only), New Search / New Tab, and New Fire Tab entries.
 -  New Search (the sidebar relabel of "New Tab") opens a regular new tab and auto-surfaces the native input focused on the Search tab; with the flag off the entry stays "New Tab" and behaves as before.
- Generalized the onboarding input-screen one-shot into InputScreenLaunchTarget (carries an InputMode instead of a DUCK_AI boolean) so both onboarding (Duck.ai tab) and New Search (Search tab) can drive the initial input mode.

### Steps to test this PR
Setup
  - [x] Internal build installed
  - [x] Native input / input screen feature is active (native omnibar input widget with the Search / Duck.ai toggle is in use)
  - [x] Do at least one duck.ai chat 
  - [x] Open the browser on a New Tab Page and focus the omnibar so the "+" plus button is available

  1. `nativeDuckAiSidebar` enabled (default) — open the "+" menu
  - [x] Tap the "+" plus button in the omnibar
  - [x] The menu entry (with the add-tab icon) reads New Search (not "New Tab")
  - [x] The New Image entry IS visible

  2. nativeDuckAiSidebar enabled — tap New Search
  - [x] A new tab opens (regular, non-Fire)
  - [x] The native input widget IS auto-surfaced (keyboard shown), without a manual omnibar tap
  - [x] The widget opens with the Search tab selected (not Duck.ai)
  - [x] The input field is empty and ready to type
  - [x] Typing a query and submitting performs a search (not a Duck.ai chat)

  3. `nativeDuckAiSidebar` disabled (via FF Inventory) — open the "+" menu
  - [x] Tap the "+" plus button in the omnibar
  - [x] The menu entry reads New Tab (not "New Search")
  - [x] The New Image entry is NOT visible

  4. nativeDuckAiSidebar disabled — tap New Tab
  - [x] A new tab opens (regular, non-Fire)
  - [x] The tab lands with the address bar unfocused (native input is NOT force-opened)
  - [ ] No Search-tab pre-selection behavior occurs (unchanged pre-existing behavior)

  Regression check
  - [ ] Duck.ai onboarding: completing the Search-path onboarding still auto-opens the input screen on the Duck.ai (chat) tab
  - [ ] Custom-AI onboarding still finishes on the Duck.ai (chat) tab
  - [ ] Focusing the omnibar on an existing tab (no menu action) still opens the input with the default toggle position — SEARCH is not spuriously forced

---

> [!NOTE]
> **Medium Risk**
> Touches new-tab creation, activity recreate/deferred actions, and native input initial mode; behavior is gated on `nativeDuckAiSidebar` but onboarding Duck.ai landing must stay correct.
> 
> **Overview**
> When **native Duck.ai sidebar** is on, the omnibar **+** menu relabels **New Tab** to **New Search**, shows a **New Image** row (action still TODO), and opening that entry starts a regular tab with the native input screen on the **Search** tab—not Duck.ai.
> 
> **Per-tab initial input mode** is threaded through `launchNewTab` → `PendingAction.OpenNewTab` (including bundle round-trip on browser-mode recreate) → `pendingInputModeTargets` / `BrowserTabFragment.inputModeTarget` → `BrowserTabViewModel.consumeInitialInputMode()`, which prefers that one-shot target and still falls back to post-onboarding **Duck.ai** via `OnboardingInputScreenLaunchTarget` in the ViewModel (removed from the fragment).
> 
> Tests cover `InputMode` persistence across pending mode switch and one-shot consumption behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aabfdda4d3522fe0d4b8ddee0f67d6d9636c44f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


karlenDimla:
* **#9773** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9773?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9761** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9761?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9761?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9760** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9760?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9752** <a href="https://app.graphite.com/github/pr/duckduckgo/Android/9752?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`


<h2></h2>

app/src/main/java/com/duckduckgo/app/browser/InputScreenLaunchTarget.kt
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Returns the armed mode and clears the signal so subsequent launches behave normally. Returns
+     * `null` when not armed.
+     */
+    fun consumeInitialInputMode(): InputMode?
+}
+
+@SingleInstanceIn(AppScope::class)
+@ContributesBinding(AppScope::class)
+class InputScreenLaunchTargetImpl @Inject constructor() : InputScreenLaunchTarget {
```
karlenDimla, line 46: @LukasPaczos  I am repurposing the OnboardingInputScreenLauncher. Do youu mind taking a look and testing the usecase of that class?

app/src/main/java/com/duckduckgo/app/browser/InputScreenLaunchTarget.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import javax.inject.Inject
+
+/**
+ * A one-shot signal for which mode the next auto-launched input screen should open in. Armed by
+ * onboarding (to land on Duck.ai) and by the "New Search" menu action (to land on Search), and
+ * cleared the moment the input screen reads it.
+ */
+interface InputScreenLaunchTarget {
```
malmstein, line 30: this should eventually by in it's own native input module. For now we can leave it here since it's replacing the Onboarding one, but we'll need to review this as API Proposal, etc...

app/src/main/java/com/duckduckgo/app/browser/InputScreenLaunchTarget.kt
malmstein: FWIW I prefer this since it takes the logic out of Onboarding and into a more purposeful API

malmstein (approved):
LGTM. Just a comment for the future API. nice work!

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -3915,10 +3915,16 @@ class BrowserTabViewModel @Inject constructor(
...
-        val shouldHideKeyboard = cta?.shouldDropAddressBarFocusWhenShown() == true ||
-            duckAiFeatureState.showInputScreen.value ||
-            currentBrowserViewState().lastQueryOrigin == QueryOrigin.FromBookmark ||
-            (settingsDataStore.omnibarType == OmnibarType.SPLIT && alreadyShownKeyboard)
+        // A pending launch target (e.g. the "New Search" menu action) wants this tab to land with its
+        // input screen already surfaced, so force focus regardless of the usual drop rules. Peek only —
+        // the target is consumed when the input screen actually opens.
+        val forceInputScreen = inputScreenLaunchTarget.peekInitialInputMode() != null
+        val shouldHideKeyboard = !forceInputScreen && (
+            cta?.shouldDropAddressBarFocusWhenShown() == true ||
```
LukasPaczos, line 3923: This probably doesn't cause issues today (though I'm not 100% sure), but can silently introduce bugs in the future.

For example, if someone introduces a new `Cta` that shows "the next time NTP is opened" and returns `shouldDropAddressBarFocusWhenShown=true`, this will silently compete with the `forceInputScreen` and the dialog could show with a keyboard open and native input shown, even though it was intended to be shown with input closed. There's nothing to tell whoever builds that `Cta` that this can happen, unless they stumble upon this piece of code.

LukasPaczos (commented):
The original `OnboardingInputScreenLaunchTarget` was deliberately scoped as narrowly as possible, and deliberately named `Onboarding*` so it wouldn't invite other uses. I think it's a dangerous API: it sets a token in _global_ scope that gets consumed by _whichever_ tab shows an input widget next. That could be the current tab regaining focus, or a new tab that opens next - the caller has no control over it. In practice it relies on the side effect that every new tab opens with the native input focused. If that side effect ever changes, every flow built on this launch target will quietly break. We also have race conditions where the outgoing tab briefly initializes while a new one is being created ([case in point](https://github.com/duckduckgo/Android/pull/9787)), which would consume the token prematurely.

Onboarding required this because it never sends an explicit "open a tab" or "open the input widget" command - it relies purely on the side effect, and we got away with it because we control that flow tightly enough. It works, but it's fragile, and if we ever change how `CtaViewModel` interacts with the `BrowserTabViewModel`, I'd like to remove it.

The new tab click handler integrated here is different: it already calls `browserActivity?.launchNewTab(browserMode = BrowserMode.REGULAR)` explicitly. That's a direct signal to the activity that we're launching a tab in this specific context. If we pass that context along to the created tab fragment and its VM, we get a much more direct and robust mechanism than the global token.

Could we explore that path first? Concretely, I mean something like:
```kotlin
browserActivity?.launchNewTab(browserMode = BrowserMode.REGULAR, inputModeTarget = InputMode.SEARCH)
```

and thread that through to `nativeInputManager.showNativeInput`, rather than widening `InputScreenLaunchTarget` to more callers and risking silent bugs and race conditions.

Only if that turns out not to be feasible would I consider extending `InputScreenLaunchTarget`.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
karlenDimla: I removed the check here since it is not needed and will not be needed when we fix keyboard behavior in NTP (we want it to show).

karlenDimla:
@LukasPaczos  @malmstein  I thought about it addressed all your comments. For now, I removed the `InputScreenLaunchTarget` as I agree that this will require a proper thinking. Because of this, I had to add another variable in the BrowserActivity -> pendingInputModeTargets. I also don't like this approach but it is the one or the other. Let me know what you think!

app/src/main/java/com/duckduckgo/app/browser/BrowserActivity.kt
```diff
@@ -272,7 +273,7 @@ open class BrowserActivity : DuckDuckGoActivity() {
 
     /**
      * The launch source extra from a genuinely new [Intent] delivery (fresh launch or [onNewIntent]),
-     * consumed by the next [onResume]. Not read off [getIntent] directly, since after process death the
+     * consumed by thce next [onResume]. Not read off [getIntent] directly, since after process death the
```
malmstein: nit: typo slipped into this line, "thce next [onResume]".

malmstein (approved):
LGTM. thanks for reworking this, threading the target per tab through the pending map is clearer than the global one-shot, and consumeInitialInputMode makes the fallback ordering obvious.

app/src/main/java/com/duckduckgo/app/browser/BrowserTabViewModel.kt
```diff
@@ -1012,9 +1015,11 @@ class BrowserTabViewModel @Inject constructor(
         initialUrl: String?,
         skipHome: Boolean,
         isExternal: Boolean,
+        inputModeTarget: InputMode? = null,
     ) {
         this.tabId = tabId
         this.skipHome = skipHome
+        this.inputModeTarget = inputModeTarget
```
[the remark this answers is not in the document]
karlenDimla: This is not medium and could only occur as an edge case. Deferring for now.

PR #9769, opened 2026-09-10
Add the missing ddg wrap when adopting a 3party-minted account_info key, and repair already-broken devices on unified device-list read.

CDRussell:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1218328293620602?focus=true

### Description

The unified device list is encrypted with one shared `account_info` key that lives on the server. That key gets stored once per browser "kind" that needs it — one copy locked for us (`ddg`), one copy locked for third-party browsers like Chrome (`3party`). You can only read the list if there's a copy locked for you.

If Chrome creates the key before we do — which happens whenever a device signs up while `canWriteUnifiedDeviceList` is off, and it also happens on Chrome-first pairings — the server ends up with a `3party` copy only. We then adopt Chrome's key, happily encrypt our own device with it, but we never made ourselves a copy we can open. So writes look fine and reads silently fall back to the old device names, logging `no_wrap_for_our_credential`.

We already had code that fixes this, but it only ran at login, so it never fired on the adopt path.

**The fix:** make our own copy of the key whenever we adopt someone else's, and for devices already stuck like this, make it the first time a device-list read complains it can't open the key.

### Steps to test this PR

_Clean install — the bug_
- [ ] Open 3rd party browser (e.g., Chrome) and visit duck ai test site (see asana task)
- [ ] Fresh install `internalDebug`. 
- In Sync Internal Settings set 
  - [ ] `canUseV2ConnectFlow` **on**, 
  - [ ] `canWriteUnifiedDeviceList` **off**
  - [ ] `canReadUnifiedDeviceList` **off**
- [ ] On Android, choose to `Sync with another device` and share the text pairing code and paste it into duck ai test site sync setup (`Settings & More` -> `Settings` -> `Sync & Backup / Turn on` -> `Sync with another device`). And tap done when finished pairing to return to device list.
- [ ] In Andriod, confirm no `signup wrote device_info + account_info key` and the list logs `0 via device_info`
- [ ] On Android, in internal sync dev screen, tap **Fetch Keys** and verify it shows `account_info` with `encrypted_with=3party` only
- [ ] Turn both unified flags **on**, then **Run migration now**
- [ ] Expect `adopted existing key` → `adding missing ddg wrap` → `added ddg wrap`, list becomes `2 via device_info`, and **Fetch Keys** shows both `ddg` and `3party` under the same kid

Logcat filter: `adb logcat -v time | grep -E "Sync-UnifiedDevices|account_info_key"`

### UI changes
No UI changes

---

---

> [!NOTE]
> **Medium Risk**
> Changes protected-key wrapping and server key uploads for sync account_info; failures are best-effort but incorrect repair could leave native devices unable to read the unified device list until a retry succeeds.
> 
> **Overview**
> Introduces **`AccountInfoDdgWrapRepairer`** to add a missing native (`ddg`) wrap on the shared `account_info` key when the server only has a `3party` wrap—unwrap via scoped password, re-encrypt for `ddg`, and upload with `setKeysIfAbsent`, including confirm-after-conflict paths and wrap pixels.
> 
> **Adoption and login:** `AccountInfoKeyManager` runs repair best-effort in `finishAdopt` (including when `Existing` has no public key and keys must be fetched). `LoginDeviceInfoWriter` drops its duplicated re-wrap block and delegates to the repairer.
> 
> **Already-broken clients:** On unified device-list read, when decryption reports `NO_WRAP_FOR_OUR_CREDENTIAL`, `SyncAccountRepository` triggers the same repair once per account (retries after failure). Tests cover the repairer, adoption, login, and list-read latch behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17d699c21e1b0424afa05746c902126a205d1e6c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/AccountInfoDdgWrapRepairer.kt
+++ b/sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/AccountInfoDdgWrapRepairer.kt
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.sync.impl
+
+import androidx.annotation.WorkerThread
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.sync.crypto.SyncLib
+import com.duckduckgo.sync.impl.Result.Error
+import com.duckduckgo.sync.impl.Result.Success
+import com.duckduckgo.sync.impl.pixels.SyncPixels
+import com.duckduckgo.sync.impl.pixels.UnifiedDeviceListPixel
+import com.duckduckgo.sync.impl.pixels.toAccountInfoKeyWrapFailureReason
+import com.duckduckgo.sync.store.SyncStore
+import com.squareup.anvil.annotations.ContributesBinding
+import dagger.SingleInstanceIn
+import logcat.LogPriority.ERROR
+import logcat.logcat
+import javax.inject.Inject
+
+@WorkerThread
+interface AccountInfoDdgWrapRepairer {
+    /**
+     * Adds a `ddg` wrap for the existing `account_info` key identified by [kid].
+     *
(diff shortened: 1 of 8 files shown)


sync/sync-impl/src/main/java/com/duckduckgo/sync/impl/AccountInfoDdgWrapRepairer.kt
```diff
@@ -0,0 +1,155 @@
...
+     */
+    fun repair(
+        kid: String,
+        entries: List<ProtectedKeyEntry>? = null,
+    ): Result<Unit>
+}
+
+@SingleInstanceIn(AppScope::class)
+@ContributesBinding(AppScope::class)
+class RealAccountInfoDdgWrapRepairer @Inject constructor(
```
CDRussell, line 49: This is logic mostly already existing in `LoginDeviceInfoWriter` before, moved to its own class because it's now used outside of login scenario too.

MiSikora (approved):

PR #9770, opened 2026-09-10
Bring the app rating prompt into the modal coordinator

aibrahim-:
Task/Issue URL:   https://app.asana.com/1/137249556945/project/72649045549333/task/1218278045050709?focus=true

### Description
Integrate the rate-our-app prompt into the prompt coordinator, use a feature flag default ON to switch back to old behaviour.
### Steps to test this PR

The following command will modify the code so that only one DDG search is needed to meet the thresholds.
```
sed -i '' 's/^const val MINIMUM_SEARCHES_THRESHOLD = 5$/const val MINIMUM_SEARCHES_THRESHOLD = 1/; s/^const val MINIMUM_DAYS_USAGE_BEFORE_FIRST_PROMPT = 3$/const val MINIMUM_DAYS_USAGE_BEFORE_FIRST_PROMPT = 0/' app/src/main/java/com/duckduckgo/app/global/rating/AppEnjoymentPromptCriteria.kt
```

Then build, and install. 

Perform one search and observe the dialog showing. 

Reset app add data, change flag value to False, then repeat.

---

> [!NOTE]
> **Medium Risk**
> Changes when and how the rating prompt appears (NTP vs app start) and modal arbitration order; mitigated by a default-on remote flag that restores legacy behavior when disabled.
> 
> **Overview**
> Routes the **app enjoyment / rating prompt** through the **prompts coordinator** instead of firing unarbitrated on every app start.
> 
> When remote flag `appRatingPromptModal` is **on** (default), new `AppRatingPromptEvaluator` decides whether to show the prompt on **`NTP_RENDER`** at coordinator priority **6**, with a short delay before emitting via the existing `AppEnjoymentPromptEmitter`. `AppEnjoymentAppCreationObserver` **no-ops** on start so the prompt is not evaluated twice or outside arbitration.
> 
> When the flag is **off**, behavior reverts to the observer’s app-start path. **`CookiePopupOptInEvaluator`** priority moves **6 → 7** so the rating prompt can win ordering against the cookie opt-in modal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ced1846c6c726e81e773982760f2651d6fe6a5ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
+++ b/app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2026 DuckDuckGo
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.duckduckgo.app.global.rating
+
+import com.duckduckgo.common.utils.DispatcherProvider
+import com.duckduckgo.di.scopes.AppScope
+import com.duckduckgo.promptscoordinator.api.ModalEvaluator
+import com.duckduckgo.promptscoordinator.api.ModalTrigger
+import com.squareup.anvil.annotations.ContributesMultibinding
+import dagger.SingleInstanceIn
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
+import logcat.logcat
+import javax.inject.Inject
+
+/**
+ * Brings the app enjoyment / rating prompt under the prompts coordinator so it stops competing with
+ * the other modals. Gated by [AppRatingPromptModalFeature]: while that flag is off this evaluator
+ * always skips and [AppEnjoymentAppCreationObserver] retains the old app-start behaviour.
+ *
+ * Uses [ModalTrigger.NTP_RENDER] so the prompt is only offered while the browser is on a clean New
+ * Tab Page. The dialog is hosted by BrowserActivity, and an [ModalTrigger.APP_RESUME] pass can land
+ * on another screen, where the dialog is silently dropped after the coordinator has already recorded
+ * the modal as shown.
+ */
+@ContributesMultibinding(
+    scope = AppScope::class,
+    boundType = ModalEvaluator::class,
+)
+@SingleInstanceIn(scope = AppScope::class)
+class AppRatingPromptEvaluator @Inject constructor(
+    private val promptTypeDecider: PromptTypeDecider,
+    private val appEnjoymentPromptEmitter: AppEnjoymentPromptEmitter,
+    private val appRatingPromptModalFeature: AppRatingPromptModalFeature,
+    private val dispatchers: DispatcherProvider,
+) : ModalEvaluator {
+
+    override val priority: Int = PRIORITY
+    override val evaluatorId: String = "app_rating_prompt"
+    override val trigger: ModalTrigger = ModalTrigger.NTP_RENDER
+
+    override suspend fun evaluate(): ModalEvaluator.EvaluationResult = withContext(dispatchers.io()) {
+        if (!appRatingPromptModalFeature.self().isEnabled()) {
+            logcat { "AppRatingPromptEvaluator: skipped, feature disabled so the app-start observer owns the prompt" }
+            return@withContext ModalEvaluator.EvaluationResult.Skipped
+        }
+
+        val promptType = promptTypeDecider.determineInitialPromptType()
+        if (promptType == AppEnjoymentPromptOptions.ShowNothing) {
+            logcat { "AppRatingPromptEvaluator: skipped, decider selected no prompt" }
+            return@withContext ModalEvaluator.EvaluationResult.Skipped
+        }
+
+        delay(MODAL_DISPLAY_DELAY)
+        withContext(dispatchers.main()) {
+            appEnjoymentPromptEmitter.promptType.value = promptType
+        }
+        logcat { "AppRatingPromptEvaluator: emitted $promptType" }
+        ModalEvaluator.EvaluationResult.ModalShown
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
```diff
@@ -0,0 +1,78 @@
...
+            logcat { "AppRatingPromptEvaluator: skipped, decider selected no prompt" }
+            return@withContext ModalEvaluator.EvaluationResult.Skipped
+        }
+
+        delay(MODAL_DISPLAY_DELAY)
+        withContext(dispatchers.main()) {
+            appEnjoymentPromptEmitter.promptType.value = promptType
+        }
+        logcat { "AppRatingPromptEvaluator: emitted $promptType" }
+        ModalEvaluator.EvaluationResult.ModalShown
```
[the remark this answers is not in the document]
aibrahim-: This is a rare scenario, and it will self-heal on the next APP_RESUME after 24h, which is fine since we want to reduce the frequency of those popups anyway.

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptModalFeature.kt
```diff
@@ -0,0 +1,41 @@
...
+ * deciding the prompt type on every app start, unarbitrated.
+ *
+ * Exactly one of the two paths runs, so the prompt can never be evaluated twice in a single start.
+ */
+@ContributesRemoteFeature(
+    scope = AppScope::class,
+    featureName = "appRatingPromptModal",
+)
+interface AppRatingPromptModalFeature {
+    @Toggle.DefaultValue(DefaultFeatureValue.TRUE)
```
catalinradoiu, line 39: Should we have the default value false for now and have this enabled internally by default? And then when we want to release it we can update the value in the privacy config. Or is this true only for testing?

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
```diff
@@ -0,0 +1,80 @@
...
+    override val evaluatorId: String = "app_rating_prompt"
+    override val trigger: ModalTrigger = ModalTrigger.NTP_RENDER
+
+    override suspend fun evaluate(): ModalEvaluator.EvaluationResult = withContext(dispatchers.io()) {
+        if (!appRatingPromptModalFeature.self().isEnabled()) {
+            logcat { "AppRatingPromptEvaluator: skipped, feature disabled so the app-start observer owns the prompt" }
+            return@withContext ModalEvaluator.EvaluationResult.Skipped
+        }
+
+        val promptType = promptTypeDecider.determineInitialPromptType()
```
catalinradoiu, line 62: I see that in the `AppEnjoymentAppCreationObserver` we are also checking for the value of the `preventDialogQueuingFeature`. Do we need that here as well? Or what was the purpose of it?

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
```diff
@@ -0,0 +1,80 @@
...
+        delay(MODAL_DISPLAY_DELAY)
+        withContext(dispatchers.main()) {
+            appEnjoymentPromptEmitter.promptType.value = promptType
+        }
+        logcat { "AppRatingPromptEvaluator: emitted $promptType" }
+        ModalEvaluator.EvaluationResult.ModalShown
+    }
+
+    companion object {
+        private const val PRIORITY = 6
```
catalinradoiu, line 77: We are fine with having this after the other modals/ prompts, right?

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
```diff
@@ -0,0 +1,80 @@
...
+class AppRatingPromptEvaluator @Inject constructor(
+    private val promptTypeDecider: PromptTypeDecider,
+    private val appEnjoymentPromptEmitter: AppEnjoymentPromptEmitter,
+    private val appRatingPromptModalFeature: AppRatingPromptModalFeature,
+    private val dispatchers: DispatcherProvider,
+) : ModalEvaluator {
+
+    override val priority: Int = PRIORITY
+    override val evaluatorId: String = "app_rating_prompt"
+    override val trigger: ModalTrigger = ModalTrigger.NTP_RENDER
```
catalinradoiu, line 54: Before this change the prompt was evaluated on app start. Is there a reason why we have this only for NTP render? The valus that is being set is observer in the browser tab view model.

catalinradoiu (commented):
I tested the changes in this PR and the functionality works as expected. Just some clarifying questions but other than this it looks good.

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
aibrahim-: Yeah it's up for discussion, I raised it here: https://app.asana.com/1/137249556945/task/1217653553813424/comment/1212839176824586?focus=true 

I think this is more appropriate than app start, and I don't think it will affect the frequency of this prompt. It might be that at the time of creating this prompt, that was the best option they had, but that's not the case now.

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
aibrahim-: Yes, it was covered in the PFR.

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptModalFeature.kt
aibrahim-: I think we can just enable and disable if something goes wrong, there's no reason to wait, and internal testing won't give us much value as it mainly affects new user over a long period of time.

app/src/main/java/com/duckduckgo/app/global/rating/AppRatingPromptEvaluator.kt
aibrahim-: we need the prompt type to trigger the prompt. 
`appEnjoymentPromptEmitter.promptType.value = promptType` 

The preventDialogQueuingFeature is redundant, and no-op here, but we need to keep since we have a Feature Flag to revert back.

catalinradoiu (approved):
Looks good and works as expected. All the other comments have been clarified.

PR #9799, opened 2026-09-17
Move tracker-detection models into tracker-detection-api

aibrahim-:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216988666064181
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/task/1216706645739239
API Proposals URL(s) (if applicable): to follow — will be added before merge

### Description

First PR of the tracker-detection extraction (Phase 1 of 5). Pure move, no behaviour change.

- `Entity`, `TrackingEvent`, `TrackerStatus` and `TrackerType` move from `browser-api` into `tracker-detection-api`, keeping their legacy `com.duckduckgo.app.trackerdetection.model` package so **no consumer file changes its imports**. A package rename is a possible follow-up.
- `browser-api` re-exports the module with `api project(':tracker-detection-api')`. This keeps the existing consumers (`:app`, `privacy-dashboard-impl`, `pir-impl`) compiling without adding the dependency to each of them.
- `tracker-detection-api` drops its `browser-api` dependency, since the models it references are now local.

**Please note (DL1 from the tech design review):** the root `build.gradle` rule that forbids `-api` modules from depending on other `-api` modules needed `:tracker-detection-api` added to its allowlist, alongside `:browser-api` and the other shared api modules. That is what makes the re-export above legal. It seems reasonable given these models are used in several places across the app, but it is a deliberate widening of that allowlist and worth a look.

### Steps to test this PR

_Tracker blocking is unaffected by the move_
- [ ] Build and run the app
- [ ] Visit a site with trackers (e.g. a news site) and open the privacy dashboard
- [ ] Confirm the blocked trackers and their company names are listed as before

### UI changes
| Before  | After |
| ------ | ----- |
No UI changes|No UI changes|


build.gradle
```diff
@@ -178,6 +178,7 @@ subprojects {
                                 && dependencyPath != ":js-messaging-api"
                                 && dependencyPath != ":browser-api"
                                 && dependencyPath != ":browser-mode-api"
+                                && dependencyPath != ":tracker-detection-api"
```
landomen, line 181: I think this is acceptable to do considering this are core entities / classes.

landomen (approved):

PR #9803, opened 2026-09-17
Replace WebTrackersBlockedRepository with a narrow api interface

aibrahim-:
Task/Issue URL: https://app.asana.com/1/137249556945/task/1216988837423281
Tech Design URL (if applicable): https://app.asana.com/1/137249556945/task/1216706645739239
API Proposals URL(s) (if applicable): to follow — will be added before merge

### Description

Phase 2 of the tracker-detection extraction, and the PR that answers **DL2** from the tech design review ("repositories should not be part of the public API — could we expose something narrower?"). No behaviour change.

Consumers reached the blocked-tracker log three different ways:

- the fire button, through `WebTrackersBlockedRepository` in `app-store`
- the tab switcher tile and its dismissal pixel, through the *concrete* `WebTrackersBlockedAppRepository`
- `WebViewRequestInterceptor` and `TrackersDbCleanerWorker`, straight through `WebTrackersBlockedDao`

`WebTrackersBlockedHistory` in `tracker-detection-api` replaces all three with the four operations that actually cross the boundary: `onTrackerBlocked`, `trackerCountForLastWeek`, `deleteAll` and `deleteEntriesOlderThan`. The Room entity, the DAO and the timestamp format stay behind it, and `WebTrackersBlockedAppRepository` is now only its implementation.

Two small deletions worth flagging:

- `WebTrackersBlockedRepository.get()` had **no callers**, so it is dropped rather than carried over into the new interface.
- `app-store`'s internal `DatabaseDateFormatter` went with it, leaving the shared `common-utils` one as the only copy.

### Steps to test this PR

_Tab switcher tile still counts blocked trackers_
- [ ] Build and run the app, browse a few tracker-heavy sites
- [ ] Open the tab switcher and confirm the "trackers blocked" info panel shows a plausible count
- [ ] Dismiss the panel and confirm its dismissal pixel still fires with a bucketed count

_Fire button still clears the log_
- [ ] Press the fire button to clear browsing data
- [ ] Reopen the tab switcher and confirm the tracker count has reset

### UI changes
| Before  | After |
| ------ | ----- |
No UI changes|No UI changes|


landomen (approved):
LGTM, let's do the API proposal before merging

